PR #2518, opened 2023-01-17
feat(shared): add layout dto to shared and relocate shared dependencies

p-fernandez:

Adds LayoutDto to shared types and reorders types and imports, mainly from `IEmailBlock` and `ITemplateVariable` types that are reused across different features.

Besides adding LayoutDto to shared package to be able to be used in the Web app for typing, I relocated many of the shared dependencies in a way that makes more sense to me and that helps to replace the imports done to `@novu/dal` that is a repository package. Therefore we avoid tight dependencies in some cases like Web app was importing the DAL package.


libs/dal/src/repositories/message-template/types.ts
```diff
@@ -1,17 +1,7 @@
...
-    textAlign?: TextAlignEnum;
-  };
-}
-
-export interface ITemplateVariable {
-  type: TemplateVariableTypeEnum;
-  name: string;
-  required: boolean;
-  defaultValue?: string | boolean;
-}
```
p-fernandez, line 17: This types are used by different entities so I moved them out to the types folder for easier sharing and importing, so there weren't dependencies of the message template entity around.

libs/shared/src/types/message-template/index.ts
```diff
@@ -8,3 +10,19 @@ export enum TextAlignEnum {
...
+    textAlign?: TextAlignEnum;
+  };
+}
+
+export interface ITemplateVariable {
+  type: TemplateVariableTypeEnum;
+  name: string;
+  required: boolean;
+  defaultValue?: string | boolean;
+}
```
p-fernandez, line 28: This is where I found more appropriate for the interfaces to live so their importing is more tidied.

ainouzgali (approved):
Looks good

PR #2520, opened 2023-01-17
feat(api): promote changes of layouts

p-fernandez:

Adds use case to be able to promote existing changes between environments in layouts.

To keep aligned our users' environments after they create and test their email notification layouts.

### DIFF
--- a/apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
+++ b/apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
@@ -10,23 +10,40 @@ export class PromoteLayoutChange {
   async execute(command: PromoteTypeChangeCommand) {
     const item = await this.layoutRepository.findOne({
       _environmentId: command.environmentId,
-      _id: command.item._id,
+      _parentId: command.item._id,
     });
 
     const newItem = command.item as LayoutEntity;
 
     if (!item) {
       const layoutEntity = {
-        ...newItem,
+        name: newItem.name,
+        content: newItem.content,
+        contentType: newItem.contentType,
+        variables: newItem.variables,
+        isDefault: newItem.isDefault,
+        channel: newItem.channel,
         _creatorId: command.userId,
         _environmentId: command.environmentId,
         _organizationId: command.organizationId,
+        _parentId: newItem._id,
       };
 
-      return this.layoutRepository.createLayout(layoutEntity);
+      return await this.layoutRepository.create(layoutEntity);
     }
 
-    return this.layoutRepository.update(
+    const count = await this.layoutRepository.count({
+      _organizationId: command.organizationId,
+      _id: command.item._id,
+    });
+
+    if (count === 0) {
+      await this.layoutRepository.deleteLayout(item._id, command.environmentId, command.organizationId);
+
+      return;
+    }
+
+    return await this.layoutRepository.update(
       {
         _environmentId: command.environmentId,
         _id: item._id,
@@ -36,6 +53,8 @@ export class PromoteLayoutChange {
         content: newItem.content,
         contentType: newItem.contentType,
         variables: newItem.variables,
+        _environmentId: command.environmentId,
+        _organizationId: command.organizationId,
       }
     );
   }
--- a/apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
+++ b/apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
@@ -1,0 +1,42 @@
+import { Injectable } from '@nestjs/common';
+import { LayoutEntity, LayoutRepository } from '@novu/dal';
+
+import { PromoteTypeChangeCommand } from '../promote-type-change.command';
+
+@Injectable()
+export class PromoteLayoutChange {
+  constructor(private layoutRepository: LayoutRepository) {}
+
+  async execute(command: PromoteTypeChangeCommand) {
+    const item = await this.layoutRepository.findOne({
+      _environmentId: command.environmentId,
+      _id: command.item._id,
+    });
+
+    const newItem = command.item as LayoutEntity;
+
+    if (!item) {
+      const layoutEntity = {
+        ...newItem,
+        _creatorId: command.userId,
+        _environmentId: command.environmentId,
+        _organizationId: command.organizationId,
+      };
+
+      return this.layoutRepository.createLayout(layoutEntity);
+    }
+
+    return this.layoutRepository.update(
+      {
+        _environmentId: command.environmentId,
+        _id: item._id,
+      },
+      {
+        name: newItem.name,
+        content: newItem.content,
+        contentType: newItem.contentType,
+        variables: newItem.variables,
+      }
+    );
+  }
+}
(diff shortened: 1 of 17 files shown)


apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.ts
```diff
@@ -0,0 +1,42 @@
...
+import { PromoteTypeChangeCommand } from '../promote-type-change.command';
+
+@Injectable()
+export class PromoteLayoutChange {
+  constructor(private layoutRepository: LayoutRepository) {}
+
+  async execute(command: PromoteTypeChangeCommand) {
+    const item = await this.layoutRepository.findOne({
+      _environmentId: LayoutRepository.convertStringToObjectId(command.environmentId),
+      _id: command.item._id,
```
ainouzgali, line 13: To me it seems that it should be searched with _parentId instead of _id. So it checks if a layout that has already been promoted exists. 
_parentId is the id of the item in a previous environment (dev)

apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    const newItem = command.item as LayoutEntity;
+
+    if (!item) {
+      const layoutEntity = {
+        ...newItem,
```
ainouzgali, line 20: Same here, if there isn't an item that has already been promoted from this item id, one should be created with:
`_parentId:  newItem._id`

libs/shared/src/entities/change/change.interface.ts
```diff
@@ -1,6 +1,7 @@
 export enum ChangeEntityTypeEnum {
-  NOTIFICATION_TEMPLATE = 'NotificationTemplate',
+  FEED = 'FEED',
```
ainouzgali, line 2:
```suggestion
  FEED = 'Feed',
```

ainouzgali (commented):
⭐️
Let me know what you think of my suggestion

libs/shared/src/entities/change/change.interface.ts
p-fernandez: I was meant to write it like your suggestion. Don't know what happened. 🤦🏻

apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.ts
```diff
@@ -0,0 +1,42 @@
+import { Injectable } from '@nestjs/common';
+import { LayoutEntity, LayoutRepository } from '@novu/dal';
+
+import { PromoteTypeChangeCommand } from '../promote-type-change.command';
+
+@Injectable()
+export class PromoteLayoutChange {
```
ainouzgali, line 7: We would also need to handle here a promote change that is delete layout.

p-fernandez:
Depending on https://github.com/novuhq/novu/pull/2521

apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
```diff
@@ -0,0 +1,42 @@
...
+import { PromoteTypeChangeCommand } from '../promote-type-change.command';
+
+@Injectable()
+export class PromoteLayoutChange {
+  constructor(private layoutRepository: LayoutRepository) {}
+
+  async execute(command: PromoteTypeChangeCommand) {
+    const item = await this.layoutRepository.findOne({
+      _environmentId: command.environmentId,
+      _id: command.item._id,
```
ainouzgali, line 13: I think this should be _parentId: command.item._id,

apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    const newItem = command.item as LayoutEntity;
+
+    if (!item) {
+      const layoutEntity = {
+        ...newItem,
+        _creatorId: command.userId,
+        _environmentId: command.environmentId,
+        _organizationId: command.organizationId,
```
ainouzgali, line 23: And here you will be creating a layout with _parentId being the _id of the item you are promoting.

apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
```diff
@@ -33,5 +35,25 @@ export class DeleteLayoutUseCase {
...
+      environmentId: command.environmentId,
+      layoutId: command.layoutId,
+      organizationId: command.organizationId,
+      userId: command.userId,
+    });
+
+    await this.createChange(command);
+  }
+
+  private async createChange(command: DeleteLayoutCommand): Promise<void> {
```
ainouzgali, line 49: Why is this function needed? Doesn't it do the same thing by creating the command? 
Maybe you meant to remove the first CreateLayoutChangeCommand in line 39?

ainouzgali (approved):
⭐️
Just please take a look at my _parent id suggestions. And let me know what you think :)

apps/api/src/app/change/usecases/promote-layout-change/promote-layout-change.use-case.ts
```diff
@@ -10,23 +10,40 @@ export class PromoteLayoutChange {
...
-    return this.layoutRepository.update(
+    const count = await this.layoutRepository.count({
+      _organizationId: command.organizationId,
+      _id: command.item._id,
+    });
+
+    if (count === 0) {
+      await this.layoutRepository.deleteLayout(item._id, command.environmentId, command.organizationId);
+
+      return;
```
p-fernandez, line 43: You can merge both lines.

p-fernandez, line 43: You can merge both lines.

apps/api/src/app/change/usecases/promote-message-template-change/promote-message-template-change.ts
```diff
@@ -24,6 +28,12 @@ export class PromoteMessageTemplateChange {
       identifier: feedDev?.identifier,
     });
 
+    const layout = await this.layoutRepository.findOne({
+      _environmentId: command.environmentId,
+      _organizationId: command.organizationId,
+      _parentId: newItem._layoutId,
+    });
```
p-fernandez, line 35: Should we check what happens if there is not layout found?

apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
```diff
@@ -21,7 +19,11 @@ export class CreateLayoutChangeUseCase {
   async execute(command: CreateLayoutChangeCommand, isDeleteChange = false): Promise<void> {
     const item = isDeleteChange
       ? await this.findDeletedLayout.execute(FindDeletedLayoutCommand.create(command))
-      : await this.layoutRepository.findById(command.layoutId, command.environmentId);
+      : await this.layoutRepository.findOne({
+          _id: command.layoutId,
+          _environmentId: command.environmentId,
+          _organizationId: command.organizationId,
```
p-fernandez, line 25: Good catch.

libs/dal/src/repositories/layout/layout.schema.ts
```diff
@@ -20,6 +20,10 @@ const layoutSchema = new Schema(
       type: Schema.Types.ObjectId,
       ref: 'User',
     },
+    _parentId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Layout',
```
p-fernandez, line 25: We have to refer to the schema itself?

p-fernandez (commented):
🌟

PR #2521, opened 2023-01-18
feat(api): create a layout change

p-fernandez:

Adds functionality for creating a layout change

To be able to promote the layouts for the production environment, first we need to create said change in order to be able to apply the changes in production, either individually or in bulk.

### DIFF
--- a/apps/api/src/app/change/usecases/create-change/index.ts
+++ b/apps/api/src/app/change/usecases/create-change/index.ts
@@ -1,0 +1,2 @@
+export * from './create-change.command';
+export * from './create-change.usecase';
--- a/apps/api/src/app/change/usecases/promote-change-to-environment/promote-change-to-environment.usecase.ts
+++ b/apps/api/src/app/change/usecases/promote-change-to-environment/promote-change-to-environment.usecase.ts
@@ -1,9 +1,9 @@
-import { Injectable, Logger, NotFoundException } from '@nestjs/common';
+import { forwardRef, Inject, Injectable, Logger, NotFoundException } from '@nestjs/common';
 import { ChangeRepository, EnvironmentRepository } from '@novu/dal';
 import { ChangeEntityTypeEnum } from '@novu/shared';
 import { PromoteChangeToEnvironmentCommand } from './promote-change-to-environment.command';
 import { PromoteTypeChangeCommand } from '../promote-type-change.command';
-import { PromoteNotificationTemplateChange } from '../promote-notification-template-change/promote-notification-template-change';
+import { PromoteNotificationTemplateChange } from '../promote-notification-template-change';
 import { PromoteMessageTemplateChange } from '../promote-message-template-change/promote-message-template-change';
 import { PromoteNotificationGroupChange } from '../promote-notification-group-change/promote-notification-group-change';
 import { applyDiff } from 'recursive-diff';
@@ -14,6 +14,7 @@ export class PromoteChangeToEnvironment {
   constructor(
     private changeRepository: ChangeRepository,
     private environmentRepository: EnvironmentRepository,
+    @Inject(forwardRef(() => PromoteNotificationTemplateChange))
     private promoteNotificationTemplateChange: PromoteNotificationTemplateChange,
     private promoteMessageTemplateChange: PromoteMessageTemplateChange,
     private promoteNotificationGroupChange: PromoteNotificationGroupChange,
--- a/apps/api/src/app/invites/e2e/bulk-invite.e2e.ts
+++ b/apps/api/src/app/invites/e2e/bulk-invite.e2e.ts
@@ -28,7 +28,7 @@ describe('Bulk invite members - /invites/bulk (POST)', async () => {
       .send({
         invitees: [
           {
-            email: 'asdasda',
+            email: 'email@bad',
             role: 'admin',
           },
         ],
--- a/apps/api/src/app/layouts/layouts.module.ts
+++ b/apps/api/src/app/layouts/layouts.module.ts
@@ -1,13 +1,14 @@
-import { Module } from '@nestjs/common';
+import { forwardRef, Module } from '@nestjs/common';
 
 import { USE_CASES } from './usecases';
 import { LayoutsController } from './layouts.controller';
 
+import { ChangeModule } from '../change/change.module';
 import { MessageTemplateModule } from '../message-template/message-template.module';
 import { SharedModule } from '../shared/shared.module';
 
 @Module({
-  imports: [SharedModule, MessageTemplateModule],
+  imports: [SharedModule, ChangeModule, MessageTemplateModule],
   providers: [...USE_CASES],
   exports: [...USE_CASES],
   controllers: [LayoutsController],
--- a/apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
@@ -1,0 +1,33 @@
+import { LayoutEntity, LayoutRepository } from '@novu/dal';
+import { ChangeEntityTypeEnum } from '@novu/shared';
+import { Injectable } from '@nestjs/common';
+
+import { CreateLayoutChangeCommand } from './create-layout-change.command';
+
+import { LayoutDto } from '../../dtos/layout.dto';
+import { ChannelTypeEnum, IEmailBlock } from '../../types';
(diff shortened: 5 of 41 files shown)


apps/api/src/app/change/usecases/create-change/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './create-change.command';
+export * from './create-change.usecase';
```
p-fernandez, line 2: Following the same pattern of the use case we apply everywhere, I moved to dedicated folder to easier export.

apps/api/src/app/layouts/use-cases/create-layout-change/create-layout-change.use-case.ts
```diff
@@ -0,0 +1,33 @@
...
+          organizationId: command.organizationId,
+          environmentId: command.environmentId,
+          userId: command.userId,
+          type: ChangeEntityTypeEnum.LAYOUT,
+          item,
+          changeId,
+        })
+      );
+    }
+  }
```
p-fernandez, line 32: To review. I assume we don't need any parentChangeId as the layout changes are not based on another change.

apps/api/src/app/layouts/use-cases/create-layout-change/create-layout-change.use-case.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Injectable } from '@nestjs/common';
+
+import { CreateLayoutChangeCommand } from './create-layout-change.command';
+
+import { LayoutDto } from '../../dtos/layout.dto';
+import { ChannelTypeEnum, IEmailBlock } from '../../types';
+import { CreateChange, CreateChangeCommand } from '../../../change/usecases';
+
+@Injectable()
+export class CreateLayoutChangeUseCase {
```
p-fernandez, line 12: Dedicate use case for layout so it can be used in the needed use cases of the layout feature.

apps/api/src/app/change/usecases/create-change/index.ts
ainouzgali: Nice!

libs/shared/src/entities/change/change.interface.ts
```diff
@@ -1,6 +1,7 @@
 export enum ChangeEntityTypeEnum {
```
ainouzgali, line 1: Its also in the first pr. Notice for conflicts

apps/api/src/app/layouts/use-cases/create-layout-change/create-layout-change.use-case.ts
ainouzgali: I agree :)

ainouzgali (approved):
Looks good.

apps/api/src/app/change/usecases/promote-change-to-environment/promote-change-to-environment.usecase.ts
```diff
@@ -14,6 +14,7 @@ export class PromoteChangeToEnvironment {
   constructor(
     private changeRepository: ChangeRepository,
     private environmentRepository: EnvironmentRepository,
+    @Inject(forwardRef(() => PromoteNotificationTemplateChange))
```
p-fernandez, line 17: There is a circular dependency for `PromoteNotificationTemplateChange`. It tells me that we probably we would need to rethink some of the dependencies between the `ChangeModule`, `LayoutModule`and `NotificationTemplateModule`.

apps/api/src/app/invites/e2e/bulk-invite.e2e.ts
```diff
@@ -28,7 +28,7 @@ describe('Bulk invite members - /invites/bulk (POST)', async () => {
       .send({
         invitees: [
           {
-            email: 'asdasda',
+            email: 'email@bad',
```
p-fernandez, line 31: To avoid CSpell to complain.

apps/api/src/app/layouts/layouts.module.ts
```diff
@@ -1,13 +1,14 @@
...
 import { USE_CASES } from './usecases';
 import { LayoutsController } from './layouts.controller';
 
+import { ChangeModule } from '../change/change.module';
 import { MessageTemplateModule } from '../message-template/message-template.module';
 import { SharedModule } from '../shared/shared.module';
 
 @Module({
-  imports: [SharedModule, MessageTemplateModule],
+  imports: [SharedModule, ChangeModule, MessageTemplateModule],
```
p-fernandez, line 11: We need `CreateChange` use case inside Layout module for being able to create all the changes through `CreateLayoutChange`.

apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
```diff
@@ -0,0 +1,33 @@
...
+import { LayoutDto } from '../../dtos/layout.dto';
+import { ChannelTypeEnum, IEmailBlock } from '../../types';
+import { CreateChange, CreateChangeCommand } from '../../../change/usecases';
+
+@Injectable()
+export class CreateLayoutChangeUseCase {
+  constructor(private createChange: CreateChange, private layoutRepository: LayoutRepository) {}
+
+  async execute(command: CreateLayoutChangeCommand): Promise<void> {
+    const item = await this.layoutRepository.findById(command.layoutId, command.environmentId);
```
p-fernandez, line 16: Generic use case to be reused in all the Layout use cases where we need to track a change.

apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
```diff
@@ -19,6 +21,7 @@ export class SetDefaultLayoutUseCase {
     try {
       if (defaultLayoutId) {
         await this.setIsDefaultForLayout(defaultLayoutId, command.environmentId, command.organizationId, false);
+        await this.createLayoutChangeForPreviousDefault(command, defaultLayoutId);
```
p-fernandez, line 24: As we make non default the default layout, we need to keep track of that change. Not totally happy as this could be easily overlooked in a change of logic. Worth discuss it.

p-fernandez:
Tests will go in the promote layout changes PR.

apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
```diff
@@ -33,5 +35,25 @@ export class DeleteLayoutUseCase {
...
+      environmentId: command.environmentId,
+      layoutId: command.layoutId,
+      organizationId: command.organizationId,
+      userId: command.userId,
+    });
+
+    await this.createChange(command);
+  }
+
+  private async createChange(command: DeleteLayoutCommand): Promise<void> {
```
ainouzgali, line 49: Why is this function needed? Doesn't it do the same thing by creating the command?
Maybe you meant to remove the first CreateLayoutChangeCommand in line 39?

ainouzgali (approved):

apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
p-fernandez: It just holds the create layout change logic to not clutter the execute method. Nothing else. 🙂

apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
ainouzgali: In the case of create change after delete-layout, this will not find the deleted one

p-fernandez: How do we handle deletions? With the soft delete should be able to pick it up. Don't know how mongo-delete works 100%.

ainouzgali: I made a suggestion in this [pr](https://github.com/novuhq/novu/pull/2574) for a find soft deleted layouts. Let me know what you think :)

p-fernandez: So the change creation for a deletion would need to use that new method?

ainouzgali: Maybe if he doesn't find the item like this, then do the findDeleted

ainouzgali (approved):

PR #2522, opened 2023-01-18
feat: add limit bar and analytics

davidsoderberg:
Screenshot 2023-01-18 at 09 34 17
Screenshot 2023-01-18 at 09 34 24


apps/web/src/components/layout/components/HeaderNav.tsx
```diff
@@ -13,6 +13,7 @@ import { Tooltip } from '../../../design-system';
 import { INTERCOM_APP_ID } from '../../../config';
 import { SpotlightContext } from '../../../store/spotlightContext';
 import { HEADER_HEIGHT } from '../constants';
+import { LimitBar } from '../../../pages/integrations/components/LImitBar';
```
p-fernandez, line 16:
```suggestion
import { LimitBar } from '../../../pages/integrations/components/LimitBar';
```
Filename spelling. 🙂

apps/web/src/components/templates/BlueprintModal.tsx
```diff
@@ -8,17 +8,34 @@ import { When } from '../utils/When';
...
 
+  const { mutateAsync: updateOnBoardingStatus } = useMutation<
+    IUserEntity,
+    { error: string; message: string; statusCode: number },
+    { showOnBoarding: boolean }
+  >(({ showOnBoarding }) => updateUserOnBoarding(showOnBoarding));
+
+  async function disableOnboarding() {
+    await updateOnBoardingStatus({ showOnBoarding: false });
+  }
```
p-fernandez, line 37: We should be coherent with the spelling to not be confusing.
Checking the right spelling in the dictionary I think we should change to `updateUserOnboarding`, `updateOnboardingStatus` and `showOnboarding`.

p-fernandez (approved):
Very helpful the screenshots to understand what's going on for this feature.

apps/web/src/components/templates/BlueprintModal.tsx
davidsoderberg: .... copied it from another part of our code and I think the object property needs to be `showOnBoarding`

p-fernandez: Ouch, I have just seen that is a schema property also. That one we shouldn't touch it. Damn. 😞

apps/web/src/components/templates/BlueprintModal.tsx
```diff
@@ -39,6 +56,7 @@ export function BlueprintModal() {
     onSuccess: (template) => {
       localStorage.removeItem('blueprintId');
       if (template) {
+        disableOnboarding();
```
LetItRock, line 59: Why we are disabling the onboarding? I feel like for the new accounts they won't still know how to do integrations, create the template, use trigger etc.

PR #2527, opened 2023-01-18
Layouts - compile layout with email content 

ainouzgali:

Added functionality of compiling layout + email content inside of layout body. 
Still missing tests for the new functionality.

I'm so sorry about the huge pr. I'll try to split it tomorrow.


apps/api/src/app/content-templates/usecases/compile-template/email/compile-email-template.usecase.ts
```diff
@@ -0,0 +1,152 @@
...
+      } catch (e) {
+        useNovuDefault = true;
+      }
+    }
+
+    const isEditorMode = command.contentType === 'editor';
+    let subject = '';
+    let content: string | IEmailBlock[] = command.content;
+    let preheader = command.preheader;
+    const layoutContent = useNovuDefault ? defaultLayout : layout.content;
```
LetItRock, line 57: this variable should be defined before `if (!useNovuDefault)` with a default set to `defaultLayout`, so it will be a string type and you won't need to cast it to `string` later... like:
```
let layoutContent = defaultLayout;

if (!command.layoutId) {
...
layout = await this.getLayoutUsecase.execute(...
layoutContent = layout.content;
}

```
then you won't also need `as LayoutId` and `layoutContent as string`

apps/api/src/app/content-templates/usecases/compile-template/email/compile-email-template.usecase.ts
```diff
@@ -0,0 +1,152 @@
...
+        content = await this.getContent(content, payload);
+      }
+      if (preheader) {
+        preheader = await this.renderContent(preheader, payload);
+      }
+    } catch (e) {
+      throw new ApiException(e?.message || `Message content could not be generated`);
+    }
+
+    const customTemplate = CompileEmailTemplateUsecase.addPreheader(layoutContent as string);
```
LetItRock, line 82: won't this preheader be included twice for the default layout?

apps/api/src/app/content-templates/content-templates.controller.ts
```diff
@@ -1,32 +1,33 @@
...
 
   @Post('/preview/email')
   public previewEmail(
     @UserSession() user: IJwtPayload,
     @Body('content') content: string | IEmailBlock[],
     @Body('contentType') contentType: MessageTemplateContentType,
     @Body('payload') payload: any,
-    @Body('subject') subject: string
+    @Body('subject') subject: string,
+    @Body('layoutId') layoutId: string
```
LetItRock, line 19: shouldn't this be optional or null?

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -209,57 +194,6 @@ export class SendMessageEmail extends SendMessageBase {
...
-        'syntax error in email editor',
-        command,
-        notification,
-        LogCodeEnum.SYNTAX_ERROR_IN_EMAIL_EDITOR
-      );
-
-      await this.createExecutionDetails.execute(
-        CreateExecutionDetailsCommand.create({
-          ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
-          detail: DetailEnum.MESSAGE_CONTENT_SYNTAX_ERROR,
```
LetItRock, line 250: should we remove this var: MESSAGE_CONTENT_SYNTAX_ERROR ?

LetItRock (approved):
left you a few small suggestions! but overall looks good! 🚀

apps/api/src/app/content-templates/usecases/compile-template/email/compile-email-template.command.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class CompileEmailTemplateCommand extends EnvironmentWithUserCommand {
+  @IsDefined()
+  content: string | IEmailBlock[];
+
+  @IsString()
+  contentType: MessageTemplateContentType;
+
+  @IsDefined()
+  payload: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
ainouzgali: Hmm yes, maybe. But that's how we do it everywhere at the moment. I actually basically just renamed the file. 
But I agree, I'm sure we can somehow better type payload and then do it consistently throughout

apps/api/src/app/content-templates/usecases/compile-template/email/compile-email-template.usecase.ts
```diff
@@ -0,0 +1,152 @@
...
+    payload: Record<string, unknown> = {}
+  ): Promise<string | IEmailBlock[]> {
+    content = [...content] as IEmailBlock[];
+    for (const block of content) {
+      block.content = await this.renderContent(block.content, payload);
+      block.url = await this.renderContent(block.url || '', payload);
+    }
+
+    return content;
+  }
```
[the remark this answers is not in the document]
ainouzgali: True. I changed the function use  to only be called with block array. Thanks, good catch

apps/api/src/app/content-templates/usecases/compile-template/email/compile-email-template.usecase.ts
ainouzgali: ❤️

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
ainouzgali: It was a bit redundant, as we already have it [here](https://github.com/novuhq/novu/blob/fd21da7a1a62a70aaba67a09c1a842417c330fe5/apps/api/src/app/events/usecases/send-message/send-message.base.ts#L49)

I now  generate the html inside `compileEmailTemplateUsecase`, any syntax error will be caught

apps/api/src/app/content-templates/usecases/compile-template/email/compile-email-template.usecase.ts
ainouzgali: Yes. But I am now changing the logic of where the default layout will be. So it will not be relevant anymore :) The novu default would not get here

PR #2531, opened 2023-01-19
Fix novu integration limit

davidsoderberg:
Screenshot 2023-01-18 at 17 08 49

### DIFF
--- a/apps/web/src/api/hooks/integrations/use-integration-limit.ts
+++ b/apps/web/src/api/hooks/integrations/use-integration-limit.ts
@@ -2,17 +2,22 @@ import { useQuery } from '@tanstack/react-query';
 import { ChannelTypeEnum } from '@novu/shared';
 import { getIntegrationLimit } from '../../integration';
 import { useActiveIntegrations } from '.';
+import { useMemo } from 'react';
 
 export function useIntegrationLimit(type: ChannelTypeEnum) {
   const selfHosted = process.env.REACT_APP_DOCKER_HOSTED_ENV === 'true';
   const { integrations = [] } = useActiveIntegrations();
+
+  const enabled = useMemo<boolean>(() => {
+    return selfHosted && integrations.filter((integration) => integration.channel === type).length === 0;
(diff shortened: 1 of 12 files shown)


davidsoderberg:
Will fix the tests

apps/web/src/api/hooks/integrations/use-integration-limit.ts
```diff
@@ -1,21 +1,25 @@
...
   const selfHosted = process.env.REACT_APP_DOCKER_HOSTED_ENV === 'true';
+  const { integrations = [] } = useActiveIntegrations();
 
   const {
     data = { limit: 0, count: 0 },
     isLoading,
     refetch,
   } = useQuery(['integrationLimit', type], () => getIntegrationLimit(type), {
-    enabled: selfHosted,
+    enabled: selfHosted && integrations.filter((integration) => integration.channel === type).length === 0,
```
p-fernandez, line 15: Worth extract this to a function or constant `isEnabled` as we reuse it in the file. 👍🏻

apps/web/src/components/templates/email-editor/EmailInboxContent.tsx
```diff
@@ -34,7 +35,7 @@ export const EmailInboxContent = ({
               margin: '5px 0px',
             }}
           >
-            {integration ? integration?.credentials?.from : 'No active email integration'}
+            <EmailIntegrationInfo integration={integration} field={'from'} />
```
p-fernandez, line 38: As the `field` default value is already `from` we wouldn't need to set it up explicitly. But I personally find it helpful, so this is ok for me.

p-fernandez (approved):

apps/web/src/components/templates/email-editor/EmailInboxContent.tsx
davidsoderberg: I know but I also find it helpful :)

LetItRock (approved):
🚀

p-fernandez (approved):
⚡

PR #2535, opened 2023-01-19
feat(api): sort and order query params for filtering layouts

p-fernandez:

Adds query params and functionality to sort the filtered results by creation date.

UI needs it. And probably users from API.


libs/dal/src/repositories/layout/layout.repository.ts
```diff
@@ -71,14 +71,18 @@ export class LayoutRepository extends BaseRepository<EnforceEnvironmentQuery, La
 
   async filterLayouts(
     query: EnforceEnvironmentQuery,
-    pagination: { limit: number; skip: number }
+    pagination: { limit: number; skip: number; sortBy: string; orderBy: OrderDirectionEnum }
   ): Promise<LayoutEntity[]> {
+    const order = pagination.orderBy ?? OrderDirectionEnum.ASC;
+    const sort = pagination.sortBy ? { [pagination.sortBy]: order } : { createdAt: 1 };
```
ainouzgali, line 77:
```suggestion
    const sort = pagination.sortBy ? { [pagination.sortBy]: order } : { createdAt: OrderDirectionEnum.DESC};
```

I think descending makes more sense as a default. So they will get first in the the last one they created. Wdyt?

ainouzgali (approved):
Yay 💪🏻

libs/dal/src/repositories/layout/layout.repository.ts
p-fernandez: Sounds good to me. We might create a task in the feature to add in the list the default first of all.

PR #2545, opened 2023-01-19
feat(api): disallow delete of assigned

p-fernandez:

Avoids that a layout is deleted if it is being used by a message template.

To avoid data corruption by users actions.


apps/api/src/app/layouts/use-cases/check-layout-is-used/check-layout-is-used.use-case.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Injectable } from '@nestjs/common';
+
+import { CheckLayoutIsUsedCommand } from './check-layout-is-used.command';
+import {
+  FindMessageTemplatesByLayoutCommand,
+  FindMessageTemplatesByLayoutUseCase,
+} from '../../../message-template/usecases';
+
+@Injectable()
+export class CheckLayoutIsUsedUseCase {
```
ainouzgali, line 11: I guess we could quite easily use this for the isUsed in the FE table we were planning to scope out  🤭
Maybe we could just do it as a query in the front or as an extra return field to the get-layout-usecase , instead of actually adding a field to the schema. Which I definitely don't like.  

Actually might have some value in also making it external API accessible.

apps/api/src/app/message-template/usecases/find-message-templates-by-layout/find-message-templates-by-layout.use-case.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { FindMessageTemplatesByLayoutCommand } from './find-message-templates-by-layout.command';
+
+const DEFAULT_PAGE_SIZE = 100;
+
+@Injectable()
+export class FindMessageTemplatesByLayoutUseCase {
+  constructor(private messageTemplateRepository: MessageTemplateRepository) {}
+
+  async execute(command: FindMessageTemplatesByLayoutCommand): Promise<MessageTemplateEntity[]> {
```
ainouzgali, line 12: I really like all these separations. You think ahead💪🏻

I see multiple possible future uses for this

ainouzgali (approved):
⭐️

apps/api/src/app/layouts/use-cases/check-layout-is-used/check-layout-is-used.use-case.ts
p-fernandez: I do agree that the `isUsed` calculation should be done on demand and not by DB flags that could go out of sync easily. And with the rest of your comments.

PR #2551, opened 2023-01-20
NV-1541: handle default novu layout

scopsy:

When creating a new environment add default layout to it.

Closes NV-1541 to make sure new organizations created are created with the default novu layout.


apps/api/nest-cli.json
```diff
@@ -20,11 +20,11 @@
...
       },
       {
-        "include": "app/content-templates/usecases/compile-template/templates/layout.handlebars",
+        "include": "app/layouts/usecases/get-novu-layout/templates/layout.handlebars",
         "outDir": "dist/src"
       },
       {
-        "include": "app/content-templates/usecases/compile-template/templates/basic.handlebars",
+        "include": "app/content-templates/usecases/compile-email-template/templates/basic.handlebars",
         "outDir": "dist/src"
```
scopsy, line 28: Refactored the locaiton of those to live closely to the relevant parts where they belong to

apps/api/src/app/content-templates/content-templates.module.ts
```diff
@@ -1,4 +1,4 @@
-import { Module } from '@nestjs/common';
+import { forwardRef, Module } from '@nestjs/common';
```
p-fernandez, line 1: This is not needed.

apps/api/src/app/layouts/usecases/get-novu-layout/get-novu-layout.usecase.ts
```diff
@@ -0,0 +1,30 @@
...
+    return template;
+  }
+
+  private async loadTemplateContent(name: string) {
+    return new Promise<string>((resolve, reject) => {
+      let path = '';
+      if (!process.env.E2E_RUNNER) {
+        path = '/src/app/layouts/usecases/get-novu-layout';
+      }
+      fs.readFile(`${__dirname}${path}/templates/${name}`, (err, content) => {
```
p-fernandez, line 21: We can keep the async/await patterns if we use the `readFile` method from the promises library of Node.js (available since v.10; now with a different import path).
`import { readFile } from 'node:fs/promises';`

apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
```diff
@@ -13,7 +13,7 @@ export class SetDefaultLayoutUseCase {
     const defaultLayoutId = await this.findDefaultLayoutId(command.environmentId, command.organizationId);
 
     try {
-      if (defaultLayoutId) {
+      if (defaultLayoutId && defaultLayoutId !== command.layoutId) {
```
p-fernandez, line 16: This condition should be out of the try/catch. If the defaultLayoutId is the same as the one wanted to be set as default, it is not needed to execute anything.

p-fernandez (commented):
Some comments. Thanks for helping us with this.

apps/api/src/app/layouts/layouts.controller.ts
```diff
@@ -45,7 +45,7 @@ import {
   SetDefaultLayoutUseCase,
   UpdateLayoutCommand,
   UpdateLayoutUseCase,
-} from './use-cases';
+} from './usecases';
```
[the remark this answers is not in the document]
scopsy: I defiently tend to aggree, but would be best to do a one of refactor for this in the repo. The inconsistency in this particular case in to a bit of a strange bug. Created a tech debt issue for this: https://linear.app/novu/issue/NV-1554/refactor-usecases-folder-to-be-use-cases-across-monorepo

apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
scopsy: Good catch.

apps/api/src/app/layouts/usecases/get-novu-layout/get-novu-layout.usecase.ts
scopsy: Good catch, didn't know that!

apps/api/src/app/environments/usecases/create-environment/create-environment.e2e.ts
```diff
@@ -35,4 +35,20 @@ describe('Create Environment - /environments (POST)', async () => {
...
+  it('should create a default layout for environment', async function () {
+    const demoEnvironment = {
+      name: 'Hello App',
+    };
+    const { body } = await session.testAgent.post('/v1/environments').send(demoEnvironment).expect(201);
+
+    const layouts = await layoutRepository.find({
+      _environmentId: body.data._id,
+      channel: 'email',
+    });
```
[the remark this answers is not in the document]
scopsy: I usually prefer not use endpoints like this in e2e tests from the thinking that I'm testing the create environment functionality and not the `GET /api/layouts`, in some cases, it's even helped to surface bugs that crippled from the other endpoints. In this particular scenario, I'm making a very simple query to double-check that not duplicate entries were created and etc. 

Wdyt?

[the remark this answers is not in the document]
p-fernandez: I am on the other side and try as much E2E as possible, even for the checks. Here you are just using the `layoutRepository` to check that the data created is right, not a step to contribute to creation flow.
Anyway, this was just a suggestion. 🙂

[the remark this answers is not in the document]
scopsy: You have convinced me ;) 

Should we even use the @novu/node package for that? Trying to think if we can somehow create some reusable methods for this instead of just using hardcoded and repeated session.call methods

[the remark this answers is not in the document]
p-fernandez: It is a possibility. So far that endpoint is not used that much, but if it is in the future we could always add it to the testing package, as we do for other heavily used functionalities. Something to consider, indeed.

PR #2572, opened 2023-01-22
Nv 1457 web move update button to top header

ainouzgali:

Add warn on unsaved changes, enforce {{{body}}} exists in layout content, some design changes 

Screen Shot 2023-01-22 at 19 11 17
Screen Shot 2023-01-22 at 19 11 23
lay


p-fernandez (approved):
🌟

apps/api/src/app/layouts/usecases/update-layout/update-layout.use-case.ts
```diff
@@ -24,6 +25,10 @@ export class UpdateLayoutUseCase {
     const databaseEntity = await this.getLayoutUseCase.execute(getLayoutCommand);
 
     const patchedEntity = this.applyUpdatesToEntity(this.mapToEntity(databaseEntity), command);
+    const hasBody = patchedEntity.content.includes('{{{body}}}');
+    if (!hasBody) {
+      throw new ApiException('Layout content must contain {{{body}}}');
+    }
```
p-fernandez, line 31: It would be better if we could validate this earlier so we can avoid to execute the database calls. It can go like this for now. We can improve performance later.

PR #2574, opened 2023-01-23
feat: show layout changes in web

ainouzgali:
Show layout changes in promote table

Screen Shot 2023-01-23 at 11 25 36

### DIFF
--- a/apps/api/src/app/change/usecases/get-changes/get-changes.usecase.ts
+++ b/apps/api/src/app/change/usecases/get-changes/get-changes.usecase.ts
@@ -6,6 +6,7 @@ import {
   NotificationGroupRepository,
   NotificationTemplateRepository,
   FeedRepository,
+  LayoutRepository,
 } from '@novu/dal';
 import { ChangeEntityTypeEnum } from '@novu/shared';
 import { ChangesResponseDto } from '../../dtos/change-response.dto';
@@ -30,7 +31,8 @@ export class GetChanges {
     private notificationTemplateRepository: NotificationTemplateRepository,
     private messageTemplateRepository: MessageTemplateRepository,
     private notificationGroupRepository: NotificationGroupRepository,
-    private feedRepository: FeedRepository
+    private feedRepository: FeedRepository,
+    private layoutRepository: LayoutRepository
   ) {}
 
   async execute(command: GetChangesCommand): Promise<ChangesResponseDto> {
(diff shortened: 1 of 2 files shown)


libs/dal/src/repositories/layout/layout.repository.ts
```diff
@@ -149,4 +149,10 @@ export class LayoutRepository extends BaseRepository<EnforceEnvironmentQuery, La
 
     return updatedEntity;
   }
+
+  async findDeleted(_id: LayoutId, _environmentId: EnvironmentId): Promise<LayoutEntity> {
+    const res = await this.layout.findDeleted({ _id, _environmentId });
+
+    return this.mapEntity(res);
+  }
```
ainouzgali, line 157: @p-fernandez I was not sure how you would like to do the mapEntity. As I don't feel this deserves a usecase.

apps/api/src/app/change/usecases/get-changes/get-changes.usecase.ts
```diff
@@ -168,4 +173,28 @@ export class GetChanges {
...
+    entityId: string,
+    environmentId: string
+  ): Promise<IViewEntity | Record<string, unknown>> {
+    let item = await this.layoutRepository.findOne({
+      _environmentId: environmentId,
+      _id: entityId,
+    });
+
+    if (!item) {
+      const items = await this.layoutRepository.findDeleted(entityId, environmentId);
```
ainouzgali, line 187: This is to show the change of a deleted layout

apps/api/src/app/change/usecases/get-changes/get-changes.usecase.ts
```diff
@@ -168,4 +173,28 @@ export class GetChanges {
...
   }
+
+  private async getTemplateDataForLayout(
+    entityId: string,
+    environmentId: string
+  ): Promise<IViewEntity | Record<string, unknown>> {
+    let item = await this.layoutRepository.findOne({
+      _environmentId: environmentId,
+      _id: entityId,
+    });
```
p-fernandez, line 184: I have seen that this module has no dependencies beyond the SharedModule and AuthModule. But I consider we shouldn't call the Layout repository here and take use of the Layout module use cases.
Here, we would need the `GetLayoutUseCase`. And then we would need to implement the use case to find a deleted layout.

libs/dal/src/repositories/layout/layout.repository.ts
p-fernandez: Replied to this in the other comment regarding the use case. But the mapping as agreed in the last tech summit should live here, so that's ok. 👍🏻

p-fernandez (approved):
🌟

PR #2575, opened 2023-01-23
fix: ensure one default layout in set as default

ainouzgali:

In update/create layout, because the db create/update of the new default happens before the SetDefaultLayoutUseCase, there are two at time. It is inconsistent which of them SetDefaultLayoutUseCase finds → so causes a state of two defaults at the same time.

### DIFF
--- a/apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
@@ -10,19 +10,19 @@ export class SetDefaultLayoutUseCase {
   constructor(private layoutRepository: LayoutRepository) {}
 
   async execute(command: SetDefaultLayoutCommand) {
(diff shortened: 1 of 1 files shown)


apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
```diff
@@ -10,9 +10,13 @@ export class SetDefaultLayoutUseCase {
   constructor(private layoutRepository: LayoutRepository) {}
 
   async execute(command: SetDefaultLayoutCommand) {
-    const defaultLayoutId = await this.findDefaultLayoutId(command.environmentId, command.organizationId);
+    const defaultLayoutId = await this.findExistingDefaultLayoutId(
```
p-fernandez, line 13:
```suggestion
    const existingDefaultLayoutId = await this.findExistingDefaultLayoutId(
```

p-fernandez (approved):
🌟 
Well spotted

PR #2576, opened 2023-01-23
fix: env for get novu integration

davidsoderberg:


scopsy (approved):

scopsy:
@davidsoderberg seems like we need to add this flag to the tests env file

PR #2579, opened 2023-01-24
fix: blueprint modal tests

davidsoderberg:


ainouzgali (approved):
Wonder if there is something we can do instead of the wait ?

davidsoderberg:
> Wonder if there is something we can do instead of the wait ?

I have tried to find something but nothing that seems to solve our problem here

PR #2585, opened 2023-01-24
fix: existing user cannot be invited to the org

LetItRock:

What issues I've encountered during the coding:
1. The original issue: existing users cannot be invited to the org: added the "Sign In & Accept" form.
2. When a new user is invited it is not possible to sign up with GitHub: added the "GitHub Sign Up" button.
3. When signing in/up with GitHub the email was not "normalized" meaning users could have two accounts one with GH, and another through the signup form, but basically using the same email. Now it's possible to create account with the signup form and login then with GH.
4. Invite issue: when sending an invite to a user we were checking if the account exists, and if not we were marking that member invite as the "new user invite" and showing the signup form, but the thing is that user could register the account in between the invite send and accept invite flow, so now it's possible to sign in with that flow.

**TODO:**
- [x] We have to write the migration that will normalize all the user emails, but before we have to check if we will have accounts with duplicated emails, and if any then think how we can merge them
- [x] For the GH Action pipeline create a new credentials: GITHUB_OAUTH_CLIENT_ID, GITHUB_OAUTH_CLIENT_SECRET which are passed as env to the API app, CYPRESS_GITHUB_USER_EMAIL, CYPRESS_GITHUB_USER_PASSWORD that are used in e2e tests to verify the sigh in/up with GitHub functionality.

**HOW TO RUN MIGRATION**
In the API folder, just run the `npm run migration`. Not sure if in the Production env, it won't be required to change the `NODE_ENV=local` to `NODE_ENV=production`.

Fixes few bugs described above.

Active session:
Screenshot 2023-01-24 at 11 53 19

Invite existing user:
Screenshot 2023-01-24 at 11 53 27

Invite a new user:
Screenshot 2023-01-24 at 11 54 21

**Videos:**
Invite a new user and signup with a form:

https://user-images.githubusercontent.com/2607232/214274136-7ff29997-3192-4d2e-89ff-b949c6c41882.mov

Invite a new user and signup with a GitHub:

https://user-images.githubusercontent.com/2607232/214274360-bb0c58eb-fdbe-4568-a7dd-47151d01384a.mov

Sign in as a new user with a GitHub and auto-accept an invite:

https://user-images.githubusercontent.com/2607232/214274566-a6a032d3-ecef-45ab-8bad-5f925eb210a3.mov

Invite existing user and sign in with a form:

https://user-images.githubusercontent.com/2607232/214274909-397f57af-25ae-446c-ab84-de19681a5b8e.mov

Invite existing user and sign in with GitHub:

https://user-images.githubusercontent.com/2607232/214275010-9a1e42e1-e57e-4f2f-bb54-394339b61707.mov


.github/workflows/reusable-web-e2e.yml
```diff
@@ -145,6 +149,9 @@ jobs:
           NODE_ENV: test
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.cypress_key }}
+          CYPRESS_GITHUB_USER_EMAIL: ${{ secrets.cypress_github_user_email }}
+          CYPRESS_GITHUB_USER_PASSWORD: ${{ secrets.cypress_github_user_password }}
+          CYPRESS_IS_CI: true
```
LetItRock, line 154: these variables are used in e2e tests to log in on the GH

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -114,6 +114,11 @@ export class AuthController {
       url += `&configurationId=${configurationId}`;
     }
 
+    const invitationToken = JSON.parse(request.query.state).invitationToken;
+    if (invitationToken) {
+      url += `&invitationToken=${invitationToken}`;
+    }
```
LetItRock, line 120: On the invitation page, we do show forms "Sign In/Up & Accept" with a GitHub button, when that button is clicked we redirect the user to GH OAuth, and in query params we do pass the `invitationToken`. When a user is successfully logged in he will be redirected to the login page, where we do read that `invitationToken` from the query params and trigger the "auto-accept invite" logic.

apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -46,14 +47,15 @@ export class AuthService {
     distinctId: string,
     origin?: SignUpOriginEnum
   ) {
-    let user = await this.userRepository.findByLoginProvider(profile.id, authProvider);
+    const email = normalizeEmail(profile.email);
+    let user = await this.userRepository.findByEmail(email);
```
LetItRock, line 51: Two things here:
- we should normalize the email
- we should look for the user by the email
Otherwise, it will be possible to create two accounts, one with sign-up form and another with a GH OAuth.

This change requires running the migration (check the description).

apps/api/src/app/invites/usecases/get-invite/get-invite.usecase.ts
```diff
@@ -15,14 +17,12 @@ export class GetInvite {
   ) {}
 
   async execute(command: GetInviteCommand) {
-    const member = await this.memberRepository.findByInviteToken(command.token);
-    if (!member) throw new ApiException('No invite found');
+    const invitedMember = await this.memberRepository.findByInviteToken(command.token);
```
LetItRock, line 20: renamed

apps/api/src/app/invites/usecases/get-invite/get-invite.usecase.ts
```diff
@@ -32,6 +32,8 @@ export class GetInvite {
     const user = await this.userRepository.findById(invitedMember.invite._inviterId);
     if (!user) throw new NotFoundException('User not found');
 
+    const invitedUser = await this.userRepository.findByEmail(normalizeEmail(invitedMember.invite.email));
```
LetItRock, line 35: Find the invited user by email and if exists set the `_userId`, this field is then used on the FE to show the "Sign In & Accept" or "Sign Up & Accept" forms.
Before this flag was set during the invite creation, which is wrong, because the user account can be created in between the invitation sent and the invitation accept (check the description for more).

apps/web/cypress/support/commands.ts
```diff
@@ -130,4 +132,29 @@ Cypress.Commands.add('clearDatabase', () => {
...
+      cy.get('body').then(($body) => {
+        if ($body.find('#js-oauth-authorize-btn').length) {
+          cy.wait(2000).then(() => {
+            $body.find('#js-oauth-authorize-btn').trigger('click');
+          });
+        }
+      });
+    }
+  );
+});
```
LetItRock, line 158: Added the new Cypress command which allows to Sign In/Up with a GitHub. It has the logic that after sign in checks if the authorize button is shown, and if then click on it.

apps/web/cypress/tests/auth.spec.ts
```diff
@@ -38,6 +40,63 @@ describe('User Sign-up and Login', function () {
       cy.getByTestId('submitButton').click();
       cy.get('.mantine-TextInput-error').contains('Please provide a valid email');
     });
+
+    it('should allow to sign-up with Github, logout, and login', function () {
+      const isCI = Cypress.env('IS_CI');
+      if (!isCI) return;
```
LetItRock, line 46: All the e2e tests that use GH button will be only running on the CI, because we don't want all the contributors to have to setup the GH app and credentials etc.

apps/web/cypress/tests/auth.spec.ts
```diff
@@ -130,12 +189,15 @@ describe('User Sign-up and Login', function () {
...
-      const todaysDate = new Date();
-      todaysDate.setDate(todaysDate.getDate() + 30); // iat - exp = 30 days
-      cy.clock(todaysDate);
+      const THIRTY_DAYS = 1000 * 60 * 60 * 24 * 30; // iat - exp = 30 days
+      const ONE_MINUTE = 1000 * 60; // adding 1 minute to be sure that token is expired
+      const date = new Date(Date.now() + THIRTY_DAYS + ONE_MINUTE);
+      cy.clock(date);
 
-      cy.visit('/templates');
+      cy.visit('/subscribers');
```
LetItRock, line 200: I had to update this test as I have had issues that the `cy.clock(date)` was before the token exp date, even if the token was generated before, weird, so I've added one additional minute to that date.

apps/web/src/components/auth/LoginForm.tsx
```diff
@@ -30,14 +31,15 @@ export function LoginForm({ email, token }: Props) {
...
 
   const { isFromVercel, code, next, configurationId } = useVercelParams();
   const vercelQueryParams = `code=${code}&next=${next}&configurationId=${configurationId}`;
   const signupLink = isFromVercel ? `/auth/signup?${vercelQueryParams}` : '/auth/signup';
   const resetPasswordLink = isFromVercel ? `/auth/reset/request?${vercelQueryParams}` : `/auth/reset/request`;
   const githubLink = isFromVercel
-    ? `${API_ROOT}/v1/auth/github?partnerCode=${code}&next=${next}&configurationId=${configurationId}&source=${SignUpOriginEnum.VERCEL}`
-    : `${API_ROOT}/v1/auth/github?source=${SignUpOriginEnum.WEB}`;
+    ? buildVercelGithubLink({ code, next, configurationId })
+    : buildGithubLink({ invitationToken });
```
LetItRock, line 42: moved to reusable functions

apps/web/src/components/auth/LoginForm.tsx
```diff
@@ -58,9 +60,21 @@ export function LoginForm({ email, token }: Props) {
...
+        setToken(token);
+
+        return;
+      }
+
+      if (invitationToken) {
+        submitToken(token, invitationToken);
+
+        return;
+      }
```
LetItRock, line 74: The LoginForm component is used on the: Login page and Invitation Page, when it's used on the second one the `invitationToken` is passed, and then when the "Sign In & Accept" button is pressed we do submit the token.

apps/web/src/components/auth/SetupLoader.tsx
```diff
@@ -4,7 +4,8 @@ import { colors, Text } from '../../design-system';
 type Props = {
   title: string;
 };
-const VercelSetupLoader = ({ title }: Props) => {
+
+const SetupLoader = ({ title }: Props) => {
```
LetItRock, line 8: renamed

apps/web/src/components/auth/SignUpForm.tsx
```diff
@@ -73,17 +74,15 @@ export function SignUpForm({ token, email }: Props) {
...
+    const token = (response as any).token;
+    applyToken(token);
 
-    if (token) {
-      const result = await submitToken(token);
-      if (!result) return;
-
-      navigate('/templates');
+    if (invitationToken) {
+      submitToken(token, invitationToken);
```
LetItRock, line 81: similar changes in the SignUpForm

apps/web/src/components/auth/githubUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+  configurationId: string | null;
+}) => {
+  const queryParams = new URLSearchParams();
+  queryParams.append('partnerCode', code ?? '');
+  queryParams.append('next', next ?? '');
+  queryParams.append('configurationId', configurationId ?? '');
+  queryParams.append('source', SignUpOriginEnum.VERCEL);
+
+  return `${API_ROOT}/v1/auth/github?${queryParams.toString()}`;
+};
```
LetItRock, line 31: @BiswaViraj you know this code well, could you verify if that doesn't have any issues?

apps/web/src/components/auth/use-accept-invite.hook.ts
```diff
@@ -14,20 +19,29 @@ export function useAcceptInvite() {
...
-      await queryClient.refetchQueries({
-        predicate: (query) => query.queryKey.includes('/v1/organizations'),
+  const submitToken = useCallback(async (token: string, invitationToken: string, refetch = false) => {
+    try {
+      // just set the header, user is logged in after token is submitted
+      applyToken(token);
+      const newToken = await mutateAsync(invitationToken);
+      if (!newToken) return;
+
+      setToken(newToken, refetch);
```
LetItRock, line 29: The `/v1/invites/${tokenItem}/accept`endpoint requires the JWT, so we call `applyToken` just to set the header, then the result is a new JWT token sent back, which we update in the app with `setToken`.

I haven't used the `setToken` instead of `applyToken` because the first one triggers additional calls to our API, which we don't want to do.

apps/web/src/config/index.ts
```diff
@@ -7,15 +7,16 @@ declare global {
   }
 }
 
+const isCypress = (isBrowser() && (window as any).Cypress) || (isBrowser() && (window as any).parent.Cypress);
```
LetItRock, line 10: I've used the `cy.origin` a new functionality that allows using Cypress cross domains, but currently there is a bug which clears the `window.Cypress` when navigating back from the origin, that's an additional check `window.parent.Cypress` is needed. 
More here: https://github.com/cypress-io/cypress/issues/25394

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -11,43 +11,48 @@ import { SignUpForm } from '../../components/auth/SignUpForm';
...
 
 export default function InvitationPage() {
+  const queryClient = useQueryClient();
   const navigate = useNavigate();
   const { token, logout, currentUser } = useContext(AuthContext);
   const location = useLocation();
   const isLoggedIn = !!token;
-  const { token: tokenParam } = useParams<{ token: string }>();
-  const { isLoading: loadingAcceptInvite, submitToken } = useAcceptInvite();
+  const { token: invitationToken } = useParams<{ token: string }>();
```
LetItRock, line 22: renamed

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -11,43 +11,48 @@ import { SignUpForm } from '../../components/auth/SignUpForm';
...
     navigate(location.pathname);
   };
 
   useEffect(() => {
-    acceptToken();
-  }, [tokenParam, data, currentUser]);
+    // auto accept invitation when logged in as invited user
+    if (isLoggedInAsInvitedUser) {
+      submitToken(tokensRef.current.token as string, tokensRef.current.invitationToken as string, true);
+    }
```
LetItRock, line 48: Updated this effect to be called only once when the current logged-in user is the invited user, like this triggers the auto-accept invite call.

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -11,43 +11,48 @@ import { SignUpForm } from '../../components/auth/SignUpForm';
...
+  }, [isLoggedInAsInvitedUser]);
+
+  useEffect(() => {
+    return () => {
+      queryClient.removeQueries(['getInviteTokenData']);
```
LetItRock, line 53: we have to remove this query, otherwise, the `queryClient.refetch` will refetch it with a failed request.

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -57,18 +62,11 @@ export default function InvitationPage() {
...
             <Center inline mb={40} mt={20}>
               <Text size="lg" color={colors.B60}>
-                <When truthy={invalidCurrentUser && !loadingAcceptInvite}>
-                  <p>The invite is not valid for the current user. Please log in with the right user.</p>
-                </When>
-
-                <When truthy={!invalidCurrentUser && !loadingAcceptInvite}>
-                  <p>Your session is currently active, use another browser or switch to incognito mode.</p>
-                  <p>Log out instead?</p>
-                </When>
```
LetItRock, line 67: We don't need to show two different descriptions, which basically means the same that you have to logout.

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -89,18 +87,18 @@ export default function InvitationPage() {
...
         <AuthContainer
-          title="Get Started"
+          title={existingUser ? 'Sign In & Accept Invite' : 'Get Started'}
           customDescription={
             inviterFirstName && organizationName ? (
-              <Center inline mb={60} mt={20}>
+              <Center inline mb={60} mt={20} data-test-id="invitation-description">
                 <Text size="lg" mr={4} color={colors.B60}>
-                  You've been invited by
+                  {`You've been invited by `}
```
LetItRock, line 95: fixed texts so it's easier to do checks in tests

apps/web/src/pages/auth/LoginPage.tsx
```diff
@@ -2,67 +2,80 @@ import { useContext, useEffect } from 'react';
...
+  const { isLoading: isLoadingAcceptInvite, submitToken } = useAcceptInvite();
 
   useEffect(() => {
     if (queryToken) {
-      setToken(queryToken);
-    }
-  }, [queryToken]);
+      const user = jwtDecode<IJwtPayload>(queryToken);
 
-  useEffect(() => {
```
LetItRock, line 32: merged two `useEffect` into one

apps/web/src/pages/auth/LoginPage.tsx
```diff
@@ -2,67 +2,80 @@ import { useContext, useEffect } from 'react';
...
   const { startVercelSetup, isLoading } = useVercelIntegration();
   const { code, isFromVercel, next } = useVercelParams();
+  const { isLoading: isLoadingAcceptInvite, submitToken } = useAcceptInvite();
 
   useEffect(() => {
     if (queryToken) {
-      setToken(queryToken);
-    }
-  }, [queryToken]);
+      const user = jwtDecode<IJwtPayload>(queryToken);
```
LetItRock, line 32: when there is a `token` in query params this effect will be triggered

apps/web/src/pages/auth/LoginPage.tsx
```diff
@@ -2,67 +2,80 @@ import { useContext, useEffect } from 'react';
...
+        navigate('/quickstart');
+
+        return;
+      }
 
-        navigate(source === 'cli' ? '/quickstart' : '/');
+      if (invitationToken) {
+        submitToken(queryToken, invitationToken);
+
+        return;
```
LetItRock, line 63: auto accept an invite

apps/api/src/app/auth/services/auth.service.ts
p-fernandez: I wasn't fully convinced of the normalized email solution but right now is the simplest way to solve this problem. Hopefully we can improve the authentication system in the future.
Many thanks for this amazing find and work. 👏🏻

apps/web/cypress/global.d.ts
```diff
@@ -43,6 +43,8 @@ declare namespace Cypress {
      */
     inviteUser(email: string): Chainable<Response>;
 
+    loginWithGithub(): Chainable<any>;
```
p-fernandez, line 46:
```suggestion
    loginWithGitHub(): Chainable<any>;
```
To be in line with their branding name and the camel casing convention.

apps/web/cypress/support/commands.ts
```diff
@@ -130,4 +132,29 @@ Cypress.Commands.add('clearDatabase', () => {
...
+  const githubPassword = Cypress.env('GITHUB_USER_PASSWORD');
+
+  cy.getByTestId('github-button').click();
+
+  return cy.origin(
+    'https://github.com',
+    { args: { githubUserEmail, githubPassword } },
+    ({ githubUserEmail, githubPassword }) => {
+      cy.get('#login_field').type(githubUserEmail);
+      cy.get('#password').type(githubPassword);
```
p-fernandez, line 146:
```suggestion
Cypress.Commands.add('loginWithGitHub', () => {
  const gitHubUserEmail = Cypress.env('GITHUB_USER_EMAIL');
  const gitHubPassword = Cypress.env('GITHUB_USER_PASSWORD');

  cy.getByTestId('github-button').click();

  return cy.origin(
    'https://github.com',
    { args: { gitHubUserEmail, gitHubPassword } },
    ({ gitHubUserEmail, gitHubPassword }) => {
      cy.get('#login_field').type(gitHubUserEmail);
      cy.get('#password').type(gitHubPassword);
```
To be in line with their branding name and the camel casing convention.

apps/web/cypress/tests/auth.spec.ts
```diff
@@ -38,6 +40,63 @@ describe('User Sign-up and Login', function () {
...
+    it('should allow to sign-up, logout, and login with Github using same email address', function () {
+      const isCI = Cypress.env('IS_CI');
+      if (!isCI) return;
+
+      const githubUserEmail = Cypress.env('GITHUB_USER_EMAIL');
+
+      cy.intercept('**/organization/**/switch').as('appSwitch');
+      cy.visit('/auth/signup');
+      cy.getByTestId('fullName').type('Test User');
+      cy.getByTestId('email').type(githubUserEmail);
```
p-fernandez, line 79:
```suggestion
    it('should allow to sign-up, logout, and login with GitHub using same email address', function () {
      const isCI = Cypress.env('IS_CI');
      if (!isCI) return;

      const gitHubUserEmail = Cypress.env('GITHUB_USER_EMAIL');

      cy.intercept('**/organization/**/switch').as('appSwitch');
      cy.visit('/auth/signup');
      cy.getByTestId('fullName').type('Test User');
      cy.getByTestId('email').type(gitHubUserEmail);
```

apps/web/cypress/tests/auth.spec.ts
```diff
@@ -130,12 +189,15 @@ describe('User Sign-up and Login', function () {
...
       cy.getByTestId('submit-btn').click();
 
+      cy.location('pathname').should('equal', '/templates');
+
       // setting current time in future, to simulate expired token
-      const todaysDate = new Date();
-      todaysDate.setDate(todaysDate.getDate() + 30); // iat - exp = 30 days
-      cy.clock(todaysDate);
+      const THIRTY_DAYS = 1000 * 60 * 60 * 24 * 30; // iat - exp = 30 days
+      const ONE_MINUTE = 1000 * 60; // adding 1 minute to be sure that token is expired
```
p-fernandez, line 196:
```suggestion
      const ONE_MINUTE = 1000 * 60; // adding 1 minute to be sure that token is expired
      const THIRTY_DAYS = ONE_MINUTE * 60 * 24 * 30; // iat - exp = 30 days
```
🏌🏻

apps/web/cypress/tests/invites.spec.ts
```diff
@@ -3,21 +3,84 @@ import * as capitalize from 'lodash.capitalize';
...
+      doRegister(this.token);
+
+      cy.getByTestId('header-profile-avatar').click();
+      cy.getByTestId('header-dropdown-organization-name').contains(capitalize(this.organization.name.split(' ')[0]));
+    });
+  });
+
+  it('should allow to login with Github if invited existing user', function () {
+    const isCI = Cypress.env('IS_CI');
+    if (!isCI) return;
```
p-fernandez, line 19: Something we can do in the future is to separate all the CI oriented tests in a specific folder and trigger them only in CI without having to do these checks in the test code.

apps/web/src/components/auth/LoginForm.tsx
```diff
@@ -107,6 +122,7 @@ export function LoginForm({ email, token }: Props) {
           required
           label="Email"
           placeholder="Type your email..."
+          disabled={!!invitationToken}
```
p-fernandez, line 125: Great little UX touch. 👏🏻

apps/web/src/components/auth/LoginForm.tsx
```diff
@@ -128,8 +144,15 @@ export function LoginForm({ email, token }: Props) {
...
-          Sign In
+        <Button
+          submit
+          mt={60}
+          inherit
+          loading={isLoading || isLoadingAcceptInvite}
+          disabled={isLoadingAcceptInvite}
+          data-test-id="submit-btn"
+        >
+          {invitationToken ? 'Sign In & Accept' : 'Sign In'}
```
p-fernandez, line 155: Would it make more sense to have the text `Accept & Sign In`? In that order.

apps/web/src/components/auth/SignUpForm.tsx
```diff
@@ -216,7 +216,7 @@ export function SignUpForm({ token, email }: Props) {
           submit
           data-test-id="submitButton"
         >
-          Sign Up {token ? '& Accept Invite' : null}
+          Sign Up {invitationToken ? '& Accept Invite' : null}
```
p-fernandez, line 219: Same as in the LoginForm, for me it makes more sense `Accept Invite & Sign Up` as it is the order of the actions happening. Maybe I am petty with the wording though.

apps/web/src/components/auth/githubUtils.ts
```diff
@@ -0,0 +1,31 @@
+import { SignUpOriginEnum } from '@novu/shared';
```
p-fernandez, line 1: Filename should be `gitHubUtils`, so I can be coherent with my previous comments. 🙂

apps/web/src/components/auth/githubUtils.ts
p-fernandez: It looks ok for me comparing with the existing code, but let's have @BiswaViraj checking it. 🙏🏻 
Maybe I wouldn't add the `queryParams` with empty string value if they are not defined for cleanliness when building the URL and to avoid any potential edge case bug if the query param is passed.

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -11,43 +11,48 @@ import { SignUpForm } from '../../components/auth/SignUpForm';
...
-  const invalidCurrentUser = existingUser && currentUser && currentUser._id !== data?._userId;
-
-  const acceptToken = async () => {
-    if (existingUser && currentUser && currentUser._id === data?._userId && isLoggedIn) {
-      const result = await submitToken(tokenParam as string, true);
-      if (result) navigate('/templates');
-    }
-  };
+  const existingUser = !!(invitationToken && data?._userId);
+  const isLoggedInAsInvitedUser = !!(isLoggedIn && existingUser && currentUser && currentUser._id === data?._userId);
```
p-fernandez, line 37:
```suggestion
  const isLoggedInAsInvitedUser = !!(isLoggedIn && existingUser && currentUser?._id === data?._userId);
```
I think checking `currentUser` on its own is superflous as part of this seeing the last branch of the condition. But this flag calculation is very brittle, so I might be wrong.

apps/web/src/pages/auth/InvitationPage.tsx
p-fernandez: This is a great catch.

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -112,14 +110,23 @@ export default function InvitationPage() {
...
+              loaderProps={{
+                color: colors.error,
+              }}
+            />
+          ) : (
+            <>
+              {existingUser ? (
+                <LoginForm email={data?.email} invitationToken={invitationToken} />
+              ) : (
+                <SignUpForm email={data?.email} invitationToken={invitationToken} />
```
p-fernandez, line 126: We can do something like:
```
const ClassForm = existingUser ? LoginForm : SignUpForm;

return (<ClassForm email={data?.email} invitationToken={invitationToken} />);
```

p-fernandez (approved):
Brilliant work. 🌟

apps/web/src/components/auth/LoginForm.tsx
LetItRock: I did it the same way like it is in the sign up form

apps/web/src/components/auth/SignUpForm.tsx
LetItRock: yeah, the action flow depends on the user's choice for signup: GitHub button or form button.
For GH it's: sign-up & accept
For form it's: accept & sign up

apps/web/src/pages/auth/InvitationPage.tsx
LetItRock: it should stay there, otherwise, if remove it, the last branch might compare `undefined === data?._userId` which is not what we want

apps/web/src/pages/auth/InvitationPage.tsx
LetItRock: sure, at the beginning I had different args there, but now it's the same, so can be changed I guess.

apps/web/src/components/auth/SignUpForm.tsx
p-fernandez: Oh yeah, you are right. Even more complicated. 😓

apps/web/src/components/auth/githubUtils.ts
BiswaViraj: This might create an issue, as those fields are essential for vercel integration

PR #2587, opened 2023-01-24
feat: ensure one default layout

ainouzgali:

 We need to make sure user can not remove the only default layout - I see two usecases that could happen : 

In web - unchecking the default 

When deleting a layout, we disallow delete of an assigned layout, but it could be unassigned and the default.

### DIFF
--- a/apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
@@ -35,7 +35,9 @@ export class DeleteLayoutUseCase {
     }
 
     if (layout.isDefault) {
-      throw new ConflictException(`Layout is being used as your default layout, so it can not be deleted`);
+      throw new ConflictException(
+        `Layout with id ${command.layoutId} is being used as your default layout, so it can not be deleted`
+      );
     }
 
     await this.layoutRepository.deleteLayout(command.layoutId, layout._environmentId, layout._organizationId);
--- a/apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
@@ -1,5 +1,5 @@
 import { ConflictException, Injectable } from '@nestjs/common';
-import { LayoutEntity, LayoutRepository } from '@novu/dal';
+import { LayoutRepository } from '@novu/dal';
 
 import { DeleteLayoutCommand } from './delete-layout.command';
 
@@ -34,14 +34,11 @@ export class DeleteLayoutUseCase {
       throw new ConflictException(`Layout with id ${command.layoutId} is being used so it can not be deleted`);
     }
 
-    await this.layoutRepository.deleteLayout(command.layoutId, layout._environmentId, layout._organizationId);
+    if (layout.isDefault) {
+      throw new ConflictException(`Layout is being used as your default layout, so it can not be deleted`);
+    }
 
-    const createLayoutChangeCommand = CreateLayoutChangeCommand.create({
-      environmentId: command.environmentId,
-      layoutId: command.layoutId,
-      organizationId: command.organizationId,
-      userId: command.userId,
-    });
+    await this.layoutRepository.deleteLayout(command.layoutId, layout._environmentId, layout._organizationId);
 
     await this.createChange(command);
   }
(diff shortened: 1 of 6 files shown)


apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
```diff
@@ -34,14 +34,11 @@ export class DeleteLayoutUseCase {
       throw new ConflictException(`Layout with id ${command.layoutId} is being used so it can not be deleted`);
     }
 
-    await this.layoutRepository.deleteLayout(command.layoutId, layout._environmentId, layout._organizationId);
+    if (layout.isDefault) {
+      throw new ConflictException(`Layout is being used as your default layout, so it can not be deleted`);
```
p-fernandez, line 38:
```suggestion
      throw new ConflictException(`Layout with id ${command.layoutId} is being used as your default layout, so it can not be deleted`);
```

apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
```diff
@@ -34,14 +34,11 @@ export class DeleteLayoutUseCase {
...
+    }
 
-    const createLayoutChangeCommand = CreateLayoutChangeCommand.create({
-      environmentId: command.environmentId,
-      layoutId: command.layoutId,
-      organizationId: command.organizationId,
-      userId: command.userId,
-    });
```
p-fernandez, line 44: Did you delete this?

apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
```diff
@@ -85,6 +84,6 @@ export class SetDefaultLayoutUseCase {
     organizationId: OrganizationId,
     isDefault: boolean
   ): Promise<void> {
-    this.layoutRepository.updateIsDefault(layoutId, environmentId, organizationId, isDefault);
+    await this.layoutRepository.updateIsDefault(layoutId, environmentId, organizationId, isDefault);
```
p-fernandez, line 87: Good catch

apps/api/src/app/layouts/usecases/update-layout/update-layout.use-case.ts
```diff
@@ -26,6 +26,10 @@ export class UpdateLayoutUseCase {
     });
     const databaseEntity = await this.getLayoutUseCase.execute(getLayoutCommand);
 
+    if (!command.isDefault && databaseEntity.isDefault) {
+      throw new ConflictException(`One default layout is required`);
+    }
```
p-fernandez, line 31: Not sure if this is the right thing to do. Here you are checking that the existing layout is a default. That doesn't mean that other layout is the default for the environment.

p-fernandez (commented):
I spotted 2 things that I think are not totally right. Let's review them together if needed.
Also we need to add tests for this. If you struggle with them I can take over this and add the tests.

apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
ainouzgali: Yes, it was not being used. You create this command inside the createChange function.

apps/api/src/app/layouts/usecases/update-layout/update-layout.use-case.ts
ainouzgali: This checks if the update (the data from the command) is to change default to false, and that previously (from the db) it was set to true. 
There could only be one default, so if it is changed from true to false, there is no other default. 
Or I might not understand what you mean :)

apps/api/src/app/layouts/usecases/delete-layout/delete-layout.use-case.ts
p-fernandez: Argh missed that. Good catch!

apps/api/src/app/layouts/usecases/update-layout/update-layout.use-case.ts
p-fernandez: Got it. I got it wrong and this is totally right. 👏🏻

p-fernandez (approved):
🌟 
Was wrong with my spots. Good work!

ainouzgali:
Adding tests. Thanks Pablo !💃🏼

PR #2595, opened 2023-01-24
feat(api): tests for layout promotion

p-fernandez:

Adds tests for the layout promotion.

Not need to explain why tests are needed. :)


ainouzgali (approved):
⭐️

PR #2596, opened 2023-01-24
fix: so changes are aggregated into one

ainouzgali:
After multiple updates in dev, there should be one row per layout in the changes table. Before, it shows all updates separately and they needed to be promoted individually (could cause unexpected behavior in the production if one change depends on another)

Now, all changes made to a layout will be aggregated into a single promotable change. 

* I still need to add tests. Will do in this pr

### DIFF
--- a/apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
@@ -1,4 +1,4 @@
-import { LayoutEntity, LayoutRepository } from '@novu/dal';
+import { LayoutRepository, ChangeRepository } from '@novu/dal';
 import { ChangeEntityTypeEnum } from '@novu/shared';
 import { Injectable } from '@nestjs/common';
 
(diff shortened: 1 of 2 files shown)


apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
```diff
@@ -13,10 +13,17 @@ export class CreateLayoutChangeUseCase {
...
+    private layoutRepository: LayoutRepository,
+    private changeRepository: ChangeRepository
   ) {}
 
   async execute(command: CreateLayoutChangeCommand, isDeleteChange = false): Promise<void> {
+    const parentChangeId: string = await this.changeRepository.getChangeId(
+      command.environmentId,
+      ChangeEntityTypeEnum.LAYOUT,
+      command.layoutId
+    );
```
p-fernandez, line 25: We should move this to inside the condition, to only execute it in case there is a layout found for the different cases.

p-fernandez (approved):
🌟

apps/api/src/app/layouts/usecases/create-layout-change/create-layout-change.use-case.ts
ainouzgali: What do you mean by condition?

p-fernandez: The condition that checks if there is item.

ainouzgali: Good point

PR #2599, opened 2023-01-25
Nv 1579 useintercom must be inside

BiswaViraj:

Refactored the `GotAQuestionButton` component to not throw error when `INTERCOM_APP_ID` & `IntercomProvider` is unavailable


scopsy (approved):
Looks good@

PR #2602, opened 2023-01-25
Dockerfile tests on PR's

scopsy:

- Create new setup-project action to handle with dependency install and environment
- Refactor all actions to use setup project
- Refactor a docker build for API action
- Add a docker build and test for API on new PR's

In some cases, changes introduced in pr's caused docker build to fail, and we would only know about it when deploying.


.github/workflows/reusable-widget-e2e.yml
```diff
@@ -84,6 +61,11 @@ jobs:
       - name: Wait on WIDGET
         run: wait-on --timeout=90000 http://localhost:3500
 
+      - name: Cypress install
+        if: steps.setup-project.outputs.cypress_cache_hit != 'true'
+        working-directory: apps/widget
+        run: pnpm cypress install
+
```
djabarovgeorge, line 68: Were we missing it before?

.github/workflows/test-api.yml
```diff
@@ -12,8 +12,23 @@ on:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 80
+    permissions:
+      contents: read
+      packages: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-project
+      - uses: ./.github/actions/docker/build-api
```
djabarovgeorge, line 31: Amazing 💯

.github/actions/docker/build-api/action.yml
```diff
@@ -0,0 +1,81 @@
...
+      shell: bash
+      run: |
+        docker push ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:$IMAGE_TAG
+        docker push ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:${{ inputs.tag }}
+
+    - name: Push Base Image
+      if: ${{ inputs.push == 'true' && inputs.tag }}
+      shell: bash
+      run: |
+        docker push ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:${{ inputs.tag }}
```
djabarovgeorge, line 76: is this one partial duplication from the previous step?

.github/actions/docker/build-api/action.yml
```diff
@@ -0,0 +1,81 @@
...
+        oidc: true
+    - name: Login To Registry
+      shell: bash
+      env:
+        GH_ACTOR: ${{ github.actor }}
+        GH_PASSWORD: ${{ inputs.github_token }}
+      run: |
+        echo $GH_PASSWORD | docker login ghcr.io -u $GH_ACTOR --password-stdin
+
+    - name: Build, tag, and push image to Amazon ECR
```
djabarovgeorge, line 41: should we do the push part here optional as we do on ghcr push (on the next step)?

djabarovgeorge (approved):
Looks amazing, loved the image build on the API test.

.github/workflows/reusable-widget-e2e.yml
scopsy: We now cache the cypress binaries so it was needed now :)

.github/workflows/test-api.yml
djabarovgeorge: Thought about something, what about creating a separate action that will be triggered in order to build the app (API for example), this way we could remove the complexity of the if statement, and it could run independently and be parallelized

@scopsy what do you think?

.github/workflows/reusable-widget-e2e.yml
LetItRock: hehe, I noticed the warning about it some time ago but totally forgot about it 😅

.github/actions/docker/build-api/action.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: |
+        docker tag ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:$IMAGE_TAG ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:${{ inputs.tag }}
+
+
+    - name: Push Base Image
+      if: ${{ inputs.push == 'true' }}
+      shell: bash
+      run: |
+        docker push ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:$IMAGE_TAG
+        docker push ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:${{ inputs.tag }}
```
LetItRock, line 70: this line shouldn't be here as we do push this tagged image in next step

.github/actions/docker/build-api/action.yml
```diff
@@ -0,0 +1,81 @@
...
+
+    - name: Push Base Image
```
LetItRock, line 72: change the name as it is a duplicated from previous

.github/actions/setup-project/action.yml
```diff
@@ -0,0 +1,75 @@
...
+      with:
+        node-version: '16.15.1'
+
+    - name: 💵 Start Redis
+      uses: supercharge/redis-github-action@1.2.0
+
+    - name: 📚 Start MongoDB
+      uses: supercharge/mongodb-github-action@1.3.0
+      with:
+        mongodb-version: 4.2.8
```
LetItRock, line 34: I'm not sure if we would like to do this for all the workflows, ex. the `dev-deploy-embed` doesn't need it.
I feel like both should be optional.

.github/workflows/reusable-api-e2e.yml
```diff
@@ -14,83 +14,47 @@ jobs:
...
-        pnpm test
-
+      - uses: actions/checkout@v3
+
+      - uses: ./.github/actions/setup-project
+
+      - uses: mansagroup/nrwl-nx-action@v3
+        with:
+          targets: lint
+          projects: "@novu/api"
```
LetItRock, line 24: after the lint we should do build, test

.github/actions/docker/build-api/action.yml
scopsy: Good catch! Exactly the problem

.github/actions/docker/build-api/action.yml
```diff
@@ -0,0 +1,81 @@
...
+      id: build-image
+      shell: bash
+      env:
+        REGISTRY_OWNER: novuhq
+        DOCKER_NAME: novu/api
+        IMAGE_TAG: ${{ github.sha }}
+        DEPOT_PROJECT_ID: 6sj0jfv0n7
+      run: |
+        cd apps/api && pnpm --silent --workspace-root pnpm-context -- apps/api/Dockerfile | depot build --build-arg PACKAGE_PATH=apps/api - -t novu-api --load
+        docker tag novu-api ghcr.io/$REGISTRY_OWNER/$DOCKER_NAME:$IMAGE_TAG
```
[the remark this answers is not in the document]
scopsy: I tag here with the PR sha so we can associate it to the PR itself. latest, prod and dev should be after the pr was merged as part of another workflow

.github/actions/setup-project/action.yml
scopsy: I aggree, in the next phase we will optimize those areas. Will create a task for  this. Wasn't part of the scope for this change

.github/workflows/reusable-api-e2e.yml
scopsy: It happens down below after the localstack section or i miss something?

.github/workflows/test-api.yml
scopsy: I'm not sure i follow 100% percent. Do you mind explaining a bit more or giving some concrete example?

.github/actions/docker/build-api/action.yml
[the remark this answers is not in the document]
LetItRock: I mean I'm not sure if the tagging here is needed as we do push the tags later in the steps

.github/workflows/reusable-api-e2e.yml
LetItRock: yeah, I meant it's better to change the order, fail earlier

.github/actions/docker/build-api/action.yml
scopsy: Yes! :) fixed it

.github/workflows/test-api.yml
djabarovgeorge: Sure thing, after looking at it again I think we can create an additional flow that will be triggered for example 'api-image-build' so we can separate the so flow and parallelize the two flows as well. it is not necessary at all just thought it will help during the testing time.

.github/actions/docker/build-api/action.yml
```diff
@@ -0,0 +1,94 @@
...
+      run: |
+        echo $GH_PASSWORD | docker login ghcr.io -u $GH_ACTOR --password-stdin
+
+    - name: Build, tag, and test
+      shell: bash
+      env:
+        REGISTRY_OWNER: novuhq
+        DOCKER_NAME: novu/api
+        IMAGE_TAG: ${{ github.sha }}
+        DEPOT_PROJECT_ID: 6sj0jfv0n7
```
p-fernandez, line 47: Where does this value come from?

.github/actions/setup-project/action.yml
```diff
@@ -0,0 +1,75 @@
...
+description: Sets up the whole monorepo and install dependencies
+
+inputs:
+  cypress_version:
+    description: 'The version of cypress to install'
+    required: false
+    default: ''
+outputs:
+  cypress_cache_hit:
+    description: 'Did cypress used binary cache'
```
p-fernandez, line 12:
```suggestion
    description: 'Did cypress use binary cache'
```

.github/actions/setup-project/action.yml
```diff
@@ -0,0 +1,75 @@
...
+    value: ${{ inputs.cypress_version != '' && steps.cache-cypress-binary-version.outputs.cache-hit || steps.cache-cypress-binary.outputs.cache-hit}}
```
p-fernandez, line 13: Where I could read about the meaning of this steps?
cache-cypress-binary-version
cache-cypress-binary

.github/actions/setup-project/action.yml
```diff
@@ -0,0 +1,75 @@
...
+  steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - name: ⚙️ Setup kernel for react native, increase watchers
+      shell: bash
+      run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+    - uses: actions/setup-node@v2
+      name: ⚙️ Setup Node Version
+      with:
+        node-version: '16.15.1'
```
p-fernandez, line 26: In the future we might want to use this GH action:
https://github.com/marketplace/actions/get-nodejs-version

We could also write one for the MongoDB and PNPM versions.

.github/workflows/dev-deploy-api.yml
```diff
@@ -34,79 +34,14 @@ jobs:
...
     steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-      - uses: actions/checkout@v2
-      - name: Setup kernel for react native, increase watchers
-        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-project
+      - uses: ./.github/actions/docker/build-api
+        id: docker_build
```
p-fernandez, line 40: I am delighted at this cleaning. 👏🏻

.github/workflows/reusable-api-e2e.yml
```diff
@@ -14,83 +14,47 @@ jobs:
...
+          until $command
+          do
+            sleep 1
+            [[ counter -eq $max_retry ]] && echo "Failed!" && exit 1
+            aws --endpoint-url=http://localhost:4566 s3 ls
+            echo "Trying again. Try #$counter"
+            ((counter++))
+          done
+          docker-compose -f docker-compose.localstack.yml logs --tail="all"
+          aws --endpoint-url=http://localhost:4566 --cli-connect-timeout 600 s3 mb s3://novu-test
```
p-fernandez, line 49: Should we set the localhost url and port as env variable and provide it to the workflow?

.github/workflows/test-docs.yml
```diff
@@ -3,40 +3,13 @@ on:
...
-        with:
-          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
-
-      - name: Install dependencies
-        run: pnpm install
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup-project
```
p-fernandez, line 12: Aren't the steps of the job `test` the same as the workflow defined `workflow_dispatch`? Should reuse the workflow for the `test` job?
Not sure if I miss anything but to me this looks like it is being executing the whole steps of the `workflow_dispatch` twice at the end.
This would happen in the other ocurrences where these same changes were done.

.github/actions/docker/build-api/action.yml
scopsy: It's from the depot identifier on their site

.github/actions/setup-project/action.yml
scopsy: This is mainly for the old widget package that uses and old version for cypress that we need to cache binaries pinned to a particular version. Unfortuently couldn't find a nice way to handle this

.github/actions/setup-project/action.yml
scopsy: that's cool! Love it

.github/workflows/reusable-api-e2e.yml
scopsy: Yes most defiently, There is another round of refactoring here to be made. Left this as is for now as out of the scope for this PR

.github/workflows/test-docs.yml
scopsy: I'm not sure i totally follow you question :( Could elaborate a bit? This addition of workflow dispatch was added so we can call this workflow manually from the UI and not only on PR's.

the job test will run only ones either on a PR or, manually triggered from UI. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

p-fernandez: My mistake when reviewing. I read it as `workflow_dispatch` was a workflow defined by us that executed some steps, and not as a property of the workflow as it is. That made me think that we were double executing steps. 😓

PR #2606, opened 2023-01-25
Nv 1568] 🐛 bug report wrong provider id stored in jobs collection #2577

### LINKED ISSUE #2577: [NV-1568] 🐛 Bug Report: Wrong providerId stored in jobs collection
### 📜 Description

When you enable multiple integrations(say discord and sendgrid) then for all jobs it is storing discord as providerId, even for email jobs. Another problem is for delay and digest jobs also storing same providerId(in this case discord).

### 👟 Reproduction steps

1. Create 2 integrations discord and sendgrid
2. Create an email notification 
3. Trigger the email notification created above and observe the jobs table, it is storing providerId as discord.

### 👍 Expected behavior

Expected to store sendgrid as providerId in jobs table

Storing wrong providerId discord in jobs table

### 🤖 Node Version

16.19

The problem is with this line:
channelType: ChannelTypeEnum[step.template.type] in src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
The problem here is using StepTypeEnum values doing reverse lookup in ChannelTypeEnum and that is always returning undefined as string enums does not support reverse lookup. 
Possible fix: we can step.template.type to upper case and /or  change both enums values to upper case.
This fix make sure to record providers for all channels but we still have problem with delay and digest steps, there also it stores one of these providers as providerId, I believe it should store null for those steps? for that we need to add additional logic to skip providerid for those steps?

<sub>[NV-1568](https://linear.app/novu/issue/NV-1568/🐛-bug-report-wrong-providerid-stored-in-jobs-collection)</sub>

tsssdev:

Fixed wrong providerid issue and also applied few performance fixes to avoid unnecessary DB calls.
With out performance fix it is issuing db queries for non channel types (like DELAY) and also issuing unnecessary queries for getting Novu integration, even after we got some valid integrations.. after fix it avoids all unnecessary calls on job creation side, as well as while sending messages.

Closes #2577


apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -113,14 +102,30 @@ export class ProcessSubscriber {
...
+        active: true,
+        organizationId: command.organizationId,
+        environmentId: command.environmentId,
+        userId: command.userId,
+      })
+    );
+    const integration = integrations[0];
+
+    return integration?.providerId ?? InAppProviderIdEnum.Novu;
+  }
```
LetItRock, line 128: please add the new line after }

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -113,14 +102,30 @@ export class ProcessSubscriber {
...
         ...(actorSubscriber && { _actorId: actorSubscriber._id }),
       });
     }
 
     return jobs;
   }
 
+  private async getProviderId(command: ProcessSubscriberCommand, stepType: StepTypeEnum): Promise<string | null> {
+    const channelType = ChannelTypeEnum[stepType.toUpperCase()];
+    if (!channelType) return null;
```
LetItRock, line 115: I feel like this might bring bugs here, but I'm not sure before we were always settings the `providerId` as `InAppProviderIdEnum.Novu` and now it can be null. And the actual type for `providerId` is `string | undefined`.

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -113,14 +102,30 @@ export class ProcessSubscriber {
...
+        providerId: providerId,
         ...(actorSubscriber && { _actorId: actorSubscriber._id }),
       });
     }
 
     return jobs;
   }
 
+  private async getProviderId(command: ProcessSubscriberCommand, stepType: StepTypeEnum): Promise<string | null> {
+    const channelType = ChannelTypeEnum[stepType.toUpperCase()];
```
LetItRock, line 114: I feel like it's better to create a map object instead of doing `.toUpperCase()`, it will be more type-safe

apps/api/src/app/integrations/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -19,7 +19,7 @@ export class GetNovuIntegration {
   ) {}
 
   async execute(command: GetNovuIntegrationCommand): Promise<IntegrationEntity | undefined> {
-    if (!process.env.NOVU_EMAIL_INTEGRATION_API_KEY) {
+    if (!process.env.NOVU_EMAIL_INTEGRATION_API_KEY || !command.channelType) {
```
LetItRock, line 22: it shouldn't be possible that there is no `command.channelType` unless it's null somehow... if it is, then please update the types

LetItRock:
hey @davidsoderberg, could you please take a look here as you have a fresh memory around the `GetDecryptedIntegrations` usecase ;)

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
tsssdev: For Null/undefined cases storing InAppProviderIdEnum.Novu does not make sense right? It is saving null fine for delay step, why you think it creates problem? should I change to undefined instead of null?

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
tsssdev: Could you explain which map object we need to create? you mean convert ChannelTypeEnum to Map?

davidsoderberg:
I think it looks good with the changes you suggest @LetItRock

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
LetItRock: It creates the TS error in line: 105, please change it to `undefined` and the return type to `Promise<string | undefined>`.
I've checked the rest of the code and it seems like it should be fine with `undefined`.

apps/api/src/app/integrations/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -51,6 +51,6 @@ export class GetDecryptedIntegrations {
       })
     );
 
-    return novuIntegration ? [novuIntegration] : integrations;
+    return [novuIntegration];
```
LetItRock, line 54: here is also TS error:
```
Type 'IntegrationEntity | undefined' is not assignable to type 'IntegrationEntity'.
  Type 'undefined' is not assignable to type 'IntegrationEntity'.ts(2322)

```

apps/api/src/app/integrations/usecases/get-novu-integration/get-novu-integration.usecase.ts
LetItRock: thanks for the clarification, let's leave it here, but please update the type of the `channelType` to be optional in the `GetNovuIntegrationCommand`

apps/api/src/app/integrations/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -38,7 +38,7 @@ export class GetDecryptedIntegrations {
         return integration;
       });
 
-    if (command.channelType === undefined) {
+    if (command.channelType === undefined || integrations.length > 0) {
```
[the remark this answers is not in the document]
tsssdev: @LetItRock This line is already covered by test cases in src/app/integrations/e2e/get-decrypted-integrations.e2e.ts('should get active decrypted integration') and src/app/integrations/e2e/get-integration.e2e.ts('should get newly created integration'), can you cross check and let me know, if we still need to add one more test case?

tsssdev:
@LetItRock  applied all changes suggested except test case(we already have test cases covering that line, please refer my comment), can you please review again and let me know if any other changes needed?

scopsy (approved):

tsssdev:
@scopsy is this released? I could not find this in v0.11.0 release changes but I see this in main(next) branch. Could you confirm? so I can delete this branch

scopsy:
@tsssdev not as part of 0.11.0 will be released as a patch in the next week 🙏

PR #2610, opened 2023-01-25
Feat: Add Logrocket Monitoring

Cliftonz:
Add LogRocket monitoring to get better analytics of what our users are using in the front-end.


apps/web/src/App.tsx
```diff
@@ -36,6 +36,13 @@ import { LinkVercelProjectPage } from './pages/partner-integrations/LinkVercelPr
 import { useBlueprint } from './hooks/useBlueprint';
 import { BrandPage } from './pages/brand/BrandPage';
 import { SegmentProvider } from './store/segment.context';
+import LogRocket from 'logrocket';
+import setupLogRocketReact from 'logrocket-react';
+
+if (LOGROCKET_ID) {
```
LetItRock, line 42: do we want to have it enabled during development?

apps/web/public/env-config.js
```diff
@@ -1,3 +1,4 @@
 window._env_ = {
   SKIP_PREFLIGHT_CHECK: 'true',
+  REACT_APP_ENVIRONMENT: 'dev',
```
LetItRock, line 3: as far as I know, this file is generated from the `.env` and we do override `REACT_APP_ENVIRONMENT` in the deploy pipeline... as far as I see it's already set in the `.env`

apps/web/.env
```diff
@@ -6,3 +6,4 @@ GLOBAL_CONTEXT_PATH=
 WEB_CONTEXT_PATH=
 REACT_APP_INTERCOM_APP_ID=
 REACT_APP_SEGMENT_KEY=
+LOGROCKET_ID='eh8t7r/novu'
```
LetItRock, line 9: shouldn't we set it from the pipeline?

scopsy: Yes, this should be removed from the .env file. And added to the deploy dev web and deploy prod web workflows on github actions.

apps/web/public/env-config.js
scopsy: Yes, totally correct. This file shouldn't be modified.

apps/web/src/App.tsx
scopsy: I don't think so, @LetItRock do you see any benefits of it?

package.json
```diff
@@ -48,7 +48,6 @@
     "build:web": "nx build @novu/web",
     "build:widget": "nx build @novu/widget",
     "build:embed": "nx build @novu/embed",
-    "build:inbound-mail": "nx build @novu/inbound-mail",
```
scopsy, line 51: Why was this removed from here?

scopsy: Oh nevermind, just seen this was a duplicate 😅

apps/web/src/App.tsx
Cliftonz: I don't think it has any benefit in dev, the analytics we need from this are only in prod.

scopsy (approved):

PR #2612, opened 2023-01-25
feat(package): add layout api to node package

p-fernandez:

Exposes Layout API in the Node package.

To keep aligned the API with the Novu package that our users install.


LetItRock (approved):
🚀

PR #2614, opened 2023-01-25
fix(package): fix tests and the implementation of the get topic

p-fernandez:

This PR introduces a breaking change in the Novu package. For the `novu.layouts.get('topic-key')` endpoint we change the response object.
Also fixes some tests and mocks that were done wrongly.

Currently the implementation of retrieving a topic by its key was calling behind the scenes the filter topics endpoint with a query param for the topic key. So the response was a paginated response. But our intention was that for a certain topic key to retrieve just the Topic DTO. Unfortunately due some changes we (I) missed that. This tries to fix it to make more sense of the API, but it is a breaking change. Due the short time we have exposed it, I consider we can do it in the new version even at the cost of annoying the users.


docs/docs/platform/topics.md
```diff
@@ -32,6 +32,18 @@ const result = await novu.topics.create({
 
 If successful this will return the internal ID generated by Novu and the topic key given by the user.
 
+That topic key can be used to retrieve the whole Topic entity:
```
p-fernandez, line 35: This info was missing from the docs. 🤦🏻 
Apologies, users.

packages/node/src/lib/topics/topics.spec.ts
```diff
@@ -87,19 +88,16 @@ describe('Novu Node.js package - Topics class', () => {
     };
 
     const mockedResponse = {
-      totalCount: 1,
-      page: 0,
-      pageSize: 10,
-      data: topic,
+      data: {
```
p-fernandez, line 91: We use Axios implementation, so users in the response they get all the information of the response including the property `data` that includes the response of the endpoint. In this case, the Topic DTO. Therefore I thought it was better to represent in the mocks the right response users receive when using the Novu package.

LetItRock (approved):
⭐

scopsy (approved):

PR #2617, opened 2023-01-26
fix: copy when there is no email integration active

davidsoderberg:
Change copy for when no email integration is active.

For it to not be misleading with the new predefined provider.

### DIFF
--- a/apps/web/src/components/templates/LackIntegrationError.tsx
+++ b/apps/web/src/components/templates/LackIntegrationError.tsx
@@ -4,14 +4,16 @@ import { Text } from '../../design-system';
 import { DoubleArrowRight } from '../../design-system/icons/arrows/CircleArrowRight';
 import { useNavigate } from 'react-router-dom';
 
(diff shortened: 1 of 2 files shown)


p-fernandez (approved):
🌟

apps/web/src/components/templates/LackIntegrationError.tsx
```diff
@@ -4,14 +4,16 @@ import { Text } from '../../design-system';
...
   const navigate = useNavigate();
 
   return (
     <>
       <WarningMessage>
         <Text>
-          {`Looks like you haven’t configured your ${channelType} provider yet, this channel will be disabled until you configure it.`}
+          {text
+            ? text
+            : `Looks like you haven’t configured your ${channelType} provider yet, this channel will be disabled until you configure it.`}
```
scopsy, line 16: Is this will be visible on the self-hosted version?

scopsy (approved):

PR #2626, opened 2023-01-27
Update README.md

Agathe-Brusset:
Add MailerSend as an email and SMS API provider.

Adding MailerSend to the README, the integration is already finished but wasn't added to the main documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Novu provides a single API to manage providers across multiple channels with a s
 - [x] [Mandrill](https://github.com/novuhq/novu/tree/main/providers/mandrill)
 - [x] [SendinBlue](https://github.com/novuhq/novu/tree/main/providers/sendinblue)
 - [x] [EmailJS](https://github.com/novuhq/novu/tree/main/providers/emailjs)
+- [x] [MailerSend](https://github.com/novuhq/novu/tree/main/providers/mailersend)
 - [ ] SparkPost
 
 #### 📞 SMS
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,6 +145,7 @@ Novu provides a single API to manage providers across multiple channels with a s
 - [x] [Telnyx](https://github.com/novuhq/novu/tree/main/providers/telnyx)
 - [x] [Termii](https://github.com/novuhq/novu/tree/main/providers/termii)
 - [x] [Gupshup](https://github.com/novuhq/novu/tree/main/providers/gupshup)
+- [x] [MailerSend](https://github.com/novuhq/novu/tree/main/providers/mailersend)
```
scopsy, line 148: @Agathe-Brusset I think mailersend only doing email right?

Agathe-Brusset: Hey @scopsy ! No, our API also allows sending SMS to US numbers.

p-fernandez: Unfortunately, the integration done is just for the email functionalities, so we can't add it to the SMS providers. 😞

Agathe-Brusset: Oh okay then, we can drop this line :D

p-fernandez: We will try in the future or with the support of the community to integrate the SMS functionalities, but that might require some work on our side as currently the provider IDs are unique and can't be repeated in integrations for different channels, such this case.

p-fernandez (approved):
🌟

Issue #2402, opened 2023-01-02
[NV-1522] 🏛️ Rename organisation


jainpawan21:
### 🔖 Feature description

Adding support of renaming organisation

### 🎤 Why is this feature needed ?

it is not possible to either delete or rename organisation. only option is to create new organisation

### ✌️ How do you aim to achieve this?

Implementing API first and then in Web

### 🔄️ Additional Information

*No response*

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

**Solution**

[Development / Workflows - Crafting updates (Figma)](https://www.figma.com/file/mxViLXxtC0PnB0ehXDcafw/Crafting-updates?type=design&node-id=2190%3A10883&mode=design&t=nRX8ll7wtl3h2uYs-1)

[Admin / Project profile  - Crafting updates (Figma)](https://www.figma.com/file/mxViLXxtC0PnB0ehXDcafw/Crafting-updates?type=design&node-id=2176%3A16990&mode=design&t=nRX8ll7wtl3h2uYs-1)

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-1522](https://linear.app/novu/issue/NV-1522/🏛️-rename-organisation)</sub>

Abobos:
@scopsy this looks like a quick win. I can work on it However, @jainpawan21 can provide more description. Thanks

scopsy:
Unfortuently we don't have a UI for this functionality yet 😦
But maybe you want to create an API endpoint for updating organizations?

jainpawan21:
@Abobos 
You can add this new endpoint here
https://github.com/novuhq/novu/blob/next/apps/api/src/app/organization/organization.controller.ts

Abobos:
@jainpawan21 , 

> @Abobos You can add this new endpoint here https://github.com/novuhq/novu/blob/next/apps/api/src/app/organization/organization.controller.ts

@jainpawan21 noted. @scopsy do we have to create a separate ticket for this?

jainpawan21:
No, No need you can create PR and reference this issue only.

Abobos:
[...]

PR #2634, opened 2023-01-28
feat: Implement rename organization

Abobos:

This PR introduce an endpoint to update the name of an organization

It is the API endpoint needed to implement [Rename Organization](https://github.com/novuhq/novu/issues/2402)


apps/api/src/app/organization/usecases/update-organization/update-organization-command.ts
```diff
@@ -0,0 +1,9 @@
+import { IsDefined } from 'class-validator';
+import { AuthenticatedCommand } from '../../../shared/commands/authenticated.command';
+
+export class UpdateOrganizationCommand extends AuthenticatedCommand {
+  @IsDefined()
+  public readonly id: string;
+
+  name: string;
```
scopsy, line 8: @Abobos due you think we should add here some maxLen for the organization name? Would use something permissive here, but maybe providing some limit will be a good idea to not break UI and etc...

apps/api/src/app/organization/organization.controller.ts
```diff
@@ -154,4 +157,15 @@ export class OrganizationController {
       })
     );
   }
+
+  @Put('/')
```
scopsy, line 161: Can we add the member role enum here: 
```
  @Roles(MemberRoleEnum.ADMIN)
```

scopsy (commented):
Looks great! Added a few small comments here let me know what you think

apps/api/src/app/organization/organization.controller.ts
Abobos: Yes. we want to restrict the update to only members with role `admin`?

apps/api/src/app/organization/usecases/update-organization/update-organization-command.ts
Abobos: Could you throw more perspective?.. How would this break this UI if we don't limit it?

apps/api/src/app/organization/organization.controller.ts
scopsy: Yes exactly 🙏

apps/api/src/app/organization/usecases/update-organization/update-organization-command.ts
scopsy: For example everywhere we show the organization name it can overflow (the settings dropdown, organization select) and maybe other places. This is quite and edge case to be honest. Really not a must.

apps/api/src/app/organization/organization.controller.ts
Abobos: @scopsy done

apps/api/src/app/organization/usecases/update-organization/update-organization-command.ts
Abobos: @scopsy yes. this makes sense.  done

apps/api/src/app/organization/organization.controller.ts
```diff
@@ -154,4 +157,16 @@ export class OrganizationController {
...
```
p-fernandez, line 161: As we are just modifying part of the resource `organization`, the HTTP method needs to be `PATCH`.
I would also recommend to rename the method name to `renameOrganization` to not ocuppy the name for a future organization update functionality that updates the whole resource. Same for the DAL repository method `updateOrganization`.

p-fernandez (changes requested):

apps/api/src/app/organization/organization.controller.ts
Abobos: Personally, I don't really subscribe to the fact that we are having different endpoints to update part of organization resource. I feel it should be a single endpoint with the possibility of updating part of the resource or the whole resource, that is why I am using the `name` column in this case. I can update the method to `PATCH`., but I think we should still keep the name as is. what do you think  @p-fernandez ?

p-fernandez: You are right now we don't have those endpoints, but in the future it is an option that I would like to have it open. 👍🏻 

Regarding the naming, when I read the method's name I would expect inside implements the functionality to update any modifiable property of the organization resource: name, logo, branding, partnerConfigurations (https://github.com/novuhq/novu/blob/4ae361d5676f007ece261698fe3447fccb9d9acf/libs/dal/src/repositories/organization/organization.schema.ts#L6)
As that is not what's happening giving a more accurate name of what the method is doing helps to any developer or user that goes through the code.

Abobos: Correct.. but the command shows it. The payload has only a `name` property, so one can still know that only `name` is being updated.

We can update the name if and only if we just want to use this method to update only name and not touch it in the future, i.e, updating other part of the organization resource in the future.

scopsy (approved):
🚀

p-fernandez (approved):
🌟

apps/api/src/app/organization/usecases/rename-organization/rename-organization-command.ts
```diff
@@ -0,0 +1,10 @@
+import { IsDefined, MaxLength } from 'class-validator';
+import { AuthenticatedCommand } from '../../../shared/commands/authenticated.command';
+
+export class RenameOrganizationCommand extends AuthenticatedCommand {
+  @IsDefined()
+  public readonly id: string;
+
+  @MaxLength(50)
```
p-fernandez, line 8: Just noticed this. I think it is better to not limit the max length because in the creation of organization we don't limit the length and it would be a misalignment. It is a good thing to have done that, but for now we can skip it until we commit to a decision to limit org name and implement it everywhere, also at database level. 👍

scopsy: I tend to aggree that it should be better implemented across the board, was initially thinking we can only implement it here. Let me revert that rule

PR #2646, opened 2023-01-30
feat(docs): add web app env variables and warning docs

p-fernandez:

Adds missing Web app information in the docs.

We have had some users struggling to understand Web app configuration by environment variables and how to apply the changes.


.markdownlint.jsonc
```diff
@@ -39,5 +39,10 @@
       "markdownlint",
       "Node.js"
     ]
+  },
+
+  // MD-46/code-block-style
+  "MD046": {
+    "style": "fenced"
```
p-fernandez, line 46: We use fenced block style all the time (backticked code blocks). `markdownlint` by default expects `indented` code blocks. I added the rule to avoid pre-commit to complain and fail.

apps/web/.env
```diff
@@ -1,8 +1,5 @@
 SKIP_PREFLIGHT_CHECK=true
 REACT_APP_API_URL=
-REACT_APP_WIDGET_SDK_PATH=
```
p-fernandez, line 3: Not used.

apps/web/.env
```diff
@@ -1,8 +1,5 @@
...
 REACT_APP_ENVIRONMENT=dev
-GLOBAL_CONTEXT_PATH=
```
p-fernandez, line 5: Not used in the Web app, as it is used to define paths in `/libs/shared`.

apps/web/.env
```diff
@@ -1,8 +1,5 @@
...
-WEB_CONTEXT_PATH=
```
p-fernandez, line 6: Not used yet, but we would need it in the future to allow different paths in the Docker solution and self hosted solution. Should I leave it even if it is not used yet?

scopsy: Let's remove it to avoid confusion

scopsy (approved):

PR #2647, opened 2023-01-30
feat: New design proposal for Novu Starting providers #2624

### LINKED ISSUE #2624: [NV-1592] 🚀 Feature: New design proposal for Novu Starting providers
### 🔖 Feature description

While working on other bugs I come across Novu starting email provider code and felt like this is adding extra complexity(some performance overhead) for main line of code multiple places(api, web etc.)..  and if we make any changes to main line of code we need to make sure this additional code not causing any issues and same way any modifications to this extra code not causing issues to main line. and later if we extend this to SMS, Chat etc. then it may add more complexity.. Another problem is these emails are stored as sendgrid emails, later if they use their own sendgrid provider then both stored as sendgrid provider.

I feel we can implement this logic with out effecting main line of code and simpler and more flexible way, I would like to present new approach for brainstorming, feel free to point out mistakes/shortcomings.

Here is the new design I propose:
We ca create new provider that can extend Sendgrid(or implement completely from scratch)  and when user enables this provider they simply provide preferred sender name in the UI(so not limited to organization name), apikey and from address we can read from environment, so user cant override these values, if they want to give different reply-to address then they can give in UI, so the replies goes to their email instead of coming to no-reply@novu.co .
And this is like any other provider, so if they don't have any email provider enabled then only they can enable this, and coming to count and limit logic, this can be implemented specific to this provider but I feel we can promote this to main line and expose this like feature for users, this will be very useful for organizations, they can set limits and track counts control their budgets, in case of attacks also this limit can help to mitigate the risk of getting big bill, so for all active providers we can expose this, not just start provider. if we need to extend this to SMS, CHAT, PUSH etc. also simple, just adding another provider. And later if we plan to start charging beyond free usage then we can do so..
I feel this is much cleaner approach as it is extending our system naturally(by adding new providers with out modifying main line code), feel free to point out if any mistakes in this approach or missed any edge cases.

Simplify the free provider logic.

Explained in the description.

### 🔄️ Additional Information

_No response_

<sub>[NV-1592](https://linear.app/novu/issue/NV-1592/🚀-feature-new-design-proposal-for-novu-starting-providers)</sub>

tsssdev:
Hey Guys, 
After discussing with David n his suggestion, I modified free/start provider code(basically removed all conditional code and made this free provider also like any other provider, so  very minimal code is required to add this free provider, we just need to add another provider and this can be easily and naturally extended, for other types like SMS, Chat etc..  Removed all old code and added new provider.
Major change in this PR is promoting the messages limit code to main line, now organisations can set the limit for any provider and if usage crosses the limit, we stop sending messages.. here I added two limits, softLimit and hardLimit, after reaching soft limit we start recording warning messages but still send messages, but after reaching hard limit we stop sending messages. For now users/organisations can select hard limit and softlimit calculating behind the scenes(90% of hard limit), I tried to give option to select both softlimit and hard limit using range slider but it looks like it may confuse the user, so revrted back to only hard limit selection and softlimit calculated in the code, if we can find a good UI component for selecting both limits we can modify UI code.. functionality all implemented, few test cases we may need to add here and there and .. please review this and give us your feedback.

Closes #2624 and global message limit feature across all providers..


providers/novu-email/src/lib/novuEmail.provider.ts
```diff
@@ -0,0 +1,33 @@
+import {
+  IEmailOptions,
+  ISendMessageSuccessResponse,
+  ICheckIntegrationResponse,
+  CheckIntegrationResponseEnum,
+} from '@novu/stateless';
+
+import { SendgridEmailProvider } from '@novu/sendgrid';
+
+export class NovuEmailProvider extends SendgridEmailProvider {
```
davidsoderberg, line 10: this was one of the things I dislike this package is useless to anyone else except Novus cloud version... My suggestion is putting this inside of the api server...

apps/api/src/app/integrations/usecases/provider-usage-limits/provider-usage-limits.usecase.ts
```diff
@@ -0,0 +1,56 @@
...
+      _organizationId: organizationId,
+      providerId: providerId,
+      createdAt: { $gte: startOfMonth(new Date()), $lte: endOfMonth(new Date()) },
+    });
+  }
+
+  async ensureLimitNotReached(integration: IntegrationEntity) {
+    if (!integration.limits) return;
+
+    const { softLimit, hardLimit } = integration.limits;
```
davidsoderberg, line 34: What is soft limit for can you explain the usecase for it?

apps/api/src/app/integrations/usecases/get-novu-integration/get-novu-integration.spec.ts
```diff
@@ -1,152 +0,0 @@
...
-import { MessageRepository, IntegrationRepository } from '@novu/dal';
-import { SharedModule } from '../../../shared/shared.module';
-import { IntegrationModule } from '../../integrations.module';
-import { GetNovuIntegration } from './get-novu-integration.usecase';
-import { GetNovuIntegrationCommand } from './get-novu-integration.command';
-import { endOfMonth, startOfMonth } from 'date-fns';
-import { CalculateLimitNovuIntegration } from '../calculate-limit-novu-integration/calculate-limit-novu-integration.usecase';
-import { ChannelTypeEnum, EmailProviderIdEnum } from '@novu/shared';
-
-describe('Get Novu Integration', function () {
```
davidsoderberg, line 15: these tests need to be moved somewhere else...

apps/api/src/app/integrations/usecases/provider-usage-limits/provider-usage-limits.usecase.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ProviderUsageLimitsCommand } from './provider-usage-limits.command';
+
+@Injectable()
+export class ProviderUsageLimits {
+  constructor(
+    private messageRepository: MessageRepository,
+    @Inject(ANALYTICS_SERVICE) private analyticsService: AnalyticsService
+  ) {}
+
+  static MAX_NOVU_INTEGRATION_MAIL_REQUESTS = parseInt(process.env.MAX_NOVU_INTEGRATION_MAIL_REQUESTS || '300', 10);
```
davidsoderberg, line 17: where is this used now?

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -20,7 +21,9 @@ export class UpdateIntegration {
   async execute(command: UpdateIntegrationCommand): Promise<IntegrationEntity> {
     const existingIntegration = await this.integrationRepository.findById(command.integrationId);
     if (!existingIntegration) throw new NotFoundException(`Entity with id ${command.integrationId} not found`);
-
+    if (NOVU_START_PROVIDERS.includes(existingIntegration.providerId)) {
+      command.limits = NOVU_START_PROVIDERS_LIMITS;
```
davidsoderberg, line 25: This needs to change when we add another novu provider because it can be different limits for different providers.

apps/web/src/components/templates/email-editor/EmailContentCard.tsx
```diff
@@ -92,11 +88,7 @@ export function EmailContentCard({
       {!isIntegrationActive ? (
         <LackIntegrationError
           channelType="E-Mail"
-          text={
-            enabled
-              ? 'Looks like you haven’t configured your E-Mail provider yet, visit the integrations page to configure.'
-              : undefined
-          }
+          text={'Looks like you haven’t configured your E-Mail provider yet, visit the integrations page to configure.'}
```
davidsoderberg, line 91: this is a bit problematic for the self hosted version because the other text gives your more info if it is self hosted then this one does

davidsoderberg:
@tsssdev would be great if you could provide a screen recording(loom) or print screens of how the ui now looks...
In general I feel that the code have become more shattered all over the place...

apps/api/src/app/integrations/usecases/provider-usage-limits/provider-usage-limits.usecase.ts
tsssdev: The idea here is, if we have one limit then after reaching the limit we stop sending messages, instead of that if we have soft limit, after crossing soft limit we can just give warning but still send messages, after reaching hard limit we stop sending message, like soft limit is a warning and they can extend limit if needed after receiving this warning.

providers/novu-email/src/lib/novuEmail.provider.ts
tsssdev: If we keep it outside providers package then again we need to write custom logic, if we keep it here... , even though many guys not using it is like any other provider. already we have many providers and many guys they do not use many providers and they use which ever providers they need/interested... Lest assume this provider is for cloud and end user is Novu cloud

p-fernandez:
@tsssdev first, thanks for this proposal and this big effort that helps to explain it.

Going through it, what I find is that there are two different features merged in this proposal. 

One being the soft/hard limits to help users to be aware of their usage of the different providers to help them to control their billing. This is a good suggestion and a step in very useful functionalities.

The other one seems to be to create another abstraction layer on top of the Sendgrid provider and move it as a new provider. I am a bit lost in the part: _and this can be easily and naturally extended, for other types like SMS, Chat etc.._
The limited trial provider is a _gift_ we implemented to allow users in the Novu cloud hosted solution to quickly try Novu without having to set up any integration or sign up for a service, so they can navigate through the features and try the product. It is disallowed in self host cloud version for obvious reasons (as we cover with its spending). Also so far we have no plans to provide other limited trial providers for the other channels (SMS, push).

My suggestion would be to split this proposal in 2 as I can spot very differentiated goals:
1 - A proposal for implementing usage features that can be aligned with a potential future billing control feature to help our users to manage their spending and also get accurate reports of their spending with us.
2 - A proposal regarding the implementation of the limited trial Novu provider addressing your concerns regarding performance and complexity, that you mention in https://github.com/novuhq/novu/issues/2624, but you haven't listed. Though without diverting our intention with the limited trial provider.

tsssdev:
@p-fernandez, Actually it has 3 parts, 
1. Removing old code related to start provider,
2. Adding new start provider for email(extending sendgrid is not mandatory)
3. Limits logic,with soft and hard limits, most of the code is related to this.

All code removals/cleanup is related to #1 and most of the new code is related to #3 and for #2 just added one new provider(novu-email) and is like any new provider with out any special logic in our main code, just added logic not to override the limits for these start providers. 

Coming to splitting, actually #2 needs some limit logic, so instead of doing separately I thought doing together make sense and reduces the amount of work required, as I mentioned for #2 not much logic involved., Hope now this is more clear and you can get complete picture. I think if you grasp limit logic, things become more clearer

Coming to extending sendgrid, that is actually not mandatory, just trying to re-use few lines of code, if needed we can remove sendgrid dependency and add that email data construction in this provider, so it will be completely independent,  and coming to other providers SMS etc.. I just mentioned it is easy to extend in this new approach with out violating OCP,  just for email also I feel this approach is better.. coming to adding other start/gift providers, if cost is concern, we can do chat provider like telegram, that is complete free of cost and useful for some organizations.

Please let me know if any other clarifications needed.

tsssdev:
> @tsssdev would be great if you could provide a screen recording(loom) or print screens of how the ui now looks... In general I feel that the code have become more shattered all over the place...

Sure @davidsoderberg I will add few screenshots, coming to code scattering, the most of the code here is for limits logic, for start provider except new provider, not much code added, just added code to block limits overriding. Please refer this https://github.com/novuhq/novu/pull/2647#issuecomment-1411627927

tsssdev:
@davidsoderberg please find the UI screenshots, if any modifications needed I can apply, this draft PR is for showcaseing my proposal, if you guys agree with this we can refine further taking all feedback, backend(api), providers side all done including test cases, but web side need to add test cases and may be some UI enhancements needed.. also if needed we can hide this on non cloud environments, alreday we have check for free providers(to disable limit set slider, we can add one more check to show based on environment., Please let me know if any other details needed.. as I mentioned my eralier comment this draft PR has 3 parts... removing old code, adding new free provider and limit logic for all providers.

limit-set
messages-usage

tsssdev:
@p-fernandez @davidsoderberg 
For making review easier I have created new branch and committed  3 parts individually in 3 different commits in this branch https://github.com/novuhq/novu/compare/next...tsssdev:novu:NV-1592-Feature-New-design-proposal-for-Novu-Starting-providers-individual-commits

**Commit 1  code cleanup of old start/free provider:** As part of this commit removed old start/free provider code (You can look at this and see how many files modified and complexity in the code and the extra slight performance overhead introduced on each message send)

**Commit 2 Added novu-email provider:** Just added another provider to use it in our cloud, nothing special or conditional code specific to free provider here, it is just like any other provider and no conditional/special logic added for free provider.

**Commit 3 Limit logic for all channels and providers:**  This commit is mainly for implementing limit logic across all channels and all providers, including the test cases.

Done.. now we got **limit logic(new feature)** and **got free provider** and with out adding any extra code in the main line of code and **zero performance overhead** on message send. Just added 3 to 4 lines of code to not to allow users to overwrite the limits of free providers, no complex code and that too that logic is in not frequently hit place, not part of message sending core logic, so zero extra overhead on message sending logic.

Please let me know if you guys need any other clarifications.

scopsy:
Following our previous conversatin regarding introducing changes in smaller PR's. Will close this one for now 🙏

PR #2659, opened 2023-01-31
fix(infra): remove standard-version package as being unused

p-fernandez:

Self explanatory.

Unused dependency. Also it is deprecated. From security and development perspective is a good thing to do.


scopsy (approved):
👋

PR #2662, opened 2023-02-01
feat(logrocket): add request sanitizing and add username& email to sessions

Cliftonz:


apps/web/src/App.tsx
```diff
@@ -40,7 +40,28 @@ import LogRocket from 'logrocket';
 import setupLogRocketReact from 'logrocket-react';
 
 if (LOGROCKET_ID && window !== undefined) {
-  LogRocket.init(LOGROCKET_ID);
+  LogRocket.init(LOGROCKET_ID, {
+    release: 'v0.11.0',
```
scopsy, line 44: This shouldn't be hardcoded like this here. Should be extracted from the package.json or injected in any other dynamic way. (env variables for example) it can get out of sync quite fast

apps/web/src/components/auth/LoginForm.tsx
```diff
@@ -59,6 +61,13 @@ export function LoginForm({ email, token }: Props) {
     try {
       const response = await mutateAsync(itemData);
       setToken((response as any).token);
+
+      const user = await getUser();
```
scopsy, line 65: I would suggest not doing LogRocket identify as part of a login form, since the user can be logged in from a lot of places. (OAuth, signup and etc...) 

You can move this to be part of the authentication hook where we set the user. Or similarly to intercom put it in the high level layout section such as `HeaderNav.tsx` (probably not the best place but at least will work for other sign in methods

.idea/misc.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JavaScriptSettings">
+    <option name="languageLevel" value="FLOW" />
```
scopsy, line 4: If i recall correctly this is configured for "flow" (ts alternative) which we don't use. Do you know how this was generated?

apps/web/src/App.tsx
```diff
@@ -40,7 +40,28 @@ import LogRocket from 'logrocket';
...
+    rootHostname: 'novu.co',
```
scopsy, line 45: Can this be extracted dynamically?

.idea/misc.xml
Cliftonz: No, but I will remove it.

apps/web/src/components/layout/components/HeaderNav.tsx
```diff
@@ -49,6 +50,21 @@ export function HeaderNav({}: Props) {
     }, [currentUser, currentOrganization]);
   }
 
+  if (LOGROCKET_ID) {
+    useEffect(() => {
+      if (currentUser && currentOrganization) {
+        let email = currentUser?.email;
+        email ??= ' ';
```
scopsy, line 57: What is this `??=` never seen such syntax in js. 

Also why are we doing an empty space string here instead of `''`? Would prefer to leave it as null/undefined or even empty string so it will be considered a falsy value.

```suggestion
        const email = currentUser?.email ?? '';
```

apps/web/src/App.tsx
```diff
@@ -38,9 +38,31 @@ import { BrandPage } from './pages/brand/BrandPage';
 import { SegmentProvider } from './store/segment.context';
 import LogRocket from 'logrocket';
 import setupLogRocketReact from 'logrocket-react';
+import packageJson from '../package.json';
```
scopsy, line 41: Did you had a chance to check if in some case this is affecting the build/dist folder structure of the web?

apps/web/src/components/layout/components/HeaderNav.tsx
p-fernandez: It is the nullish coalescing operator. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing
It is useful to check for existence without being triggered by 0 or false.

Cliftonz: The reason I did this is that logrocket traits does not support undefined value which '' is also considered undefined.

apps/web/src/App.tsx
Cliftonz: What do you mean?

LetItRock: @Cliftonz his suspicion was that the `package.json` won't be available after the build... but I've checked it and actually the `package.json` file is bundled to the app as the JSON string and then parsed... it just adds ~5.14kb to the bundled main JS chunk...

LetItRock: I don't think so

apps/web/src/components/layout/components/HeaderNav.tsx
```diff
@@ -49,6 +50,31 @@ export function HeaderNav({}: Props) {
...
+            name: currentUser?.firstName + ' ' + currentUser?.lastName,
+            organizationId: currentOrganization._id,
+            organization: currentOrganization.name,
+          };
+        }
+
+        LogRocket.identify(currentUser?._id, logrocketTraits);
+      }
+    }, [currentUser, currentOrganization]);
+  }
```
LetItRock, line 76: please move the `if (LOGROCKET_ID)` to be inside the `useEffect`

```suggestion
    useEffect(() => {
      if (!LOGROCKET_ID) return;
      
      if (currentUser && currentOrganization) {
        let logrocketTraits;

        if (currentUser?.email !== undefined) {
          logrocketTraits = {
            name: currentUser?.firstName + ' ' + currentUser?.lastName,
            organizationId: currentOrganization._id,
            organization: currentOrganization.name,
            email: currentUser?.email ? currentUser?.email : ' ',
          };
        } else {
          logrocketTraits = {
            name: currentUser?.firstName + ' ' + currentUser?.lastName,
            organizationId: currentOrganization._id,
            organization: currentOrganization.name,
          };
        }

        LogRocket.identify(currentUser?._id, logrocketTraits);
      }
    }, [currentUser, currentOrganization]);
```

apps/web/src/App.tsx
Cliftonz: @scopsy This should be good to merge then

scopsy (approved):

scopsy:
@Cliftonz let's fix the prettier errors and feel free to merge once test passed 🎉

PR #2669, opened 2023-02-02
Updating Cypress from version 12.1.0 to 12.3.0

RobertoCassino:

This PR updates the cypress at apps/web/package.json (line 129) from version 12.1.0 to version 12.3.0.

4 CVEs were found in novu, the version of the Cypress testing framework is the root cause here, as the version used is 12.1.0 and it includes and old version of the packages like simple-git and others.


p-fernandez (commented):
Mind you to update it also in `/apps/widget` so we have the versions used in both apps aligned?

scopsy:
Let's see of tests will pass, last time i tried updating widget it didn't worked out :(

apps/widget/package.json
```diff
@@ -69,7 +69,7 @@
     "@types/react-router-dom": "^5.1.7",
     "craco-antd": "^1.19.0",
     "cross-env": "^7.0.3",
-    "cypress": "10.11.0",
+    "cypress": "12.3.0",
```
p-fernandez, line 72:
```suggestion
    "cypress": "^12.3.0",
```
So we both get it updated (web and widget) and just one single version in the dependency all the time.

p-fernandez (approved):
🌟

RobertoCassino:
> 🌟

Thanks @p-fernandez for approval, the end to end test for widget is still failing and I am not sure what the problem is for the api migration. We focused on replacing `.should()` with `.then()` in accordance to the manual provided by Cypress https://docs.cypress.io/guides/references/migration-guide#should. The test is now failing at `getNotifications()` on line 161 in `apps/widget/cypress/e2e/shell-embed.spec.ts`. This is weird, as the function is being used in multiple tests which are passing successfully. 

The exact point of failure is on line 167 where is trying to execute `cy.wrap(body).find(`[data-test-id="notification-list-item"]`);`. What would the next steps be?

p-fernandez:
> > 🌟
> 
> Thanks @p-fernandez for approval, the end to end test for widget is still failing and I am not sure what the problem is for the api migration. We focused on replacing `.should()` with `.then()` in accordance to the manual provided by Cypress https://docs.cypress.io/guides/references/migration-guide#should. The test is now failing at `getNotifications()` on line 161 in `apps/widget/cypress/e2e/shell-embed.spec.ts`. This is weird, as the function is being used in multiple tests which are passing successfully.
> 
> The exact point of failure is on line 167 where is trying to execute `cy.wrap(body).find(`[data-test-id="notification-list-item"]`);`. What would the next steps be?

We might need to revisit the test. Just double checking it is not a timeout issue from Cypress.

LetItRock:
@RobertoCassino I've fixed failing widget tests, hope they will pass now ;)

LetItRock:
@RobertoCassino @p-fernandez @scopsy
I see that the test is failing because of the race conditions in the cache, I'll inform @djabarovgeorge

scopsy:
@LetItRock seems like it's a flaky tests that also happens regardless of this PR. My suggestion is to merge this one and deal with the race condition in another PR. WDYT?

scopsy:
@pieterge end of the month are the usual monthly cadence, however I plan to release a hotfix release until the end of the week 🙏

Issue #2642, opened 2023-01-30
[NV-1608] 🐛 Bug Report: `subject` as a variable is not working in subject


chavda-bhavik:
### 📜 Description

We have the facility to change variables like below.
![image](https://user-images.githubusercontent.com/50201755/215398675-b1aeddf0-ed83-4f11-a52b-78f4939dc5dc.png)
But variables are not getting replaced in `subject`.

### 👟 Reproduction steps

1. Create a notification
2. Add `Email` in the notification template.
3. Edit the `Email` and define the variable in a subject like, `Hello {{name}}` with any content.
4. Trigger the email with a variable value in payload defined like, `{ "name": "John" }`
5. See the received email

Received email will be having subject as `Hello {{name}}` even though `name` is passed in the payload.

### 👍 Expected behavior

Payload variables should be replaced in the `subject` of the email.

### 👎 Actual Behavior with Screenshots

It send email with a subject like `Hello {{name}}`

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

<sub>[NV-1608](https://linear.app/novu/issue/NV-1608/🐛-bug-report-variables-in-subject-are-not-getting-replaced)</sub>

oba2311:
@LetItRock  / @scopsy  - how long has this been not working? 🤔

scopsy:
Need to reproduce this, to make sure. There supposed to be a test case for this, so it would be very strange if it's not working. Wll need to investigate

scopsy:
@ainouzgali could add here you latest findings regarding this ticket?

jainpawan21:
{{subject}} is not working, rest other variables are working.
Bhavik said he wants to work on this

PR #2676, opened 2023-02-05
Fix: Subject named variable not assigning

chavda-bhavik:
variable with the `subject` was getting overridden in the `compile-template usecase`. So, I removed the line overriding the `subject` variable.

#2642 

Added a test case covering the `subject` variable is getting replaced properly in `email subject` and `email body`.


ainouzgali (approved):
Thanks @chavda-bhavik ! Great catch!

PR #2691, opened 2023-02-07
feat: exclude actor from topic event

ainouzgali:

To exclude the actor from getting the events, exclude him from the subscribers of the topic. 

As usually, the actor of an action will not expect to receive notification on the action they just performed.


scopsy (approved):

scopsy:
@ainouzgali could you please also mention this behaviour in the docs section?

PR #2694, opened 2023-02-07
fix: missing indexes for count query

scopsy:

Add environmentId to the query to avoid collections scans on it and limited the amount of objects scanned until reached limit

The current query is not utilizing indexes properly since an index on organizationId does not exists


scopsy:
@davidsoderberg I know this will make it per environment and not per organization, but due to indexing and performance issue we must make this change. Queries here are taking 60 seconds in some cases

davidsoderberg:
> @davidsoderberg I know this will make it per environment and not per organization, but due to indexing and performance issue we must make this change. Queries here are taking 60 seconds in some cases

It is okay as long as we are aware of now giving 1000 emails each month :)

davidsoderberg (approved):

scopsy:
@davidsoderberg why 1000 and not 600 total?

davidsoderberg:
> @davidsoderberg why 1000 and not 600 total?

haha 600 I mean, did read the changed test right before writing the comment :P

PR #2701, opened 2023-02-08
NV-1582 bug report node package does not allow

### LINKED ISSUE #2603: [NV-1582] 🐛 Bug Report: Node package does not allow data key in overrides
### 📜 Description

Trying to pass a data key in `fcm` overrides as described by the [docs](https://docs.novu.co/channels/push/fcm) causes type errors because of a missing data key in `ITriggerOverrideFCM` (see [here](https://github.com/novuhq/novu/blob/next/packages/node/src/lib/events/events.interface.ts#L64)).

You get `Object literal may only specify known properties, and 'data' does not exist in type 'ITriggerOverrideFCM`.

### 👟 Reproduction steps

Attempt to use the `data` key as described by the docs:

```

const novu = new Novu(process.env.NOVU_API_KEY);

novu.trigger('event-name', {
  to: {
    subscriberId: '...',
  },
  payload: {
    abc: 'def', // If the notification is a data notification, the payload will be sent as the data
  },
  overrides: {
    fcm: {
      type: 'notification',
      // type: 'data' => will turn this into a FCM data notification, where the payload is sent as a data notification
      data: {
        key: 'value',
        // If type is "notification", you can use the "data" override to send notification messages with optional data payload
      },
    },
  },
});
```

### 👍 Expected behavior

It should not produce type errors.

It produces type errors.

_No response_

None

<sub>[NV-1582](https://linear.app/novu/issue/NV-1582/🐛-bug-report-node-package-does-not-allow-data-key-in-overrides)</sub>

scopsy:
- Added a new data property to the overrides of fcm
- Made override keys as optional
- Added a test for it

This closes #2603

### DIFF
--- a/packages/node/src/lib/events/events.interface.ts
+++ b/packages/node/src/lib/events/events.interface.ts
@@ -75,6 +75,7 @@ export type ITriggerOverrideFCM = {
   clickAction?: string;
   titleLocKey?: string;
   titleLocArgs?: string;
+  data?: Record<string, any>;
 };
 
 export type IAPNSAlert = {
--- a/packages/node/src/lib/events/events.interface.ts
+++ b/packages/node/src/lib/events/events.interface.ts
@@ -47,13 +47,13 @@ export type ITriggerOverrides = {
     | 'plivo'
     | 'postmark'
     | 'sendgrid'
-    | 'twilio']: object;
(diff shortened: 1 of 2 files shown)


BiswaViraj (approved):
🌟

PR #2702, opened 2023-02-08
feat: add in app activated api support

djabarovgeorge:

adds a new endpoint on the API.

we need it in order to help us indicate if a new user initialized the in-app on this side.

### DIFF
--- a/apps/api/src/app/organization/organization.controller.ts
+++ b/apps/api/src/app/organization/organization.controller.ts
@@ -33,6 +33,8 @@ import { IGetOrganizationsDto } from './dtos/get-organizations.dto';
 import { GetMyOrganization } from './usecases/get-my-organization/get-my-organization.usecase';
 import { GetMyOrganizationCommand } from './usecases/get-my-organization/get-my-organization.command';
 import { IGetMyOrganizationDto } from './dtos/get-my-organization.dto';
+import { GetInAppActivated } from './usecases/get-In-app-activated/get-In-app-activated.usecase';
+import { GetInAppActivatedCommand } from './usecases/get-In-app-activated/get-In-app-activated.command';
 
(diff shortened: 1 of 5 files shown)


apps/api/src/app/organization/organization.controller.ts
```diff
@@ -154,4 +157,14 @@ export class OrganizationController {
       })
     );
   }
+
+  @Get('/in-app/activated')
```
djabarovgeorge, line 161: I am still not sure under what controller to add validation of the in-app initialization. Because neither the organization nor the subscriber feels entirely correct. it is kinda related to both of them and neither of them.
will love to hear new perspectives.

apps/api/src/app/organization/usecases/get-In-app-activated/get-In-app-activated.usecase.ts
```diff
@@ -0,0 +1,18 @@
...
+
+@Injectable()
+export class GetInAppActivated {
+  constructor(private readonly subscriberRepository: SubscriberRepository) {}
+
+  async execute(command: GetInAppActivatedCommand): Promise<boolean> {
+    const inAppSubscriberCount = await this.subscriberRepository.find({
+      _organizationId: command.organizationId,
+      isOnline: { $exists: true },
+      subscriberId: { $not: /cli-/ },
```
djabarovgeorge, line 13: This part is good because I added additional code on the CLI package to the subscriber id, **but** it will be updated to equal to the one from snippets on the web client of the subscriber id (which will be reflated here)

apps/api/src/app/organization/organization.controller.ts
djabarovgeorge: Moved it to integration controller

PR #2711, opened 2023-02-09
Nv 1576 make default layout changes dependent

BiswaViraj:

- Makes default layout changes dependent
- Created a new change type: `DefaultLayout`
- 

image

### DIFF
--- a/apps/api/src/app/layouts/usecases/create-layout/create-layout.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/create-layout/create-layout.use-case.ts
@@ -39,9 +39,9 @@ export class CreateLayoutUseCase {
         userId: dto._creatorId,
       });
       await this.setDefaultLayout.execute(setDefaultLayoutCommand);
+    } else {
+      await this.createChange(command, dto._id);
     }
-
-    await this.createChange(command, dto._id);
 
     return dto;
   }
--- a/apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
+++ b/apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
@@ -26,14 +26,7 @@ export class SetDefaultLayoutUseCase {
     );
 
     if (!existingDefaultLayoutId) {
-      await this.createDefaultChange(
-        CreateDefaultLayoutChangeCommand.create({
-          environmentId: command.environmentId,
-          layoutId: command.layoutId,
-          organizationId: command.organizationId,
-          userId: command.userId,
-        })
-      );
+      await this.createDefaultChange(command);
 
       return;
     }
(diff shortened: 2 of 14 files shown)


apps/api/src/app/layouts/usecases/create-layout/create-layout.use-case.ts
```diff
@@ -39,10 +39,10 @@ export class CreateLayoutUseCase {
         userId: dto._creatorId,
       });
       await this.setDefaultLayout.execute(setDefaultLayoutCommand);
+    } else {
+      await this.createChange(command, dto._id);
```
ainouzgali, line 43: I wonder what would happen if there is an update/create of layout(name,content) + setting to default. Would it create 2 changes? One for default layout change and one for layout change? If so, are we sure they are also dependent on promote?

ainouzgali (commented):
Great job😍 Wrote a few questions.

apps/api/src/app/layouts/usecases/create-layout/create-layout.use-case.ts
BiswaViraj: In that case, It will create a single(`1`) change under `DefaultLayout`,
This way it maintains the dependency between other layout changes too.

apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
```diff
@@ -1,40 +1,65 @@
...
+      await this.createDefaultChange(
+        CreateDefaultLayoutChangeCommand.create({
+          environmentId: command.environmentId,
+          layoutId: command.layoutId,
+          organizationId: command.organizationId,
+          userId: command.userId,
+        })
+      );
+
       return;
```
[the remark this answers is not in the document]
BiswaViraj: For the very first layout, the `existingDefaultId` will be false,
This will handle that scenario

apps/api/src/app/layouts/usecases/set-default-layout/set-default-layout.use-case.ts
```diff
@@ -93,4 +109,27 @@ export class SetDefaultLayoutUseCase {
   ): Promise<void> {
     await this.layoutRepository.updateIsDefault(layoutId, environmentId, organizationId, isDefault);
   }
+
+  private async createDefaultChange(command: CreateDefaultLayoutChangeCommand) {
+    const createLayoutChangeCommand = CreateDefaultLayoutChangeCommand.create({
```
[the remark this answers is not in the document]
BiswaViraj: 😅 Removing them

ainouzgali:
Looks good! Then my only concern is the message in the changes table. It shows it only as one row with Default Layout Change and just one of the layouts (I think the one that used to be default) with all the fields changes, not only the default change. Also, it will include changes made in between to another template - I just worry it will be a bit confusing for users to understand what they are promoting.

apps/web/src/components/changes/ChangesTableLayout.tsx
```diff
@@ -70,6 +70,9 @@ export const ChangesTable = ({
           {type === ChangeEntityTypeEnum.LAYOUT && (
             <Text color={colorScheme === 'dark' ? colors.B40 : colors.B70}>Layout Change</Text>
           )}
+          {type === ChangeEntityTypeEnum.DEFAULT_LAYOUT && (
+            <Text color={colorScheme === 'dark' ? colors.B40 : colors.B70}>Default Layout Change</Text>
+          )}
```
LetItRock, line 75: same

LetItRock (approved):

PR #2713, opened 2023-02-09
fix: so previous change id is used on delete change for notification template

davidsoderberg:
Same behaviour as on update notification template

To fix so template is deleted in prod env when using promote all


ainouzgali (approved):
Is it possible to add a test? Or will it be too flaky?

davidsoderberg:
> Is it possible to add a test? Or will it be too flaky?

I am looking into this right now 😄

PR #2719, opened 2023-02-09
fix(api): implement distributed lock to fix digest concurrency issues

p-fernandez:

Fixes the Digest implementation in a way that avoids when concurrent calls the notifications are not duplicated anymore and the digest is working as expected.
For that takes advantage of a distributed lock implementation with Redlock (Redis based).

We were having multiple reports of Digest not working as expected. This is one step forward to fix some of the problems addressed.


apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -52,13 +54,13 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+    workflowQueueService = session.testServer?.getService(WorkflowQueueService);
 
     runJob = new RunJob(
       jobRepository,
-      session.testServer.getService(SendMessage),
-      session.testServer.getService(QueueNextJob),
-      session.testServer.getService(StorageHelperService)
+      session.testServer?.getService(SendMessage),
+      session.testServer?.getService(QueueNextJob),
+      session.testServer?.getService(StorageHelperService)
```
p-fernandez, line 63: Linter warnings.

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -1046,4 +1044,122 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+      }),
+      triggerEvent({
+        customVar: 'concurrent-call-4',
+      }),
+      triggerEvent({
+        customVar: 'concurrent-call-5',
+      }),
+    ]);
+
+    await session.awaitRunningJobs(template?._id, false, 5);
```
p-fernandez, line 1085: We trigger 5 events so we would expect 5 running jobs, therefore the third param of this helper, `unfinishedJobs` is set to 5 to move on with the test execution.

apps/api/src/app/events/usecases/add-job/add-delay-job.usecase.ts
```diff
@@ -74,8 +74,8 @@ export class AddDelayJob {
    */
   private async noExistingDelayedJobForDate(
     data: JobEntity,
-    currDelayPath: string,
-    currDelayDate: string
+    currentDelayPath: string,
+    currentDelayDate: string
```
p-fernandez, line 78: To avoid `cSpell` complaining.

apps/api/src/app/events/usecases/add-job/add-digest-job.usecase.ts
```diff
@@ -1,55 +1,100 @@
...
+    const shouldDelayDigestDigestJobOrMerge = async () => this.jobRepository.shouldDelayDigestJobOrMerge(job);
+
+    return await applyLock<IFindAndUpdateResponse>(
+      {
+        resource,
+        ttl: TTL,
+      },
+      shouldDelayDigestDigestJobOrMerge
+    );
+  }
```
p-fernandez, line 86: Lock application. We surround the starting point of code we want to lock, so any call to that point of code, if the lock is taken, will be halted and retried once the lock is being freed.

apps/api/src/app/events/usecases/add-job/add-digest-job.usecase.ts
```diff
@@ -1,55 +1,100 @@
...
 
-    return AddJob.toMilliseconds(data.digest.amount, data.digest.unit);
+    return undefined;
+  }
+
+  private async shouldDelayDigestOrMergeWithLock(job: JobEntity): Promise<IFindAndUpdateResponse> {
+    const resource = `environment:${job._environmentId}:template:${job._templateId}:subscriber:${job._subscriberId}`;
+    const TTL = 250;
+
+    const shouldDelayDigestDigestJobOrMerge = async () => this.jobRepository.shouldDelayDigestJobOrMerge(job);
```
p-fernandez, line 77: The reason of setting the arrow function like this instead directly as an argument is to have the function named and therefore use the name of the functions for the logs inside of the distributed lock to identify in the future potential failures.

apps/api/src/app/events/usecases/add-job/add-digest-job.usecase.ts
```diff
@@ -1,55 +1,100 @@
...
+
+      return AddJob.toMilliseconds(digest.amount, digest.unit);
     }
 
-    return AddJob.toMilliseconds(data.digest.amount, data.digest.unit);
+    return undefined;
+  }
+
+  private async shouldDelayDigestOrMergeWithLock(job: JobEntity): Promise<IFindAndUpdateResponse> {
+    const resource = `environment:${job._environmentId}:template:${job._templateId}:subscriber:${job._subscriberId}`;
```
p-fernandez, line 74: Opinionated structure. But the key params that identify if we have to merge digests that happen at the same time is based on these 3 values: `environmentId`, `templateId` and `subscriberId`.
If not using them we would expect faulty behaviour of the lock and the digest implementation.

apps/api/src/app/events/usecases/add-job/add-digest-job.usecase.ts
```diff
@@ -1,55 +1,100 @@
...
+    const TTL = 250;
```
p-fernandez, line 75: Arbitrary. Can be tuned.

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
```diff
@@ -27,14 +30,17 @@ export class AddJob {
...
+        ? await this.addDigestJob.execute(AddDigestJobCommand.create({ job }))
+        : undefined;
+    const delayAmount = job.type === StepTypeEnum.DELAY ? await this.addDelayJob.execute(command) : undefined;
 
     if (job.type === StepTypeEnum.DIGEST && digestAmount === undefined) {
       return;
     }
 
-    if (digestAmount === undefined && delayAmount == undefined) {
+    if (digestAmount === undefined && delayAmount === undefined) {
```
p-fernandez, line 43: This was simplified based on we are returning now `undefined` in the negate brace of the ternaries of calculating `digestAmount` and `delayAmount`.

apps/api/src/app/events/usecases/add-job/should-add-digest-job.usecase.ts
```diff
@@ -1,28 +0,0 @@
...
-@Injectable()
-export class ShouldAddDigestJob {
-  constructor(private jobRepository: JobRepository) {}
-
-  public async execute(command: AddJobCommand): Promise<boolean> {
-    const data = await this.jobRepository.findById(command.jobId);
-
-    const isValidDigestStep = data.type === StepTypeEnum.DIGEST && data.digest.amount && data.digest.unit;
-    if (!isValidDigestStep || !data) {
-      return true;
```
p-fernandez, line 15: Arbitrarily I decided to remove this use case and integrate the logic in the `AddDigestJob` use case to be able to simplify the code. We were hitting the database extra times without any need (in my opinion for this flow) and therefore having to do many validations that were already checked, just because in the case this use case was used standalone.

apps/api/src/bootstrap.ts
```diff
@@ -39,6 +40,8 @@ if (process.env.SENTRY_DSN) {
 validateEnv();
 
 export async function bootstrap(expressApp?): Promise<INestApplication> {
+  startupDistributedLock();
```
p-fernandez, line 43: We add the initalisation of the distributed lock.
We should add the shutdown of the distributed lock (implemented but not integrated). @scopsy  @Cliftonz where do we do the shutdown of the Nest server? I can't find it.

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+let distributedLock: Redlock;
+let lockCounter = 0;
+
+export const startup = (
+  settings: Settings = {
+    automaticExtensionThreshold: 500,
+    driftFactor: 0.01,
+    retryCount: 50,
+    retryDelay: 100,
+    retryJitter: 200,
```
p-fernandez, line 23: Default settings (as recommended by Redlock).

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+): void => {
+  if (distributedLock) {
+    return;
+  }
+
+  // TODO: Implement distributed nodes (at least 3 Redis instances)
+  const rawInstances = [redisUrl].filter((instance) => !!instance).map((url) => new Redis(url));
```
p-fernandez, line 31: This needs an architecture and resource change needed to discuss. We might be able to sustain the functionality right now for a while.
@Cliftonz @scopsy we need to discuss this.

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+   * ignored in its normal operation.
+   *
+   * This function serves to prevent Node's default behavior of crashing
+   * when an "error" event is emitted in the absence of listeners.
+   */
+  distributedLock.on('error', (error) => {
+    // Ignore cases where a resource is explicitly marked as locked on a client.
+    if (error instanceof ResourceLockedError) {
+      // TODO: Hide or move to trace.
+      Logger.error(error, LOG_CONTEXT);
```
p-fernandez, line 53: This logs the times some call hits an used lock. We do not need to log them but right now while we see how well the implementation works, might be useful to understand the process. Move it to `debug` level could be a good idea.

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -162,4 +166,42 @@ export class JobRepository extends BaseRepository<EnforceEnvironmentQuery, JobEn
...
+            status: JobStatusEnum.DELAYED,
+          },
+        }
+      );
+
+      execution.modified = updatedDigestJob.modifiedCount;
+    }
+
+    return execution;
+  }
```
p-fernandez, line 206: We moved the previous domain logic to the DAL layer to be able to fully lock it at DAL layer with the distributed lock. Unfortunately I haven't been able to implement with Mongoose/MongoDB functionalities the logic: 
find if there is a existing delayed job for this subscriber-template-enviroment -> if not found make this certain job as the digest one delayed.
That was my goal. But weren't able to fix it with workflow changes or MongoDB functionalities (API, transactions and locks, etc)

That logic was implemented before in the domain. When concurrent calls hit it, there was a race condition as both calls when trying to find a existing delayed job didn't find anything, creating more than one single delayed job, therefore digest not working properly and duplicating the number of notifications sent.

apps/api/src/app/events/usecases/add-job/add-digest-job.usecase.ts
scopsy: Should we also add here the `digestKey` as part of the lock? Since different digest keys might not affect each other and other is no need to lock them.

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+): void => {
+  if (distributedLock) {
+    return;
+  }
+
+  // TODO: Implement distributed nodes (at least 3 Redis instances)
+  const rawInstances = [redisUrl].filter((instance) => !!instance).map((url) => new Redis(url));
+
+  distributedLock = new Redlock(rawInstances, settings);
+  Logger.log('Redlock started', LOG_CONTEXT);
```
scopsy, line 34: Love the usage of context

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    Logger.log(`Lock ${resource} for ${handler.name}`, LOG_CONTEXT);
+
+    const result = await handler();
+
+    return result;
+  } finally {
+    await releaseLock();
+    Logger.log(`Lock ${resource} released for ${handler.name}`, LOG_CONTEXT);
+  }
+};
```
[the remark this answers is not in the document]
scopsy: Love the implementtation with the callback here

apps/api/src/bootstrap.ts
scopsy: Could we maybe move this logic to shared.module.ts: 
```
{
    provide: DalService,
    useFactory: async () => {
      await dalService.connect(process.env.MONGO_URL);

      return dalService;
    },
  },
```

And follow a similar pattern of the Dal service so than nest will bootstrap it as part of it's lifecycle of modules instead of here?

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+export const startup = (
+  settings: Settings = {
+    automaticExtensionThreshold: 500,
+    driftFactor: 0.01,
+    retryCount: 50,
+    retryDelay: 100,
+    retryJitter: 200,
+  }
+): void => {
+  if (distributedLock) {
```
scopsy, line 26: Could we maybe also check that the appropriate REDIS_HOST env variable exists? And skip the lock functionality so the system can also operate potentially with the locking mechanism as usual?

apps/api/src/services/distributed-lock/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    Logger.log(`Lock ${resource} acquired for ${ttl} ms`, LOG_CONTEXT);
+
+    return createLockRelease(acquiredLock);
+  } catch (error) {
+    Logger.error(`Lock ${resource} threw an error: ${error.message}`, LOG_CONTEXT);
+    throw error;
+  }
+};
+
+export const applyLock = async <T>({ resource, ttl }: ILockOptions, handler: () => Promise<T>): Promise<T> => {
```
scopsy, line 122: Could we maybe introduce an e2e test for the lock mechanism? To validate it works as expected?

scopsy (commented):
Phenomenal work on this one! Adding a few notes here

apps/api/src/services/distributed-lock/index.ts
p-fernandez: That's a good point. Though it won't work properly, but at least would somehow work.

apps/api/package.json
```diff
@@ -104,6 +104,7 @@
     "passport-jwt": "^4.0.0",
     "passport-oauth2": "^1.6.1",
     "recursive-diff": "^1.0.8",
+    "redlock": "5.0.0-beta.2",
```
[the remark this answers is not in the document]
p-fernandez: I need to change the implementation then.

apps/api/src/app/events/usecases/add-job/add-digest-job.usecase.ts
p-fernandez: I can temporarily do it in case `digestKey` is provided. But I need to revisit the `digestKey` feature after this anyway. There have been feedback is not working properly.

scopsy: So let's leave it for later than

apps/api/package.json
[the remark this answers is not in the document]
scopsy: Might not be a problem, just wondering how stable the version is

apps/api/src/app/events/e2e/delay-events.e2e.ts
```diff
@@ -190,13 +190,18 @@ describe('Trigger event - Delay triggered events - /v1/events/trigger (POST)', f
...
       _templateId: template._id,
       type: StepTypeEnum.DELAY,
     });
 
-    const diff = differenceInMilliseconds(new Date(delayedJob.payload.sendAt), new Date(delayedJob?.updatedAt));
+    expect(delayedJobs.length).to.eql(1);
+
+    const delayedJob = delayedJobs[0];
+
+    const updatedAt = delayedJob?.updatedAt as string;
```
p-fernandez, line 203: TS warnings removal

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps-regular.usecase.ts
```diff
@@ -10,16 +10,13 @@ export class DigestFilterStepsRegular {
...
+
     for (const step of command.steps) {
-      if (step.template.type === StepTypeEnum.DIGEST) {
+      if (step?.template?.type === StepTypeEnum.DIGEST) {
         delayedDigests = await this.getDigest(command, step);
       }
 
-      if (delayedDigests) {
-        continue;
-      }
```
p-fernandez, line 21: This bit with the old `shouldAddDigestUseCase` were creating a double race condition that were mutually dependent.
When creating steps we were checking if there was a delayed digest and when deciding if adding the digest or not we were checking if existed or not.
We have yet to clarify the digest code flow a bit (in a forthcoming PR) but we decided to go through the direction of always creating all the steps for a digest job, no matter what, and at the time of deciding if adding the Digest job or not to the queue, either mark as merged/digested the steps for the digest jobs that are merged.

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -42,7 +42,7 @@ export class DigestFilterSteps {
         _creatorId: command.userId,
         type: StepTypeEnum.TRIGGER,
         content: '',
-      } as any,
+      } as MessageTemplateEntity,
```
p-fernandez, line 45: Just to remove the `any` and the warning. This object misses 2 properties to not raise the warn.

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps-regular.usecase.ts
scopsy: @davidsoderberg cc

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps-regular.usecase.ts
```diff
@@ -41,16 +38,17 @@ export class DigestFilterStepsRegular {
   }
 
   private async getDigest(command: DigestFilterStepsCommand, step: NotificationStepEntity) {
-    const where: any = {
```
scopsy, line 44: 👏

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts
```diff
@@ -0,0 +1,149 @@
...
+import * as Redlock from 'redlock';
+import { setTimeout } from 'timers/promises';
+import { Logger } from '@nestjs/common';
+
+const LOG_CONTEXT = 'DistributedLock';
+
+const redisUrl =
+  process.env.REDIS_HOST && process.env.REDIS_PORT
+    ? `${process.env.REDIS_HOST}:${process.env.REDIS_PORT}`
+    : 'localhost:6379';
```
scopsy, line 11: I actually think we can throw an error here instead of falling back to localhost, wdyt? We also have a validation in the env-validator for this value as a required value. wdyt?

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts
```diff
@@ -0,0 +1,149 @@
...
+
+interface ILockOptions {
+  resource: string;
+  ttl: number;
+}
+
+export class DistributedLockService {
+  public distributedLock: Redlock;
+  public instances: Redis[];
+  public lockCounter = {};
```
scopsy, line 21: Because the system is distributed and this will be on each instance of the service, what is the user of this lockCounter? Should we maybe remove it?

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  constructor() {
+    this.startup();
+  }
+
+  public startup(
+    settings = {
+      driftFactor: 0.01,
+      retryCount: 50,
+      retryDelay: 100,
+      retryJitter: 200,
```
scopsy, line 33: Can we make the TTL and retry delay and count higher? (maybe just retryCount)

scopsy (approved):
Amazing work! Added a few minor comments, other than that feel free to merge

apps/api/src/app/events/services/distributed-lock-service/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+@Injectable()
+export class EventsDistributedLockService {
+  private readonly distributedLockService: DistributedLockService;
+
+  constructor() {
+    this.distributedLockService = new DistributedLockService();
+  }
+
+  public async applyLock<T>(settings, handler): Promise<T> {
```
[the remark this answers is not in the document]
p-fernandez: For settings we had it in Redlock 5. When downgrading we lost it. I can add it for `handler`.

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts
p-fernandez: Let's go for it, to have extra safety in case environment validator somehow stops working.

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts
p-fernandez: The `lockCounter` is needed for the shutdown operation to avoid the system to shutdown while a lock is in operation (safety measure). What we need to do, and I think it got lost in my previous comments, is to know where I can engage the `shutdown` functionality in our Nest app, because I haven't found any way where we do operations before app Nest shutdown. Can you point me where we could do that?

PR #2726, opened 2023-02-11
fix: Digest key does not allow specifying a nested payload key #2697

### LINKED ISSUE #2697: [NV-1641] 🐛 Bug Report: Digest key does not allow specifying a nested payload key
### 📜 Description

When creating a digest, digest key doesn't work when you set the digestKey to `object.name` for example for a payload like `{ "object": {"name": "test" }}`.

Other parts of Novu do allow access to a nested payload key so this seems like it should be supported.

Also, if Novu does not find the digest key (like currently if you use a nested key), instead of falling back to simply using the subscriberId as a default, the digest will be turned off entirely.

### 👟 Reproduction steps

Create a digest and specify a nested payload key as the digest key.

### 👍 Expected behavior

It should accept a nested payload key as a digest key

It allows entering "object.name" for example but the digest turns off.

_No response_

None

<sub>[NV-1641](https://linear.app/novu/issue/NV-1641/🐛-bug-report-digest-key-does-not-allow-specifying-a-nested-payload)</sub>

tsssdev:

We have total 3 issues fixed in this PR:
1. Digest not working for multiple digest values for digest key
2. When digest key missing in payload should fallback to group by subscriber
3. Nested digest keys.

Modified and Added few more test cases to cover these cases, please review and let me know, all the weird issues caused by the bugs in the code, I  think lack of locking is not causing for these problems.

BTW, while working on this bug I felt the current digest code is too complex and excessive and can be simplified significantly, also performance also can be improved by refactoring and reducing the DB calls. I will create a bug and create draft PR and we can discuss in detail.

Closes #2697  and a few other open issues related to digest.


scopsy:
Tagging here @p-fernandez since he is already doing a lot of work on those files in his PR, so we can merge those two PR's together 🙏

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -812,14 +826,110 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
     expect(messages[1].content).to.include(digests[1].payload.postId);
+    expect(messages[2].content).to.include(digests[2].payload.postId);
+    const jobCount = await jobRepository.count({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+    });
+    const allJobs = await jobRepository.find({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+    });
```
p-fernandez, line 838: You don't need this call. 🙂

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -812,14 +826,110 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+        },
+      ],
+    });
+    await triggerEvent({
+      customVar: 'No digest key',
+    });
+
+    await triggerEvent({
+      customVar: 'digest key1',
+      nested: { postId: postId },
```
p-fernandez, line 870:
```suggestion
      nested: { postId },
```

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -812,14 +826,110 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+        RunJobCommand.create({
+          jobId: digest._id,
+          environmentId: digest._environmentId,
+          organizationId: digest._organizationId,
+          userId: digest._userId,
+        })
+      );
+    }
+
+    await session.awaitRunningJobs(template?._id, false, 0);
```
p-fernandez, line 915: You should be able to use the same solution as in the previous test you modified and just use the `awaitRunningJobs` testing utility.

```suggestion
    await session.awaitRunningJobs(template?._id, false, 3);
```
3 as you are expecting to have 3 digest jobs unfinished.

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -812,14 +826,110 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+    expect(messages[2].content).to.include(digests[2].payload.nested.postId);
     const jobCount = await jobRepository.count({
       _environmentId: session.environment._id,
       _templateId: template._id,
     });
-    expect(jobCount).to.equal(6);
+    const allJobs = await jobRepository.find({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+    });
```
p-fernandez, line 931: Again, not needed. 🙂

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -812,14 +826,110 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+
+    let digests = await jobRepository.find({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+      type: StepTypeEnum.DIGEST,
+    });
+    expect(digests[1].payload.nested.postId).not.to.equal(digests[2].payload.nested.postId);
+    expect(digests[0].payload.nested?.postId).to.equal(undefined);
+
+    expect(digests.length).to.equal(3);
```
p-fernandez, line 897: Makes more sense this check to be done before the ones of the nested postId as if this one is not accomplished the others won't and checking it later is not totally useful (I always expect tests to fail as early as possible).
Can you change the order in this test and the other ones where the lenght of the digests are checked after doing other array checks?

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -927,8 +1046,133 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
       _environmentId: session.environment._id,
       _templateId: template._id,
     });
+    const allJobsBackoff = await jobRepository.find({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+    });
```
p-fernandez, line 1052: Not needed, again. 🙂

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -927,8 +1046,133 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+    expect(jobCount).to.equal(15);
+  });
+
+  it('should create multiple digests based on different nested digestKeys with backoff', async function () {
+    const postId = MessageRepository.createObjectId();
+    const postId2 = MessageRepository.createObjectId();
```
p-fernandez, line 1058: Thank you very much for improving the testing suites with more solid and thorough tests. 👏🏻

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -927,8 +1046,133 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+    for (const digest of digests) {
+      await runJob.execute(
+        RunJobCommand.create({
+          jobId: digest._id,
+          environmentId: digest._environmentId,
+          organizationId: digest._organizationId,
+          userId: digest._userId,
+        })
+      );
+    }
```
p-fernandez, line 1141: I am still unsure that you would need to execute this to accomplish the goal of the test. What are the reasons behind forcing the job executions for digest type of jobs?

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -927,8 +1046,133 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
...
+    expect(contents).to.include(`Hello world ${postId2}`);
+
+    const jobCount = await jobRepository.count({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+    });
+    const allJobsBackoff = await jobRepository.find({
+      _environmentId: session.environment._id,
+      _templateId: template._id,
+    });
```
p-fernandez, line 1174: 🗑️

apps/api/src/app/events/usecases/add-job/should-add-digest-job.usecase.ts
```diff
@@ -14,14 +15,19 @@ export class ShouldAddDigestJob {
...
+    const where: any = {
       status: JobStatusEnum.DELAYED,
       type: StepTypeEnum.DIGEST,
       _subscriberId: data._subscriberId,
       _templateId: data._templateId,
       _environmentId: data._environmentId,
-    });
+    };
+    if (digestKey) {
+      //where['payload.' + digestKey] = data.payload[digestKey];
```
p-fernandez, line 27: 🗑️

apps/api/src/app/events/usecases/add-job/should-add-digest-job.usecase.ts
```diff
@@ -14,14 +15,19 @@ export class ShouldAddDigestJob {
...
+      where['payload.' + digestKey] = DigestFilterSteps.getNestedValue(data.payload, digestKey);
+    }
+    const delayedDigest = await this.jobRepository.findOne(where);
```
p-fernandez, line 30: In the PR where I am fixing the digest job problems reported we are moving this to the DAL. I will take care of it as this use case will disappear.

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -46,4 +46,19 @@ export class DigestFilterSteps {
       _templateId: command.templateId,
     };
   }
+  public static getNestedValue<ObjectType>(payload: ObjectType, path: string) {
```
p-fernandez, line 49: Please add a blank line to separate the methods.

Would you mind to add a unit test suite located in this same folder as `digest-filter-steps.usecase.spec.ts` where we can test this function?

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -46,4 +46,19 @@ export class DigestFilterSteps {
...
+      const keys = path.split('.');
+      for (const key of keys) {
+        if (!result) return undefined;
+        result = result[key];
+      }
+    } catch (error) {
+      return undefined;
+    }
+
+    return result;
```
p-fernandez, line 62: I think we could simplify the implementation, but that's why I asked you to add a unit test suite so we can have full confidence in this suggestion because I am not 100% sure:
```suggestion
    if (!path || !payload) {
      return undefined;
    }
    
    try {
      let result;
      const keys = path.split('.');
      for (const key of keys) {
        result = payload[key];
      }
      
      return result;
    } catch (error) {
      return undefined;
    }
```
Let me know what you think or if I missed any nuance of the fix.

apps/api/src/app/events/usecases/send-message/digest/get-digest-events.usecase.ts
```diff
@@ -6,19 +6,19 @@ import {
...
-    if (batchValue) {
-      jobs = jobs.filter((job) => {
-        return job.payload[currentJob.digest.digestKey] === batchValue;
-      });
-    }
-
+    const batchValue = currentJob?.payload
+      ? DigestFilterSteps.getNestedValue(currentJob.payload, currentJob?.digest?.digestKey)
+      : undefined;
+    jobs = jobs.filter((job) => {
```
p-fernandez, line 19: You removed by accident the check of the condition if `batchValue` exists that avoids to run the loop as it would be useless.

p-fernandez (commented):
Thank you very much for helping us to fix this. 🙌🏻 

I made some suggestions and comments to the code.
I do agree the implementation of the digest jobs can be simplified and improved. I am currently working on that. Might need to include this PR on that work. Will keep you updated.

apps/api/src/app/events/e2e/digest-events.e2e.ts
tsssdev: Sorry, these statements were added for debugging tests and forgot to remove :)

apps/api/src/app/events/usecases/send-message/digest/get-digest-events.usecase.ts
tsssdev: No @p-fernandez This is removed intentionally, batchvalue missing jobs should have separate digest

apps/api/src/app/events/e2e/digest-events.e2e.ts
tsssdev: @p-fernandez all the places these digest/delayed jobs we are triggering manually, so we can test the cases what happens after the digest test cases executes(after delay), we cant wait jobs to finish, as we have 10 seconds time out and test cases failing with timeout, so triggering all 3 jobs manually and waiting for all jobs to finish, so zero. make sense?

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
tsssdev: Actually this suggested code has small bug, this line result = payload[key]; should be result = result[key];

p-fernandez:
I rebased this PR on top of the changes merged in `next` and did some extra fixes to your good work @tsssdev. I need someone else from the team to give it a look and we can merge it.

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -46,4 +46,25 @@ export class DigestFilterSteps {
...
+      for (const key of keys) {
+        result = payload[key];
+      }
+
+      return result;
+    } catch (error) {
+      Logger.error('Failure when parsing digest payload nested key');
+
+      return undefined;
+    }
```
scopsy, line 68: I think we can use `lodash.get` using a similar approach introduced in: https://github.com/novuhq/novu/pull/2794 it will also support arrays and etc...

scopsy (approved):
Looks great, added a small comment for improvement. 🎉

PR #2749, opened 2023-02-13
feat: add what filters was used for step to analytic event

davidsoderberg:
Analytics for step processing

Add what filter was used and what passed or failed

### DIFF
--- a/apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
+++ b/apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
@@ -64,7 +64,7 @@ export class SendMessage {
     const stepType = command.step?.template?.type;
 
     if (!command.payload?.$on_boarding_trigger) {
-      const usedFilters = shouldRun.conditions.reduce(this.sumFilters, {
+      const usedFilters = shouldRun.conditions.reduce(MessageMatcher.sumFilters, {
         stepFilters: [],
         failedFilters: [],
         passedFilters: [],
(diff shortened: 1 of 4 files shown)


apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
```diff
@@ -215,4 +222,33 @@ export class SendMessage {
...
+
+  private sumFilters(
+    summary: {
+      stepFilters: string[];
+      failedFilters: string[];
+      passedFilters: string[];
+    },
+    condition: ICondition
+  ) {
+    let type: string = condition.filter.toLowerCase();
```
scopsy, line 234: Is there any case that the filter will be undefined? Maybe better to add a 
```suggestion
    let type: string = condition.filter?.toLowerCase();
```

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
```diff
@@ -215,4 +222,33 @@ export class SendMessage {
 
     return !channels.find((channel) => channel === job.type);
   }
+
+  private sumFilters(
```
scopsy, line 226: Would be great to have a unit test on this one, while doing so it might be required to extract this to a public function on a filters service perhaps? and than can be unitested for all variations

scopsy (commented):
Great Work! Could you also check how this is affecting the Activity feed UI? Pawel and Biswa added there a widget to show the filters and their results, just to make sure it hasn't been broken in the process

davidsoderberg:
> Great Work! Could you also check how this is affecting the Activity feed UI? Pawel and Biswa added there a widget to show the filters and their results, just to make sure it hasn't been broken in the process

it does not, feed items are not depending on my code, I use the same object as the feed items does but nothing changes them, just reading them 😄

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
davidsoderberg: According to the type there should always be a string...

scopsy: Yes seen it, but is it possible that there will be some data inside the DB that will be set to null or undefined?

scopsy (approved):

PR #2750, opened 2023-02-13
docs(topic): Replace enum with string for simplified understanding

unicodeveloper:

Updates the Topic docs to do away with enums and simply use strings. 

This change is needed to simplify the understanding of developers trying to trigger notifications to topics. Before now, they need to go into the codebase to check the value of the enum. With this change, they can just copy and paste and trigger.

Furthermore, it makes it easy for non-JavaScript/non-Node developers to flow with the topics docs.


p-fernandez (approved):
🌟

PR #2489, opened 2023-01-12
feat(wip): added cached entity

djabarovgeorge:
### What change does this PR introduce?

It is the POC of dividing the Cached Interceptor into CachedRequest and CachedEntity, this one is the CachedEntity part, I a bit carried away and added a couple of extra queries which were not so necessary at the moment please look at the trigger flow part.

In this solution, we cache under subscriber._id && subscriber.subscriberId separately, and when we need to invalidate we invalidate by key and not by pattern. 

### Why was this change needed?

In order to reduce the number of scans on the Redis server.

PR #2754, opened 2023-02-13
fix: proper error handling for missing subscriber on preference update

scopsy:

- Refactor usage of subscriberId and internal subscriber id on fetch preference
- Throw exception when subscriber not found
- Added missing change

NV-1459

### DIFF
--- a/apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
+++ b/apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
@@ -37,6 +37,7 @@ import { ANALYTICS_SERVICE } from '../../../shared/shared.module';
 import { Cached } from '../../../shared/interceptors';
 import { CacheKeyPrefixEnum } from '../../../shared/services/cache';
 import { MessageMatcher } from '../trigger-event/message-matcher.service';
+import { ApiException } from '../../../shared/exceptions/api.exception';
 
 @Injectable()
 export class SendMessage {
@@ -174,11 +175,15 @@ export class SendMessage {
       environmentId: job._environmentId,
     });
 
+    const subscriber = await this.subscriberRepository.findById(job._subscriberId);
+    if (!subscriber) throw new ApiException('Subscriber not found with id ' + job._subscriberId);
+
     const buildCommand = GetSubscriberTemplatePreferenceCommand.create({
       organizationId: job._organizationId,
-      subscriberId: job._subscriberId,
+      subscriberId: subscriber.subscriberId,
       environmentId: job._environmentId,
       template,
+      subscriber,
     });
 
     const { preference } = await this.getSubscriberTemplatePreferenceUsecase.execute(buildCommand);
--- a/apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
+++ b/apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
@@ -30,7 +30,7 @@ export class GetSubscriberTemplatePreference {
       command.subscriber ??
       (await this.subscriberRepository.findBySubscriberId(command.environmentId, command.subscriberId));
     if (!subscriber) {
-      throw new ApiException('Subscriber not found');
+      throw new ApiException(`Subscriber ${command.subscriberId} not found`);
     }
 
     const subscriberPreference = await this.subscriberPreferenceRepository.findOne({
--- a/apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
+++ b/apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
@@ -13,6 +13,7 @@ import {
   ISubscriberPreferenceResponse,
 } from '../get-subscriber-preference/get-subscriber-preference.usecase';
 import { GetSubscriberTemplatePreferenceCommand } from './get-subscriber-template-preference.command';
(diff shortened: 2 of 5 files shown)


apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
```diff
@@ -174,11 +175,15 @@ export class SendMessage {
       environmentId: job._environmentId,
     });
 
+    const subscriber = await this.subscriberRepository.findById(job._subscriberId);
+    if (!subscriber) throw new ApiException('Subscriber not found with id ' + job._subscriberId);
+
     const buildCommand = GetSubscriberTemplatePreferenceCommand.create({
       organizationId: job._organizationId,
-      subscriberId: job._subscriberId,
+      subscriberId: subscriber.subscriberId,
```
scopsy, line 183: to make sure we only pass here the public subscriberId to avoid confusion

apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -25,11 +26,16 @@ export class GetSubscriberTemplatePreference {
 
   async execute(command: GetSubscriberTemplatePreferenceCommand): Promise<ISubscriberPreferenceResponse> {
     const activeChannels = await this.queryActiveChannels(command);
-    const subscriber = await this.subscriberRepository.findBySubscriberId(command.environmentId, command.subscriberId);
+    const subscriber =
```
scopsy, line 29: To avoid a query when not needed

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
djabarovgeorge: that one will help, I refactored the job entity in https://github.com/novuhq/novu/pull/2489
maybe I will look into finalizing it on the cool down in order to merge it as well.

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
```diff
@@ -174,11 +175,15 @@ export class SendMessage {
       environmentId: job._environmentId,
     });
 
+    const subscriber = await this.subscriberRepository.findById(job._subscriberId);
```
djabarovgeorge, line 178: In this PR #2489 we added user's subscriberId to the job entity. I will add a not note on this PR to get the subscriber from the cache first.

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
djabarovgeorge: This flow is too confusing! that's why I refactored it on the #2489 😀

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
scopsy: Awesome so we can fix it there later 🎉

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
```diff
@@ -174,11 +175,15 @@ export class SendMessage {
...
+    if (!subscriber) throw new ApiException('Subscriber not found with id ' + job._subscriberId);
```
djabarovgeorge, line 179: Btw I am not sure we need to add the exception here because if the flow got till here then we are after ProcessSubscriber usecase, in that case, the execute function already fetched/created a subscriber and passed it. 

the possible exception we can get is when the user tries to update an existing subscriber through  
 /subscribers/:subscriberid/preferences
 on the GetPreferences usecase flow. 
 wdyt?

djabarovgeorge: never mind i can see that you added a check on the [get-subscriber-template-preference.usecase.ts](https://github.com/novuhq/novu/pull/2754/files#diff-cca4106add4716bff4349e085b13e832c6baa61c597d32169309e016e7794cb0). :)

apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -25,11 +26,16 @@ export class GetSubscriberTemplatePreference {
...
+      command.subscriber ??
+      (await this.subscriberRepository.findBySubscriberId(command.environmentId, command.subscriberId));
+    if (!subscriber) {
+      throw new ApiException('Subscriber not found');
```
djabarovgeorge, line 33:
```suggestion
      throw new ApiException(`Subscriber ${command.subscriberId} not found`);
```

apps/api/src/app/subscribers/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -25,11 +26,16 @@ export class GetSubscriberTemplatePreference {
...
+    }
 
     const subscriberPreference = await this.subscriberPreferenceRepository.findOne({
       _environmentId: command.environmentId,
-      _subscriberId: subscriber !== null ? subscriber._id : command.subscriberId,
+      _subscriberId: subscriber._id,
```
djabarovgeorge, line 38: 🤩

djabarovgeorge (approved):
Looks great to me :)

apps/api/src/app/events/usecases/send-message/send-message.usecase.ts
djabarovgeorge: Yeah i left a note there so we wont forget 😅

PR #2755, opened 2023-02-13
Add Logging to Api, Mail, Webhook, and Websocket

Cliftonz:

This PR introduces standardized logging for the application. The creation of the loggers can be found in packages -> application-generic where a logger can be instantiated with either createLogger for non-nest apps and create NestLogger for nest based apps.

This change will allow us to start creating quality logs for better remediation of issues on premise and in novu cloud.

### Notes
This PR also introduces the env variable LOGGING_LEVEL to all backends and we will need to ensure that new relic licenses, app name, and logging level to all deployments.

This deployment is also blocked by the eu pr as it contains updates to the workflow.

We will also want to look at being able to dynamically update the logging level while the system is running so we can collect debug or verbose logs while the system is running without redeploying it.


apps/api/src/app/auth/auth.controller.ts
```diff
@@ -59,13 +61,19 @@ export class AuthController {
...
+  githubAuth(context: ExecutionContext) {
+    Logger.verbose('Checking Github Auth');
+    Logger.debug('Context: ' + context);
     if (!process.env.GITHUB_OAUTH_CLIENT_ID || !process.env.GITHUB_OAUTH_CLIENT_SECRET) {
-      throw new ApiException(
+      const err = new ApiException(
         'GitHub auth is not configured, please provide GITHUB_OAUTH_CLIENT_ID and GITHUB_OAUTH_CLIENT_SECRET as env variables'
       );
+      Logger.error(err);
+      throw err;
```
djabarovgeorge, line 72: @Cliftonz @scopsy 
Do we want to log all the exceptions that are thrown? if we do maybe it will be a good idea to place this log on nest ResponseInterceptor or ExceptionInterceptor. wdyt?

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -59,13 +61,19 @@ export class AuthController {
   ) {}
 
   @Get('/github')
-  githubAuth() {
+  githubAuth(context: ExecutionContext) {
+    Logger.verbose('Checking Github Auth');
```
djabarovgeorge, line 65: Do we have any metadata on the log that will be added to those logs?  for example transactionId/ environmentId/ userId etc...

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
```diff
@@ -22,25 +22,39 @@ export class AddJob {
...
+    Logger.verbose('Getting Job');
     const job = command.job ?? (await this.jobRepository.findById(command.jobId));
+    Logger.debug('job is: ' + job);
     if (!job) {
+      Logger.warn('job was null in both the input and search');
+
       return;
     }
 
+    Logger.log('Starting New Job of type: ' + job.type);
```
djabarovgeorge, line 36: Is there a logger level 'log'?

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
```diff
@@ -69,6 +89,10 @@ export class AddJob {
   }
 
   public static toMilliseconds(amount: number, unit: DigestUnitEnum): number {
+    Logger.debug('Amount is: ' + amount);
+    Logger.debug('Unit is: ' + unit);
```
djabarovgeorge, line 93: What do you think about combining those log to one execution?

djabarovgeorge (commented):
Looks like awesome work, left a couple of questions in order to understand better.

apps/api/package.json
```diff
@@ -96,9 +96,11 @@
     "ioredis": "^5.2.4",
     "jsonwebtoken": "9.0.0",
     "lodash": "^4.17.15",
+    "logform": "^2.5.1",
     "nanoid": "^3.1.20",
     "nest-raven": "8.2.0",
     "newrelic": "9.10.0",
+    "nest-winston": "^1.8.0",
```
scopsy, line 103: A small ideation here, did you had a chance to check out https://github.com/pinojs/pino ? Seems like the also have the same easy integration with NewRelic but have a much lower performance impact on the application running it and comes with a built in interceptor for HTTP requests aswell. Wdyt?

apps/api/src/app/auth/auth.controller.ts
scopsy: Good point! I think that we should add the metadata for : environmentId, userId, organizationId whenever a jwt token is available in the request, @Cliftonz how can we do that?

apps/api/src/app/auth/auth.controller.ts
scopsy: I think logging exceptions in the middleweare/interceptor is a better approach. We can still log error when we are catching them and not rethrowing.

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -59,13 +61,19 @@ export class AuthController {
...
     if (!process.env.GITHUB_OAUTH_CLIENT_ID || !process.env.GITHUB_OAUTH_CLIENT_SECRET) {
-      throw new ApiException(
+      const err = new ApiException(
         'GitHub auth is not configured, please provide GITHUB_OAUTH_CLIENT_ID and GITHUB_OAUTH_CLIENT_SECRET as env variables'
       );
+      Logger.error(err);
+      throw err;
     }
 
+    Logger.log('Github Auth has all variables.');
```
scopsy, line 75: What is the difference between info and log?

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
```diff
@@ -22,25 +22,39 @@ export class AddJob {
   ) {}
 
   public async execute(command: AddJobCommand): Promise<void> {
+    Logger.debug('command is: ' + command);
```
scopsy, line 25: I'm a bit concerned with PII here, and also, in Javascript, this will show: `command is: [Object object]` due to the way how strings works.

How are we dealing with PII leaking to those logs?

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
```diff
@@ -22,25 +22,39 @@ export class AddJob {
...
+
+    Logger.verbose('Getting Job');
```
scopsy, line 27: Without the proper context such as `environmentId`, orgIf, transactionId those logs seem to me quite useless since we will have millions of them without any context of a request, or am i missing something?

apps/api/src/newrelic.ts
```diff
@@ -29,13 +31,18 @@ exports.config = {
...
+    },
+  },
   logging: {
     /**
      * Level at which to log. 'trace' is most useful to New Relic when diagnosing
      * issues with the agent, 'info' and higher will impose the least overhead on
      * production applications.
      */
-    level: 'info',
+    level: 'trace',
```
scopsy, line 45: They are mentioning that trace should be used when diagnosing things, and that higher trace levels should be used. What are the performance implications? Should we do here log and above here?

apps/api/src/newrelic.ts
```diff
@@ -4,6 +4,8 @@
  * See lib/config/default.js in the agent distribution for a more complete
  * description of configuration variables and their potential values.
  */
+console.log('Newrelic.js file');
```
scopsy, line 7: I think this was left by a mistake here :)

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
Cliftonz: If we do that then new relic will parse the values together instead of apart.

apps/api/src/app/auth/auth.controller.ts
Cliftonz: they are the same thing its just how the nest logger interface is.

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
Cliftonz: Well to be fair, we do need to be carefull about that. 
The best approach I have seen is create dto variable annotations that will mask the type for us.

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
Cliftonz: It is the nestjs version of info

apps/api/src/app/auth/services/passport/jwt.strategy.ts
```diff
@@ -19,6 +20,12 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
       throw new UnauthorizedException();
     }
 
+    this.logger.assign({
+      userId: user._id,
+      environmentId: payload.environmentId,
+      organizationId: payload.organizationId,
+    });
```
djabarovgeorge, line 27: Does that mean that this data will be propagated to all of the logs related to this request? do you think that we should add data related to running the machine as well?

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -22,10 +23,19 @@ export class TriggerEvent {
     private jobRepository: JobRepository,
     private addJobUsecase: AddJob,
     private notificationRepository: NotificationRepository,
-    protected createExecutionDetails: CreateExecutionDetails
+    protected createExecutionDetails: CreateExecutionDetails,
+    private logger: PinoLogger
   ) {}
 
   async execute(command: TriggerEventCommand) {
+    const child = this.logger.logger.child({
```
djabarovgeorge, line 31: What does the creation of the child mean here? 
In addition, what do you think about creating usecase base class in order to assign more info like where the log printed (the usecase name for example)

apps/api/src/app/auth/services/passport/jwt.strategy.ts
scopsy: Data for machine is auto added by newrelic, and yes this will be added to all logs of the request

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
scopsy: Was just a test we've done, was removed :)

apps/inbound-mail/src/.env.development
```diff
@@ -7,3 +7,8 @@ HOST="0.0.0.0"
 REDIS_DB_INDEX="2"
 REDIS_HOST="localhost"
 REDIS_PORT="6379"
+
+NEW_RELIC_ENABLED=true
+NEW_RELIC_APP_NAME="[DEV] - INBOUND-MAIL"
```
scopsy, line 12: Can we make those variable to be passed from terraform? so we can have consistent naming for Prod environments (US/EU and etc...)

Cliftonz: Absolutely

Cliftonz: Done

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
Cliftonz: @scopsy I think this should be addressed in an update soon that provides masking annotations to the dtos.

Cliftonz:
@scopsy @djabarovgeorge Here is an example of masking outputs with the maskValue function.
This function supports recursively masking a value or an object.
Screenshot 2023-02-28 at 10 31 03 AM

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
Cliftonz: This is now covered by the Log decorator.

apps/api/package.json
```diff
@@ -105,11 +105,14 @@
     "passport-github2": "^0.1.12",
     "passport-jwt": "^4.0.0",
     "passport-oauth2": "^1.6.1",
+    "pino": "^8.10.0",
+    "pino-http": "^8.3.3",
+    "pino-pretty": "^9.2.0",
```
scopsy, line 110: If we have moved everything to application generic, can we remove it from the API service?

packages/application-generic/src/logging/log.decorator.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      return await originalMethod.apply(this, args);
+    };
+  };
+}
+
+// FIXME
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+export const logDecorator = () => DecorateAll(log(), { deep: true });
```
scopsy, line 38:
```suggestion
export const LogDecorator = () => DecorateAll(log(), { deep: true });
```

for style alignments with other providers would be great to have it Capitalized

packages/application-generic/src/logging/log.decorator.ts
```diff
@@ -0,0 +1,38 @@
...
+export function log() {
+  return (
+    target: any,
+    propertyKey: string,
+    propertyDescriptor: PropertyDescriptor
+  ) => {
+    const originalMethod = propertyDescriptor.value;
+
+    propertyDescriptor.value = async function (...args: any[]) {
+      // { methodName: "name", args: [{"name": "value"}] }
```
scopsy, line 14: Can this be removed?

packages/application-generic/src/logging/log.decorator.ts
```diff
@@ -0,0 +1,38 @@
...
+        args: [],
+      };
+
+      const promises = args.map(async (value) => {
+        item.args.push({
+          [value.constructor.name]: maskValue(value.constructor.name, value),
+        });
+      });
+
+      await Promise.all(promises);
```
scopsy, line 26: Not entirely sure why all this complexity is needed, I think this could be easily refactored to:

```suggestion
for (const argument of args) {
    item.args.push({
      [value.constructor.name]: maskValue(value.constructor.name, value),
    });
}
```

Or perhaps I'm missing something

packages/application-generic/src/logging/masking.ts
```diff
@@ -0,0 +1,157 @@
+import MaskData from 'maskdata';
+
+const cardFields = ['credit', 'debit', 'creditCard', 'debitCard'];
```
scopsy, line 3: Can we also mask: avatar and photo fields here?

packages/application-generic/src/middleware/logger.middleware.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable, NestMiddleware, Logger } from '@nestjs/common';
+
+@Injectable()
+export class LoggerMiddleware implements NestMiddleware {
+  use(req: any, res: any, next: () => void) {
+    Logger.verbose(req);
+    Logger.verbose(res);
+    next();
+  }
```
scopsy, line 9: Isn't Pino-nest already have this functionality built in?

packages/application-generic/src/services/analytics.service.ts
```diff
@@ -84,6 +85,7 @@ export class AnalyticsService {
   }
 
   track(name: string, userId: string, data: Record<string, unknown> = {}) {
+    Logger.log('User: ' + userId + ' Did: ' + name, data);
```
scopsy, line 88: What does DiD stands for? maybe just call it: event ?

packages/application-generic/src/logging/index.ts
```diff
@@ -0,0 +1,115 @@
...
+  };
+}
+
+export function createNestLoggingModuleOptions(settings: ILoggerSettings) {
+  const values = getLoggingVariables();
+
+  return {
+    pinoHttp: {
+      customLevels: loggingLevelSet,
+      // useOnlyCustomLevels: true,
```
scopsy, line 91: If not relevant we can remove this comment here

packages/application-generic/src/logging/index.ts
```diff
@@ -0,0 +1,115 @@
+import {
+  LoggerErrorInterceptor,
+  Logger,
+  LoggerModule,
+  PinoLogger,
+} from 'nestjs-pino';
+import { storage, Store } from 'nestjs-pino/storage';
+import * as process from 'process';
```
scopsy, line 8: This import line should be removed, process is available to the service and no need to import it.

apps/ws/src/bootstrap.ts
```diff
@@ -18,9 +19,14 @@ if (process.env.SENTRY_DSN) {
...
 
 export async function bootstrap() {
-  const app = await NestFactory.create(AppModule);
+  const app = await NestFactory.create(AppModule, { bufferLogs: true });
   const redisIoAdapter = new RedisIoAdapter(app);
 
+  app.useLogger(app.get(Logger));
+  app.flushLogs();
+
+  app.useGlobalInterceptors(getErrorInterceptor());
```
scopsy, line 28: Is this realy needed when using pino nestjs?

apps/inbound-mail/src/.env.development
scopsy: I guess we can remove this from here if that's the case :)

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -19,10 +19,15 @@ export class UpdateIntegration {
   ) {}
 
   async execute(command: UpdateIntegrationCommand): Promise<IntegrationEntity> {
+    Logger.verbose('Excuting Update Integration Command');
```
scopsy, line 22:
```suggestion
    Logger.verbose('Eexcuting Update Integration Command');
```

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -39,6 +42,14 @@ export class TriggerEvent {
       },
     });
 
+    this.logger.assign({
+      transactionId: command.transactionId,
+      environmentId: command.environmentId,
+      organizationId: command.organizationId,
+    });
+
+    await this.validateTransactionIdProperty(command.transactionId, command.organizationId, command.environmentId);
```
scopsy, line 51: Why is this line here was added? It doesn't look like it exists on next branch

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -29,8 +29,16 @@ export class ParseEventRequest {
 
   async execute(command: ParseEventRequestCommand) {
     const transactionId = command.transactionId || uuidv4();
+    Logger.log('Starting Trigger');
 
     const mappedActor = command.actor ? this.mapTriggerRecipients.mapSubscriber(command.actor) : undefined;
+
+    if (mappedActor === undefined) {
+      Logger.warn('mappedActor is undefined');
```
scopsy, line 37: It's not a warning in this case, not having actor is a totally normal state of the system.

scopsy (changes requested):
Amazing effort! Adding a few comments 🙏 My biggest concern is about the masking functionality, why are we not using https://github.com/pinojs/pino/blob/master/docs/redaction.md and implementing this from scratch?

apps/ws/src/bootstrap.ts
Cliftonz: Yes, because we are using an async logger we need to make sure it is up and running so we dont lose any logs when the system starts up.

apps/api/src/app/widgets/widgets.controller.ts
```diff
@@ -54,6 +54,7 @@ import { RemoveMessage } from './usecases/remove-message/remove-message.usecase'
 
 @Controller('/widgets')
 @ApiExcludeController()
+@logDecorator()
```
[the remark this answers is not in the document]
Cliftonz: Yes, this is the custom method parameter logger.

apps/ws/src/bootstrap.ts
scopsy: Was not referring to flushLogs, mainly to the `useGlobalInterceptor`, from what I recall nest-pino adds one by default when autoLogging is enabled. (sorry for the confusion)

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -42,14 +42,6 @@ export class TriggerEvent {
       },
     });
 
-    this.logger.assign({
-      transactionId: command.transactionId,
-      environmentId: command.environmentId,
-      organizationId: command.organizationId,
-    });
-
-    await this.validateTransactionIdProperty(command.transactionId, command.organizationId, command.environmentId);
```
scopsy: Was it removed by accident?

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -42,14 +42,6 @@ export class TriggerEvent {
       },
     });
 
-    this.logger.assign({
-      transactionId: command.transactionId,
-      environmentId: command.environmentId,
-      organizationId: command.organizationId,
-    });
```
scopsy: Don't we still want to have this assigned here to be available on all the scope logs?

packages/application-generic/src/logging/index.ts
```diff
@@ -88,7 +83,6 @@ export function createNestLoggingModuleOptions(settings: ILoggerSettings) {
   return {
     pinoHttp: {
```
scopsy, line 84: @Cliftonz probably lost in my previous comments, but why are we implementing masking from scratch and not using the built in redact for pino? https://github.com/pinojs/pino/blob/master/docs/redaction.md

Cliftonz: Because the redact functionality requires that you know the exact path to react in every json object and it just blanket redacts without any knowledge of what we are masking. For example, email should be masked differently then a password.

apps/ws/src/bootstrap.ts
Cliftonz: Autologging is only for req/res not capturing errors, that is where the ErrorInterceptor comes in, thus this is needed to capture errors.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
Cliftonz: Yes added it back in.

packages/application-generic/src/logging/index.ts
Cliftonz: I attempted to implement this but the wildcard pathing was not working. @scopsy If you can take a look at this I think that would be best.
In this pr, I have kept my original implementation and made sure that the request logging processing is only done if log level is set to debug and uses promises to ensure non-synchronous blocking.

.idea/runConfigurations/API_E2E.xml
scopsy, line 1: Was this removed by mistake? @Cliftonz

apps/api/src/app/events/usecases/add-job/add-job.usecase.ts
```diff
@@ -57,10 +71,16 @@ export class AddJob {
...
+      Logger.warn('variable delay is null which is not apart of the definition');
+    }
+
+    Logger.verbose('Adding Job to Queue');
     await this.workflowQueueService.addToQueue(job._id, job, delay);
 
     if (delay) {
-      this.createExecutionDetails.execute(
+      Logger.verbose('Delay is active, Creating execution details');
+      await this.createExecutionDetails.execute(
```
scopsy, line 83: @Cliftonz those are not awaited by design here to avoid blocking the response to the user

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
scopsy: @Cliftonz saw that you resolved this issue, but seems like this line is still here, could you elaborate?

apps/api/src/app/events/events.controller.ts
```diff
@@ -22,12 +22,14 @@ import { TriggerEventToAll, TriggerEventToAllCommand } from './usecases/trigger-
...
 import { ExternalApiAccessible } from '../auth/framework/external-api.decorator';
 import { JwtAuthGuard } from '../auth/framework/auth.guard';
+import { LogDecorator } from '@novu/application-generic';
 
 @Controller({
   path: 'events',
   scope: Scope.REQUEST,
 })
 @ApiTags('Events')
+@LogDecorator()
```
scopsy, line 32: @Cliftonz not sure if you have answered this previously but couldn't find it, if nest-pino already have an auto logger for HTTP requests, why do we need the LogDecorator?

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
```diff
@@ -90,14 +91,21 @@ export class WorkflowQueueService {
...
-      return await this.runJob.execute(
-        RunJobCommand.create({
-          jobId: data._id,
-          environmentId: data._environmentId,
-          organizationId: data._organizationId,
-          userId: data._userId,
-        })
-      );
+      return await new Promise(async (resolve, reject) => {
+        storage.run(new Store(PinoLogger.root), () => {
```
scopsy, line 95: @Cliftonz this what was missing and caused the unit tests to fail ^

apps/api/src/app/shared/shared.module.ts
```diff
@@ -138,7 +139,14 @@ const PROVIDERS = [
 ];
 
 @Module({
-  imports: [],
+  imports: [
+    LoggerModule.forRoot(
+      createNestLoggingModuleOptions({
+        serviceName: 'novu/api',
+        version: '0.12.0',
```
scopsy, line 146: @Cliftonz why are we initializing it here twice? Once in the app module and the shared module? Should be only once, and if it's here let's make sure not to hard code those values

packages/application-generic/src/logging/LogDecorator.ts
scopsy, line 1: Seems like the log decorator is breaking a lot of existing modules, as it converts all non promise methods to be promise methods. For example the `toArray` method provided by in `widgets.controller.ts` file. 

Mine suggestion is, Don't do a global decorator to decorate ALL methods in a class, but have a more targeted decorator that will decorate individual methods to unwanted side effects.

PR #2757, opened 2023-02-14
fix: so empty string can not be used as default value

davidsoderberg:
so empty string can not be used as default value

To be more logical with how we handle default values

### DIFF
--- a/apps/api/src/app/message-template/usecases/create-message-template/create-message-template.spec.ts
+++ b/apps/api/src/app/message-template/usecases/create-message-template/create-message-template.spec.ts
@@ -1,7 +1,7 @@
 import { Test } from '@nestjs/testing';
 import { UserSession } from '@novu/testing';
 import { MessageTemplateRepository } from '@novu/dal';
-import { EmailBlockTypeEnum, StepTypeEnum } from '@novu/shared';
+import { EmailBlockTypeEnum, StepTypeEnum, TemplateVariableTypeEnum } from '@novu/shared';
 import { expect } from 'chai';
 
 import { CreateMessageTemplate } from './create-message-template.usecase';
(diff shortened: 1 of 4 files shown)


apps/api/src/app/message-template/usecases/create-message-template/create-message-template.usecase.ts
```diff
@@ -20,7 +20,13 @@ export class CreateMessageTemplate {
     let item = await this.messageTemplateRepository.create({
       cta: command.cta,
       name: command.name,
-      variables: command.variables,
+      variables: command.variables?.filter((variable) => {
```
scopsy, line 23: Is it possible maybe to avoid duplication a and have a helper function to clean to prepare the variables?

scopsy (commented):
Looks great, Is it possible to add a test for this functionality?

p-fernandez (approved):
🌟

apps/api/src/app/message-template/usecases/create-message-template/create-message-template.spec.ts
```diff
@@ -57,6 +65,7 @@ describe('Create Message Template', function () {
     expect(result.name).to.eql('test-message-template');
     expect(result.title).to.eql('test');
     expect(result.content).to.eql(content);
-    expect(result.variables).to.eql([]);
+    expect(result.variables?.length).to.eql(1);
+    expect(result.variables?.at(0)?.defaultValue).to.eql('test');
```
scopsy, line 69: Shouldn't the message be created with the variable even if the default value is false? And we just shouldn't override the sent message with this information? 

Because this variable is still was used in the template (if that's the usage of the variables array)

And only during trigger we should filter out to not override it with a string. Or maybe I'm missing something @davidsoderberg

PR #2763, opened 2023-02-14
feat: add overrides for email providers

davidsoderberg:


apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -205,9 +205,13 @@ export class SendMessageEmail extends SendMessageBase {
       to: email,
       subject,
       html,
-      from: command.payload.$sender_email || integration?.credentials.from || 'no-reply@novu.co',
+      from:
+        overrides?.email?.from || command.payload.$sender_email || integration?.credentials.from || 'no-reply@novu.co',
       attachments,
+      text: overrides?.email?.text,
```
p-fernandez, line 211: Does this fix any bug or it is a leftover from other task?

providers/mailersend/src/lib/mailersend.provider.ts
```diff
@@ -52,6 +52,14 @@ export class MailersendEmailProvider implements IEmailProvider {
       .setText(options.text)
       .setAttachments(attachments);
 
+    if (options.cc) {
```
p-fernandez, line 55: Should we do an extra check that it is an array?

providers/mailgun/src/lib/mailgun.provider.ts
```diff
@@ -44,6 +44,8 @@ export class MailgunEmailProvider implements IEmailProvider {
       to: emailOptions.to,
       subject: emailOptions.subject,
       html: emailOptions.html,
+      cc: emailOptions.cc ? emailOptions.cc[0] : undefined,
+      bcc: emailOptions.bcc ? emailOptions.bcc[0] : undefined,
```
p-fernandez, line 48: Mailgun doesn't accept `cc` and `bcc` as a list?

providers/postmark/src/lib/postmark.provider.ts
```diff
@@ -72,6 +72,8 @@ export class PostmarkEmailProvider implements IEmailProvider {
       HtmlBody: options.html,
       TextBody: options.html,
       Subject: options.subject,
+      Cc: options.cc ? options.cc[0] : undefined,
+      Bcc: options.bcc ? options.bcc[0] : undefined,
```
p-fernandez, line 76: Same question as for Mailgun. Doesn't Postmark take `cc` and `bcc` as list of emails?

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -119,7 +119,7 @@ export class SendMessageEmail extends SendMessageBase {
       return;
     }
 
-    const overrides = command.overrides[integration?.providerId] || {};
+    const overrides: Record<string, any> = command.overrides[integration?.providerId] || {};
```
scopsy, line 122: Aren't we taking the information from the "email" field and not from the providerId field? 

So when the user is sending a trigger he specifies it as:
```
novu.trigger('template', {
...data,
overrides: {
   email: { // <---- Here it's an email and not a provider Id 
     cc: []
   }
}
})
```

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -205,9 +205,13 @@ export class SendMessageEmail extends SendMessageBase {
       to: email,
       subject,
       html,
-      from: command.payload.$sender_email || integration?.credentials.from || 'no-reply@novu.co',
+      from:
```
scopsy, line 208: Dp we have an e2e test that shows this overrides functionality works? Would be great to have it

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -205,9 +205,13 @@ export class SendMessageEmail extends SendMessageBase {
...
+        overrides?.email?.from || command.payload.$sender_email || integration?.credentials.from || 'no-reply@novu.co',
```
scopsy, line 209:
```suggestion
        overrides?.email?.from || integration?.credentials.from || 'no-reply@novu.co',
```

$sender_email can be deprecated we don't use it since we move to the new novu system from stateless, so i think we can remove this.

providers/netcore/src/lib/netcore.provider.ts
```diff
@@ -57,6 +57,20 @@ export class NetCoreProvider implements IEmailProvider {
     body.personalizations[0].to[0] = new netcoreLib.EmailStruct();
     body.personalizations[0].to[0].email = options.to;
 
+    if (options.cc) {
+      body.personalizations[0].cc = options.cc.map((ccItem, index) => {
+        body.personalizations[0].cc[index] = new netcoreLib.EmailStruct();
+        body.personalizations[0].cc[index].email = ccItem;
+      });
```
scopsy, line 64: `map` needs to return a value, here `undefined` is returned so i don't think it will work, we should either use a forEach or a for loop if we are not returning .

I would suggest a cleaner approach perhaps:

```suggestion
      body.personalizations[0].cc = options.cc.map((ccItem) => {
        const emailStruct = new netcoreLib.EmailStruct();
        emailStruct.email = ccItem;
        
        return emailStruct;
      });
```

providers/netcore/src/lib/netcore.provider.ts
```diff
@@ -57,6 +57,20 @@ export class NetCoreProvider implements IEmailProvider {
...
+    }
+
+    if (options.bcc) {
+      body.personalizations[0].bcc = options.bcc.map((ccItem, index) => {
+        body.personalizations[0].bcc[index] = new netcoreLib.EmailStruct();
+        body.personalizations[0].bcc[index].email = ccItem;
+      });
```
scopsy, line 71: Same suggestion as from the cc section

providers/mailgun/src/lib/mailgun.provider.ts
davidsoderberg: Not what I could find...

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: I thought this was the wished structure....

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: we need to pass the override down if we like it to be sent in the providers

providers/mailgun/src/lib/mailgun.provider.ts
p-fernandez: Then I wouldn't apply BCC and CC to Mailgun and the other provider. Or if doing it inform the user that only one email is allowed per field. We will get a lot of questions from people asking why the mail was not delivered to all BCC and CC. 😭

providers/postmark/src/lib/postmark.provider.ts
davidsoderberg: Not according to the interface from the postmark package.

p-fernandez: Same comment as for Mailgun then. 🙁 
https://github.com/novuhq/novu/pull/2763#discussion_r1105756766

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: a good point but right now we expect this: 
```
novu.trigger('template', {
...data,
overrides: {
  providerId: {
     email: { // <---- Here it's an email and not a provider Id 
       cc: []
     }
  }
}
})
```
Will remove so email is not needed

providers/mailgun/src/lib/mailgun.provider.ts
davidsoderberg: @scopsy should we mention this in the docs maybe?
ping @iampearceman and @jainpawan21

providers/postmark/src/lib/postmark.provider.ts
davidsoderberg: @scopsy should we mention this in the docs maybe?
ping @iampearceman and @jainpawan21

scopsy:
@davidsoderberg could we also add a piece on the documentation site on this new functionality and available keys to override?

providers/mailgun/src/lib/mailgun.provider.ts
davidsoderberg: Never mind found a way of doing it interfaces was not clear enough

providers/postmark/src/lib/postmark.provider.ts
davidsoderberg: Never mind found a way of doing it interfaces was not clear enough

apps/api/src/app/events/e2e/send-message-email.e2e.ts
```diff
@@ -1,7 +1,49 @@
...
+      text: 'text',
+      cc: ['cc@nouv.co'],
+      bcc: ['bcc@nouv.co'],
+    });
+
+    expect(result.to).to.deep.equal(['to-reply@novu.co', 'override-to@novu.co']);
+    expect(result.from).to.equal('override-from@novu.co');
+    expect(result.text).to.equal('text');
+    expect(result.cc).to.deep.equal(['cc@nouv.co']);
+    expect(result.bcc).to.deep.equal(['bcc@nouv.co']);
```
p-fernandez, line 28:
```suggestion
      cc: ['cc@novu.co'],
      bcc: ['bcc@novu.co'],
    });

    expect(result.to).to.deep.equal(['to-reply@novu.co', 'override-to@novu.co']);
    expect(result.from).to.equal('override-from@novu.co');
    expect(result.text).to.equal('text');
    expect(result.cc).to.deep.equal(['cc@novu.co']);
    expect(result.bcc).to.deep.equal(['bcc@novu.co']);
```
Not that the test was bad. It was checking the values rightfully. 🙂

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -412,6 +419,23 @@ export class SendMessageEmail extends SendMessageBase {
   }
 }
 
+export const createMailData = (options: IEmailOptions, overrides: Record<string, any>): IEmailOptions => {
```
p-fernandez, line 422:
```suggestion
export const createMailData = (options: IEmailOptions, overrides: Record<string, string>): IEmailOptions => {
```
Do we have any override that is not string value?

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -10,14 +10,16 @@ export interface IProvider {
 }
 
 export interface IEmailOptions {
-  to: string | string[];
+  to: string[];
```
p-fernandez, line 13: I really like this change as simplifies everything. 👏🏻

providers/mailersend/src/lib/mailersend.provider.ts
```diff
@@ -52,6 +52,14 @@ export class MailersendEmailProvider implements IEmailProvider {
       .setText(options.text)
       .setAttachments(attachments);
 
+    if (options.cc && Array.isArray(options.cc)) {
+      emailParams.setCc(options.cc.map((ccItem) => new Recipient(ccItem)));
+    }
+
+    if (options.bcc && Array.isArray(options.bcc)) {
+      emailParams.setBcc(options.bcc.map((ccItem) => new Recipient(ccItem)));
+    }
```
p-fernandez, line 61: Slick!

p-fernandez (approved):
Great PR. We needed this. 🙌🏻

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: Some of them can be undefined, arrays or strings...

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
scopsy, line 122: I think we need to change this, since line 224 will never execute because of this.

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -218,6 +221,10 @@ export class SendMessageEmail extends SendMessageBase {
       }
     }
 
+    if (overrides?.email?.replyTo) {
```
scopsy, line 224: overrides here is always empty because we do the `const overrides: Record<string, any> = command.overrides[integration?.providerId] || {};` command earlier.

Should it be :
```suggestion
    if (command.overrides?.email?.replyTo) {
```
?

docs/docs/channels/email/index.mdx
```diff
@@ -14,6 +14,21 @@ When creating an email provider integration you will be asked to provide additio
 
 For some email providers including SendGrid you will have to authenticate the **From email address** to make sure you will send email messages using an authorized address.
 
+## Sending Email Overrides
+
+The overrides field supports an email property that should have the interface of this:
+
+```ts
```
scopsy, line 21: Could we maybe provide the full novu.trigger snippet here for people to better understand the context?

scopsy (approved):
Amazing work, left couple of comments 🎉

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: Not sure I follow...

PR #2769, opened 2023-02-15
feat: make sender name an input field for email card

davidsoderberg:
So sender name can be changed from each email template if needed empty field uses default from integration.

### DIFF
--- a/apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
+++ b/apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
@@ -362,14 +362,7 @@ export class SendMessageEmail extends SendMessageBase {
   ) {
     const mailFactory = new MailFactory();
     const mailHandler = mailFactory.getHandler(
-      {
-        ...integration,
-        credentials: {
-          ...integration.credentials,
-          senderName: senderName && senderName.length > 0 ? senderName : integration.credentials.senderName,
-        },
-        providerId: GetNovuIntegration.mapProviders(ChannelTypeEnum.EMAIL, integration.providerId),
-      },
+      SendMessageEmail.buildFactoryIntegration(integration, senderName),
       mailData.from
     );
 
@@ -428,6 +421,17 @@ export class SendMessageEmail extends SendMessageBase {
 
       return;
     }
+  }
+
+  public static buildFactoryIntegration(integration: IntegrationEntity, senderName?: string) {
+    return {
+      ...integration,
+      credentials: {
+        ...integration.credentials,
+        senderName: senderName && senderName.length > 0 ? senderName : integration.credentials.senderName,
+      },
+      providerId: GetNovuIntegration.mapProviders(ChannelTypeEnum.EMAIL, integration.providerId),
+    };
   }
 }
 
--- a/apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
+++ b/apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
@@ -361,10 +361,7 @@ export class SendMessageEmail extends SendMessageBase {
     senderName?: string
   ) {
     const mailFactory = new MailFactory();
-    const mailHandler = mailFactory.getHandler(
-      SendMessageEmail.buildFactoryIntegration(integration, senderName),
-      mailData.from
-    );
+    const mailHandler = mailFactory.getHandler(this.buildFactoryIntegration(integration, senderName), mailData.from);
 
     try {
       const result = await mailHandler.send(mailData);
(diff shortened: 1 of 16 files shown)


apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -350,14 +357,12 @@ export class SendMessageEmail extends SendMessageBase {
...
+    notification: NotificationEntity,
+    senderName?: string
   ) {
     const mailFactory = new MailFactory();
     const mailHandler = mailFactory.getHandler(
-      {
-        ...integration,
-        providerId: GetNovuIntegration.mapProviders(ChannelTypeEnum.EMAIL, integration.providerId),
-      },
+      SendMessageEmail.buildFactoryIntegration(integration, senderName),
```
p-fernandez, line 365: What's the goal of using a static method rather than the instance method considering where it is called it will be from a method of an instantiated class?

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -417,6 +422,17 @@ export class SendMessageEmail extends SendMessageBase {
       return;
     }
   }
+
+  public static buildFactoryIntegration(integration: IntegrationEntity, senderName?: string) {
+    return {
+      ...integration,
+      credentials: {
+        ...integration.credentials,
+        senderName: senderName && senderName.length > 0 ? senderName : integration.credentials.senderName,
```
p-fernandez, line 431:
```suggestion
        senderName: senderName?.length > 0 ? senderName : integration.credentials.senderName,
```
Just saving characteres.

apps/api/src/app/notification-template/e2e/create-notification-templates.e2e.ts
```diff
@@ -297,6 +299,92 @@ describe('Create Notification template - /notification-templates (POST)', async
...
+
+    expect(body.data).to.be.ok;
+    const template: INotificationTemplate = body.data;
+
+    expect(template._notificationGroupId).to.equal(testTemplate.notificationGroupId);
+    const message = template.steps[0];
+    expect(message.template?.senderName).to.equal('test');
+  });
+
+  it('should build factory integration', () => {
```
p-fernandez, line 333: Maybe this unit test should live closer to the SendMessageEmail use case and not in this E2E test suite? Like creating a `send-message-email.usecase.spec.ts` file, even if will only hold one test suite so far (for the `buildFactoryIntegration` method) but we might split this test in 3 so the expectations get documented.

p-fernandez (approved):
🌟 

2 considerations, but looks good!

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: Doing this typescript compliance about senderName possible being undefined...

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
davidsoderberg: No one really... I was thinking that we might reuse it in other instances but figured out now it is email specific...

apps/api/src/app/notification-template/e2e/create-notification-templates.e2e.ts
davidsoderberg: I kind of needing it to be in e2e so I can use the testserver to get an instance of the SendMessageEmail class

scopsy:
@davidsoderberg could you please send a loom/photo of how this new input looks like and behaves?

davidsoderberg:
Screenshot 2023-02-16 at 10 11 25

Screenshot 2023-02-16 at 10 11 33

Sure I am not sure the user will understand what the field is used for...

scopsy:
@davidsoderberg aggree, what if we will add the input title similar to the "Email Layout" one? And remove the Inbox view header and will keep it only the preview tab ?

davidsoderberg:
> @davidsoderberg aggree, what if we will add the input title similar to the "Email Layout" one? And remove the Inbox view header and will keep it only the preview tab ?

Right now we do not have in preview so lets think about that part when we get to it... I was thinking of tooltip instead to save space, what do you think about that?

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
p-fernandez: You are right. I missed that for not trying the PR in my local.

davidsoderberg:
Screenshot 2023-02-16 at 12 22 13

scopsy:
I think it makes much more sense @davidsoderberg, we can have the "date" in the preview section.

scopsy:
@davidsoderberg could you add a tooltip near the Sender Name to say that if empty we will use the one from integration page?

davidsoderberg:
Screenshot 2023-02-16 at 12 38 42

PR #2783, opened 2023-02-15
feat: add listmonk email provider

Codepowercode:

Added new provider for Listmonk Transactional Email
To be able to send email using Listmonk Transactional Email


apps/api/package.json
```diff
@@ -72,6 +72,7 @@
     "@novu/termii": "^0.11.0",
     "@novu/testing": "^0.11.0",
     "@novu/twilio": "^0.11.0",
+    "@novu/listmonkEmail": "^0.11.0",
```
scopsy, line 75: Could you please change the camelCasing to be kebab-cased in the package and file names? (sorry for the confusion in the generator, updating it soon.

scopsy (changes requested):
Thank you! To be honest i'm not quite sure if listmonk is valid at the current state of novu since we manage the content and listmonk only works with a template_id. 

What will happen is that all the workflows will use the same templates if it's only configured at the provider level.

We have some ideas of allowing defining external templates on nodes, but for now we cannot have a provider who doesn't accept HTML content when sending requires a template id. 

Do they have a way of sending HTML directly?

apps/api/package.json
Codepowercode: > Thank you! To be honest i'm not quite sure if listmonk is valid at the current state of novu since we manage the content and listmonk only works with a template_id.
> 
> What will happen is that all the workflows will use the same templates if it's only configured at the provider level.
> 
> We have some ideas of allowing defining external templates on nodes, but for now we cannot have a provider who doesn't accept HTML content when sending requires a template id.
> 
> Do they have a way of sending HTML directly?

No as of version 2.3.0 they don't have that functionality

jainpawan21:
Closing this PR as list monk has templateId as required field
https://listmonk.app/docs/apis/transactional/

PR #2789, opened 2023-02-15
NV-946 - Scheduled Delay step - add to trigger snippet

### LINKED ISSUE #2183: [NV-946] Scheduled Delay step - add to trigger snippet
With scheduled delay, a novu user can declare a path in payload of where the delayed date would be in the payload when the workflow triggers. 

I think it would be nice to add that path to the trigger snippet we show in the web UI especially as we show all other variables used in the templates. It will also help reduce misunderstandings and the need to debug. 

So, if this is defined on the web:

This will be the updated trigger snippet:

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-946](https://linear.app/novu/issue/NV-946/scheduled-delay-step-add-to-trigger-snippet)</sub>

gitstart-novu:
### What change does this PR introduce
This change adds the scheduled delay path to the trigger code snippet

### Why was this change needed

closes [#2183](https://github.com/novuhq/novu/issues/2183)

[Loom Video](https://www.loom.com/share/6c73cc7c208d404089042bfd2dfbab5e)


apps/web/src/components/templates/TriggerSnippetTabs.tsx
```diff
@@ -9,7 +14,13 @@ import * as get from 'lodash.get';
 const NODE_JS = 'Node.js';
 const CURL = 'Curl';
 
-export function TriggerSnippetTabs({ trigger }: { trigger: INotificationTrigger }) {
+export function TriggerSnippetTabs({
+  trigger,
+  template,
```
ainouzgali, line 19: Instead of adding template to props and the multiple changes that follows that change, might be better to just use getValues of steps. It will not cause rerenders :) 

Something like this:

Screen Shot 2023-02-19 at 17 46 18

gitstart-novu:
@ainouzgali this is ready for another review.

ainouzgali:
Hi @gitstart-novu, thank you so much. But, I have decided to reevaluate how to add variables to the trigger - instead of listening to the form steps, I would do it in the backend on create/update notification template. 
Your approach added unnecessary fields and inconsistencies that are not needed with a bit more deep refactor. 
Thank you again!

gitstart-novu:
@ainouzgali Thanks for the feedback. After a quick look, I can see how it would be cleaner to do it in the backend.

Will get to work on the new approach and update this in due time.

ainouzgali:
Thank you so much @gitstart-novu ! I have several ideas to expand this. I will take this one from here 💙

gitstart-novu:
@ainouzgali No problem then. Hope to see this resolved soon.

Looking forward to your solution :)

ainouzgali:
Hi @gitstart-novu, I have implemented this as part of [this pr](https://github.com/novuhq/novu/pull/2908) . Feel free to share your thoughts💙 Thanks again

PR #2794, opened 2023-02-15
fix: allow step filtering by nested payload/subscriber data

### LINKED ISSUE #2756: [NV-1670] 🐛 Bug Report: Cannot filter steps by custom subscriber data
### 📜 Description

Hi - first time contributing, so please be gracious! 

I noticed that we're not able to filter event trigger steps by custom subscriber data. In the following screenshot, if a subscriber has custom attributes `{customKey: customValue}` and the event is triggered to them, they will not receive the notification. 

Screenshot 2023-02-13 at 4 57 36 PM

### 👟 Reproduction steps

1. Create a subscriber with custom attributes, eg `"data": { "customKey": "customValue" }`
2. Add a step filter to a notification template as the above screenshot
3. Trigger the notification to the subscriber

### 👍 Expected behavior

I expect the app to apply step filters to custom subscriber data. In the scenario described above, a subscriber with matching customKey/customValue pair should receive the notification. 

If you add a filter for `data.customKey`, `EQUALS`, `customValue`, and there is a subscriber with `data.customKey: customValue`, that subscriber will not receive a notification from the event. 

Screenshot 2023-02-13 at 5 06 34 PM

It's not pictured, but the same happens if I use  `customValue` instead of `data.customValue` as the key in the step filter. 

I may be missing something since I'm new to this codebase, but it looks like a one-line code change in the `processFilterEquality(variables, fieldFilter, filterProcessingDetails)` function in 
 `api/src/app/events/usecases/trigger-event/message-matcher.service.ts`.

We currently extract the field filter value from the subscriber data using `_.get(variables, [fieldFilter.on, fieldFilter.field])` where 
- `variables` contains the keys: `payload`, `subscriber`, and `webhook` 
- `fieldFilter.on` contains the string value 'subscriber' or 'payload'
- `fieldFilter.field`  contains user input to search by (in the above example, `data.customKey`). 

lodash .get() can iterate through the keys of an object with either a string (eg `"key1[0].key2.key3"`) or with an array (eg `["key1", "0", "key2", "key3]`). The current implementation uses an array, but since the user passes a string input (eg `"data.customKey"`), we call the function as `_.get( {...}, ["subscriber", "data.customKey"] )`, using a mix of array and string values to define the path. It can be fixed if we build a single string for the path (`"subscriber.data.customKey"`) or if we `.split('.')` the field and build an array for the path (`["subscriber", "data", "customKey"]`)

Locally, I replaced 
```ts
const actualValue = _.get(variables, [fieldFilter.on, fieldFilter.field]);
```
with 
```ts
const actualValue = _.get(variables, `${fieldFilter.on}.${fieldFilter.field}`);
```
and it works as expected - but I haven't tested it very thoroughly.

<sub>[NV-1670](https://linear.app/novu/issue/NV-1670/🐛-bug-report-cannot-filter-steps-by-custom-subscriber-data)</sub>

chauandrew:

This change passes a string to lodash.get() instead of an array so that it is able to access nested data. For example, instead of calling `_.get(obj, ['subscriber', 'data.myKey'])`, we now call `_.get(obj, 'subscriber.data.myKey')`. 

Closes #2756, fixing a bug to allow filter steps to filter by nested data.

In the following screenshot, you can see that the notification has a step filter of (`subscriber.data.customKey`, `EQUALS`, `"customValue"`) and there is a user with custom data `{customKey: 'customValue'}`. When the notification is triggered, the user receives the notification. 

Screenshot 2023-02-15 at 12 44 04 PM

### DIFF
--- a/apps/api/src/app/events/usecases/trigger-event/message-matcher.service.spec.ts
+++ b/apps/api/src/app/events/usecases/trigger-event/message-matcher.service.spec.ts
@@ -252,17 +252,25 @@ describe('Message filter matcher', function () {
           {
             operator: 'EQUAL',
             value: 'nestedValue',
-            field: 'data.nestedKey.childKey',
+            field: 'data.nestedKey',
             on: FilterPartTypeEnum.SUBSCRIBER,
           },
         ]),
       }),
       {
         subscriber: {
+          firstName: '',
+          lastName: '',
+          email: '',
+          subscriberId: '',
+          deleted: false,
+          createdAt: '',
+          updatedAt: '',
+          _id: '',
+          _organizationId: '',
+          _environmentId: '',
           data: {
-            nestedKey: {
-              childKey: 'nestedValue',
-            },
+            nestedKey: 'nestedValue',
           },
         },
       }
@@ -279,12 +287,12 @@ describe('Message filter matcher', function () {
             operator: 'EQUAL',
             value: 'nestedValue',
             field: 'data.nestedKey.doesNotExist',
-            on: FilterPartTypeEnum.SUBSCRIBER,
+            on: FilterPartTypeEnum.PAYLOAD,
           },
         ]),
       }),
       {
-        subscriber: {
+        payload: {
           data: {
             nestedKey: {
               childKey: 'nestedValue',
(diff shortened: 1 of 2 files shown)


scopsy (approved):
Looks great! And loved the added tests

scopsy:
@chauandrew seems like we have some type errors on the spec files, mind taking a look at those? 🙏

scopsy (approved):

PR #2797, opened 2023-02-16
feat: converting apostrophes in variables correctly

davidsoderberg:


scopsy (approved):
🎉

apps/api/src/app/content-templates/usecases/compile-template/compile-template.usecase.ts
```diff
@@ -44,6 +44,8 @@ export class CompileTemplate {
 
     const template = Handlebars.compile(templateContent);
 
-    return template(command.data);
+    const result = template(command.data, {});
+
+    return result.replace(/&#x27;/g, "'");
```
ainouzgali, line 49: @davidsoderberg are we sure its only for ' ?
What about lets say &, I think that one also has a problem. And maybe others?

PR #2798, opened 2023-02-16
feat: add more integration fields for apns

davidsoderberg:
Two new fields for apns config in integration store, production existed before to the apns provider but was always set to true if cloud version was used. Topic was an required field from start and was missed during first implementation

Screenshot 2023-02-16 at 10 00 33

### DIFF
--- a/apps/web/src/pages/integrations/components/IntegrationInput.tsx
+++ b/apps/web/src/pages/integrations/components/IntegrationInput.tsx
@@ -47,8 +47,6 @@ export function IntegrationInput({
   }
 
   if (credential.type === 'switch') {
-    console.log(field);
-
     return (
       <Switch
         styles={() => ({
--- a/apps/web/src/pages/integrations/components/IntegrationInput.tsx
+++ b/apps/web/src/pages/integrations/components/IntegrationInput.tsx
@@ -1,5 +1,5 @@
 import { CredentialsKeyEnum, IConfigCredentials, secureCredentials } from '@novu/shared';
-import { Input, PasswordInput, Textarea } from '../../../design-system';
+import { Input, PasswordInput, Switch, Textarea } from '../../../design-system';
 
 export function IntegrationInput({
   credential,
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/integrations/components/IntegrationInput.tsx
```diff
@@ -46,6 +46,30 @@ export function IntegrationInput({
     );
   }
 
+  if (credential.type === 'switch') {
+    console.log(field);
```
p-fernandez, line 50: 🗑️

apps/web/src/pages/integrations/components/IntegrationInput.tsx
```diff
@@ -46,6 +46,30 @@ export function IntegrationInput({
...
+
+    return (
+      <Switch
+        styles={() => ({
+          root: {
+            display: 'block !important',
+            maxWidth: '100% !important',
```
p-fernandez, line 57: Do we have any other way of implement the style without using `!important`?

providers/apns/src/lib/apns.provider.ts
```diff
@@ -25,7 +27,7 @@ export class APNSPushProvider implements IPushProvider {
         keyId: config.keyId,
         teamId: config.teamId,
       },
-      production: process.env.NODE_ENV === 'prod',
+      production: config.production,
```
p-fernandez, line 30: What could be the outcome if a user enables the production switch while testing in the dev environment?

apps/web/src/pages/integrations/components/IntegrationInput.tsx
davidsoderberg: no this is what I know the closest way to do it and it did not work without it 😢

providers/apns/src/lib/apns.provider.ts
davidsoderberg: it uses production instead of sandbox for apns at Apple (I guess it is apple that is hosting this)

p-fernandez (approved):
🌟

PR #2800, opened 2023-02-16
fix(infra): upgrade and clean dependencies for provider creation script

p-fernandez:

It fixes dependencies upgrades, removal of unused ones and aligns the versions to the current monorepo ones.
Also gives better hint for the naming pattern (kebab-case) that we adopted recently.

Yesterday while @unicodeveloper was investigating the process of creating a new provider he ran into problems by mismatching versions, mainly of our `stateless` package. From there I decided it was better to upgrade the script until we find a way to automatically inject the latest versions of the needed dependencies.

### DIFF
--- a/_templates/provider/new/package.ejs.t
+++ b/_templates/provider/new/package.ejs.t
@@ -5,7 +5,7 @@
 
 {
   "name": "@novu/<%= name %>",
-  "version": "",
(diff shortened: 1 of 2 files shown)


scopsy (approved):
dios lo bendiga (novu notification center i18n style ;)

Issue #2714, opened 2023-02-09
[NV-540] After clearing notification name field - error is not shown again


ainouzgali:
After entering input into notification name in W.E and then deleting it. When clicking on workflow button, the required error should pop up. 

pIt does not navigate to the editor, so it is not recognized as valid, \\ which is good. But the error doesn't show.

[https://www.loom.com/share/96bfda1bff5d421492ed7e0045cb2186](https://www.loom.com/share/96bfda1bff5d421492ed7e0045cb2186)

For now, I added an astrick and it moves focus to the field - to be more clear dor the user.

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-540](https://linear.app/novu/issue/NV-540/after-clearing-notification-name-field-error-is-not-shown-again)</sub>

ainouzgali:
This is such a great ticket super super clear.

indeed it is a bug and we can get to it later - BL for now and would add to a small batch.

PR #2801, opened 2023-02-16
NV-540 - After clearing notification name field - error is not shown again

gitstart-novu:
### What change does this PR introduce
This PR introduces a new CSS rule that applies a border color of `colors.error` to an input field when it is in the invalid state and has focus. This change was made to provide a more visible indication to the user that there is an error with the input field, specifically when the notification name field in Notification settings is deleted and the workflow button is clicked. The new rule helps to focus the user's attention on the error state of the field and encourages them to correct the error.

### Why was this change needed

closes [#2714](https://github.com/novuhq/novu/issues/2714)

[Video demonstration of this fix](https://user-images.githubusercontent.com/57623705/217866864-4ffc352a-72c7-432e-b774-662566cc6694.mov)


p-fernandez (approved):
🌟

PR #2811, opened 2023-02-16
Nv 942 bug report no way to reset destroy the

### LINKED ISSUE #1289: 🐛 Bug Report: No way to reset / destroy the `widget_user_auth_token` in the iFrame embed.
### 📜 Description

I have a Vue app that embeds the iFrame option for displaying the notification center. When I log out the existing user and log in as a new user, the notifications don't get updated. 

The `widget_user_auth_token` is not reset and remains the same. Only after a hard refresh on the page does the notification list update. 

### 👟 Reproduction steps

1. Embed the iFrame in your app.
2. Log in & check the notification center (also see `widget_user_auth_token`).
3. Log out of the current user.
4. log in as a new user.
5. Check the notification center (and `widget_user_auth_token`).

### 👍 Expected behavior

Step 5 should result in different notifications (and `widget_user_auth_token`) from step 2.

Both the notifications and `widget_user_auth_token` for steps 2 and for remaining the same. 

### 💻 Operating system

MacOS

### 🤖 Node Version

16.4

_No response_

_No response_

BiswaViraj:

closes #1289 

- Added a logout method that removes the widget auth token, disconnects the WebSocket

### DIFF
--- a/packages/notification-center/src/shared/interfaces/index.ts
+++ b/packages/notification-center/src/shared/interfaces/index.ts
@@ -89,6 +89,7 @@ export interface INovuProviderContext {
   fetchingStrategy: IFetchingStrategy;
   setFetchingStrategy: (strategy: Partial<IFetchingStrategy>) => void;
   onLoad: (data: { organization: IOrganizationEntity }) => void;
+  logout: VoidFunction;
 }
 
 export interface INotificationsContext {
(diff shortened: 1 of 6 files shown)


packages/notification-center/src/shared/interfaces/index.ts
```diff
@@ -89,6 +89,7 @@ export interface INovuProviderContext {
   fetchingStrategy: IFetchingStrategy;
   setFetchingStrategy: (strategy: Partial<IFetchingStrategy>) => void;
   onLoad: (data: { organization: IOrganizationEntity }) => void;
+  logout: VoidFunction;
```
p-fernandez, line 92: Is this a TypeScript built in type for void functions? I didn't know about it.

p-fernandez (approved):
🌟

PR #2816, opened 2023-02-17
feat: hide content of textarea with multiline keys

davidsoderberg:

Fix so private key in apns can be multiline

Screenshot 2023-02-17 at 08 20 17

Screenshot 2023-02-17 at 08 24 54


apps/web/src/pages/integrations/components/IntegrationInput.tsx
```diff
@@ -13,6 +14,10 @@ export function IntegrationInput({
   register: any;
 }) {
   if (isNeededToHide(credential.key)) {
```
scopsy, line 16: I wonder if we can refactor this to have the PasswordInput if statement not here, but rather on the "string" type input 🤔

scopsy (approved):

apps/web/src/pages/integrations/components/IntegrationInput.tsx
davidsoderberg: yes lets create a tech debt for this 😄

PR #2825, opened 2023-02-17
feat(api): refactor of trigger event

p-fernandez:

Refactor of trigger event.
Also moves `ISubscribersDefine` interface to `@novu/shared` as it was being exported from `@novu/node`. Felt weird our API had that dependency from a client package. 

Trigger event calls ProcessSubscriber that inside was hosting the creation of the notification and the  building of the job steps for the jobs of the notification. Felt like a confusing implementation so decided to isolate both use cases into ProcessSubscriber and CreateNotification.

### DIFF
--- a/apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
@@ -5,11 +5,12 @@ import {
   NotificationTemplateEntity,
   SubscriberEntity,
 } from '@novu/dal';
-import { ISubscribersDefine, ITopic, TriggerRecipients } from '@novu/node';
+import { ITopic, TriggerRecipients } from '@novu/node';
 import {
   ChannelTypeEnum,
   StepTypeEnum,
   IEmailBlock,
+  ISubscribersDefine,
   TopicId,
   TopicKey,
   TopicName,
--- a/apps/api/src/app/events/usecases/create-notification/create-notification.command.ts
+++ b/apps/api/src/app/events/usecases/create-notification/create-notification.command.ts
@@ -1,0 +1,34 @@
+// TODO: We shouldn't be importing from DAL here. Needs big refactor throughout monorepo.
+import { NotificationTemplateEntity, SubscriberEntity } from '@novu/dal';
+import { ISubscribersDefine } from '@novu/shared';
+import { IsDefined, IsString, IsOptional } from 'class-validator';
+
+import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
+
+export class CreateNotificationCommand extends EnvironmentWithUserCommand {
+  @IsDefined()
+  @IsString()
+  identifier: string;
+
+  @IsDefined()
+  overrides: Record<string, Record<string, unknown>>;
+
+  @IsDefined()
+  payload: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+
+  @IsDefined()
+  subscriber: SubscriberEntity;
+
+  @IsDefined()
+  template: NotificationTemplateEntity;
+
+  @IsDefined()
+  to: ISubscribersDefine;
+
+  @IsString()
+  @IsDefined()
+  transactionId: string;
+
+  @IsOptional()
+  actor?: SubscriberEntity;
+}
--- a/apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
+++ b/apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
@@ -1,142 +1,35 @@
 import { Injectable } from '@nestjs/common';
-import {
-  NotificationRepository,
-  SubscriberRepository,
-  NotificationTemplateRepository,
-  SubscriberEntity,
-  JobEntity,
-  JobStatusEnum,
-  NotificationStepEntity,
-} from '@novu/dal';
-import { STEP_TYPE_TO_CHANNEL_TYPE, InAppProviderIdEnum, StepTypeEnum } from '@novu/shared';
-import { CreateSubscriber, CreateSubscriberCommand } from '../../../subscribers/usecases/create-subscriber';
-import { CreateLog } from '../../../logs/usecases';
+import { SubscriberRepository, SubscriberEntity } from '@novu/dal';
+import { ISubscribersDefine } from '@novu/shared';
+
 import { ProcessSubscriberCommand } from './process-subscriber.command';
-import { DigestFilterSteps } from '../digest-filter-steps/digest-filter-steps.usecase';
-import { DigestFilterStepsCommand } from '../digest-filter-steps/digest-filter-steps.command';
-import { CacheKeyPrefixEnum } from '../../../shared/services/cache';
-import { Cached } from '../../../shared/interceptors';
-import { ApiException } from '../../../shared/exceptions/api.exception';
-import {
-  GetDecryptedIntegrations,
-  GetDecryptedIntegrationsCommand,
-} from '../../../integrations/usecases/get-decrypted-integrations';
+
 import { subscriberNeedUpdate } from '../../../subscribers/usecases/update-subscriber';
+import { CreateSubscriber, CreateSubscriberCommand } from '../../../subscribers/usecases/create-subscriber';
 
 @Injectable()
 export class ProcessSubscriber {
-  constructor(
-    private subscriberRepository: SubscriberRepository,
-    private notificationRepository: NotificationRepository,
-    private createSubscriberUsecase: CreateSubscriber,
-    private createLogUsecase: CreateLog,
-    private notificationTemplateRepository: NotificationTemplateRepository,
-    private filterSteps: DigestFilterSteps,
-    private getDecryptedIntegrations: GetDecryptedIntegrations
-  ) {}
+  constructor(private createSubscriberUsecase: CreateSubscriber, private subscriberRepository: SubscriberRepository) {}
 
-  public async execute(command: ProcessSubscriberCommand): Promise<Omit<JobEntity, '_id'>[]> {
-    const template =
-      command.template ??
-      (await this.getNotificationTemplate({
-        _id: command.templateId,
-        environmentId: command.environmentId,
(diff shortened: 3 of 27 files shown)


apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
```diff
@@ -5,11 +5,12 @@ import {
   NotificationTemplateEntity,
   SubscriberEntity,
 } from '@novu/dal';
-import { ISubscribersDefine, ITopic, TriggerRecipients } from '@novu/node';
```
p-fernandez, line 8: Didn't make sense to me an interface that is shared in the monorepo to be exported from the Node.js client package.

apps/api/src/app/events/usecases/create-notification/create-notification.command.ts
```diff
@@ -0,0 +1,34 @@
...
+export class CreateNotificationCommand extends EnvironmentWithUserCommand {
+  @IsDefined()
+  @IsString()
+  identifier: string;
+
+  @IsDefined()
+  overrides: Record<string, Record<string, unknown>>;
+
+  @IsDefined()
+  payload: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
p-fernandez, line 17: I think here we would need to also make shareable the interface `ITriggerPayload` so we can remove all any set for payload property. But not 100% sure if it will be right interface.

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -1,161 +1,56 @@
...
-        environmentId: command.environmentId,
-        organizationId: command.organizationId,
-        userId: command.userId,
-        templateId: template._id,
-        notificationId: notification._id,
-      })
-    );
-
-    const jobs: Omit<JobEntity, '_id'>[] = [];
+  public async execute(command: ProcessSubscriberCommand): Promise<SubscriberEntity | undefined> {
```
p-fernandez, line 14: Now this use case does just one thing, process the subscriber.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -1,65 +1,109 @@
...
       data: {
-        triggerIdentifier: command.identifier,
+        triggerIdentifier: identifier,
       },
     });
 
+    const template = await this.notificationTemplateRepository.findByTriggerIdentifier(
+      command.environmentId,
+      command.identifier
+    );
```
p-fernandez, line 50: Potentially we could send the template from the queued data as during ParseEventRequest we already check the template exists, but I find sensible to retrieve it again in case during the queued time for the event, the template is being somehow deleted.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -1,65 +1,109 @@
...
+
+    /*
+     * Makes no sense to execute anything if template doesn't exist
+     * TODO: Send a 404?
+     */
+    if (!template) {
+      const message = 'Notification template could not be found';
+      Logger.error(message, LOG_CONTEXT);
+      throw new ApiException(message);
+    }
```
p-fernandez, line 60: Saving execution time and DB calls if we reach this scenario.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -1,65 +1,109 @@
...
+    if (actor) {
+      actorProcessed = await this.processSubscriber.execute(
+        ProcessSubscriberCommand.create({
+          environmentId,
+          organizationId,
+          userId,
+          subscriber: actor,
+        })
+      );
+    }
```
p-fernandez, line 75: Before, when sending the actor, we were searching for it in the DB inside the loop. Therefore we would be searching for the actor _number of subscribers times_. Not efficient. 😅

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -1,65 +1,109 @@
...
+        ProcessSubscriberCommand.create({
+          environmentId,
+          organizationId,
+          userId,
+          subscriber,
+        })
       );
+
+      // If no subscriber makes no sense to try to create notification
+      if (subscriberProcessed) {
```
p-fernandez, line 88: Comment says it all. Before we would try to create a notification even if the subscriber didn't exist in the database.

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -1,161 +1,56 @@
...
-        userId: command.userId,
-      })
-    );
-    const integration = integrations[0];
-
-    return integration?.providerId ?? InAppProviderIdEnum.Novu;
-  }
-  @Cached(CacheKeyPrefixEnum.NOTIFICATION_TEMPLATE)
-  private async getNotificationTemplate({ _id, environmentId }: { _id: string; environmentId: string }) {
-    return await this.notificationTemplateRepository.findById(_id, environmentId);
```
p-fernandez, line 131: I didn't move this code at all and removed it because we are passing the template all the time to the notification creation so not needed to cache its retrieval as we only need to look for it once.
Before it made sense as we were checking in the loop for every subscriber if the template existed.

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -1,161 +1,56 @@
...
   }
 
   private async createOrUpdateSubscriber(
-    command: Pick<ProcessSubscriberCommand, 'environmentId' | 'organizationId'>,
+    environmentId: string,
+    organizationId: string,
     subscriberPayload,
-    subscriber: SubscriberEntity | null
-  ) {
+    // TODO: Getting rid of this null would be amazing
```
LetItRock, line 47: after we do a `findBySubscriberId` we should have a check and throw if null

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
LetItRock: The use-case `TriggerEvent` is used by the `TriggerHandlerQueueService` and throwing that error won't work for the case when the trigger queue events are processed, because there is no REST context anymore... so I feel like it should throw some generic runtime error and we should catch and wrap it in other places... wdyt?

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
LetItRock: should we then return even before we enter this for loop?

LetItRock (approved):
awesome! 🚀 left some small suggestions

apps/api/src/app/events/usecases/create-notification/create-notification.command.ts
p-fernandez: Soon to be fixed.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
p-fernandez: Not totally understand the point you explain. `ApiException` is a generic runtime error. We already throw something similar in the method `validateTransactionIdProperty` that validates the entry data. And the worker at the end it has its own context, though mostly of the information provided will come from the database. I am happy to discuss this privately if it is easier because I couldn't understand what you mean, sorry. 😓

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
p-fernandez: Do you mean instead to do:
```
if (!subscriberProcessed) {
  continue;
}
```
?
RIght now I am surrounding the loop code to execute with the condition so nothing is executed if no subscriber. And the query we have to do it, to double check in case data stored in the queue and database data is not consistent.

apps/api/src/app/events/usecases/process-subscriber/process-subscriber.usecase.ts
p-fernandez: If we throw we would break the loop. That would mean that some notifications will be processed from the trigger but not all, leading to inconsistency. Unless we set up a rollback for the full trigger, for now I think it is better to leave the current behaviour of sending to the existing subscribers the notifications, and just skip it for non existent subscribers or failed subscribers.
Something worth to bring to tech summit to see what's the best direction we want to have for best fault tolerance. 👍🏻

LetItRock: ahhh ok sorry I was confused about this TODO comment, I think it's not possible to get rid of that null type in this function then because the subscriber is created in the `createSubscriberUsecase.execute` call when it's null...

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
LetItRock: `APIException` extends `BadRequestException` which is `HttpException`, but we are not in that HTTP context anymore, so I think that this is not an appropriate exception here.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
LetItRock: nvm, I feel like I didn't notice that the code is surrounded within the same if...

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
p-fernandez: 🤦🏻 We have a bigger problem then. We would need to change the definition of ApiException and the use in the monorepo. The are more places where it is used under the queue context. Adding to technical debt.

PR #2830, opened 2023-02-19
feat: add userHashForIntercom field in user schema

jainpawan21:

Add new field userHashForIntercom
For new users, calculate hash during register and store in user document
For old users, calculate hash during login

Closes [NV-1548](https://linear.app/novu/issue/NV-1548/add-identity-verification-for-intercom)


apps/api/src/app/auth/usecases/login/login.usecase.ts
```diff
@@ -62,6 +63,19 @@ export class Login {
       throw new UnauthorizedException(`Incorrect email or password provided.`);
     }
 
+    if (!user.userHashForIntercom) {
+      const intercomSecretKey = process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY as string;
+      const userHashForIntercom = createHmac('sha256', intercomSecretKey).update(email).digest('hex');
```
scopsy, line 68: Could we use user._id instead of the email here?

apps/api/src/app/auth/usecases/register/user-register.usecase.ts
```diff
@@ -29,11 +30,14 @@ export class UserRegister {
     if (existingUser) throw new ApiException('User already exists');
 
     const passwordHash = await bcrypt.hash(command.password, 10);
+    const intercomSecretKey = process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY as string;
+    const userHashForIntercom = createHmac('sha256', intercomSecretKey).update(email).digest('hex');
```
scopsy, line 34: Let's extract this to a helper function so we can reuse this logic in a single place

apps/api/src/app/auth/usecases/register/user-register.usecase.ts
```diff
@@ -29,11 +30,14 @@ export class UserRegister {
...
     const user = await this.userRepository.create({
       email,
       firstName: command.firstName.toLowerCase(),
       lastName: command.lastName?.toLowerCase(),
       password: passwordHash,
+      userHashForIntercom: userHashForIntercom,
```
scopsy, line 40: Can we only set this property when there is a `userHashForIntercom` is present?

scopsy (changes requested):
Looks great! I've added a few comments 🎉

apps/api/src/app/auth/usecases/register/user-register.usecase.ts
```diff
@@ -36,6 +37,19 @@ export class UserRegister {
       password: passwordHash,
     });
 
+    const intercomSecretKey = process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY as string;
```
scopsy, line 40: This whole flow needs to be wrapped with 
```
if (process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY) {
...
```

apps/api/src/app/auth/usecases/login/login.usecase.ts
```diff
@@ -62,6 +63,19 @@ export class Login {
...
+      const intercomSecretKey = process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY as string;
+      const userHashForIntercom = createHash(intercomSecretKey, user._id);
+      await this.userRepository.update(
+        { _id: user._id },
+        {
+          $set: {
+            userHashForIntercom: userHashForIntercom,
+          },
+        }
+      );
```
scopsy, line 76: This also needs to be wrapped with an if statement for
```
process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY
```

scopsy (changes requested):
Great work! Let's make sure that the whole intercom verification flow is wrapped in an `if` statement to not execute on self hosted or intercom-less environments

libs/dal/src/repositories/user/user.entity.ts
```diff
@@ -47,4 +47,6 @@ export class UserEntity {
     times: number;
     lastFailedAttempt: string;
   };
+
+  userHashForIntercom?: string;
```
scopsy, line 51: thinking out loud because of: https://linear.app/novu/issue/NV-1456/enable-hmac-encryption-on-our-clients-notification-center

Should we changes this to:

```
servicesHashes: { intercom: string }
```
to support other service in the future including "Novu" ?

jainpawan21: Yes this option will be better to accomodate future requirements

scopsy: @jainpawan21 do you want to tackle this already in this PR?

jainpawan21: Yes, I will update this PR
do we need to run any migration to add extra property in user schema?

apps/api/src/app/auth/usecases/login/login.usecase.ts
```diff
@@ -62,6 +63,19 @@ export class Login {
...
 
+    if (process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY && !user.servicesHashes?.intercom) {
+      const intercomSecretKey = process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY as string;
+      const userHashForIntercom = createHash(intercomSecretKey, user._id);
+      await this.userRepository.update(
+        { _id: user._id },
+        {
+          $set: {
+            servicesHashes: { intercom: userHashForIntercom },
+          },
```
scopsy, line 74:
```suggestion
          $set: {
            'servicesHashes.intercom': userHashForIntercom },
          },
```

The previous method could erase all the serviceHashes contents with this object

apps/api/src/app/auth/usecases/register/user-register.usecase.ts
```diff
@@ -36,6 +37,21 @@ export class UserRegister {
...
+    if (process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY) {
+      const intercomSecretKey = process.env.INTERCOM_IDENTITY_VERIFICATION_SECRET_KEY as string;
+      const userHashForIntercom = createHash(intercomSecretKey, user._id);
+      await this.userRepository.update(
+        { _id: user._id },
+        {
+          $set: {
+            servicesHashes: {
+              intercom: userHashForIntercom,
+            },
```
scopsy, line 49: Same here :)

libs/dal/src/repositories/user/user.entity.ts
scopsy: No, mongo db  doesn't need that :)

scopsy (approved):

scopsy:
@jainpawan21 the tests seem to consistenly fail :( could you try resolving the conflicts here?

jainpawan21:
@scopsy resolved conflict and failing tests.
only build docker is failing

PR #2832, opened 2023-02-19
feat: add analytics event to track usage of provider webhooks

davidsoderberg:
Two tracking events so we can track how much provider webhooks is used.

To be able to understand usage of our provider webhook feature

### DIFF
--- a/apps/webhook/src/shared/shared.module.ts
+++ b/apps/webhook/src/shared/shared.module.ts
@@ -1,4 +1,5 @@
 import { Module } from '@nestjs/common';
+import { AnalyticsService } from '@novu/application-generic';
 import {
   DalService,
   UserRepository,
@@ -40,6 +41,16 @@ const PROVIDERS = [
     },
   },
   ...DAL_MODELS,
+  {
+    provide: AnalyticsService,
+    useFactory: async () => {
+      const analyticsService = new AnalyticsService(process.env.SEGMENT_TOKEN);
+
+      await analyticsService.initialize();
+
+      return analyticsService;
+    },
+  },
 ];
 
 @Module({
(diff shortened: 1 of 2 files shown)


apps/webhook/src/shared/shared.module.ts
```diff
@@ -40,6 +41,16 @@ const PROVIDERS = [
     },
   },
   ...DAL_MODELS,
+  {
+    provide: AnalyticsService,
+    useFactory: async () => {
+      const analyticsService = new AnalyticsService(process.env.SEGMENT_TOKEN);
```
davidsoderberg, line 47: This needs to be added in terraform I guess? @scopsy

apps/webhook/src/webhooks/usecases/webhook/webhook.usecase.ts
```diff
@@ -34,6 +36,17 @@ export class Webhook {
       channel,
     });
 
+    const member = await this.memberRepository.getOrganizationAdminAccount(command.organizationId);
```
davidsoderberg, line 39: We do not know who created the Webhook so I am using admin account here...

apps/webhook/src/webhooks/usecases/webhook/webhook.usecase.ts
```diff
@@ -44,7 +57,19 @@ export class Webhook {
...
+
+    if (member) {
+      this.analyticsService.track('[Webhook] - Provider Webhook events parsed', member._userId, {
+        _organization: command.organizationId,
+        _environmentId: command.environmentId,
+        providerId,
+        channel,
+        events,
+      });
+    }
```
scopsy, line 70: I think actually that this one will be enough (no need to have the top one. wdyt?

scopsy (approved):
Looks great! I will add the terraform entry aswell

apps/webhook/src/webhooks/usecases/webhook/webhook.usecase.ts
davidsoderberg: I added both to be able to track if we get more webhooks the we parse...

PR #2836, opened 2023-02-19
Nv 1130 generic error message should point at

ainouzgali:

Show more explicit errors on step node. on update click and on Test workflow. 

https://www.loom.com/share/dd2fba28845544b48263c64edb1fd432


apps/web/src/components/workflow/FlowEditor.tsx
```diff
@@ -162,6 +162,8 @@ export function FlowEditor({
         const oldNode = nodes[i + 1] || { position: { x: 0, y: 120 } };
         const newId = (step._id || step.id) as string;
 
+        await triggerErrors('steps');
```
scopsy, line 165: Why this one is needed here? It triggers the form validation if i understand correctly?

apps/web/src/design-system/template-button/ChannelButton.tsx
```diff
@@ -109,12 +115,23 @@ export function ChannelButton({
...
   const { classes: popoverClasses } = usePopoverStyles();
+  const viewport = useViewport();
+
+  const {
+    watch,
+    formState: { errors },
+  } = useFormContext();
 
-  const { watch } = useFormContext();
+  let showErrors = initialErrors;
```
scopsy, line 125: This one sounds like a boolean to me, maybe we can rename this to: `formErrorsContent` or similar?

scopsy (approved):
Amazing work! Added a feq comments, other than that looks good to me. 🎉

apps/web/src/components/workflow/FlowEditor.tsx
ainouzgali: This is to get the meaningful errors before the submit of the form. So we would know what are the errors immediately on adding a node @scopsy

PR #2851, opened 2023-02-20
refactor: auth and spotlight context to be own hooks

davidsoderberg:
To not reuse auth context or spotlight context all over the place, instead we create a specific hook for each of them.
if we in the future needs to hide other logic behind the hook, we can easily do it if we do it like this...


apps/web/src/App.tsx
```diff
@@ -154,218 +151,216 @@ function App() {
...
-                <SpotLightProvider>
-                  <Routes>
-                    <Route path={ROUTES.AUTH_SIGNUP} element={<SignUpPage />} />
-                    <Route path={ROUTES.AUTH_LOGIN} element={<LoginPage />} />
-                    <Route path={ROUTES.AUTH_RESET_REQUEST} element={<PasswordResetPage />} />
-                    <Route path={ROUTES.AUTH_RESET_TOKEN} element={<PasswordResetPage />} />
-                    <Route path={ROUTES.AUTH_INVITATION_TOKEN} element={<InvitationPage />} />
-                    <Route path={ROUTES.AUTH_APPLICATION} element={<CreateOrganizationPage />} />
+            <AuthProvider>
+              <SpotLightProvider>
```
scopsy, line 155: I think that the spotlight shouldn't work on the non registered pages, maybe we can move it to be only part of the layout?

apps/web/src/App.tsx
```diff
@@ -154,218 +151,216 @@ function App() {
...
+                            <NotificationList />
+                          </SpotLight>
+                        </RequiredAuth>
+                      }
+                    />
+                    <Route
+                      path="/general-started"
+                      element={
+                        <RequiredAuth>
+                          <SpotLight>
```
scopsy, line 232: Is it possible to wrap all logged in routes with those? so we won't have to do it manually, (where the layout is defined) I remember it was possible in previous versions.

apps/web/src/App.tsx
```diff
@@ -154,218 +151,216 @@ function App() {
...
+                  />
+                  <Route
+                    path={ROUTES.PARTNER_INTEGRATIONS_VERCEL_LINK_PROJECTS_EDIT}
+                    element={
+                      <RequiredAuth>
+                        <LinkVercelProjectPage type="edit" />
+                      </RequiredAuth>
+                    }
+                  />
+                  <Route element={<AppLayout />}>
```
scopsy, line 179: I think we can define all the reuseable providers here. Wdyt?

apps/web/src/components/auth/CreateOrganizationForm.tsx
```diff
@@ -22,7 +22,7 @@ export function CreateOrganization({}: Props) {
   } = useForm({});
 
   const navigate = useNavigate();
-  const { setToken, token } = useContext(AuthContext);
+  const { setToken, token } = useAuthContext();
```
scopsy, line 25: Nice!

apps/web/src/components/layout/components/HeaderNav.tsx
```diff
@@ -27,11 +27,11 @@ const menuItem = [
 const headerIconsSettings = { color: colors.B60, width: 30, height: 30 };
 
 export function HeaderNav({}: Props) {
-  const { currentOrganization, currentUser, logout } = useContext(AuthContext);
+  const { currentOrganization, currentUser, logout } = useAuthContext();
   const { colorScheme, toggleColorScheme } = useMantineColorScheme();
   const { themeStatus } = useLocalThemePreference();
   const dark = colorScheme === 'dark';
-  const { addItem } = useContext(SpotlightContext);
+  const { addItem } = useSpotlightContext();
```
scopsy, line 34: Love the refactoring!

scopsy (commented):
Amazing work! Added a few small suggestions <3

apps/web/src/App.tsx
davidsoderberg: this is only storing actions for the spotlight that is wrapping each route...

apps/web/src/App.tsx
davidsoderberg: Not sure I follow what you suggest...

scopsy: Sorry for the confusion :P 

I meant that instead of copy pasting Those two provider wrappers on all routes individually, we can wrap them once at: https://github.com/novuhq/novu/pull/2851#discussion_r1111564308

this will affect all routes under the `<AppLayout/>` component

scopsy (approved):
🎉

apps/web/src/components/layout/AppLayout.tsx
```diff
@@ -8,52 +8,61 @@ import { colors } from '../../design-system';
...
+                    <small style={{ color: 'lightGrey' }}>
+                      Event Id: {eventId}.
+                      <br />
+                      {error.toString()}
+                    </small>
+                  </code>
+                </>
+              )}
+            >
+              <RequiredAuth>
```
LetItRock, line 56: I feel like this `RequiredAuth` should wrap all other components here... we should do a auth check logic as early as possible... otherwise other components, like header, and side nav can start running it's logic which in case of not logged in user should not happen

apps/web/src/store/spotlightContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    const newItems = [...items, ...item];
+    newItems.sort((a, b) => (b.order || 0) - (a.order || 0));
+
+    setItems(newItems);
+  };
+
+  const removeItem = (id: string) => {
+    setItems(items.filter((item) => item.id !== id));
+  };
```
LetItRock, line 38: I would recommend wrapping these functions with the `useCallback`

apps/web/src/store/spotlightContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return <SpotlightContext.Provider value={{ items, addItem, removeItem }}>{children}</SpotlightContext.Provider>;
```
LetItRock, line 40: and memoize the `{ items, addItem, removeItem }` object here

apps/web/src/store/authContext.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <AuthContext.Provider
+      value={{
+        currentUser: user,
+        currentOrganization: organization,
+        organizations,
+        token,
+        logout,
+        setToken,
+        jwtPayload,
+      }}
```
LetItRock, line 40: this object should be memoized

LetItRock (approved):
awesome work! 🎉 left a few small suggestions ;)

apps/web/src/store/spotlightContext.tsx
davidsoderberg: this is just moved code and will not be part of this refactoring

apps/web/src/store/authContext.tsx
davidsoderberg: this is just moved code and will not be part of this refactoring

PR #2855, opened 2023-02-20
Nv 1471 mark all as read does not mark all

BiswaViraj:

- Added a new endpoint to mark all messages as read by Feed.
- Fixes the issues where clicking `mark all as read` on the notification-center` didn't marked all the messages.


packages/notification-center/src/hooks/useMarkNotificationAsReadByFeed.ts
```diff
@@ -0,0 +1,58 @@
...
+
+            return {
+              pageParams: infiniteData.pageParams,
+              pages,
+            };
+          }
+        );
+        queryClient.refetchQueries(INFINITE_NOTIFICATIONS_QUERY_KEY, {
+          exact: false,
+        });
```
ainouzgali, line 45: I think you should only do one of those. The set query data actually replaces the data in the query's cache. So you either replace the data or do the refetch on line 43

ainouzgali (commented):
@BiswaViraj great work :) 
I'm wondering if instead of another endpoint + hook, if we could use the existing markAllNotificationsAsRead with some modifications? Or at least the markNotificationsAs we get from the useMarkNotificationsAs hook?  
It seems the only difference from what we already have is doing it by feed , and maybe we could add it as a parameter 
@BiswaViraj @LetItRock  wdyt?

packages/notification-center/src/hooks/useMarkNotificationAsReadByFeed.ts
BiswaViraj: Agreed, will remove the refetch one

packages/notification-center/src/hooks/useMarkNotificationAsReadByFeed.ts
```diff
@@ -0,0 +1,58 @@
...
+        queryClient.refetchQueries(FEED_UNSEEN_COUNT_QUERY_KEY, {
+          exact: false,
+        });
+        queryClient.refetchQueries(UNSEEN_COUNT_QUERY_KEY, {
+          exact: false,
+        });
```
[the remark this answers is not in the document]
BiswaViraj: While testing the unseen_count didn't get updated until I closed and reopened the widget :/
I will give it another go,

BiswaViraj:
@ainouzgali ,
The old implementation only updates the `read, seen` status of the messages whose `ids` has been passed as a payload,
This new endpoint will mark all the unread messages and doesn't require the `messageIds`.

apps/api/src/app/widgets/usecases/mark-all-message-as-read-by-feed/mark-all-message-as-read-by-feed.usecase.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+    const response = await this.messageRepository.markAllUnreadAsReadByFeed(
+      subscriber._id,
+      command.environmentId,
+      feedId
+    );
+
+    this.queueService.wsSocketQueue.add({
+      event: 'unseen_count_changed',
+      userId: command.subscriberId,
```
ainouzgali, line 40: Notice you are using subscriber._id to get count and command.subscriberId to update socket. I do not think they are the same. 
I think you need to use subscriber._id. 
Maybe that's the reason your unseen_count was not updated?

BiswaViraj: @ainouzgali, Thanks, it fixed it :D

packages/notification-center/src/hooks/useMarkNotificationAsReadByFeed.ts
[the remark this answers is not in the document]
BiswaViraj: Removed them, it works after updating the subscriber id at the socket event

ainouzgali (commented):
Hi @BiswaViraj, I tried a few things and found some usecases that aren't working as expected. Let me know if you'd like to go over it :)

apps/api/src/app/widgets/widgets.controller.ts
```diff
@@ -224,6 +227,25 @@ export class WidgetsController {
     return await this.markMessageAsUsecase.execute(command);
   }
 
+  @ApiOperation({
+    summary: "Mark subscriber feed's all unread messages as read",
+  })
+  @UseGuards(AuthGuard('subscriberJwt'))
+  @Post('/messages/feed/markAsRead')
```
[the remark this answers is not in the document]
LetItRock: I'm wondering why can't we modify the `widgets/messages/markAs` endpoint to allow marking all the messages read/seen whatever?
So like if the message ids won't be provided then we will mark all read/seen, otherwise only selected.
@BiswaViraj

BiswaViraj:
@LetItRock , We can modify the `widgets/messages/markAs`, but then we will have to add the support for `feedIds` as well which might be too much if this or that for a single API
wdyt?

LetItRock:
> @LetItRock , We can modify the `widgets/messages/markAs`, but then we will have to add the support for `feedIds` as well which might be too much if this or that for a single API wdyt?

I think it's not too much actually and it might be useful if someone would like to mark notifications for a feed.

@scopsy wdyt

BiswaViraj:
Didn't went with the `widgets/messages/markAs` because of the following:

- how to handle when both `messageIDs` and `FeedId` are provided
- single endpoint handling
  - messageIds (mark messages with the given ids)
  - feedIds (mark messages where feedid is the given id)
  - absence of both messageID & feedId (mark all messages)

Instead I saw we already have a `messages/seen` endpoint and created a `messages/read` endpoint
also modified it's usecase to handle both `seen` and `read` and serves a single responsibility which is marking all the messages as seen or read.

packages/notification-center/src/store/notifications-provider.context.tsx
```diff
@@ -133,6 +140,7 @@ export function NotificationsProvider({
       markAllNotificationsAsRead,
       markAllNotificationsAsSeen,
```
BiswaViraj, line 141: Also, I feel the names for `markAllNotificationAsRead` and `markAllNotificationsAsSeen` are incorrect.
It should be appropriate to rename these to `markFetchedNotificationsAs**` according to what they do.
And then we can rename the `markAllNotificationsByFeed` to just `markAllNotifications`.
But renaming these will be a breaking change but much needed one.

apps/api/src/app/widgets/usecases/mark-all-messages-as/mark-all-messages-as.usecase.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    await this.queueService.wsSocketQueue.add({
+      event: 'unread_count_changed',
+      userId: subscriber._id,
+      payload: {
+        unreadCount: 0,
+      },
+    });
+
```
scopsy, line 52: Shouldn't we conditionaly dire those depending on the `command.markAs` ?

scopsy (approved):

apps/api/src/app/widgets/usecases/mark-all-messages-as/mark-all-messages-as.usecase.ts
BiswaViraj: ah yes, added them

PR #2862, opened 2023-02-20
fix: reset page number on tab switch

BiswaViraj:

- Resets the page number on tab switch on the changes page


apps/web/src/pages/changes/PromoteChangesPage.tsx
```diff
@@ -43,6 +43,7 @@ export function PromoteChangesPage() {
       value: PENDING,
       content: (
         <ChangesTable
+          key={page}
```
BiswaViraj, line 46: The page number indicator at the bottom of the page sometimes will be out of sync with the current page number,
Adding a `key` fixed that issue

scopsy (approved):

Issue #2795, opened 2023-02-16
[NV-1675] EmptyState prop on PopoverNoificationCenter bad typing


ainouzgali:
## Reproduction Steps

* Create a react project and use the notification center with an empty state component:

```
<PopoverNotificationCenter
    emptyState={() => <LookingGlassIcon/>}
/>
```

* When using the following syntax react send a warning message:

```
Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it
```

## Expected Behaviour

We need to modify the TS definition and change the signature to be:

```
<PopoverNotificationCenter
    emptyState={<LookingGlassIcon/>}
/>
```

And update the relevant documentation section where needed.

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-1675](https://linear.app/novu/issue/NV-1675/emptystate-prop-on-popovernoificationcenter-bad-typing)</sub>

nikitagrossman:
id like to help

ainouzgali:
Amazing! @nikitagrossman feel free to reach out if you need anything🚀

PR #2866, opened 2023-02-20
fix: EmptyState prop on PopoverNoificationCenter bad typing

nikitagrossman:
EmptyState prop on PopoverNoificationCenter fix

EmptyState prop on PopoverNoificationCenter bad typing

closes: #2795


scopsy:
Need to update the docs aswell: https://docs.novu.co/notification-center/react/api-reference/#popovernotificationcenter

@LetItRock do you think we should update props aswell here? footer header and etc..?

ainouzgali:
Also missing change in https://github.com/novuhq/novu/blob/next/packages/notification-center/src/shared/interfaces/index.ts#L60

scopsy (approved):

PR #2873, opened 2023-02-21
Nv 1743 refactor location of hooks

davidsoderberg:
Move all hooks into hooks folder from store and api

To make it easier to find our custom hooks


apps/web/src/utils/debounce.ts
```diff
@@ -1,18 +0,0 @@
-import { useEffect, useState } from 'react';
-
-export default function useDebounce(value, delay = 500) {
```
p-fernandez, line 3: Not used anymore?

p-fernandez (approved):
🌟

apps/web/src/utils/debounce.ts
davidsoderberg: No it was not so I removed it ;)

LetItRock (approved):

PR #2876, opened 2023-02-21
fix(api): test email send with the novu provider issue fix

LetItRock:
When using the test email send with a Novu provider the API app raises an error.

https://novu-r9.sentry.io/issues/3908824474/?project=6248811&query=is%3Aunresolved&referrer=issue-stream

Fixes the issue.

Screenshot 2023-02-21 at 12 59 38

Screenshot 2023-02-21 at 13 02 13


apps/api/src/.env.test
```diff
@@ -64,4 +64,4 @@ FF_IS_TOPIC_NOTIFICATION_ENABLED=true
 STORE_NOTIFICATION_CONTENT=true
 
 MAX_NOVU_INTEGRATION_MAIL_REQUESTS=300
-NOVU_EMAIL_INTEGRATION_API_KEY=
+NOVU_EMAIL_INTEGRATION_API_KEY=test
```
LetItRock, line 67: for the e2e tests to run the logic that calculates the Novu provider limits

apps/api/src/app/events/usecases/send-message/test-send-email.usecase.ts
```diff
@@ -65,25 +70,40 @@ export class SendTestEmail {
...
+    );
 
     try {
       await mailHandler.send(mailData);
+      this.analyticsService.track('Test Email Sent - [Events]', command.userId, {
+        _organization: command.organizationId,
+        _environment: command.environmentId,
+        channel: ChannelTypeEnum.EMAIL,
+        providerId,
+      });
```
LetItRock, line 106: added a new segment analytics event - please note it's sent to analytics after the email is sent

apps/api/src/app/integrations/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -54,7 +56,7 @@ export class GetNovuIntegration {
         providerId: CalculateLimitNovuIntegration.getProviderId(command.channelType),
         ...limit,
       });
-      throw new Error(`Limit for Novus ${command.channelType.toLowerCase()} provider was reached.`);
+      throw new ApiException(`Limit for Novus ${command.channelType.toLowerCase()} provider was reached.`);
```
LetItRock, line 59: when the Novu provider limit is reached we should throw the `ApiException` not the general `Error`

apps/api/src/app/events/usecases/send-message/test-send-email.usecase.ts
```diff
@@ -65,25 +70,40 @@ export class SendTestEmail {
...
-    const mailHandler = mailFactory.getHandler(integration, mailData.from);
+  private async sendMessage(
+    integration: IntegrationEntity,
+    mailData: IEmailOptions,
+    mailFactory: MailFactory,
+    command: TestSendMessageCommand
+  ) {
+    const { providerId } = integration;
+    const mailHandler = mailFactory.getHandler(
+      { ...integration, providerId: GetNovuIntegration.mapProviders(ChannelTypeEnum.EMAIL, providerId) },
```
LetItRock, line 95: this line actually fixes the original issue

apps/api/src/app/events/usecases/send-message/test-send-email.usecase.ts
```diff
@@ -12,13 +15,15 @@ import { TestSendMessageCommand } from './send-message.command';
 import { ApiException } from '../../../shared/exceptions/api.exception';
 import { CompileEmailTemplate } from '../../../content-templates/usecases/compile-email-template/compile-email-template.usecase';
 import { CompileEmailTemplateCommand } from '../../../content-templates/usecases/compile-email-template/compile-email-template.command';
+import { GetNovuIntegration } from '../../../integrations/usecases/get-novu-integration';
 
 @Injectable()
 export class SendTestEmail {
```
p-fernandez, line 21: I am sorry for this but I had renamed the class and file name in a PR that just got merged yesterday. I can help you with the right changes if needed.

apps/api/src/app/events/usecases/send-message/test-send-email.usecase.ts
p-fernandez, line 106: 👏🏻

apps/api/src/app/events/usecases/send-message/test-send-email.usecase.ts
```diff
@@ -65,25 +70,40 @@ export class SendTestEmail {
...
+  private async sendMessage(
+    integration: IntegrationEntity,
+    mailData: IEmailOptions,
+    mailFactory: MailFactory,
+    command: TestSendMessageCommand
+  ) {
+    const { providerId } = integration;
+    const mailHandler = mailFactory.getHandler(
+      { ...integration, providerId: GetNovuIntegration.mapProviders(ChannelTypeEnum.EMAIL, providerId) },
+      mailData.from
```
p-fernandez, line 96: I would move all this bit of code inside the `try` so in case of misconfiguration for the MailHandler or parsing error in the providers mapping we would still get signalled that the error was `Unexpected provider error`. Easier to identify at debugging as downstream we have no error identification.

apps/api/src/app/events/e2e/test-email.e2e.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  it('should not allow sending test email when Novu provider limit is reached', async function () {
+    await deleteEmailIntegration();
+    await reachNovuProviderLimit();
+
+    try {
+      await sendTestEmail(requestDto);
+    } catch (e) {
+      expect(e.response.status).to.equal(400);
```
p-fernandez, line 102: I don't think that 400 Bad Request is the right code when hitting the limit. Find either 409 Conflict or 422 Unprocessable Entity more adequate, as any of them indicate that the request is valid and understood by the server but resource status doesn't allow to proceed. Rather than confuse the client stating the request in wrongly formatted somehow.
Anyway, this is inherited from existing functionality but might be worth to improve it. Feel free to ignore it for this PR and we can address it in a different one.

LetItRock: I do agree, I'll update it ;)

apps/api/src/app/events/usecases/send-message/test-send-email.usecase.ts
p-fernandez: I saw the conflicted file in the pipeline and I wrote that comment. Good wasn't too hard to fix!

p-fernandez (approved):
🌟

PR #2878, opened 2023-02-21
feat(api): performance service

p-fernandez:

Implements a service that allows to gather performance data and Node.js internals data for certain areas of the API.
Also adds a performance test to understand the duration of different executions under different loads.

Will be helpful to study performance of the sensible parts of the Trigger Event and what is the overhead of the distributed locking mechanism. Also can help to set the base for load performance test suites (in a way).

Screenshot 2023-02-21 at 12 32 20


SebastianStehle:
Why are you not just using OTLP?

p-fernandez:
> Why are you not just using OTLP?

I guess you are referring to OpenTelemetry, if not I don't know what you might be talking about. Acronyms are hard.

To be honest I didn't think about it because this had a clear purpose of measuring and not to be able to export data. @Cliftonz is working on general analytics so I don't want to step on his work. In case he decides to integrate OpenTelemetry, downstream this implementation `perf_hooks` could be easily replaced and integrate it.

SebastianStehle:
Yes, I am about open telemetry.

The most important primitive of OTLP is the span. It represents a unit of work and you annotate your code to track spans. For example method calls, db calls, http requests and so on. Spans have a correlation Ids to represent the parent span and there is a protocol to add this correlation IDs to http requests for distributed tracing.

Most implementations for like this:

1. You have a way to create spans.
2. You define one or many outputs (exporters), e.g. NewRelic, OTLP (Open Telemetry protocol) and so on.
3. You define one or more processors, e.g. to enrich the spans or to filter them.

So what I would do:

1. Configure Open Telemetry.
2. Annotate your code with custom spans.
3. Write a custom exporter that provides the performance data (e.g. for tests) or so.

You can then get rid of a lot of code and the performance data can also be used for self hosting, e.g. in Azure Application Insights or Google Stackdriver and so on.

Cliftonz:
@p-fernandez I agree, I think this will be fine for now. It's a good start.

However in the long run I agree with @SebastianStehle that we should be doing open-telemetry, However doing this in the trigger engine and mail parser is much harder then the nest based apps as there are easy OS modules for it.

 https://github.com/pragmaticivan/nestjs-otel

@SebastianStehle I will put adding open-telemetry in the backlog to add for the OS deployments.
This would also be accompanied by setting up Prometheus metrics too.

apps/api/src/app/events/events.controller.ts
```diff
@@ -69,6 +71,8 @@ export class EventsController {
     @UserSession() user: IJwtPayload,
     @Body() body: TriggerEventRequestDto
   ): Promise<TriggerEventResponseDto> {
+    const mark = this.performanceService.buildEndpointTriggerEventMark(body.transactionId as string);
```
LetItRock, line 74: maybe we should hold all this logic in the decorator? I can imagine it like this:
```
@MeasurePerformance('EventsController.trackEvent')
async trackEvent(
```
then I don't have to deal with the service

apps/api/src/app/events/services/performance-service/index.ts
```diff
@@ -0,0 +1,139 @@
...
+
+enum MarkFunctionNameEnum {
+  CREATE_NOTIFICATION_JOBS = 'createNotificationJobs',
+  DIGEST_FILTER_STEPS = 'digestFilterSteps',
+  ENDPOINT_TRIGGER_EVENT = 'endpoint:triggerEvent',
+  TRIGGER_EVENT = 'triggerEvent',
+}
+
+@Injectable()
+export class EventsPerformanceService {
```
LetItRock, line 13: I don't think that we need to create an events-specific performance service, I feel like all this might be generic code, that can be reused in any context. This way we won't end up writing per context performance services. So I see it with the decorator as I mentioned before.

The `publishResults` function can be defined as the helper function that will be called from anywhere we need to get the stats.

apps/api/src/app/events/services/performance-service/index.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    this.performanceService.calculateAverage(MarkFunctionNameEnum.ENDPOINT_TRIGGER_EVENT, endpointTriggerEventMeasures);
+    this.performanceService.calculateAverage(MarkFunctionNameEnum.TRIGGER_EVENT, triggerEventMeasures);
+    this.performanceService.calculateAverage(
+      MarkFunctionNameEnum.CREATE_NOTIFICATION_JOBS,
+      createNotificationJobsMeasures
+    );
+    this.performanceService.calculateAverage(MarkFunctionNameEnum.DIGEST_FILTER_STEPS, digestFilterStepsMeasures);
+  }
```
LetItRock, line 75:
```suggestion
  private publishMeasures(showLogs: boolean): void {
    const measures: Record<string, number[]> = {};

    this.performanceService.measures.forEach((measure) => {
      const functionName = measure.name.split(':')[0];
      if (!measures[functionName]) {
        measures[functionName] = [measure.duration];

        return;
      }

      measures[functionName].push(measure.duration);

      if (showLogs) {
        Logger.debug(
          `Duration: ${measure.duration.toFixed(2)} ms| Id: ${measure.name}`,
          PerformanceContextEnum.CONTEXT_MEASURE
        );
      }
    });

    Object.keys(measures).forEach((key) => {
      this.performanceService.calculateAverage(MarkFunctionNameEnum.ENDPOINT_TRIGGER_EVENT, measures[key]);
    });
  }
```

apps/api/src/app/events/services/performance-service/index.ts
```diff
@@ -0,0 +1,139 @@
...
+    subscriberId: string
+  ): IMark {
+    const mark = {
+      id: `${MarkFunctionNameEnum.DIGEST_FILTER_STEPS}:event:${transactionId}:template:${templateId}:notification:${notificationId}:subscriber:${subscriberId}:steps`,
+    };
+
+    return this.setStart(mark);
+  }
+
+  public buildEndpointTriggerEventMark(transactionId: string): IMark {
```
LetItRock, line 104: these all won't be needed if we will create generic function with mark like:
```
const mark = {
      id: `${functionName}:${uuid}`,
    };
```

apps/api/src/app/events/services/performance-service/index.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  public buildDigestFilterStepsMark(
+    transactionId: string,
+    templateId: string,
+    notificationId: string,
+    subscriberId: string
+  ): IMark {
+    const mark = {
+      id: `${MarkFunctionNameEnum.DIGEST_FILTER_STEPS}:event:${transactionId}:template:${templateId}:notification:${notificationId}:subscriber:${subscriberId}:steps`,
+    };
```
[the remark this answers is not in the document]
p-fernandez: Those are needed for having individual marks and to help to debug different executions if needed. And to have the right data as they are being accumulated I needed to make it unique, if not they would get overriden everytime it was executed. I saw your mention regarding generating a random UUID but that would make more difficult to debug any of the executions to be honest.
Right now the publishing of the individual marks is disabled as it was cluttering the log output and don't need it.

apps/api/src/app/events/services/performance-service/index.ts
p-fernandez: I try to think of the different modules as if they were its own microservice, so making this domain related service allows to:
- Keep isolated in the domain the specific mark generation.
- Have their own kind of calculations and what publish and what not (think in a different domain we might want to have different calculations, instead of average other algorithm, publish different results, etc)

That doesn't mean that's going to happen, to be reused elsewhere, so I can understand why you are suggesting something more generic.

apps/api/src/app/events/services/performance-service/index.ts
p-fernandez: Addressing this comment here: https://github.com/novuhq/novu/pull/2878#discussion_r1117205847

tsssdev:
> > Why are you not just using OTLP?
> 
> I guess you are referring to OpenTelemetry, if not I don't know what you might be talking about. Acronyms are hard.
> 
> To be honest I didn't think about it because this had a clear purpose of measuring execution times and was not meant to be able to export data. Mostly to add a quick tool for @tsssdev and his request about how to be able to measure Trigger Event execution performance. @Cliftonz is working on general analytics so I don't want to step on his work. In case he decides to integrate OpenTelemetry, downstream this implementation `perf_hooks` could be easily replaced and integrate it.

@p-fernandez How we generate load? do we have load testing infrastructure in place? Under decent to high load only we can see real issues with locking or with out locking.. if we have Dev/QA type of environment that can take decent load then we can implement quick load tests with K6(https://github.com/grafana/k6 ) or any other similar tool.

Cliftonz:
@tsssdev Our testing bed is dev.web.novu.co and we are looking to add k6 load tests in the api e2e test folder. If you want to create a pr for that we would be more then happy to look at it and merge it.

apps/api/src/app/events/services/performance-service/index.ts
LetItRock: I don't know to me it would be more to maintain... but yeah maybe let's start with something simple and then adjust to the new requirements in the future...

LetItRock (approved):

apps/api/src/app/events/events.controller.ts
Cliftonz: @LetItRock  This is something nest-otel can cover for us

Cliftonz (approved):
I think this is a good start, lets get it merged in and we can go from here.

PR #2882, opened 2023-02-22
Nv 1754 move page specific components to page

davidsoderberg:
Move components to components folders inside of pages

To create a better structure for our web app


LetItRock (approved):
🔥

PR #2885, opened 2023-02-22
feat(api): homogenize trigger event use cases and fix tech debt

p-fernandez:

Homogenise the Trigger Event module use cases and fixes some tech debt related.
Also moves MessageMatcher to its own location as being an use case and not being used by TriggerEvent usecase.

Tech debt.


apps/api/src/app/events/usecases/index.ts
```diff
@@ -56,7 +54,6 @@ export const USE_CASES = [
   SendMessageDelay,
   SendTestEmail,
   MapTriggerRecipients,
-  MapTriggerRecipientsCommand,
```
p-fernandez, line 59: This was a mistake I introduced time ago.

apps/api/src/app/events/usecases/send-message/send-test-email.usecase.ts
```diff
@@ -21,7 +23,7 @@ export class SendTestEmail {
     private getDecryptedIntegrationsUsecase: GetDecryptedIntegrations
   ) {}
 
-  public async execute(command: TestSendMessageCommand) {
+  public async execute(command: SendTestEmailCommand) {
```
p-fernandez, line 26: We had a mixing naming style for this use case and command. Homogenised.

libs/shared/src/types/builder/filter.types.ts
```diff
@@ -1,5 +1,11 @@
 import type { BuilderFieldOperator } from './builder.types';
 
+export enum TimeOperatorEnum {
```
p-fernandez, line 3: Added for tighter validation.

apps/api/src/app/events/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -458,7 +460,7 @@ async function findAsync<T>(array: T[], predicate: (t: T) => Promise<boolean>):
 }
 
 async function filterAsync<T>(arr: T[], callback: (item: T) => Promise<boolean>): Promise<T[]> {
-  const fail = Symbol();
+  const fail = Symbol('Filter Async failure');
```
p-fernandez, line 463: DeepSource flagged this.

apps/api/src/app/events/usecases/send-message/send-test-email.command.ts
```diff
@@ -0,0 +1,23 @@
...
+export class SendTestEmailCommand extends EnvironmentWithUserCommand {
+  @IsDefined()
+  contentType: 'customHtml' | 'editor';
+  @IsDefined()
+  payload: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+  @IsDefined()
+  @IsString()
+  subject: string;
+  preheader?: string;
+  senderName?: string;
```
djabarovgeorge, line 15: Should we add IsOptional() here?

libs/shared/src/types/builder/filter.types.ts
djabarovgeorge: Loved it 👍🏻

djabarovgeorge (approved):
Looks good to me, added small comment there

PR #2886, opened 2023-02-22
feat: add addition params for onboarding setup

djabarovgeorge:

Adds addition params to run on the onboarding script.

In order to trigger event on the demo app.


apps/web/src/pages/quick-start/steps/Setup.tsx
```diff
@@ -146,8 +147,11 @@ const LoaderWrapper = styled.div`
   margin-top: 10px;
 `;
 
-function updateCodeSnipped(codeSnippet: string, environmentIdentifier: string) {
-  return codeSnippet.replace(APPLICATION_IDENTIFIER, environmentIdentifier);
+function updateCodeSnipped(codeSnippet: string, environmentIdentifier: string, apiKey: string) {
+  return codeSnippet
+    .replace(APPLICATION_IDENTIFIER, environmentIdentifier)
+    .replace(API_KEY, apiKey ?? '')
```
djabarovgeorge, line 153: Added API key in order to pass it to the demo app.

apps/web/src/pages/quick-start/steps/Setup.tsx
```diff
@@ -146,8 +147,11 @@ const LoaderWrapper = styled.div`
...
+    .replace(ENVIRONMENT, process.env.REACT_APP_API_URL === 'https://api.novu.co' ? '' : 'dev');
```
djabarovgeorge, line 154: We need this one in order to provide a more fluent flow for the product team so they won't need to update the API and WS URLs on the demo app.

I was not sure what indication to add here at the moment I am using the API URL, if some one have any other suggestion i would love to hear about them :)

apps/web/src/pages/quick-start/steps/Setup.tsx
```diff
@@ -72,12 +78,7 @@ export function Setup() {
         {
           template: {
             type: StepTypeEnum.IN_APP,
-            content:
-              'Welcome to Novu! Click on this notification to <b>visit the cloud admin panel</b> managing this message',
-            cta: {
-              type: ChannelCTATypeEnum.REDIRECT,
-              data: { url: `/templates/edit/${notificationTemplate?._id}` },
-            },
+            content: 'Welcome to Novu! <b>visit the cloud admin panel</b> managing this message',
```
djabarovgeorge, line 81: I removed the CTA because the snippets do not have the click handlers, @scopsy let me know what you think.

apps/web/src/pages/quick-start/steps/Setup.tsx
```diff
@@ -21,13 +24,17 @@ import { When } from '../../../components/utils/When';
...
+import { getApiKeys } from '../../../api/environment';
 
 export function Setup() {
   const [notificationTemplate, setNotificationTemplate] = useState<INotificationTemplate>();
   const { framework } = useParams();
   const { groups } = useNotificationGroup();
   const { templates = [], loading } = useTemplates();
   const { environment } = useEnvController();
+  const { data: apiKeys } = useQuery<{ key: string }[]>(['getApiKeys'], getApiKeys);
+  const apiKey = apiKeys?.length ? apiKeys[0].key : '';
```
p-fernandez, line 36: Are we always going to receive an array? If there are other potential types we could receive in this call might worth to check it is array rather than its length, as if receiving a string the condition would pass.

apps/web/src/pages/quick-start/steps/Setup.tsx
```diff
@@ -146,8 +148,14 @@ const LoaderWrapper = styled.div`
   margin-top: 10px;
 `;
 
-function updateCodeSnipped(codeSnippet: string, environmentIdentifier: string) {
-  return codeSnippet.replace(APPLICATION_IDENTIFIER, environmentIdentifier);
+function updateCodeSnipped(codeSnippet: string, environmentIdentifier: string, apiKey: string) {
```
p-fernandez, line 151:
```suggestion
function updateCodeSnippet(codeSnippet: string, environmentIdentifier: string, apiKey: string) {
```
Typo.

apps/web/src/pages/quick-start/steps/Setup.tsx
```diff
@@ -146,8 +148,14 @@ const LoaderWrapper = styled.div`
...
+  const concatUrls = process.env.REACT_APP_ENVIRONMENT !== 'prod' || !!process.env.REACT_APP_DOCKER_HOSTED_ENV;
+
+  return codeSnippet
+    .replace(APPLICATION_IDENTIFIER, environmentIdentifier)
+    .replace(API_KEY, apiKey ?? '')
+    .replace(BACKEND_API_URL, concatUrls ? process.env.REACT_APP_API_URL || 'http://localhost:3000' : '')
+    .replace(BACKEND_SOCKET_URL, concatUrls ? process.env.REACT_APP_WS_URL || 'http://localhost:3002' : '');
```
p-fernandez, line 158: Shouldn't we use `window._env_.REACT_APP_API_URL` and `window._env_.REACT_APP_WS_URL` to avoid to have spread all the Web app the URL calculation?

https://github.com/novuhq/novu/blob/6e5b03ef3fadca0121957d1dd4eab13e863831e9/apps/web/src/config/index.ts#L12

If for any reason they are not the right variables to use here, maybe centralising in the `config` the calculation would be a good idea to have all that isolated in one place and not through the codebase. What do you think?

apps/web/src/pages/quick-start/steps/Setup.tsx
djabarovgeorge: At the moment it should be always an array, as it is on the environment entity.

apps/web/src/pages/quick-start/steps/Setup.tsx
djabarovgeorge: Good catch! I do it all the time, but not always catch it 🤦‍♂️

apps/web/src/pages/quick-start/steps/Setup.tsx
djabarovgeorge: Good catch, I will add it to the config file. Thanks!

p-fernandez (approved):
🌟

LetItRock (approved):

PR #2887, opened 2023-02-22
feat(api): move the providers decryption for the template out of loop

p-fernandez:

I moved the template providers decryption out of the step generation loop when creating a notification for every subscriber.

I consider we were doing an expensive operation multiple times when we could just do it once for all the template steps. Downstream, the decryption is using synchronous calls to the `crypto` module (`createDecipheriv` functionality), so we were doing (number of subscribers * number of steps) synchronous calls. We can expect they might be resource expensive as being cryptographic functionalities so it seems sensible to only do the minimum needed to achieve our goals. And because for now, the flow is set to just have one template when triggering an event we can safely reuse the providers of said template for the different subscribers that will be passed in the `to` property.

### DIFF
--- a/apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.command.ts
+++ b/apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.command.ts
@@ -1,6 +1,6 @@
 // TODO: We shouldn't be importing from DAL here. Needs big refactor throughout monorepo.
 import { NotificationTemplateEntity, SubscriberEntity } from '@novu/dal';
-import { ISubscribersDefine } from '@novu/shared';
+import { ChannelTypeEnum, ISubscribersDefine } from '@novu/shared';
 import { IsDefined, IsString, IsOptional } from 'class-validator';
 
 import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
@@ -21,6 +21,9 @@ export class CreateNotificationJobsCommand extends EnvironmentWithUserCommand {
 
   @IsDefined()
   template: NotificationTemplateEntity;
+
+  @IsDefined()
+  templateProviderIds: Map<ChannelTypeEnum, string>;
 
   @IsDefined()
   to: ISubscribersDefine;
--- a/apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
+++ b/apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
@@ -1,6 +1,19 @@
-import { JobEntity, JobRepository, NotificationTemplateRepository, NotificationRepository } from '@novu/dal';
+import {
+  JobEntity,
+  JobRepository,
+  NotificationTemplateEntity,
+  NotificationTemplateRepository,
+  NotificationRepository,
+} from '@novu/dal';
 import { Injectable, Logger } from '@nestjs/common';
-import { StepTypeEnum, ExecutionDetailsSourceEnum, ExecutionDetailsStatusEnum } from '@novu/shared';
+import {
+  ChannelTypeEnum,
+  ExecutionDetailsSourceEnum,
+  ExecutionDetailsStatusEnum,
+  InAppProviderIdEnum,
+  StepTypeEnum,
+  STEP_TYPE_TO_CHANNEL_TYPE,
+} from '@novu/shared';
 import * as Sentry from '@sentry/node';
 
 import { TriggerEventCommand } from './trigger-event.command';
@@ -14,6 +27,10 @@ import {
   CreateExecutionDetailsCommand,
   DetailEnum,
 } from '../../../execution-details/usecases/create-execution-details/create-execution-details.command';
+import {
+  GetDecryptedIntegrations,
+  GetDecryptedIntegrationsCommand,
+} from '../../../integrations/usecases/get-decrypted-integrations';
 import { ApiException } from '../../../shared/exceptions/api.exception';
 
 const LOG_CONTEXT = 'TriggerEventUseCase';
@@ -27,6 +44,7 @@ export class TriggerEvent {
     private jobRepository: JobRepository,
     private notificationRepository: NotificationRepository,
     private notificationTemplateRepository: NotificationTemplateRepository,
+    private getDecryptedIntegrations: GetDecryptedIntegrations,
     protected createExecutionDetails: CreateExecutionDetails
   ) {}
 
@@ -56,6 +74,13 @@ export class TriggerEvent {
       Logger.error(message, LOG_CONTEXT);
       throw new ApiException(message);
     }
+
+    const templateProviderIds = await this.getProviderIdsForTemplate(
+      command.userId,
+      command.organizationId,
+      command.environmentId,
+      template
+    );
 
     const jobs: Omit<JobEntity, '_id'>[][] = [];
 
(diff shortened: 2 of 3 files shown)


apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -175,4 +201,52 @@ export class TriggerEvent {
...
+      const type = step.template?.type;
+      if (type) {
+        const channelType = STEP_TYPE_TO_CHANNEL_TYPE.get(type);
+
+        if (channelType) {
+          const provider = await this.getProviderId(userId, organizationId, environmentId, channelType);
+          if (provider) {
+            providers.set(channelType, provider);
+          } else {
+            providers.set(channelType, InAppProviderIdEnum.Novu);
```
p-fernandez, line 223: I am keeping this for retrocompatibility but can't grasp the need of setting a default provider if no provider for the channel. Is related to our free trial provider? @davidsoderberg

djabarovgeorge: I am not sure (i do think we need to add the novu integration by default as the in-app channel, that way we can use it more securely and not as a default)

Small note maybe we should add it only if it is in-app channelType. (i am pretty sure it could be otherwise but still)

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -175,4 +201,52 @@ export class TriggerEvent {
...
+  }
+
+  private async getProviderId(
+    userId: string,
+    organizationId: string,
+    environmentId: string,
+    channelType: ChannelTypeEnum
+  ): Promise<string> {
+    const integrations = await this.getDecryptedIntegrations.execute(
+      GetDecryptedIntegrationsCommand.create({
```
djabarovgeorge, line 239: I think we can pass here findOne:true as a param so on the command, that way you wont fetch by 'find'

djabarovgeorge:
@p-fernandez Looks awesome and great catch on the unnecessary compute.

I really liked the refactor, but (not related to this PR at all!! 🙂) I am not sure if we need the provider id on the job entity at all, more than that I think it is a bug. 

The reason is that for chat and push we allow to have multiple integration providers (maybe to other channels in the future as well), more then that and the thing is that we refetch the integration on the send-message-type(sms/email, etc..) (which is good which means that if the integration was updated we get the most updated one).

So due to the refetch i am not sure we need it on the job entity at all unless we use it somewhere (i had not a chance to validate)

p-fernandez:
> @p-fernandez Looks awesome and a great catch on the unnecessary compute.
> 
> I really liked the refactor, but (not related to this PR at all!! 🙂) I am not sure if we need the provider id on the job entity at all, more than that I think it is a bug.
> 
> The reason is that for chat and push we allow to have multiple integration providers (maybe to other channels in the future as well), more then that and the thing is that we refetch the integration on the send-message-type(sms/email, etc..) (which is good which means that if the integration was updated we get the most updated one).
> 
> So due to the refetch i am not sure we need it on the job entity at all unless we use it somewhere (i had not a chance to validate)

Indeed for multiprovider channels this should change and probably the way of fetching the integrations information. If adding the provider to the job is a bug I have not enough information right now to comment. Would need a deeper dig. 🙂

djabarovgeorge:
> > @p-fernandez Looks awesome and a great catch on the unnecessary compute.
> > I really liked the refactor, but (not related to this PR at all!! 🙂) I am not sure if we need the provider id on the job entity at all, more than that I think it is a bug.
> > The reason is that for chat and push we allow to have multiple integration providers (maybe to other channels in the future as well), more then that and the thing is that we refetch the integration on the send-message-type(sms/email, etc..) (which is good which means that if the integration was updated we get the most updated one).
> > So due to the refetch i am not sure we need it on the job entity at all unless we use it somewhere (i had not a chance to validate)
> 
> Indeed for multiprovider channels this should change and probably the way of fetching the integrations information. If adding the provider to the job is a bug I have not enough information right now to comment. Would need a deeper dig. 🙂

It could be a bug or that could be redundant data or none, maybe we should ask on some channel about it or/and create a ticket about it. 🤔

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
p-fernandez: Should I add that check? Do not mind if we find it is safe to do. If not, I'd rather leave it for a different PR.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
p-fernandez: Checking `GetDecryptedIntegrations` code I do not find a big difference between both options as downstream still operates as an array and does all the checks adding the one entity found to an array.
https://github.com/novuhq/novu/blob/e32b5c2bb4d48570342769a9305fbe1e2d2a1209/apps/api/src/app/integrations/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts#L30

djabarovgeorge: You are right, the only difference as far as I know will be the query operation find/findOne when 'find' will query all the collection, and 'findOne' will stop after the first find.

apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
djabarovgeorge: Tend to agree it was a mall thought I had.
That's why I think we need to treat 'novu' in-app as a regular integration.

p-fernandez:
> > > @p-fernandez Looks awesome and a great catch on the unnecessary compute.
> > > I really liked the refactor, but (not related to this PR at all!! 🙂) I am not sure if we need the provider id on the job entity at all, more than that I think it is a bug.
> > > The reason is that for chat and push we allow to have multiple integration providers (maybe to other channels in the future as well), more then that and the thing is that we refetch the integration on the send-message-type(sms/email, etc..) (which is good which means that if the integration was updated we get the most updated one).
> > > So due to the refetch i am not sure we need it on the job entity at all unless we use it somewhere (i had not a chance to validate)
> > 
> > 
> > Indeed for multiprovider channels this should change and probably the way of fetching the integrations information. If adding the provider to the job is a bug I have not enough information right now to comment. Would need a deeper dig. 🙂
> 
> It could be a bug or that could be redundant data or none, maybe we should ask on some channel about it or/and create a ticket about it. 🤔

As discussed with the team the goal of providerId in the job entity:
- analytics
- information in activity feed in the future
- potential fallbacks between multiple providers

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.command.ts
```diff
@@ -22,6 +22,9 @@ export class CreateNotificationJobsCommand extends EnvironmentWithUserCommand {
   @IsDefined()
   template: NotificationTemplateEntity;
 
+  @IsDefined()
+  templateProviderIds: Map<ChannelTypeEnum, string>;
```
LetItRock, line 26: there is nothing related to the template, so I feel it should be called `channelToProviderIds`?

LetItRock (approved):
hawkeye, I would never notice that🎖️

PR #2890, opened 2023-02-23
chore(dal): fixed the mongo schema types and repositories to return the correct mapped types

LetItRock:

This is the first part of fixing the dal types. At the end of the second part we would have functions to return the correct types when using repository functions, for ex. `findById` will return `Entity | null`. Right now it's just `Entity` which results in errors in the API app when we forget to add the null check.

The Mongoose [docs](https://mongoosejs.com/docs/typescript.html) are suggesting using the `ObjectId` for the field that is "id", but we do type them with `string` type. Then our repositories are querying DB and doing a mapping to the entity type just to have strings as ids. But having the wrong types on the schema breaks all the Mongoose types system and model functions.

In this PR I've introduced the `DBModel` interfaces that the Mongoose Schema uses and the `Entity` type is returned back from the repositories like it was before, so nothing will have to be changed on the API app side.


apps/api/migrations/changes-migration.ts
```diff
@@ -12,8 +12,8 @@ import {
   OrganizationRepository,
 } from '@novu/dal';
 import { ChangeEntityTypeEnum, MemberRoleEnum } from '@novu/shared';
-import { CreateChange } from '../src/app/change/usecases/create-change.usecase';
-import { CreateChangeCommand } from '../src/app/change/usecases/create-change.command';
+import { CreateChange } from '../src/app/change/usecases/create-change/create-change.usecase';
+import { CreateChangeCommand } from '../src/app/change/usecases/create-change/create-change.command';
```
LetItRock, line 16: this is the API migration file, I've fixed the imports and ts issues here

apps/api/src/app/events/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -46,9 +47,16 @@ export class SendMessageInApp extends SendMessageBase {
...
+    protected getDecryptedIntegrationsUsecase: GetDecryptedIntegrations
   ) {
-    super(messageRepository, createLogUsecase, createExecutionDetails, subscriberRepository);
+    super(
+      messageRepository,
+      createLogUsecase,
+      createExecutionDetails,
+      subscriberRepository,
+      getDecryptedIntegrationsUsecase
+    );
```
LetItRock, line 59: added `GetDecryptedIntegrations` as it's required by the super class, check below

apps/api/src/app/events/usecases/send-message/send-message.base.ts
```diff
@@ -21,7 +21,7 @@ export abstract class SendMessageBase extends SendMessageType {
     protected createLogUsecase: CreateLog,
     protected createExecutionDetails: CreateExecutionDetails,
     protected subscriberRepository: SubscriberRepository,
-    protected getDecryptedIntegrationsUsecase?: GetDecryptedIntegrations
+    protected getDecryptedIntegrationsUsecase: GetDecryptedIntegrations
```
LetItRock, line 24: this should not be optional as we have the code that uses this use-case

libs/dal/src/repositories/base-repository.ts
```diff
@@ -1,11 +1,11 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { ClassConstructor, plainToInstance } from 'class-transformer';
-import { Document, Model, Query, Types, ProjectionType, FilterQuery } from 'mongoose';
+import { Model, Types, ProjectionType, FilterQuery, UpdateQuery } from 'mongoose';
 
-export class BaseRepository<T_Query, T_Response> {
-  public _model: Model<any & Document>;
+export class BaseRepository<T_DBModel, T_MappedEntity> {
```
LetItRock, line 5: Did a lot of type fixes here (no logic changes), but basically, the `BaseRepository` requires now to pass the `T_DBModel` which is a DB model, and `T_MappedEntity` which is the type that we map the DB object to. 
So like the Mongoose model return `DBModel` objects and we map them to `Entity`, just to have ids as strings, not the ObjectIds.

libs/dal/src/repositories/change/change.entity.ts
```diff
@@ -22,3 +26,18 @@ export class ChangeEntity {
...
+  _creatorId: Types.ObjectId;
+
+  _environmentId: Types.ObjectId;
+
+  _organizationId: Types.ObjectId;
+
+  _entityId: Types.ObjectId;
+
+  _parentId?: Types.ObjectId;
+};
```
LetItRock, line 43: This is a `DBModel` that we pass to the Mongoose schema, please note that id fields have types as `Types.ObjectId`. So we will have here just to override types for the ids.

libs/dal/src/repositories/change/change.entity.ts
```diff
@@ -1,13 +1,17 @@
+import { Types } from 'mongoose';
 import { ChangeEntityTypeEnum } from '@novu/shared';
 
+import type { EnvironmentId } from '../environment';
+import type { OrganizationId } from '../organization';
+
 export class ChangeEntity {
```
LetItRock, line 7: this type we do return from the repositories to the outside world

libs/dal/src/repositories/change/change.schema.ts
```diff
@@ -48,4 +45,5 @@ changeSchema.virtual('user', {
 });
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export const Change = mongoose.models.Change || mongoose.model<IChangeDocument>('Change', changeSchema);
+export const Change =
+  (mongoose.models.Change as mongoose.Model<ChangeDBModel>) || mongoose.model<ChangeDBModel>('Change', changeSchema);
```
LetItRock, line 49: I'm not sure why we have here "or", but I guess it's for the case when the model already is created.
So when the first part is true it means that the model is created so we cast it to the correct type.
The second part creates the model passing the correct type and schema.

libs/dal/src/repositories/execution-details/execution-details.repository.ts
```diff
@@ -1,18 +1,13 @@
 import { ExecutionDetailsStatusEnum } from '@novu/shared';
-import { ExecutionDetailsEntity } from './execution-details.entity';
-import { ExecutionDetails } from './execution-details.schema';
-import { BaseRepository, Omit } from '../base-repository';
-import { Document, FilterQuery } from 'mongoose';
-
-class PartialExecutionDetailsEntity extends Omit(ExecutionDetailsEntity, ['_environmentId', '_organizationId']) {}
 
-type EnforceEnvironmentQuery = FilterQuery<PartialExecutionDetailsEntity & Document> &
-  ({ _environmentId: string } | { _organizationId: string });
```
LetItRock, line 10: Removed these hacks to force the `_environmentId` or `_organizationId`, because it breaks all the Mongoose types.
When the query is passed to the model function, the query fields are optional, but we force them to be required and that breaks types.

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -150,21 +153,16 @@ export class JobRepository extends BaseRepository<EnforceEnvironmentQuery, JobEn
...
+  }) {
+    const job = this.MongooseModel.findOne(query, select)
       .populate('template', selectTemplate)
       .populate('notification', selectNotification)
       .populate('subscriber', selectSubscriber)
       .populate('environment', selectEnvironment)
       .lean()
       .exec();
+
+    return this.mapEntity(job) as JobEntityPopulated;
```
LetItRock, line 165: this mapping part was missing here...

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -150,21 +153,16 @@ export class JobRepository extends BaseRepository<EnforceEnvironmentQuery, JobEn
...
+  }) {
+    const job = this.MongooseModel.findOne(query, select)
       .populate('template', selectTemplate)
       .populate('notification', selectNotification)
       .populate('subscriber', selectSubscriber)
       .populate('environment', selectEnvironment)
       .lean()
       .exec();
+
+    return job as unknown as JobEntityPopulated;
```
LetItRock, line 165: here we don't do mapping, I don't know why, but I see its failing when calling `this.mapEntity(job)` and returns `{}` empty object

apps/api/src/app/events/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -32,6 +32,7 @@ import { CacheKeyPrefixEnum, InvalidateCacheService } from '../../../shared/serv
 import { SendMessageBase } from './send-message.base';
 import { ApiException } from '../../../shared/exceptions/api.exception';
 import { OrganizationEntity } from '../../../../../../../libs/dal/src/repositories/organization/organization.entity';
```
scopsy, line 34: I know not related but this import is bad, need to be refactored to be from @novu/dal, don't we have a lint rule for banned imports 🤔

apps/api/src/app/events/usecases/send-message/send-message.base.ts
scopsy: CC. @djabarovgeorge

apps/api/src/app/notification-template/usecases/update-notification-template/update-notification-template.usecase.ts
```diff
@@ -1,5 +1,6 @@
 // eslint-ignore max-len
 import { BadRequestException, Inject, Injectable, NotFoundException } from '@nestjs/common';
+import { AnyKeys } from 'mongoose';
```
scopsy, line 3: Unfortuently we should not use `mongoose` related things in API. Can we re export this thing from the dal package? Ideally we would want the application to know as less as possible about mongoose or mongo

libs/dal/src/repositories/change/change.schema.ts
scopsy: It is mainly for e2e testin for the exact reason you have mentioned 🙏

libs/dal/src/repositories/log/log.repository.ts
```diff
@@ -1,14 +1,8 @@
-import { BaseRepository, Omit } from '../base-repository';
-import { LogEntity } from './log.entity';
+import { BaseRepository } from '../base-repository';
+import { LogEntity, LogDBModel } from './log.entity';
 import { Log } from './log.schema';
-import { Document, FilterQuery } from 'mongoose';
 
-class PartialLogEntity extends Omit(LogEntity, ['_environmentId', '_organizationId']) {}
-
-type EnforceEnvironmentQuery = FilterQuery<PartialLogEntity & Document> &
```
scopsy, line 8: How the enforcement is happening right now? Maybe just missed it

libs/dal/src/repositories/job/job.repository.ts
p-fernandez: Try to replace `this.MongooseModel` for `this._model` in line 157 and see if that's the reason.

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -1,16 +1,17 @@
...
-const jobSchema = new Schema(
+const jobSchema = new Schema<JobDBModel>(
   {
     identifier: {
       type: Schema.Types.String,
     },
     status: {
       type: Schema.Types.String,
-      default: 'pending',
+      default: JobStatusEnum.PENDING,
```
p-fernandez, line 14: Good one!

libs/dal/src/repositories/organization/organization.entity.ts
```diff
@@ -17,6 +17,8 @@ export class OrganizationEntity {
   partnerConfigurations?: IPartnerConfiguration[];
 }
 
+export type OrganizationDBModel = OrganizationEntity;
```
p-fernandez, line 20: I find weird that the DBModel and the Entity are the same in this case. Shouldn't the entity return with the `OrganizationId` for the `_id`?
Also `DbModel._id` should be of type `Types.ObjectId` for coherence with the rest of the changes done.

libs/dal/src/repositories/change/change.entity.ts
LetItRock: the query enforcement was to force using `_organizationId` or `_environmentId` when querying the model... but as I commented somewhere else this strict rule breaks the types, like the Mongoose model raises TS errors because these fields should be optional

libs/dal/src/repositories/job/job.repository.ts
LetItRock: @p-fernandez that is not the reason, base repository sets the `this._model` in the constructor, the same as it does for `this.MongooseModel`. The problem might be the `JSON.stringify(entity)` call that does the `this.mapEntity`

apps/api/src/app/events/usecases/send-message/send-message-in-app.usecase.ts
LetItRock: I'll fix this one, but agree we should enable that rule.

apps/api/src/app/notification-template/usecases/update-notification-template/update-notification-template.usecase.ts
LetItRock: There is a small difference, `Partial` still requires the correct types for values to be passed, but the `AnyKeys` allows either for the correct value type or for any other type. So for this file it might work, but I've used it in another place when we do assign the `_feedId: null` but the type is `string | undefined`, so it raises the issue there, but yeah I'll fix that place too.

libs/dal/src/repositories/log/log.repository.ts
LetItRock: answered here: https://github.com/novuhq/novu/pull/2890#discussion_r1116646908

apps/api/src/app/change/usecases/get-changes/get-changes.usecase.ts
```diff
@@ -181,7 +182,7 @@ export class GetChanges {
     entityId: string,
     environmentId: string
   ): Promise<IViewEntity | Record<string, unknown>> {
-    let item = await this.layoutRepository.findOne({
+    let item: LayoutEntity | undefined = await this.layoutRepository.findOne({
```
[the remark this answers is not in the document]
LetItRock: yes, the `findOne` return type is `LayoutEntity` which is the issue I aim to solve with both PRs. 
But in the second PR these types are removed ;)

libs/dal/src/repositories/organization/organization.entity.ts
LetItRock: The `DbModel._id` type is not required by the schema, it's created automatically, so the `OrganizationEntity` just has to have it defined as the `_id: string` as it does, and because of that the `OrganizationDBModel = OrganizationEntity` for this case.

p-fernandez (commented):
Great big work again.
For me is approved, but because the query enforcement might be sensitive I'd like to have full agreement with @scopsy and @djabarovgeorge. 👍🏻

apps/api/src/app/events/usecases/send-message/send-message.base.ts
djabarovgeorge: That's true, I made him optional only because at the moment in-app does not require integration.

libs/dal/src/repositories/execution-details/execution-details.repository.ts
djabarovgeorge: We added it mainly to help us (and the community) from missing those fields. While working on the implementation I encountered a couple of those issues in the project that was missing environment fields.

LetItRock: @djabarovgeorge @scopsy @p-fernandez I think I've found the solution here will update PR with a change soon ;) so the enforcement will stay

LetItRock:
@djabarovgeorge @scopsy @p-fernandez I've found the solution here for the env or org ids enforcement so it will stay ;) please check the latest commit

djabarovgeorge:
Amazing! Looks great to me will review it again tomorrow 🙃, did you had the chance to test it locally it behaves as expected?

LetItRock:
> Amazing! Looks great to me will review it again tomorrow 🙃, did you had the chance to test it locally it behaves as expected?

yes, I've tested it and it works the exact same way ;)

libs/dal/src/types/enforce.ts
```diff
@@ -0,0 +1,5 @@
+import type { EnvironmentId } from '../repositories/environment';
+import type { OrganizationId } from '../repositories/organization';
+
+export type EnforceOrgId = { _organizationId: OrganizationId };
+export type EnforceEnvOrOrgIds = { _environmentId: EnvironmentId } | EnforceOrgId;
```
p-fernandez, line 5: Great! Only still missing thing (related to what I have pending) is that you could pass any string of any type and it will pass the validation, but for the enforcement query right now is enough. We can make it stricter in the future. 💪🏻

libs/dal/src/repositories/subscriber/subscriber.schema.ts
```diff
@@ -38,11 +39,7 @@ subscriberSchema.index({ _environmentId: 1, subscriberId: 1 });
 
 subscriberSchema.plugin(mongooseDelete, { deletedAt: true, deletedBy: true, overrideMethods: 'all' });
 
-interface ISubscriberDocument extends SubscriberEntity, Document {
-  _id: never;
-  __v: never;
-}
```
p-fernandez, line 44: 👋🏻

libs/dal/src/repositories/topic/topic-subscribers.entity.ts
```diff
@@ -18,4 +19,17 @@ export class TopicSubscribersEntity {
...
+  '_environmentId' | '_organizationId' | '_subscriberId' | '_topicId'
+> & {
+  _environmentId: Types.ObjectId;
+
+  _organizationId: Types.ObjectId;
+
+  _subscriberId: Types.ObjectId;
+
+  _topicId: Types.ObjectId;
+};
```
p-fernandez, line 33: Would be so cool if we could implement a generic that could do this automatic type (pseudocode):
```
type CustomOmit<T, <A extends ReadonlyArray<string>, V = Types.ObjectId>> = Omit<T, A> & {
  [K in (A extends ReadonlyArray<infer A> ? A : never)]: V
}
``` 
So we could do something like this:
```
type TopicSubscriberDBModel = CustomOmit<TopicSubscribersEntity, <'_environmentId' | '_organizationId' | '_subscriberId' | '_topicId', Types.ObjectId>>;
```

p-fernandez (approved):
🌟 

So for future reference the use will be:
 Mongoose -> FooEntity -> FooDBModel -> [MongoDB] -> FooDBModel -> FooEntity -> Mongoose?
 Keeping all the transformations inside of the repositories?

libs/dal/src/repositories/topic/topic-subscribers.entity.ts
LetItRock: your wish is my order :D 
```
type ChangePropsValueType<T, K extends keyof T, V = Types.ObjectId> = Omit<T, K> & {
  [P in K]: V;
};

type TopicSubscribersDBModel = ChangePropsValueType<
  TopicSubscribersEntity,
  '_environmentId' | '_organizationId' | '_subscriberId' | '_topicId'
>;
```

LetItRock:
> 🌟
> 
> So for future reference the use will be: Mongoose -> FooEntity -> FooDBModel -> [MongoDB] -> FooDBModel -> FooEntity -> Mongoose? Keeping all the transformations inside of the repositories?

exactly!

PR #2893, opened 2023-02-23
feat(dal,api): strict null check and type fixes

LetItRock:

This is the second part, the previous is #2890.
Here I've enabled the strict null check flag in the TS build config for DAL and API, resulting in many null/undefined errors in the API app.
Solved ~300 TS issues.

Some of these errors were visible in the Sentry, ex. here: https://novu-r9.sentry.io/issues/3841226005/?project=6248811&query=is%3Aunresolved&referrer=issue-stream


libs/dal/src/repositories/base-repository.ts
```diff
@@ -77,7 +77,7 @@ export class BaseRepository<T_DBModel, T_MappedEntity> {
       })
       .batchSize(batchSize)
       .cursor()) {
-      yield this.mapEntities(doc);
+      yield this.mapEntity(doc);
```
LetItRock, line 80: the doc is a single entity

apps/api/src/app/events/services/chat-service/chat.factory.ts
```diff
@@ -7,7 +7,7 @@ import { MSTeamsHandler } from './handlers/msteams.handler';
 export class ChatFactory implements IChatFactory {
   handlers: IChatHandler[] = [new SlackHandler(), new DiscordHandler(), new MSTeamsHandler()];
 
-  getHandler(integration: IntegrationEntity): IChatHandler {
+  getHandler(integration: IntegrationEntity) {
```
p-fernandez, line 10: What's the reason behind removing some of the return types?

apps/api/src/app/events/services/chat-service/handlers/slack.handler.ts
```diff
@@ -1,20 +1,15 @@
...
   constructor() {
     super('slack', ChannelTypeEnum.CHAT);
   }
 
-  buildProvider(credentials: ICredentials) {
-    const config: {
-      applicationId: string;
-      clientId: string;
-      secretKey: string;
-    } = { applicationId: credentials.applicationId, clientId: credentials.clientId, secretKey: credentials.secretKey };
```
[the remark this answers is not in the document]
p-fernandez: Good spot.

apps/api/src/app/events/services/chat-service/handlers/base.handler.ts
```diff
@@ -15,7 +15,7 @@ export abstract class BaseChatHandler implements IChatHandler {
 
   async send(chatContent: IChatOptions) {
     if (process.env.NODE_ENV === 'test') {
-      return null;
+      return {};
```
p-fernandez, line 18: I particularly do not like this. Returning an empty object I think it can cause a lot of problems and doesn't make much sense, to be honest.

apps/api/src/app/events/services/push-service/handlers/fcm.handler.ts
```diff
@@ -11,13 +11,14 @@ export class FCMHandler extends BasePushHandler {
   buildProvider(credentials: ICredentials) {
     const credentialConfig: IFcmConfig = { user: credentials.user, serviceAccount: credentials.serviceAccount };
 
-    if (!credentials.user && !credentials.serviceAccount) {
-      throw new Error('Config is not valid for fcm');
-    }
```
p-fernandez, line 16: What's the reason behind removing this check? I do not find it harmful.

apps/api/src/app/events/usecases/send-message/digest/digest.usecase.ts
```diff
@@ -61,7 +61,7 @@ export class Digest extends SendMessageType {
   private async getEvents(command: SendMessageCommand) {
     const currentJob = await this.jobRepository.findOne({ _environmentId: command.environmentId, _id: command.jobId });
 
-    if (currentJob.digest?.type === DigestTypeEnum.BACKOFF) {
+    if (currentJob && currentJob.digest?.type === DigestTypeEnum.BACKOFF) {
```
p-fernandez, line 64:
```suggestion
    if (currentJob?.digest?.type === DigestTypeEnum.BACKOFF) {
```
In case it was warning for potentially undefined.

apps/api/src/app/integrations/usecases/create-integration/create-integration.command.ts
```diff
@@ -19,5 +19,5 @@ export class CreateIntegrationCommand extends EnvironmentCommand {
   check: boolean;
 
   @IsOptional()
-  userId?: string;
+  userId: string;
```
p-fernandez, line 22: Mmmm... unless this is a mistake or if not you should remove the decorator also.

apps/api/src/app/integrations/usecases/get-webhook-support-status/get-webhook-support-status.usecase.ts
```diff
@@ -36,15 +35,15 @@ export class GetWebhookSupportStatus {
 
     return true;
   }
-  private async getIntegration(command: GetWebhookSupportStatusCommand): Promise<IntegrationEntity> {
+  private async getIntegration(command: GetWebhookSupportStatusCommand) {
```
p-fernandez, line 38: In this PR you were removing some of the return types for certain methods. What's the goal behind that? I can't find any pattern in those changes so I could understand why it is needed.

apps/api/src/app/layouts/usecases/create-default-layout-change/create-default-layout-change.usecase.ts
```diff
@@ -21,7 +21,7 @@ export class CreateDefaultLayoutChangeUseCase {
   ) {}
 
   async execute(command: CreateDefaultLayoutChangeCommand): Promise<void> {
-    let item: LayoutEntity | LayoutDto = await this.layoutRepository.findOne({
+    let item: LayoutEntity | LayoutDto | null = await this.layoutRepository.findOne({
```
p-fernandez, line 24: It doesn't feel right this typing. LayoutEntity and LayoutDto shouldn't be used in the same place even if then they are interchangeable because same structure. I know it is inherited though.

apps/api/src/app/logs/usecases/create-log/create-log.usecase.ts
```diff
@@ -7,7 +7,7 @@ export class CreateLog {
   constructor(private logRepository: LogRepository) {}
 
   async execute(command: CreateLogCommand): Promise<LogEntity> {
-    let rawData: string = null;
+    let rawData: string | null = null;
```
p-fernandez, line 10:
```suggestion
    let rawData = '';
```
This should fix it and should be ok. If we don't want to risk it we can set as:
```
let rawData;
```
and still will be ok and warning free. Just checked it.

apps/api/src/app/message-template/usecases/update-message-template/update-message-template.usecase.ts
```diff
@@ -1,3 +1,4 @@
+import { AnyKeys } from 'mongoose';
```
p-fernandez, line 1: Domain application should be agnostic of the ORM. We mentioned this in the previous PR but for a different file. Just a reminder. 🙏🏻

apps/api/src/app/notifications/activity.controller.ts
```diff
@@ -38,18 +38,18 @@ export class ActivityController {
     @UserSession() user: IJwtPayload,
     @Query() query: ActivitiesRequestDto
   ): Promise<ActivitiesResponseDto> {
-    let channelsQuery: ChannelTypeEnum[];
+    let channelsQuery: ChannelTypeEnum[] | undefined;
```
p-fernandez, line 41:
```suggestion
    let channelsQuery;
```

apps/api/src/app/notifications/activity.controller.ts
```diff
@@ -38,18 +38,18 @@ export class ActivityController {
...
 
     if (query.channels) {
       channelsQuery = Array.isArray(query.channels) ? query.channels : [query.channels];
     }
 
-    let templatesQuery: string[];
+    let templatesQuery: string[] | undefined;
```
p-fernandez, line 47:
```suggestion
    let templatesQuery;
```

apps/api/src/app/notifications/activity.controller.ts
```diff
@@ -38,18 +38,18 @@ export class ActivityController {
...
     if (query.templates) {
       templatesQuery = Array.isArray(query.templates) ? query.templates : [query.templates];
     }
 
-    let emailsQuery: string[];
+    let emailsQuery: string[] | undefined;
```
p-fernandez, line 52:
```suggestion
    let emailsQuery;
```

apps/api/src/app/shared/interceptors/shared-cache.ts
```diff
@@ -124,9 +124,9 @@ export function getCredentialsKeys() {
  */
 export function buildCachedQuery(args: unknown[]): Record<string, unknown> {
   const fromStringArray = { id: args[0], environmentId: args[1] };
-  const fromObjectArray = args.reduce((obj, item) => Object.assign(obj, item), {});
+  const fromObjectArray = args.reduce<Record<string, unknown>>((obj, item) => Object.assign(obj, item), {});
```
p-fernandez, line 127: 👏🏻

libs/dal/src/repositories/layout/layout.repository.ts
```diff
@@ -66,27 +66,27 @@ export class LayoutRepository extends BaseRepository<LayoutDBModel, LayoutEntity
     return await this.findOne({ _environmentId, _organizationId, isDefault: true });
   }
 
-  async findDeleted(id: LayoutId, environmentId: EnvironmentId): Promise<LayoutEntity | undefined> {
+  async findDeleted(id: LayoutId, environmentId: EnvironmentId): Promise<LayoutEntity | null> {
```
p-fernandez, line 69: What's the reason behind moving from undefined returns to null? Just asking as I deliberately chose in this implementation (and I would like to do so in general) to go for `undefined` rather than use `null` for the type implications (being `null` an `Object` I'd rather avoid to use it as much as possible; e.g. default values in arguments when passing `null` wouldn't be used).

providers/burst-sms/src/lib/burst-sms.provider.ts
```diff
@@ -14,8 +14,8 @@ export class BurstSmsProvider implements ISmsProvider {
 
   constructor(
     private config: {
-      apiKey: string;
-      secretKey: string;
+      apiKey?: string;
+      secretKey?: string;
```
p-fernandez, line 18: I do think we have an implementation problem if we need to make optional the keys for authenticating in the providers. At this point we should have already checked we have the values set.

apps/api/src/app/events/services/chat-service/chat.factory.ts
LetItRock: the functions are defined in the interface `IChatFactory`, so you don't have to redefine them here

apps/api/src/app/events/usecases/send-message/digest/get-digest-events-regular.usecase.ts
```diff
@@ -10,12 +13,14 @@ export class GetDigestEventsRegular extends GetDigestEvents {
...
     const amount =
-      typeof currentJob?.digest.amount === 'number'
-        ? currentJob?.digest.amount
-        : parseInt(currentJob.digest.amount, 10);
+      typeof currentJob.digest?.amount === 'number'
+        ? currentJob.digest.amount
+        : parseInt(currentJob.digest?.amount ?? '0', 10);
     const earliest = sub(new Date(currentJob.createdAt), {
-      [currentJob.digest.unit]: amount,
+      [currentJob.digest?.unit ?? DigestUnitEnum.MINUTES]: amount,
```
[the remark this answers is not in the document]
LetItRock: @p-fernandez so what's your suggestion? I can add the `// @ts-ignore` comments...
I'll revert it because it works with `NaN` and `undefined` preserving the date:
Screenshot 2023-02-24 at 10 50 59

apps/api/src/app/events/services/push-service/handlers/fcm.handler.ts
LetItRock: this code was not removed but the if condition changed a little, but basically does the same

apps/api/src/app/integrations/usecases/create-integration/create-integration.command.ts
LetItRock: the userId is always passed to this command

apps/api/src/app/integrations/usecases/get-webhook-support-status/get-webhook-support-status.usecase.ts
LetItRock: the return type is inferred, in this case it's `IntegrationEntity | null`

apps/api/src/app/layouts/usecases/create-default-layout-change/create-default-layout-change.usecase.ts
LetItRock: I don't know "why's" here, but I've fixed just the type here. We do assign different values to the `item` variable.

apps/api/src/app/logs/usecases/create-log/create-log.usecase.ts
LetItRock: I've tried to not touch the code itself, but rather the types, that's why I didn't assign an empty string to it.
If you do `let rawData;` then the type is `any` and you can assign whatever you want. The code below `JSON.stringify` assigns the string, that's why the appropriate type is `let rawData: string | null = null;`.

apps/api/src/app/notifications/activity.controller.ts
LetItRock: it will be any which is not appropriate type

apps/api/src/app/notifications/activity.controller.ts
LetItRock: any type

apps/api/src/app/notifications/activity.controller.ts
LetItRock: any type

apps/api/src/app/shared/shared.module.ts
```diff
@@ -77,7 +77,7 @@ const cacheService = {
   provide: CacheService,
   useFactory: async () => {
     return new CacheService({
-      host: process.env.REDIS_CACHE_SERVICE_HOST,
+      host: process.env.REDIS_CACHE_SERVICE_HOST ?? 'localhost',
```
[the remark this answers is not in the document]
LetItRock: yeah, I've removed that and changed the type of `host?: string`. There is a logic that checks if the `host` is set when the cache is used.

libs/dal/src/repositories/layout/layout.repository.ts
LetItRock: The types that are returned by the find model methods are like `Entity | null`. And line 75 does return null too.

providers/burst-sms/src/lib/burst-sms.provider.ts
LetItRock: This is because the interface for the `ICredentials` has these fields as optional, I didn't want to touch that as it might require even more work to fix.

apps/api/src/app/events/services/push-service/handlers/fcm.handler.ts
p-fernandez: My bad. Apologies for the mistake.

apps/api/src/app/logs/usecases/create-log/create-log.usecase.ts
p-fernandez: I truly thought TS would interpret it as `undefined` type. Wow, my bad. 😓 

I know I am insisting a bit on the issue in a different commands but not keen on using `null`. It is not directly passing to `JSON.stringify` but `JSON.stringify(null)` becomes `'null'` while `JSON.stringify(undefined)` becomes `undefined`. In this case it would be in the `command.raw` but if arrives to `rawData` we can have problems thinking that could get an empty value while it could hold null stringified.
Just wanted to highlight those risks of using `null` because they can lead to unexpected bugs. So apologies for the repetition. 🙏🏻

apps/api/src/app/events/usecases/send-message/digest/get-digest-events-regular.usecase.ts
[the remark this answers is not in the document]
p-fernandez: We should throw an exception to be honest. But it is not responsibility of this PR.
Please add the ignore comment and a TODO, please. 🙏🏻

apps/api/src/app/integrations/usecases/create-integration/create-integration.command.ts
p-fernandez: Then let's remove the `isOptional`. 💪🏻

p-fernandez (approved):
🌟 
Big big work. Thank you for that.

apps/api/src/app/logs/usecases/create-log/create-log.usecase.ts
LetItRock: I've checked DB and I don't see strings like `"null"`, but I see a lot `raw: null`, but you're right it's better to change it to undefined here, I will do that

PR #2894, opened 2023-02-24
Use Buildx when building images from forked repositories

kylegalbraith:
There is a GitHub limitation that restricts access to secrets when workflows are triggered from repositories that have been forked. This means that the Test API workflow can't build an image because it doesn't have access to the secret that would allow it to authenticate to Depot.

This PR adds the ability to check if the PR is coming from a fork, and if so, it will fall back to local Docker Buildx for the image build. If it is not coming from a fork, the image build will happen on Depot.


p-fernandez (approved):
🌟 
Thank you very much for this contribution. It will help our contributors a lot.

PR #2908, opened 2023-02-26
feat: extract filter digest and delay variables to trigger

### LINKED ISSUE #2183: [NV-946] Scheduled Delay step - add to trigger snippet

ainouzgali:

Extract variables of steps - like filters on payload, digestKey or a delay path to be returned as variables as well. And not only message content variables ( {{...}} ) . So all the variables needed in the trigger will be included for the user. 

Closes [NV-1755](https://linear.app/novu/issue/NV-1755/trigger-variables-expand-variables-extraction-from-action-steps-and)
Closes #2183


apps/web/src/pages/templates/components/TriggerSnippetTabs.tsx
```diff
@@ -61,17 +66,19 @@ novu.trigger('${identifier}', ${JSON.stringify(
...
      --header 'Authorization: ApiKey <REPLACE_WITH_API_KEY>' \\
      --header 'Content-Type: application/json' \\
      --data-raw '${JSON.stringify(
        {
          name: identifier,
-         to: { ...getSubscriberValue(subscriberVariables, (variable) => variable.value || '<REPLACE_WITH_DATA>') },
-         payload: { ...getPayloadValue(variables) },
+         to,
+         payload,
+         overrides,
```
scopsy, line 81: Can we only show the overrides object here if it contains any keys?

scopsy (approved):
Looks great! Just a small note regarding overrides

apps/web/src/pages/templates/components/TriggerSnippetTabs.tsx
ainouzgali: Because of the JSON.stringify it won't show the overrides if it is an empty object

PR #2909, opened 2023-02-26
fix: content service not iterating over push title 

ainouzgali:
…n execution detail

content service does not iterate over the push title - vars are not saved for trigger and handlebars errors in title are not shown to user on update(allows submit with errors)
also - fix push icon not shown in execution steps in activity feed


scopsy (approved):
🎉

PR #2911, opened 2023-02-27
feat(api): create store job use case

p-fernandez:

Adds a new use case to isolate part of the flow that was living directly written in the trigger event use case.

Keeps clearing the Trigger Event flow to be able to identify the different problems happening in the functionalities (mainly digest).

Document to guide through the Trigger Event flow revision and analyse the different problems and the reasoning behind them:
https://www.notion.so/novuhq/Trigger-Event-revision-8a9126af57ca4b4d9cd618fd84f7f442?pvs=4


davidsoderberg (approved):
Great refactor 🥳

PR #2915, opened 2023-02-27
feat(api): create non existing topic on the fly when adding subscribers

p-fernandez:

Allows users to create a topic on the fly from the selected `topicKey` if that key doesn't exist for that environment and organization.

To decrease friction for the users to allow them to do the topic creation and the addition of the subscribers in just one call.


apps/api/src/app/topics/use-cases/add-subscribers/add-subscribers.use-case.ts
```diff
@@ -16,19 +18,28 @@ interface ISubscriberGroups {
...
-    const topic = await this.topicRepository.findTopicByKey(
+    let topic = await this.topicRepository.findTopicByKey(
       command.topicKey,
       command.organizationId,
       command.environmentId
     );
+
     if (!topic) {
-      throw new NotFoundException(`Topic with key ${command.topicKey} not found in current environment`);
+      const createTopicCommand = CreateTopicCommand.create({
```
ainouzgali, line 35: Should we check that the command.topicKey is valid? Like not null or empty string or a certain min length?

ainouzgali (approved):
🤩

apps/api/src/app/topics/use-cases/add-subscribers/add-subscribers.use-case.ts
p-fernandez: Good points!
If `topicKey` is not sent (empty string) when doing the call the URL would be the kind of `POST /api/topics//subscribers` and the App would return a 404 Not Found as it doesn't exist. If they send a null that will be converted to a string by their client so we will be ok. Also we have the validation in the command that would throw a 400 error: https://github.com/novuhq/novu/blob/b5cf0dcb26f2e938b25a3c4799a794d7f0d5e4ea/apps/api/src/app/topics/use-cases/add-subscribers/add-subscribers.command.ts#L9

I am not too much worried about the length string and so as they have the option to rename the topic with the API.

apps/api/src/app/topics/use-cases/add-subscribers/add-subscribers.use-case.ts
```diff
@@ -16,19 +18,28 @@ interface ISubscriberGroups {
...
+        environmentId: command.environmentId,
+        key: command.topicKey,
+        // TODO: Maybe make more clear that is a provisional name
+        name: `Topic-On-The-Fly-${command.topicKey}`,
```
scopsy, line 39: My suggestion is either use `topicKey` or keeping this at null value. With a preference to null, since if the user display this name somewhere better to be empty than a strange name in my opinion. Wdyt?

p-fernandez: I wanted to allow a way to help users to differentiate quickly how a topic was created in case they created it by error not being aware of the on the fly functionality.

p-fernandez: Agreement to use the substring `Autogenerated` to notify users of the origin of the topic creation.

Issue #2727, opened 2023-02-11
[NV-1649] ✨ Polishing: Redesigning/refactoring the digest code


tsssdev:
### 📜 Description

While working on digest bug #2697  I felt the current digest code is too complex and excessive and can be simplified significantly, also performance also can be improved by refactoring and reducing the DB calls. I will create a draft PR and we can discuss in detail.

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

<sub>[NV-1649](https://linear.app/novu/issue/NV-1649/✨-polishing-redesigningrefactoring-the-digest-code)</sub>

oba2311:
@tsssdev alright if you are up for it, then don't see why not. @davidsoderberg  WDYT?

SebastianStehle:
Btw: Are you really sure that you need a distributed lock for that? I am not sure about the backoff strategy, but I assume that you only exactly one digest group per subscriberId and digest-id. So you could just work with good primary keys to ensure that you do not have duplicates.

tsssdev:
Sure @oba2311 I will submit a draft PR so it will be easy to understand what I am trying to do.. @SebastianStehle I do not think we need distributed lock as we are grouping by subscriber id, may be in very rare cases we may get dupes for that I think adding locks makes system more complicated and slows down significantly, I think we should avoid locking where ever possible. 
Also Right now I think we do not have clarity on what is expected if we get multiple digest nodes and and multiple delay and digest nodes together in same chain, as part of this we can also come up with clear expected behavior.  Also right now we are calling relation between jobs as parent child, but I relly think it is just chain of jobs, WDYT?

SebastianStehle:
I think the same, therefore I was asking. With the normal strategy, I think you can just do something like this

Pseudo Code:

UPSERT
   { "_id": "{tenantId}_{subscriberId}_{...}", "expiration": { $lt: ... }}
   { "$push": { "events": "newEvent" } }
  
CASE 1: No group exists or is not expired.
1. Will be just created

CASE 2: Group is expired but not processed yet.
1. If you have another element that is already expired, you get a primary key exception. 
2. Then you just remove the element from mongodb and you push the events to the next step in the workflow.
3. Then just retry with the insert.

oba2311:
[...]

PR #2919, opened 2023-02-28
feat(api): remove dal dependencies from workflow queue service

p-fernandez:

Extracts some logic from the Workflow Queue service that led to DAL dependency and inverts the control through dependency injection of said logic inside use cases.
Also adds some tests for the job status update callbacks.

Proper inversion control and move forward to make the Workflow Queue service domain logic agnostic, so in the future when replaced by Kakfka/Pulsar (@Cliftonz is investigating), the replacement should be easier to do.


apps/api/src/app/events/services/workflow-queue/workflow.queue.service.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  it('should a job added to the queue be updated as failed if it fails', async () => {
+    const transactionId = uuid();
+    const _environmentId = session.environment._id;
+    const _notificationId = NotificationRepository.createObjectId();
+    const _organizationId = session.organization._id;
+    const _subscriberId = subscriber._id;
+    const _templateId = NotificationTemplateRepository.createObjectId();
+    const _userId = session.user._id;
+
+    // Job with fake notification template and fake notification.
```
p-fernandez, line 132: Job created on purpose to trigger a failure and the workflow queue service `failed` event listener callback.

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
```diff
@@ -32,40 +40,45 @@ export class WorkflowQueueService {
...
   @Inject()
   private runJob: RunJob;
+  @Inject()
+  private setJobAsCompleted: SetJobAsCompleted;
+  @Inject()
+  private setJobAsFailed: SetJobAsFailed;
   private readonly queueScheduler: QueueScheduler;
   readonly DEFAULT_ATTEMPTS = 3;
 
-  constructor(private createExecutionDetails: CreateExecutionDetails, private queueNextJob: QueueNextJob) {
```
p-fernandez, line 41: Let's take advantage of Nest.js dependency injection functionality.

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
```diff
@@ -92,10 +105,47 @@ export class WorkflowQueueService {
...
+          organizationId: job.data._organizationId,
+        })
+      );
+    }
+
+    const lastWebhookFilterRetry = job.attemptsMade === this.DEFAULT_ATTEMPTS && hasToBackoff;
+
+    if (lastWebhookFilterRetry) {
+      await this.handleLastFailedWebhookFilter(job, error);
+    }
```
p-fernandez, line 136: Once understanding what's happening here and found the proper naming, this should be moved to an use case to finalise the domain logic extraction.

apps/api/src/app/events/usecases/run-job/run-job.usecase.ts
```diff
@@ -2,14 +2,15 @@ import { Injectable } from '@nestjs/common';
...
+import { SendMessage, SendMessageCommand } from '../send-message';
+
 import { ApiException } from '../../../shared/exceptions/api.exception';
 import { StorageHelperService } from '../../services/storage-helper-service/storage-helper.service';
-import { QueueNextJobCommand } from '../queue-next-job/queue-next-job.command';
-import { QueueNextJob } from '../queue-next-job/queue-next-job.usecase';
-import { SendMessageCommand } from '../send-message/send-message.command';
-import { SendMessage } from '../send-message/send-message.usecase';
-import { RunJobCommand } from './run-job.command';
-import { shouldBackoff } from '../../services/workflow-queue/workflow.queue.service';
```
p-fernandez, line 12: Probably thanks to Nest.js dependency injection this circular dependency didn't trigger off.

apps/api/src/app/events/usecases/update-job-status/update-job-status.usecase.ts
```diff
@@ -0,0 +1,13 @@
...
+import { JobRepository } from '@novu/dal';
+
+import { UpdateJobStatusCommand } from './update-job-status.command';
+
+@Injectable()
+export class UpdateJobStatus {
+  constructor(private jobRepository: JobRepository) {}
+
+  public async execute(command: UpdateJobStatusCommand): Promise<void> {
+    await this.jobRepository.updateStatus(command.organizationId, command._jobId, command.status);
```
p-fernandez, line 11: Someone would say it is ridiculous to have this small use case... I'd rather have this simple and functionalities separated (and above all, out of the workflow queue service) and be able to easily modify them as needed.

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
tsssdev: All he was trying to do is, not to fail the job when number of attempts  not reached(number of attempts 3 for jobs with webhook filters), so those jobs should skip the failure path for first and 2nd failure and on 3rd failure(last failure) it should go to regular failure path and decide to schedule next job or no, based on the flag shouldStopOnFailt.. this is also taken care very cleanly in #2727 and much simpler to understand in new code

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
```diff
@@ -1,18 +1,26 @@
...
 import {
   CreateExecutionDetailsCommand,
   DetailEnum,
 } from '../../../execution-details/usecases/create-execution-details/create-execution-details.command';
-import { EXCEPTION_MESSAGE_ON_WEBHOOK_FILTER } from '../../../shared/constants';
-import { QueueNextJobCommand } from '../../usecases/queue-next-job/queue-next-job.command';
-import { QueueNextJob } from '../../usecases/queue-next-job/queue-next-job.usecase';
-import { ConnectionOptions } from 'tls';
+
+export const WORKER_NAME = 'standard';
```
scopsy, line 23: Great addition

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
```diff
@@ -32,40 +40,45 @@ export class WorkflowQueueService {
...
-  private jobRepository: JobRepository;
+  private createExecutionDetails: CreateExecutionDetails;
+  @Inject()
+  private queueNextJob: QueueNextJob;
   @Inject()
   private runJob: RunJob;
+  @Inject()
+  private setJobAsCompleted: SetJobAsCompleted;
+  @Inject()
+  private setJobAsFailed: SetJobAsFailed;
```
scopsy, line 51: Would suggest moving all of those to the constructor with @Inject, easier to read in my opinion and follows the pattern of all our other services

apps/api/src/app/events/usecases/update-job-status/update-job-status.usecase.ts
scopsy: I think that's totally fine here.

scopsy (approved):
Looks good! Added a few minor comments

apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
p-fernandez: Doing this way we are forced to use the `forwardRef` functionality as injecting them as arguments generates a circular dependency with the Events module due `AddJob` depends on the Workflow Queue service.

Issue #2043, opened 2022-11-17
[NV-1186] 🚀 Feature: add a way to unset a subcriber's credentials


amosbastian:
### 🔖 Feature description

It would be nice if you could unset credentials for e.g. a Slack workspace

### 🎤 Why is this feature needed ?

Currently you need to do:

```
await novu.subscribers.setCredentials('subscriberId', ChatProviderIdEnum.Slack, {
  webhookUrl: undefined,
});
```

which isn't too much of a hassle, but is potentially confusing

### ✌️ How do you aim to achieve this?

Having something like `unsetCredentials`, such as

```
await novu.subscribers.unsetCredentials('subscriberId', ChatProviderIdEnum.Slack);
```

would be nice

### 🔄️ Additional Information

I'm not sure if you guys want to, or already do support subscribing to multiple Slack work spaces / Discord servers, but that could be something to keep in mind

### 👀 Have you spent some time to check if this feature request has been raised before?

* [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

* [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

*No response*

> From [SyncLinear.com](https://synclinear.com)

p-fernandez:
[https://github.com/novuhq/novu/pull/2039](https://github.com/novuhq/novu/pull/2039)

oba2311:
So @scopsy  this seems done in GH?

p-fernandez:
> So @scopsy this seems done in GH?

No, this was me linking that PR in a wrong issue. Let me fix. Apologies.

PR #2940, opened 2023-03-03
NV-2043 - add a way to unset a subcriber's credentials

gitstart:
- adds the `unsetCredentials` method to subscribers
- Updates the tests to test the functionality of the new method

Closes https://github.com/novuhq/novu/issues/2043

Currently you need to do:

```javascript
await novu.subscribers.setCredentials('subscriberId', ChatProviderIdEnum.Slack, {
  webhookUrl: undefined,
});
```
to unset credentials...


p-fernandez (approved):
As a temporary solution should be fine but unless we have a blocker in the backend we should implement a DELETE endpoint or POST action based endpoint for the unset.

PR #2942, opened 2023-03-04
[NV-1804] Fix: Allow Distributed Cache to use Redis Password

### LINKED ISSUE #2941: [NV-1804] 🐛 Bug Report: Distributed Lock System Doesn't Work With Redis Password
### 📜 Description

The distributed lock system does not support Redis passwords like the rest of the app does. This is because [getRedisUrl](https://github.com/novuhq/novu/blob/next/apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts#L12) only uses `process.env.REDIS_HOST` and `process.env.REDIS_PORT`.

### 👟 Reproduction steps

1. Set a Redis password
2. Start the API project

### 👍 Expected behavior

It should allow you to have a Redis password like the rest of the app.

```text
[ioredis] Unhandled error event: ReplyError: NOAUTH Authentication required.
```

I fixed it already, just opening an issue to reference in the PR.

<sub>[NV-1804](https://linear.app/novu/issue/NV-1804/🐛-bug-report-distributed-lock-system-doesnt-work-with-redis-password)</sub>

JoeyEamigh:

Supports Redis passwords in the distributed cache system.

Fixes #2941.

I just used the same connection object y'all were using for Bull. Figured it was probably closest to what y'all wanted.


JoeyEamigh:
I am still getting errors in my container off of `next` about `ioredis` authentication. I am going to continue investigating, since the errors now start a few minutes after the container has been running, and I haven't gotten a local reproduction yet. I may be adding to this PR assuming there is no user error going on here.

JoeyEamigh:
No longer getting the error after changing the tag on my docker image - seems to have been user error.

The pipeline that failed seems to be a docker bug since I did not touch any docker stuff in this commit. Will fast-forward and see if that helps.

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.ts
```diff
@@ -47,7 +56,7 @@ export class DistributedLockService {
     }
 
     // TODO: Implement distributed nodes (at least 3 Redis instances)
-    this.instances = [getRedisUrl()].filter((instance) => !!instance).map((url) => new Redis(url));
+    this.instances = [getRedisObject()].filter((instance) => !!instance).map((url) => new Redis(url));
```
p-fernandez, line 59:
```suggestion
    this.instances = [getRedisObject()].filter((instanceSetings) => !!instanceSettings).map((redisSettings) => new Redis(redisSettings));
```
We also need to change the filter function. A helper method that checks the existence at least of the host and the port (unless `ioredis` defaults properly when passing the settings object.

Cliftonz: @p-fernandez Correct me if I am wrong, but I believe line 15 covers this.

p-fernandez: @Cliftonz yes, I missed that. With the suggestion should be good to go.

JoeyEamigh:
Applied the suggestions - thank y'all!

p-fernandez (approved):
🌟

PR #2943, opened 2023-03-04
[NV-1803] Fix: Update `env.sh` to Properly Read Runtime Vars

### LINKED ISSUE #2939: 🐛 Bug Report: `API_ROOT_URL` not being read from .env file
### 📜 Description

Version - 0.12.0

The frontend serving app ie. FRONT_BASE_URL makes calls to API_ROOT_URL. 

Issue is, `API_ROOT_URL` variable value doesn't seem to be read from `.env` file, but rather it seems to be hardcoded to localhost.

### 👟 Reproduction steps

1. Fetch the file https://github.com/novuhq/novu/blob/next/docker/local/deployment/docker-compose.yml
2. Edit compose file and replace versions 0.11.0 with 0.12.0 (since I wanted to make an upgrade to use new featues)
3. In the same directory, have an `.env` file.
4. for variable `API_ROOT_URL`, try using a different hostname -- other than localhost
5. build image & start containers `docker-compose -f docker-compose.yml up`
6. Open `FRONT_BASE_URL` in a browser
7. The frontend serving app (FRONT_BASE_URL) should use localhost for `API_ROOT_URL`. <-- that's the bug.

### 👍 Expected behavior

Expected behavior: API_ROOT_URL value to be read from `.env` file

Actual behavior: API_ROOT_URL is being set as localhost, although the value in `.env` is set to my public URL of EC2.
image

Original discord discussion: https://discord.com/channels/895029566685462578/1081138836605775912/1081183315249336360

None

JoeyEamigh:

The `env.sh` script should loop through the `.env` file and check if there are runtime variables present in the current shell. If there are, it should write them to the `env-config.js` file.

The issue was that instead of continuing the loop if there wasn't a value in the shell, the loop would continue if there wasn't a value in `.env`, requiring the environment variables to be set at build time and defeating the whole purpose of the script.

Now it will continue only if there is no value in the shell, and no default value in `.env`.

Closes #2939.

### Other information

I also added other env vars that were used in `index.ts` but were not present in the `.env` file for it to work with this `env.sh` script.


apps/web/env.sh
```diff
@@ -5,34 +5,34 @@ rm -rf ./env-config.js
 touch ./env-config.js
 
 # Add assignment
-echo "window._env_ = {" >> ./env-config.js
+echo "window._env_ = {" >>./env-config.js
```
LetItRock, line 8: We should do exactly the same also for the `widget` app. Could you adjust that file too?

vipulsharma144:
Hey! Can you folks suggest any alternatives till the time this is merged? I am unable to use the web app using docker deployment due to this bug. It always points to localhost:3000.

JoeyEamigh:
The widget `env.sh` seemed to be different and actually working, however I did update it to be the same.

JoeyEamigh:
For @vipulsharma144 and others who would like a working build until this gets merged, I have `ghcr.io/joeyeamigh/novu/web:nv-1803` public so anyone can pull it.

scopsy (approved):

katyaterletskaya:
Hi @scopsy 😊 Do you have any plans to publish a new image containing this fix?

vishnu-kyatannawar:
> Hi @katyaterletskaya the new version will be available on mid-next week pray

@scopsy any update on new version?

katyaterletskaya:
Hi @scopsy 😊 Would it be possible to add a `0.13.0` tag to the images?

scopsy:
@katyaterletskaya yes! Was just about to publish it :P

PR #2944, opened 2023-03-04
Feat/Fix: Reduce Size of Web Docker Container

JoeyEamigh:

Static web apps, such as the React app being used by Novu do not need their `node_modules` at runtime. This allows for the ability to build in a `builder` container, then copy over only the static resources to a container that will serve the files.

Currently, `novu/web:0.12.0` is 2.42GB according to `docker image ls -a`. This change reduces the size of the container to 175MB, and keeps intact the features required for dynamic environment variable injection (upon its fix in #2943).

### Other information

I am currently running this image (`ghcr.io/joeyeamigh/novu/web:slim`) in my environments, and it is working just as well as the live one. If someone knows something I don't about a system in the project that keeps this from being possible, let me know.


scopsy:
Amazing work @JoeyEamigh! thank you. Will review this with the team and merge shortly

p-fernandez (commented):
Looks good to me, but I will let @Cliftonz to be the one approving it if he finds everything right.

PR #2951, opened 2023-03-06
fix: jobs undefined

BiswaViraj:

### DIFF
--- a/apps/web/src/pages/activities/hooks/useNotificationStatus.ts
+++ b/apps/web/src/pages/activities/hooks/useNotificationStatus.ts
@@ -5,7 +5,7 @@ export const useNotificationStatus = (notification: any): JobStatusEnum => {
   const [status, setStatus] = useState(JobStatusEnum.PENDING);
 
   useEffect(() => {
-    const jobs = notification.jobs || [];
+    const jobs = notification?.jobs || [];
     const result = jobs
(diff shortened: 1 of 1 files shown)


scopsy (approved):

scopsy:
@BiswaViraj in what cases jobs can be undefined? Should we fallback to an empty array in that case? Also find it strange that a job can have undefined value in it.

BiswaViraj:
@scopsy , Not sure how job was undefined, 
adding a fallback array as well

scopsy:
@BiswaViraj could you do a bit of code investigation to try to find potential paths this could happen in? Maybe we have some other more deep problem we need to fix here.

BiswaViraj:
@scopsy , Yup, I'm on it

apps/web/src/pages/activities/hooks/useNotificationStatus.ts
```diff
@@ -5,8 +5,9 @@ export const useNotificationStatus = (notification: any): JobStatusEnum => {
   const [status, setStatus] = useState(JobStatusEnum.PENDING);
 
   useEffect(() => {
-    const result = notification?.jobs
```
scopsy, line 8: I think this should happen in the backend, so the API actually returns an array not undefined here

PR #2954, opened 2023-03-06
Nv 1787 notifications page placeholder

LetItRock:

The Notifications page table placeholder will be shown when there are no templates created yet.
The "Try Digest Playground" tile creates the example "Digest Workflow" and currently it redirects to the Workflow Editor but will navigate to the Digest Workflow Playground later on.

Screenshot 2023-03-06 at 11 20 34


apps/web/.storybook/preview-head.html
```diff
@@ -1,3 +1,8 @@
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
 <link href="https://fonts.googleapis.com/css2?family=Lato:wght@300;400;700&display=swap" rel="stylesheet">
+<script>
+window._env_ = {
+  SKIP_PREFLIGHT_CHECK: 'true',
+};
+</script>
```
LetItRock: fix for the storybook, it was failing because this env did not exist

apps/web/cypress/plugins/index.ts
```diff
@@ -85,7 +89,7 @@ module.exports = (on, config) => {
       );
 
       let templates;
-      if (!settings?.noEnvironment) {
+      if (!settings?.noTemplates) {
```
LetItRock, line 92: the `noEnvironment` was not used... 
if `noTemplates` flag is set we won't create the templates in Cypress e2e tests...

apps/web/cypress/tests/notifications.spec.ts
```diff
@@ -24,4 +24,28 @@ describe('Notification Templates Screen', function () {
...
+    cy.intercept('**/notification-templates**').as('notificationTemplates');
+    cy.visit('/templates');
+    cy.wait('@notificationTemplates');
+
+    cy.getByTestId('no-workflow-templates-placeholder').should('be.visible');
+    cy.getByTestId('create-workflow-tile').should('not.be.disabled');
+    cy.getByTestId('create-workflow-tile').click();
+
+    cy.url().should('include', '/templates/create');
+  });
```
LetItRock, line 50: for now only two tests...
I'll add the test for the `Try Digest Workflow` tile a little later when it will point to the appropriate page

apps/web/src/design-system/cards/Cards.tsx
```diff
@@ -3,22 +3,26 @@ import { useNavigate } from 'react-router-dom';
...
   description?: React.ReactNode | string;
   navigateTo?: string;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
   imagePath?: string;
   href?: string;
+  disabled?: boolean;
+  testId?: string;
 }
 
 export function Cards({ cells }: { cells: ICardCell[] }) {
```
LetItRock, line 21: moved this component to the `design-system` folder and adjusted it a little to be more close with the designs

apps/web/src/design-system/icons/general/DigestGradient.tsx
```diff
@@ -1,19 +1,8 @@
 import React from 'react';
 
-export interface IIconProps {
-  width?: string;
-  height?: string;
-}
-/* eslint-disable */
-export function DigestGradient({ width = '30', height = '31' }: IIconProps) {
+export function DigestGradient(props: React.ComponentPropsWithoutRef<'svg'>) {
```
LetItRock, line 3: allow to override the width and height like we do in other icons

apps/web/src/pages/quick-start/steps/Quickstart.tsx
```diff
@@ -39,6 +38,8 @@ export function Quickstart() {
   );
 }
 
+const BellGradientIcon = () => <BellGradient style={{ width: '40px', height: '40px' }} />;
```
LetItRock, line 41: our icons are looking wrong when you do inspect its viewBox... that's why I had to do this hack in order to have them look the same... I'll talk to Nikolay about it

apps/web/src/design-system/table/Table.tsx
```diff
@@ -25,6 +45,7 @@ export function Table({
   pagination = false,
   loading = false,
   onRowClick,
+  noDataPlaceholder,
```
LetItRock, line 48: allow to pass the table placeholder

apps/web/src/design-system/table/Table.tsx
```diff
@@ -79,7 +100,7 @@ export function Table({
         : {}),
     } as any,
     usePagination
-  );
+  ) as UseTableProps;
```
LetItRock, line 103: adjusted types for the `useTable` hook... casted it here to the `UseTableProps` because there is no another way to do this when the hook uses other plugins, like `usePagination`

apps/web/src/pages/templates/TemplatesListPage.tsx
```diff
@@ -31,6 +51,39 @@ function NotificationList() {
     navigate(ROUTES.TEMPLATES_CREATE);
   };
 
+  const handleTryDigestClick = () => {
```
LetItRock, line 54: when the `Try Digest Workflow` button is clicked we do create an example Digest Workflow

apps/web/src/design-system/cards/Cards.tsx
```diff
@@ -48,15 +52,20 @@ function Card({ cell }: { cell: ICardCell }) {
   };
 
   return (
-    <StyledCard dark={colorScheme === 'dark'} onClick={handleOnClick}>
+    <StyledCard
+      dark={colorScheme === 'dark'}
+      onClick={cell.onClick ?? handleOnClick}
```
p-fernandez, line 57: I think this logic should be moved to inside `handleOnClick`.
```suggestion
      onClick={handleOnClick}
```
And inside of the method:
```
const handleOnClick = () => {
    if (cell.onClick) {
       onClick();
       return;
     }
     
     if (cell.navigateTo) {
       navigate(cell.navigateTo);
     }
  };
```

apps/web/src/design-system/icons/general/DigestGradient.tsx
p-fernandez: Wasn't working before?

apps/web/src/pages/templates/TemplatesListNoData.tsx
```diff
@@ -0,0 +1,76 @@
...
+              testId: 'try-digest-playground-tile',
+            },
+          ]}
+        />
+      </CardsContainer>
+    </NoDataHolder>
+  );
+};
+
+const DigestGradientIcon = () => <DigestGradient style={{ height: '40px', width: '40px' }} />;
```
p-fernandez, line 76: Shouldn't this live as a reusable component within the new FE structure?

apps/web/src/pages/templates/TemplatesListPage.tsx
p-fernandez: Not sure if this should be driven (the example workflow content) by the backend. 🤔

p-fernandez (approved):

apps/web/src/design-system/icons/general/DigestGradient.tsx
LetItRock: it was only allowing for the width/height, but not for ex. `styles`

apps/web/src/pages/templates/TemplatesListPage.tsx
LetItRock: we do the same for the in-app workflow and I think that for now, it's ok to have it like this... 
I feel like we should use what is available to accomplish the goal and not overengineer with the FE needs...

PR #2969, opened 2023-03-07
feat(regression): integrate doppler secrets with regression tests

p-fernandez:

Integrates Doppler with the regression test suite in order to incorporate the secrets for the E2E testing for the email providers.
It also creates a separate NODE_ENV `regression` to avoid clash with the current test environment set up. It also keeps building the structure to be able to loop all the email providers in the regression test suite to avoid test duplication.

In order to try the different email providers in E2E tests without any mocking we will need to provide the secrets in a safe way to the codebase. Considering Doppler doesn't have a Node.js SDK yet, I went through the path of using their API by a read-only token that just gives access to a separated Doppler project I created to store the secrets of the email providers.


apps/api/src/config/env-validator.ts
```diff
@@ -91,7 +91,19 @@ if (process.env.STORAGE_SERVICE === 'AWS' || !process.env.STORAGE_SERVICE) {
...
+if (process.env.NODE_ENV === 'regression') {
+  const noEmpty = makeValidator((dopplerToken) => {
+    if (!dopplerToken) {
+      throw new Error('Expected a non empty Doppler token');
+    }
+  });
+
+  validators.DOPPLER_TOKEN = noEmpty({
+    desc: 'Doppler read-only access token to get the secrets needed to run regression tests',
+  });
```
p-fernandez, line 103: This validator will help to remind us to add the DOPPLER_TOKEN value when in need to execute the regression tests. Should only be available in this environment purposedly created.

apps/api/src/providers/helpers/index.ts
```diff
@@ -61,3 +66,59 @@ export const createRegressionNotificationTemplate = async (session: UserSession,
...
+  const _environmentId = session.environment._id;
+  const _organizationId = session.organization._id;
+
+  const activeIntegration = await integrationRepository.findActiveIntegrationForChannel(_environmentId, channel); 
+
+  if (activeIntegration) {
+    await integrationRepository.delete({
+      _environmentId, _organizationId, _id: activeIntegration._id
+    });
+  }
```
p-fernandez, line 88: We can only have one integration per channel for Email channel. So delete before creating one.

apps/api/src/providers/provider.regression.ts
```diff
@@ -1,23 +1,55 @@
+import { ChannelTypeEnum, EmailProviderIdEnum } from '@novu/shared';
 import { UserSession } from '@novu/testing';
 import { expect } from 'chai';
 
-import { createRegressionNotificationTemplate } from './helpers';
+import { checkProviderIntegration, createProviderIntegration, createRegressionNotificationTemplate } from './helpers';
+import { getProviderSecrets } from './secrets';
 
-const providerId = 'sendgrid';
+const providers = [EmailProviderIdEnum.SendGrid];
```
p-fernandez, line 8: I will keep working with one provider only and then we can slowly add them and check that the tests don't need touch ups.

apps/api/src/providers/provider.regression.ts
```diff
@@ -1,23 +1,55 @@
...
+              credentials: { apiKey: secrets.apiKey, secretKey: secrets.secretKey },
+              active: true,
+              deleted: false,
+            });
+          });
 
-      it('should create notification template properly', async () => {
-        expect(1).to.eql(0);
+          it('should create notification template properly', async () => {
+            expect(1).to.eql(0);
```
p-fernandez, line 50: I am leaving this on purpose until more work is created as a flag that is not work done. Also I am PRing against a Draft PR as a safety measure.

apps/api/src/providers/secrets/index.ts
```diff
@@ -0,0 +1,39 @@
+import { EmailProviderIdEnum } from '@novu/shared';
+import axios from 'axios';
+
+const getDopplerSecrets = async () => {
+  const response = await axios.get(
+    `https://${process.env.DOPPLER_TOKEN}@api.doppler.com/v3/configs/config/secrets/download?format=json`
+  );
+
+  return response.data;
+};
```
p-fernandez, line 10: This returns an object with the project and secrets related to the doppler token provided. As you can see there is no mention of project names and so and it is good as a safety measure. But at the very same time one has to be careful with the token provided to not inject non desired secrets.

apps/api/src/providers/secrets/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const injectDopplerSecrets = async (): Promise<void> => {
+  const dopplerSecrets = await getDopplerSecrets();
+
+  const providerSecrets = Object.entries(dopplerSecrets).filter((dopplerSecret): boolean => {
+    return dopplerSecret[0].startsWith('REGRESSION_');
+  });
```
p-fernandez, line 17: The endpoint retrieves some extra key / value pairs with information of the project. As I decided to name all the secrets for this environment with `REGRESSION_<SECRET>` we only inject those that match the pattern.

libs/dal/src/repositories/integration/types.ts
```diff
@@ -0,0 +1,7 @@
+import { FilterQuery } from 'mongoose';
+
+import { IntegrationDBModel } from './integration.entity';
+
+import type { EnforceEnvOrOrgIds } from '../../types/enforce';
+
+export type IntegrationQuery = FilterQuery<IntegrationDBModel> & EnforceEnvOrOrgIds;
```
p-fernandez, line 7: A bit of tidying up.

libs/testing/src/user.session.ts
```diff
@@ -121,8 +121,6 @@ export class UserSession {
       await this.addOrganization();
     }
 
-    await this.fetchJWT();
```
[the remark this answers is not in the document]
scopsy: I think it is needed because after adding an organization we need to generate a new jwt token with organizationId, because at this point it won't exist.

apps/api/src/providers/secrets/index.ts
```diff
@@ -0,0 +1,39 @@
+import { EmailProviderIdEnum } from '@novu/shared';
+import axios from 'axios';
+
+const getDopplerSecrets = async () => {
```
scopsy, line 4: What do you think about the following:
Instead of fetching doppler secrets in code we can just rely on the env variables to be present. And run the doppler CLI `run` command when running the tests. It will populate all env variables automatically for the run project. In this way we benefit The ability that our code doesn't know about doppler, we can run this code with env variables from GitHub actions, ECS, Doopler and etc...

so we can later easily switch where the variables are coming from. 

https://docs.doppler.com/docs/install-cli#usage

apps/api/src/providers/secrets/index.ts
scopsy: Following my comment above, we could create a new project/service in doppler that will be called regression tests and have all the variables defined there.

p-fernandez: That's what is done already. 😄

libs/testing/src/user.session.ts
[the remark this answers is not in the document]
p-fernandez: Will add it back.

apps/api/src/providers/secrets/index.ts
p-fernandez: I was considering that possibility but the need of having to log in to run the Doppler CLI pushed me back as it would be annoying for any user having to have their own Doppler account. And for me still makes the code to be aware of Doppler existence in a tighter way. Because of this one of the reasons of going with the API is the ability of quickly revoke the Doppler access token if compromised, thus we should only have an entry vector. Compared with the login that we would have any member added to the Doppler organisation.

scopsy: Got ya, that makes sense. Thanks for the explanation 🙏

scopsy (approved):
🎉

PR #2970, opened 2023-03-07
chore(infra): upgrade to support ts 4.9 satisfies

p-fernandez:

Upgrades dependencies so the linters can support TS 4.9 `satisfies` keyword.

https://prettier.io/blog/2022/11/23/2.8.0.html


scopsy (approved):
Just need to update lock file 🎉

scopsy:
Hope the prettier change won't force us to re run prettier on all the projects :P

p-fernandez:
> Hope the prettier change won't force us to re run prettier on all the projects :P

Tried it in API and no changes pushed. Hopefully elsewhere will be the same.

PR #2976, opened 2023-03-08
chore(infra): remove octokit as unused dependency

p-fernandez:

Removes unused dependency.


scopsy (approved):
A legacy from where I took this boilerplate from XD 

It was a project called releasly previously that was sending notifications when new github version were released on npm XD

p-fernandez:
@scopsy this lasted long clinged to the monorepo!

scopsy:
The legacy will live forever!

PR #2981, opened 2023-03-08
fix(NV-1650): remove the animation in the workflow editor

### LINKED ISSUE #2728: [NV-1650] ✨ Polishing: Remove the animation in the workflow editor
### 📜 Description

Remove the animation in the workflow editor. It is super annoying and you have to wait like a second each time you go to the editor.

### 👟 Reproduction steps

_No response_

### 👍 Expected behavior

No animation

_No response_

<sub>[NV-1650](https://linear.app/novu/issue/NV-1650/✨-polishing)</sub>

adarshajit:

This PR removes the animation in the workflow editor.

This was done as a part of the polishing initiative. The changes introduced in this PR fixes #2728 

demo

### DIFF
--- a/apps/web/src/components/workflow/FlowEditor.tsx
+++ b/apps/web/src/components/workflow/FlowEditor.tsx
@@ -84,7 +84,7 @@ export function FlowEditor({
       const middle = clientWidth ? clientWidth / 2 - 100 : 0;
       const zoomView = nodes.length > 4 ? 0.75 : 1;
       const xyPos = reactFlowInstance?.project({ x: middle, y: 0 });
-      setViewport({ x: xyPos?.x ?? 0, y: xyPos?.y ?? 0, zoom: zoomView }, { duration: 800 });
+      setViewport({ x: xyPos?.x ?? 0, y: xyPos?.y ?? 0, zoom: zoomView });
     }
   }, [reactFlowInstance]);
 


p-fernandez (approved):
Looks ok. Just fix the conflict. Also thanks for sharing a gif of the fix.
CC @LetItRock seems good right?

LetItRock:
> Looks ok. Just fix the conflict. Also thanks for sharing a gif of the fix. CC @LetItRock seems good right?

@adarshajit @p-fernandez 
I'm not sure about this change, in the case of the existing template it blinks, and I do notice it.
That's all because we do reposition the flow editor after it's mounted, and the position we want to have is the top center.
The flow editor doesn't allow doing that with props, unfortunately.

p-fernandez:
@adarshajit might to check what @LetItRock mentions when editing an existing workflow? 🙏🏻

LetItRock:
@adarshajit @p-fernandez 
This is what I meant, it's a small blink when the editor is initialized, it would be nice if we can remove it, but if not let's leave it. My last comment says why it's happening.

https://user-images.githubusercontent.com/2607232/225120086-9eeb70c3-a643-4d0e-b8bd-debf507426b4.mov

p-fernandez:
@scopsy I am not very keen in doing a change that breaks the UX for the Workflow edition. I'd rather keep the current status behaviour rather than patch it. Your call. 🙂

scopsy:
@adarshajit I do aggree that the flicker looks worse than the previously animation :( 

Maybe we can hide the nodes until the first reposition happens to avoid this flicker?

LetItRock:
> @adarshajit I do aggree that the flicker looks worse than the previously animation :(
> 
> Maybe we can hide the nodes until the first reposition happens to avoid this flicker?

the flick is a result because of these steps happen:
- we do always render a trigger node even if the workflow is still fetching
- when the workflow is fetched we do re-render the tree

scopsy:
@LetItRock So seems like we can hide the trigger node until the template is fetched right? I'm totally okey with having it empty until all data is available to perform a proper first render.

LetItRock:
> @LetItRock So seems like we can hide the trigger node until the template is fetched right? I'm totally okey with having it empty until all data is available to perform a proper first render.

@scopsy yes, that's correct

scopsy:
Sounds like a plan, @adarshajit want to give this a try?

jainpawan21:
As we have changed this UI in 0.14.0
I think this issue is now fixed now automatically.

scopsy:
@jainpawan21 good point. Closing this 🙏

PR #2979, opened 2023-03-08
feat: Add fcmOptions to Firebase provider

tomcastro:
### What change does this PR introduce?

This PR introduces the `fcmOptions` (link [here](https://firebase.google.com/docs/reference/admin/node/firebase-admin.messaging.fcmoptions.md#fcmoptions_interface)) optional override for Firebase, as this allows us to implement analytics labels for Firebase tracking. 

### Why was this change needed?

Closes #2965

Issue #2965, opened 2023-03-07
🚀 Feature: add analytics_label option in FCM


jainpawan21:
### 🔖 Feature description

Novu Push channel has FCM provider. One option used for analytics `analytics_label` is missing in FCM options.

### 🎤 Why is this feature needed ?

analytics_label option can be used for analytics purpose.

### ✌️ How do you aim to achieve this?

By adding extra field in either FCM data or overrides.

### 🔄️ Additional Information

https://discord.com/channels/895029566685462578/943775848509620234/1082670692832641044

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

oba2311:
@jainpawan21  the original poster indicated that they are willing to submit the PR - could you share the link to the issue with them and assign?  thanks

PR #2982, opened 2023-03-09
FCM support add extra options

### LINKED ISSUE #2233: 🚀/🐛 Feature/Bug: No possibility of sending FCM data payload
### 🔖 Feature description

Dear all

With the commit `8c30377ddd806d4c06d51ba4f9ee6541ab593e38` there was the introduction of the new FCM API, where novu uses now `sendMulticast`.

I'd like to send some payload data to handle clicks on push notifications, i.e. I want to use the `data` property for a received push notification. 

In `fcm.provider.ts`, there's this:

```typescript
  async sendMessage(
    options: IPushOptions
    delete (options.overrides as any)?.deviceTokens;
    const res = await this.messaging.sendMulticast({
      tokens: options.target,
      notification: {
        title: options.title,
        body: options.content,
        ...options.overrides,
      },
    });
```

So, Novu only uses the `notification` property of the `MulticastMessage`, which is an extension of the a [BaseMessage](https://firebase.google.com/docs/reference/admin/node/firebase-admin.messaging.basemessage.md#basemessage_interface). The `BaseMessage` also allows for a `data` property, which then can be accessed when the notification is received. 

Be aware, that just using `payload` as the `data` property might be an issue, since payload can be of any type, e.g. `payload : { someList : [] }` would fail, because the `data` property must be `{ [key: string]: string; }` according to the docs.

Now, I would do and push this feature myself but sadly I'm not experienced enough with JS to build the application... 

Thanks!

To act according to the received data when clicking on a received FCM push notification, e.g. show a different screen than usual etc.

See above.

### 🔄️ Additional Information

_No response_

None

tienthanh2509:

Support adding additional options for FCM message

- ~FCM option with analytics_label~ It already done on #2979
- Priority for android message
- `apns` with additional headers include priority header, Ref https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns
- Add test case for above changes

Closes #2233, #2965

- Update IPushOptions for new extra fields
- ~add fcmOptions~ It already done on #2979

### DIFF
--- a/packages/stateless/src/lib/provider/provider.interface.ts
+++ b/packages/stateless/src/lib/provider/provider.interface.ts
@@ -56,8 +56,13 @@ export interface IPushOptions {
     channelId?: string;
     categoryId?: string;
     mutableContent?: boolean;
-    android?: { [key: string]: { [key: string]: string } };
-    apns?: { payload: { aps: { [key: string]: { [key: string]: string } } } };
+    android?: { [key: string]: { [key: string]: string } | string };
+    apns?: {
+      headers?: { [key: string]: string };
+      payload: {
+        aps: { [key: string]: { [key: string]: string } | string };
+      };
+    };
     fcmOptions?: { analyticsLabel?: string };
   };
 }
--- a/providers/fcm/src/lib/fcm.provider.spec.ts
+++ b/providers/fcm/src/lib/fcm.provider.spec.ts
@@ -167,3 +167,131 @@ test('should trigger fcm with apns (ios) override', async () => {
     },
   });
 });
+
+test('should trigger fcm data for ios with headers options', async () => {
+  await provider.sendMessage({
+    title: 'Test',
+    content: 'Test push',
+    target: ['tester'],
+    payload: {
+      key_1: 'val_1',
+      key_2: 'val_2',
+    },
+    overrides: {
+      type: 'data',
+      apns: {
+        headers: {
+          'apns-priority': '5',
+        },
+        payload: {
+          aps: {
+            alert: {
+              'loc-key': 'some_body',
+              'title-loc-key': 'some_title',
+            },
+            sound: 'demo.wav',
+          },
+        },
+      },
+    },
+  });
+  expect(app.initializeApp).toHaveBeenCalledTimes(1);
+  expect(app.cert).toHaveBeenCalledTimes(1);
+  expect(spy).toHaveBeenCalled();
+  expect(spy).toHaveBeenCalledWith({
+    tokens: ['tester'],
+    apns: {
+      headers: {
+        'apns-priority': '5',
+      },
+      payload: {
+        aps: {
+          alert: {
+            'loc-key': 'some_body',
+            'title-loc-key': 'some_title',
+          },
+          sound: 'demo.wav',
+        },
+      },
+    },
+    data: {
+      key_1: 'val_1',
+      key_2: 'val_2',
+    },
+  });
+});
+
+test('should trigger fcm data for both android and ios with override options', async () => {
+  await provider.sendMessage({
+    title: 'Test',
+    content: 'Test push',
+    target: ['tester'],
+    payload: {
+      key_1: 'val_1',
+      key_2: 'val_2',
+    },
+    overrides: {
+      type: 'data',
+      android: {
+        data: {
+          body_loc_key: 'body_loc_value',
+          title_loc_key: 'title_loc_value',
+          sound: 'demo.wav',
+        },
+        priority: 'high',
+      },
+      apns: {
+        headers: {
+          'apns-priority': '5',
+        },
+        payload: {
+          aps: {
+            alert: {
+              'loc-key': 'some_body',
+              'title-loc-key': 'some_title',
+            },
+            sound: 'demo.wav',
+          },
+        },
+      },
+      fcmOptions: {
+        analyticsLabel: 'test_label',
+      },
+    },
+  });
+  expect(app.initializeApp).toHaveBeenCalledTimes(1);
+  expect(app.cert).toHaveBeenCalledTimes(1);
+  expect(spy).toHaveBeenCalled();
+  expect(spy).toHaveBeenCalledWith({
+    tokens: ['tester'],
+    android: {
+      data: {
+        body_loc_key: 'body_loc_value',
+        title_loc_key: 'title_loc_value',
+        sound: 'demo.wav',
+      },
+      priority: 'high',
+    },
+    apns: {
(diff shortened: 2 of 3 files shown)


p-fernandez (approved):
🌟

p-fernandez:
@tienthanh2509 I just noticed the issue you are trying to close was assigned to a different contributor. The PR was added here: https://github.com/novuhq/novu/pull/2979
But I noticed you added a small thing extra, the APNS headers. I will try to merge the other PR so you can modify this PR from `next` branch and just add that change.

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -56,8 +56,12 @@ export interface IPushOptions {
     channelId?: string;
     categoryId?: string;
     mutableContent?: boolean;
-    android?: { [key: string]: { [key: string]: string } };
-    apns?: { payload: { aps: { [key: string]: { [key: string]: string } } } };
+    fcmOptions?: { [key: string]: string };
+    android?: object;
+    apns?: {
+      headers?: { [key: string]: string };
```
p-fernandez, line 62: Update the PR to provide just this change. And please also update the tests or add a new one to check this new property is passed properly.

p-fernandez (changes requested):

tienthanh2509:
Hi @p-fernandez 
I updated the interface, the extra option is now under `fcm` fields.'
Please help to review
Thanks

```http
POST /api/v1/events/trigger HTTP/1.1
Content-Type: application/json
Authorization: ApiKey redacted
Host: 192.168.128.62:3080
Content-Length: 792

{
	"name": "some-template",
	"to": {
		"subscriberId": "redacted"
	},
	"payload": {
		"action_type": "some data"
	},
	"overrides": {
		"fcm": {
			"type": "data",
			"android": {
				"data": {
					...
				},
				"priority": "high"
			},
			"apns": {
				"headers": {
					"apns-priority": "5"
				},
				"payload": {
					...
				}
			}
		}
	}
}
```

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -56,6 +56,7 @@ export interface IPushOptions {
     channelId?: string;
     categoryId?: string;
     mutableContent?: boolean;
+    fcmOptions?: { [key: string]: string };
     android?: { [key: string]: { [key: string]: string } };
     apns?: { payload: { aps: { [key: string]: { [key: string]: string } } } };
```
p-fernandez, line 61: In this interface you need to add to the `apns` property the `headers` type.
So far you can leave it like:
```
    apns?: { headers: { [key: string]: string }, payload: { aps: { [key: string]: { [key: string]: string } } } };
```

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -56,6 +56,7 @@ export interface IPushOptions {
     channelId?: string;
     categoryId?: string;
     mutableContent?: boolean;
+    fcmOptions?: { [key: string]: string };
```
p-fernandez, line 59: You are duplicating the property in the interface (line 61) and removing its functionality. When merging, please don't remove `{ analyticsLabel?: string };` and include the generic record.

providers/fcm/src/lib/fcm.provider.spec.ts
```diff
@@ -167,3 +167,78 @@ test('should trigger fcm with apns (ios) override', async () => {
     },
   });
 });
+
+test('should trigger fcm data for both android and ios with override options', async () => {
```
p-fernandez, line 171: We have already an ios test so leave it just for Android and pass only the Android properties, not the APNS ones, please.
In the iOS test, include the APNS `headers` part of the payload to complete the test the other contributor created. 👍🏻

p-fernandez (changes requested):

packages/stateless/src/lib/provider/provider.interface.ts
tienthanh2509: Hi @p-fernandez 
I clean up the PR and added headers to apns

packages/stateless/src/lib/provider/provider.interface.ts
tienthanh2509: Hi @p-fernandez 
I clean up the PR, removed duplicated changes

providers/fcm/src/lib/fcm.provider.spec.ts
tienthanh2509: Hi @p-fernandez 
Test cases were added, since almost change duplicate #2979 , this PR only contains small change for missing headers (apns) and priority fields (android)

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -56,8 +56,13 @@ export interface IPushOptions {
     channelId?: string;
     categoryId?: string;
     mutableContent?: boolean;
-    android?: { [key: string]: { [key: string]: string } };
-    apns?: { payload: { aps: { [key: string]: { [key: string]: string } } } };
+    android?: { [key: string]: { [key: string]: string } | string };
+    apns?: {
+      headers?: { [key: string]: string };
```
tienthanh2509, line 61: This is new field, required for change apns headers

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -56,8 +56,13 @@ export interface IPushOptions {
     channelId?: string;
     categoryId?: string;
     mutableContent?: boolean;
-    android?: { [key: string]: { [key: string]: string } };
-    apns?: { payload: { aps: { [key: string]: { [key: string]: string } } } };
+    android?: { [key: string]: { [key: string]: string } | string };
```
tienthanh2509, line 59: Android FCM message contains both map{} and string fields

tienthanh2509 (commented):
The PR was cleaned up due to duplicate #2979 
Only additional fields was added

providers/fcm/src/lib/fcm.provider.spec.ts
p-fernandez: What I meant was to separate the iOS and Android testing in separate tests.
```
test('should trigger fcm data for android with override options',
```
and remove from the payload the `apns` object, as we already test it in the test of line 171.

tienthanh2509: The test case was separated for android and ios

p-fernandez (approved):
🌟

PR #2985, opened 2023-03-09
fix: only throw error if response contains any failed data

### LINKED ISSUE #2966: [NV-1817] 🐛 Bug Report: APNS delivered notifications marked as failed (v 0.12.0)
### 📜 Description

We discovered something interesting while testing APNS integration for sending ios push notifications.

When we use a valid token, the notification is being received BUT we get an error (with an empty message) and the transaction is marked as failed.

When we use a wrong token then we get an error with the correct message and the transaction is also being marked as failed.

So, in both cases the transaction is marked as failed but in the first case we actually saw it coming in.

### 👟 Reproduction steps

1. Setup APNS provider
2. Create subscriber and notification template
3. Send ios PUSH notification to the subscriber's device token

### 👍 Expected behavior

Message should be received and transaction should be marked as successful. 

Message is being received but transaction is marked as failed with an empty error.

```
{
  “stack”: “Error: \n    at APNSPushProvider.<anonymous> (/usr/src/app/providers/apns/build/main/lib/apns.provider.js:84:35)\n    at step (/usr/src/app/providers/apns/build/main/lib/apns.provider.js:44:23)\n    at Object.next (/usr/src/app/providers/apns/build/main/lib/apns.provider.js:25:53)\n    at fulfilled (/usr/src/app/providers/apns/build/main/lib/apns.provider.js:16:58)\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)“,
  “message”: “”
}
```
Screenshot 2023-03-07 at 5 33 25 PM

_No response_

None

<sub>[NV-1817](https://linear.app/novu/issue/NV-1817/🐛-bug-report-apns-delivered-notifications-marked-as-failed-v-0120)</sub>

BiswaViraj:
Closes #2966 

Throws error only when the notification fails, and should be able to send success response on notification being delivered successfully
Notifications sent successfully were being marked as failed.

### DIFF
--- a/providers/apns/src/lib/apns.provider.ts
+++ b/providers/apns/src/lib/apns.provider.ts
@@ -44,7 +44,7 @@ export class APNSPushProvider implements IPushProvider {
     });
     const res = await this.provider.send(notification, options.target);
 
-    if (res.failed) {
+    if (res.failed.length > 0) {
       throw new Error(
         res.failed
           .map(


providers/apns/src/lib/apns.provider.ts
```diff
@@ -44,7 +44,7 @@ export class APNSPushProvider implements IPushProvider {
     });
     const res = await this.provider.send(notification, options.target);
 
-    if (res.failed) {
+    if (res.failed.length > 0) {
```
p-fernandez, line 47: Reference for this change:
https://github.com/node-apn/node-apn/blob/master/doc/provider.markdown#failed

p-fernandez (approved):
🌟

PR #2987, opened 2023-03-09
fix(infra): when upgrading to pnpm 7.28.0 we didn't update the images

p-fernandez:

Homogenise the PNPM version all throughout the monorepo.

PNPM it was upgraded to 7.28.0 (https://github.com/novuhq/novu/blob/56ddf76a915aba4bb26c70bfef47ef6984b08f98/package.json#L4) but not in the Docker images and some scripts. This fixes that.
Also it is an investigation step to see why this CI check fails for a dependency update:
https://github.com/novuhq/novu/actions/runs/4373578609/jobs/7651925951


BiswaViraj (approved):
🚀

PR #2995, opened 2023-03-10
feat(api): clarify the notification steps creation

p-fernandez:

Small refactor to aggregate the logic accordingly.

Inside of the use case that was in charge on deciding if a digest step should be created or not we were doing more things, like filtering the active steps for the notification, adding extra complexity to that use case.
These changes should facilitate the vision if we are committing to decisions on creating or not digest steps before we should.

### DIFF
--- a/apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
+++ b/apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
@@ -1,12 +1,14 @@
 import { Injectable, Logger } from '@nestjs/common';
-import { MessageTemplateEntity, NotificationStepEntity } from '@novu/dal';
-import { StepTypeEnum, DigestTypeEnum } from '@novu/shared';
+import { NotificationStepEntity } from '@novu/dal';
+import { DigestTypeEnum, StepTypeEnum } from '@novu/shared';
+
 import { DigestFilterStepsCommand } from './digest-filter-steps.command';
 import { DigestFilterStepsBackoff } from './digest-filter-steps-backoff.usecase';
 import { DigestFilterStepsRegular } from './digest-filter-steps-regular.usecase';
 
 import { EventsPerformanceService } from '../../services/performance-service';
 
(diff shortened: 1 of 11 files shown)


apps/api/src/app/message-template/usecases/update-message-template/update-message-template.usecase.ts
```diff
@@ -110,23 +110,25 @@ export class UpdateMessageTemplate {
       );
     }
 
-    const changeId = await this.changeRepository.getChangeId(
-      command.environmentId,
-      ChangeEntityTypeEnum.MESSAGE_TEMPLATE,
-      item._id
-    );
+    if (item._id) {
```
p-fernandez, line 113: Change to comply with setting _id as optional for MessageTemplateEntity.

apps/api/src/app/message-template/usecases/create-message-template/create-message-template.usecase.ts
```diff
@@ -37,7 +37,10 @@ export class CreateMessageTemplate {
       actor: command.actor,
     });
 
-    item = (await this.messageTemplateRepository.findById(item._id)) as MessageTemplateEntity;
+    if (item?._id) {
```
p-fernandez, line 40: Change to comply with setting _id as optional for MessageTemplateEntity.

LetItRock (approved):
looks good to me 🥇

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -23,44 +25,35 @@ export class DigestFilterSteps {
...
     return {
       template: {
         _environmentId: command.environmentId,
         _organizationId: command.organizationId,
         _creatorId: command.userId,
+        _layoutId: null,
         type: StepTypeEnum.TRIGGER,
         content: '',
-      } as MessageTemplateEntity,
+      },
```
djabarovgeorge, line 56: Not relevant to this PR but are you aware of why _layoutId is nullable and not optional?

PR #2996, opened 2023-03-10
feat: set inherit as deault font family for the notification center t…

BiswaViraj:
…o be able to inherit the host font family

- sets the font-family of notification-center to `inherit` instead of `Lato`, this will allow inApp to inherit the host's font family.
- Added `inherit` option to the branding form


packages/notification-center/src/components/notification-center/components/AppContent.tsx
```diff
@@ -29,7 +29,7 @@ export function AppContent() {
 
 const wrapperClassName = (primaryColor: string, fontFamily: string, dir: string) => css`
   margin: 0;
-  font-family: ${fontFamily}, Helvetica, sans-serif;
```
p-fernandez, line 32: Not sure if it is a good thing to remove the fallback font families. If it doesn't stop for our users to add the font family they chose if it can be loaded, I would leave the fallbacks.

libs/shared/src/consts/notification-item-buttons/notificationItemButton.ts
```diff
@@ -33,21 +33,21 @@ export const darkButtonStyle: IStyleButtons = {
...
+  clicked: { backGroundColor: 'white', fontColor: '#FFFFFF', removeCircleColor: '#525266', fontFamily: 'inherit' },
 };
 
 export const lightButtonStyle: IStyleButtons = {
   primary: {
     backGroundColor: 'linear-gradient(99deg,#DD2476 0% 0%, #FF512F 100% 100%)',
     fontColor: '#FFFFFF',
     removeCircleColor: 'white',
-    fontFamily: 'Lato',
+    fontFamily: 'inherit',
```
LetItRock, line 47: font-family is inherited css prop, I don't think that you have to specify everywhere `inherit`.

packages/notification-center/src/components/notification-center/components/AppContent.tsx
BiswaViraj: added the fallbacks 🚀

libs/shared/src/consts/notification-item-buttons/notificationItemButton.ts
BiswaViraj: removed them and made the `fontFamily`optional type

p-fernandez (approved):

PR #3000, opened 2023-03-11
feat(otel/analytics): Add Open-Telemetry and Promethus Montioring to Novu

Cliftonz:
This PR adds the initial commit of nestjs-otel to Novu's api, ws, and webhook containers.
This system set up a prometheus metric server on 9464 on the path '/metrics' and the ability to add spans in the future.

The otel output is supported by NewRelic and will allow logs to be connected with spans and traced for performance on the open source (with jaeger and prometheus) and saas deployments as discussed in #2878.

image


apps/api/src/app.module.ts
```diff
@@ -58,6 +59,7 @@ const modules: Array<Type | DynamicModule | Promise<DynamicModule> | ForwardRefe
   MessagesModule,
   PartnerIntegrationsModule,
   TopicsModule,
+  getOTELModule(),
```
LetItRock, line 62: it would be better if you could just return the telemetry module instead of returning the getter function that then you call like we have other modules above here

apps/api/src/bootstrap.ts
```diff
@@ -19,6 +19,8 @@ import { ResponseInterceptor } from './app/shared/framework/response.interceptor
 import { RolesGuard } from './app/auth/framework/roles.guard';
 import { SubscriberRouteGuard } from './app/auth/framework/subscriber-route.guard';
 import { validateEnv } from './config/env-validator';
+import { getOTELSDK } from '@novu/application-generic/build/main/tracing/tracing';
+import * as packageJson from '../package.json';
```
LetItRock, line 23: we had some issues with this pattern before, so maybe let's remove it and hardcode the name in the `getOTELSDK` function call?

apps/webhook/src/bootstrap.ts
```diff
@@ -2,19 +2,24 @@ import './config';
...
+import * as packageJson from '../package.json';
 
 import { AppModule } from './app.module';
 
 if (process.env.SENTRY_DSN) {
   Sentry.init({
     dsn: process.env.SENTRY_DSN,
     environment: process.env.NODE_ENV,
-    release: `v${version}`,
+    release: `v${packageJson.version}`,
```
LetItRock, line 14: we should also pass the version from env variables

apps/api/src/bootstrap.ts
Cliftonz: Can you point me to thous issues?

I am doing the same thing with the logging PR.

p-fernandez: https://discord.com/channels/895029566685462578/1045289346598703164/1073211111874248754

packages/application-generic/src/tracing/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  MetricService,
+  OtelInstanceCounter,
+  OtelUpDownCounter,
+  OtelHistogram,
+  OtelObservableGauge,
+  OtelObservableCounter,
+  OtelObservableUpDownCounter,
+  OtelCounter,
+  Span,
+} from 'nestjs-otel';
```
p-fernandez, line 15: I do like to have a single point of export for all the functionalities needed so there the swapping of the third party is quick. Might suggest to have a generic wrapper name that surrounds the functionality such as:
```
setCounter,
setHistogram,
setGauge,
```

apps/webhook/src/bootstrap.ts
Cliftonz: An environment variable can easily get forgotten or not updated, however, the package.json is the source of truth which is why I used it.

apps/api/src/bootstrap.ts
scopsy: It was actually another thing that causing this previously. So I believe it should be a problem id the API is building succesfully

apps/webhook/src/bootstrap.ts
LetItRock: true, but I just mentioned that because we have had issues before with this approach and we wanted to update it somewhere later on

scopsy:
@Cliftonz closing this for the time being 🙏

jahnavisana2812:
Can i know the reason why was this pr closed

Cliftonz:
@jahnavisana2812 We closed this as this was a POC and we are unsure how to correctly move forward at the moment on metrics and analytics gathering.

We have it on the list to discuss in the near future.

PR #3005, opened 2023-03-12
Get Started - Create integration providers section

djabarovgeorge:

Adds integration configuration component.
 
image

image


apps/web/src/pages/quick-start/consts.tsx
```diff
@@ -234,3 +238,59 @@ export enum FlowTypeEnum {
...
+      }
+    },
+  },
+  {
+    Icon: Mail,
+    title: 'Email',
+    displayName: 'Email',
+    type: ChannelTypeEnum.EMAIL,
+    description: '🎉  Try our gift: 300 emails Use Novu provider for free or change the provider to yours',
+    clickHandler: () => {},
```
LetItRock, line 261: this way it might be hard to integrate with a integrations store modal, like from here you are not able to say open modal

LetItRock (approved):

apps/web/src/pages/quick-start/consts.tsx
djabarovgeorge: From here we could say open modal if this handler will get 'setter' from the upper scope (like I did on another handler with 'navigation'), do you think it will work?

PR #3009, opened 2023-03-13
feat(regression): add mailtrap integration

p-fernandez:

Integration with Mailtrap.io (tool to be able to inspect and test emails sent).

It is the tool we will use to check the emails sent to the different integrations are right.

### DIFF
--- a/apps/api/src/providers/provider.regression.ts
+++ b/apps/api/src/providers/provider.regression.ts
@@ -1,18 +1,119 @@
 import { ChannelTypeEnum, EmailProviderIdEnum } from '@novu/shared';
-import { UserSession } from '@novu/testing';
+import { MailtrapInboxStatusEnum, MailtrapService, UserSession } from '@novu/testing';
 import { expect } from 'chai';
 
 import { checkProviderIntegration, createProviderIntegration, createRegressionNotificationTemplate } from './helpers';
-import { getProviderSecrets } from './secrets';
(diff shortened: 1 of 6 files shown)


libs/testing/src/notifications.service.ts
```diff
@@ -1,10 +1,6 @@
 import axios from 'axios';
-import { ChannelTypeEnum } from '@novu/shared';
-import { MessageRepository } from '@novu/dal';
```
p-fernandez, line 3: Leftovers

apps/api/src/providers/provider.regression.ts
```diff
@@ -1,18 +1,119 @@
...
+      expect(inbox).to.deep.include({
+        id: inboxId,
+        name: 'Regression',
+        status: MailtrapInboxStatusEnum.ACTIVE,
+        used: true,
+        domain: 'sandbox.smtp.mailtrap.io',
+        email_domain: 'inbox.mailtrap.io',
+        emails_count: 0,
+        emails_unread_count: 0,
+      });
```
LetItRock, line 57: I don't get it, do we test here the Mailtrap APIs and their responses?

LetItRock: yes, but there is no logic in that layer :P

LetItRock (approved):

PR #3021, opened 2023-03-15
feat(api): endpoint to check if subscriber belongs to a topic

p-fernandez:

Implement and endpoint that provided a topicKey and the external subscriber ID retrieves the relationship if existing or a 404 Not Found if any of both are not existing.
Implement it in Novu package.
Add tests and documentation.

We have been requested by users to provide with a way for them to check if a subscriber belongs to a certain topic. This could improve the engagement with the feature and traction of potential new users that could implement their use cases and use Novu.

### DIFF
--- a/apps/api/src/app/topics/dtos/topic-subscriber.dto.ts
+++ b/apps/api/src/app/topics/dtos/topic-subscriber.dto.ts
@@ -2,7 +2,7 @@ import { ApiProperty } from '@nestjs/swagger';
 
 import { EnvironmentId, ExternalSubscriberId, OrganizationId, SubscriberId, TopicId, TopicKey } from '../types';
 
-export class TopicSubscribersDto {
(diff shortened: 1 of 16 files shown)


apps/api/src/app/topics/dtos/topic-subscriber.dto.ts
```diff
@@ -2,7 +2,7 @@ import { ApiProperty } from '@nestjs/swagger';
 
 import { EnvironmentId, ExternalSubscriberId, OrganizationId, SubscriberId, TopicId, TopicKey } from '../types';
 
-export class TopicSubscribersDto {
+export class TopicSubscriberDto {
```
p-fernandez, line 5: For the DTO the singular made more sense.

LetItRock (approved):
awesome, we're extending topics' functionality🥇

PR #3029, opened 2023-03-15
fix(api): reduce activity stats queries and improve performance

p-fernandez:

Reduce the number of queries against the DB for the Activity Stats endpoint (it calculates 3 numbers, yearly total notifications, monthly and weekly).
Also introduces an index for Notification schema for the `_organizationId` and a test.

We were being notified by performance degradation. Slow responses in the Activity feed. Thanks to one of our users we could spot one of the slow endpoints was `/notifications/stats`. Investigating the code we can find that 3 count queries were done. It is acknowledged that MongoDB count queries for large collections can be quite slow. This tries to mitigate that for large collections reducing the queries and calculating the numbers in an alternative way.


apps/api/src/app/notifications/dtos/activities-response.dto.ts
```diff
@@ -173,7 +173,7 @@ export class ActivityNotificationResponseDto {
   transactionId: string;
 
   @ApiPropertyOptional()
-  createdAt?: Date;
+  createdAt?: string;
```
p-fernandez, line 176: I think this doesn't break anything as I think we were truly returning strings (this DTO is used in API responses and I couldn't find any transformation to Date in the code). Worth review this in the Web app. Running in my local couldn't find any problem with this change though. 😕

apps/api/src/app/notifications/e2e/get-activity-stats.e2e.ts
```diff
@@ -43,7 +53,7 @@ describe('Get activity stats - /notifications/stats (GET)', async () => {
       {
         _environmentId: session.environment._id,
       },
-      null,
+      undefined,
```
p-fernandez, line 56: TS warning.

apps/api/src/app/notifications/e2e/get-activity-stats.e2e.ts
```diff
@@ -69,4 +79,107 @@ describe('Get activity stats - /notifications/stats (GET)', async () => {
...
+    const channels = [StepTypeEnum.IN_APP];
+    const to = 'no-reply@novu.co';
+    const payload = {};
+
+    const notifications: unknown[] = [];
+
+    /*
+     * We generate notifications avoiding clashes of leap years and different month lengths
+     * so this test can be executed any time with same results
+     * Create 7 notifications during the week
```
p-fernandez, line 97: I just wanted to avoid dates edge cases based on the execution time.

apps/api/src/app/notifications/usecases/get-activity-stats/get-activity-stats.usecase.ts
```diff
@@ -1,42 +1,61 @@
...
-      },
-    });
+    const now = new Date();
 
-    const monthly = await this.notificationRepository.count({
-      _environmentId: command.environmentId,
-      _organizationId: command.organizationId,
-      createdAt: {
-        $gte: subMonths(new Date(), 1),
+    const yearlyNotifications = await this.notificationRepository.find(
```
p-fernandez, line 16: We retrieve all the notifications in a year and then loop them to calculate the monthly and the weekly rather than hitting the DB again. Also to help for large collections we just retrieve 2 fields: `_id` and `createdAt`. We could even just retrieve `createdAt`.

libs/dal/src/repositories/base-repository.ts
```diff
@@ -88,6 +88,20 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
...
+    const result = await this.MongooseModel.insertMany(data, { ordered: false });
+
+    const insertedIds = result.map((inserted) => inserted._id);
+
+    return {
+      acknowledged: true,
+      insertedCount: result.length,
+      insertedIds,
+    };
+  }
```
p-fernandez, line 103: Introduced to help with the testing.

libs/dal/src/repositories/notification/notification.schema.ts
```diff
@@ -19,6 +19,7 @@ const notificationSchema = new Schema<NotificationDBModel>(
     _organizationId: {
       type: Schema.Types.ObjectId,
       ref: 'Organization',
+      index: true,
```
p-fernandez, line 22: The query for the stats is done with `_environmentId` and `_organizationId`. I would expect this field not being indexed, to slow down the query. Another option would be to just query based on the indexed field `_environmentId` as it should be enough.

scopsy: I have introduced the compound index of:
```
{ _environmentId: 1, createdAt: -1} 
```

yesterday, because of the cardinality of the environment field it will be better the the organizationId, and mongoDB is smart enough to use the new compound index for the new query even if it contains the organizationId.

One thing we can actually do is to remove the organizationId from the query.

scopsy (commented):
Added a few comments regarding the selected query, I'm afraid it might introduce more performance issues once implemented :(

apps/api/src/app/notifications/usecases/get-activity-stats/get-activity-stats.usecase.ts
p-fernandez: @scopsy this was meant more like a short term fix. I do think we need to review the UI and what's the benefit of showing one year long usage all the time.
But you are right for a long term your proposal should perform better. Let me try to experiment a bit with it in MongoDB.

libs/dal/src/repositories/notification/notification.schema.ts
p-fernandez: @scopsy The compound index was a great move. 👍🏻 
I think we should also introduce the index for organization for future multitenancy purposes. We might forget.

scopsy (approved):
Looking good!

PR #3043, opened 2023-03-19
feat: remove onboarding history management

djabarovgeorge:

Remove all the management of the steps history logic, switched it to the parent component to be responsible for the knowledge of what is the previous page, in addition, changed the route of '/quickstart/notification-center/trigger' to '/quickstart/notification-center/set-up/:framework/trigger' in order to have the indication from what page we got to the '.../trigger' route so we could manage the route back.

Due to the possibility of bugs (and maintaining overhead)


scopsy (approved):
🔥

PR #3048, opened 2023-03-19
fix(integration): tlsOptions for nodemailer integration

### LINKED ISSUE #3010: [NV-1838] 🐛 Bug Report: tlsOptions are not mapped to NodeMailer integration
### 📜 Description

When a user tries to set tlsOptions for an integration via the UI such as `{ "rejectUnauthorized": false }` it seems that it is not written to the DB and is not propagated to the nodemailer integration.

It seems that the NodeMailer config for tls is always set to `undefined` here: https://github.com/novuhq/novu/blob/4727c8eda6228dea4f4f48c583c081ba292a41a2/providers/nodemailer/src/lib/nodemailer.provider.ts#L58

### 👟 Reproduction steps

1. Go to integrations
2. Select Custom SMTP 
3. Set the host, port, `secure=true` and `tlsOptions={ "rejectUnauthorized": false }`
4. Click "save"

### 👍 Expected behavior

1. When the integration is opened again, the `tlsOptions` text field should contain the value that was saved, i.e. it should not be empty
2. The `tlsOptions` should be used in the NodeMailer integration

image

image

With this bug, it is not possible to test out `novu` in productive environments of intranets that have their own internal SMTP servers :( 

None

<sub>[NV-1838](https://linear.app/novu/issue/NV-1838/🐛-bug-report-tlsoptions-are-not-mapped-to-nodemailer-integration)</sub>

akhill10:
-  added `requireTls` , `ignoreTls` and `tlsOptions`  in the Intergration Db model and credentialsDto for create integration api.

- Fixes #3010 : While saving the configuration for custom SMTP integration  the following fields `requireTls` , `ignoreTls` and `tlsOptions` aren't saved in the Db. Hence unable test out the custom smtp with tls configurations.


libs/dal/src/repositories/integration/integration.schema.ts
```diff
@@ -37,6 +37,9 @@ const integrationSchema = new Schema<IntegrationDBModel>(
       clientId: Schema.Types.String,
       projectName: Schema.Types.String,
       serviceAccount: Schema.Types.String,
+      requireTls: Schema.Types.String,
+      ignoreTls: Schema.Types.String,
+      tlsOptions: Schema.Types.String,
```
p-fernandez, line 42: Unfortunately we shouldn't do this. `tlsOptions` is a quite complicated object (https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options) so for a _quick_ implementation we would need for now to use `Schema.Types.Mixed` in the Schema.

Then also we have the problem that Web app is capturing strings and `requireTls` and `ignoreTls` should be stored as booleans (besides what's mentioned for `tlsOptions`).
That implies proper mapping in the appropriate DTOs of API and validation.

So unfortunately we can't do a simple string 2 string mapping throughout all the system. 😞

p-fernandez (changes requested):

libs/dal/src/repositories/integration/integration.schema.ts
akhill10: Thanks for the review. Sure, will make the changes accordingly.

A quick question though, why are we maintaining the schema type as String even though it's boolean? I have seen the `secure` key is also capturing it as a string in web app, but DTO is defined as a boolean, and in the schema it's defined as a String. So followed the same approach with `requireTls` and `ignoreTls`

 Should we not maintain consistency overall as web app, DTO and Schema ?

p-fernandez: I personally think with `secure` the wrong decision was taken for a quick development. 🙁 Besides that the content is not being validated, so even if with TS we have a different type, at runtime is passing strings.
And right now we shouldn't change it without doing a migration because it would break any integration users had already set up so we can only do the proper thing with the new additions.

@scopsy @LetItRock tagging you to align as we would need to change some of the frontend and overall we have a misalignment of existing properties with their proper type DB wise.

akhill10: @p-fernandez Made the schema changes.

Also added a checkbox option on the web app for the fields `requireTls` and `ignoreTls`.  These values will be captured as a boolean in web app. 
Didn't make any changes to `secure` as it would break existing things.
 
Let me know if this works out.  

Screenshot 2023-03-22 at 4 35 46 PM

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -95,6 +95,39 @@ describe('Create Integration - /integration (POST)', function () {
...
+  it('should create custom SMTP integration with TLS options successfully', async function () {
+    const payload = {
+      providerId: 'nodemailer',
+      channel: 'email',
+      credentials: {
+        host: 'smtp.example.com',
+        port: '587',
+        secure: 'true',
+        requireTls: true,
+        tlsOptions: JSON.stringify({ rejectUnauthorized: false }),
```
p-fernandez, line 108:
```suggestion
        tlsOptions: { rejectUnauthorized: false },
```
We are expecting to send an object for TLS options so the test should reflect that.
Also we need to review the API calls from the Web App, as it would need to transform from the string value stored in the form to the object we need to send. Same when retrieving the integration to convert from object to string to fill the form field when editing the integration. And for the update case.
https://github.com/novuhq/novu/blob/371c1381afb3e6b919a4355a649585880da65150/apps/web/src/api/integration.ts#L16

And not forgetting handle cases of malformed JSON. Not a straight forward implementation was this task. 🙁

p-fernandez (changes requested):

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -95,6 +95,52 @@ describe('Create Integration - /integration (POST)', function () {
...
+    const nodeMailerStoredIntegration = (await integrationRepository.findByEnvironmentId(environmentId)).find(
+      (i) => i.providerId.toString() === 'nodemailer'
+    );
+    expect(nodeMailerStoredIntegration?.credentials?.host).to.equal(payload.credentials.host);
+    expect(nodeMailerStoredIntegration?.credentials?.port).to.equal(payload.credentials.port);
+    expect(nodeMailerStoredIntegration?.credentials?.secure).to.equal(payload.credentials.secure);
+    expect(nodeMailerStoredIntegration?.credentials?.requireTls).to.equal(payload.credentials.requireTls);
+    expect(nodeMailerStoredIntegration?.credentials?.tlsOptions).to.instanceOf(Object);
+    expect(nodeMailerStoredIntegration?.credentials?.tlsOptions).to.eql(payload.credentials.tlsOptions);
+    expect(nodeMailerStoredIntegration?.active).to.equal(true);
```
p-fernandez, line 142: All these checks are already done between lines 121 and 130. Not need to do them again. 🙂

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
```diff
@@ -79,4 +79,39 @@ describe('Update Integration - /integrations/:integrationId (PUT)', function ()
...
+      active: true,
+      check: false,
+    };
+
+    // create integration
+    const nodeMailerIntegrationId = (await session.testAgent.post('/v1/integrations').send(nodeMailerProviderPayload))
+      .body.data._id;
+
+    // update integration
+    await session.testAgent.put(`/v1/integrations/${nodeMailerIntegrationId}`).send(nodeMailerProviderPayload);
```
p-fernandez, line 103: For the update, change the payload update so we can check the changes have been made. For example, change `requireTls` to false and change `tlsOptions` object. That should be enough to test the update. 👍🏻

apps/web/src/pages/integrations/IntegrationsStorePage.tsx
```diff
@@ -218,7 +218,9 @@ function initializeProviders(integrations: IntegrationEntity[]): IIntegratedProv
     if (integration?.credentials && Object.keys(clonedCredentials).length !== 0) {
       clonedCredentials.forEach((credential) => {
         // eslint-disable-next-line no-param-reassign
-        credential.value = integration.credentials[credential.key]?.toString();
+        if (credential.type === 'object') {
+          credential.value = JSON.stringify(integration.credentials[credential.key]);
+        } else credential.value = integration.credentials[credential.key]?.toString();
```
p-fernandez, line 223:
```suggestion
        if (credentialcredential.type === 'object') {
          credential.value = JSON.stringify(integration.credentials[credential.key]);
        } else {
        credential.value = integration.credentials[credential.key]?.toString();
       }
```
Just to clarify the else.

p-fernandez (approved):
Just last thing the tests suggested changes and we should be good to go to merge this. Though I want to test it a bit in my local when you are done. 👍🏻

akhill10:
@p-fernandez Thanks.  Requested changes are done. You can test it locally. Let me know if anything needs to be taken care of.

apps/web/src/pages/integrations/components/ConnectIntegrationForm.cy.tsx
```diff
@@ -116,7 +122,9 @@ it('shows the configuration for the selected provider', () => {
 
   // We may use a for-loop here since order of checks is not important
   for (const cred of credentials) {
-    cy.get(`input[name="${cred.key}"]`).should('have.attr', 'placeholder', cred.displayName);
+    if (cred.key === CredentialsKeyEnum.RequireTls)
+      cy.get(`input[name="${cred.key}"]`).should('have.attr', 'type', 'checkbox');
```
p-fernandez, line 126: This test is failing, might need to review the implementation as something is off.

akhill10: Ah! My bad. Credential type defined in the test case must be `switch` instead of `boolean`.
Fixed it. It should pass now.

p-fernandez:
@akhill10 I tested this locally generally works well. 
There are few bugs for certain values do the operations we do to transform a stringified JSON to object and back but they are minor and can be addressed in a further PR.
For example if you leave empty the field TLS options and update the integration you will get:
Screenshot 2023-03-31 at 11 05 58

Many thanks for this PR. This is a great contribution. 🌟

PR #3056, opened 2023-03-20
feat(web): integrations store modal analytics

LetItRock:

Integration Store Modal - segment analytics.

Screenshot 2023-03-20 at 17 22 11
Screenshot 2023-03-20 at 17 18 13


apps/web/src/pages/integrations/constants.ts
```diff
@@ -0,0 +1,6 @@
+export enum IntegrationsStoreModalAnalytics {
+  CREATE_INTEGRATION_FORM_SUBMIT = '[Integrations Store Modal] Create Integration Form Submit',
```
scopsy, line 2:
```suggestion
  CREATE_INTEGRATION_FORM_SUBMIT = 'Create Integration Form Submit - [Integrations Store Modal]',
```

A small suggestion is to move the category to the end. It's just easier to read on mixpanel because their events dropdown are quite small, so you would only read the repeated beginning :)

PR #3063, opened 2023-03-21
feat(regression): add test for cc and bcc overrides

p-fernandez:

Adds test to the suite to test the CC and BCC emails overrides.

It is one of our pain points by breaking this functionality so we need to have confidence in its behaviour.

### DIFF
--- a/apps/api/src/providers/provider.regression.ts
+++ b/apps/api/src/providers/provider.regression.ts
@@ -265,12 +265,9 @@ describe('Regression test - Providers', () => {
             let ccMessage;
             let bccMessage;
             for (const inboxMessage of inboxMessages) {
-              const {
-                smtp_information: {
-                  data: { rcpt_to_addrs: rcptToAddrs },
-                },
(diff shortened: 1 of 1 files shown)


apps/api/src/providers/provider.regression.ts
```diff
@@ -238,6 +240,82 @@ describe('Regression test - Providers', () => {
...
+
+            let mainMessage;
+            let ccMessage;
+            let bccMessage;
+            for (const inboxMessage of inboxMessages) {
+              const {
+                smtp_information: {
+                  data: { rcpt_to_addrs: rcptToAddrs },
+                },
+              } = inboxMessage;
```
scopsy, line 272:
```suggestion
              const recipientToAddress = inboxMessage?.smtp_information?.data?.rcpt_to_addrs;
```

A very personal thing, but I find this one more readable here :)

p-fernandez: I can live with it. 🙂

PR #3065, opened 2023-03-21
fix: remove activity and add notification tag

jainpawan21:
Remove the deprecated APIs `Activity` tag and add the `Notification` tag

To remove https://docs.novu.co/api/get-activity-feed/ APIs and add Notification apis
Because these activity APIs are deprecated, they are not returning the expected response.


LetItRock (approved):

jay777-777:
Hi css

PR #3066, opened 2023-03-22
fix: Node client bulk trigger

szbartnik:

Fixes Node client's `bulkTrigger` operation.

Node client's `bulkTrigger` is throwing an error:
```
TypeError: novu.bulkTrigger is not a function
```

Temporary workaround I'm using:
```typescript
novuClient.events.bulkTrigger
```


scopsy (approved):
Good catch! thank you!

PR #3069, opened 2023-03-22
Fix type definition

oskar-kupski-elpassion:

Added proper types definition based on type used in `novu/packages/notification-center/src/store/notifications.context.ts`

The actual value returned by the hook is sometimes `null` ( probably on init ). Based on types it should not be `null`.


LetItRock:
hey @oskar-kupski-elpassion 👋 
The `useNotifications` hook will never return the `null` if you wrap your components with the `NovuProvider`, please check the docs here: https://docs.novu.co/notification-center/react/react-components#custom-ui
If it returns `null` for you sometimes it means that the component that is rendering `NovuProvider` is unmounted.

PR #3075, opened 2023-03-24
fix: get latest version of actor details while retreiving messages/feed list

BiswaViraj:
- Serves the latest available avatar of the actor while fetching messages/feeds

- Previously, during the queue job processing the avatar was fetched and served to the notification list.
This caused a scenario where after updating the avatar, the notification list will still show the old avatar for older messages

### DIFF
--- a/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
+++ b/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
@@ -54,7 +54,7 @@ export class GetMessages {
 
     for (const message of data) {
       if (message._actorId && message.actor?.type === ActorTypeEnum.USER) {
-        message.actor.data = await this.processAvatar(message._actorId, command.environmentId);
+        message.actor.data = this.processUserAvatar(message.actorSubscriber);
       }
     }
 
@@ -83,15 +83,7 @@ export class GetMessages {
     return await this.subscriberRepository.findBySubscriberId(_environmentId, subscriberId);
   }
 
-  private async processAvatar(actorId: string, environmentId: string): Promise<string | null> {
-    const actorSubscriber: SubscriberEntity | null = await this.subscriberRepository.findOne(
-      {
-        _environmentId: environmentId,
-        _id: actorId,
-      },
-      'avatar'
-    );
-
+  private processUserAvatar(actorSubscriber?: SubscriberEntity): string | null {
     return actorSubscriber?.avatar || null;
   }
 }
--- a/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
+++ b/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
@@ -1,5 +1,15 @@
 import { BadRequestException, Injectable } from '@nestjs/common';
 import { MessageEntity, MessageRepository, SubscriberRepository, SubscriberEntity } from '@novu/dal';
+import {
+  ChannelTypeEnum,
+  IMessageButton,
+  ExecutionDetailsSourceEnum,
+  ExecutionDetailsStatusEnum,
+  IEmailBlock,
+  InAppProviderIdEnum,
+  ActorTypeEnum,
+  IActor,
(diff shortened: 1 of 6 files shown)


libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -112,6 +112,11 @@ const messageSchema = new Schema<MessageDBModel>(
       },
       data: Schema.Types.Mixed,
     },
+    _actorId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Subscriber',
+      index: true,
```
scopsy, line 118: This index could be removed no need to have it.

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -50,6 +50,12 @@ export class GetNotificationsFeed {
       });
     }
 
+    for (const message of feed) {
+      if (message._actorId && message.actor?.type === ActorTypeEnum.USER) {
+        message.actor.data = await this.processAvatar(message._actorId, command.environmentId);
+      }
+    }
```
scopsy, line 57: Unfortunately, this approach is not very performant is we don't utilize parallelization here, nor using caching. 

Could you please this logic to use "populate" from mongoose on the dal level instead of looping here on this?

scopsy (changes requested):
🎉 Added a small note regarding performance here

libs/dal/src/repositories/message/message.schema.ts
BiswaViraj: removed

libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -130,6 +134,13 @@ messageSchema.virtual('template', {
   justOne: true,
 });
 
+messageSchema.virtual('actorSubscriber', {
```
scopsy, line 137:
```suggestion
messageSchema.virtual('actor', {
```

Should we maybe just call it "actor" or there might some reason for this?

apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
```diff
@@ -40,11 +50,29 @@ export class GetMessages {
       skip: command.page * LIMIT,
     });
 
+    for (const message of data) {
+      if (message._actorId && message.actor?.type === ActorTypeEnum.USER) {
+        message.actor.data = await this.processAvatar(message._actorId, command.environmentId);
+      }
+    }
```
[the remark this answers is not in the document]
BiswaViraj: I'm moving this one to DAL currently

libs/dal/src/repositories/message/message.schema.ts
BiswaViraj: we already have an existing `actor` field on the message entity with the shape

```js
actor: { 
type: ActorTypeEnum, // USER, SYSTEM ICON, SYSTEM CUSTOM URL
data: string | null
}
```

libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -69,11 +69,21 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
...
       sort: '-createdAt',
-    }).populate('subscriber', '_id firstName lastName avatar subscriberId');
+    })
+      .populate('subscriber', '_id firstName lastName avatar subscriberId')
+      .populate({
+        path: 'actorSubscriber',
+        match: {
+          'actor.type': ActorTypeEnum.USER,
+          _actorId: { $exists: true },
+        },
```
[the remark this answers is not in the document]
BiswaViraj: We need to populate actor, only when actorID exists and the actor type is USER, as we have different actor types as well i.e,```USER, SYSTEM_ICON, SYSTEM_CUSTOM```

scopsy (approved):

PR #3077, opened 2023-03-25
fix: add missing authmodule import in layout module

jainpawan21:

1. Add AuthModule import in LayoutModule [suggested by @ainouzgali ]
2. Add  @UseGuards(JwtAuthGuard) in layouts controller apis

Layouts APIs are not accessible using api key. Only accessible with bearer token


scopsy:
@jainpawan21 i'm not sure this is needed due to `@UseGuards(JwtAuthGuard)` is located on the controller level and is applied to each controller endpoint. This pattern exists in a lot of our controller.

jainpawan21:
> @jainpawan21 i'm not sure this is needed due to `@UseGuards(JwtAuthGuard)` is located on the controller level and is applied to each controller endpoint. This pattern exists in a lot of our controller.

Did not notice that we are using this guard at controller level. Removed from route level ✅

ainouzgali (approved):
⭐️

PR #3079, opened 2023-03-26
feat: refactor cache functions

djabarovgeorge:


apps/api/src/app/shared/interceptors/invalidate-cache.interceptor.ts
djabarovgeorge, line 1: removed this decorator so no one will use the evil scan on the cache by mistake.

apps/api/src/app/shared/services/cache/invalidate-cache.service.ts
```diff
@@ -1,43 +1,12 @@
...
-
-      Logger.debug('invalidate promises are: ' + invalidatePromises);
-      Logger.verbose('Removing all keys with prefix');
-      await Promise.all(invalidatePromises);
-    } else {
-      Logger.warn('StoreKeyPrefix is not in array format');
-      await this.clearByPattern(storeKeyPrefix, credentials);
-    }
-  }
-
```
djabarovgeorge, line 40: same here removed this function so no one will use the evil scan on the cache by mistake.
I still left the 'clearByPattern' function because it is private, I may delete it in near future.

apps/api/src/app/shared/services/cache/key-builders/entities.ts
djabarovgeorge, line 1: split the entities keys builder functions to separate files.

apps/api/src/app/shared/services/cache/key-builders/queries.spec.ts
djabarovgeorge, line 1: split the quieries keys builder functions to separate files.

djabarovgeorge:
Still did not have the time to check if we can optimize the notification templates invalidation. hopefully will look into it this week.

apps/api/src/app/shared/services/cache/index.ts
```diff
@@ -1,3 +1,2 @@
 export * from './invalidate-cache.service';
 export * from './cache.service';
-export { CacheKeyPrefixEnum } from './keys';
```
p-fernandez, line 3: I would export the key builders so the imports would be simplified in the rest of the codebase.

p-fernandez (commented):
This is a great refactor. Even if it is quite subtle.
👏🏻

PR #3082, opened 2023-03-27
docs: change introduction messaging

sumitsaurabh927:

It changes the text of 'introduction' in the 'overview' in docs.

This has better messaging, in my opinion. It is divided into two paragraphs, so that instead of being one large block of text, it is two smaller ones. This helps in improving readability.


p-fernandez (approved):
🌟

sumitsaurabh927:
> 🌟

♥️

PR #3086, opened 2023-03-27
feat(api): create webhook filter backoff strategy use case

p-fernandez:

Extracts from the WorkflowService the logic to calculate the backoff strategy and it also helps to move a step forward to remove the ExecutionDetails dependency from it.

Keep on the work to make WorkflowService driven by dependency injection so the use cases injected drive the behaviour of it in a more formal clean architecture.

### DIFF
--- a/apps/api/src/app/events/e2e/trigger-event.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event.e2e.ts
@@ -38,7 +38,9 @@ import { v4 as uuid } from 'uuid';
 
 const axiosInstance = axios.create();
 
-describe('Trigger event - /v1/events/trigger (POST)', function () {
+const eventTriggerPath = '/v1/events/trigger';
+
+describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
   let session: UserSession;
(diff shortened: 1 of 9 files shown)


p-fernandez:
@LetItRock this PR might be interesting for the work you are doing right now with the new Job service. Hopefully I didn't clash with it. 😅

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -779,14 +781,14 @@ describe('Trigger event - /v1/events/trigger (POST)', function () {
...
-    const block = message.content[0] as IEmailBlock;
+    const block = message?.content[0] as IEmailBlock;
 
     expect(block.content).to.equal('Hello John Doe, Welcome to Umbrella Corp');
   });
 
   it('should throw an error when workflow identifier provided is not in the database', async () => {
     const response = await session.testAgent
-      .post(`/v1/events/trigger`)
+      .post(eventTriggerPath)
```
scopsy, line 791: One of the things I like about you is that you always leave the place cleaner than when you got to it :)

scopsy (approved):

PR #3088, opened 2023-03-27
chore(application-generic): move shared services to application-generic package

LetItRock:

Moved some shared services to the `application-generic` package:
- caching service
- performance
- storage

We want to move the jobs processing logic into a separate service that will be easier to dynamically scale based on the queue size.


apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -11,7 +11,7 @@ import {
   UserRepository,
 } from '@novu/dal';
 import { AuthProviderEnum, IJwtPayload, ISubscriberJwt, MemberRoleEnum, SignUpOriginEnum } from '@novu/shared';
-import { AnalyticsService } from '@novu/application-generic';
+import { AnalyticsService, CacheKeyPrefixEnum, Cached } from '@novu/application-generic';
```
LetItRock, line 14: moved caching service and all the interceptors to the `application-generic` package

packages/application-generic/src/services/storage/storage-helper.service.spec.ts
```diff
@@ -0,0 +1,245 @@
+import { File } from '@google-cloud/storage';
```
LetItRock, line 1: had to rewrite it to jest test

packages/application-generic/src/services/cache/cache-service.spec.ts
```diff
@@ -1,13 +1,13 @@
-import { expect } from 'chai';
 import { CachingConfig, ICacheService } from './cache.service';
 
 describe('cache-service', function () {
   let cacheService: ICacheService;
-  before(function () {
+
+  beforeEach(function () {
```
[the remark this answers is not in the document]
scopsy: @djabarovgeorge just to get your second eyes here

scopsy (approved):
Amazing work, added a few comments

packages/application-generic/src/services/cache/cache.service.ts
```diff
@@ -37,14 +37,24 @@ export class CacheService implements ICacheService {
...
 
-      this.client = new Redis(Number(this.config.port || 6379), this.config.host, {
-        password: this.config.password,
-        connectTimeout: this.config.connectTimeout ? Number(this.config.connectTimeout) : this.DEFAULT_CONNECT_TIMEOUT,
-        keepAlive: this.config.keepAlive ? Number(this.config.keepAlive) : this.DEFAULT_KEEP_ALIVE,
-        family: this.config.family ? Number(this.config.family) : this.DEFAULT_FAMILY,
-        keyPrefix: this.config.keyPrefix ?? this.DEFAULT_KEY_PREFIX,
-        tls: this.config.tls,
-      });
+      this.client = new Redis(
```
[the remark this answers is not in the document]
LetItRock: I've found the problem, the ESLint config has `'max-len': ['warn', { code: 140 }],` and Prettier `"printWidth": 120,`
So it looks like when the code is not formatted and is committed to Git the ESLint prefers its own rules ;)

packages/application-generic/src/services/storage/storage-helper.service.spec.ts
LetItRock: yeah, it will be more consistent with the rest...

PR #3105, opened 2023-03-29
fix: upgrade bullmq versions

scopsy:

We have mismatched version between the difference services. Updated all versions to be the same.

### DIFF
--- a/apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
+++ b/apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
@@ -1,5 +1,5 @@
 import { forwardRef, Inject, Injectable } from '@nestjs/common';
-import { Job, JobsOptions, Queue, QueueBaseOptions, QueueScheduler, Worker } from 'bullmq';
+import { Job, JobsOptions, QueueBaseOptions, WorkerOptions } from 'bullmq';
 // TODO: Remove this DAL dependency, maybe through a DTO or shared entity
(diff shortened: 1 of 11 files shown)


apps/api/src/app/events/services/workflow-queue/workflow.queue.service.ts
```diff
@@ -89,15 +84,15 @@ export class WorkflowQueueService {
...
 
-  private getWorkerOpts() {
+  private getWorkerOpts(): WorkerOptions {
     return {
       ...this.bullConfig,
       lockDuration: 90000,
       concurrency: 200,
       settings: {
-        backoffStrategies: this.getBackoffStrategies(),
+        backoffStrategy: this.getBackoffStrategies(),
```
scopsy, line 93: Was refactored to a function instead of object since version 2

packages/application-generic/src/services/bull-mq.service.ts
```diff
@@ -66,12 +63,6 @@ export class BullmqService {
     return this._worker;
   }
 
-  public createScheduler(name: string, config: QueueBaseOptions) {
-    this._queueScheduler = new QueueScheduler(name, config);
-
-    return this._queueScheduler;
-  }
-
```
scopsy, line 74: Scheduler is not needed because it is now taken care of by the Worker class

p-fernandez (approved):
Good improvements. 
🌟

PR #3106, opened 2023-03-29
Index Cleanup - Changes

djabarovgeorge:

Why? (Context)

### DIFF
--- a/libs/dal/src/repositories/change/change.repository.ts
+++ b/libs/dal/src/repositories/change/change.repository.ts
@@ -1,23 +1,23 @@
 import { ChangeEntityTypeEnum } from '@novu/shared';
 
-import { EnforceEnvOrOrgIds } from '../../types/enforce';
+import { EnforceEnvId } from '../../types/enforce';
 import { BaseRepository } from '../base-repository';
 import { ChangeEntity, ChangeDBModel } from './change.entity';
 import { Change } from './change.schema';
 
-export class ChangeRepository extends BaseRepository<ChangeDBModel, ChangeEntity, EnforceEnvOrOrgIds> {
+export class ChangeRepository extends BaseRepository<ChangeDBModel, ChangeEntity, EnforceEnvId> {
   constructor() {
     super(Change, ChangeEntity);
   }
 
   public async getEntityChanges(
-    organizationId: string,
+    environmentId: string,
     entityType: ChangeEntityTypeEnum,
     entityId: string
   ): Promise<ChangeEntity[]> {
     return await this.find(
       {
-        _organizationId: organizationId,
+        _environmentId: environmentId,
         _entityId: entityId,
         type: entityType,
       },
--- a/libs/dal/src/repositories/change/change.schema.ts
+++ b/libs/dal/src/repositories/change/change.schema.ts
@@ -43,13 +43,10 @@ changeSchema.virtual('user', {
 });
 
 changeSchema.index({
-  _environmentId: 1,
   _parentId: 1,
   _entityId: 1,
+  _environmentId: 1,
   createdAt: 1,
-  type: 1,
-  enabled: 1,
-  _creatorId: 1,
 });
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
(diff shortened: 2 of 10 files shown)


libs/dal/src/repositories/change/change.repository.ts
```diff
@@ -1,23 +1,23 @@
 import { ChangeEntityTypeEnum } from '@novu/shared';
 
-import { EnforceEnvOrOrgIds } from '../../types/enforce';
+import { EnforceEnvId } from '../../types/enforce';
 import { BaseRepository } from '../base-repository';
 import { ChangeEntity, ChangeDBModel } from './change.entity';
 import { Change } from './change.schema';
 
-export class ChangeRepository extends BaseRepository<ChangeDBModel, ChangeEntity, EnforceEnvOrOrgIds> {
+export class ChangeRepository extends BaseRepository<ChangeDBModel, ChangeEntity, EnforceEnvId> {
```
djabarovgeorge, line 8: Decided to remove the queries with OrganizationId where is not needed to reduce the amount of indexes, still not entirely sure regarding this collection.

libs/dal/src/repositories/change/change.schema.ts
```diff
@@ -44,6 +42,16 @@ changeSchema.virtual('user', {
   justOne: true,
 });
 
+changeSchema.index({
+  _environmentId: 1,
+  _parentId: 1,
+  _entityId: 1,
+  createdAt: 1,
+  type: 1,
+  enabled: 1,
```
djabarovgeorge, line 51: Still not sure about those two, because I am not entirely sure if the compound index will be unique here.

scopsy: Those are not really needed because their cardinality is VERY low (boolean). So suggest removing both type and enabled from there

libs/dal/src/repositories/change/change.schema.ts
```diff
@@ -44,6 +42,16 @@ changeSchema.virtual('user', {
   justOne: true,
 });
 
+changeSchema.index({
+  _environmentId: 1,
+  _parentId: 1,
+  _entityId: 1,
+  createdAt: 1,
```
scopsy, line 49: Are we doing sorts with "-createdAt" ? depending on the order you might need to switch to "-1"

libs/dal/src/repositories/change/change.schema.ts
djabarovgeorge: i think i will remove them because the scan should be really fast in those cases.

libs/dal/src/repositories/change/change.schema.ts
djabarovgeorge: We do, that's why I added it here so we don't do blocking sort.
And yes we do by 1

libs/dal/src/repositories/change/change.repository.ts
djabarovgeorge: Yeah that why i raised here, i wanted to check first on the tests level and then validate with David :)

libs/dal/src/repositories/change/change.schema.ts
```diff
@@ -44,6 +42,13 @@ changeSchema.virtual('user', {
   justOne: true,
 });
 
+changeSchema.index({
+  _parentId: 1,
+  _entityId: 1,
+  _environmentId: 1,
```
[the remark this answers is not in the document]
djabarovgeorge: Long story short let me copy an explanation from the report we got:

● A compound index can be utilized to satisfy multiple queries. For example, if there is
a compound index like { a : 1, b : 1, c:1 }, this can be utilized to satisfy all the
following queries -
* db.coll.find({ a : 3 })
* db.coll.find({ a : 3, b : 5})
* db.coll.find({ a : 3, b : 5, c : 8})

Let me know if it helps to understand the logic behind it better. @p-fernandez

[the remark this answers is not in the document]
p-fernandez: I missed an `all together` in my question. So MongoDB does the magic behind scenes. Then good for me. 🤐

p-fernandez (approved):

djabarovgeorge:
Hey @davidsoderberg 
Do you see any drawbacks to removing the organizationId from the change queues?

PR #3108, opened 2023-03-29
Initial setting of TTL index

ainouzgali:

WIP : an initial version of setting a ttl index for Message and Notification schemes. 
I will of course, improve the function of expire date calculation- I first want to progress on how we will handle a notification/job that is delayed/ not completed on the time of its creation. And therefore its expiration date should take that into account.

### DIFF
--- a/apps/api/src/app/events/e2e/trigger-event.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event.e2e.ts
@@ -36,6 +36,10 @@ import { EmailEventStatusEnum } from '@novu/stateless';
 import { v4 as uuid } from 'uuid';
 import { differenceInMilliseconds, subMonths } from 'date-fns';
 
+const IN_APP_MESSAGE_EXPIRE_MONTHS = 6;
+const MESSAGE_EXPIRE_MONTHS = 1;
+const NOTIFICATION_EXPIRE_MONTHS = 1;
+
 const axiosInstance = axios.create();
 
 const eventTriggerPath = '/v1/events/trigger';
@@ -285,7 +289,7 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
     let expireAt = new Date(notification?.expireAt as string);
     let createdAt = new Date(notification?.createdAt as string);
 
-    let subExpireMonths = subMonths(expireAt, 1);
+    let subExpireMonths = subMonths(expireAt, NOTIFICATION_EXPIRE_MONTHS);
     let diff = differenceInMilliseconds(subExpireMonths, createdAt);
 
     expect(diff).to.approximately(0, 1);
@@ -302,7 +306,7 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
     expireAt = new Date(message?.expireAt as string);
     createdAt = new Date(message?.createdAt as string);
 
-    subExpireMonths = subMonths(expireAt, 6);
+    subExpireMonths = subMonths(expireAt, IN_APP_MESSAGE_EXPIRE_MONTHS);
     diff = differenceInMilliseconds(subExpireMonths, createdAt);
 
     expect(diff).to.approximately(0, 1);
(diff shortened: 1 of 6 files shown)


libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -112,10 +112,13 @@ const messageSchema = new Schema<MessageDBModel>(
       },
       data: Schema.Types.Mixed,
     },
+    expireAt: Schema.Types.Date,
   },
-  { ...schemaOptions }
+  schemaOptions
 );
 
+messageSchema.index({ expireAt: 1 }, { expires: '1h' });
```
ainouzgali, line 120: 1 hour after its expiration date - to allow buffer between archiving and its actual deletion. When we get to the archive will rethink that time amount.

libs/dal/src/repositories/base-repository.ts
```diff
@@ -81,7 +82,27 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
     }
   }
 
+  private calcExpireDate(modelName: string, data: FilterQuery<T_DBModel> & T_Enforcement) {
+    const now: number = Date.now();
```
Cliftonz, line 86: Looks great!

libs/dal/src/repositories/message/message.schema.ts
Cliftonz: @ainouzgali  I dont think this implementation is correct. By what the docs say this will ignore our actually ttl we set in the app and set every document to 1hour after creation. 
https://www.mongodb.com/docs/manual/tutorial/expire-data/#expire-documents-at-a-specific-clock-time

scopsy:
@ainouzgali can we please have tests added to this aswell? Or is it part of another PR?

libs/dal/src/repositories/message/message.schema.ts
ainouzgali: It deletes by the expireAt field, not the 1h. But it does ignore the expire. So it does not work right. Will try a different option

ainouzgali:
I will add tests in a  separate PR @scopsy

libs/dal/src/repositories/message/message.schema.ts
Cliftonz: The key thing for us is that the ttl should be set by the app not the db.

ainouzgali: Never mind @Cliftonz , It does work as expected - deletes by the expire date we set and then waits the amount waited in the expires field.

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -255,6 +254,77 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
...
+    expect(emails.length).to.equal(1);
+    const email = emails[0];
+
+    expireAt = new Date(email?.expireAt as string);
+    createdAt = new Date(email?.createdAt as string);
+
+    subExpireMonths = subMonths(expireAt, 1);
+    diff = differenceInMilliseconds(subExpireMonths, createdAt);
+
+    expect(diff).to.approximately(0, 1);
```
scopsy, line 325: Is it possible to create constants somehwere in the top of the case so we can easily understand the intended behaviour? For example:
```
const IN_APP_MESSAGE_EXPIRE_MONTHS = 6;
const MESSAGE_EXPIRE_MONTHS = 1;
```

Right now need to manually do the logic to understand that the second param of the subMonths function is for this value.

Some more read about it: https://www.pluralsight.com/tech-blog/avoiding-magic-numbers/ :)

libs/dal/src/repositories/base-repository.ts
```diff
@@ -81,7 +82,27 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
...
+        return addMonths(now, 1);
+      default:
+        return null;
+    }
+  }
+
   async create(data: FilterQuery<T_DBModel> & T_Enforcement): Promise<T_MappedEntity> {
+    const expireAt = this.calcExpireDate(this.MongooseModel.modelName, data);
+    if (expireAt) {
+      data = { ...data, expireAt };
```
scopsy, line 104:
```suggestion
      data.expireAt = expireAt;
```

Is there a particular reason for the ...destructuring? 

I think it is not needed as we just can assign the needed property and save some memory :)

scopsy (approved):

PR #3109, opened 2023-03-29
Index Cleanup -  Execution Details

djabarovgeorge:

Why? (Context)

### DIFF
--- a/libs/dal/src/repositories/execution-details/execution-details.schema.ts
+++ b/libs/dal/src/repositories/execution-details/execution-details.schema.ts
@@ -72,23 +72,39 @@ const executionDetailsSchema = new Schema<ExecutionDetailsDBModel>(
 
 /*
  * This index was initially created to optimize:
- * message matcher use case
+ *
+ * Path : libs/dal/src/repositories/job/job.schema.ts
+ *    Context : The _jobId is here because of JobSchema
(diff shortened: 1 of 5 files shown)


libs/dal/src/repositories/execution-details/execution-details.schema.ts
```diff
@@ -79,6 +70,12 @@ const executionDetailsSchema = new Schema<ExecutionDetailsDBModel>(
   schemaOptions
 );
 
+executionDetailsSchema.index({
+  _messageId: 1,
+  _jobId: 1,
+  _environmentId: 1,
+});
+
```
djabarovgeorge, line 78: Would love to raise this index because he is very very heavy index storage-wise( every index is more than 1G ), therefore I tried to minimize the amount of the indexes. In addition, IMO most of the fetches are on notification.populate which is under _id.

I left indexes {  _messageId: 1,  _jobId: 1} because of the use of executionDetailsRepository.count in MessageMatcher.

The addition indexes that could have need added are:
  webhookStatus: 1,
  _notificationId: 1,
  _notificationTemplateId: 1,
  _subscriberId: 1,
  transactionId: 1,

p-fernandez (approved):

scopsy (approved):

PR #3110, opened 2023-03-29
chore: add nb and sv translations for notification actions

katyaterletskaya:

provide nb and sv translations for notification actions


scopsy (approved):
🎉 Thank you!

PR #3112, opened 2023-03-30
feat: hash reset tokens

### LINKED ISSUE #2978: 🚀 Feature: hash reset tokens to improve security
### 🔖 Feature description

This is an opinion based issue, but I think it's worth discussing. I have observed that Novu stores reset tokens as plain text in the database. I think hashing reset tokens can increase security in the event of a database breach as accounts with an active reset token will be exposed.

In case of database breach accounts with an active reset token will be exposed. Exposed reset tokens can be used to reset passwords by attackers.

### Reset password request
1. Generate SHA256 hash from generated reset token
2. Store hashed token in database
3. Send to the user plain reset token

### Reset password
1. Take plain reset token from the body
4. Hash reset token with SHA256
5. Use hashed token to query database

### Challenges
It will break tests because we cannot get plain reset token anymore.
```ts
  it('should change a password after reset', async () => {
    const { body } = await session.testAgent.post('/v1/auth/reset/request').send({
      email: session.user.email,
    });

    expect(body.data.success).to.equal(true);
    const foundUser = await userRepository.findById(session.user._id);

    const { body: resetChange } = await session.testAgent.post('/v1/auth/reset').send({
      password: 'ASd3ASD$Fdfdf',
      // resetToken is hashed and test will fail
      token: foundUser.resetToken,
    });

    expect(resetChange.data.token).to.be.ok;

```

### My suggestions
I think it's fine to implement this maybe in repository layer? It's all about storage.

#### Ways to avoid broken tests:
- some kind of field in `UserSchema` like `_plainResetToken` and could be set only in testing environments.
- hash only in production environment?

```ts
export class UserRepository extends BaseRepository<UserDBModel, UserEntity> {
  constructor() {
    super(User, UserEntity);
  }

  private hashResetToken(token: string) {
    if (process.env.NODE_ENV === 'prod') {
      return createHash('sha256').update(token).digest('hex');
    }
    // in tests we will be able to obtain plain token
    return token;
  }

  async findUserByToken(token: string) {
    return await this.findOne({
      resetToken: hashResetToken(token),
    });
  }

  async updatePasswordResetToken(userId: string, token: string, resetTokenCount: IUserResetTokenCount) {
    return await this.update(
      {
        _id: userId,
      },
      {
        $set: {
          resetToken: hashResetToken(token),
          resetTokenDate: new Date(),
          resetTokenCount,
        },
      }
    );
  }
```

### 🔄️ Additional Information

_No response_

michaldziuba03:

PR introduces change in reset password flow - the reset token will be stored as SHA256 hash which is commonly used for this purpose, Obviously user will receive plain reset token in the email message.

The goal of this feature is to improve security. We don't want to expose users with active reset tokens in case of database breach.

Closes #2978


scopsy (commented):
Amazing work! will be taking a closer look a bit later, but for now, does this handle backward compatibility? For example after deploying users still have the old tokens there

michaldziuba03:
@scopsy It doesn't provide backward compatibility.

I have two ideas:
1. Create script to update users with active plain reset token in batch and hash their tokens. It's not perfect and probably some users will need to request again reset password anyway.

2. We can query user by hashed reset token **OR** plain reset token. SHA256 has fixed length and uuids has their own fixed length so no collisions possible :)
```ts
{ resetToken: { $in: ['hashedToken', 'plainToken' ] } }
```

I can implement backward compatibility using second option if you agree with this solution.

scopsy:
@michaldziuba03 we can do the second option definitely, you can add there a note to remove this in a second minor version so we don't have this all the time there. Good idea!

michaldziuba03:
@scopsy PR is ready to review.

Everything seems to work - user receives link with plain token and database stores hashed reset token :)

results

scopsy (approved):
🚀

PR #3120, opened 2023-03-30
feat(api): in-memory provider service

p-fernandez:

Created a new service that isolates the implementation of any in-memory provider (in this case `ioredis`) so we can easily swap the mode (from normal provider to cluster mode) or to even swap a provider as needed. It will also would allow us to have different providers available depending on environment.
So far we leave it connected only to the cache service.

It is one of the first steps of the performance improvements by targeting to move cache service and distributed lock service to a Redis cluster setting.

### DIFF
--- a/apps/api/src/app/shared/services/cache/cache.service.ts
+++ b/apps/api/src/app/shared/services/cache/cache.service.ts
@@ -1,9 +1,9 @@
-import Redis from 'ioredis';
-import { ConnectionOptions } from 'tls';
 import { Logger } from '@nestjs/common';
 import { QUERY_PREFIX } from './key-builders/shared';
 
-const STORE_CONNECTED = 'ready';
+import { InMemoryProviderClient, InMemoryProviderService } from '../in-memory-provider';
+
+const LOG_CONTEXT = 'CacheService';
 
 export interface ICacheService {
   set(key: string, value: string, options?: CachingConfig);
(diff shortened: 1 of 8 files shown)


apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -0,0 +1,73 @@
...
+export class InMemoryProviderService {
+  public inMemoryProviderClient: InMemoryProviderClient;
+  public inMemoryProviderConfig: InMemoryProviderConfig;
+
+  constructor() {
+    if (!this.inMemoryProviderClient) {
+      this.inMemoryProviderClient = process.env.IN_MEMORY_CLUSTER_MODE
+        ? this.inMemoryClusterProviderSetup()
+        : this.inMemoryProviderSetup();
+    }
```
scopsy, line 21: From a very quick look at this, it seems as if Caching is required to run novu, where previously we had an optional for this (You can run novu without the cache service, if you haven't passed the REDIS_HOST params).

How the current behaviour will work? Can we still allow running novu without a cache service?

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -0,0 +1,73 @@
...
+   * https://docs.nestjs.com/fundamentals/lifecycle-events#application-shutdown
+   *
+   * Enabled by:
+   *   app.enableShutdownHooks();
+   *
+   * in /apps/api/src/bootstrap.ts
+   */
+  public async onApplicationShutdown(signal): Promise<void> {
+    await this.shutdown();
+  }
```
scopsy, line 72: Very valid point! We need to add something similar to bullmq to gracefully finish processing jobs before shut down.

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -0,0 +1,73 @@
...
+import { getRedisInstance, getRedisProviderConfig, IRedisProviderConfig, Redis } from './redis-provider';
+
+import { ApiException } from '../../exceptions/api.exception';
+
+const LOG_CONTEXT = 'InMemoryCluster';
+
+export type InMemoryProviderClient = Redis;
+type InMemoryProviderConfig = IRedisProviderConfig;
+
+export class InMemoryProviderService {
```
scopsy, line 12: One very little thing here regarding naming, maybe we can consider a different name for it, since initially I was sure that this service is an inMemory provider for testing purposes (instead of testing against a real db)

I know we want something more general than redis here, don't have a better unfortuently. Just want to point this, not necessarily a problem.

scopsy (approved):

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
p-fernandez: If we move away from in-memory databases we can easily do a different adapter service of other kind and just plug the dependant services (Cache, DistributedLock). I find this and other db adapter services to be able to live parallelly.

apps/api/src/app/shared/services/cache/cache.service.ts
```diff
@@ -1,9 +1,9 @@
-import Redis from 'ioredis';
-import { ConnectionOptions } from 'tls';
 import { Logger } from '@nestjs/common';
 import { QUERY_PREFIX } from './key-builders/shared';
 
-const STORE_CONNECTED = 'ready';
```
p-fernandez, line 6: As the connected status would vary from provider to provider we move the responsibility to the InMemoryProviderService.

apps/api/src/app/shared/services/cache/cache.service.ts
```diff
@@ -20,125 +20,111 @@ export type CachingConfig = {
...
   }
 
   public getStatus() {
     return this.client?.status;
   }
 
-  public cacheEnabled() {
-    return this.client?.status === STORE_CONNECTED;
+  public cacheEnabled(): boolean {
+    const isEnabled = this.inMemoryProviderService.isClientReady();
```
p-fernandez, line 38: Here, about this:
https://github.com/novuhq/novu/pull/3120/files#r1156254550

PR #3121, opened 2023-03-31
refactor(docs) Update CLI prompt message

guiruiz:
Updates the provider name CLI prompt message on docs.

It seems that the standard case style for the CLI provider name changed and the docs seems outdated.
So I'm updating just to avoid any mistake.

N/A


p-fernandez (approved):
🌟 
Thanks for this!

PR #3123, opened 2023-03-31
feat(api): distributed lock service should reuse in memory provider svc

p-fernandez:

Makes the distributed lock service to reuse the in-memory provider service.

We plan to implement Redis cluster as in-memory provider service for the Cache service and the Distributed Lock service. So far we had both connections to the in-memory provider (ioredis) separatedly even if using the same instance.


Cliftonz:
LGTM

djabarovgeorge (approved):
🚀

PR #3131, opened 2023-04-03
fix: add reply to support for aws ses provider

### LINKED ISSUE #3122: [NV-1955] 🐛 Bug Report: `replyTo` email override isn't working for AWS SES
### 📜 Description

I am self-hosting Novu version v0.13.0 using Docker, and I use AWS SES as my email provider.

I am following this guide (https://docs.novu.co/channels/email/#sending-email-overrides) to set replyTo override for email notifications.

The replyTo isn't working, I don't see `replyTo` address in the email that gets sent (attached a SS to provide just enough context)

image

### 👟 Reproduction steps

1. Set AWS SES as your email provider
2. While triggering an email notification, set `replyTo` address in email overrides object.
3. The email that gets sent will not have `replyTo` header set <-- that's the bug.

### 👍 Expected behavior

The email that gets sent should have `replyTo` header set.

The email that gets sent does not have `replyTo` header set. (check SS uploaded in description)

_No response_

None

<sub>[NV-1955](https://linear.app/novu/issue/NV-1955/🐛-bug-report-reply-to-email-override-isnt-working-for-aws-ses)</sub>

deep-bhatt-pmc:

- This PR sets the replyTo parameter in nodemailer SES transport.

- `replyTo` wasn't working for AWS SES provider. This PR fixes that bug.

Closes #3122

### DIFF
--- a/providers/ses/src/lib/ses.provider.spec.ts
+++ b/providers/ses/src/lib/ses.provider.spec.ts
@@ -12,6 +12,7 @@ const mockConfig = {
 
 const mockNovuMessage = {
   to: ['test@test2.com'],
+  replyTo: 'test@test1.com',
   subject: 'test subject',
   html: '<div> Mail Content </div>',
   attachments: [
--- a/providers/ses/src/lib/ses.provider.spec.ts
+++ b/providers/ses/src/lib/ses.provider.spec.ts
@@ -93,7 +93,8 @@ test('should trigger ses library correctly', async () => {
   const provider = new SESEmailProvider(mockConfig);
   const response = await provider.sendMessage(mockNovuMessage);
 
-  const bufferArray = spy.mock.calls[0][0].input.RawMessage.Data;
+  // eslint-disable-next-line
+  const bufferArray = spy.mock.calls[0][0].input['RawMessage']['Data'];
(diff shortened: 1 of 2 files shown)


p-fernandez (commented):
I would suggest to add test for this addition for confidence.

providers/ses/src/lib/ses.provider.spec.ts
```diff
@@ -12,6 +12,7 @@ const mockConfig = {
 
 const mockNovuMessage = {
   to: ['test@test2.com'],
+  replyTo: 'test@test1.com',
```
p-fernandez, line 15: I think that this is not enough to test if `replyTo` is being configured in SES. 
In line 96 we should modify the spy to check that the proper params and values are passed by checking with the method `toHaveBeenCalledWith` and the values passed in the `mockNovuMessage`.
Unfortunately it wasn't done properly before but if not doing it we can't validate this change. 🙁

p-fernandez (changes requested):

deep-bhatt-pmc:
> I would suggest to add test for this addition for confidence.

btw. @p-fernandez  Not sure if you have been notified, but I have updated the PR, can you please check?

scopsy:
Looks great!

PR #3133, opened 2023-04-03
Index Cleanup - Job

djabarovgeorge:

Why? (Context)

### DIFF
--- a/libs/dal/src/repositories/job/job.repository.ts
+++ b/libs/dal/src/repositories/job/job.repository.ts
@@ -1,5 +1,5 @@
 import { ProjectionType } from 'mongoose';
-import { ChannelTypeEnum, StepTypeEnum } from '@novu/shared';
+import { StepTypeEnum } from '@novu/shared';
 
 import { BaseRepository } from '../base-repository';
 import { JobEntity, JobDBModel, JobStatusEnum } from './job.entity';
@@ -73,15 +73,6 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
         `There was a problem when trying to set an error for the job ${jobId} in the organization ${organizationId}`
       );
     }
-  }
-
-  public async findInAppsForDigest(organizationId: string, transactionId: string, subscriberId: string) {
-    return await this.find({
-      _organizationId: organizationId,
-      type: ChannelTypeEnum.IN_APP,
-      _subscriberId: subscriberId,
-      transactionId,
-    });
   }
 
   public async findJobsToDigest(from: Date, templateId: string, environmentId: string, subscriberId: string) {
--- a/libs/dal/src/repositories/job/job.schema.ts
+++ b/libs/dal/src/repositories/job/job.schema.ts
@@ -345,8 +345,7 @@ jobSchema.index({
 /*
  * This index was initially created to optimize:
  *
- * Path : apps/api/src/app/events/usecases/send-message/send-message<CHANNEL>.usecase.ts
- *    Context : The reason for this Index is because of
+ *    Context : The reason for this Index is that it used by the activity feed with populate,
  *              Notification scheme virtual localField: '_id', foreignField: '_notificationId', one to many
  */
 jobSchema.index({
--- a/libs/dal/src/repositories/job/job.schema.ts
+++ b/libs/dal/src/repositories/job/job.schema.ts
@@ -159,16 +159,6 @@ jobSchema.virtual('environment', {
  *           'step.uuid': filter.step,
  *        })
  *
- *
- */
-jobSchema.index({
-  transactionId: 1,
-  _subscriberId: 1,
-});
-
-/*
- * This index was initially created to optimize:
- *
  * Path : apps/api/src/app/events/usecases/trigger-event/trigger-event.usecase.ts
  *    Context : validateTransactionIdProperty()
  *       Query : findOne(
@@ -204,8 +194,7 @@ jobSchema.index({
  */
 jobSchema.index({
   transactionId: 1,
-  _environmentId: 1,
-  _id: 1,
+  _subscriberId: 1,
 });
 
 /*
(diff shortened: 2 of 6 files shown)


libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -75,15 +75,6 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
...
 
-  public async findInAppsForDigest(organizationId: string, transactionId: string, subscriberId: string) {
-    return await this.find({
-      _organizationId: organizationId,
-      type: ChannelTypeEnum.IN_APP,
-      _subscriberId: subscriberId,
-      transactionId,
-    });
-  }
-
```
djabarovgeorge, line 86: Removed because this query is not in use

libs/dal/src/repositories/job/job.schema.ts
djabarovgeorge, line 1: An initial proposal I still want to overview again and see if we can make it fewer indexes.

djabarovgeorge: Left the notes so ill have a ref on the desition making will update them once ill finish.

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -143,5 +134,80 @@ jobSchema.virtual('environment', {
...
+  _id: 1,
+});
+
+/*
+ * This index was initially created to optimize:
+ * apps/api/src/app/events/usecases/send-message/digest/get-digest-events-backoff.usecase.ts * updatedAt should be first range operator
+ * repo
+ * repo
+ * repo
+ * repo
```
p-fernandez, line 192:
```suggestion
 * repo
```

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -143,5 +134,80 @@ jobSchema.virtual('environment', {
...
+  type: 1,
+  status: 1,
+  updatedAt: 1,
+  transactionId: 1,
+});
+
+/*
+ * This index was initially created to optimize:
+ */
+jobSchema.index({});
```
p-fernandez, line 210: Is this still needed?

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -143,5 +134,80 @@ jobSchema.virtual('environment', {
   justOne: true,
 });
 
+/*
+ * This index was initially created to optimize:
+ * apps/api/src/app/events/usecases/add-job/add-delay-job.usecase.ts
+ */
```
p-fernandez, line 140: Leaving the explanations for the different index optimisations are very helpful. 🙌🏻

p-fernandez (approved):
🌟

libs/dal/src/repositories/job/job.schema.ts
scopsy: I think that even better would be to provide here some query examples? Because this comment is already not accurate as @LetItRock moved this file :P 

Having some query examples here stay accurate even when code moves.

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -143,5 +134,80 @@ jobSchema.virtual('environment', {
...
+jobSchema.index({
+  transactionId: 1,
+  _subscriberId: 1,
```
scopsy, line 143: Because of the cardinality of those 2 fields together I feel like we don't need all the other properties here of `_templateId` and `_environmentId` unless this could become a covered query. So having `transactionId, _subscriberId, status` should be enough in my opinion

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -143,5 +134,80 @@ jobSchema.virtual('environment', {
...
+  type: 1,
+});
+
+/*
+ * This index was initially created to optimize:
+ * apps/api/src/app/events/usecases/message-matcher/message-matcher.usecase.ts
+ */
+jobSchema.index({
+  transactionId: 1,
+  _subscriberId: 1,
```
scopsy, line 156: Following the top comment regarding the above query, I think this can be omitted?

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -180,18 +253,102 @@ jobSchema.index({
...
+ *          _environmentId: this.convertStringToObjectId(job._environmentId),
+ *          status: JobStatusEnum.DELAYED,
+ *          type: StepTypeEnum.DIGEST,
+ *          ...(digestKey && { [`payload.${digestKey}`]: digestValue }),
+ *        }
  */
 jobSchema.index({
-  _templateId: 1,
   _subscriberId: 1,
+  _templateId: 1,
```
djabarovgeorge: switched _subscriberId with _templateId because it has a higher cardinality

djabarovgeorge:
@scopsy 

what do you think of changing 
jobSchema.index({
  _subscriberId: 1,
  _templateId: 1,
  status: 1,
  type: 1,
  transactionId: 1,
});

 to 
 
 jobSchema.index({
  _subscriberId: 1,
  _templateId: 1,
  _environmentId : 1
  type: 1,
  transactionId: 1,
});



jobSchema.index({
  _subscriberId: 1,
  _templateId: 1,
  _environmentId: 1,
  type: 1,
  status: 1,
  updatedAt: 1,
  transactionId: 1,
});

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -143,5 +134,227 @@ jobSchema.virtual('environment', {
...
+ *       Query : find({
+ *          _subscriberId: this.convertStringToObjectId(job._subscriberId),
+ *          _templateId: job._templateId,
+ *          _environmentId: this.convertStringToObjectId(job._environmentId),
+ *          status: JobStatusEnum.DELAYED,
+ *          type: StepTypeEnum.DIGEST,
+ *          ...(digestKey && { [`payload.${digestKey}`]: digestValue }),
+ *        }
+ */
+jobSchema.index({
```
Cliftonz, line 349: I think this index may be "overrided" by the first query.
When you run this query, can you run https://www.mongodb.com/docs/manual/reference/method/cursor.explain/ to see which index the above query is running?

libs/dal/src/repositories/job/job.schema.ts
Cliftonz: I agree, the next index below is a copy of this one.
From our discussion with mongo the order of the fields by the esr rule is the most important. Mongo should be smart enough to use the below index if this one is removed.

libs/dal/src/repositories/job/job.schema.ts
djabarovgeorge: No removed this one.

scopsy:
@djabarovgeorge we also need to add a single key index to support _notificationId queries, this is used  by the activity feed populate from mongoose

PR #3136, opened 2023-04-04
fix(chat): Try all chat channels

ryshoooo:

Added a try-catch wrapper around sending chat messages for each channel.

In case the subscriber has multiple chat channels configured and 1 of them malfunctions, the current execution would halt and thus all the subsequent chat channels would not be executed. It might be that there is a temporary error with the first chat channel, but the others work fine. This PR changes this behavior and all chat channels are tried to be executed.

NA


apps/api/src/app/events/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -115,7 +115,14 @@ export class SendMessageChat extends SendMessageBase {
...
 
     for (const channel of chatChannels) {
-      await this.sendChannelMessage(command, channel, notification, chatChannel, content);
+      try {
+        await this.sendChannelMessage(command, channel, notification, chatChannel, content);
+      } catch (e) {
+        /*
+         * Do nothing, one chat channel failed, perhaps another one succeeds
+         * The failed message has been created
+         */
```
scopsy, line 124: @ryshoooo maybe we can add here an execution detail for failed delivery? Or maybe throw if all channels have failed? Because otherwise, the trigger will appear as delivered, when in reality it wasn't

p-fernandez: Also let's log the ones that fail. You can do it with:
```
const LOG_CONTEXT = 'SendMessageChat'; // Make this available in the top of the file.
...
Logger.error(`Sending chat message to the chat channel ${channel}`, LOG_CONTEXT);
```
You can import the `Logger` from `@nestjs/common` package.

ryshoooo: Updated :)

scopsy (approved):
🔥

jainpawan21:
@scopsy Can we merge this PR?

scopsy:
Yes @jainpawan21 feel free to merge when tests passing :)

PR #3143, opened 2023-04-04
Read pref optimizations

scopsy:
Add more read preferences for the mongo driver for some high thruput queries


apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -153,7 +153,6 @@ export class AuthService {
         organizationId: subscriber._organizationId,
         environmentId: subscriber._environmentId,
         subscriberId: subscriber.subscriberId,
-        organizationAdminId: userId,
```
scopsy, line 156: This one was not used, and we can remove un needed db query

ainouzgali (approved):
😍

PR #3156, opened 2023-04-06
fix: show termii sms provider in integration store

jainpawan21:

fix: termii sms provider in integration store

Closes [NV-2062](https://linear.app/novu/issue/NV-2062)


ainouzgali (approved):
Looks good @jainpawan21⭐️

PR #3159, opened 2023-04-06
chore(docs): change the the api docs domain for redirecting

andrewgolovanov:
This PR updates the API documentation domain to which we redirect the user when he visits the API link - `https://docs.novu.co/api/`

it will allow to give up Gatsby hosting and host the API documentation on Netlify

N/A


andrewgolovanov:
Hey @scopsy! Hope you’re doing great! 
I'm waiting to build a preview page to test the redirects work in more detail with the domain from Netlify

as we may have some problems with this
The idea is that for redirection to work, the documentation project (`https://docs.novu.co/`) should be in the same account as the API documentation project, then we can use the domain `https://novu-api-docs.netlify.app/ `to redirect to it. If not, we need to register a domain for the API documentation that is different from the base domain with the `.netlify` prefix

andrewgolovanov:
@scopsy To test this behavior, I put the documentation and the API documentation on the same account with completely identical settings - and it works

it works for example and testing:
[https://novudocs.netlify.app/](https://novudocs.netlify.app/)
[https://novudocs.netlify.app/api/overview/](https://novudocs.netlify.app/api/overview/)

so in our case, we have to register the domain for the API documentation to make the redirection work correctly

andrewgolovanov:
This PR requires adjustments and isn't ready for merger

scopsy (approved):

PR #3166, opened 2023-04-09
TTL - execution details, jobs and notifications

ainouzgali:

Calculate and add expireAt date at creation of a notification - add the expected delay to the expiration. And then take that expire date into every job and execution details that are created


apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -78,6 +88,7 @@ export class CreateNotificationJobs {
         digest: step.metadata,
         type: step.template.type,
         providerId: providerId,
+        expireAt: notification.expireAt,
```
ainouzgali, line 91: All jobs will expire at the same time of the notification they are related to

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -102,19 +113,6 @@ export class CreateNotificationJobs {
     return steps.filter((step) => step.active === true);
   }
 
-  private createTriggerStep(command: CreateNotificationJobsCommand): NotificationStepEntity {
```
ainouzgali, line 105: Wasn't used

apps/api/src/app/events/usecases/digest-filter-steps/digest-filter-steps-regular.usecase.ts
```diff
@@ -29,6 +29,7 @@ export class DigestFilterStepsRegular {
         {
           $set: {
             _digestedNotificationId: delayedDigests._notificationId,
+            expireAt: delayedDigests.expireAt,
```
ainouzgali, line 32: If notification has been digested into an existing one, will expire at the same date as the original one.

apps/api/src/app/execution-details/usecases/create-execution-details/create-execution-details.command.ts
```diff
@@ -72,6 +76,7 @@ export class CreateExecutionDetailsCommand extends EnvironmentWithSubscriber {
       providerId: job.providerId,
       transactionId: job.transactionId,
       channel: job.type,
+      expireAt: job.expireAt,
```
ainouzgali, line 79: Will add the expireAt of a job to all the related execution details

libs/dal/src/repositories/base-repository.ts
```diff
@@ -83,7 +83,11 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
   }
 
   private calcExpireDate(modelName: string, data: FilterQuery<T_DBModel> & T_Enforcement) {
-    const now: number = Date.now();
+    let now: Date = new Date();
```
scopsy, line 86: Maybe rename this to `startDate`? When we override it with expiredAt the "now" might be a bit misleading :)

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -101,10 +101,13 @@ const jobSchema = new Schema<JobDBModel>(
       type: Schema.Types.ObjectId,
       ref: 'Subscriber',
     },
+    expireAt: Schema.Types.Date,
   },
   schemaOptions
 );
 
+jobSchema.index({ expireAt: 1 }, { expires: '48h' });
```
scopsy, line 109: Maybe we can extract the 48H to a reusable variable so we can update all instances easily?

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -142,4 +140,51 @@ export class CreateNotificationJobs {
...
+    const delayedSteps = command.template.steps.filter(
+      (step) => step.template?.type === StepTypeEnum.DIGEST || step.template?.type === StepTypeEnum.DELAY
+    );
+
+    const delay = delayedSteps
+      .map((step) => this.calculateDelayAmount(step, command.payload, command.overrides))
+      .reduce((partialSum, a) => partialSum + a, 0);
+
+    return addMilliseconds(Date.now(), delay);
+  }
```
scopsy, line 154: @ainouzgali where does all of this functionality tested? would suggest extracting this to a class with public methods, and tests those public methods with unit tests, there is a lot of code here and would be great to test for it :)

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -142,4 +140,51 @@ export class CreateNotificationJobs {
...
   }
+
+  private calculateExpireAt(command: CreateNotificationJobsCommand) {
+    const delayedSteps = command.template.steps.filter(
+      (step) => step.template?.type === StepTypeEnum.DIGEST || step.template?.type === StepTypeEnum.DELAY
+    );
+
+    const delay = delayedSteps
+      .map((step) => this.calculateDelayAmount(step, command.payload, command.overrides))
+      .reduce((partialSum, a) => partialSum + a, 0);
```
scopsy, line 151: Would really suggest to avoid naming things with single letters:
```suggestion
      .reduce((sum, delayAmount) => sum + delayAmount, 0);
```

Or similar, this becomes easier to read, wdyt?

scopsy (commented):
@ainouzgali looks great, where does the tests for this functionality exists? Is it in a different PR?

apps/api/src/app/events/usecases/add-job/add-delay-job.usecase.ts
```diff
@@ -22,71 +21,6 @@ export class AddDelayJob {
...
-  }
-
-  /**
-   * To handle case of Scheduled Delay (triggered multiple times with the exact same future date) followed by a Digest.
-   * To avoid duplicate pending digests which would result in duplicate messages sent.
-   */
-  private async noExistingDelayedJobForDate(
-    data: JobEntity,
-    currentDelayPath: string,
-    currentDelayDate: string
```
ainouzgali, line 78: Was not needed anymore. It is not happening, thanks to Pablo

ainouzgali:
Working on tests now. Both for the TTL and for the new service @scopsy . But I might want to maybe push to dev before all tests are ready so we can try to integrate the online archive by tomorrow (to use the indexes)

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
ainouzgali: extracted it, this was duplicated from the logic in add-delay-job

apps/api/src/app/events/usecases/add-job/add-delay-job.usecase.ts
djabarovgeorge: I think we should remove it only if we are 100% sure this issue won't happen and we have a test that can simulate the possible error otherwise I think it could be risky.

apps/api/src/app/events/services/calculate-delay/delay.service.ts
```diff
@@ -0,0 +1,65 @@
+import { DigestUnitEnum, DelayTypeEnum } from '@novu/shared';
+
+import { Logger } from '@nestjs/common';
+import { ApiException } from '../../../shared/exceptions/api.exception';
+import { differenceInMilliseconds } from 'date-fns';
+import { NotificationStepEntity } from '@novu/dal';
+
+export class DelayService {
+  calculateDelay(step: NotificationStepEntity, payload: any, overrides: any): number {
```
djabarovgeorge, line 9: What do you think about wrapping the input here in an object?
In addition, get rid of the 'any' types.

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -142,4 +133,16 @@ export class CreateNotificationJobs {
 
     return steps;
   }
+
+  private calculateExpireAt(command: CreateNotificationJobsCommand) {
```
djabarovgeorge, line 137: I think it will be a good idea to make this one public or external function so we could add unit tests to it.

apps/api/src/app/events/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -142,4 +133,16 @@ export class CreateNotificationJobs {
...
+    const delayedSteps = command.template.steps.filter(
+      (step) => step.template?.type === StepTypeEnum.DIGEST || step.template?.type === StepTypeEnum.DELAY
+    );
+
+    const delay = delayedSteps
+      .map((step) => this.delayService.calculateDelay(step, command.payload, command.overrides))
+      .reduce((sum, delayAmount) => sum + delayAmount, 0);
```
djabarovgeorge, line 144: Is our business logic to make the expireAt the exact sum of all the delay/digest steps? we don't want to add any error rate time? (like network latency etc) Or TTL Variant as we have on the caching? 
Or do we have the TTL expiration logic somewhere else in the code that takes it to the account?

djabarovgeorge (commented):
Looks great added some comments 🤘

apps/api/src/app/events/usecases/add-job/add-delay-job.usecase.ts
ainouzgali: The tests created for that use case passed, and I did multiple manual checks to make sure. 
I agree, I'm not crazy about removing it, but it really seems to be resolved

libs/dal/src/repositories/base-repository.ts
```diff
@@ -87,14 +87,20 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
...
       case 'Message':
         if (data.channel === 'in_app') {
-          return addMonths(now, 6);
+          return addMonths(startDate, 6);
         } else {
-          return addMonths(now, 1);
+          return addMonths(startDate, 1);
         }
+      case 'Notification':
+        return addMonths(startDate, 1);
```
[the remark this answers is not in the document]
ainouzgali: Thats because, for jobs and execution details, I am taking the expireAt date of the notification. Which already has the month added. I want the notification, and all records that are related to it to have same expiration date.

apps/api/src/app/events/services/calculate-delay/delay.service.ts
```diff
@@ -0,0 +1,65 @@
+import { DigestUnitEnum, DelayTypeEnum } from '@novu/shared';
+
+import { Logger } from '@nestjs/common';
+import { ApiException } from '../../../shared/exceptions/api.exception';
+import { differenceInMilliseconds } from 'date-fns';
+import { NotificationStepEntity } from '@novu/dal';
+
+export class DelayService {
```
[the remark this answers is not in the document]
ainouzgali: Yes, CalculateDelayService is what I started with. You are right, I'll change it

apps/api/src/app/events/services/calculate-delay/delay.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    if (this.checkValidDelayOverride(overrides)) {
+      return this.toMilliseconds(overrides.delay.amount as number, overrides.delay.unit as DigestUnitEnum);
+    }
+
+    return this.toMilliseconds(step.metadata.amount as number, step.metadata.unit as DigestUnitEnum);
+  }
+
+  toMilliseconds(amount: number, unit: DigestUnitEnum): number {
```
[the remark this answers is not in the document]
ainouzgali: I was looking to remove the AddJob.toMilliseconds. Also from its uses in digest.. Thats why its not private and added into the service. But didn't get to it yet

libs/dal/src/repositories/base-repository.ts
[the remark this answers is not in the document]
LetItRock: ahhh I see now, it's hard to notice that because we have that logic hidden in the repo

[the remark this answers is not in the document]
ainouzgali: Yes🙈 Do you have a suggestion for me how I can make it clearer ?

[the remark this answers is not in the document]
LetItRock: not sure, maybe easier to read would be to have it in the `create-notification-jobs.usecase.ts` `calculateExpireAt` function instead of repo, because that the place where we calculate the base date, so I don't expect that one to be overriden?

ainouzgali:
Closing this one. Reopened in #3210

Issue #3022, opened 2023-03-15
[NV-1847] 🐛 Bug Report: Update API accepts null or '' as values but email is not unset


FabianGutierrezOpengov:
### 📜 Description

When using something like 

```
channel = await this.novu.subscribers.update(id, {
        email: null,
        phone: data.phoneNumber,
      });
```

Email doesn't change, but request is also not rejected.
      

### 👟 Reproduction steps

1) create subcriber

2) update subscriber with null or empty string

### 👍 Expected behavior

Either disallow null as value, or allow me to unset emails (I highly prefer the second one, as an user could want to stop getting notifications to their email)

### 👎 Actual Behavior with Screenshots

getting the previous email

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

<sub>[NV-1847](https://linear.app/novu/issue/NV-1847/🐛-bug-report-update-api-accepts-null-or-as-values-but-email-is-not)</sub>

jayavardhan3112:
Can you assign me this issue @FabianGutierrezOpengov

jainpawan21:
Sure @jayavardhan3112 
Assigning this to you!

jainpawan21:
Hi @jayavardhan3112 
You need help in this?

jayavardhan3112:
Yes, i tried hitting this API
`PUT: http://localhost:3000/v1/subscribers/:subscriberId`
`{
    "email": "",
    "phone": "2109269563"
}`

Getting the response - 

`
{
    "statusCode": 400,
    "message": [
        "email must be an email"
    ],
    "error": "Bad Request"
}
`
isnt this what we want ?

jayavardhan3112:
@jainpawan21 Please Review.

PR #3169, opened 2023-04-10
fix:  Update Subscribers API allowing null

jayavardhan3112:

Changed the update subscriber API, thus when there is a null or empty string, it will unset the email from the database

Closes [#3022](https://github.com/novuhq/novu/issues/3022)


scopsy (commented):
Because currently the endpoint behaving like a partial update endpoint with all fields this can lead to situation where the email is removed when updating other fields. Can we maybe perform a check like it was previously, but instead of `!= null` we can check for `!== undefined` so in case null was passed, we will update the field to `null` instead of unsettling it. wdyt?

jayavardhan3112:
> Because currently the endpoint behaving like a partial update endpoint with all fields this can lead to situation where the email is removed when updating other fields. Can we maybe perform a check like it was previously, but instead of `!= null` we can check for `!== undefined` so in case null was passed, we will update the field to `null` instead of unsettling it. wdyt?

So when a user passes an empty string on Null, then only the email would be unset, (The reason I unset the Email Key is what if its empty string , or Null, this would make the code to add redundant checks), thus I thought of unsetting the key,

jayavardhan3112:
> Because currently the endpoint behaving like a partial update endpoint with all fields this can lead to situation where the email is removed when updating other fields. Can we maybe perform a check like it was previously, but instead of `!= null` we can check for `!== undefined` so in case null was passed, we will update the field to `null` instead of unsettling it. wdyt?

So when a user passes an empty string on Null, then only the email would be unset, (The reason I unset the Email Key is what if its empty string , or Null, this would make the code to add redundant checks), thus I thought of unsetting the key,

jayavardhan3112:
> @jayavardhan3112 yes exactly, when it's undefined it should be untouched. I think setting it as null can help make queries for those users vs unset (meaning it was never set) similar to the semantic undefined vs null situation. Wdyt?

makes sense, I will update this and let you know, Thnx

p-fernandez (approved):
🌟 
@jayavardhan3112 please solve the conflicts so we can merge it. Sorry it took us so long. 🙏🏻

jayavardhan3112:
Updated, @p-fernandez . am not able to resolve them, could you please check the access

p-fernandez:
@jayavardhan3112 could you give me write permissions in your repository to push the conflict fix in your next branch? The PR was closed because I could only push Novu's `next` branch so the PR gets invalidated.

jayavardhan3112:
> @jayavardhan3112 could you give me write permissions in your repository to push the conflict fix in your next branch? The PR was closed because I could only push Novu's `next` branch so the PR gets invalidated.
> By the way don't worry as I have your PR fixed in a branch in my repository.

Hi @p-fernandez , Have invited you to collaborate to the PR

p-fernandez:
@jayavardhan3112 done. Can you do a quick review and confirm all your changes are there?

jayavardhan3112:
> @jayavardhan3112 done. Can you do a quick review and confirm all your changes are there?

Looks good to me, you can go ahead and merge it, Thank you

PR #3171, opened 2023-04-10
feat: remove yearly stats due to performance issues

scopsy:

This removes the yearly aggregation from the stats endpoint that causes performance issues when doing the queries against notification collection. 

For larger organizations this results in a very long operations, we will review our analytical approach there in the future, but for now disabling this to avoid bottlenecks.

### DIFF
--- a/libs/dal/src/repositories/notification/notification.repository.ts
+++ b/libs/dal/src/repositories/notification/notification.repository.ts
@@ -126,28 +126,33 @@ export class NotificationRepository extends BaseRepository<
   }
 
   async getActivityGraphStats(date: Date, environmentId: string) {
-    return await this.aggregate([
-      {
-        $match: {
-          createdAt: { $gte: date },
-          _environmentId: new Types.ObjectId(environmentId),
-        },
-      },
-      { $unwind: '$channels' },
-      {
-        $group: {
-          _id: {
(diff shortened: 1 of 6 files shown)


libs/dal/src/repositories/notification/notification.repository.ts
```diff
@@ -170,7 +169,6 @@ export class NotificationRepository extends BaseRepository<
           _id: null,
           weekly: { $sum: { $cond: [{ $gte: ['$createdAt', weekBefore] }, 1, 0] } },
           monthly: { $sum: { $cond: [{ $gte: ['$createdAt', monthBefore] }, 1, 0] } },
```
djabarovgeorge, line 171:
```suggestion
          monthly: { $sum: 1 },
```

I think we could do something like this in the month now.

djabarovgeorge (approved):
🚀

djabarovgeorge:
@scopsy what do you think about moving this one to secondary?

libs/dal/src/repositories/notification/notification.repository.ts
scopsy: Good point!

scopsy:
@djabarovgeorge aggree here, pushing it now 🙏

PR #3174, opened 2023-04-11
Index Cleanup - Notification

djabarovgeorge:

Why? (Context)

### DIFF
--- a/libs/dal/src/repositories/notification/notification.schema.ts
+++ b/libs/dal/src/repositories/notification/notification.schema.ts
@@ -82,6 +82,78 @@ notificationSchema.virtual('jobs', {
   foreignField: '_notificationId',
 });
 
+/*
+ *
+ * Path: libs/dal/src/repositories/notification/notification.repository.ts
+ *    Context: findBySubscriberId()
+ *        Query: find({_environmentId: environmentId,
+ *                    _subscriberId: subscriberId,});
+ *
+ */
+notificationSchema.index({
+  _subscriberId: 1,
+  _environmentId: 1,
+});
+
+/*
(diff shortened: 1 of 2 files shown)


libs/dal/src/repositories/notification/notification.schema.ts
```diff
@@ -82,6 +82,78 @@ notificationSchema.virtual('jobs', {
...
+ * Path: libs/dal/src/repositories/notification/notification.repository.ts
+ *    Context: findBySubscriberId()
+ *        Query: find({_environmentId: environmentId,
+ *                    _subscriberId: subscriberId,});
+ *
+ */
+notificationSchema.index({
+  _subscriberId: 1,
+  _environmentId: 1,
+});
```
p-fernandez, line 96: Just a doubt, the same as in the other index cleanup PR. Is not this index included in the following one?

djabarovgeorge: They are including one another but the order matters and the prefix of the index make one redundant. 
for example 

{
  _subscriberId: 1,
  _environmentId: 1,
}

{
  transactionId: 1,
  _subscriberId: 1,
  _environmentId: 1,
}

are not the same, 
but the following example has the same prefix so we can remove the first one.

{
  _subscriberId: 1,
  _environmentId: 1,
}

{
  _subscriberId: 1,
  _environmentId: 1,
    transactionId: 1,
}

Hopefully, i got your question right and my answer helped 😅

p-fernandez: Yes, it does. I forgot that the order matters. Thanks for explaining it to me again!

p-fernandez (approved):
🌟

PR #3177, opened 2023-04-11
nv-2054-implement-redis-cluster-service

p-fernandez:

Adds two in memory cluster providers (Redis based configuration and Elasticache provider; both using Redis Cluster downstream) and connects it to the API, enabling Cache Service and Distributed Lock service to take advantage of a Redis Cluster instance for improving the performance.
It also adds the tests and the CI/CD changes needed for those to run. 

One of the performance improvements we are executing.

Several environment variables need to be added and set depending on the different environments configuration.
For DEV environment we would like to set:
IN_MEMORY_CLUSTER_MODE_ENABLED=true
ELASTICACHE_CLUSTER_SERVICE_HOST=<ElastiCache_URL>
ELASTICACHE_CLUSTER_SERVICE_PORT=<ElastiCache_PORT>

For CI/CD and tests environments we would like to set:
IN_MEMORY_CLUSTER_MODE_ENABLED=true
REDIS_CLUSTER_SERVICE_HOST=localhost
REDIS_CLUSTER_SERVICE_PORTS=[7000,7001,7002,7003,7004,7005]


.github/actions/setup-redis-cluster/action.yml
```diff
@@ -0,0 +1,15 @@
+name: Setup Novu Redis Cluster
```
p-fernandez, line 1: Action to create a Redis Cluster to be used by the CI/CD in the tests.

apps/api/src/.example.env
```diff
@@ -11,7 +11,7 @@ REDIS_PREFIX=
 REDIS_HOST=localhost
 REDIS_DB_INDEX=2
 
-REDIS_CACHE_SERVICE_HOST=
+REDIS_CACHE_SERVICE_HOST=localhost
 REDIS_CACHE_SERVICE_PORT=6379
```
scopsy, line 15: Would prefer if by default cache will be disabled locally, wdyt?

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.spec.ts
```diff
@@ -10,6 +10,8 @@ import { DistributedLockService } from './distributed-lock.service';
 
 const originalRedisCacheServiceHost = process.env.REDIS_CACHE_SERVICE_HOST;
 const originalRedisCacheServicePort = process.env.REDIS_CACHE_SERVICE_PORT;
+const originalRedisClusterServiceHost = process.env.REDIS_CLUSTER_SERVICE_HOST;
+const originalRedisClusterServicePorts = process.env.REDIS_CLUSTER_SERVICE_PORTS;
```
scopsy, line 14: Should this be?
```suggestion
const originalRedisClusterServicePorts = process.env.REDIS_CLUSTER_SERVICE_PORT;
```

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -1,41 +1,156 @@
...
+      inMemoryProviderClient.on('end', () => {
+        Logger.warn('In-memory cluster end', LOG_CONTEXT);
+      });
+
+      inMemoryProviderClient.on('error', (error) => {
+        Logger.error(error, LOG_CONTEXT);
+      });
+
+      inMemoryProviderClient.on('ready', () => {
+        Logger.warn('In-memory cluster ready', LOG_CONTEXT);
```
scopsy, line 142:
```suggestion
        Logger.log('In-memory cluster ready', LOG_CONTEXT);
```

I think better suited as it's a success one

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -1,41 +1,156 @@
...
     Logger.verbose('In-memory cluster service set up', LOG_CONTEXT);
 
-    return {} as InMemoryProviderClient;
+    const { getConfig, getClient } = this.getClientAndConfigForCluster();
+
+    this.inMemoryProviderConfig = getConfig();
+    const { host, ttl } = getConfig();
+
+    if (!host) {
+      Logger.log('Missing host for in-memory cluster provider', LOG_CONTEXT);
```
scopsy, line 110:
```suggestion
      Logger.warn('Missing host for in-memory cluster provider', LOG_CONTEXT);
```

scopsy (approved):

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -1,41 +1,156 @@
...
+    return process.env.IN_MEMORY_CLUSTER_MODE_ENABLED === 'true';
+  }
+
+  public getClusterOptions(): ClusterOptions | undefined {
+    if (this.inMemoryProviderClient && this.isClusterMode()) {
+      return this.inMemoryProviderClient.options;
+    }
+  }
+
+  public getOptions(): RedisOptions | undefined {
```
p-fernandez, line 63: Downstream the Redis Cluster can have Redis Options set up and configured just like a single instance, but they live in a different place.

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -1,41 +1,154 @@
...
+        getClient: getElasticacheCluster,
+        getConfig: getElasticacheClusterProviderConfig,
+      },
+      redis: {
+        getClient: getRedisCluster,
+        getConfig: getRedisClusterProviderConfig,
+      },
+    };
+
+    return this.isElasticacheEnabled() ? clusterProviders.elasticache : clusterProviders.redis;
```
p-fernandez, line 96: If Elasticache configuration is set in the environment we use that provider and if not we fall back to the Redis Cluster provider that needs automated setting in the CI/CD and in your local environment when running the tests. If the Cluster mode is enabled in the configuration through `IN_MEMORY_CLUSTER_MODE_ENABLED`.

apps/api/src/config/env-validator.ts
```diff
@@ -50,11 +50,17 @@ const validators: { [K in keyof any]: ValidatorSpec<any[K]> } = {
...
   }),
   REDIS_CACHE_SERVICE_TLS: json({
     default: undefined,
   }),
+  REDIS_CLUSTER_SERVICE_HOST: str({
+    default: '',
+  }),
+  REDIS_CLUSTER_SERVICE_PORTS: str({
+    default: '',
+  }),
```
p-fernandez, line 63: Not sure if we should validate the Elasticache variables as they are optional and depending on the environment deployed.

libs/testing/src/user.session.ts
```diff
@@ -36,16 +36,18 @@ import { UserService } from './user.service';
 /**
  * TODO: move this to a reusable area
  */
+const connection = {
+  db: Number(process.env.REDIS_DB_INDEX || '1'),
+  port: Number(process.env.REDIS_PORT || 6379),
```
p-fernandez, line 41: This and other hardcodes regarding the port need to be addressed in another PR.

apps/api/src/app/shared/services/distributed-lock/distributed-lock.service.spec.ts
p-fernandez: No, the name is right as this is for the case where we provide the Redis Cluster with the ports nodes into an array. Nothing to do with the Elasticache cluster set up. This is the fallback.

apps/api/src/app/shared/services/in-memory-provider/in-memory-provider.service.ts
p-fernandez: I set it up higher as the readiness was being a problem because my infra problems. I can revert it back. 👍🏻

apps/api/src/.env.test
```diff
@@ -22,6 +22,19 @@ REDIS_CACHE_KEEP_ALIVE=
 REDIS_CACHE_FAMILY=
 REDIS_CACHE_KEY_PREFIX=
 
+IN_MEMORY_CLUSTER_MODE_ENABLED=true
+ELASTICACHE_CLUSTER_SERVICE_HOST=
+ELASTICACHE_CLUSTER_SERVICE_PORT=
+REDIS_CLUSTER_SERVICE_HOST=localhost
+REDIS_CLUSTER_SERVICE_PORTS=[7000,7001,7002,7003,7004,7005]
```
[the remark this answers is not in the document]
p-fernandez: We can do that just by setting: `IN_MEMORY_CLUSTER_MODE_ENABLED` to `false`. Just want to highlight that `in-memory-provider.service.spec.ts` won't pass.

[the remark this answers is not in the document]
scopsy: Sure, we can do that

apps/api/src/app/shared/services/in-memory-provider/elasticache-cluster-provider.ts
```diff
@@ -84,15 +85,17 @@ export const getElasticacheCluster = (): Cluster | undefined => {
...
     // scaleReads: 'slave' as NodeRole, // Enable it for improved performance to read only in replicas
     /*
      *  Disabled in Prod as affects performance
      */
     showFriendlyErrorStack: process.env.NODE_ENV !== 'prod',
-    slotsRefreshTimeout: 5000,
+    slotsRefreshTimeout: 10000,
   };
 
+  Logger.log(`Initializing Elasticache Cluster with ${instances?.length} instances`);
```
p-fernandez, line 98: This is not going to count the real instances as the set up for Elasticache is different. Not sure if through the URL or any API call we would know the exact number.

apps/api/src/app/shared/services/in-memory-provider/redis-cluster-provider.ts
```diff
@@ -86,6 +87,8 @@ export const getRedisCluster = (): Cluster | undefined => {
     slotsRefreshTimeout: 2000,
   };
 
+  Logger.log(`Initializing Redis Cluster Provider with ${instances?.length} instances`);
```
p-fernandez, line 90: This is ok. 👍🏻

scopsy (approved):

.github/workflows/reusable-api-e2e.yml
```diff
@@ -58,9 +58,17 @@ jobs:
...
         run: CI='' pnpm build:api
 
-      # Runs a set of commands using the runners shell
-      - name: Run a test
+      - name: Run E2E tests
+        env:
+          IN_MEMORY_CLUSTER_MODE_ENABLED: true
         run: |
           cd apps/api && pnpm test:e2e
           pnpm test
```
LetItRock, line 66: this like will run unit tests

scopsy: Ooops forgot to remove it

PR #3193, opened 2023-04-13
feat: use plain responses from mongoose to improve performance

scopsy:

Uses lean on the dal level to avoid mongoose performance overhead

there is a performance overhead to create mongoose object, since we anyway create a POJO out of it, we can eliminate the necessary conversion to it in the first place


p-fernandez (approved):
Probably a tech debt task of moving the DALs to use the base repository implementations rather than using the Mongoose models functionalities would have made these changes smaller and only in the base repository. 🙁

scopsy:
@p-fernandez totally aggree, we just need to expose a more flexible API on the dal with to support optional populates and etc...

djabarovgeorge (approved):
What @p-fernandez said.

PR #3194, opened 2023-04-13
add ttl index configuration 

ainouzgali:

- No ttl expiration for managed users 
- add env variable to allow self hosted users to disable ttl


libs/dal/src/shared/consts/ttl.ts
```diff
@@ -1 +1,9 @@
 export const TTL_EXPIRE_AFTER_AMOUNT = '48h';
+
+export const TTL_INDEX_ENABLED = !(process.env.NOVU_MANAGED_SERVICE === 'true' || process.env.DISABLE_TTL === 'true');
```
LetItRock, line 3:
```suggestion
export const TTL_INDEX_ENABLED = process.env.NOVU_MANAGED_SERVICE === 'false' && process.env.DISABLE_TTL === 'false';
```

LetItRock (approved):

libs/dal/src/shared/consts/ttl.ts
ainouzgali: I want to allow both of those env variables to be undefined. That only of its actually there and true. Thats why I went that ! route :)

PR #3195, opened 2023-04-13
perf: improve auth flow performance

scopsy:

### DIFF
--- a/apps/api/src/app/auth/framework/auth.guard.ts
+++ b/apps/api/src/app/auth/framework/auth.guard.ts
@@ -29,7 +29,7 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
       return this.authService.apiKeyAuthenticate(key).then<any>((result) => {
         request.headers.authorization = `Bearer ${result}`;
 
-        return super.canActivate(context);
+        return true;
       });
     }
 
--- a/apps/api/src/app/auth/services/auth.service.ts
+++ b/apps/api/src/app/auth/services/auth.service.ts
@@ -11,7 +11,7 @@ import {
   UserRepository,
 } from '@novu/dal';
 import { AuthProviderEnum, IJwtPayload, ISubscriberJwt, MemberRoleEnum, SignUpOriginEnum } from '@novu/shared';
-import { AnalyticsService, Instrument } from '@novu/application-generic';
+import { AnalyticsService, Instrument, PinoLogger } from '@novu/application-generic';
 
 import { CreateUserCommand } from '../../user/usecases/create-user/create-user.dto';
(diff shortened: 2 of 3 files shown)


apps/api/src/app/auth/framework/auth.guard.ts
```diff
@@ -29,7 +29,7 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
       return this.authService.apiKeyAuthenticate(key).then<any>((result) => {
         request.headers.authorization = `Bearer ${result}`;
 
-        return super.canActivate(context);
+        return true;
```
scopsy, line 32: No need to validate again since we already validated the token above

libs/dal/src/repositories/member/member.repository.ts
```diff
@@ -161,10 +161,16 @@ export class MemberRepository extends BaseRepository<MemberDBModel, MemberEntity
...
+    return !!(await this.findOne(
+      {
+        _organizationId: organizationId,
+        _userId: userId,
+      },
+      '_id',
+      {
+        readPreference: 'secondaryPreferred',
+      }
+    ));
```
djabarovgeorge, line 173: Noice!

apps/api/src/app/auth/framework/auth.guard.ts
djabarovgeorge: Can you elaborate where exactly? I think I am missing it.

djabarovgeorge (approved):
Looks good to me, i had just one question there

scopsy:
@djabarovgeorge so basically the super method was calling the jwt.strategy validate function behind the scenes, and we already validated the API token during the apiKeyAuthenticate method, so a double validation was performed.

apps/api/src/app/auth/framework/auth.guard.ts
LetItRock: in the `apiKeyAuthenticate` function

apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -259,12 +266,16 @@ export class AuthService {
 
   @Instrument()
   async validateUser(payload: IJwtPayload): Promise<UserEntity> {
-    const user = await this.getUser({ _id: payload._id });
-    if (!user) throw new UnauthorizedException('User not found');
+    // We run these in parallel to speed up the query time
+    const userPromise = this.getUser({ _id: payload._id });
+    const isMemberPromise = payload.organizationId
```
LetItRock, line 271: wondering when there is no `organizationId`? after the registration?

scopsy: Yep exactly, before the user creates the actual organization

PR #3199, opened 2023-04-14
refactor: remove duplicate cache service initialization

scopsy:

Previously the caching service was initialized 3 times because it was injected another un-needed time.

I'm concerned that we have a split connection for some of the redis calls.

### DIFF
--- a/apps/api/src/app/shared/shared.module.ts
+++ b/apps/api/src/app/shared/shared.module.ts
@@ -102,7 +102,6 @@ const distributedLockService = {
 };
 
 const PROVIDERS = [
-  inMemoryProviderService,
   cacheService,
   distributedLockService,
   {
--- a/apps/api/src/app/events/services/distributed-lock-service/index.ts
+++ b/apps/api/src/app/events/services/distributed-lock-service/index.ts
@@ -5,11 +5,7 @@ import { InMemoryProviderService } from '../../../shared/services/in-memory-prov
 
 @Injectable()
 export class EventsDistributedLockService {
-  private readonly distributedLockService;
-
-  constructor(private inMemoryProviderService: InMemoryProviderService) {
-    this.distributedLockService = new DistributedLockService(inMemoryProviderService);
-  }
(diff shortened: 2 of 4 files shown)


scopsy:
@p-fernandez is there any specific reason for the previous implementation? Or is it okey if we change it to this?

p-fernandez:
> @p-fernandez is there any specific reason for the previous implementation? Or is it okey if we change it to this?

The current implementation was to follow our instantiation pattern and the way I quickly found the less times the services were initialised.
What I do not understand from the PR is that there is no Cache Service changes related even if that is the goal of the PR.

apps/api/src/app/shared/shared.module.ts
```diff
@@ -102,7 +102,6 @@ const distributedLockService = {
 };
 
 const PROVIDERS = [
-  inMemoryProviderService,
```
LetItRock, line 105: @scopsy @p-fernandez 
I'm wondering why can't we just use the DI mechanism to inject the `InMemoryProviderService` to the `CacheService` and `DistributedLockService`?... we have here above the provider object and we do call `useFactory` two times which creates two instances of `InMemoryProviderService`...

LetItRock:
> > @p-fernandez is there any specific reason for the previous implementation? Or is it okey if we change it to this?
> 
> The current implementation was to follow our instantiation pattern and the way I quickly found the less times the services were initialised. What I do not understand from the PR is that there is no Cache Service changes related even if that is the goal of the PR.

the `CacheService` is initialized in the shared module where we do pass the instance on `InMemoryProviderService`

p-fernandez:
> the `CacheService` is initialized in the shared module where we do pass the instance on `InMemoryProviderService`

I might miss any nuance but I do not think these changes would remove the double initialisation. Unless the goal of the task is to reduce the initialisations of the InMemoryProviderService to create less instances of the ioredis client downstream, that indeed is happening.

p-fernandez (approved):

PR #3207, opened 2023-04-16
fix: missing email override

djabarovgeorge:

Update the override to be taken from override and not override[providerId].

Not sure if it was an intended/bug or feature but we need the update in order for us to be provider agnostic on the email override object.


apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -124,7 +124,7 @@ export class SendMessageEmail extends SendMessageBase {
       return;
     }
 
-    const overrides: Record<string, any> = command.overrides[integration?.providerId] || {};
```
scopsy, line 127: Previously this was the actual `overrides` object here, instead of the global overrides. Maybe we can optionally merge both providerId overrides and `email` overrides to a unified override? 

The providerId should overtak the "email" once.

And then we can use this object instead of command.overrides downstream. Wdyt?

djabarovgeorge: I think we could, one sec i will add the additional provider to it.

apps/api/src/app/events/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -124,7 +124,11 @@ export class SendMessageEmail extends SendMessageBase {
...
+    const overrides: Record<string, any> = Object.assign(
+      {},
+      command.overrides,
+      command.overrides[integration?.providerId]
```
scopsy, line 130:
```suggestion
      command.overrides['email'] || {},
      command.overrides[integration?.providerId] || {}
```

djabarovgeorge: if we will add '|| {}' then only the last one will merge and it will result in: 
command.overrides[integration?.providerId] OR {}
The previous one merged all the entities when the last one is the most will override the first one.

scopsy (approved):

scopsy:
@djabarovgeorge could you merge this from next with Pawel's work? this file has been moved

scopsy (approved):

PR #3209, opened 2023-04-16
fix: dont store deleted template that are missing in next env(prod)

djabarovgeorge:

do not store templates that were in dev and then were deleted. (without being promoted to prod before)

At the moment if the user will create a template work/edit it (without promoting to prod) then decides he does not need it in prod and delete it. once he will promote it for the first time a new template will be created (not deleted)

In the beginning, i created a solution where we store the templated but in the deleted state, but then I thought about we will probably not want to store templates that were not in prod env.

@scopsy  LMKWYT  maybe we do want to store then :)


p-fernandez (approved):
🌟

scopsy:
@djabarovgeorge Can we merge this?

djabarovgeorge:
> @djabarovgeorge Can we merge this?

Definitely just was wondering if we want to archive this template state as deleted prod.

PR #3215, opened 2023-04-17
Added Plausible

nevo-david:
Added plausible to Novu

We have it on our website but not on our dashboard


scopsy:
@nevo-david seems like all our test suite is failing regarding process is not defined, we might need to check the type of it prior to accessing it

apps/web/.env
```diff
@@ -10,4 +10,4 @@ REACT_APP_INTERCOM_APP_ID=
 REACT_APP_SEGMENT_KEY=
 REACT_APP_SENTRY_DSN=
 LOGROCKET_ID=
-
+REACT_APP_PLAUSIBLE_DOMAIN=
```
scopsy, line 13: @nevo-david what this value should be in prod?

PR #3226, opened 2023-04-17
Inject CONTEXT_PATH variables defined in .env to images for reverse proxy routing

RtiM0:

The proposed modification aims to inject the *CONTEXT_PATH variables defined in the .env file into the images.

The deployment docker-compose file was not including the environment variables required to set the CONTEXT_PATH feature, which created a problem for those who wanted to use routes with a reverse proxy such as Nginx instead of having separate subdomains for each app. The proposed modification addresses this issue by ensuring that the necessary environment variables are included in the deployment file.

https://discord.com/channels/895029566685462578/1097448355237462017


p-fernandez (commented):
We might need also to add context path for the new Worker app we are splitting from the API app.
CC: @scopsy @Cliftonz @LetItRock I found this ok but might need to wait for the Worker app.

LetItRock (approved):

PR #3233, opened 2023-04-18
feat(infra): allow to enable autopipelining in cache service via ff

p-fernandez:

Adds feature flag to temporarily enable auto-pipelining in Redis Cluster mode for the Cache Service.

We expect improvements in the management of the commandQueue of ioredis when enabling this feature, but we want to test it controlledly and load test it before moving into production, thus this temp implementation.


apps/api/src/app/shared/shared.module.ts
```diff
@@ -78,15 +78,17 @@ const dalService = new DalService();
 
 const inMemoryProviderService = {
   provide: InMemoryProviderService,
-  useFactory: () => {
-    return new InMemoryProviderService();
+  useFactory: (enableAutoPipelining?: boolean) => {
+    return new InMemoryProviderService(enableAutoPipelining);
```
scopsy, line 82: @p-fernandez maybe worth doing an object here? In case we will want to control other aspects in the future. Wdyt?

scopsy (approved):

apps/api/src/app/shared/shared.module.ts
p-fernandez: Is this going to be permanent? I was expecting this to be temporary and we would enable it for Cache and DistributedLock after testing for Cache.

scopsy: I think it could become, so we can reuse this service in the future with different hosts and etc... Not a must, but was just thinkinof this for the future. You can leave it as sis for now

p-fernandez: Let's try it quick and then I can clean it if it works for our purpose and we decide it will stay.

scopsy:
@p-fernandez could you ignor the .env.test files for the cspell? And let's merge this 🙏

PR #3236, opened 2023-04-19
feat: switch to using production node env instead of env

scopsy:

Switched `prod` to `production` when specified in the `NODE_ENV` as the optimizations only happened when `production` is set not `prod`. 

Switching the ﻿NODE_ENV environment variable to “production” in Node.js is important for performance because it enables various optimizations within the Node.js runtime and third-party libraries designed to improve performance in production environments. For example, it may disable certain debugging features, enable caching of compiled templates, and enable more aggressive garbage collection. Some libraries may behave differently or provide additional optimizations when running in production mode.


LetItRock (approved):
🥇

PR #3237, opened 2023-04-19
refactor: update reactflow

davidsoderberg:
Updates react flow to latest version.

Because we have not updated it since we installed it first time.


p-fernandez (approved):
👏🏻

scopsy (approved):
🔥

LetItRock:
@davidsoderberg @djabarovgeorge 
the `useStartFromScratchTour` hook has a selector for a DOM node with `[class="react-flow__controls"]` but because we've updated the library there is no such DOM node anymore... we should update it to `[data-testid="rf__controls"]` or something else, because I'm not sure if it's ok to rely on the library `data-testid` attribute

davidsoderberg:
Closing this for now because a lot of stuff have happend since April and it is easier to do a new PR instead.

PR #3238, opened 2023-04-19
fix(api): improve the flow of mark message as use case for performance

p-fernandez:

Parallelises 2 calls mutually independent for quicker execution and avoids an unnecessary loop if user is not admin.

Noticed this flow improvement when analysing calls pre yesterday's deployment where calls to `markAs` endpoint were quite slow. The biggest improvement has come from our infrastructure changes but decided to add this small thing as an extra.


LetItRock (approved):
🎖️

PR #3239, opened 2023-04-19
Activity Feed: show subscriberId instead of novu's internal id 

ainouzgali:
Fix - show the user's defined id for the subscriber (subscriberId). Currently in Activity Feed we are showing novu's id for that subscriber(subscriber._id). 

Screen Shot 2023-04-19 at 13 31 20


scopsy (approved):
🔥

PR #3250, opened 2023-04-20
feat: speed up eslint parser timing

scopsy:
This PR introduces a performance improvement to eslint by localizing the ts parser options to the scanned object rather than the whole repo from the root. This change resulted in a significant reduction in test case running time, from 20s to 5s. Additionally, cspell has been removed from eslint, as it caused more than 70% of the time spent when linting.

This change was needed to improve the performance of eslint, which was previously slow due to the ts parser options being applied to the entire repo from the root. This caused husky pre commit take a forever amount of time to complete

Before:
CleanShot 2023-04-20 at 09 40 04@2x

After:
CleanShot 2023-04-20 at 09 39 35@2x

Cspell took almost all the time, now we will only run it in CI. 
CleanShot 2023-04-20 at 09 42 51@2x


.eslintrc.js
```diff
@@ -7,7 +7,6 @@ module.exports = {
     'prettier',
     'plugin:prettier/recommended',
     'plugin:promise/recommended',
-    'plugin:@cspell/recommended',
```
p-fernandez, line 10: You should remove .cspell.json file too. Or are we keeping it in the CI?

apps/api/src/app/invites/usecases/invite-member/invite-member.usecase.ts
```diff
@@ -37,7 +37,6 @@ export class InviteMember {
     if (process.env.NOVU_API_KEY && (process.env.NODE_ENV === 'dev' || process.env.NODE_ENV === 'production')) {
       const novu = new Novu(process.env.NOVU_API_KEY);
 
-      // eslint-disable-next-line @cspell/spellchecker
       // cspell:disable-next
```
p-fernandez, line 40: You should remove all the ocurrences of cspell anotations.

.eslintrc.js
scopsy: @p-fernandez keeping it in CI, just making sure locally our linter will run faster

apps/api/src/app/invites/usecases/invite-member/invite-member.usecase.ts
scopsy: The second annotation helps the CI skip this one. I only removed the eslint one here

PR #3259, opened 2023-04-20
Fixing multiple push providers issue NV-2156

### LINKED ISSUE #3258: [NV-2156] 🐛 Bug Report:  When using multiple push channels, it always get the first integration
### 📜 Description

When having 2 push providers active (I'm working on OneSignal and a Webhook for Push), then both get triggered with the same provider instead of the correct one. 

The credentials are the correct one, but the provider is incorrect. What helps in the future is adding the provider on the screens on the screenshots.

### 👟 Reproduction steps

Activate 2 push providers, then set channels for both on a subscriber and send a push notification.

### 👍 Expected behavior

It should trigger the correct provider.

First: image

Second: image

I checked, it seems to always provide me with the first push integration's providerId.

<sub>[NV-2156](https://linear.app/novu/issue/NV-2156/🐛-bug-report)</sub>

TNAJanssen:

It produces a way to get the correct integration per channel

Closes #3258

### Screenshots

image


apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -153,48 +126,74 @@ export class SendMessagePush extends SendMessageBase {
...
+        channel.providerId
+      );
+    }
+
+    if (sent) {
       return;
     }
 
-    await this.sendErrors(pushChannels, command, notification, integration);
+    await this.sendErrors(pushChannels, command, notification);
```
p-fernandez, line 193: This is a good catch.

p-fernandez, line 193: This is a good catch.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -153,48 +126,74 @@ export class SendMessagePush extends SendMessageBase {
...
+      await this.sendMessage(
+        integration,
+        channel.credentials.deviceTokens,
+        title,
+        content,
+        command,
+        notification,
+        command.payload,
+        overrides,
+        channel.providerId
```
p-fernandez, line 185: As you headed to get the `providerId` from `integration` inside the method, we can remove this argument and in line 238 set the `providerId` property from the value of `integration.providerId`.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
p-fernandez: But I noticed that the functionality implemented here is the same as it is happening in lines 114-124.
I'd remove from line 115 to 124 and replace for this call `await this.sendErrors` and early return.
This will help us to remove the `flag` you added for the loop in line 129, that I find it a bit useless.

p-fernandez (changes requested):
Good addition. Improves UX in Execution Details.
I think we can do some code improvements while we are there.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
TNAJanssen: Done, just need to push

TNAJanssen:
I changed the code as you suggested, i have tested this change on a test environment and it works really well.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -112,21 +112,11 @@ export class SendMessagePush extends SendMessageBase {
...
-        CreateExecutionDetailsCommand.create({
-          ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
-          detail: DetailEnum.SUBSCRIBER_NO_ACTIVE_CHANNEL,
-          source: ExecutionDetailsSourceEnum.INTERNAL,
-          status: ExecutionDetailsStatusEnum.FAILED,
-          isTest: false,
-          isRetry: false,
-        })
-      );
+      await this.sendErrors(pushChannels, command, notification);
```
p-fernandez, line 115: Forgot to mention we can avoid to pass `pushChannels` to this method as we are doing the check before and it is not needed to do it inside.

p-fernandez (approved):
🌟

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
TNAJanssen: @p-fernandez done, pushing in a moment

SerafeimOikonomou:
Could anyone summarize how the process will work for multiple providers from now on please?

PR #3272, opened 2023-04-22
Add message count limit

djabarovgeorge:

In this PR we introduce the count limit param.

We need this additional limit query option, in order to reduce the number of scanned objects/returned on MongoDB.


apps/api/src/app/widgets/dtos/message-response.dto.ts
```diff
@@ -201,6 +201,9 @@ export class MessagesResponseDto {
   @ApiProperty()
   totalCount: number;
 
+  @ApiProperty()
+  hasMore: boolean;
```
LetItRock, line 205: missing tests for this flag

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    res = pipe.transform(limit, metadata);
+    expect(res).to.equal(limit);
+
+    limit = 1000;
+    res = pipe.transform(limit, metadata);
+    expect(res).to.equal(limit);
+
+    limit = 1001;
+    expect(() => pipe.transform(limit, metadata)).to.throw(`${METADATA.DATA} must not be greater than 1000`);
+  });
```
LetItRock, line 47: I think its better to split this test into smaller tests, like with the `it.each`` `, so when a few cases will fail you would know which tests to fix... same about the below code

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
```diff
@@ -0,0 +1,36 @@
+import { ArgumentMetadata, BadRequestException, Injectable, PipeTransform } from '@nestjs/common';
+
+@Injectable()
+export class LimitPipe implements PipeTransform {
+  private readonly minInt: number;
+  private readonly maxInt: number;
+  private readonly isOptional: boolean = false;
```
LetItRock, line 7: please add empty line between variables and constructor, and constructor and transform function

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
```diff
@@ -0,0 +1,36 @@
+import { ArgumentMetadata, BadRequestException, Injectable, PipeTransform } from '@nestjs/common';
+
+@Injectable()
```
LetItRock, line 3: does it need to be injectable?

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
```diff
@@ -0,0 +1,36 @@
...
+export class LimitPipe implements PipeTransform {
+  private readonly minInt: number;
+  private readonly maxInt: number;
+  private readonly isOptional: boolean = false;
+  constructor(minInt: number, maxInt: number, isOptional?: boolean) {
+    this.minInt = minInt;
+    this.maxInt = maxInt;
+    if (isOptional != null) {
+      this.isOptional = isOptional;
+    }
```
LetItRock, line 13: if you will set the default value on the constructor argument you won't need to have this check...

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+  transform(value: number | undefined, metadata: ArgumentMetadata) {
+    if (this.isOptional && value == null) {
```
LetItRock, line 16: if you compare with null then the type of value should be: `value: number | undefined | null`

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
```diff
@@ -0,0 +1,36 @@
...
+      return value;
+    }
+
+    if (!this.isOptional && value == null) {
```
LetItRock, line 20: loose equality comparison is bad practice... I think it's hard to spot that it's loose and I feel like not everyone might remember that `undefined == null` is also true...

apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
```diff
@@ -48,14 +47,35 @@ export class GetMessages {
...
     return {
       page: command.page,
       totalCount,
+      hasMore,
       pageSize: LIMIT,
       data,
     };
   }
 
+  private getHasMore(page: number, LIMIT: number, feed, totalCount) {
```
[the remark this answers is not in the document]
LetItRock: maybe it should be moved to the base repository and reused?

apps/widget/cypress/plugins/index.ts
```diff
@@ -109,3 +128,16 @@ module.exports = (on, config) => {
     },
   });
 };
+
+async function awaitRunningJobs({
+  apiKey,
```
LetItRock, line 133: this one should be called `serverUrl`

apps/widget/cypress/plugins/index.ts
```diff
@@ -99,6 +112,12 @@ module.exports = (on, config) => {
       };
     },
 
+    async awaitRunningJobs({ templateId, organizationId }: { templateId: string; organizationId: string }) {
+      await awaitRunningJobs({ apiKey: config.env.API_URL, templateId, organizationId });
+
+      return true;
+    },
```
LetItRock, line 119: you don't need to have this wrapper function, instead assign the default value to the `awaitRunningJobs:132`, like 
```
async function awaitRunningJobs({
  apiKey = config.env.API_URL,
...
```

LetItRock (commented):
hey, left you a few comments ;) also you forgot to assign the ticket ;)

apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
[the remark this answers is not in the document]
djabarovgeorge: I am not sure because this logic will be changed in version 0.16 , after the update, we could rethink it and add the combined query in the repository (for pagination) and the common part in base repository.

apps/api/src/app/widgets/dtos/message-response.dto.ts
djabarovgeorge: Good catch, thanks! 🤘 
I create a test for the pagination itself but not the flag, I will add one as well.

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.spec.ts
djabarovgeorge: Yeah, I was not sure here is to split those or not. Do you think split ones will be easier to read?

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
djabarovgeorge: True

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
djabarovgeorge: Do you mean `value === null || value === undefined` will be more easy to read?

apps/widget/cypress/plugins/index.ts
djabarovgeorge: It is, good catch 🤦‍♂️

apps/widget/cypress/plugins/index.ts
djabarovgeorge: Created this one to use it in `notifications-list.spec.ts` in order to try to solve the flaky tests in the CI. :(

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.spec.ts
LetItRock: like easier to spot the failing parts when you will run the test... so like if you will have let's say 10 tests, you will immediately see that for example three are failing and you will know which are these three tests. 
If you would have one test you would have to run it many times, like run, spot the issue, fix the issue, then run again, spot the issue, fix etc.

apps/api/src/app/widgets/pipes/limit-pipe/limit-pipe.ts
LetItRock: like it's more text, but it has a clear meaning... with `value == null` I might not notice that it's a `==` and not `===`

apps/widget/cypress/plugins/index.ts
LetItRock: sorry what I meant is that you don't have to call `awaitRunningJobs` in line 116... instead you can do:
```
 const session = new UserSession(serverUrl);
  await session.awaitRunningJobs(templateId, undefined, 0, organizationId);
```
like you don't need to have that additional function `awaitRunningJobs`

LetItRock (approved):

PR #3274, opened 2023-04-23
Add notification group endpoints

### LINKED ISSUE #1321: `@novu/node`: complete the notification groups module
- [x] #1741 (Get all, create)
- [x] Get one (no endpoint)
- [x] Update (no endpoint)
- [x] Delete (no endpoint)
- [x] https://github.com/novuhq/api-docs/pull/32

Feel free to tell me if there's more I'm missing and I'll update the list.

peoray:
Related to  #1321


peoray:
@p-fernandez please review :)
cc @florian-lefebvre 

Leaving it as a draft for now but all that's left is testing and reviews

florian-lefebvre:
I'm glad someone is working on this again! I'm no able to review your PR because I'm not familiar enough with the codebase, but I'm sure someone from the Novu team will eventually takes care of it

apps/api/src/app/notification-groups/usecases/index.ts
```diff
@@ -1,8 +1,14 @@
...
+import { GetNotificationGroup } from './get-notification-group/get-notification-group.usecase';
+import { DeleteNotificationGroup } from './delete-notification-group/delete-notification-group.usecase';
+import { UpdateNotificationGroup } from './update-notification-group/update-notification-group.usecase';
 
 export const USE_CASES = [
   GetNotificationGroups,
   CreateNotificationGroup,
+  GetNotificationGroup,
+  DeleteNotificationGroup,
+  UpdateNotificationGroup,
```
p-fernandez, line 12: Remove the `//` please.

apps/api/src/app/notification-groups/usecases/update-notification-group/update-notification-group.usecase.ts
```diff
@@ -0,0 +1,45 @@
...
+    const { id, environmentId, name } = command;
+
+    const item = await this.notificationGroupRepository.findOne({
+      _environmentId: environmentId,
+      _id: id,
+    });
+
+    if (!item) {
+      throw new NotFoundException();
+    }
```
p-fernandez, line 20: Instead of doing this you can compose this use case by using the `GetNotificationGroup` use case you created as there you are already handling the situation of a group not found. Therefore if the logic would need to be changed to manage that use case you would only have to change it in one place as this use case depends on the safe guard of a group existing.

apps/api/src/app/notification-groups/usecases/update-notification-group/update-notification-group.usecase.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    await await this.notificationGroupRepository.update(
```
p-fernandez, line 22:
```suggestion
   await this.notificationGroupRepository.update(
```

apps/api/src/app/notification-groups/usecases/update-notification-group/update-notification-group.usecase.ts
```diff
@@ -0,0 +1,45 @@
...
+    );
+
+    const result = await this.notificationGroupRepository.findOne({
+      _environmentId: environmentId,
+      _id: id,
+    });
+
+    if (!result) {
+      throw new NotFoundException(`Notification group with id ${id} not found`);
+    }
```
p-fernandez, line 41: Instead of doing this you can check inside the NotificationGroupRepository if the response from the update method returns that an entry has been changed and manage the logic accordingly.
Also it feels a bit weird that you check again if the group exists after updating it when you have already checked it before updating it. 🤔

apps/api/src/app/notification-groups/e2e/patch-notification-group.e2e.ts
p-fernandez, line 1: As the use case is called Update, rename the `patch` part of the file to `update`, please. 🙏🏻

apps/api/src/app/notification-groups/e2e/patch-notification-group.e2e.ts
```diff
@@ -0,0 +1,37 @@
+import { expect } from 'chai';
+import { UserSession } from '@novu/testing';
+
+describe('Patch Notification Group - /notification-groups/:id (PATCH)', async () => {
```
p-fernandez, line 4:
```suggestion
describe('Update Notification Group - /notification-groups/:id (PATCH)', async () => {
```

apps/api/src/app/notification-groups/e2e/patch-notification-group.e2e.ts
p-fernandez: I'd also add a test to check the behaviour if we are trying to update a notification group that doesn't exist.

apps/api/src/app/notification-groups/e2e/get-notification-group.e2e.ts
```diff
@@ -0,0 +1,41 @@
...
+    const { body } = await session.testAgent.get(`/v1/notification-groups/${id}`);
+
+    const group = body.data;
+
+    expect(group.name).to.equal(`Test name`);
+    expect(group._id).to.equal(postNotificationGroup1.body.data.id);
+    expect(group._environmentId).to.equal(session.environment._id);
+  });
+
+  it('should get 404 when notification group is not present with id', async function () {
```
p-fernandez, line 30:
```suggestion
  it('should get 404 when notification group is not present with the requested id', async function () {
```

apps/api/src/app/notification-groups/e2e/delete-notification-group.e2e.ts
p-fernandez, line 1: Same as in update, I would add a test case to see the behaviour if we try to delete a notification group with an id that it doesn't exist.

apps/api/src/app/notification-groups/dtos/delete-notification-group-response.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+  @IsDefined()
+  acknowledged: boolean;
+
+  @ApiProperty({
+    description: 'The status enum for the performed action',
+    enum: ['deleted'],
+  })
+  @IsString()
+  @IsDefined()
+  status: string;
```
p-fernandez, line 18: What's the reason behind adding this status to the API response of a call that from the method we know it is a delete call? Did you have something in mind that in the future we could benefit from having a status?

p-fernandez (commented):
Thanks for this work! Some comments to discuss and few improvements.

apps/api/src/app/notification-groups/usecases/update-notification-group/update-notification-group.usecase.ts
peoray: I need to return an object of the update code as per the `NotificationGroupResponseDto`

apps/api/src/app/notification-groups/dtos/delete-notification-group-response.dto.ts
peoray: I got this from the issue, @scopsy mentioned it somewhere. Also, we have similar in the remove-subscribe usecase

p-fernandez: I am not particularly fan of returning 200 status for successful delete calls as I consider it complicates consumer's logic besides not adding extra useful information. As it is an implementation detail requested I will discuss it in the core team. 🙂

p-fernandez (approved):
🌟

apps/api/src/app/notification-groups/dtos/delete-notification-group-response.dto.ts
peoray: Alright then.

Unrelated, but I admired your eye for detail and thorough reviews, especially for the PRs I have raised so far. It helps me to understand the codebase better and overall, a skill I want to emulate.

PR #3275, opened 2023-04-23
docs: add faq for ttl

ainouzgali:

Add the ttl expiration to docs - for now only faq. I am not sure where else it would be right to add it.


scopsy (approved):
Perfect! Great addition!

PR #3276, opened 2023-04-23
fix: suscriber response dto

jainpawan21:

fix few dtos issue


djabarovgeorge (commented):
Looks good to me, do you think we should update the 
`subscribers/:subscriberId/notifications/feed` endpoint in this PR as well?

In addition, what do you think about a small test that checks if the limit from the query will work?

jainpawan21:
> Looks good to me, do you think we should update the `subscribers/:subscriberId/notifications/feed` endpoint in this PR as well?
> 
> In addition, what do you think about a small test that checks if the limit from the query will work?

Fixed this endpoint also

PR #3280, opened 2023-04-24
remove worker from docker-compose.yml

diegoquintanav:
Removes service `worker` from docker-compose.yml because image for `ghcr.io/novuhq/novu/worker:0.13.0` does not exist

I'm following the instructions from <https://docs.novu.co/overview/docker-deploy#quick-start>, and it seems that these docs are outdated (there's no docker-compose.redis-cluster.yml at that location). 

If I check the repository, the README says that I should use deployment/docker-compose.yml  instead, which is pulling an image for worker that does not exist (there's a `worker-ee` at <https://github.com/orgs/novuhq/packages> but no luck with that one).

`worker` image does not exist and breaks deployment

### DIFF
--- a/docker/local/deployment/docker-compose.yml
+++ b/docker/local/deployment/docker-compose.yml
@@ -49,29 +49,6 @@ services:
       API_CONTEXT_PATH: ${API_CONTEXT_PATH}
     ports:
       - '3000:3000'
-  worker:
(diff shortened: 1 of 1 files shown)


yndawmarc:
Hi, is this done? I can't still find the **docker-compose.redis-cluster.yml**  in deployment.

PR #3281, opened 2023-04-24
Feat[NV-1868]: Initial Queue Metric Monitoring 

Cliftonz:
This pr adds a metric queue and repeatable job to sample each queue in Novu every second to have the ability to later dynamically scale on the number of messages in the queue.


apps/api/package.json
```diff
@@ -57,6 +58,7 @@
     "class-validator": "^0.14.0",
     "compression": "^1.7.4",
     "cross-env": "^7.0.3",
+    "d": "^1.0.1",
```
scopsy, line 61: @Cliftonz what this is?

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+      });
+    };
+  }
+
+  private async jobHasCompleted(job): Promise<void> {
+    try {
+      Logger.log('Metric job Completed');
+    } catch (error) {
+      Logger.error('Failed to set job as completed', error);
+    }
```
scopsy, line 115: In which scenario this should throw? 
```suggestion
      Logger.log('Metric job Completed');
```

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  private async jobHasFailed(job, error): Promise<void> {
+    try {
+      Logger.log('Metric job failed', error);
+    } catch (anotherError) {
+      Logger.error(`Failed to set job as failed ${anotherError}`);
+    }
```
scopsy, line 123:
```suggestion
      Logger.error('Metric job failed', error);
```

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -110,6 +116,17 @@ const SERVICES = [
     provide: TriggerQueueService,
     useClass: TriggerProcessorQueueService,
   },
+  {
+    provide: WsQueueService,
+    useClass: WsQueueService,
+  },
```
scopsy, line 122: Why is this needed here?

packages/application-generic/src/services/bull-mq.service.ts
```diff
@@ -64,8 +78,15 @@ export class BullmqService {
     const WorkerClass = !BullmqService.pro
       ? Worker
       : require('@taskforcesh/bullmq-pro').WorkerPro;
+
+    let internalOptions: WorkerOptions;
+    if (options) {
+      internalOptions = options;
+    }
+    internalOptions.metrics = { maxDataPoints: MetricsTime.ONE_MONTH };
```
scopsy, line 86:
```suggestion
    let internalOptions: WorkerOptions = {};
    if (options) {
      internalOptions = options;
    }
    internalOptions.metrics = { maxDataPoints: MetricsTime.ONE_MONTH };
```

internalOptions.metrics might fail because internal options can be potentially undefined.

packages/application-generic/src/services/queue.service.ts
```diff
@@ -22,10 +21,9 @@ export class QueueService<T = unknown> {
   public readonly bullMqService: BullmqService;
   public readonly DEFAULT_ATTEMPTS = 3;
 
-  constructor() {
+  constructor(public name = 'standard') {
```
scopsy, line 24: We shouldn't allow to override this property. Why is it needed? this can lead to some really bad bugs and back compatibility issues.

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function sumOfSquares(values: number[]): number {
+  const meanVal = mean(values);
+
+  return sum(values.map((value) => (value - meanVal) ** 2));
+}
+
+@Injectable()
+export class MetricQueueService extends QueueService<Record<string, never>> {
```
scopsy, line 22: I'm not sure regarding this extension here from QueueService, because the QueueService previously served the job processing queue, and I'm afraid extending it like this can create a lot of problems. @LetItRock could you help with reviewing this one?

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+      await this.jobHasCompleted(job);
+    });
+
+    this.bullMqService.worker.on('failed', async (job, error) => {
+      await this.jobHasFailed(job, error);
+    });
+
+    void this.addToQueue('metric-job', {}, '', {
+      repeat: {
+        immediately: true,
```
scopsy, line 40: Note that immediatly can cause some weired behaviour when a new image is deployed. Would suggest to keep this on the pattern only.

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+      await this.bullMqService.queue.drain();
+      await this.bullMqService.worker.close();
+    }
+  }
+
+  private getWorkerOpts(): WorkerOptions {
+    return {
+      ...this.bullConfig,
+      lockDuration: 90000,
+      concurrency: 200,
```
scopsy, line 58: The concurreny here should be 1, and lockDuration much lower rate in my opinion

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+      settings: {},
+    } as WorkerOptions;
+  }
+
+  private getWorkerProcessor() {
+    return async () => {
+      return await new Promise(async (resolve, reject) => {
+        for (const queueService of this.token_list) {
+          this.bullMqService.createQueue(queueService.name, {
```
scopsy, line 67: This creates a global instance on the parent class, and can lead to a lot of weird behaviour. We should create a new bullMq service instance for each Queue we have here.

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,125 @@
...
+          Logger.log(`${process.env.NOVU_MANAGED_SERVICE}: managed`);
+          Logger.log(`MetricQueueService/${queueService.name}/completed`, JSON.stringify(successMetric));
+          Logger.log(`MetricQueueService/${queueService.name}/completed`, JSON.stringify(failMetric));
+          if (process.env.NOVU_MANAGED_SERVICE === 'true') {
+            Logger.log(`MetricQueueService/${queueService.name}/completed`, JSON.stringify(successMetric));
+            Logger.log(`MetricQueueService/${queueService.name}/completed`, JSON.stringify(failMetric));
+          } else {
+            nr.recordMetric(`MetricQueueService/${queueService.name}/completed`, successMetric);
+            nr.recordMetric(`MetricQueueService/${queueService.name}/failed`, failMetric);
+          }
```
scopsy, line 104: I'm quite confused here, could you eloborate on why we are only logging it on the NOBU MANAGED SERVICe vs actually sending this to newrelic? we need the exact opposite here. Or Am I missing something here?

scopsy (commented):
@LetItRock could you also help with a review here?

apps/worker/src/app/workflow/services/workflow-queue.service.ts
```diff
@@ -43,6 +43,7 @@ export class WorkflowQueueService extends QueueService<IJobData> {
     @Inject(forwardRef(() => CreateExecutionDetails)) private createExecutionDetails: CreateExecutionDetails
   ) {
     super();
+    Logger.warn('Workflow queue service created');
```
LetItRock, line 46: leftover?

apps/worker/src/app/workflow/workflow.module.ts
LetItRock: The `WsQueueService` doesn't have the worker logic and yes it's to get the metrics

packages/application-generic/src/services/bull-mq.service.ts
LetItRock:
```suggestion
    const internalOptions: WorkerOptions = {
       ...options,
       metrics: { maxDataPoints: MetricsTime.ONE_MONTH },
    }
```

packages/application-generic/src/services/queue.service.ts
LetItRock: explained in the first comment in `MetricQueueService` file

apps/worker/src/app/workflow/services/metric-queue.service.ts
scopsy: Thank you for the explanation @LetItRock this makes sense now, appreciate it!

apps/worker/package.json
```diff
@@ -27,36 +27,39 @@
...
     "@sentry/tracing": "^7.40.0",
+    "@types/newrelic": "^9.13.0",
     "axios": "^1.3.3",
     "body-parser": "^1.20.0",
     "bull": "^4.4.0",
     "bullmq": "^3.10.2",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.0",
     "cross-env": "^7.0.3",
+    "d": "^1.0.1",
```
scopsy, line 49: @Cliftonz Another ping regarding this dependency, where is it used?

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -0,0 +1,135 @@
...
+            count: metrics.failed.count,
+            total: failNumber == 0 ? 0 : sum(metrics.failed.data),
+            min: failNumber == 0 ? 0 : min(metrics.failed.data),
+            max: failNumber == 0 ? 0 : max(metrics.failed.data),
+            sumOfSquares: failNumber == 0 ? 0 : sumOfSquares(metrics.failed.data),
+          };
+
+          if (process.env.NOVU_MANAGED_SERVICE === 'true') {
+            nr.recordMetric(`MetricQueueService/${queueService.name}/completed`, successMetric);
+            nr.recordMetric(`MetricQueueService/${queueService.name}/failed`, failMetric);
```
scopsy, line 117: @Cliftonz maybe I'm missing something, but wasn't the most important thing for this is to understand how many jobs we have in flight and pending right now? The completed and failed is quite nice, but I think it lacks the important metrics of how many pending, and how many processing in parallel at the moment. 

Also would be great to know how many are "scheduled" to the future.

apps/worker/package.json
Cliftonz: This should be removed now

apps/worker/src/app/workflow/services/metric-queue.service.ts
Cliftonz: You are correct, I have updated the code to including getting waiting, active, and delayed jobs.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -136,6 +136,8 @@ export class SendMessageChat extends SendMessageBase {
       await this.createExecutionDetails.execute(
         CreateExecutionDetailsCommand.create({
           ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
ainouzgali, line 140: I believe this one is not needed

ainouzgali (approved):

PR #3284, opened 2023-04-25
Migration script for expireAt 

ainouzgali:

create migration script for existing records - add expireAt


apps/api/migrations/expire-at/expire-at-delay.migration.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+      _environmentId: session.environment._id,
+      _templateId: digestTemplate?._id,
+    });
+    const executionDetails = await executionDetailsRepository.find({
+      _environmentId: session.environment._id,
+      _notificationTemplateId: digestTemplate?._id,
+    });
+
+    notifications.forEach((msg) => {
+      expect(msg.expireAt).to.exist;
```
LetItRock, line 95: should we also check if the date value is what we expect it to be?

apps/api/migrations/expire-at/expire-at.migration.ts
```diff
@@ -0,0 +1,94 @@
...
+  );
+  await jobRepository.update(
+    { ...query, _notificationId: { $nin: excludedIds } },
+    { $set: { expireAt: expireAtOneMonth } }
+  );
+  await executionDetailsRepository.update(
+    { ...query, _notificationId: { $nin: excludedIds } },
+    { $set: { expireAt: expireAtOneMonth } }
+  );
+}
```
LetItRock, line 80: won't we override the entities which already have the `expireAt` set? Like we've deployed the changes with `expireAt` and now we want to run this migration as far as I understand.

apps/api/migrations/expire-at/expire-at.migration.ts
```diff
@@ -0,0 +1,94 @@
...
+    status: JobStatusEnum.PENDING,
+  });
+
+  // digested events stays pending, leaving the notification and deleting the actually digested could cause errors
+  return await notificationRepository._model.distinct('_id', {
+    ...query,
+    _id: { $in: pendingNotifications },
+    _digestedNotificationId: { $exists: false, $eq: null },
+  });
+}
```
LetItRock, line 94: so we exclude all the notifications that were created without the digest... is that correct?
when the digest is created for the first time, the notification won't have this field `_digestedNotificationId` set... so we also exclude these notifications?

can you maybe explain this a little? thanks

apps/api/migrations/expire-at/expire-at.migration.ts
ainouzgali: In query there is { expireAt: { $exists: false }}, so it will only be for those that do not have the field yet

apps/api/migrations/expire-at/expire-at.migration.ts
ainouzgali: @LetItRock you are right. But, for notifications that have been digested - all jobs for the steps after the digest stays pending (we never change their status) while for the first digest notification all jobs are run to completion. So, first digest will not be excluded in lines 83-36, and those that have been digested will not be excluded by lines 89-93. Meaning, all those cases expireAt will be added for them.

apps/api/migrations/expire-at/expire-at-delay.migration.spec.ts
ainouzgali: Sure, I can add that

apps/api/migrations/expire-at/expire-at.migration.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const messageRepository = new MessageRepository();
+const notificationRepository = new NotificationRepository();
+const jobRepository = new JobRepository();
+const executionDetailsRepository = new ExecutionDetailsRepository();
+const organizationRepository = new OrganizationRepository();
+const environmentRepository = new EnvironmentRepository();
+const now = Date.now();
+const expireAtOneMonth = addMonths(now, 1);
+const expireAtSixMonths = addMonths(now, 6);
```
djabarovgeorge, line 20: It could be a good idea to add some VARIANT here,
meaning instead of adding the expire at exact same time we would add diff so Mongo won't expire them all at once in the future.

apps/api/migrations/expire-at/expire-at.migration.ts
```diff
@@ -0,0 +1,94 @@
...
+    {
+      ...query,
+      channel: 'in_app',
+    },
+    { $set: { expireAt: expireAtSixMonths } }
+  );
+}
+
+export async function notificationExpireAt(query) {
+  const excludedIds = await getExcludedNotificationIds(query);
```
scopsy, line 66: @ainouzgali I'm a bit concerned with this query when a customer has millions of events, won't this array will be huge? Is there other wise we can make the query without this find clause here?

ainouzgali: It will only include pending jobs (that are not digest related) - for example, a delay that did not finish yet. It shouldn't be big unless I'm missing something @scopsy

djabarovgeorge: Maybe we could use findBatch() here?

djabarovgeorge: Not here exactly but the idea in general in the flow.

djabarovgeorge (approved):
⭐

scopsy (approved):

PR #3291, opened 2023-04-26
Update digest.md

david-morris:
Explicitly state behavior of digest step with only one item

Update digest.md to better explain what happens to a digest with one item.

I spent a lot of time trying to figure out how Novu reacts to digests of one item from the documentation, since the sample template includes the digest status as a condition.  I'm guessing that's for back-off digests?

Anyway, this would explicitly note what happens to a digest with one item.


scopsy (approved):
Makes sense, thank you!

PR #3232, opened 2023-04-18
fix: remove example folder

jainpawan21:
### What change does this PR introduce?
Moved to example repo with this PR
https://github.com/novuhq/examples/pull/1

PR #3303, opened 2023-04-29
fix: add sender name option in mailjet

jainpawan21:
Add senderName option in mailjet provider

With mailjet as active integration, sender name was not visible in email

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "start:docker:embed": "cd libs/embed && npm run start:docker",
     "start:docker:widget": "cross-env nx run-many --target=start:docker --projects=@novu/widget",
     "docker:build": "pnpm -r --if-present --parallel docker:build",
-    "build": "nx run-many --target=build --exclude=angular-notification-center-example,vue-notification-center-example --all",
+    "build": "nx run-many --target=build --all",
     "prebuild": "nx run-many --target=prebuild --all",
     "build:api": "nx build @novu/api",
     "build:worker": "nx build @novu/worker",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "start:docker:embed": "cd libs/embed && npm run start:docker",
     "start:docker:widget": "cross-env nx run-many --target=start:docker --projects=@novu/widget",
     "docker:build": "pnpm -r --if-present --parallel docker:build",
-    "build": "nx run-many --target=build --exclude=angular-notification-center-example,vue-notification-center-example --all",
```
ainouzgali, line 44: Was this removed on purpose? What is behind this?

ainouzgali (approved):
@jainpawan21 looks good :) 
I saw that there are other providers not supporting senderName, maybe we should open tickets about them? Maybe community members could help us with those

package.json
jainpawan21: I recently removed these examples in this [PR](https://github.com/novuhq/novu/pull/3232) and moved them to our new repo novuhq/examples in this [PR](https://github.com/novuhq/examples/pull/1)

unicodeveloper:
> @jainpawan21 looks good :) I saw that there are other providers not supporting senderName, maybe we should open tickets about them? Maybe community members could help us with those

Yeah we should open tickets for them. @sumitsaurabh927 this might be your chance to dig in to the code. WDYT?

sumitsaurabh927:
Sure @unicodeveloper 
Is there something specific you want me to do?

PR #3304, opened 2023-04-30
feat: add redirect url to slack

djabarovgeorge:

In order to allow the user to provide a redirect URL to their application, the default will be to close the tab.

In order to let the user add additional internal logic if needed.


ainouzgali (approved):
⭐️

PR #3305, opened 2023-04-30
Create Novu redirect url for Slack

djabarovgeorge:

An API endpoint for the user to pass in order to handle the OAuth flow.

In order to handle slack redirects, in the  OAuth flow.


apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
```diff
@@ -0,0 +1,126 @@
...
+        subscriberId: command?.subscriberId,
+      })
+    );
+
+    const subscriberCredentials: IChannelCredentialsCommand = { webhookUrl: webhookUrl, channel: command.providerId };
+
+    await this.updateSubscriberChannelUsecase.execute(
+      UpdateSubscriberChannelCommand.create({
+        organizationId: environment._organizationId,
+        environmentId: command.environmentId,
```
p-fernandez, line 57: Could we get rid of the need to use the `organizationId` so that would simplify a lot the code? Also checking `UpdateSubscriberChannel` use case I can see we have no use for `organizationId`: https://github.com/novuhq/novu/blob/f1ab32fef885a5422bde34572ca7d794261c3960/apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.command.ts#L20
https://github.com/novuhq/novu/blob/f1ab32fef885a5422bde34572ca7d794261c3960/apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.usecase.ts#L10

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
```diff
@@ -0,0 +1,126 @@
...
+    private environmentRepository: EnvironmentRepository,
+    private createSubscriberUsecase: CreateSubscriber
+  ) {}
+
+  async execute(command: HandleChatOauthCommand) {
+    const { credentials: integrationCredentials } = await this.getIntegration(command);
+
+    const webhookUrl = await this.getWebhook(command, integrationCredentials);
+
+    const environment = await this.getEnvironment(command);
```
p-fernandez, line 33: In case my request to not to have to use this to retrieve the organizationId is not possible I would swap the order of the execution because if no environment is found we save to do an HTTP call to Slack in `this.getWebhook`.
We can even do a parallel call of it with `this.getIntegration` to save a bit of time as they are independent of each other and if any fails would stop the execution throwing an API exception.

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
```diff
@@ -0,0 +1,126 @@
...
+      client_secret: integrationCredentials.secretKey,
+    };
+    const config = {
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+    };
+
+    const res = await axios.post(this.SLACK_ACCESS_URL, body, config);
+    const webhook = res.data?.incoming_webhook?.url;
```
p-fernandez, line 94: I consider this should be an application generic Slack service. But so far could it live here by now.

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
```diff
@@ -0,0 +1,126 @@
...
+        oauthHandler: OAuthHandlerEnum.NOVU,
+      })
+    );
+  }
+
+  private async getEnvironment(command: HandleChatOauthCommand) {
+    const environment = await this.environmentRepository.findById(command.environmentId);
+
+    if (environment == null) {
+      throw new ApiException(`Environment ID: ${command.environmentId} not found`);
```
p-fernandez, line 70:
```suggestion
      throw new NotFoundException(`Environment ID: ${command.environmentId} not found`);
```
Though maybe we could make the SubscribersModule dependent of the EnvironmentModule and use here the `GetEnvironment` use case instead of reimplementing it.

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.command.ts
```diff
@@ -0,0 +1,50 @@
...
+  IsMongoId,
+  IsString,
+  registerDecorator,
+  ValidationArguments,
+  ValidationOptions,
+} from 'class-validator';
+import { ChatProviderIdEnum } from '@novu/shared';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function IsNotEmpty(validationOptions?: ValidationOptions) {
```
p-fernandez, line 13: What's the difference of this implementation with the `class-validator` `IsNotEmpty`?

apps/api/src/app/subscribers/e2e/handle-chat-oauth.e2e.ts
```diff
@@ -0,0 +1,157 @@
...
+import axios from 'axios';
+import * as sinon from 'sinon';
+
+import { UserSession } from '@novu/testing';
+import { ChannelTypeEnum, ChatProviderIdEnum } from '@novu/shared';
+import { SubscriberRepository } from '@novu/dal/';
+
+const axiosInstance = axios.create();
+
+describe('HandleCharOauth - /:subscriberId/:providerId/:environmentId (GET)', function () {
```
p-fernandez, line 11:
```suggestion
describe('HandleChatOauth - /:subscriberId/:providerId/:environmentId (GET)', function () {
```

libs/shared/src/dto/integration/construct-integration.interface.ts
```diff
@@ -13,11 +13,15 @@ export interface ICredentialsDto {
...
+  applicationId?: string;
+  clientId?: string;
   projectName?: string;
+  serviceAccount?: string;
+  baseUrl?: string;
   requireTls?: boolean;
   ignoreTls?: boolean;
   tlsOptions?: Record<string, unknown>;
-  baseUrl?: string;
+  redirectUrl?: string;
```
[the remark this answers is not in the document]
djabarovgeorge: Not sure what you mean here by reuse.

apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.command.ts
```diff
@@ -25,4 +25,12 @@ export class UpdateSubscriberChannelCommand extends EnvironmentCommand implement
...
   @ValidateNested()
   credentials: IChannelCredentialsCommand;
+
+  @IsNotEmpty()
+  oauthHandler: OAuthHandlerEnum;
+}
+
+export enum OAuthHandlerEnum {
+  NOVU = 'novu',
+  EXTERNAL = 'external',
```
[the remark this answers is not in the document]
djabarovgeorge: Tend to agree, here thanks.

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
```diff
@@ -0,0 +1,126 @@
...
+    const query: Partial<IntegrationEntity> & { _environmentId: string } = {
+      _environmentId: command.environmentId,
+      channel: ChannelTypeEnum.CHAT,
+      providerId: command.providerId,
+    };
+
+    const integration = await this.integrationRepository.findOne(query);
+
+    if (integration == null) {
+      throw new ApiException(
```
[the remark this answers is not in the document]
djabarovgeorge: Good catch!

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
djabarovgeorge: Yeah once we will have this logic in multiple providers we could refactor it.

apps/api/src/app/subscribers/usecases/handle-chat-oauth/handle-chat-oauth.usecase.ts
djabarovgeorge: Yeah totally agree here, i would love to remove it, but then i will need to make it optional on all the other commands. the UserSession is not an option because we don't have it.
But but we could get it as a param from the user. i will look into it in the refactoring phase now i need to merge it to the next branch so i could test with dev env and not with localhost and ngrok :(

apps/api/src/app/subscribers/e2e/handle-chat-oauth.e2e.ts
```diff
@@ -0,0 +1,157 @@
...
+      throw new Error('Exception should have been thrown');
+    } catch (e) {
+      expect(e.response.data.message[0]).to.equal('subscriberId should not be undefined');
+    }
+
+    try {
+      await handleChat(session.serverUrl, 'undefined', session.environment._id);
+      throw new Error('Exception should have been thrown');
+    } catch (e) {
+      expect(e.response.data.message[0]).to.equal('subscriberId should not be undefined');
```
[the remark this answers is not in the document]
djabarovgeorge: Definitely but actually I want to remove this one and create a unit test with those cases directly to the new @IsNotEmpty validator.
Created this one for now for the development.

apps/api/src/app/subscribers/e2e/handle-chat-oauth.e2e.ts
djabarovgeorge: Good catch this one was on all the files 😂 missed this one on the typo fixes

p-fernandez (approved):
⚡

PR #3306, opened 2023-04-30
fix: merge template status to parent change

### LINKED ISSUE #3287: [NV-2170] 🐛 Bug Report: Broken workflow on the Changes page
### 📜 Description

Promoting an old change on a now-deleted template will re-create the template in prod in a non-deletable state.
If a template reaches a state where it exists only in prod and not in development, it is impossible to disable, delete, or modify.

### 👟 Reproduction steps

1. Disable a notification in Development
2. Delete a notification in Development
3. Go on the Changes page
4. Promote the "delete" change
5. Promote the "disable" change

### 👍 Expected behavior

The template should be deleted in Production and all related-pending changes to that template should be gone from the Changes list
It would also be greatly beneficial to be able to "Discard" changes and not only "Promote" them

- The template no longer exists in development, but exists in production and cannot be deleted (not even through the API)
- All older pending changes that have not been promoted are still lingering in the Changes table and can be promoted
- The template which now only exists in production cannot be modified at all

image

_No response_

None

<sub>[NV-2170](https://linear.app/novu/issue/NV-2170/🐛-bug-report-broken-workflow-on-the-changes-page)</sub>

djabarovgeorge:

This PR adds a merge of status change to parent change, which will help the user to promote the changes in historical order. 
 
At the moment status changes create separate changes which can cause issues if the order of the promotions is different from the historical order.
Closes #3287

### DIFF
--- a/apps/api/src/app/notification-template/usecases/change-template-active-status/change-template-active-status.usecase.ts
+++ b/apps/api/src/app/notification-template/usecases/change-template-active-status/change-template-active-status.usecase.ts
@@ -1,5 +1,5 @@
 import { BadRequestException, Injectable, NotFoundException } from '@nestjs/common';
-import { NotificationTemplateEntity, NotificationTemplateRepository } from '@novu/dal';
+import { NotificationTemplateEntity, NotificationTemplateRepository, ChangeRepository } from '@novu/dal';
 import { ChangeEntityTypeEnum } from '@novu/shared';
 import {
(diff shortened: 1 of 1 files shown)


huangchengshun:
💼
![1710373010580](https://github.com/novuhq/novu/assets

①huangchengsheng√
+🗝~🗝~+ m🔒~🔒

PR #3308, opened 2023-05-01
feat: add instrument, combine subscriber+job processing

ainouzgali:

### DIFF
--- a/packages/application-generic/src/usecases/store-subscriber-jobs/store-subscriber-jobs.usecase.ts
+++ b/packages/application-generic/src/usecases/store-subscriber-jobs/store-subscriber-jobs.usecase.ts
@@ -1,4 +1,4 @@
-import { JobRepository, NotificationRepository } from '@novu/dal';
+import { JobRepository, NotificationRepository, JobEntity } from '@novu/dal';
 import { Injectable } from '@nestjs/common';
 import {
   StepTypeEnum,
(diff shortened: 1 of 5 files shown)


packages/application-generic/src/usecases/store-subscriber-jobs/store-subscriber-jobs.usecase.ts
```diff
@@ -53,24 +44,41 @@ export class StoreSubscriberJobs {
...
+        CreateExecutionDetailsCommand.create({
+          ...CreateExecutionDetailsCommand.getDetailsFromJob(job),
+          detail: DetailEnum.STEP_CREATED,
+          source: ExecutionDetailsSourceEnum.INTERNAL,
+          status: ExecutionDetailsStatusEnum.PENDING,
+          isTest: false,
+          isRetry: false,
+        })
+      );
+    }
```
scopsy, line 82: Wonder if we can do a bulk write here in the future

ainouzgali: I already did it

ainouzgali: I want to push it in a different PR, and also thats why I added the Instrument here, I want to have a more precise understanding of the affect

davidsoderberg (approved):

PR #3312, opened 2023-05-01
fix: serialization and primitives validation (no. 1)

michaldziuba03:

This is first PR in **validation fix** pull requests series.
It fixes some problems I mentioned in DMs to @scopsy :)

#### Affected modules (some of them only partially)
- auth
- integration
- notification-template
- execution-detail
- change
- invite
- organization
- subscribers

#### Changes:
- add validation for pagination limit parameter
- add primitives validation (`@IsString`, `@IsEnum`...) for affected modules to avoid data corruption and some other dangerous things.
-  add field selection for joined user in `Change` entity because API returns whole user object with sensitive data (password hash, reset token hash). Now it returns only necessary for UI data.
- add `@IsMongoId` in some places, because API often returns `HTTP Status: 500` when Mongo Id is expected but some random value is provided. **During review pay attention to whether the right fields have been annotated as Mongo Ids**.
- replace inline `@Query('someQueryParam') param: string` with `@Query() query: QueryDto` to perform validation and improve Swagger documentation :) I fixed HTTP Parameter Pollution in affected modules.

image

image

> HTTP Parameter Pollution. Keep in mind query param is not always string - it's possible to inject object or array.

- create `transformers.ts` file in shared modules. It actually will be used in future PRs in the series.
- create `validators.ts` file for custom class validators
- move some dtos from organization to invites. Why? Because some dtos defined in organization module was ONLY used in invites module.
- change some tests in `integration` module. In most cases I try to fix my solutions first without changing existing tests - but in some tests "EMAIL" was used for channel instead `ChannelTypeEnum.EMAIL`. I checked frontend and some other tests (they rely on enum) - I decided to accept only ChannelTypeEnum as is without transformations to lowercase etc.

To improve security and prevent some crazy things with corrupted data to happen in the future.


scopsy:
Amazing work! A small suggestion for future PR's is a one I learned from @LetItRock, is to create inline comments on the PR code diff to explain why a particular change is made. In the way you have written in the PR body, just makes it more contextual to the actual code change. Not a must, but just something I find makes easier to review bigger PR's 🚀

apps/api/src/app/change/dtos/change-request.dto.ts
```diff
@@ -1,24 +1,14 @@
...
-    required: false,
-    default: 10,
-    maximum: 100,
-  })
-  limit?: number = 10;
+import { ApiProperty } from '@nestjs/swagger';
+import { IsDefined, IsString } from 'class-validator';
+import { PaginationRequestDto } from '../../shared/dtos/pagination-request';
 
+export class ChangesRequestDto extends PaginationRequestDto(10, 100) {
```
djabarovgeorge, line 5: Good catch! :)

apps/api/src/app/organization/organization.controller.ts
```diff
@@ -140,10 +143,7 @@ export class OrganizationController {
   }
 
   @Put('/branding')
-  async updateBrandingDetails(
-    @UserSession() user: IJwtPayload,
-    @Body() body: { color: string; logo: string; fontColor: string; contentBackground: string; fontFamily: string }
-  ) {
+  async updateBrandingDetails(@UserSession() user: IJwtPayload, @Body() body: UpdateBrandingDetailsDto) {
```
djabarovgeorge, line 146: 🥳

libs/dal/src/repositories/change/change.entity.ts
```diff
@@ -12,6 +12,8 @@ export class ChangeEntity {
 
   _environmentId: EnvironmentId;
 
+  user: Pick<UserEntity, '_id' | 'firstName' | 'lastName' | 'profilePicture'>;
```
djabarovgeorge, line 15: Should we add this one here only for the ChangeRepository.getList populate?
Maybe instead we can create a response type with it for getList, wdyt?

djabarovgeorge (commented):
Looks like amazing work! left a small question for your review :)

apps/api/src/app/shared/validators.ts
```diff
@@ -0,0 +1,7 @@
+import { IsIn } from 'class-validator';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const IsBooleanAny = () =>
+  IsIn([true, false, 'true', 'false'], {
+    message: '$property must be a boolean value',
+  });
```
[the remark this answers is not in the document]
michaldziuba03: Sure. I made this validator because frontend sends "true"/"false" for some fields - maybe we will change this behaviour to send only real boolean and use built-in `@IsBoolean`?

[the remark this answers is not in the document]
michaldziuba03: but for example in some e2e tests for integration - regular boolean is sent.

libs/dal/src/repositories/change/change.entity.ts
michaldziuba03: I agree - I will create type for this particular method

scopsy:
Looking good! Let's finish with @djabarovgeorge suggestions and we can merge this :)

michaldziuba03:
@scopsy @djabarovgeorge new changes are ready to review

libs/dal/src/repositories/change/change.repository.ts
```diff
@@ -6,6 +6,10 @@ import { ChangeEntity, ChangeDBModel } from './change.entity';
 import { Change } from './change.schema';
 import { UserEntity } from '../user';
 
+type ChangeEntityPopulated = ChangeEntity & {
```
djabarovgeorge, line 9: Could we export it one level outside 
as we have in environment types? 
libs/dal/src/repositories/environment/types.ts

djabarovgeorge (approved):
Loved your work here, and left a small suggestion.

P.S. I was inspired a bit by you and created a @IsNotEmpty validator in one of my tasks, once we merge this one I will move it to your validators folder :)

michaldziuba03:
@djabarovgeorge @scopsy I think it's ready to merge - check out last small changes I made.

scopsy (approved):

djabarovgeorge (approved):

PR #3315, opened 2023-05-02
removed typo

fadkeabhi:
Removed typo

Its not necessary, but always good not to have typos


djabarovgeorge (approved):
Good catch!  🚀

PR #3317, opened 2023-05-03
Remove readonly on allowed domains

djabarovgeorge:

Allow updating allowed domains on prod env as well.

So users could add addition dns environments for their domain.


p-fernandez (approved):
🌟

PR #3321, opened 2023-05-03
Bulk write for execution details

ainouzgali:
…update env id

execution details: insertMany, 
job update many -> updateOne
job update status - environmentId vs organizationId 
no loop over integrations for in app

Note - I will tidy up bulk execution code, first I want to see the effect in performance in dev

### DIFF
--- a/packages/application-generic/src/usecases/bulk-create-execution-details/bulk-create-execution-details.usecase.ts
+++ b/packages/application-generic/src/usecases/bulk-create-execution-details/bulk-create-execution-details.usecase.ts
@@ -1,0 +1,38 @@
+import { InstrumentUsecase } from '../../instrumentation';
+import { Injectable } from '@nestjs/common';
+import { ExecutionDetailsRepository, ExecutionDetailsEntity } from '@novu/dal';
+
+import { BulkCreateExecutionDetailsCommand } from './bulk-create-execution-details.command';
+import { mapExecutionDetailsCommandToEntity } from '../create-execution-details';
+
+@Injectable()
+export class BulkCreateExecutionDetails {
+  constructor(private executionDetailsRepository: ExecutionDetailsRepository) {}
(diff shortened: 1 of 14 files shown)


packages/application-generic/src/usecases/bulk-create-execution-details/bulk-create-execution-details.usecase.ts
djabarovgeorge, line 1: A am wondering if we actually need this additional usecase here or maybe we could do this logic inside the CreateExecutionDetails 🤔

djabarovgeorge (approved):
Looks great, I left couple of comments there for your review 🙃

p-fernandez (approved):

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -211,16 +211,18 @@ export class TriggerEvent {
...
-          const provider = await this.getProviderId(
-            userId,
-            organizationId,
-            environmentId,
-            channelType
-          );
-          if (provider) {
-            providers.set(channelType, provider);
-          } else {
+          if (channelType === 'in_app') {
```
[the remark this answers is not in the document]
ainouzgali: Niceeee, great idea. Thank you! I think this deserves a different PR to see the effect of that.

packages/application-generic/src/usecases/bulk-create-execution-details/bulk-create-execution-details.usecase.ts
ainouzgali: I could definitely improve the Command of Bulk and reuse some stuff from create executions details. But, in the same usecase I think it will be too much as one has insertMany and the other per one. I will reiterate on it and tag you to see what you  think

PR #3323, opened 2023-05-03
feat: validate notification group

djabarovgeorge:

validate notification group on template id.

in order to make sure we won't get edge cases when we get notification groups from other environments.


p-fernandez (approved):
How hard would be to add a test for this to check it throws the error?
Othe than that... 🌟

PR #3325, opened 2023-05-04
fix: keep unapplied changes

djabarovgeorge:

Keep the changes if the change was not applied because of an exception.

In order to not confuse the user, and keep the change so it could be merged once the issue is resolved. 

The downside is that the changes will accumulate to a large number


jainpawan21 (approved):
🎉

LetItRock (approved):
looks good :)

PR #3326, opened 2023-05-04
feat(infra): graceful shutdown for bull mq

p-fernandez:

Introduces graceful shutdown for the Bull MQ and all the services dependant on it.

Reliability on system scaling up/down, avoiding to lose jobs while closing. These changes should help us to avoid information when shutting down an instance.

### DIFF
--- a/apps/api/src/app/health/health.controller.ts
+++ b/apps/api/src/app/health/health.controller.ts
@@ -1,32 +1,33 @@
 import { Controller, Get } from '@nestjs/common';
 import { ApiExcludeController } from '@nestjs/swagger';
-import { HealthCheck, HealthCheckService } from '@nestjs/terminus';
+import { HealthCheck, HealthCheckResult, HealthCheckService } from '@nestjs/terminus';
 import { HealthIndicatorFunction } from '@nestjs/terminus/dist/health-indicator';
-import { DalService } from '@novu/dal';
+import {
+  CacheServiceHealthIndicator,
+  DalServiceHealthIndicator,
+  InMemoryProviderServiceHealthIndicator,
+  TriggerQueueServiceHealthIndicator,
+} from '@novu/application-generic';
 
 import { version } from '../../../package.json';
-import { CacheServiceHealthIndicator } from './cache-health-indicator';
 
 @Controller('health-check')
 @ApiExcludeController()
 export class HealthController {
   constructor(
     private healthCheckService: HealthCheckService,
-    private dalService: DalService,
-    private cacheHealthIndicator: CacheServiceHealthIndicator
+    private cacheHealthIndicator: CacheServiceHealthIndicator,
+    private dalHealthIndicator: DalServiceHealthIndicator,
+    private inMemoryHealthIndicator: InMemoryProviderServiceHealthIndicator,
+    private triggerQueueHealthIndicator: TriggerQueueServiceHealthIndicator
   ) {}
 
   @Get()
   @HealthCheck()
-  healthCheck() {
+  healthCheck(): Promise<HealthCheckResult> {
     const checks: HealthIndicatorFunction[] = [
-      async () => {
-        return {
-          db: {
-            status: this.dalService.connection.readyState === 1 ? 'up' : 'down',
-          },
-        };
-      },
+      () => this.dalHealthIndicator.isHealthy(),
+      () => this.inMemoryHealthIndicator.isHealthy(),
       async () => {
         return {
           apiVersion: {
@@ -35,6 +36,7 @@ export class HealthController {
           },
         };
       },
+      () => this.triggerQueueHealthIndicator.isHealthy(),
     ];
 
     if (process.env.ELASTICACHE_CLUSTER_SERVICE_HOST) {
--- a/apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
+++ b/apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
@@ -4,7 +4,7 @@ import { getRedisPrefix } from '@novu/shared';
 import { InboundEmailParse } from '../usecases/inbound-email-parse/inbound-email-parse.usecase';
 import { InboundEmailParseCommand } from '../usecases/inbound-email-parse/inbound-email-parse.command';
 import { ConnectionOptions } from 'tls';
-import { BullmqService } from '@novu/application-generic';
+import { BullMqService } from '@novu/application-generic';
 
 @Injectable()
 export class InboundParseQueueService {
@@ -25,10 +25,10 @@ export class InboundParseQueueService {
   };
   public readonly queue: Queue;
   public readonly worker: Worker;
-  private readonly bullMqService: BullmqService;
+  private readonly bullMqService: BullMqService;
 
   constructor(private emailParseUsecase: InboundEmailParse) {
-    this.bullMqService = new BullmqService();
+    this.bullMqService = new BullMqService();
     this.queue = this.bullMqService.createQueue(this.QUEUE_NAME, {
       ...this.bullConfig,
       defaultJobOptions: {
--- a/apps/api/src/app/shared/constants.ts
+++ b/apps/api/src/app/shared/constants.ts
@@ -1,3 +1,2 @@
-export const QUEUE_SERVICE = 'QueueService';
 export const DAL_SERVICE = 'DalService';
 export const EXCEPTION_MESSAGE_ON_WEBHOOK_FILTER = 'Exception while performing webhook request.';
--- a/apps/api/src/app/shared/shared.module.ts
+++ b/apps/api/src/app/shared/shared.module.ts
@@ -36,6 +36,7 @@ import {
   WsQueueService,
   DistributedLockService,
   PerformanceService,
+  TriggerQueueService,
 } from '@novu/application-generic';
 
 import * as packageJson from '../../../package.json';
@@ -106,6 +107,7 @@ const distributedLockService = {
 const PROVIDERS = [
   cacheService,
   distributedLockService,
+  inMemoryProviderService,
   {
     provide: WsQueueService,
     useClass: WsQueueService,
@@ -118,7 +120,6 @@ const PROVIDERS = [
       return dalService;
     },
   },
-  cacheService,
   {
     provide: PerformanceService,
     useFactory: () => {
@@ -141,6 +142,7 @@ const PROVIDERS = [
       return analyticsService;
     },
   },
+  TriggerQueueService,
 ];
 
 @Module({
--- a/apps/worker/src/app/health/health.controller.ts
+++ b/apps/worker/src/app/health/health.controller.ts
@@ -1,18 +1,29 @@
 import { Controller, Get } from '@nestjs/common';
 import { HealthCheck, HealthCheckResult, HealthCheckService } from '@nestjs/terminus';
-import { DalServiceHealthIndicator } from '@novu/application-generic';
+import {
+  DalServiceHealthIndicator,
+  QueueServiceHealthIndicator,
+  TriggerQueueServiceHealthIndicator,
+} from '@novu/application-generic';
 
 import { version } from '../../../package.json';
 
 @Controller('health-check')
 export class HealthController {
-  constructor(private healthCheckService: HealthCheckService, private dalHealthIndicator: DalServiceHealthIndicator) {}
+  constructor(
+    private healthCheckService: HealthCheckService,
+    private dalHealthIndicator: DalServiceHealthIndicator,
+    private queueHealthIndicator: QueueServiceHealthIndicator,
+    private triggerQueueHealthIndicator: TriggerQueueServiceHealthIndicator
+  ) {}
 
   @Get()
   @HealthCheck()
   healthCheck(): Promise<HealthCheckResult> {
     return this.healthCheckService.check([
       () => this.dalHealthIndicator.isHealthy(),
+      () => this.queueHealthIndicator.isHealthy(),
+      () => this.triggerQueueHealthIndicator.isHealthy(),
       async () => {
         return {
           apiVersion: {
--- a/apps/worker/src/app/health/health.controller.ts
+++ b/apps/worker/src/app/health/health.controller.ts
@@ -1,24 +1,18 @@
 import { Controller, Get } from '@nestjs/common';
-import { HealthCheck, HealthCheckService } from '@nestjs/terminus';
-import { DalService } from '@novu/dal';
+import { HealthCheck, HealthCheckResult, HealthCheckService } from '@nestjs/terminus';
+import { DalServiceHealthIndicator } from '@novu/application-generic';
 
 import { version } from '../../../package.json';
 
 @Controller('health-check')
 export class HealthController {
-  constructor(private healthCheckService: HealthCheckService, private dalService: DalService) {}
+  constructor(private healthCheckService: HealthCheckService, private dalHealthIndicator: DalServiceHealthIndicator) {}
 
   @Get()
   @HealthCheck()
-  healthCheck() {
+  healthCheck(): Promise<HealthCheckResult> {
     return this.healthCheckService.check([
-      async () => {
-        return {
-          db: {
-            status: this.dalService.connection.readyState === 1 ? 'up' : 'down',
-          },
-        };
-      },
+      () => this.dalHealthIndicator.isHealthy(),
       async () => {
         return {
           apiVersion: {
--- a/apps/worker/src/app/workflow/services/metric-queue.service.ts
+++ b/apps/worker/src/app/workflow/services/metric-queue.service.ts
@@ -18,6 +18,9 @@ function sumOfSquares(values: number[]): number {
   return sum(values.map((value) => (value - meanVal) ** 2));
 }
 
+const LOG_CONTEXT = 'MetricQueueService';
+const METRIC_JOB_ID = 'metric-job';
+
 @Injectable()
 export class MetricQueueService extends QueueService<Record<string, never>> {
   constructor(@Inject('BULLMQ_LIST') private token_list: QueueService[]) {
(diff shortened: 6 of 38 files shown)


apps/api/src/app/health/health.controller.ts
```diff
@@ -3,17 +3,23 @@ import { ApiExcludeController } from '@nestjs/swagger';
...
 @Controller('health-check')
 @ApiExcludeController()
 export class HealthController {
   constructor(
     private healthCheckService: HealthCheckService,
     private dalService: DalService,
-    private cacheHealthIndicator: CacheServiceHealthIndicator
+    private cacheHealthIndicator: CacheServiceHealthIndicator,
+    private inMemoryHealthIndicator: InMemoryProviderServiceHealthIndicator,
+    private triggerQueueHealthIndicator: TriggerQueueServiceHealthIndicator
```
p-fernandez, line 22: First steps to be able to complete in a different PR the checks for the graceful bootstrap.

apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
```diff
@@ -4,7 +4,7 @@ import { getRedisPrefix } from '@novu/shared';
 import { InboundEmailParse } from '../usecases/inbound-email-parse/inbound-email-parse.usecase';
 import { InboundEmailParseCommand } from '../usecases/inbound-email-parse/inbound-email-parse.command';
 import { ConnectionOptions } from 'tls';
-import { BullmqService } from '@novu/application-generic';
+import { BullMqService } from '@novu/application-generic';
```
p-fernandez, line 7: Renaming based on pattern.

apps/api/src/app/shared/constants.ts
```diff
@@ -1,3 +1,2 @@
-export const QUEUE_SERVICE = 'QueueService';
```
p-fernandez, line 1: Unused. Leftover from refactor to application-generic.

apps/api/src/app/shared/shared.module.ts
```diff
@@ -118,7 +120,6 @@ const PROVIDERS = [
       return dalService;
     },
   },
-  cacheService,
```
p-fernandez, line 121: Duplication.

apps/worker/src/app/workflow/services/workflow-queue.service.ts
```diff
@@ -55,14 +55,6 @@ export class WorkflowQueueService extends QueueService<IJobData> {
     });
   }
 
-  public async gracefulShutdown() {
-    // Right now we only want this for testing purposes
-    if (process.env.NODE_ENV === 'test') {
-      await this.bullMqService.queue.drain();
-      await this.bullMqService.worker.close();
-    }
-  }
```
p-fernandez, line 64: Used just for testing. Now this will be implemented in the BullMqService that this service is dependant.

apps/ws/src/shared/constants.ts
```diff
@@ -1,2 +1 @@
-export const QUEUE_SERVICE = 'QueueService';
```
p-fernandez, line 1: Unused. Leftover from refactor to application-generic.

packages/application-generic/src/health/queue.health-indicator.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor(private queueService: QueueService) {
+    super();
+  }
+
+  async isHealthy(): Promise<HealthIndicatorResult> {
+    const runningStatus =
+      await this.queueService.bullMqService.getRunningStatus();
+
+    if (!runningStatus.queueIsPaused && runningStatus.workerIsRunning) {
```
p-fernandez, line 22: Need to review this condition.

packages/application-generic/src/health/trigger-queue.health-indicator.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor(private triggerQueueService: TriggerQueueService) {
+    super();
+  }
+
+  async isHealthy(): Promise<HealthIndicatorResult> {
+    const runningStatus =
+      await this.triggerQueueService.bullMqService.getRunningStatus();
+
+    if (!runningStatus.queueIsPaused) {
```
p-fernandez, line 22: We only create a queue for the TriggerQueueService.

packages/application-generic/src/health/ws-queue.health-indicator.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  constructor(private wsQueueService: WsQueueService) {
+    super();
+  }
+
+  async isHealthy(): Promise<HealthIndicatorResult> {
+    const runningStatus =
+      await this.wsQueueService.bullMqService.getRunningStatus();
+
+    if (!runningStatus.queueIsPaused && runningStatus.workerIsRunning) {
```
p-fernandez, line 22: Need to review this condition.

packages/application-generic/src/services/bull-mq.service.ts
```diff
@@ -114,4 +117,50 @@ export class BullmqService {
         : {}),
     });
   }
+
+  public async gracefulShutdown(): Promise<void> {
+    Logger.log('Shutting the BullMQ service down', LOG_CONTEXT);
+
+    await this._queue.close();
```
p-fernandez, line 124: Might we want to check if `this._queue` exists in case BullMqService is only used to create a worker and not a queue in another service dependant on this.

packages/application-generic/src/services/ws-queue.service.ts
```diff
@@ -1,18 +1,21 @@
...
+
+@Injectable()
 export class WsQueueService extends QueueService<Record<string, never>> {
   public static queueName = 'ws_socket_queue';
+
   constructor() {
     super('ws_socket_queue');
   }
 
-  public readonly bullMqService: BullmqService;
```
p-fernandez, line 14: This is set by `QueueService` so this shouldn't be needed here.

BiswaViraj (approved):
🎉

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -33,47 +36,44 @@ export class MetricQueueService extends QueueService<Record<string, never>> {
...
 
-  public async gracefulShutdown() {
-    // Right now we only want this for testing purposes
-    if (process.env.NODE_ENV === 'test') {
-      await this.bullMqService.queue.drain();
-      await this.bullMqService.worker.close();
-    }
+        return undefined;
+      })
+      .catch((error) => Logger.error('Metric Job Exists function errored', LOG_CONTEXT, error));
```
p-fernandez, line 76: Catching to avoid unexpected unhandled exceptions.

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -88,58 +88,63 @@ export class MetricQueueService extends QueueService<Record<string, never>> {
...
+              Logger.log(`MetricQueueService/${queueService.name}/failed`, JSON.stringify(failMetric));
+              Logger.log(`MetricQueueService/${queueService.name}/waiting`, waitCount);
+              Logger.log(`MetricQueueService/${queueService.name}/delayed`, delayedCount);
+              Logger.log(`MetricQueueService/${queueService.name}/active`, activeCount);
+            }
           }
+
+          return resolve();
+        } catch (error) {
+          return reject(error);
```
p-fernandez, line 137: Catching to avoid unexpected unhandled exceptions. This could be logged only.

Cliftonz (approved):

packages/application-generic/src/services/queue.service.ts
```diff
@@ -31,14 +34,16 @@ export class QueueService<T = unknown> {
     });
   }
 
-  public async gracefulShutdown() {
-    // Right now we only want this for testing purposes
-    if (process.env.NODE_ENV === 'test') {
-      await this.bullMqService.queue.drain();
```
djabarovgeorge, line 37: With the new change we will be missing the 'queue.drain()', is that ok on the testing env?

packages/application-generic/src/health/trigger-queue.health-indicator.ts
p-fernandez: I wasn't able to find in the Bull MQ API any other way to know if the queue was running besides this property and to check if the queue was instantiated.

apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -33,47 +36,44 @@ export class MetricQueueService extends QueueService<Record<string, never>> {
...
-
-      return true;
-    });
-  }
+    )
+      .then(async (exists): Promise<void> => {
+        Logger.debug('metric job exists: ' + exists, LOG_CONTEXT);
+
+        if (!exists) {
+          Logger.debug(`metricJob doesn't exist, creating it`, LOG_CONTEXT);
```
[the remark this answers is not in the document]
p-fernandez: Maybe something to discuss in the tech summit.

packages/application-generic/src/services/bull-mq.service.ts
```diff
@@ -114,4 +117,50 @@ export class BullmqService {
...
+
+    Logger.log('Shutting down the BullMQ service has finished', LOG_CONTEXT);
+  }
+
+  public async getRunningStatus(): Promise<{
+    queueIsPaused: boolean;
+    workerIsRunning: boolean;
+  }> {
+    const queueIsPaused =
+      (this._queue && (await this._queue.isPaused())) || undefined;
```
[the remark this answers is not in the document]
p-fernandez: The problem is that BullMQ service is used both for creating queues and workers. We have some services that only create a queue and other that only create a worker. I wanted to do a generic functionality that showed the running status of the service that instantiates the BullMQ service. Therefore the undefined for the values to return to handle those cases. Open to a better suggestion though!

packages/application-generic/src/health/in-memory.health-indicator.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  async isHealthy(): Promise<HealthIndicatorResult> {
+    if (this.inMemoryProviderService.isClientReady()) {
+      return this.getStatus(this.INDICATOR_KEY, true);
+    }
+
+    throw new HealthCheckError(
+      'In-memory Health',
+      this.getStatus(this.INDICATOR_KEY, false)
```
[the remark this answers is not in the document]
p-fernandez: Would be a good future improvement. 👍🏻

apps/worker/src/app/workflow/services/metric-queue.service.ts
p-fernandez: I expect this rejection to be caught by the `failed` event (line 35) and to be logged through `jobHasFailed` function. But not 100% sure.

packages/application-generic/src/services/queue.service.ts
p-fernandez: I was the one implementing that as a hack to make one Workflow Queue test pass. Seems that hack is not needed anymore.

apps/api/src/app/health/health.controller.ts
```diff
@@ -1,32 +1,33 @@
...
     const checks: HealthIndicatorFunction[] = [
-      async () => {
-        return {
-          db: {
-            status: this.dalService.connection.readyState === 1 ? 'up' : 'down',
-          },
-        };
-      },
+      () => this.dalHealthIndicator.isHealthy(),
+      () => this.inMemoryHealthIndicator.isHealthy(),
```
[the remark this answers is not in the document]
p-fernandez: NestJS Terminus package takes care of that behind the scenes: https://github.com/nestjs/terminus/blob/410e07bf5e96d38285bc244225563d633ee1b2b5/lib/health-check/health-check-executor.service.ts#L64

apps/worker/src/app/health/health.controller.ts
```diff
@@ -1,24 +1,18 @@
...
 
 @Controller('health-check')
 export class HealthController {
-  constructor(private healthCheckService: HealthCheckService, private dalService: DalService) {}
+  constructor(private healthCheckService: HealthCheckService, private dalHealthIndicator: DalServiceHealthIndicator) {}
 
   @Get()
   @HealthCheck()
-  healthCheck() {
+  healthCheck(): Promise<HealthCheckResult> {
```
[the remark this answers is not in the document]
p-fernandez: Yes. I was planning to do it when tackling the implementation of MemoryDB for the Worker but maybe it is better I totally do it in this PR.

apps/worker/src/app/workflow/services/metric-queue.service.ts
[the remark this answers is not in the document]
djabarovgeorge: Added :)

packages/application-generic/src/services/bull-mq.service.ts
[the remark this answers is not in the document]
djabarovgeorge: Obviously not for this cycle but maybe we can think of splitting the logic into two classes that extend this service one producer and one consumer.

djabarovgeorge (approved):
💫

apps/api/src/app/shared/shared.module.ts
```diff
@@ -106,6 +107,7 @@ const distributedLockService = {
 const PROVIDERS = [
   cacheService,
   distributedLockService,
+  inMemoryProviderService,
```
LetItRock, line 110: Is there any reason why we are creating 2 instances of the `InMemoryProviderService`?
we can inject it like this: 
```
{
    provide: CacheService,
    useFactory: (inMemoryProviderService: InMemoryProviderService) => {
      return new CacheService(inMemoryProviderService);
    },
    inject: [InMemoryProviderService],
  },
```

apps/api/src/app/shared/shared.module.ts
```diff
@@ -141,6 +142,7 @@ const PROVIDERS = [
       return analyticsService;
     },
   },
+  TriggerQueueService,
```
LetItRock, line 145: isn't this done by mistake? we are using the `TriggerHandlerQueueService` in the `events.module`... we can change it to inject the `TriggerQueueService` instead if you wish ;)

apps/worker/src/app/workflow/services/metric-queue.service.ts
LetItRock: @p-fernandez @djabarovgeorge yes, this is what will happen after `reject`

apps/worker/src/bootstrap.ts
```diff
@@ -64,6 +64,9 @@ export async function bootstrap(): Promise<INestApplication> {
   app.use(bodyParser.json());
   app.use(bodyParser.urlencoded({ extended: true }));
 
+  // Starts listening for shutdown hooks
+  app.enableShutdownHooks();
```
LetItRock, line 68: you told me some time ago to remove this 😛 hahah

packages/application-generic/src/services/bull-mq.service.ts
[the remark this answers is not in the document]
LetItRock: IMO we do always call `createQueue` when initializing any queue, and the workers are tied to the queue, and won't do anything without it... so my suggestion is to move the queue creation to the constructor of the `BullMqService`, but of course we can do this in the separate ticket in the cooldown...

packages/application-generic/src/health/trigger-queue.health-indicator.ts
LetItRock: we should also check here is the worker is running by using the flag that you do return

apps/api/src/app/shared/shared.module.ts
p-fernandez: We are configuring the provider with auto pipelining for the CacheService and without for the DistributedLockService. That is one of the reasons. 🙁

apps/api/src/app/shared/shared.module.ts
p-fernandez: I recall it was NestJS complaining as not set as provider due the health indicators. Let me try to remove `TriggerHandlerQueueService` as dependency and see it doesn't complain. But I have a feeling that will require it too.

apps/worker/src/bootstrap.ts
p-fernandez: I don't remember it 🙈 . This is needed for the graceful shutdown hooks to operate after the SIGTERM so I would need the context on why I could have said that. Maybe I mentioned what NestJS says about not optimal performance when enabled it?

packages/application-generic/src/services/bull-mq.service.ts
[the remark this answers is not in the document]
p-fernandez: I ignore the reasons why it was done this way and not like you suggest. Maybe @davidsoderberg could give more context.
Though I can infer from the PR he did that the intention was to provide flexibility to create multiple queues independently as we will only have one worker that will consume them all. If we move all the queue creation to the BullMQ service it would make harder to be able to configure every queue independently as we have implemented right now.
Also to be able to pass a `groupId` that at the end is to take advantage of BullMQ grouping feature when consuming jobs, that was one of the performance improvement actions we did in previous release.

apps/api/src/app/shared/shared.module.ts
p-fernandez: I have just tried and we need this because the HealthModule where the health indicators are used depend on this SharedModule and TriggerQueueService is a dependency for TriggerQueueServiceHealthIndicator.
How would you suggest to do it instead? I can't think right now of other choice we have and also not sure what I understand from your suggestion would work.

packages/application-generic/src/health/trigger-queue.health-indicator.ts
p-fernandez: Unfortunately we can't do it as only a queue is created in `TriggerQueueService` therefore the worker comes as undefined. 🙁

PR #3331, opened 2023-05-04
Nv 2197 create UI for design for timed digest

davidsoderberg:
Screenshot 2023-05-04 at 11 30 15
Screenshot 2023-05-04 at 11 30 22
Screenshot 2023-05-04 at 11 30 28
Screenshot 2023-05-04 at 11 30 35
Screenshot 2023-05-04 at 11 30 42
Screenshot 2023-05-04 at 11 30 48

### DIFF
--- a/apps/web/src/pages/templates/workflow/digest/SentHeader.tsx
+++ b/apps/web/src/pages/templates/workflow/digest/SentHeader.tsx
@@ -1,0 +1,107 @@
+import { DigestTypeEnum, DigestUnitEnum } from '@novu/shared';
+import { useCallback } from 'react';
+import { useFormContext } from 'react-hook-form';
+import { colors } from '../../../../design-system';
+import * as capitalize from 'lodash.capitalize';
+
+const getOrdinal = (num: string | number) => {
+  if (typeof num === 'string') {
+    num = parseInt(num, 10);
+  }
+  const ord = ['st', 'nd', 'rd'];
+  const exceptions = [11, 12, 13];
+  let nth = ord[(num % 10) - 1];
+
+  if (ord[(num % 10) - 1] == undefined || exceptions.includes(num % 100)) {
+    nth = 'th';
+  }
+
+  return num + nth;
+};
+
+const Highlight = ({ children }) => (
+  <b
+    style={{
+      color: colors.B80,
+    }}
+  >
+    {children}
+  </b>
+);
+
+export const SentHeader = ({ index }) => {
+  const { watch } = useFormContext();
+
+  const type = watch(`steps.${index}.metadata.type`);
+  const unit = watch(`steps.${index}.metadata.unit`);
+  const backoff = watch(`steps.${index}.metadata.backoff`);
+  const at = watch(`steps.${index}.metadata.timed.at`);
+  const daysOfWeek = watch(`steps.${index}.metadata.timed.dayOfWeek`) || [];
+  const day = watch(`steps.${index}.metadata.timed.day`) || [];
+  const every = watch(`steps.${index}.metadata.timed.every`);
+  const amount = watch(`steps.${index}.metadata.amount`);
+  const BackoffText = useCallback(() => {
+    return backoff ? (
+      <>
+        only if <Highlight>frequent events</Highlight> occur
+      </>
+    ) : null;
+  }, [backoff]);
+
+  if (type === DigestTypeEnum.TIMED && unit == DigestUnitEnum.HOURS) {
+    return (
+      <>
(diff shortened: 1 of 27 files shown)


LetItRock (commented):
hey 👋 I've found a few things to fix:

1. The min tab value is not selected, and I feel like we should provide here a default every value, and we should have a FE validation logic for min/max (it partially works now).
Screenshot 2023-05-05 at 00 26 26
2. Hourly, similar default every value and the min/max validation logic.
3. Daily, similar default every value and the min/max validation logic. When every is ex. 13, then the label should be like: `Every 13th day`. If the time value is not selected the `at` part should not be shown.
Screenshot 2023-05-05 at 00 27 26
4. Weekly, similar default every value and the min/max validation logic. And I feel like we should select by default the current weekday, because daily without selected days means - every week at the current day.
Screenshot 2023-05-05 at 00 28 27
5. Monthly, similar default every value and the min/max validation logic. Similarly, I think we should select the current month's day.
Screenshot 2023-05-05 at 00 30 12
6. When the "On the" monthly option is selected the label is not updated.
Screenshot 2023-05-05 at 00 30 36

apps/web/src/pages/templates/workflow/digest/SentHeader.tsx
```diff
@@ -0,0 +1,112 @@
...
+  <b
+    style={{
+      color: colors.B80,
+    }}
+  >
+    {children}
+  </b>
+);
+
+export const SentHeader = ({ index }) => {
```
LetItRock, line 37: maybe `WillBeSentHeader` ?

apps/web/src/pages/templates/workflow/digest/SentHeader.tsx
```diff
@@ -0,0 +1,112 @@
...
+        <Highlight>Weekly</Highlight> on <Highlight>{daysOfWeek?.map((item) => capitalize(item)).join(', ')}</Highlight>{' '}
+        at <Highlight>{at}</Highlight>{' '}
+        <div>
+          <BackoffText />
+        </div>
+      </>
+    );
+  }
+
+  if (unit === DigestUnitEnum.MONTHS) {
```
LetItRock, line 91: a little above you have an additional check `type === DigestTypeEnum.TIMED` did you miss it here?

apps/web/src/pages/templates/workflow/digest/WeekDaySelect.tsx
```diff
@@ -0,0 +1,99 @@
...
+        overflow: 'hidden',
+        marginTop: 24,
+        opacity: disabled ? 0.4 : 1,
+      }}
+    >
+      {items.map((day, index) => {
+        return (
+          <Day
+            label={day.label}
+            value={day.value}
```
LetItRock, line 82: missed key prop

apps/web/src/pages/templates/workflow/digest/WeekDaySelect.tsx
```diff
@@ -0,0 +1,99 @@
+import { Group, UnstyledButton } from '@mantine/core';
+import { DaysEnum } from '@novu/shared';
+import { useEffect, useState } from 'react';
```
LetItRock, line 3: please update the imports to what we agreed some time ago: first library imports, then our monorepo, then relative

apps/web/src/pages/templates/workflow/digest/WeekDaySelect.tsx
```diff
@@ -0,0 +1,99 @@
...
+    value: DaysEnum.SUNDAY,
+  },
+];
+
+export const WeekDaySelect = ({ onChange = (value: DaysEnum[]) => {}, value = [], disabled = false }) => {
+  const [selected, setSelected] = useState<DaysEnum[]>(value);
+
+  useEffect(() => {
+    onChange(selected);
+  }, [selected]);
```
LetItRock, line 63: this component should be controlled component, and you don't need to have a state and `useEffect` here

LetItRock (commented):
the `timed` metadata should be updated with appropriate values when submitting it to the BE, now we do send both weekdays and month days


https://user-images.githubusercontent.com/2607232/236349384-7cf13392-8462-4c52-a804-fd20df4addd3.mov

apps/web/src/pages/templates/workflow/digest/WeekDaySelect.tsx
davidsoderberg: a great catch we should have a lint for this...

LetItRock (approved):

Issue #3234, opened 2023-04-18
[NV-2136] 🐛 Bug Report: Novu Provider is not working properly.


xiangdev1016:
### 📜 Description
export const Notification = () => {
  const wallet = useWallet(FANTOM);

  return wallet.connected ? (
    <NovuProvider
      subscriberId={wallet.address.toLocaleLowerCase()}
      applicationIdentifier={NOVU_APP_ID}
      initialFetchingStrategy={{ fetchNotifications: true }}
    >
      <CustomNotificationCenter />
    </NovuProvider>
  ) : (
    <></>
  );
};

This is my code snippet that retrieves a wallet address. When I change the wallet, I can obtain a different wallet address.

I am currently utilizing the wallet address as the subscribeId. It works correctly the first time. As you can see, I'm using wallet address for each subscribeId

However, when I change the wallet, the notification payload data does not update based on the new subscribeId. I have checked the onLoad function in NovuProvider and it is only called once.

Additionally, I am using the useNotification() hook function on the CustomNotificationCenter component using the @novu/notification-center library.

I have attempted to use the refetch function in useNotification, which works, but the NovuProvider is not functioning as expected.

Are there any other options to manually refetch notification payload  data based on subscribeId using useNotification or other methods?

### 👟 Reproduction steps

1. connect wallet
2. fetch notification first time
3. change wallet
At that time, it's not fetching notification payload data from novu provider

### 👍 Expected behavior

It should be clear the old subscriber's notifications, and then have to be fetch new notification payload data based on new wallet.

### 👎 Actual Behavior with Screenshots

![image](https://user-images.githubusercontent.com/94132651/232785527-d8e3e6a7-4d3b-4008-96c0-9c3f628985fd.png)

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-2136](https://linear.app/novu/issue/NV-2136/🐛-bug-report-novu-provider-is-not-working-properly)</sub>

LetItRock:
hey @xiangdev1016! 👋 I do confirm that the issue exists. 
The session is initialized, but we don't refetch the organization info, unseen notifications, and the feed.

@xiangdev1016 do you want to contribute?

gitstart:
[...]

PR #3338, opened 2023-05-04
NV-2136 - 🐛 Bug Report: Novu Provider is not working properly.

gitstart:
 Refetches organizations info, unseen count and notifications feeds when `subscriberId` or `applicationIdentifier` changes.

Closes: https://github.com/novuhq/novu/issues/3234

### Other information

---


gitstart:

packages/notification-center/src/hooks/useFetchOrganization.ts
```diff
@@ -1,17 +1,32 @@
...
+  const [intialize, setInitialize] = useState(false);
+  const queryClient = useQueryClient();
+
+  useEffect(() => {
+    if (intialize) {
+      queryClient.refetchQueries(UNSEEN_COUNT_QUERY_KEY, { exact: false });
+      queryClient.refetchQueries(FEED_UNSEEN_COUNT_QUERY_KEY, { exact: false });
+      queryClient.refetchQueries(ORGANIZATION_QUERY_KEY, { exact: false });
+    }
+  }, [subscriberId, applicationIdentifier]);
```
LetItRock, line 21: This hook is supposed to be only responsible for fetching the organization. Now it has additional logic that refetches other queries after we fetched org. I think this is not a good place for this logic to be and I wouldn't expect that if I will debug it. And why do we only refetch these and not for ex. the user preferences?

---

I haven't spend much time digging into the code, but I feel like we should do this in another way. 
In simple words, we just need to update the "query keys" with the `subscriberId, subscriberHash, applicationIdentifier` like the `useSession` does, so any changes to these fields will create the new "query keys" resulting in new fetches for all queries (if the cache is stale).

Please ensure that all the query hooks have these fields in the keys and that the WS session is also re-initialized.

gitstart:
@LetItRock thanks for the feedback, your comment has been implemented.

LetItRock (approved):
amazing work! 🥇

jainpawan21:
@gitstart few tests are failing.
Can you please look into these failing tests?

packages/notification-center/src/hooks/useInitializeSocket.ts
```diff
@@ -11,6 +12,7 @@ type IUseInitializeSocket = (args: { socketUrl: string }) => {
 
 export const useInitializeSocket: IUseInitializeSocket = ({ socketUrl }) => {
   const socketRef = useRef<Socket | null>(null);
+  const setQueryKey = useSetQueryKey();
```
ainouzgali, line 15: Hi @gitstart I believe your problem is coming from here. `useInitializeSocket` is used in NovuProvider.ts but `useSetQueryKey` uses the context of that provider.

gitstart: I have effected the changes @ainouzgali

ainouzgali (approved):
Thank you @gitstart ! This looks great!

donnguyen:
Hi guys, just wonder when will this fix be released?

Thanks again for the great work!

ainouzgali:
Hi @donnguyen, this was included in the release of v0.16.0 which was deployed 2 days ago :) 
Could you please try it out and let us know if it works out for you?

donnguyen:
Hi @ainouzgali, thanks for getting back to me. I have just tried the new version, but seems it hasn't solved the whole issues yet.

The issue now I can find is that the notification/organization/unseen calls might happen before the initialise (where we get and set token), and if the case happens -> these API calls still get data of the old/previous subscriber/application.

Issue #3342, opened 2023-05-05
[NV-2222] 🐛 Bug Report: mailgun baseurl is missing in handler


raheelriax:
### 📜 Description

I was trying to set up Mail gun integration with Novu, and the integration got created OK, but when I tried to send emails, it consistently failed. The error was also a bit misleading as it always shows `providerId` as null or `SendGrid` whereas I only had one integration which was `Mailgun`.
I looked into the issue and [found out](https://stackoverflow.com/questions/63489555/mailgun-401-forbidden/63967799#63967799) that the domains in the `EU Region` need to specify the EU `host`. I looked into the codebase and found some issues

1. [mailgun-provider](https://github.com/novuhq/novu/blob/next/providers/mailgun/src/lib/mailgun.provider.ts#L35) does take custom `url` as a `baseUrl` argument and pass that on to `mailgun-client` but the [mailgun-handler](https://github.com/novuhq/novu/blob/next/packages/application-generic/src/factories/mail/handlers/mailgun.handler.ts#L24) does not pass `baseUrl`  to the `mailgun-provider`. 

2. Integration Create API does not let you specify `baseUrl`, but it's defined in the interface [here](https://github.com/novuhq/novu/blob/next/libs/dal/src/repositories/integration/integration.entity.ts#L26)

### 👟 Reproduction steps

1. Setup a Mailgun account with a `EU` region domain
2. Setup Mailgun integration on Novu.
3. Try to send an email, you will get an error.

### 👍 Expected behavior

- Should let you define custom host URL in case of `EU` 
- The mail should be sent

### 👎 Actual Behavior with Screenshots

I get this error in my last step of `executionDetails` of the `job`

```json
{
  "detail": "Unexpected provider error",
  "providerId": "sendgrid",
  "raw": "{\"stack\":\"\",\"status\":401,\"message\":\"Unauthorized\",\"details\":\"Forbidden\"}",
  "source": "INTERNAL",
  "status": "FAILED",
  "__typename": "NotificationActivityJobExecutionDetailsModel"
}

```

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

[...]

PR #3344, opened 2023-05-05
feat(integrations): Add baseUrl to mailgun-handler

raheelriax:
modify the integration credentials DTO to include the baseUrl property.

NV-2222 #3342

Fixes https://github.com/novuhq/novu/issues/3342

To let users with `EU` region domain use the mail gun integration smoothly

### DIFF
--- a/packages/application-generic/src/factories/mail/handlers/mailgun.handler.ts
+++ b/packages/application-generic/src/factories/mail/handlers/mailgun.handler.ts
@@ -14,10 +14,12 @@ export class MailgunHandler extends BaseHandler {
       username: string;
       domain: string;
       from: string;
+      baseUrl?: string;
     } = {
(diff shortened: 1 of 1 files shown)


apps/api/src/app/integrations/dtos/credentials.dto.ts
```diff
@@ -43,6 +43,4 @@ export class CredentialsDto {
   tlsOptions?: Record<string, unknown>;
   @ApiPropertyOptional()
   baseUrl?: string;
-  @ApiPropertyOptional()
```
jainpawan21, line 46: Hi @raheelriax 
Any reason of removing this property?

jainpawan21 (changes requested):

p-fernandez (approved):
🌟

PR #3346, opened 2023-05-06
[WIP] feat(api,web,worker): update digest metadata types

LetItRock:

**Still WIP:**

In this PR:
- updated digest metadata types, now we will use the same interface between all the places in the code
- updated swagger API step types for the create/update notification template endpoints
- fixed some form issues for timed digest after saving the template
- do not send the unnecessary fields in the request when saving the template with timed digest


apps/worker/src/app/workflow/usecases/send-message/digest/digest.usecase.ts
```diff
@@ -58,7 +58,10 @@ export class Digest extends SendMessageType {
   private async getEvents(command: SendMessageCommand) {
     const currentJob = await this.jobRepository.findOne({ _environmentId: command.environmentId, _id: command.jobId });
 
-    if (currentJob?.digest?.type === DigestTypeEnum.BACKOFF || currentJob?.digest?.backoff) {
+    if (
+      currentJob?.digest?.type === DigestTypeEnum.BACKOFF ||
+      (currentJob?.digest && 'backoff' in currentJob?.digest)
```
davidsoderberg, line 63: Will not this be true even if backoff is false?

apps/web/src/pages/templates/components/templateToFormMappers.ts
```diff
@@ -49,8 +49,6 @@ const mapDigestStep = (item: IStepEntity): IStepEntity => ({
   ...item,
   metadata: {
     ...item.metadata,
-    type: DigestTypeEnum.REGULAR,
```
davidsoderberg, line 52: This will break old backoff steps

apps/web/src/pages/templates/workflow/DigestMetadata.tsx
```diff
@@ -232,7 +232,8 @@ export const DigestMetadata = ({ control, index, readonly }) => {
                     <span>at</span>
                     <Controller
                       control={control}
-                      name={`steps.${index}.metadata.timed.at`}
+                      name={`steps.${index}.metadata.timed.atTime`}
+                      shouldUnregister
```
davidsoderberg, line 236: I guess this remove data if it is not selected?

apps/web/src/pages/templates/components/templateToFormMappers.ts
LetItRock: can you say more about why? the previous code was also breaking things... after you saved the scheduled digest and did a page refresh it erased that information and set it to regular with backoff, which is not true

apps/web/src/pages/templates/workflow/DigestMetadata.tsx
LetItRock: yes it removes it from the form timed metadata object

apps/worker/src/app/workflow/usecases/send-message/digest/digest.usecase.ts
LetItRock: 🎖️ good catch

apps/web/src/pages/templates/components/templateToFormMappers.ts
davidsoderberg: Because if we do not add the backoff: true if it has type backoff since before it will not open backoff panel as it should...

LetItRock: so we have to do that by checking the "old" backoff type

davidsoderberg (approved):

davidsoderberg (approved):

davidsoderberg (approved):

PR #3348, opened 2023-05-07
Remove unnecessary find notification

ainouzgali:

- Remove notification.find for each send message.
- Remove duplicate find organization.
- Add instrument tags.


apps/worker/src/app/workflow/usecases/send-message/send-message-type.usecase.ts
```diff
@@ -21,7 +21,6 @@ export abstract class SendMessageType {
     errorId: string,
     errorMessageFallback: string,
     command: SendMessageCommand,
-    notification: NotificationEntity,
```
ainouzgali, line 24: notification was not used in send error

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -182,11 +176,11 @@ export class SendMessageChat extends SendMessageBase {
     }
 
     const message: MessageEntity = await this.messageRepository.create({
-      _notificationId: notification._id,
+      _notificationId: command.notificationId,
       _environmentId: command.environmentId,
       _organizationId: command.organizationId,
       _subscriberId: command._subscriberId,
-      _templateId: notification._templateId,
+      _templateId: command.templateId,
```
scopsy, line 183:
```suggestion
      _templateId: command._templateId,
```

For brewity would suggest to add the "_" for all internal id's we having. This has caused some confusion previously specifically with regard to `subsriberId`

davidsoderberg (approved):

PR #3353, opened 2023-05-08
feat: add sharable url to the integration store

djabarovgeorge:
The Sharable URL is the URL that our users will need to share in their product in order to invite subscribers to provide access to send messages to their channel.

In order to help the users with the Sharable URL generation.

Short loom how it looks in the dashboard
https://www.loom.com/share/4d6469c9df904a43a063ca31012417c1

### DIFF
--- a/apps/web/src/pages/integrations/components/Modal/ConnectIntegrationForm.tsx
+++ b/apps/web/src/pages/integrations/components/Modal/ConnectIntegrationForm.tsx
@@ -1,11 +1,17 @@
-import { useEffect, useReducer, useRef } from 'react';
-import { Controller, useForm } from 'react-hook-form';
-import styled from '@emotion/styled/macro';
+import { useEffect, useReducer, useRef, useState } from 'react';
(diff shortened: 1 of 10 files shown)


apps/web/src/pages/integrations/components/Modal/ConnectIntegrationForm.tsx
```diff
@@ -460,3 +507,36 @@ const CenterDiv = styled.div`
...
+          <Input
+            label="Sharable URL"
+            value={oauthUrl}
+            rightSection={
+              <CopyWrapper onClick={() => oauthUrlClipboard.copy(oauthUrl)}>
+                {oauthUrlClipboard.copied ? <Check /> : <Copy />}
+              </CopyWrapper>
+            }
+            disabled={true}
+            description={'Use the URL below to share your app with any Slack workspace.'}
```
djabarovgeorge, line 536: I wonder if we need to update the copy here and add more information.
because this sharable URL is pointing to Novu API handler.

Issue #3255, opened 2023-04-20
🐛 Bug Report: useFeedUnseenCount shows total feed count


david-morris:
### 📜 Description

Despite the name and documentation, `useFeedUnseenCount` returns the total count associated with a query.

If you want to fetch the unseen count for a feed, you need to add `seen: false` to the query.

This is a minor annoyance for people who want to keep IStores as constants in some central place.

### 👟 Reproduction steps

1. send one notification in an empty feed and read it
2. call `useFeedUnseenCount({ query: { feedIdentifier: <feed identifier> } })`
3. call `useFeedUnseenCount({ query: { feedIdentifier: <feed identifier>, seen: false } })`

### 👍 Expected behavior

Something called `useFeedUnseenCount` should count _unseen_ items in a feed.

### 👎 Actual Behavior with Screenshots

If you don't add `seen: false` to the query, you see 1 item instead of the correct 0 in `result.data.count`.

### 📃 Provide any additional context for the Bug.

Since this has hit live, it makes sense to solve in the interim before a major version update with a note on the documentation and ideally in a `/** description */` if it's possible to get that into the `dist/`.

Since this allows functionality that the described behavior doesn't, it probably makes sense to keep the existing hook as `useQueryCount` and export a new `useFeedUnseenCount` as a wrapper.

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

gitstart:
@david-morris I would like to pick this up cc @scopsy

scopsy:
@gitstart just assigned you 🙏

PR #3358, opened 2023-05-08
NV-2154 - 🐛 Bug Report: useFeedUnseenCount shows total feed count

gitstart:
Updates the `useFeedUnseenCount` hook to return count for unseen items in a feed without having to pass `seen: false` to the query.

Despite the name and documentation, `useFeedUnseenCount` returns the total count associated with a query.
If you want to fetch the unseen count for a feed, you need to add `seen: false` to the query.
closes https://github.com/novuhq/novu/issues/3255

### Other information

---


gitstart:

packages/notification-center/src/hooks/useFeedUnseenCount.ts
```diff
@@ -10,10 +10,11 @@ export const useFeedUnseenCount = (
   options: UseQueryOptions<ICountData, Error, ICountData> = {}
 ) => {
   const { apiService, isSessionInitialized } = useNovuContext();
+  const unseenQuery = { ...query, seen: query.seen ?? false };
```
p-fernandez, line 13: I understand this was trying to keep retrocompatibility but we can safely ignore the query passed and just set the `unseenQuery` to `seen: false`.
What I would do then is explain in the documentation ([docs/docs/notification-center/react/api-reference.md](https://github.com/novuhq/novu/blob/next/docs/docs/notification-center/react/api-reference.md#usefeedunseencount)) that from earlier versions of `v.0.15.0` (as this would be merged in that version) as deprecated the query needs to be passed but from that version forward is not needed, in the same way as `useUnseenCount` works. Therefore users can slowly clean their code at their pace but withouth the functionality be affected.
@scopsy @unicodeveloper what's your take on this?

gitstart:
@p-fernandez your comment has been addressed, thanks for the feedback.

p-fernandez (approved):

p-fernandez:
Seems this change was somehow applied from a different PR while this was on review. Thanks for your contribution. 👏🏻

PR #3361, opened 2023-05-09
fix: sent header

davidsoderberg:


apps/web/src/pages/templates/workflow/digest/WillBeSentHeader.tsx
```diff
@@ -61,6 +61,16 @@ export const WillBeSentHeader = ({ index }) => {
...
   const amount = watch(`steps.${index}.metadata.amount`);
 
+  useEffect(() => {
+    if (!backoff) {
+      return;
+    }
+    if (type !== DigestTypeEnum.TIMED) {
+      return;
+    }
+    setValue(`steps.${index}.metadata.backoff`, false);
```
LetItRock, line 71: maybe these fields should be set when creating a new node? but not sure if it will help

LetItRock (approved):

apps/web/src/pages/templates/workflow/digest/WillBeSentHeader.tsx
davidsoderberg: It will not because this is when switching type and not on create...

PR #3365, opened 2023-05-09
fix(provider): Remove sender name field and make boolean type a switch in nodemailer

### LINKED ISSUE #1753: 🐛 Bug Report: nodemailer sendername field & secure field
### 📜 Description

1. Nodemailer has a `Sender name` field in UI, even though it's not used in 
https://github.com/novuhq/novu/blob/2ee01b8c97d77b9f3295e46720385c137a0bfd57/providers/nodemailer/src/lib/nodemailer.provider.ts
image

2. Secure field is a textfield, therefore not intuitive, user should leave this field empty if it's `false` and provide some value if `true`

### 👟 Reproduction steps

1. Go to integration Store
2. Click on Nodemailer integration

- Check https://github.com/novuhq/novu/blob/2ee01b8c97d77b9f3295e46720385c137a0bfd57/providers/nodemailer/src/lib/nodemailer.provider.ts

### 👍 Expected behavior

1. There should be no `Sender name` field in UI.
2. Secure field should be checkbox or switch 

image
image

### 💻 Operating system

MacOS

### 🤖 Node Version

v16.14.2

From my point of view it's better to rename it to custom SMTP connection and not Nodemailer :)

peoray:

Closes #1753


libs/shared/src/consts/providers/credentials/provider-credentials.ts
```diff
@@ -181,7 +181,7 @@ export const nodemailerConfig: IConfigCredentials[] = [
     type: 'string',
     required: false,
   },
-  ...mailConfigBase,
+  mailConfigBase[0],
```
p-fernandez, line 184: I consider we shouldn't remove the option to set the sender name from Nodemailer. If it is not used in the provider package we should implement it. Probably the change for the credentials would be to make the sender name optional instead of mandatory as it is right now.

apps/web/src/pages/integrations/components/IntegrationInput.tsx
```diff
@@ -73,6 +73,28 @@ export function IntegrationInput({
...
+        placeholder={credential.displayName}
+        description={credential.description ?? ''}
+        data-test-id={credential.key}
+        error={errors[credential.key]?.message}
+        {...register(credential.key)}
+        checked={field.value}
+        onChange={field.onChange}
+      />
+    );
+  }
```
p-fernandez, line 96: This change will make all boolean values for all integrations to become a Switch. We should review all the providers to see nothing breaks in the functionality.
Adding screenshots of how they would look like now would be also helpful for the PR reviewers.

peoray: yes, although, it makes sense for a boolean to be a toggle rather than an input field. I know I have been confused by that when creating documentation for ([sparkpost](https://docs.novu.co/channels/email/sparkpost/)). 

The only providers affected are Sparkpost and Nodemailer (I checked). The others have a type of switch, which frankly I don't understand why it is so, `boolean` sounds like a better option since it's only true or false. (Note that this switch type is only included in the nodemailer config as well so it's the same behaviour as the boolean.

peoray:
Screen Shot 2023-05-09 at 10 30 54 AM
@p-fernandez

p-fernandez (approved):
🌟 
Well done!

p-fernandez:
@peoray could you do a `pnpm install` and `pnpm build` in your branch and commit the changes? The lock file is not working as expected and blocking the CI. 🙏🏻

peoray:
@p-fernandez done

PR #3371, opened 2023-05-09
feat: add novu sharable url usecase

djabarovgeorge:

New endpoint for requests from the user with an API key. this endpoint can reuse the same use case from the current logic use case.

For the user to use only one URL in order to set credentials for the chat channels under the same URL.


apps/api/src/app/subscribers/usecases/chat-oauth/chat-oauth.usecase.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  private async getClientId(command: ChatOauthCommand): Promise<string> {
+    const query: Partial<IntegrationEntity> & { _environmentId: string } = {
+      _environmentId: command.environmentId,
+      channel: ChannelTypeEnum.CHAT,
+      providerId: command.providerId,
+    };
+
+    const integration = await this.integrationRepository.findOne(query);
```
ainouzgali, line 32: @djabarovgeorge what would happen if here he finds a different chat integration and not slack? Also, does it matter here if its active or not?

apps/api/src/app/subscribers/subscribers.controller.ts
```diff
@@ -64,11 +64,13 @@ import { PaginatedResponseDto } from '../shared/dtos/pagination-response';
 import { GetSubscribersDto } from './dtos/get-subscribers.dto';
 import { GetInAppNotificationsFeedForSubscriberDto } from './dtos/get-in-app-notification-feed-for-subscriber.dto';
 import { ApiResponse } from '../shared/framework/response.decorator';
-import { HandleChatOauth } from './usecases/handle-chat-oauth/handle-chat-oauth.usecase';
-import { HandleChatOauthCommand } from './usecases/handle-chat-oauth/handle-chat-oauth.command';
 import { HandleChatOauthRequestDto } from './dtos/handle-chat-oauth.request.dto';
 import { LimitPipe } from '../widgets/pipes/limit-pipe/limit-pipe';
 import { OAuthHandlerEnum } from './types';
+import { ChatOauthCallback } from './usecases/chat-oauth-callback/chat-oauth-callcack.usecase';
```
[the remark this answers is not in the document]
djabarovgeorge: Good catch , I fixed it in the next PR 😅

apps/api/src/app/subscribers/usecases/chat-oauth/chat-oauth.usecase.ts
djabarovgeorge: I think it will be ok, Although we only support Slack at the moment this controller/usecase should be provider agnostic meaning we could add support to other providers as well.
Regarding the second question, you raised a good point, this implementation does not care if the integration is active or not, leaving the responsibility to the user.

ainouzgali: But if the query findOne returns the discord integration?

ainouzgali: sorry sorry

ainouzgali (approved):

PR #3375, opened 2023-05-10
fix: so analytic event is not created if the job is filtered out

davidsoderberg:
Moving of analytic track event `Process Workflow Step - [Triggers]`

Because the statistics will not be correct otherwise

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
@@ -63,6 +63,12 @@ export class SendMessage {
 
     const stepType = command.step?.template?.type;
 
+    if (!shouldRun.passed || !preferred) {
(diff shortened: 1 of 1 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -63,6 +63,12 @@ export class SendMessage {
 
     const stepType = command.step?.template?.type;
 
+    if (!shouldRun.passed || !preferred) {
```
LetItRock, line 66: as far as I remember we wanted to send that analytic event even if filters did not pass 
cc @scopsy

scopsy: @davidsoderberg i have mis understood your message earlier on Discord. The event should be still send even if the job is canceled. I was thinking you meant the other way in the message. Sorry for the confusion.

PR #3376, opened 2023-05-10
fix: Update node subscriber feed params typings

shanehandley-lived:

According to the OpenAPI documentation and testing, the `feedIdentifier` param is [optional](https://github.com/novuhq/novu/blob/next/apps/api/src/app/subscribers/dtos/get-in-app-notification-feed-for-subscriber.dto.ts#L9) when fetching a subscriber's notification feed.

The node client typings have this parameter as required. This updates the node client typings to reflect the API definition, so that the param can be excluded.

The `feedIdentifier` is an optional parameter, and the API accepts an empty object as the second argument to `getNotificationsFeed`, but it requires suppressing the following typescript error:

```
Argument of type '{}' is not assignable to parameter of type 'IGetSubscriberNotificationFeedParams'.
  Property 'feedIdentifier' is missing in type '{}' but required in type 'IGetSubscriberNotificationFeedParams'.ts(2345)
subscriber.interface.d.ts(27, 5): 'feedIdentifier' is declared here.
```

### DIFF
--- a/packages/node/src/lib/subscribers/subscriber.interface.ts
+++ b/packages/node/src/lib/subscribers/subscriber.interface.ts
@@ -44,6 +44,6 @@ export interface IUpdateSubscriberPreferencePayload {
 }
 export interface IGetSubscriberNotificationFeedParams {
   page?: number;
-  feedIdentifier: string;
+  feedIdentifier?: string;
   seen?: boolean;
 }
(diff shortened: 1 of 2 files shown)


p-fernandez (approved):
🌟 
Great addition with the tests.

PR #3378, opened 2023-05-10
feat(infra): improve cache service logging

p-fernandez:

Improves the logging and information provided in the logging for the CacheService.

We are experiencing problems with the Elasticache instances when load testing in DEV that are returning some auto pipelining errors in the Cache Service. As it is right now is hard to understand where the potential problem comes from so I am trying to improve the information we have at hand to be able to identify if there is any codebase problem that throws those errors at high loads, before digging deeper in the investigation of the infra resources.


ainouzgali (approved):
🚀

PR #3385, opened 2023-05-10
feat: 🚀 Feature — Add Web PushAPI support as Push Provider

### LINKED ISSUE #3459: 🚀 Feature: Add Web Push-API as a Push Provider for those that only build web apps
### 🔖 Feature description

Implement the Web Push-API standard to receive a push subscription generated by a client browser and to allow the sending to mobile or desktop devices through the browser's Push API.

[MDN Doc](https://developer.mozilla.org/en-US/docs/Web/API/Push_API)

I only build web applications and don't pay a vendor to process push requests. Implementing the browser push standard allows users to bypass any vendors and to directly send push notifications to supported browsers.

Add a Web Push-API provider using the create provider docs and implement the Push API standard.

### 🔄️ Additional Information

_No response_

johnmarksilly:

Add a new push provider to support the web [PushAPI standard](https://developer.mozilla.org/en-US/docs/Web/API/Push_API)

Support for alternative push method for browsers when external vendors aren't used.

Closes #3459

N/A


libs/shared/src/consts/providers/credentials/provider-credentials.ts
```diff
@@ -450,6 +450,24 @@ export const expoConfig: IConfigCredentials[] = [
   ...pushConfigBase,
 ];
 
+export const pushApiConfig: IConfigCredentials[] = [
+  {
+    key: CredentialsKeyEnum.PrivateKey,
```
djabarovgeorge, line 455: I wonder if we can reuse SecretKey instead of creating a new one, the name won't need to be the same just to have the same purpose.

libs/shared/src/consts/providers/credentials/secure-credentials.ts
```diff
@@ -6,4 +6,5 @@ export const secureCredentials: CredentialsKeyEnum[] = [
   CredentialsKeyEnum.Token,
   CredentialsKeyEnum.Password,
   CredentialsKeyEnum.ServiceAccount,
+  CredentialsKeyEnum.PrivateKey,
```
djabarovgeorge, line 9: Good catch adding this one here! :)

providers/push-api/src/lib/push-api.provider.ts
```diff
@@ -0,0 +1,112 @@
...
+  ) {}
+
+  async sendPush(
+    title: string,
+    content: string,
+    subscription: webpush.PushSubscription
+  ) {
+    const pushOptions: webpush.RequestOptions = {
+      vapidDetails: {
+        subject: 'mailto:support@novu.co',
```
djabarovgeorge, line 49: What is the purpose of this subject?

providers/push-api/src/lib/push-api.provider.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  async sendMessage(
+    options: IPushOptions
+  ): Promise<ISendMessageSuccessResponse> {
+    const subscriptions = options.target;
+
+    const failures: webpush.SendResult[] = [];
+    const id = this.getId();
+
+    for (const subscriptionString of subscriptions) {
```
djabarovgeorge, line 75: Why do we have their list of subscriptions? on this level, we should have the Novu subscriber as a target and it should be a single recipient.

djabarovgeorge (commented):
Great job adding a new provider, I left a couple of questions regarding topics I was not sure about.

libs/shared/src/consts/providers/credentials/provider-credentials.ts
johnmarksilly: I thought about that too. At first I had it the same, but I switched it since I was already adding a new public key as well. I figured in the future there might be another use case for a private key as well. It's up to your team formally. It would be little effort to switch it out.

providers/push-api/src/lib/push-api.provider.ts
johnmarksilly: Vapid subject is to identify the web push server. Technically it should be another credential. I can add this one in there if there isn't already an email credential.

providers/push-api/src/lib/push-api.provider.ts
johnmarksilly: So with web push, you can have multiple devices which all need to receive a push. 

How I understood the targets, each target represented a push end point for a single subscriber. So in the case of mobile push through APS, they could have multiple iOS devices, but they are logged in with the same credential (subscriber id). 

In this case, I have a desktop browser and an PWA browser running on my phone. Both would be subscribed with the same user.

johnmarksilly: The other thing to note, is web push requires an endpoint and key information. So rather than a single endpoint/target string, the endpoint is really an encoded string with an endpoint and the key signatures.

inveve:
Any movement on this @johnmarksilly

jainpawan21:
@johnmarksilly 
Please look into merge conflcts

johnmarksilly:
@inveve @jainpawan21 @djabarovgeorge See conflict resolutions.


@djabarovgeorge If you have strong opinions about the private/public key credentials let me know and I'll update, but IMO I believe that these will be useful to other devs and are more clear than using other properties. I understand this is hidden from end-users, but devs may find it useful.

johnmarksilly:
@jainpawan21 Is this ready for merge?

docs/docs/channels/push/push-api.md
```diff
@@ -0,0 +1,58 @@
...
+import { Novu, PushProviderIdEnum } from '@novu/node';
+
+const novu = new Novu(process.env.NOVU_API_KEY);
+
+// request access to notifications API and get subscription
+const subscriptionJSON = JSON.stringify(subscription);
+const subscriptionBase64 = btoa(subscriptionJSON) // browser
+// const subscriptionBase64 = Buffer.from(subscriptionJSON, 'ascii').toString('base64') // nodejs
+
+await novu.subscribers.setCredentials('subscriberId', PushProviderIdEnum.FCM, {
```
jainpawan21, line 35: PushProviderIdEnum.FCM => PushProviderIdEnum.PushAPI

jainpawan21 (changes requested):

inveve:
Any update @jainpawan21

jainpawan21:
@djabarovgeorge WDYT what should we do with this PR as there are lot of merge conflicts due to inactivity.

johnmarksilly:
I can go through and resolve the conflicts tomorrow. Every time I get it ready though, it gets forgotten!

providers/push-api/src/lib/push-api.provider.ts
davidsoderberg: @johnmarksilly please go ahead and add this as an email credential.

bart-sofomo:
Wondering if this will ever get merged 🤔

mistycaseyward12:
I would merge it

scopsy:
Hi all, my apologies as this PR got quite lost in between, quite a lot of things has changed in the monorepo since, and unfortuently I will have to merge it. In case it will be reopened and updated I will personaly make sure it will be reviewed in time and merged. I'm sorry for closing this :(

PR #3388, opened 2023-05-11
feat: add light themed icons for digest

davidsoderberg:
Screenshot 2023-05-09 at 15 16 30
Screenshot 2023-05-10 at 15 23 51
Screenshot 2023-05-11 at 06 33 02
Screenshot 2023-05-11 at 06 33 09


apps/web/src/pages/templates/workflow/digest/BackOffTooltipIcon.tsx
```diff
@@ -1,4 +1,89 @@
...
+          x="20.25"
+          y="27.79"
+          width="51.5"
+          height="0.5"
+          rx="0.25"
+          stroke="#BEBECC"
+          stroke-width="0.5"
+          stroke-dasharray="2 2"
+        />
+        <g clip-path="url(#clip1_809_136)">
```
ainouzgali, line 25:
```suggestion
          strokeWidth="0.5"
          strokeDasharray="2 2"
        />
        <g clipPath="url(#clip1_809_136)">
```

And those other places. If I'm not wrong, it causes some errors

ainouzgali (approved):
@davidsoderberg Could we maybe move all the icons to design-system/icons? even if we create a digest folder in icons I think it will be nicer -  so we can easily reuse and to have all icons in the same place

ainouzgali:
Its not possible to reuse the dark one with changing the stroke/fill etc?

davidsoderberg:
> @davidsoderberg Could we maybe move all the icons to design-system/icons? even if we create a digest folder in icons I think it will be nicer - so we can easily reuse and to have all icons in the same place

Not sure these will be reused as other because they are very specific for this location that is why I did not put them i design-system. I agree that we could have an icons folder in digest thu 😄

apps/web/src/pages/templates/workflow/digest/BackOffTooltipIcon.tsx
```diff
@@ -1,4 +1,89 @@
+import { useMantineTheme } from '@mantine/core';
+
 export const BackOffTooltipIcon = () => {
+  const { colorScheme } = useMantineTheme();
+
+  if (colorScheme === 'light') {
+    return (
+      <svg xmlns="http://www.w3.org/2000/svg" width="280" height="57" viewBox="0 0 280 57" fill="none">
```
LetItRock, line 8: please run all the svg's through this tool: https://jakearchibald.github.io/svgomg/, to make them smaller in size

LetItRock: nope, because it's svg's vector graphics :) it just removes unnecessary staff and improved the paths

PR #3393, opened 2023-05-11
feat: add scheduled digest data to analytics event

davidsoderberg:


ainouzgali (approved):

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -71,6 +73,18 @@ export class SendMessage {
       });
 
       const digest = command.job.digest;
+      let timedInfo: any = {};
+
+      if (digest && digest.type === DigestTypeEnum.TIMED && digest.timed) {
```
LetItRock, line 78:
```suggestion
      if (digest?.type === DigestTypeEnum.TIMED) {
```
I feel like `timed` should be required

davidsoderberg: I agree but I like to be better safe then sorry in this case because if it fails here the digest will not be running and mess up the entire workflow for analytics data...

PR #3396, opened 2023-05-11
feat(infra): pause workers in bootstrap until queues are ready

p-fernandez:

Pauses the workers till the queues are ready, checked by the health checks, in order to be able to start to process jobs.

Is part of the improvements for the startup and the shutdown of the apps.

Manually tested. Might require a better implementation.

### DIFF
--- a/apps/worker/src/app/health/health.module.ts
+++ b/apps/worker/src/app/health/health.module.ts
@@ -1,12 +1,6 @@
 import { Module } from '@nestjs/common';
 import { TerminusModule } from '@nestjs/terminus';
-import {
-  DalServiceHealthIndicator,
-  QueueService,
-  QueueServiceHealthIndicator,
-  TriggerQueueService,
-  TriggerQueueServiceHealthIndicator,
-} from '@novu/application-generic';
+import { DalServiceHealthIndicator } from '@novu/application-generic';
 
 import { HealthController } from './health.controller';
 
@@ -15,12 +9,6 @@ import { SharedModule } from '../shared/shared.module';
 @Module({
   imports: [SharedModule, TerminusModule],
   controllers: [HealthController],
-  providers: [
-    DalServiceHealthIndicator,
-    QueueService,
-    QueueServiceHealthIndicator,
-    TriggerQueueService,
-    TriggerQueueServiceHealthIndicator,
-  ],
+  providers: [DalServiceHealthIndicator],
 })
 export class HealthModule {}
--- a/apps/worker/src/app/shared/shared.module.ts
+++ b/apps/worker/src/app/shared/shared.module.ts
@@ -36,6 +36,12 @@ import {
   GCSStorageService,
   AzureBlobStorageService,
   S3StorageService,
+  ReadinessService,
+  QueueServiceHealthIndicator,
+  TriggerQueueServiceHealthIndicator,
+  WsQueueServiceHealthIndicator,
+  QueueService,
+  TriggerQueueService,
 } from '@novu/application-generic';
 
 import * as packageJson from '../../../package.json';
@@ -104,6 +110,22 @@ const distributedLockService = {
   },
 };
 
+const readinessService = {
+  provide: ReadinessService,
+  useFactory: (
+    queueServiceHealthIndicator: QueueServiceHealthIndicator,
+    triggerQueueServiceHealthIndicator: TriggerQueueServiceHealthIndicator,
+    wsQueueServiceHealthIndicator: WsQueueServiceHealthIndicator
+  ) => {
+    return new ReadinessService(
+      queueServiceHealthIndicator,
+      triggerQueueServiceHealthIndicator,
+      wsQueueServiceHealthIndicator
+    );
+  },
+  inject: [QueueServiceHealthIndicator, TriggerQueueServiceHealthIndicator, WsQueueServiceHealthIndicator],
+};
+
 const PROVIDERS = [
   cacheService,
   distributedLockService,
@@ -128,14 +150,17 @@ const PROVIDERS = [
   InvalidateCacheService,
   CreateLog,
   {
-    provide: WsQueueService,
-    useClass: WsQueueService,
-  },
-  {
     provide: StorageService,
     useClass: getStorageServiceClass(),
   },
+  QueueServiceHealthIndicator,
+  TriggerQueueServiceHealthIndicator,
+  WsQueueServiceHealthIndicator,
+  QueueService,
+  TriggerQueueService,
+  WsQueueService,
   StorageHelperService,
+  readinessService,
   ...DAL_MODELS,
 ];
 
--- a/apps/worker/src/app/workflow/services/trigger-worker.service.ts
+++ b/apps/worker/src/app/workflow/services/trigger-worker.service.ts
@@ -22,6 +22,7 @@ export class TriggerWorkerService extends TriggerQueueService implements INovuWo
   private getWorkerOpts() {
     return {
       ...this.bullConfig,
+      autorun: false,
       lockDuration: 90000,
       concurrency: 200,
     };
--- a/apps/worker/src/app/workflow/services/trigger-worker.service.ts
+++ b/apps/worker/src/app/workflow/services/trigger-worker.service.ts
@@ -22,7 +22,6 @@ export class TriggerWorkerService extends TriggerQueueService implements INovuWo
   private getWorkerOpts() {
     return {
       ...this.bullConfig,
-      autorun: false,
       lockDuration: 90000,
       concurrency: 200,
     };
--- a/apps/worker/src/app/workflow/services/trigger-worker.service.ts
+++ b/apps/worker/src/app/workflow/services/trigger-worker.service.ts
@@ -1,16 +1,19 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, Logger } from '@nestjs/common';
 const nr = require('newrelic');
 import {
-  TriggerQueueService,
+  INovuWorker,
   PinoLogger,
   storage,
   Store,
   TriggerEvent,
   TriggerEventCommand,
+  TriggerQueueService,
 } from '@novu/application-generic';
 
+const LOG_CONTEXT = 'TriggerWorkerService';
+
 @Injectable()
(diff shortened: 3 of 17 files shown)


apps/worker/src/app/health/health.module.ts
```diff
@@ -15,12 +9,6 @@ import { SharedModule } from '../shared/shared.module';
 @Module({
   imports: [SharedModule, TerminusModule],
   controllers: [HealthController],
-  providers: [
-    DalServiceHealthIndicator,
-    QueueService,
-    QueueServiceHealthIndicator,
-    TriggerQueueService,
-    TriggerQueueServiceHealthIndicator,
```
p-fernandez, line 23: Moved to `SharedModule` as I needed them there also. NestJS recommends to import it as provider in just one module.

apps/worker/src/app/workflow/services/trigger-worker.service.ts
```diff
@@ -1,16 +1,19 @@
...
   TriggerEvent,
   TriggerEventCommand,
+  TriggerQueueService,
 } from '@novu/application-generic';
 
+const LOG_CONTEXT = 'TriggerWorkerService';
+
 @Injectable()
-export class TriggerProcessorQueueService extends TriggerQueueService {
+export class TriggerWorkerService extends TriggerQueueService implements INovuWorker {
```
p-fernandez, line 16: Renaming as here we are instantiating a worker from BullMQ so for me the naming was confusing. Opinionated and debatable. Can roll back if needed.

apps/worker/src/app/workflow/services/workflow-worker.service.ts
```diff
@@ -31,8 +32,10 @@ interface IJobData {
   _userId: string;
 }
 
+const LOG_CONTEXT = 'WorkflowWorkerService';
+
 @Injectable()
-export class WorkflowQueueService extends QueueService<IJobData> {
+export class WorkflowWorkerService extends QueueService<IJobData> implements INovuWorker {
```
p-fernandez, line 38: Renaming as here we are instantiating a worker from BullMQ so for me the naming was confusing. Opinionated and debatable. Can roll back if needed.

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -110,14 +110,6 @@ const SERVICES: Provider[] = [
...
     useClass: MetricQueueService,
   },
-  {
-    provide: QueueService,
-    useClass: WorkflowQueueService,
-  },
-  {
-    provide: TriggerQueueService,
-    useClass: TriggerProcessorQueueService,
-  },
```
p-fernandez, line 120: I think this didn't make sense. Might be very wrong.
CC: @davidsoderberg ?

apps/worker/src/bootstrap.ts
```diff
@@ -71,6 +95,8 @@ export async function bootstrap(): Promise<INestApplication> {
 
   await app.listen(process.env.PORT);
 
+  await startAppInfra(app);
```
p-fernandez, line 98: Doubting if we should enable the worker before or after letting the application to listen to requests.

packages/application-generic/src/services/readiness.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    Logger.log('Enabling queues as workers are meant to be ready', LOG_CONTEXT);
+
+    try {
+      const healths = await Promise.all([
+        this.queueServiceHealthIndicator.isHealthy(),
+        this.triggerQueueServiceHealthIndicator.isHealthy(),
+        this.wsQueueServiceHealthIndicator.isHealthy(),
+      ]);
+
+      return healths.every((health) => !!health === true);
```
p-fernandez, line 39: In a µservices world I would do HTTP/gRPC calls to the health check endpoints of the different services we are dependant. In this case we share the health indicators in our app generic package. Underneath all them use the same BullMQ instance with its dependency to a dedicated in-memory database. So we check the 3 even though with one would be enough. This would let it prepared if we create a dedicate in-memory instance for every queue in the future.

apps/worker/src/app/shared/shared.module.ts
```diff
@@ -127,15 +149,18 @@ const PROVIDERS = [
...
     useClass: getStorageServiceClass(),
   },
+  QueueServiceHealthIndicator,
+  TriggerQueueServiceHealthIndicator,
+  WsQueueServiceHealthIndicator,
+  QueueService,
+  TriggerQueueService,
+  WsQueueService,
   StorageHelperService,
+  readinessService,
```
p-fernandez, line 163: Order matters. If a provider is dependant of other providers, needs to be added later in the array. Learnt it the hard way. 🙁

apps/worker/src/bootstrap.ts
scopsy: I think that this order makes sense here

packages/application-generic/src/services/readiness.service.ts
```diff
@@ -0,0 +1,68 @@
...
+      );
+
+      return false;
+    }
+  }
+
+  async pauseWorkers(workers: INovuWorker[]): Promise<void> {
+    for (const worker of workers) {
+      Logger.log(`Pausing worker ${worker.name}...`, LOG_CONTEXT);
+      await worker.pauseWorker();
```
scopsy, line 54: I assume that pausing might take time while finishing processing inflights calls, maybe we should do Promise.all here?

scopsy (approved):
Looks good! Would love to get @davidsoderberg opinion regarding the open question of the WorkflowService imports just in case, but otherwise looks good.

apps/worker/src/bootstrap.ts
```diff
@@ -40,6 +62,8 @@ export async function bootstrap(): Promise<INestApplication> {
   app.useLogger(app.get(PinoLogger));
   app.flushLogs();
 
+  await prepareAppInfra(app);
```
p-fernandez, line 65: We pause the workers here.

packages/application-generic/src/services/readiness.service.ts
p-fernandez: Not sure if you are thinking this is for the graceful shutdown. This is used here: https://github.com/novuhq/novu/pull/3396/files#r1193495088

So it is meant to be used before we even accept calls (before app is ready to listen).

apps/worker/src/app/workflow/services/trigger-worker.service.ts
LetItRock: The old name was like that because at some point we had two queues for the trigger: producer and processor. 
I feel like the name should be similar to what we have for the other queues `WorkflowQueueService, WsQueueService`, so like `TriggerQueueService`, but I'm not sure why we have `TriggerQueueService` and not are reusing the `QueueService`

apps/worker/src/app/workflow/workflow.module.ts
LetItRock: @p-fernandez these are injected to the `BULLMQ_LIST` in line 120... also please check all the places where `QueueService` and `TriggerQueueService` were used before and change there accordingly... from what I remember is that `AddJob` use-case is using `private queueService: QueueService` in the constructor because the `WorkflowQueueService` is not the part of the `application-generic` package

apps/worker/src/app/workflow/services/trigger-worker.service.ts
p-fernandez: I can give it a look regarding not using QueueService. But I still think using Queue when we are instantiating a worker is confusing. Maybe here I am thinking more in AWS terms of architecture (SQS, SNS). Also we can raise that as an open question in the team. What do you think?

p-fernandez: You worded nicely @djabarovgeorge. Indeed it is the terminology that I was aiming with these changes as latest years have worked a lot with Pub/Sub tools and patterns and I personally enjoy that separation of concerns. I also see we have that separation in our implementation (it can be seen in the services created and point of discussion) so to me made sense heading that direction with the naming. But again, something worth to bring to the team.

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -108,22 +108,10 @@ const USE_CASES = [
...
-  {
-    provide: WsQueueService,
-    useClass: WsQueueService,
-  },
+  QueueService,
+  TriggerQueueService,
```
LetItRock, line 112: It looks wrong to me (but I'm not sure) that we do provide: `QueueService, TriggerQueueService` and then also `WorkflowWorkerService, TriggerWorkerService`. This way we will have two instances of the same queues because the `WorkflowWorkerService extends QueueService` and `TriggerWorkerService extends TriggerQueueService`. The `WorkflowWorkerService, TriggerWorkerService` instances were injected by their base class because some use cases leave in the `application-generic` package but these classes are defined in the worker app.
I feel it's better to have what was there before.

packages/application-generic/src/services/readiness.service.ts
p-fernandez: Had to revert this change back as I wasn't able to find the reason tests were not working fully. We can try it in the future.

apps/worker/src/app/workflow/services/trigger-worker.service.ts
p-fernandez: I have been pairing with @LetItRock and we agreed to tackle the proposed refactor in a different PR, as the separation of concerns is a good thing and we can have it less confusing as right now.

apps/worker/src/app/workflow/services/workflow-worker.service.ts
p-fernandez: Renaming and refactor will be done in a different PR.

apps/worker/src/app/workflow/workflow.module.ts
p-fernandez: For future readers, I have been pairing with @LetItRock and we agreed to tackle the proposed refactor in a different PR, as the separation of concerns is a good thing and we can have it less confusing as right now. Same for these dependencies and renaming. We will clarify it in that other PR.

PR #3402, opened 2023-05-12
Update Queue metrics accuracy

Cliftonz:
This pr corrects the times and granularity for metrics with bullmq. 

This pr will allow use to accurately autoscale our workers by sending the historical job processing data to a 3rd party New relic.


djabarovgeorge:
Hey @Cliftonz Is this PR ready for review?

Cliftonz:
> Hey @Cliftonz Is this PR ready for review?

No, not yet I need to go over this either tonight or tomorrow to make sure it's ready.

Cliftonz:
> Currently seeing some weird errors, asked Pawel to see if he gets the same ones.

No errors where found on Pawel side

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -89,51 +73,37 @@ export class MetricQueueService extends QueueService<Record<string, never>> {
...
             const activeCount = await queueService.bullMqService.queue.getActiveCount();
 
-            if (process.env.NOVU_MANAGED_SERVICE === 'true') {
-              nr.recordMetric(`MetricQueueService/${queueService.name}/completed`, successMetric);
-              nr.recordMetric(`MetricQueueService/${queueService.name}/failed`, failMetric);
-              nr.recordMetric(`MetricQueueService/${queueService.name}/waiting`, waitCount);
-              nr.recordMetric(`MetricQueueService/${queueService.name}/delayed`, delayedCount);
-              nr.recordMetric(`MetricQueueService/${queueService.name}/active`, activeCount);
+            if (process.env.NOVU_MANAGED_SERVICE === 'true' && process.env.NEW_RELIC_LICENSE_KEY.length !== 0) {
+              Logger.verbose('active length', process.env.NEW_RELIC_LICENSE_KEY.length);
```
scopsy, line 85: This might throw an error when undefined
```suggestion
            if (process.env.NOVU_MANAGED_SERVICE === 'true' && process.env.NEW_RELIC_LICENSE_KEY) {
              Logger.verbose('active length', process.env.NEW_RELIC_LICENSE_KEY.length);
```

Suggest doing it like this

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -1,30 +1,14 @@
...
 
 const LOG_CONTEXT = 'MetricQueueService';
 const METRIC_JOB_ID = 'metric-job';
 
 @Injectable()
-export class MetricQueueService extends QueueService<Record<string, never>> {
+export class MetricQueueActiveService extends QueueService<Record<string, never>> {
   constructor(@Inject('BULLMQ_LIST') private token_list: QueueService[]) {
-    super('metric');
+    super('metric-active');
```
djabarovgeorge, line 11: What about adding the 'job' context as well??
because maybe in the future we will add other `metric-{SOMETHING}-active`
same for the other statuses.

```suggestion
    super(`${METRIC_JOB_ID}-active`);
```

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -79,7 +63,7 @@ export class MetricQueueService extends QueueService<Record<string, never>> {
   private getWorkerOpts(): WorkerOptions {
     return {
       ...this.bullConfig,
-      lockDuration: 500,
+      lockDuration: 900,
```
djabarovgeorge, line 66: i wonder if there were any particular reasons for this change.

apps/worker/src/app/workflow/services/metric-queue-completed.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    };
+  }
+
+  private async jobHasCompleted(job): Promise<void> {
+    Logger.verbose('Metric job Completed', job.id, LOG_CONTEXT);
+  }
+
+  private async jobHasFailed(job, error): Promise<void> {
+    Logger.verbose('Metric job failed', LOG_CONTEXT, error);
+  }
```
djabarovgeorge, line 117: Are those methods in use/future use?

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
Cliftonz: since the job happens every second I boosted the lock time to 900 seconds so that the job had 900ms to finish before the next one kicks off

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
Cliftonz: I dont think we should do this, this job is specificly for getting the active metric. 
I think if we need to add something to another job down the road we can do it then.

apps/worker/src/app/shared/utils/cronHealth.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger } from '@nestjs/common';
+
+const url = 'https://uptime.betterstack.com/api/v1/heartbeat/';
+const LOG_CONTEXT = 'cronHealth';
+export function checkinForCronJob(cronId?: string) {
+  if (process.env.NOVU_MANAGED_SERVICE && process.env.NODE_ENV === 'production' && cronId) {
+    Logger.verbose(`Calling health endpoint for ${cronId}`);
```
Cliftonz, line 7: Added check to betterhealth to make sure the cron jobs are working

apps/worker/src/app/shared/utils/cronHealth.ts
```diff
@@ -0,0 +1,12 @@
...
+    fetch(url + cronId)
+      .then((response) => response.json())
```
djabarovgeorge, line 9: do we need to do something with the response?
looks like we fetch, and then log 'Metric job Completed' Is it the expected behavior?

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -0,0 +1,115 @@
...
+import { QueueService } from '@novu/application-generic';
+import { checkinForCronJob } from '../../shared/utils/cronHealth';
+import * as process from 'process';
+
+const LOG_CONTEXT = 'MetricQueueService';
+const METRIC_JOB_ID = 'metric-job';
+
+@Injectable()
+export class MetricQueueActiveService extends QueueService<Record<string, never>> {
+  constructor(@Inject('BULLMQ_LIST') private token_list: QueueService[]) {
```
djabarovgeorge, line 12: Do we have all of the available Queues in this list, from Worker and API?

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -0,0 +1,115 @@
...
+        Logger.verbose('metric job started', LOG_CONTEXT);
+        const deploymentName = process.env.FLEET_NAME ?? 'default';
+
+        try {
+          for (const queueService of this.token_list) {
+            const waitCount = await queueService.bullMqService.queue.getWaitingCount();
+            const delayedCount = await queueService.bullMqService.queue.getDelayedCount();
+            const activeCount = await queueService.bullMqService.queue.getActiveCount();
+
+            if (process.env.NOVU_MANAGED_SERVICE === 'true' && process.env.NEW_RELIC_LICENSE_KEY) {
```
djabarovgeorge, line 87: Should we validate here the environment here as well or is it redundant?

apps/worker/src/app/shared/utils/cronHealth.ts
djabarovgeorge, line 1: cron-health.ts

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -0,0 +1,115 @@
...
+          if (jobElement.id === METRIC_JOB_ID) {
+            exists = true;
+          }
+        }
+
+        return exists;
+      })
+    )
+      .then(async (exists): Promise<void> => {
+        Logger.debug('metric job exists: ' + exists, LOG_CONTEXT);
```
djabarovgeorge, line 43: maybe we should add it to the if-else statement below because there is a case where we log 'metric job exists' , and then after it log `metricJob doesn't exist, creating it`

apps/worker/src/app/shared/utils/cronHealth.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger } from '@nestjs/common';
+
+const url = 'https://uptime.betterstack.com/api/v1/heartbeat/';
+const LOG_CONTEXT = 'cronHealth';
+export function checkinForCronJob(cronId?: string) {
```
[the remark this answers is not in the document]
Cliftonz: I set it up as optional as the ids are coming in as env variables thus we need to handle in the function if it is undefined then we do not send a good health signal to an "empty" path.

apps/worker/src/app/shared/utils/cronHealth.ts
```diff
@@ -0,0 +1,12 @@
+import { Logger } from '@nestjs/common';
+
+const url = 'https://uptime.betterstack.com/api/v1/heartbeat/';
```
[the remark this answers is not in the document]
Cliftonz: I do not know, what decisions would you make here?

apps/worker/src/app/shared/utils/cronHealth.ts
```diff
@@ -0,0 +1,12 @@
...
+const LOG_CONTEXT = 'cronHealth';
+export function checkinForCronJob(cronId?: string) {
+  if (process.env.NOVU_MANAGED_SERVICE && process.env.NODE_ENV === 'production' && cronId) {
+    Logger.verbose(`Calling health endpoint for ${cronId}`);
+    fetch(url + cronId)
```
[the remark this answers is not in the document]
Cliftonz: agree

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
Cliftonz: @titenkov Could we get your input on this?
I remember there was a specific reason we did this but I did not write it down or remember.

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
Cliftonz: which one are you referring to?
Here I just need to make sure we only send in the cloud version of Novu.

apps/worker/src/app/shared/utils/cronHealth.ts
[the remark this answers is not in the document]
djabarovgeorge: I would put it in env variable that way we can update it more easily if needed, and encapsulate by env in the future if needed.

apps/worker/src/app/shared/utils/cronHealth.ts
djabarovgeorge: If we don't need the response, maybe we can remove the `.json()` in order to remove the redundant parsing.

scopsy:
@Cliftonz what this PR is for? Should we close it?

Cliftonz:
@scopsy This PR is for updating the Queue metrics to be accurate. 
I will make the updates and we can go from there.

apps/worker/src/app/shared/utils/cron-health.ts
```diff
@@ -1,12 +1,12 @@
 import { Logger } from '@nestjs/common';
 
-const url = 'https://uptime.betterstack.com/api/v1/heartbeat/';
+const url = process.env.CRON_CHECKIN_URL;
```
djabarovgeorge, line 3: i know it was my idea to extract it to the env variable, but I think that it will be nice to have a fallback to the origin URL as well just in case. Apologies for not raising it before i will push this small change if you don't mine so you won't have to. :)

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -120,8 +121,12 @@ const REPOSITORIES = [JobRepository];
...
 const SERVICES: Provider[] = [
   {
-    provide: MetricQueueService,
-    useClass: MetricQueueService,
+    provide: MetricQueueCompletedService,
+    useClass: MetricQueueCompletedService,
+  },
+  {
+    provide: MetricQueueActiveService,
+    useClass: MetricQueueActiveService,
```
Cliftonz, line 129: @LetItRock Is the below code on line 144 still the best way to get all of the queues?

```
 provide: 'BULLMQ_LIST',
    useFactory: (workflowQueue: QueueService, triggerQueue: TriggerQueueService, wsQueue: WsQueueService) => {
      return [workflowQueue, triggerQueue, wsQueue];
},
```

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
LetItRock: @djabarovgeorge I think that we are missing the one from the `inbound-mail`, its name is `JobTopicNameEnum.INBOUND_PARSE_MAIL`. 
@Cliftonz However, I'm not sure if we should trace it here.

Cliftonz: @LetItRock we should but I will have to put in a ticket for that as we need to do a few things to get analytics and metrics from the inbound-mail service.

PR #3404, opened 2023-05-13
fix: update subscriber docs

jainpawan21:
Update Subscriber Docs
 - Add ahead of trigger and inline of trigger info
 - Add Few FAQ
 - Add Subscriber Response


docs/docs/platform/subscribers.md
```diff
@@ -31,48 +45,96 @@ await novu.subscribers.identify(user.id, {
...
+
+await novu.subscribers.update(user.id, {
+  // new email
+  email: user.email,
+});
+```
+
+## Delete a subscriber
+
+To stop a subscriber from receiving notifications, you can delete the subscriber. This will hard delete the subscriber means you will not be able to access this subscriber later. You will have to create this subscriber again.
```
p-fernandez, line 137:
```suggestion
To stop a subscriber from receiving notifications, you can delete the subscriber. This will hard delete the subscriber and means you will not be able to access this subscriber later. You will have to create this subscriber again.
```
Also we should explain what happens with the existing data (activity feed) when a hard delete happens.

p-fernandez (approved):
🌟

docs/docs/platform/subscribers.md
jainpawan21: @p-fernandez We have deleted property in subscriber response
https://github.com/novuhq/novu/blob/4b0fa1d2e69dd6e44d883d88b5caefc8306d2cb1/apps/api/src/app/subscribers/dtos/subscriber-response.dto.ts#L61

I want to confirm are we hard deleting the subscriber in this API or not?

docs/docs/platform/subscribers.md
```diff
@@ -82,18 +144,126 @@ const novu = new Novu(process.env.NOVU_API_KEY);
...
-  <p>Subscribers have fixed schema. Storing custom properties is not supported but here is a work around, you can store that property in your database and send those values in payload option of trigger using variables. Read more about variables <a href="./templates#variable-usage"> here </a>.</p>
+  <summary>How to store custom properties in subscriber?</summary>
+  <p>To store custom properties, use <code>data</code> field of subscriber.</p>
+</details>
+
+<details>
+  <summary>Why Novu stores user's PII?</summary>
+  
+  <p>Novu stores user personal information like first name, last name, email, phone, locale, avatar, etc to provide a multichannel notification experience to users. Once these values are stored, Novu automatically configures these values required as per different channels.</p>
+</details>
```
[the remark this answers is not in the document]
jainpawan21: I think we should create a separate page for security (how we handle user data, compliance, our cert like GDPR, SOC 2 etc) and add that page link here

docs/docusaurus.config.js
```diff
@@ -159,7 +159,7 @@ const codeTheme = require('./src/utils/prism');
         },
         prism: {
           theme: codeTheme,
-          additionalLanguages: ['php', 'ruby', 'java'],
+          additionalLanguages: ['php', 'ruby', 'java', 'csharp', 'swift', 'kotlin'],
```
unicodeveloper, line 162: Please add `Go` and `Python`

unicodeveloper: Maybe you should remove `swift`. Right now we don't have any swift sdk

PR #3406, opened 2023-05-14
feat: remove backward compatibility for reset token

michaldziuba03:
It removes backward compatibility with old plain reset token format. It changes single temporary mongodb query to it's final version.

We don't need it anymore as tokens in old format should be already expired in database. I assume hashed reset token feature lives since `v0.14` release [PR#3112](https://github.com/novuhq/novu/pull/3112).


scopsy (approved):

scopsy:
🚀

p-fernandez (approved):
🌟

PR #3407, opened 2023-05-14
fix: bulk insert execution details should be ordered

ainouzgali:

Change insertMany to be ordered. Caused in some cases to execution details in the activity feed to present in the wrong order.


libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -34,7 +34,7 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
       stored.push(this.mapEntity(created));
     }
 
-    await this.MongooseModel.insertMany(stored, { ordered: true });
+    await this.insertMany(stored, true);
```
djabarovgeorge, line 37: I wonder how will we handle if there will be an exception on the bulk error, for example, if we have an array of 100 inserts and the first one causes an error. 
At the moment with an unordered insert, the other ones will be executed in parallel while in an ordered insert the execution will stop.

[Reference](https://www.mongodb.com/docs/manual/reference/method/db.collection.insertMany/#error-handling)

ainouzgali: I believe that is the behaviour we want, no? Why do you feel that would be a problem? 

[A previous discussion ](https://github.com/novuhq/novu/pull/3329#discussion_r1184748745)

djabarovgeorge: Sorry actually, I thought I was commenting executionDetailsRepository.insertMany.
I will continue the comment there :)

packages/application-generic/src/usecases/bulk-create-execution-details/bulk-create-execution-details.usecase.ts
```diff
@@ -19,9 +19,8 @@ export class BulkCreateExecutionDetails {
 
       entities.push(entity);
     });
-    // TODO: Which checks to do? If the notification and job belong to the environment and organization provided?
 
-    await this.executionDetailsRepository.insertMany(entities);
+    await this.executionDetailsRepository.insertMany(entities, true);
```
djabarovgeorge, line 23: [Continued](https://github.com/novuhq/novu/pull/3407/files#r1193123461) 
What I am not sure about here is the change of the behavior previously if we were inserting 100 documents and one failed. we still saved 99 documents. 
Now if we insert 100 documents and the first one failed all of them are not saved.
Did I got it right?

Could be related to storeJobs as well where if the first job is not saved(for example the email) all of the jobs won't be saved.

ainouzgali: Yes, it does mean that. It will save until there is an error in one. I'll think about it/another solution.

p-fernandez: I'd go with @scopsy suggestion. I think that is something we broadly miss in the DAL service. We tend to blindly trust bulk (and even single) inserts.

LetItRock: agree, so we will do the `unordered` and we can check errors like this: https://www.mongodb.com/docs/manual/reference/method/db.collection.insertMany/#unordered-inserts

p-fernandez (approved):
I would also add 2 tests to see the exception. management for both use cases.

PR #3408, opened 2023-05-14
feat: find one integration and only select provider id, remove decrypt

ainouzgali:

No need to find all integrations and do decrypt.


djabarovgeorge (approved):
🚀

PR #3411, opened 2023-05-14
fix: toggle in field value for critical

jainpawan21:
Fixed toggle for critical templates


p-fernandez (approved):
🌟

PR #3413, opened 2023-05-14
fix(novu/node): Get all Notification templates without any parameter

codingmickey:
- Also adding additional unit tests for page and limit on Getting All Notification templates

Just when I was trying to use the `novu/node` package and getting all templates.. I found that I made an error in making the SDK last year that Getting all the notification templates was throwing..

```node

const novu = new Novu('<API_KEY>');

async function main() {
  const result = await novu.notificationTemplates.getAll();
  console.log(result.data);
}

main();
```

- [X] Also added a few unit tests for this function

No issue is open currently for this...

Output of the above code

image


p-fernandez (approved):
You break, you fix. ❤️ 
Thanks for the extra tests added.

codingmickey:
> You break, you fix. ❤️ Thanks for the extra tests added.

Yeah 😂 No problem!!

PR #3417, opened 2023-05-15
chore(application-generic): moved the timed digest delay calculations

LetItRock:

We've decided to move the calculations for the timed digest delay to the main repo. This is what is this PR about.

Tested it locally and it works fine.


davidsoderberg (approved):

packages/application-generic/src/services/calculate-delay/timed-digest-delay.service.ts
```diff
@@ -0,0 +1,186 @@
...
+  dateStart?: Date;
+  unit: DigestUnitEnum;
+  amount: number;
+  timeConfig?: ITimedConfig;
+}
+
+export class TimedDigestDelayService {
+  /**
+   * Calculates the delay time in miliseconds between the time for the next schedule and current time
+   * @returns the delay time in miliseconds
```
p-fernandez, line 75:
```suggestion
   * @returns the delay time in milliseconds
```

p-fernandez (approved):
🌟

PR #3423, opened 2023-05-15
fix: notification template

### LINKED ISSUE #3386: [NV-2292] 🐛 Bug Report: dual active states in UI, if template is created with api
### 📜 Description

When I create a Novu Notification Template using the Novu Node SDK the UI shows that it is in both active and disabled state. I can't trigger notifications using this template until I toggle the active state off and back on in the UI.
err

### 👟 Reproduction steps

1. Use the Novu Node SDK to create a notification group and then a notification template with the same name.
2. Get the triggerID for this notification template and trigger a notification to subscribers (I used 1) with this template.
3. Go to UI and see no notification in the activity and the notification template is in a dual active/disabled state.

### 👍 Expected behavior

It should create the Notification Template in active state and allow me to trigger notifications using this template.

Screenshots attached in description

_No response_

None

<sub>[NV-2292](https://linear.app/novu/issue/NV-2292/🐛-bug-report-dual-active-states-in-ui-if-template-is-created-with-api)</sub>

michaldziuba03:

It "drops" (not actually) `draft` in notification template. What I mean by that is basically remove ability to provide draft value as a user (if user provides `draft` it will be ignored).

After my changes `draft` is always set as opposite of `active` value provided by the user.

**Additional changes**:
- removed `active` from update template dto - it exists **ONLY** in dto but is not even passed into use-case (Novu updates `active` value with seperated endpoint).
- add some missing validators
- add validation for `preferenceSettings`, because it's possible to break template with update

image

image

Closes #3386

- If user create template with `active: true` and `draft: true` - UI will show dual states
- If user create template with `active: false` and `draft: false` - UI will show no state

Screenshot from 2023-05-15 14-24-11


apps/api/src/app/notification-template/notification-template.controller.ts
```diff
@@ -202,7 +202,7 @@ export class NotificationTemplateController {
         steps: body.steps,
         notificationGroupId: body.notificationGroupId,
         active: body.active ?? false,
-        draft: body.draft ?? true,
+        draft: !body.active,
```
michaldziuba03, line 205: `draft` should be always opossite of `active` - otherwise bugs happen

apps/api/src/app/notification-template/usecases/change-template-active-status/change-template-active-status.usecase.ts
```diff
@@ -55,7 +55,7 @@ export class ChangeTemplateActiveStatus {
       {
         $set: {
           active: command.active,
-          draft: command.active === false,
+          draft: !command.active,
```
michaldziuba03, line 58: shorter version suggested by webstorm :)

apps/api/src/app/shared/dtos/preference-channels.ts
```diff
@@ -1,14 +1,29 @@
 import { ApiPropertyOptional } from '@nestjs/swagger';
+import { IsBoolean, IsOptional } from 'class-validator';
```
michaldziuba03, line 2: I decided to reuse this dto for validation, because it's possible to create completely corrupted `preferenceSettings` in database (in update at least).

apps/api/src/app/notification-template/dto/create-notification-template.request.dto.ts
```diff
@@ -1,10 +1,20 @@
...
+  MaxLength,
+  ValidateNested,
+} from 'class-validator';
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { ICreateNotificationTemplateDto, IPreferenceChannels } from '@novu/shared';
 
 import { StepFilter } from '../../shared/dtos/step-filter';
 import { PreferenceChannels } from '../../shared/dtos/preference-channels';
 import { MessageTemplate } from '../../shared/dtos/message-template';
+import { Type } from 'class-transformer';
```
LetItRock, line 17: no need in this additional import

apps/api/src/app/notification-template/dto/update-notification-template-request.dto.ts
```diff
@@ -1,8 +1,18 @@
...
+  IsOptional,
+  IsString,
+  MaxLength,
+  ValidateNested,
+} from 'class-validator';
 import { ICreateNotificationTemplateDto, IPreferenceChannels } from '@novu/shared';
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { PreferenceChannels } from '../../shared/dtos/preference-channels';
 import { NotificationStep } from '../../shared/dtos/notification-step';
+import { Type } from 'class-transformer';
```
LetItRock, line 15: same

LetItRock (commented):
hey, thanks for the contribution! 🌟 
I'm missing here the fix on the FE side, as I understand it shouldn't be possible to see two labels simultaneously: `disabled` and `active`. In the future, if we decide to use both we might hit the same problem again.
Also, we should have a script to update the existing cases in the prod DB.

PR #3424, opened 2023-05-16
feat: add migration to create in app integration

davidsoderberg:

Add migration to create in app integration for existing orgs.

### DIFF
--- a/apps/api/migrations/in-app-integration/in-app-integration.migration.ts
+++ b/apps/api/migrations/in-app-integration/in-app-integration.migration.ts
@@ -23,16 +23,27 @@ export async function createInAppIntegration() {
 
     const environments = await environmentRepository.findOrganizationEnvironments(organization._id);
     for (const environment of environments) {
-      await integrationRepository.create({
+      const count = await integrationRepository.count({
         _environmentId: environment._id,
         _organizationId: organization._id,
         providerId: InAppProviderIdEnum.Novu,
         channel: ChannelTypeEnum.IN_APP,
-        credentials: encryptCredentials({
-          hmac: environment.widget.notificationCenterEncryption,
-        }),
-        active: true,
       });
+
+      if (count === 0) {
+        const response = await integrationRepository.create({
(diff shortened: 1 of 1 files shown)


apps/api/migrations/in-app-integration/in-app-integration.migration.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    const environments = await environmentRepository.findOrganizationEnvironments(organization._id);
+    for (const environment of environments) {
+      await integrationRepository.create({
+        _environmentId: environment._id,
+        _organizationId: organization._id,
+        providerId: InAppProviderIdEnum.Novu,
+        channel: ChannelTypeEnum.IN_APP,
+        credentials: encryptCredentials({
+          hmac: environment.widget.notificationCenterEncryption,
```
scopsy, line 32: The hmac field is not encrypted right?

davidsoderberg: No

scopsy (approved):

apps/api/migrations/in-app-integration/in-app-integration.migration.ts
```diff
@@ -0,0 +1,58 @@
...
+      });
+
+      if (count === 0) {
+        const response = await integrationRepository.create({
+          _environmentId: environment._id,
+          _organizationId: organization._id,
+          providerId: InAppProviderIdEnum.Novu,
+          channel: ChannelTypeEnum.IN_APP,
+          credentials: encryptCredentials({
+            hmac: environment.widget.notificationCenterEncryption,
```
djabarovgeorge, line 40: Should we remove the environment.widget.notificationCenterEncryption under the environment? if yes would it be part of this cycle?

davidsoderberg: I think we should, not sure we should do it this cycle. My suggestion is to do it in a later release...

PR #3430, opened 2023-05-16
fix: secure flag persisted as string

### LINKED ISSUE #3360: [NV-2291] 🐛 Bug Report: APNs integration dialog does not allow turning Production off
### 📜 Description

When configuring the APNs integration, clicking the `Production` toggle sets the slider in the UI correctly, but after saving the value is not persisted.

### 👟 Reproduction steps

1. Go to the integrations store
2. Configure the push provider for APNS
3. Toggle the 'Production' value off
4. Save
5. Re-open the integration settings for APNs
6. The 'Production' value is still set to true.

### 👍 Expected behavior

The 'Production' value should be persisted as 'Off'

https://user-images.githubusercontent.com/123625424/236993527-4a8c226d-d4c3-4984-b825-61da48997e25.mov

Recent related issue: https://github.com/novuhq/novu/issues/2967

None

<sub>[NV-2291](https://linear.app/novu/issue/NV-2291/🐛-bug-report-apns-integration-dialog-does-not-allow-turning)</sub>

### LINKED ISSUE #3416: [NV-2329] 🐛 Bug Report: Issues with integration `secure` option
### 📜 Description

After wrestling with some APNs issues over the past week or so, I was not able to get any meaningful debugging insights from the Novu web app, but I realised that there are some potential issues with how the `secure` property is being handled between Novu <> apns.

Currently, it looks to be an optional property, so integrations created without specifying a value for this are effectively `undefined`.

This can have unfortunate consequences in the case of the `apns` package, since it will default to `production` if `NODE_ENV` is set to production (which I believe [is set](https://github.com/novuhq/novu/blob/next/apps/api/src/.env.production#L1) for Novu api) so if no value exists, production will be assumed by the apns package.

The workaround for this seemed to be ensuring that the property is set to `false`, but this didn't seem to resolve issues. I believe it's an issue with how the `credentials` data is being stored, mainly because the `secure` value is being persisted as a string ("true" || "false"). This has the effect in the `apns` package where `"false"` is actually interpreted as [true for any string value](https://github.com/node-apn/node-apn/blob/master/lib/config.js#L94)

I created a quick test in the apn package to verify that this is the case:

Screenshot 2023-05-15 at 6 53 58 pm

### 👟 Reproduction steps

1. Create an APNs integration via the API excluding the "secure" property
2. Fetch the integraton, observing that the "secure" property is absent
3. Update the integration via an API call, setting the "secure" property to (Boolean) true
4. Observe that the "secure" property is being persisted as a string in mongoDB. 

### 👍 Expected behavior

When creating an integration, the `secure` property should be defaulted to Boolean `false` if no value is provided by teh user.

When the value is updated, the value should be persisted as a boolean value, to prevent the kind of ambiguity introduced when using the apns package, and potentially others

Screenshot 2023-05-15 at 7 01 46 pm

I'm using the API to update notifications, but this may impact the fix for https://github.com/novuhq/novu/issues/3360

I've been unable to get e2e tests working to test a fix for this, but I feel like the `secure` property in the DTO should be `@IsBoolean` over `@IsBooleanAny` which was changed in https://github.com/novuhq/novu/commit/4e5b082b0654d6e78b322917a5d74df001e6dd2b - I don't have context on those changes however, and there was probably a reason for that change that core devs know about.

None

<sub>[NV-2329](https://linear.app/novu/issue/NV-2329/🐛-bug-report-issues-with-integration-secure-option)</sub>

michaldziuba03:
It makes `secure` flag persisted as boolean.

To be more specific it changes Integration mongoose schema, because `credentials.secure` had String type instead of Boolean and for that reason secure was persisted as string.

Second thing - I added transformer `@TransformToBoolean`, because frontend sends "true"/"false" values for credentials and DTO is typed as boolean what makes dto not reliable. Transformer transforms only "true"/"false" it doesn't work like `value === "true"` where any other value than "true" returns false.

New transformer is supposed to be used in combination with `@IsBoolean` validator. I also dropped my previous validator `@IsBooleanAny` because new transformer + regular `@IsBoolean` makes it obsolete and is just better.

New transformer can be used in places like dtos for query params where booleans are expected, because their values are always strings.

Closes #3360


p-fernandez:
@michaldziuba03 at first sight looks like a very good work but I am missing the way the current data is going to be migrated to be able to not break any user using Novu. How would you solve that issue in this PR?

michaldziuba03:
@scopsy I think it depends because if `.lean()` is used somewhere in repository then mongoose returns "raw" data as is.

michaldziuba03:
@p-fernandez @scopsy 

I decided to write migration script for changing type...but I have problem.

Mongoose casts types during updates so when I query something like `{ "credentials.secure": "true" }` mongoose casts it to `{ "credentials.secure": true }` and it  updates from true to true :)

I know `@novu/dal` tries to abstract data access layer but I think I need direct access to connection and perform query using underlying mongodb driver.

jainpawan21:
@p-fernandez / @scopsy Can we please prioritize this PR?

apps/api/migrations/secure-to-boolean/secure-to-boolean-migration.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  console.log('Updating from "true" to true...');
+  const resultTrue = await updateTrueValues();
+  console.log(`Matched: ${resultTrue.matchedCount}\nModified: ${resultTrue.modifiedCount} \n`);
+
+  console.log('Updating from "false" to false...');
+  const resultFalse = await updateFalseValues();
+  console.log(`Matched: ${resultFalse.matchedCount}\nModified: ${resultFalse.modifiedCount}`);
+
+  console.log('End migration.');
```
p-fernandez, line 24: Use `Logger` from `@nestjs/common` rather than `console.log` so we can track in out telemetry tools.

p-fernandez:
@scopsy it looks ok from my side

apps/api/migrations/secure-to-boolean/secure-to-boolean-migration.ts
michaldziuba03: sure. In other migration scripts I have seen console.logs so I decided to use them in my script as well.

I switched to Nest.js Logger in recent changes.

michaldziuba03:
@scopsy @p-fernandez

scopsy (approved):

PR #3432, opened 2023-05-16
NV-2332 - Template Store - Add blueprint module

djabarovgeorge:

Add a dedicated blueprint module to fetch blueprints.

Separate the blueprint logic from the notification-templates.


apps/api/src/app/blueprint/blueprint.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+  ): Promise<NotificationTemplateResponse> {
+    return this.createBlueprintNotificationTemplate.execute(
+      GetBlueprintNotificationTemplateCommand.create({
+        environmentId: user.environmentId,
+        organizationId: user.organizationId,
+        userId: user._id,
+        templateId,
+      })
+    );
+  }
```
djabarovgeorge, line 49: Will probably be removed. need to update the client to use /notification-template post request.

djabarovgeorge: Created a [task](https://linear.app/novu/issue/NV-2335/[template-store]-refactor-web-on-blueprint-creation-to-use) for it.

apps/api/src/app/blueprint/e2e/create-blueprint-notification-templates.e2e.ts
```diff
@@ -1,11 +1,12 @@
...
+
 import { UserSession } from '@novu/testing';
 import { EmailBlockTypeEnum, StepTypeEnum, INotificationTemplate, FilterPartTypeEnum } from '@novu/shared';
 import { NotificationTemplateRepository, EnvironmentRepository } from '@novu/dal';
 
-import { CreateNotificationTemplateRequestDto } from '../dto';
+import { CreateNotificationTemplateRequestDto } from '../../notification-template/dto';
 
-describe('Create Notification template from blueprint - /notification-templates/:templateId/blueprint (POST)', async () => {
+describe('Create Notification template from blueprint - /blueprints/:templateId (POST)', async () => {
```
p-fernandez, line 9: I assume this is not a breaking change as is for Novu org.

apps/api/src/app/notification-template/notification-template.controller.ts
```diff
@@ -149,36 +144,6 @@ export class NotificationTemplateController {
...
-    return this.getBlueprintNotificationTemplate.execute(
-      GetBlueprintNotificationTemplateCommand.create({
-        environmentId: user.environmentId,
-        organizationId: user.organizationId,
-        userId: user._id,
-        templateId,
-      })
-    );
-  }
-
```
p-fernandez, line 166: What we are trying to achieve moving them to an own module? As far as I remember when @davidsoderberg implemented this, blueprints will be base notification templates, so it felt right to live in the notifications template module and its API.

LetItRock: I tend to agree, the blueprints are just the copy of the original notification template from another org. So I feel like there is no sense to move it to a separate module and make the breaking changes in the public `API` (not sure if anyone is using it).

apps/api/src/app/blueprint/e2e/create-blueprint-notification-templates.e2e.ts
djabarovgeorge: It will break changes for the pixel point project, i check already with Nevo he said it will be fine, they could update it on their side once we deploy.

apps/api/src/app/notification-template/notification-template.controller.ts
djabarovgeorge: So the change was made for several reasons, you guys are right to say that the blueprint is in a sense are notification- template, but logically they are not. [I was not sure what to name this new module](https://discord.com/channels/895029566685462578/1105123620650487919/1107993657384116355
) as well 
I wanted to separate those two because maybe tomorrow we will get this data from separate CMS and then it won't make sense to keep it in the notification-template namespace.
The most crucial part of why I created a separate module was the [blueprint storage](https://www.notion.so/novuhq/Technical-spec-bd4583dff7f04a63b51001fb3dd684ba?pvs=4#4d463327ce564c419c4e931c6be02bf0)
but in the end, it was not relevant but I still decided to keep it separate.

Let me know what you think at the current state I am not tied in this decision and I can revert it if you think it is less intuitive.
@p-fernandez @LetItRock

LetItRock: @djabarovgeorge 
I'm ok with leaving it, but all the reasons are not valid anymore as you said. 
So my approach is to not overengineer if not needed and if you are not sure because the future changes might never happen. But anyway, this is just my personal opinion, and I feel it's ok to leave it if you already did a big chunk of work.

LetItRock (approved):

apps/api/src/app/notification-template/notification-template.controller.ts
djabarovgeorge: The revert should not be too much work, I will merge for now and revisit this decision in middle of this week

PR #3433, opened 2023-05-16
fix: corrected apple dev link in apns docs

michaldziuba03:
changes link from [developer.apple.com.](https://developer.apple.com.) to [developer.apple.com](https://developer.apple.com) in [APNs docs](https://docs.novu.co/channels/push/apns/).

because we don't want invalid links in docs :)

screenshot


p-fernandez (approved):
🌟

PR #3438, opened 2023-05-17
Nv 2333 update api to use new hmac flag on

davidsoderberg:


apps/api/src/app/channels/dtos/update-settings-response.dto.ts
```diff
@@ -22,6 +22,7 @@ export class UpdateSettingsResponseDto {
 
   @ApiProperty({
     type: WidgetSettings,
+    deprecated: true,
```
davidsoderberg, line 25: Should we deprecate it or just remove it from dtos?

apps/api/src/app/environments/environments.controller.ts
```diff
@@ -152,23 +148,4 @@ export class EnvironmentsController {
...
-    @Body() body: UpdateWidgetSettingsRequestDto
-  ): Promise<EnvironmentResponseDto> {
-    const command = UpdateWidgetSettingsCommand.create({
-      organizationId: user.organizationId,
-      environmentId: user.environmentId,
-      notificationCenterEncryption: body.notificationCenterEncryption,
-    });
-
-    return await this.updateWidgetSettingsUsecase.execute(command);
-  }
```
p-fernandez, line 173: 👋🏻

apps/api/src/app/widgets/usecases/initialize-session/initialize-session.usecase.ts
```diff
@@ -26,7 +34,23 @@ export class InitializeSession {
...
+      GetDecryptedIntegrationsCommand.create({
+        findOne: true,
+        active: true,
+        environmentId: environment._id,
+        organizationId: environment._organizationId,
+        userId: command.subscriberId,
+      })
+    );
+
+    const item = Array.isArray(response) ? response[0] : response;
```
p-fernandez, line 47: Not sure if I am missing any detail but this bit of code can lead to bugs. If we now make the In App as a provider, that means it could be not enabled. And at the very same time a different channel provider enabled. So we should check that from the decrypted active integrations we have the In App one enabled.

packages/node/src/lib/environments/environments.ts
```diff
@@ -31,8 +30,4 @@ export class Environments extends WithHttp implements IEnvironments {
   async regenerateApiKeys() {
     return await this.http.post('/environments/api-keys/regenerate');
   }
-
-  async updateWidget(payload: IWidgetUpdatePayload) {
-    return await this.http.put('/environments/widget/settings', payload);
-  }
```
p-fernandez, line 37: 👋🏻

p-fernandez (changes requested):
Good cleaning but I think we might have a bug in the implementation.

apps/api/src/app/widgets/usecases/initialize-session/initialize-session.usecase.ts
davidsoderberg: Yeah you found it at the same time as I did ;)

p-fernandez (approved):
🌟

PR #3441, opened 2023-05-17
feat(api): implement launch darkly node sdk

p-fernandez:

Implements a Feature Flags service that instantiates a selected provider service, in the current case, the Launch Darkly service. Is implemented in a way that won't block execution if the feature flags provider is not set.

Abstraction of the feature flags provider.


apps/api/src/app/feature-flags/feature-flags.module.ts
```diff
@@ -2,12 +2,23 @@ import { Module } from '@nestjs/common';
...
 import { AuthModule } from '../auth/auth.module';
 
+const featureFlagsService = {
+  provide: FeatureFlagsService,
+  useFactory: async () => {
+    return new FeatureFlagsService();
+  },
+};
+
+const providers = [LaunchDarklyService, featureFlagsService];
```
LetItRock, line 17: I'm wondering whether these two should be added to the `application-generic` package instead

scopsy: Good point, in that way we can also use it inside the `worker`, `ws` and other services

p-fernandez: I have a fear of `application-generic` becoming a trash can where everything shared ends up there.
I am biased regarding the usage of feature flags in the worker, as it is the lowest level of our system, and I consider feature flags discrimination should be as early as possible in the API layer. On the other side I would understand the need to use feature flags for the WS app. Though one could say that should be controlled in the client (Web app) and to retrieve the logic from the client we have 2 options: to call to our API and have a one place of responsibility or use the Launch Darkly client SDK package.

LetItRock: This is how the `application-generic` is used right now 🥲
I feel like it might be useful for all the apps, I can imagine a situation where we want the request to go through the API layer and just feature flag something in the Worker layer, let's say new calculations for the timed digest functionality, for some users we will calculate it with "old logic" and for others using the "new logic" (just the example).

LetItRock: maybe let's start like this, and then if we will see the need then we can move to the `application-generic`

LetItRock (approved):

apps/api/src/app/feature-flags/services/feature-flags.service.ts
```diff
@@ -0,0 +1,51 @@
...
+  [FeatureFlagsProvidersEnum.LAUNCH_DARKLY]: LaunchDarklyService,
+};
+
+@Injectable()
+export class FeatureFlagsService {
+  private service: IFeatureFlagsService;
+
+  constructor() {
+    // TODO: In the future we can replace the object key here for an environment variable
+    const service = featureFlagsProviders[FeatureFlagsProvidersEnum.LAUNCH_DARKLY];
```
djabarovgeorge, line 23: If we want to let the user have the ability to choose a provider, maybe it will be a good idea to introduce an environment variable with the provider name e.g
FEATURE_FLAG_PROVIDER=LaunchDarkly.
And update the current environment variable key LAUNCH_DARKLY_SDK_KEY to a generic name like FEATURE_FLAG_SDK_KEY, that way the user could choose from existing sdk's a provide that he wants to use.

p-fernandez: Exactly is what I'd like to do in the future. 👍🏻

Issue #2964, opened 2023-03-07
[NV-1815] 🐛 Bug Report: GitHub email change = new account


michaldziuba03:
### 📜 Description

I know it's quite rare case but...when you register in Novu using "Sign in with GitHub" and later you change GitHub primary email, Novu creates new account and previous data is "gone".

Btw. I can reset password while using GitHub provider which is passwordless - it's expected behaviour?

### 👟 Reproduction steps

1. Sign In with GitHub on Novu
2. Change primary email on GitHub
3. Log out from Novu
4. Login with the same GitHub account
5. You end up with new Novu account :)

### 👍 Expected behavior

Most applications perform account lookup by id from provider instead just email address. So email change shouldn't affect user experience directly. The main drawback could be stale email address saved in database during registration.

### 👎 Actual Behavior with Screenshots

**X** - email address before change
**Y** - email address after change

### View from account with X mail address

![image](https://user-images.githubusercontent.com/43048524/223445981-a9bc9c03-6c2d-4af7-80c7-71c2e6d6bafd.png)

### View from account with Y email address

![image](https://user-images.githubusercontent.com/43048524/223446680-0ee482df-93ae-4fcc-9fc6-b154627f246a.png)

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-1815](https://linear.app/novu/issue/NV-1815/🐛-bug-report-github-email-change-=-new-account)</sub>

scopsy:
I think that one of the potential things we should do is instead of using the email to identify on oauth, we can use the profiled returned by GitHub instead.

oba2311:
@scopsy  @BiswaViraj  if you guys are busy this could get into other cycle IMO.

scopsy:
@oba2311 added a help wanted for this issue.

scopsy:
@peoray sure! Assigning you 🙏

oba2311:
@unicodeveloper  could you help here? thanks

scopsy:
[...]

PR #3443, opened 2023-05-17
fix: prevent github email change from creating a new account

gitstart:
 This changes ensure users retain their account when their github's primary email is updated.

 New account is created for users that uses github authentication provider, when they update their github account's primary email address.

### Other information
closes https://github.com/novuhq/novu/issues/2964

---


gitstart:

p-fernandez (changes requested):
This is a quite sensitive change. I would expect to include tests for confidence.

apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -89,20 +93,45 @@ export class AuthService {
...
+              {
+                $set: {
+                  ...(withoutUsername && { 'tokens.$.username': profile.login }),
+                  ...(email !== user.email && { email }),
+                },
+              }
+            )
           );
+        }
+        if (withoutToken) {
```
djabarovgeorge, line 116: Here we will update the credentials only if the user does not have a GitHub token?
Could there be a case when we would want to update the credentials we the user have GitHub token but it is not updated?

apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -89,20 +93,45 @@ export class AuthService {
...
+        // Checks if user's github primary email already exists but from a different authentication provider.
+        const withoutToken = user.tokens.findIndex((i) => i.provider === AuthProviderEnum.GITHUB) === -1;
+        const userUpdates: Promise<any>[] = [];
+
+        if (withoutUsername || email !== user.email) {
+          userUpdates.push(
+            this.userRepository.update(
+              {
+                _id: user._id,
+                'tokens.providerId': profile.id,
```
djabarovgeorge, line 105: Following the next if statement, could there be a case when the user is withoutUsername AND withoutToken? if it could be I think this query will return an empty result.

apps/api/src/app/auth/e2e/github-auth.e2e.ts
djabarovgeorge, line 1: 🔥

djabarovgeorge (commented):
Left a couple of questions, due mine lack of knowledge in the GitHub OAuth. Tend to agree with Pablo we should cover this change with tests as much as possible in order to keep this code right in the future.

libs/dal/src/repositories/user/user.repository.ts
```diff
@@ -47,6 +47,15 @@ export class UserRepository extends BaseRepository<UserDBModel, UserEntity> {
     });
   }
 
+  async findByEmailOrLoginProvider(
+    email: string,
+    { profileId, provider }: { profileId: string; provider: AuthProviderEnum }
+  ): Promise<UserEntity | null> {
+    return this.findOne({
+      $or: [{ email }, { 'tokens.providerId': profileId, 'tokens.provider': provider }],
+    });
```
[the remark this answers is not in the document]
gitstart: @djabarovgeorge No chance for that.

The email that gets passed in here is from the provider profile data of the user.

apps/api/src/app/auth/services/auth.service.ts
gitstart: @djabarovgeorge This was an existing logic, although I get the additional concern of updating an expired token (if it expires in the first place). Maybe @p-fernandez could help out here.

gitstart: Not entirely sure about that. @p-fernandez WDYT?

apps/api/src/app/auth/services/auth.service.ts
gitstart: Given we already have this condition
https://github.com/novuhq/novu/pull/3443/files#diff-7b90561f1765cc8061f3065abde564abfdf3891dac81d2e99577cd44285b8061R92

It will update the user info if any of the below holds true
https://github.com/novuhq/novu/pull/3443/files#diff-7b90561f1765cc8061f3065abde564abfdf3891dac81d2e99577cd44285b8061R100

gitstart: @djabarovgeorge 

Given we already have this condition
https://github.com/novuhq/novu/blob/5e1a23fdaa143d5bc23ace875d34f35afd8ce3b2/apps/api/src/app/auth/services/auth.service.ts#L92

It will update the user info if any of the below condition is the case
https://github.com/novuhq/novu/blob/5e1a23fdaa143d5bc23ace875d34f35afd8ce3b2/apps/api/src/app/auth/services/auth.service.ts#L100

gitstart:
@djabarovgeorge sorry for the delay in replying your comments. Could you please go through the responses.

cc: @p-fernandez

PR #3444, opened 2023-05-17
NV-2336 - Template Store -Add grouped by category blueprints

djabarovgeorge:

Adds get grouped by category blueprint list.
refactored some old code.
switched to fetch by prod env and not the organization.

So the client could get all the available blueprints.

### DIFF
--- a/apps/api/src/app/blueprint/blueprint.controller.ts
+++ b/apps/api/src/app/blueprint/blueprint.controller.ts
@@ -8,6 +8,7 @@ import { GroupedBlueprintResponse } from './dto/grouped-blueprint.response.dto';
 import { GetBlueprint, GetBlueprintCommand } from './usecases/get-blueprint';
 import { CreateBlueprintCommand, CreateBlueprint } from './usecases/create-blueprint';
 import { GetGroupedBlueprints } from './usecases/get-blueprints';
+import { GetBlueprintResponse } from './dto/get-blueprint.response.dto';
 
 @Controller('/blueprints')
 @UseInterceptors(ClassSerializerInterceptor)
@@ -19,7 +20,7 @@ export class BlueprintController {
   ) {}
 
   @Get('/:templateId')
-  getBlueprintById(@Param('templateId') templateId: string): Promise<NotificationTemplateEntity> {
+  getBlueprintById(@Param('templateId') templateId: string): Promise<GetBlueprintResponse> {
     return this.getBlueprintUsecase.execute(
       GetBlueprintCommand.create({
         templateId,
--- a/apps/api/src/app/blueprint/blueprint.controller.ts
+++ b/apps/api/src/app/blueprint/blueprint.controller.ts
@@ -1,45 +1,44 @@
-import { ClassSerializerInterceptor, Controller, Get, Param, Post, UseGuards, UseInterceptors } from '@nestjs/common';
+import { ClassSerializerInterceptor, Controller, Get, Param, Post, UseInterceptors } from '@nestjs/common';
 
 import { IJwtPayload } from '@novu/shared';
+import { NotificationTemplateEntity } from '@novu/dal';
 
 import { UserSession } from '../shared/framework/user.decorator';
-import { NotificationTemplateResponse } from '../notification-template/dto/notification-template-response.dto';
-import {
-  GetBlueprintNotificationTemplate,
-  GetBlueprintNotificationTemplateCommand,
-} from './usecases/get-blueprint-notification-template';
-import { CreateBlueprintNotificationTemplate } from './usecases/create-blueprint-notification-template';
+import { GroupedBlueprintResponse } from './dto/grouped-blueprint.response.dto';
+import { GetBlueprint, GetBlueprintCommand } from './usecases/get-blueprint';
+import { CreateBlueprintCommand, CreateBlueprint } from './usecases/create-blueprint';
+import { GetGroupedBlueprints } from './usecases/get-blueprints';
 
 @Controller('/blueprints')
 @UseInterceptors(ClassSerializerInterceptor)
 export class BlueprintController {
   constructor(
-    private getBlueprintNotificationTemplate: GetBlueprintNotificationTemplate,
-    private createBlueprintNotificationTemplate: CreateBlueprintNotificationTemplate
+    private getBlueprintUsecase: GetBlueprint,
+    private getGroupedBlueprintsUsecase: GetGroupedBlueprints,
+    private createBlueprintUsecase: CreateBlueprint
   ) {}
 
   @Get('/:templateId')
-  getNotificationTemplateBlueprintById(
-    @UserSession() user: IJwtPayload,
-    @Param('templateId') templateId: string
-  ): Promise<NotificationTemplateResponse> {
-    return this.getBlueprintNotificationTemplate.execute(
-      GetBlueprintNotificationTemplateCommand.create({
-        environmentId: user.environmentId,
-        organizationId: user.organizationId,
-        userId: user._id,
+  getBlueprintById(@Param('templateId') templateId: string): Promise<NotificationTemplateEntity> {
+    return this.getBlueprintUsecase.execute(
+      GetBlueprintCommand.create({
         templateId,
       })
     );
+  }
+
+  @Get('/group-by-category')
+  getGroupedBlueprints(@UserSession() user: IJwtPayload): Promise<GroupedBlueprintResponse[]> {
+    return this.getGroupedBlueprintsUsecase.execute();
   }
 
   @Post('/:templateId/blueprint')
   createNotificationTemplateFromBlueprintById(
     @UserSession() user: IJwtPayload,
     @Param('templateId') templateId: string
-  ): Promise<NotificationTemplateResponse> {
-    return this.createBlueprintNotificationTemplate.execute(
-      GetBlueprintNotificationTemplateCommand.create({
+  ): Promise<NotificationTemplateEntity> {
+    return this.createBlueprintUsecase.execute(
+      CreateBlueprintCommand.create({
         environmentId: user.environmentId,
         organizationId: user.organizationId,
         userId: user._id,
--- a/apps/api/src/app/blueprint/blueprint.controller.ts
+++ b/apps/api/src/app/blueprint/blueprint.controller.ts
@@ -19,6 +19,11 @@ export class BlueprintController {
     private createBlueprintUsecase: CreateBlueprint
   ) {}
 
+  @Get('/group-by-category')
+  getGroupedBlueprints(): Promise<GroupedBlueprintResponse[]> {
+    return this.getGroupedBlueprintsUsecase.execute();
+  }
+
   @Get('/:templateId')
   getBlueprintById(@Param('templateId') templateId: string): Promise<GetBlueprintResponse> {
     return this.getBlueprintUsecase.execute(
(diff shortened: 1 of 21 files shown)


apps/api/src/app/blueprint/usecases/get-blueprints/get-grouped-blueprints.usecase.ts
```diff
@@ -0,0 +1,20 @@
...
+
+@Injectable()
+export class GetGroupedBlueprints {
+  constructor(private notificationTemplateRepository: NotificationTemplateRepository) {}
+
+  async execute(): Promise<GroupedBlueprintResponse[]> {
+    const blueprints = await this.notificationTemplateRepository.findAllGroupedByCategory();
+    if (!blueprints) {
+      throw new NotFoundException(
+        `Blueprints with id ${NotificationTemplateRepository.getBlueprintEnvironmentId()} was not found`
```
ainouzgali, line 14:
```suggestion
        `Blueprints for environment id ${NotificationTemplateRepository.getBlueprintEnvironmentId()} were not found`
```

apps/api/src/app/blueprint/blueprint.controller.ts
```diff
@@ -1,45 +1,44 @@
...
-        userId: user._id,
+  getBlueprintById(@Param('templateId') templateId: string): Promise<NotificationTemplateEntity> {
+    return this.getBlueprintUsecase.execute(
+      GetBlueprintCommand.create({
         templateId,
       })
     );
   }
 
+  @Get('/group-by-category')
```
djabarovgeorge, line 30: This one is no good 😞 because it is the same as @Get('/:templateId') , need to think about other way to represent it in the API.

apps/api/src/app/blueprint/blueprint.controller.ts
```diff
@@ -1,45 +1,44 @@
...
-  getNotificationTemplateBlueprintById(
-    @UserSession() user: IJwtPayload,
-    @Param('templateId') templateId: string
-  ): Promise<NotificationTemplateResponse> {
-    return this.getBlueprintNotificationTemplate.execute(
-      GetBlueprintNotificationTemplateCommand.create({
-        environmentId: user.environmentId,
-        organizationId: user.organizationId,
-        userId: user._id,
+  getBlueprintById(@Param('templateId') templateId: string): Promise<NotificationTemplateEntity> {
```
p-fernandez, line 22: Can we return a DTO like `NotificationStepDto` or similar instead of returning the DAL entity? It doesn't feel right we have to have a dependency to DAL in this layer.
Same for the method `createNotificationTemplateFromBlueprintById`.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -46,22 +46,56 @@ export class NotificationTemplateRepository extends BaseRepository<
...
     };
 
     const item = await this.MongooseModel.findOne(requestQuery).populate('steps.template');
 
     return this.mapEntity(item);
   }
 
+  async findAllGroupedByCategory(): Promise<{ name: string; blueprints: NotificationTemplateEntity[] }[]> {
+    if (!NotificationTemplateRepository.getBlueprintEnvironmentId()) {
+      return {} as { name: string; blueprints: NotificationTemplateEntity[] }[];
```
p-fernandez, line 59:
```suggestion
      return [];
```
This doesn't feel right. To trick TS to return an empty object instead of an empty array as it should be or an undefined value.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -46,22 +46,56 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+      _environmentId: NotificationTemplateRepository.getBlueprintEnvironmentId() as string,
     };
 
     const item = await this.MongooseModel.findOne(requestQuery).populate('steps.template');
 
     return this.mapEntity(item);
   }
 
+  async findAllGroupedByCategory(): Promise<{ name: string; blueprints: NotificationTemplateEntity[] }[]> {
+    if (!NotificationTemplateRepository.getBlueprintEnvironmentId()) {
```
p-fernandez, line 58:
```suggestion
   const blueprintId = NotificationTemplateRepository.getBlueprintEnvironmentId();
    if (!blueprintId) {
```
We need the `blueprintId` further in this method so we can reuse the value.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -46,22 +46,56 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+
+    const requestQuery: NotificationTemplateQuery = {
+      isBlueprint: true,
+      _environmentId: NotificationTemplateRepository.getBlueprintEnvironmentId() as string,
+    };
+
+    const items = await this.MongooseModel.find(requestQuery)
+      .populate('steps.template')
+      .populate('notificationGroup')
+      .lean();
```
p-fernandez, line 70: If here we won't have duplicates I think we could simplify and just use a map function rather than a reduce when calculated the groupedItems.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -46,22 +46,56 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+
+      return acc;
+    }, {});
+
+    return Object.values(groupedItems);
+  }
+
   async getBlueprintList(skip = 0, limit = 10) {
-    if (!NotificationTemplateRepository.getBlueprintOrganizationId()) {
+    if (!NotificationTemplateRepository.getBlueprintEnvironmentId()) {
```
p-fernandez, line 92: Same as comment for line 58.

apps/api/src/app/change/usecases/promote-notification-template-change/promote-notification-template-change.usecase.ts
```diff
@@ -129,7 +129,7 @@ export class PromoteNotificationTemplateChange {
         _environmentId: command.environmentId,
         _organizationId: command.organizationId,
         _notificationGroupId: notificationGroup._id,
-        isBlueprint: command.organizationId === this.blueprintOrganizationId,
+        isBlueprint: !!newItem.blueprintId,
```
djabarovgeorge, line 132: @davidsoderberg I wonder if this one is equivalent to the previous flag.

apps/api/src/app/blueprint/blueprint.controller.ts
LetItRock: maybe it should be a `GET /blueprints` and the `groupBy` should be a query param that allows to group by something?

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -115,7 +149,7 @@ export class NotificationTemplateRepository extends BaseRepository<
     return this.mapEntity(res);
   }
 
-  public static getBlueprintOrganizationId(): string | undefined {
+  public static getBlueprintEnvironmentId(): string | undefined {
```
LetItRock, line 152: how would I know the `environment id` if I've just forked and deployed it for the first time?

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -46,22 +46,56 @@ export class NotificationTemplateRepository extends BaseRepository<
...
   }
 
+  async findAllGroupedByCategory(): Promise<{ name: string; blueprints: NotificationTemplateEntity[] }[]> {
+    if (!NotificationTemplateRepository.getBlueprintEnvironmentId()) {
+      return {} as { name: string; blueprints: NotificationTemplateEntity[] }[];
+    }
+
+    const requestQuery: NotificationTemplateQuery = {
+      isBlueprint: true,
+      _environmentId: NotificationTemplateRepository.getBlueprintEnvironmentId() as string,
```
LetItRock, line 64: please move `NotificationTemplateRepository.getBlueprintEnvironmentId()` to const, then you won't need to do `as string`

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -46,22 +46,56 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+    const items = await this.MongooseModel.find(requestQuery)
+      .populate('steps.template')
+      .populate('notificationGroup')
+      .lean();
+
+    const groupedItems = items.reduce((acc, item) => {
+      const notificationGroupId = item._notificationGroupId;
+      const notificationGroupName = item.notificationGroup?.name;
+
+      if (!acc[notificationGroupId as unknown as string]) {
```
LetItRock, line 76: the type of `notificationGroupId` is `ObjectId`, so you don't have to cast to `as unknown as string`, but you can do `notificationGroupId.toString()`

libs/dal/src/repositories/notification-template/notification-template.repository.ts
LetItRock: not sure if we can do that, example plz?

apps/api/src/app/change/usecases/promote-notification-template-change/promote-notification-template-change.usecase.ts
djabarovgeorge: Never mind i will revet this change.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: I am reverting this change to the organization, but the question will be still valid, and it is good I will add it to the open questions.

apps/api/src/app/blueprint/blueprint.controller.ts
djabarovgeorge: Yeah definitely I will create one, tho createNotificationTemplateFromBlueprintById will be deleted in this cycle so I won't add it there.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: > If here we won't have duplicates I think we could simplify and just use a map function rather than a reduce when calculated the groupedItems.

Not sure if I got the 'map' right in your suggestion.

djabarovgeorge: > not sure if we can do that, example plz?

Was this comment directed to me or Pablo? :)

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: Dima's [answer](https://www.notion.so/Technical-spec-bd4583dff7f04a63b51001fb3dd684ba?d=a1ae851c7d134e99b35a957fe1be4fa0#28cf65d0f0764628949cb364a6f51f15) ,the newly forked project won't point to the local API it will point to the cloud version.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
LetItRock: to @p-fernandez

p-fernandez: I just noticed that there could be multiple blueprints for different notification groups. I missed that point, sorry. 😓

LetItRock (approved):

apps/api/src/app/blueprint/blueprint.controller.ts
djabarovgeorge: Decided to keep it this way because we won't have other group options at the moment, and creating the query will make the caching more complicated.

PR #3449, opened 2023-05-17
feat(api): Add template object to subscriber notification feed response

### LINKED ISSUE #3381: 🐛 Bug Report: template object missing in GET feed response
### 📜 Description

the API docs show an optional `template` object in response to the `GET feed` endpoint, but there is no way to specify it in the request and it is missing from the resopnse

### 👟 Reproduction steps

call the `GET feed` endpoint here: https://docs.novu.co/api/get-a-notification-feed-for-a-particular-subscriber/

### 👍 Expected behavior

expected to have the `template` object or at least be able to specify that we want it in the request

it is missing from the response completely :( 

_No response_

None

peoray:
Fixes  #3381


p-fernandez (approved):
🌟 
Would be interesting to add an E2E test that checks this addition.

jainpawan21:
@peoray please add test in this file
https://github.com/novuhq/novu/blob/next/apps/api/src/app/subscribers/e2e/get-notifications-feed.e2e.ts

scopsy:
For some of the performance considerations here I would prefer if we return only the needed fields and not the whole template. @peoray maybe we can ask the issue opened about those?

peoray:
@scopsy I have a question about that already. It is unanswered

scopsy:
@peoray let's start with name, identifier and _id for now. Wdyt?

scopsy:
Closing as stale 🙏

peoray:
@scopsy how can this be reopened?

PR #3450, opened 2023-05-17
feat(web): templates store modal workflow preview

LetItRock:

In this PR I've implemented the workflow preview in the Templates Store Modal.
The preview is using the same `FlowEditor` that we use in the Template Editor, so it's been a little refactored by moving the code that shouldn't be there outside of the component. I'll leave the explanation in the comments. 

Templates Store PRD.

Screenshot 2023-05-17 at 23 05 58
Screenshot 2023-05-17 at 23 05 49
Screenshot 2023-05-17 at 23 05 34
Screenshot 2023-05-17 at 23 05 12

https://github.com/novuhq/novu/assets/2607232/f56d3fa5-399b-421a-8ca5-9849c61d5760


apps/web/src/components/quick-start/digest-demo-flow/NodeStepWithPopover.tsx
```diff
@@ -16,7 +16,7 @@ import { parseUrl } from '../../../utils/routeUtils';
 import { OnBoardingAnalyticsEnum } from '../../../pages/quick-start/consts';
 import { useSegment } from '../../providers/SegmentProvider';
 import { useDigestDemoFlowContext } from './DigestDemoFlowProvider';
-import { NodeStep } from '../common';
+import { NodeStep } from '../../workflow-nodes';
```
LetItRock, line 19: moved to another folder, not specific to the quick-start

apps/web/src/components/workflow-nodes/NodeStep.tsx
```diff
@@ -1,21 +1,22 @@
...
+  className?: string;
   data: { label: string; email?: string };
   Handlers: React.FC<any>;
   Icon: React.FC<any>;
   ActionItem?: React.ReactNode;
   ContentItem?: React.ReactNode;
 };
 
-export function NodeStep({ Handlers, Icon, data, ActionItem, ContentItem }: NodeStepProps) {
+export function NodeStep({ className, Handlers, Icon, data, ActionItem, ContentItem }: NodeStepProps) {
```
LetItRock, line 14: allow to style it with `className` and styled

apps/web/src/pages/templates/components/templates-store/ChannelNode.tsx
```diff
@@ -0,0 +1,30 @@
+import { NodeProps, Handle, Position, getOutgoers, useReactFlow } from 'react-flow-renderer';
+import styled from '@emotion/styled';
+
+import { NodeStep } from '../../../../components/workflow-nodes';
+
+const NodeStepStyled = styled(NodeStep)`
+  width: 200px;
+`;
```
LetItRock, line 8: that's how easy is to update the styles

apps/web/src/pages/templates/components/templates-store/TemplatesStoreModal.tsx
```diff
@@ -23,6 +24,69 @@ import {
...
+  },
+  {
+    _id: 'push',
+    name: 'Push',
+    template: {
+      type: StepTypeEnum.PUSH,
+      content: 'Test push',
+    },
+  },
+];
```
LetItRock, line 89: mock steps

apps/web/src/pages/templates/components/templates-store/TemplatesStoreModal.tsx
```diff
@@ -164,7 +243,27 @@ export const TemplatesStoreModal = ({ isOpened, onClose }: ITemplatesStoreModalP
...
+                preventScrolling={false}
+                nodesDraggable={false}
+                elementsSelectable={false}
+                nodesConnectable={false}
+                selectNodesOnDrag={false}
+                wrapperStyles={{
+                  height: '100%',
+                  width: '100%',
+                  minHeight: '300px',
+                }}
```
LetItRock, line 264: reusing the same `FlowEditor` component as we do use in the Template Editor, allowing to pass any `react-flow` props

apps/web/src/pages/templates/components/templates-store/templateStoreStyles.ts
```diff
@@ -91,13 +92,18 @@ export const TemplateName = styled.div`
...
+  flex: 1 1 auto;
+  width: 100%;
   height: 100%;
 `;
 
 export const NovuButtonHolder = styled.div`
   position: absolute;
   bottom: 4px;
   right: 4px;
+  z-index: 10000;
```
LetItRock, line 106: the `Use template` button should be above the workflow editor

apps/web/src/pages/templates/workflow/WorkflowEditor.tsx
```diff
@@ -37,6 +50,20 @@ const WorkflowEditor = () => {
...
+
+      if (node.type === 'channelNode') {
+        navigate(basePath + `/${node.data.channelType}/${node.data.uuid}`);
+      }
+      if (node.type === 'triggerNode') {
+        navigate(basePath + '/test-workflow');
+      }
+    },
+    [basePath]
+  );
```
LetItRock, line 65: this code is moved from the `FlowEditor` to make it move "clean" and reusable

apps/web/src/pages/templates/workflow/WorkflowEditor.tsx
```diff
@@ -86,6 +113,10 @@ const WorkflowEditor = () => {
     setToDelete(uuid);
   };
 
+  const onStepInit = async () => {
+    await trigger('steps');
+  };
```
LetItRock, line 118: same

apps/web/src/pages/templates/workflow/WorkflowEditor.tsx
```diff
@@ -17,6 +17,18 @@ import { UpdateButton } from '../components/UpdateButton';
...
+import AddNode from './workflow/node-types/AddNode';
+import { AddNodeEdge } from './workflow/edge-types/AddNodeEdge';
+
+const nodeTypes = {
+  channelNode: ChannelNode,
+  triggerNode: TriggerNode,
+  addNode: AddNode,
+};
+
+const edgeTypes = { special: AddNodeEdge };
```
LetItRock, line 31: These were hardcoded before in the `FlowEditor` and they are using a lot of different functionalities, like `react-hook-form` etc., but it's only specific to the Template Editor. Now we can define whatever steps we want and we can reuse `FlowEditor`.

apps/web/src/pages/templates/workflow/workflow/FlowEditor.tsx
```diff
@@ -50,27 +40,59 @@ const initialNodes: Node[] = [
 
 const initialEdges: Edge[] = [];
 
+const DEFAULT_WRAPPER_STYLES = {
+  height: '100%',
+  width: '100%',
+  minHeight: '600px',
+};
```
LetItRock, line 47: were defined on the first div wrapper before

apps/web/src/pages/templates/workflow/workflow/FlowEditor.tsx
```diff
@@ -50,27 +40,59 @@ const initialNodes: Node[] = [
...
+    channelNode: ComponentType<NodeProps>;
+    addNode?: ComponentType<NodeProps>;
+  };
+  edgeTypes?: { special: ComponentType<EdgeProps> };
+  withControls?: boolean;
+  wrapperStyles?: React.CSSProperties;
+  onDelete?: (id: string) => void;
+  onStepInit?: (step: IFormStep) => Promise<void>;
+  addStep?: (channelType: StepTypeEnum, id: string, index?: number) => void;
+}
```
LetItRock, line 64: extends `ReactFlowProps` and adds some other useful props

apps/web/src/pages/templates/workflow/workflow/FlowEditor.tsx
```diff
@@ -50,27 +40,59 @@ const initialNodes: Node[] = [
...
+  defaultEdgeOptions = {
+    type: 'smoothstep',
+  },
+  zoomOnScroll = false,
+  preventScrolling = true,
+  nodesConnectable = false,
+  nodesDraggable = true,
+  minZoom = 0.5,
+  maxZoom = 1.5,
+  defaultZoom = 1,
```
LetItRock, line 79: before these were defined in the const `reactFlowDefaultProps`

apps/web/src/components/workflow/FlowEditor.tsx
```diff
@@ -159,19 +165,18 @@ export function FlowEditor({
...
-        await triggerErrors('steps');
+        await onStepInit?.(step);
 
         const newNode = buildNewNode(newId, oldNode, parentId, step, i);
-
-        const newEdge = buildNewEdge(parentId, newId);
-
         setNodes((nds) => nds.concat(newNode));
+
+        const newEdge = buildNewEdge(parentId, newId, edgeTypes ? 'special' : 'default');
```
p-fernandez, line 173: The condition of the third argument is a bit hidden and can be missed for someone not familiar with code. I'd suggest to extract it so it is a bit more clear.

apps/web/src/components/workflow/FlowEditor.tsx
```diff
@@ -332,9 +327,17 @@ export function FlowEditor({
...
+            preventScrolling={preventScrolling}
+            nodesConnectable={nodesConnectable}
+            nodesDraggable={nodesDraggable}
+            minZoom={minZoom}
+            maxZoom={maxZoom}
+            defaultZoom={defaultZoom}
+            {...restProps}
           >
-            <Controls />
+            {withControls && <Controls />}
```
p-fernandez, line 340: Thinking if we should pass `withControls` as a `show` property to the Controls component and the proper component to handle their rendering rather than the parent. What do you think?

apps/web/src/pages/templates/components/templates-store/ChannelNode.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  return (
+    <NodeStepStyled
+      data={data}
+      Icon={data.Icon}
+      Handlers={() => {
+        return (
+          <>
+            <Handle type="target" id="b" position={Position.Top} />
+            <Handle style={noChildStyle} type="source" id="a" position={Position.Bottom} />
```
p-fernandez, line 24: Would be a problem if multiple components share the same id? Even if being part of a parent component? Thinking in the reconciliation algorithm and when debugging.

p-fernandez (approved):
Just a code bits review. Looks good overall! 🌟

apps/web/src/components/workflow/FlowEditor.tsx
LetItRock: This `Controls` component is from the `react-flow-renderer` library and it doesn't have the `show` prop.
But still, I think it's better to do it on the parent level, otherwise, if you do it on the child level then you might register some hooks unnecessarily and the React will create a fiber for that component and will do all its logic aka reconciliation work.

apps/web/src/pages/templates/components/templates-store/ChannelNode.tsx
LetItRock: I'm not sure if this is even needed, but I've copied it from the other place 🥲
This `id` is not used as the HTML el `id` but as the internal library `handleId`.
It shouldn't have any impact on the reconciliation.

Issue #3180, opened 2023-04-11
[NV-2097] 🚀 Feature: Allow sender name to accept dynamic values


nocodeventure:
### 🔖 Feature description

The ability to add custom variables to buttons inside email templates, as well as the ability for the sender name to pick up on {{ORGANIZATION_NAME}}.

Buttons should also accept dynamic variables.

### 🎤 Why is this feature needed ?

This feature should be implemented because it provides more flexibility and customization options for email templates. With the ability to add custom variables to buttons, users can personalize the button text and link URLs based on the recipient's information or behavior. For example, if you want to create a personalized call-to-action button that leads to a specific landing page for each customer, you can use custom variables such as {{FIRST_NAME}} and {{LAST_NAME}} to generate unique URLs for each customer.

### ✌️ How do you aim to achieve this?

I'm not a developer

### 🔄️ Additional Information

Additionally, the ability for the sender name to pick up on {{ORGANIZATION_NAME}} can help users maintain a consistent branding experience for their email campaigns. For example, if you want to send an email campaign on behalf of your organization, you can use the {{ORGANIZATION_NAME}} variable in the sender name field, so that the recipient can easily recognize who the email is from.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-2097](https://linear.app/novu/issue/NV-2097/🚀-feature)</sub>

gitstart:
@scopsy I would like to pick this up

nocodeventure:
@gitstart How is this going? Appreciate your efforts.

gitstart:
> @gitstart How is this going? Appreciate your efforts.

We're still on it, @nocodeventure

gitstart:
> @gitstart How is this going? Appreciate your efforts.

@nocodeventure you can find the implemented fix on this [PR](https://github.com/novuhq/novu/pull/3453)

nocodeventure:
Amazing thank you @gitstart

yndawmarc:
Hi is there a way to override the subject?

PR #3453, opened 2023-05-18
NV-2097 - 🚀 Feature: Allow sender name to accept dynamic values

gitstart:
 Allows sender name to accept dynamic values, similar to `Subject` and `Preheader` on the email template.

closes https://github.com/novuhq/novu/issues/3180

### Other information
Found Out that, We can already use `custom variables` on for `button texts`, also we can use `existing variables` for URL that the button click will redirect to.
#### We can see the use of `custom variables` in button text in template 
image

#### Generated Template with variable's default value
image

[Loom Video](https://www.loom.com/share/31420c18129c4dc7993f5e307be5c8b5)

---


gitstart:

PR #3454, opened 2023-05-18
NV-2337 - Template Store - Add grouped blueprint cache

djabarovgeorge:


apps/api/src/app/blueprint/e2e/create-blueprint.e2e.ts
```diff
@@ -0,0 +1,115 @@
...
+        ],
+      },
+    ],
+  };
+
+  const { body: normalTemplate } = await session.testAgent.post(`/v1/notification-templates`).send(testTemplate);
+
+  process.env.BLUEPRINT_CREATOR = session.organization._id;
+
+  const { body } = await session.testAgent.post(`/v1/notification-templates`).send(testTemplate);
```
LetItRock, line 94: not sure why you are sending the two times same request, can you explain?

apps/api/src/app/blueprint/e2e/create-blueprint.e2e.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  before(async () => {
+    session = new UserSession();
+    await session.initialize();
+  });
+
+  it('should create template from blueprint', async function () {
+    const prodEnv = await getProductionEnvironment();
+
+    const { testTemplate, normalTemplate, prodVersionNotification, createdTemplate } =
```
LetItRock, line 22: these are super confused I'm not following what you are trying to do here

apps/api/src/app/blueprint/e2e/get-grouped-blueprints.e2e.ts
```diff
@@ -0,0 +1,43 @@
...
+    const data = await session.testAgent.get(`/v1/blueprints/group-by-category`).send();
+
+    expect(data.statusCode).to.equal(200);
+
+    const groupedBlueprints = data.body.data as GroupedBlueprintResponse[];
+
+    expect(groupedBlueprints[0].name).to.equal('General');
+
+    for (const grouped of groupedBlueprints) {
+      for (const blueprint of grouped.blueprints) {
```
LetItRock, line 32:
```suggestion
    for (const group of groupedBlueprints) {
      for (const blueprint of group.blueprints) {
```

LetItRock: I feel like we should also check some base fields that are needed for the FE, like id, name, description at least

packages/application-generic/src/services/cache/key-builders/entities.ts
```diff
@@ -79,11 +80,26 @@ const buildKeyById = ({
...
+const buildGroupedBlueprintsKey = ({
+  identifier,
+}: {
+  identifier: string;
+}): string =>
+  buildCommonKey({
+    type: CacheKeyTypeEnum.ENTITY,
+    keyEntity: CacheKeyPrefixEnum.GROUPED_BLUEPRINTS,
+    environmentIdPrefix: OrgScopePrefixEnum.ORGANIZATION_ID,
+    environmentId: process.env.BLUEPRINT_CREATOR,
```
LetItRock, line 92: isn't the `BLUEPRINT_CREATOR` env the `organizationId` ?

apps/api/src/app/blueprint/e2e/create-blueprint.e2e.ts
djabarovgeorge: Tend to agree it is a bit confusing it is a legacy code I refactored in the following PRs and hope to refactor more so it will be a more intuitive flow. The thing there is that those requests are different because of the  `process.env.BLUEPRINT_CREATOR = session.organization._id;` part that means that the second request mimics the creation inside of the blueprint organization.

apps/api/src/app/blueprint/e2e/create-blueprint.e2e.ts
djabarovgeorge: The entire test or those lines?
This is a legacy code that I refactored a bit in the following PRs, here I created createTemplateFromBlueprint in order to extract the logic of the creation of the template from the blueprint.

packages/application-generic/src/services/cache/key-builders/entities.ts
djabarovgeorge: Sorry for the confusion, I changed it to prod environmentId (because it is more accurate to fetch from prod env), and then (following PR) I reverted it back to organizationId in order to make it more intuitive and if we will need to prod env id we will fetch it.

djabarovgeorge: You are right and sorry for the confusion, I changed it to prod environmentId (because it is more accurate to fetch from prod env), and then (following PR) I reverted it back to organizationId in order to make it more intuitive and if we will need to prod env id we will fetch it.

apps/api/src/app/blueprint/e2e/get-grouped-blueprints.e2e.ts
djabarovgeorge: Good suggestion, I will add some.

apps/api/src/app/blueprint/e2e/create-blueprint.e2e.ts
LetItRock: I feel like we need to split this test into two separate, rn it checks if the blueprint will/won't be created, so these could be two tests.

djabarovgeorge: Good point, I will create a note for myself to revisit this test.

PR #3456, opened 2023-05-18
feat(api): implement feature flag generic use case

p-fernandez:

Implements a generic use case to retrieve a feature flag using the feature flag service and under the hood, Launch Darkly.

This use case can be reused across our system once the feature flags are set and tested.

Missing tests. If I don't do them for this PR while it is reviewed I will do them when implementing https://linear.app/novu/issue/NV-2325/create-template-store-feature-flag

### DIFF
--- a/apps/api/src/app/feature-flags/services/feature-flags.service.ts
+++ b/apps/api/src/app/feature-flags/services/feature-flags.service.ts
@@ -2,7 +2,7 @@ import { Injectable, Logger } from '@nestjs/common';
 
 import { LaunchDarklyService } from './launch-darkly.service';
 
-import { IFeatureFlagsService } from '../types';
+import { FeatureFlagKey, IFeatureFlagContext, IFeatureFlagsService } from '../types';
 
 const LOG_CONTEXT = 'FeatureFlagsService';
 
@@ -47,5 +47,9 @@ export class FeatureFlagsService {
         Logger.error('Feature Flags service has failed when shutted down', LOG_CONTEXT, error);
       }
     }
+  }
+
+  public async get<T>(key: FeatureFlagKey, context: IFeatureFlagContext, defaultValue: T): Promise<T> {
+    return await this.service.get(key, context, defaultValue);
   }
 }
--- a/apps/api/src/app/feature-flags/services/launch-darkly.service.ts
+++ b/apps/api/src/app/feature-flags/services/launch-darkly.service.ts
@@ -1,7 +1,7 @@
-import { init, LDClient } from 'launchdarkly-node-server-sdk';
+import { init, LDClient, LDContext, LDFlagValue, LDUser } from 'launchdarkly-node-server-sdk';
 import { Injectable, Logger } from '@nestjs/common';
 
-import { IFeatureFlagsService } from '../types';
+import { FeatureFlagKey, IFeatureFlagContext, IFeatureFlagsService } from '../types';
 
 const LOG_CONTEXT = 'LaunchDarklyService';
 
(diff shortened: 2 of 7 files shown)


apps/api/src/app/feature-flags/services/feature-flags.service.ts
```diff
@@ -48,4 +48,8 @@ export class FeatureFlagsService {
       }
     }
   }
+
+  public async get<T>(key: FeatureFlagKey, context: IFeatureFlagContext, defaultValue: T): Promise<T> {
+    return await this.service.get(key, context, defaultValue);
+  }
```
LetItRock, line 54: You don't need to declare this function `async` because you do return `Promise`, unless you plan to have another code here, and you don't need to `await`.

apps/api/src/app/feature-flags/services/launch-darkly.service.ts
```diff
@@ -39,4 +37,18 @@ export class LaunchDarklyService implements IFeatureFlagsService {
...
+      await this.client.waitForInitialization();
+      Logger.log('Launch Darkly SDK has been successfully initialized', LOG_CONTEXT);
+    } catch (error) {
+      Logger.error('Launch Darkly SDK has failed when initialized', LOG_CONTEXT, error);
+      throw error;
+    }
+  }
+
+  private mapToLaunchDarklyContext(context: IFeatureFlagContext): LDContext {
+    return context as LDContext;
```
LetItRock, line 52: I'm getting here the TS error:
Screenshot 2023-05-18 at 22 20 09

LetItRock: btw do you plan to have a logic here, because this casting doesn't feel right?

apps/api/src/app/feature-flags/use-cases/get-feature-flag/get-feature-flag.use-case.ts
```diff
@@ -0,0 +1,23 @@
...
+    const { key, environmentId, organizationId, userId } = command;
+
+    const context = {
+      environmentId,
+      organizationId,
+      userId,
+    };
+
+    // TODO: The type needs to be dynamic. String so far by default
+    await this.featureFlagsService.get<string>(key, context, '');
```
LetItRock, line 21: you don't have to specify generic type `<string>` here as it would be inferred from the default value

apps/api/src/app/feature-flags/services/launch-darkly.service.ts
p-fernandez: This bit is meant to be touched in this task: https://linear.app/novu/issue/NV-2324/mappers-around-launch-darkly-api
But I am unsure how the Launch Darkly context is going to be the best way, therefore this useless function.
I am fixing the build error though. Thanks for raising it.

apps/api/src/app/feature-flags/use-cases/get-feature-flag/get-feature-flag.use-case.ts
p-fernandez: Didn't know this one. Thanks!

apps/api/src/app/feature-flags/services/feature-flags.service.ts
LetItRock: `this.featureFlagsService.get<string>(key, context, '');` this will pass the compiler because you don't care about the promise result, but the promise will run, even with your code (async/await) it will pass.
But this `const test = this.featureFlagsService.get<string>(key, context, '');` will assign the promise to the constant and then TS will tell you that. Anyway, from what I remember in the compilation time compiler can improve that code ;)

LetItRock (approved):

apps/api/src/app/feature-flags/services/launch-darkly.service.ts
```diff
@@ -17,14 +17,12 @@ export class LaunchDarklyService implements IFeatureFlagsService {
...
 
-  public async initialize(): Promise<void> {
-    try {
-      await this.client.waitForInitialization();
-      Logger.log('Launch Darkly SDK has been successfully initialized', LOG_CONTEXT);
-    } catch (error) {
-      Logger.error('Launch Darkly SDK has failed when initialized', LOG_CONTEXT, error);
-      throw error;
-    }
+  public async get<T>(key: FeatureFlagKey, context: IFeatureFlagContext, defaultValue: T): Promise<T> {
```
djabarovgeorge, line 20: Now sure if the following suggestion is manageable 😅.
What do you think we will create a backoff functionality?
1. it could be another SDK of memory storage hardcoded/environment variable/or other.
2. it could be a provider base class that points to memory data.
3. it could be a feature-flag-service backoff if `get` failed.
4. or any other way.

The purpose is to backoff to a predefined flag if the service is not responsive.

The tricky part here is to manage this data. Maybe we can retrieve a JSON from launch darkly and store it as it is. But once again the managing part could be tricky.

Another option could be manual backoff per flag in the code. but this option could be less convenient because we will need to update the code => deploy

If you feel functionality is needed we can create a task for it.

PR #3458, opened 2023-05-18
[NV-2360] fix(web): fixed the issue when visiting the same workflow template editor twice

LetItRock:

This PR fixes the issue that happens when visiting the template editor twice for the same template. Currently, the form state is not "set", which results in the template steps not being loaded into the workflow. 

Also, there is another issue, like when the template is updated and the user navigates to the templates list page and back, then the old template state is shown.

This PR fixes both issues.

Fixes the above issue.

https://github.com/novuhq/novu/assets/2607232/c3d460b3-e78b-4e85-afb7-8b4b802c6631

https://github.com/novuhq/novu/assets/2607232/0b3891ab-49cc-498c-a72e-f7237b4348f2

### DIFF
--- a/apps/web/cypress/support/commands.ts
+++ b/apps/web/cypress/support/commands.ts
@@ -71,7 +71,11 @@ Cypress.Commands.add('clickNodeButton', (selector: string) => {
 
 Cypress.Commands.add(
   'initializeSession',
-  (settings: { disableLocalStorage?: boolean; noTemplates?: boolean } = { disableLocalStorage: false }) => {
+  (
+    settings: { disableLocalStorage?: boolean; noTemplates?: boolean; showOnBoardingTour?: boolean } = {
+      disableLocalStorage: false,
+    }
+  ) => {
     return cy.task('getSession', settings).then((response: any) => {
       if (!settings.disableLocalStorage) {
         window.localStorage.setItem('auth_token', response.token);
--- a/apps/web/cypress/tests/notification-editor/main-functionality.spec.ts
+++ b/apps/web/cypress/tests/notification-editor/main-functionality.spec.ts
@@ -386,4 +386,76 @@ describe('Workflow Editor - Main Functionality', function () {
 
     cy.getByTestId('control-add').first();
   });
+
+  it('should load successfully the recently created notification template, when going back from editor -> templates list -> editor', function () {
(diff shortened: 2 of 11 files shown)


apps/web/cypress/support/commands.ts
```diff
@@ -71,7 +71,11 @@ Cypress.Commands.add('clickNodeButton', (selector: string) => {
 
 Cypress.Commands.add(
   'initializeSession',
-  (settings: { disableLocalStorage?: boolean; noTemplates?: boolean } = { disableLocalStorage: false }) => {
+  (
+    settings: { disableLocalStorage?: boolean; noTemplates?: boolean; showOnBoardingTour?: boolean } = {
+      disableLocalStorage: false,
+    }
+  ) => {
```
LetItRock, line 78: Fixed the issue with flaky e2e tests that were failing in the case when the "start from the scratch tour" was shown which was covering the workflow editor drop-box, so the drop action couldn't perform.

apps/web/cypress/tests/notification-editor/main-functionality.spec.ts
```diff
@@ -386,4 +386,76 @@ describe('Workflow Editor - Main Functionality', function () {
 
     cy.getByTestId('control-add').first();
   });
+
+  it('should load successfully the recently created notification template, when going back from editor -> templates list -> editor', function () {
```
LetItRock, line 390: added two e2e tests to cover the bugs

apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
```diff
@@ -164,16 +164,18 @@ const TemplateEditorFormProvider = ({ children }) => {
...
+    useTemplateController(templateId);
+
+  useEffectOnce(() => {
+    if (!template) return;
+
+    // we don't call this on success because the templates might be fetched before
+    setTrigger(template.triggers[0]);
+    const form = mapNotificationTemplateToForm(template);
+    reset(form);
+  }, !!template);
```
LetItRock, line 177: The thing is that the `onFetchSuccess` aka `onSuccess` is called then the request completes successfully, but the thing is that the template might be loaded before and is cached by the react query, so it won't then trigger `onSuccess`, that's why the form state is not updated and in the result, steps are not set.

p-fernandez (approved):
Great fixes!
🌟

PR #3464, opened 2023-05-19
@novu/node: Add transaction id to broadcast payload options

RtiM0:

This will let the users set a custom transactionId in the broadcast event.

Setting transactionId is useful for broadcast events so that events can be easily programmatically cancelled if the need arises. Transaction ID is already supported by the API:
https://github.com/novuhq/novu/blob/next/apps/api/src/app/events/events.controller.ts#LL117C22-L117C22


scopsy:
@jainpawan21 indeed! @RtiM0 have you had a chance to test this? Just to verify that the controller on the API actually passes it downstream

RtiM0:
@scopsy I raised this PR just to get the sdk to be able to send a transaction ID to the API, which API already accepts as seen here:
https://github.com/novuhq/novu/blob/next/apps/api/src/app/events/events.controller.ts#L133

But if I remember correctly, transaction ids were not assigned to all notifications as intended when the notifications were actually triggered.
This was on an older version of Novu, so I'm sorry I can't be conclusive here.

scopsy (approved):

scopsy (approved):

PR #3465, opened 2023-05-19
feat(api): implement mappers for the different contexts in feature flags

p-fernandez:

Adds mapping for the different potential contexts we might have for the feature flags (organization, environment, user).

We need to adapt Novu's context to Launch Darkly's one.

I am missing for now adding other contexts such as version and region that were discussed but we still don't have the implementation for them in the apps. The implementation of the contexts in this PR should be enough to give an idea how to implement them in the future.

### DIFF
--- a/apps/api/src/app/feature-flags/services/feature-flags.service.ts
+++ b/apps/api/src/app/feature-flags/services/feature-flags.service.ts
@@ -49,7 +49,8 @@ export class FeatureFlagsService {
     }
   }
 
-  public async get<T>(key: FeatureFlagKey, context: IFeatureFlagContext, defaultValue: T): Promise<T> {
-    return await this.service.get(key, context, defaultValue);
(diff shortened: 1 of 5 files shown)


apps/api/src/app/feature-flags/services/feature-flags.service.ts
```diff
@@ -49,7 +49,8 @@ export class FeatureFlagsService {
     }
   }
 
-  public async get<T>(key: FeatureFlagKey, context: IFeatureFlagContext, defaultValue: T): Promise<T> {
-    return await this.service.get(key, context, defaultValue);
+  public async get<T>(key: FeatureFlagKey, defaultValue: T, context: IFeatureFlagContext): Promise<T> {
```
p-fernandez, line 52: swapped the order because I found it was a clearer API as the key and the default value are related and context might be optional in a further decision.

PR #3467, opened 2023-05-21
NV-2335 - Template Store - Template store refactor web on blueprint creation to use

djabarovgeorge:

At the moment we are using the /notification-template/:templateId/blueprint POST, we need to update this flow to create the template using the regular /notification-template POST request that is used by the client.

The additional step that we will need to refactor is instead of sending the id of the blueprint we should pass the template snapshot.

1. Reuse the existing route.
2. In case we get calls from different environments (US/EU...) the current logic won't work because not all the environments we have the blueprint stored, in order to solve this issue we can send a snapshot of the template, the same way we do in the  /notification-template POST request. In order to simplify the flow we can use the same route.


apps/web/src/pages/templates/components/BlueprintModal.tsx
```diff
@@ -118,7 +126,8 @@ export function BlueprintModal() {
             data-test-id="create-from-blueprint"
             onClick={() => {
               if (blueprintId) {
-                mutate(blueprintId);
+                const templatePayload = mapBlueprintToTemplateCreatePayload();
+                mutate(templatePayload);
```
LetItRock, line 130: I would move this `mapBlueprintToTemplateCreatePayload` logic to the `useMutation` first argument, like:
```
useMutation(() => {
    const templatePayload =  { ...blueprint };
    templatePayload.notificationGroupId = templatePayload._notificationGroupId;
    templatePayload.blueprintId = templatePayload._id;

    return createTemplate(templatePayload);
}, {
...
});


```

apps/api/src/app/notification-template/dto/create-notification-template.request.dto.ts
```diff
@@ -14,7 +14,7 @@ export class CreateNotificationTemplateRequestDto implements ICreateNotification
   @ApiProperty()
   @IsString()
   @IsDefined({
-    message: 'Notification group must be provided',
+    message: 'Notification group must be provided ',
```
LetItRock, line 17: do we need that space?

LetItRock (approved):

apps/api/src/app/notification-template/dto/create-notification-template.request.dto.ts
djabarovgeorge: I found that if we won't have it that the next argument will miss a space, so I added it here but if it is persistent in all cases maybe looking at how to fix it globally will be a better idea.

PR #3468, opened 2023-05-21
feat: client points to novu cms centralized server

djabarovgeorge:

### DIFF
--- a/apps/web/src/api/api.client.ts
+++ b/apps/web/src/api/api.client.ts
@@ -2,9 +2,9 @@ import axios from 'axios';
 import { API_ROOT } from '../config';
 
 export const api = {
-  get(url: string) {
+  get(url: string, absoluteUrl = false) {
     return axios
-      .get(`${API_ROOT}${url}`)
+      .get(buildUrl(url, absoluteUrl))
       .then((response) => {
         return response.data?.data;
       })
@@ -61,3 +61,7 @@ export const api = {
       });
   },
 };
+
+function buildUrl(url: string, absoluteUrl: boolean) {
+  return absoluteUrl ? url : `${API_ROOT}${url}`;
+}
--- a/apps/web/src/api/api.client.ts
+++ b/apps/web/src/api/api.client.ts
@@ -1,10 +1,13 @@
 import axios from 'axios';
 import { API_ROOT } from '../config';
 
+interface IOptions {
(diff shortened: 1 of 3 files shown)


apps/web/src/config/index.ts
```diff
@@ -14,6 +14,11 @@ export const API_ROOT =
     ? window._env_.REACT_APP_API_URL || process.env.REACT_APP_API_URL || 'http://localhost:1336'
     : window._env_.REACT_APP_API_URL || process.env.REACT_APP_API_URL || 'http://localhost:3000';
 
+export const CMS_ROOT_URL =
+  window._env_.REACT_APP_CMS_ROOT_URL || isCypress
+    ? window._env_.REACT_APP_CMS_ROOT_URL || process.env.REACT_APP_CMS_ROOT_URL || 'http://localhost:1336'
+    : 'https://api.novu.co';
```
djabarovgeorge, line 20: This way all the clients including the self-hosted will point to the same server.

djabarovgeorge, line 20: This way all the clients including the self-hosted (besides the test env) will point to the same server.

apps/web/src/api/api.client.ts
```diff
@@ -2,9 +2,9 @@ import axios from 'axios';
 import { API_ROOT } from '../config';
 
 export const api = {
-  get(url: string) {
+  get(url: string, absoluteUrl = false) {
```
LetItRock, line 5: please make this second arg as options object like: `{ absoluteUrl }: { absoluteUrl?: boolean } = {}`, then it will be more readable from the plate where the `get` is used

apps/web/src/config/index.ts
LetItRock: To me the CMS is confusing still, generally it is associated with the headless CMS services like Contentful etc.
Maybe we should change it to the `BLUEPRINTS_API_URL` ?

LetItRock (approved):

apps/web/src/api/api.client.ts
djabarovgeorge: Good idea!!

apps/web/src/config/index.ts
djabarovgeorge: Yes, I can understand that I have no objections to the suggestion :) will update it.

PR #3476, opened 2023-05-23
fix: create and delete feed changes 

ainouzgali:

create and then delete feed creates two changes to promote, depending on the order of promotion the deleted feed will be created in prod with no option of deleting it.


apps/api/src/app/change/usecases/promote-feed-change/promote-feed-change.ts
```diff
@@ -17,6 +17,10 @@ export class PromoteFeedChange {
     }
 
     if (!item) {
+      if (command.item.deleted) {
+        return;
+      }
```
p-fernandez, line 22: I feel it doesn't make sense to have it here. If the item passed in the command has the deleted property, it will come already so we can check it and do the early return before even we do the repository call, saving one hit to the database.

p-fernandez (commented):
I have one concern about the logic.

apps/api/src/app/change/usecases/promote-feed-change/promote-feed-change.ts
p-fernandez: But for the case you mention I would expect `item.deleted` to not be set. If it is passed I think something would be very wrong.

ainouzgali: Notice that item is the prod feed and command.item is the dev feed. 

Say a feed was created in dev, then deleted in dev. Then promoted.

The command.item is the aggregated item of all changes in dev-> which means that command.item should also have the deleted flag. 

It looks for the feed in prod, which doesn't exist - as it was never created. And then if the promotion is of a feed that was deleted anyway - there is no reason to do anything for prod env. 
 
But I might be missing something.

p-fernandez: Ok, this changes things a bit. I would rename item to `productionItem` as from the variables we have no way to know that we are looking for an item in production.

p-fernandez (approved):
🌟

PR #3480, opened 2023-05-23
Nv 2212 predefined sms provider

BiswaViraj:

Novu SMS provider using Twilio


apps/web/src/hooks/integrations/useIntegrationLimit.ts
```diff
@@ -5,22 +5,14 @@ import { useMemo } from 'react';
 import { IS_DOCKER_HOSTED } from '../../config/index';
 import { useActiveIntegrations } from './useActiveIntegrations';
 
+const isLimitFetchingEnabled = !IS_DOCKER_HOSTED;
 export function useIntegrationLimit(type: ChannelTypeEnum) {
-  const { integrations = [] } = useActiveIntegrations();
-
-  const isLimitFetchingEnabled = useMemo<boolean>(() => {
-    return !IS_DOCKER_HOSTED && integrations.filter((integration) => integration.channel === type).length === 0;
```
p-fernandez, line 12: What's the reason behind removing the check if there are integrations or no?

apps/web/src/pages/integrations/IntegrationsStoreModal.tsx
```diff
@@ -21,6 +22,7 @@ import { Close } from '../../design-system/icons/actions/Close';
 import { useProviders } from './useProviders';
 import { useSegment } from '../../components/providers/SegmentProvider';
 import { IntegrationsStoreModalAnalytics } from './constants';
+import { NovuSMSProviderModal } from './components/NovuSMSProviderModal';
```
p-fernandez, line 25:
```suggestion
import { NovuSmsProviderModal } from './components/NovuSmsProviderModal';
```
So we follow the convention we tend to use (like in `SmsProviderIdEnum`.

apps/web/src/pages/integrations/components/LimitBar.tsx
```diff
@@ -1,85 +1,103 @@
...
 }: {
   limit: number;
   count: number;
   channel?: ChannelTypeEnum;
-  label?: string | null;
+  showDescription?: boolean;
 }) => {
   const unit = useMemo<string>(() => {
-    return channel === ChannelTypeEnum.EMAIL ? 'emails' : 'sms';
+    return channel === ChannelTypeEnum.EMAIL ? 'emails' : 'messages';
```
p-fernandez, line 60: I think since we removed the Tooltip component this `useMemo` is not needed anymore as there won't be child components rerenders and we can just calculate the string.
CC: @LetItRock am I wrong with this?

apps/web/src/pages/integrations/components/NovuProviderBase.tsx
```diff
@@ -0,0 +1,90 @@
...
+      }}
+    >
+      <CloseButton data-test-id="connection-integration-form-close" type="button" onClick={onClose}>
+        <Close />
+      </CloseButton>
+
+      <Stack spacing={24}>
+        <div>
+          <img src={logoSrc} alt="logo" style={{ maxWidth: 130, maxHeight: 32 }} />
+        </div>
```
p-fernandez, line 45: I think we should make the logo an own component if it doesn't exist yet, with the `maxWidth` and `maxHeight` properties passed and maybe automatically calculating the path and file name of the logo for dark and light versions, so we can standardise the logo naming.

apps/web/src/pages/integrations/components/NovuSMSProviderModal.tsx
```diff
@@ -0,0 +1,6 @@
+import { ChannelTypeEnum } from '@novu/shared';
+import { NovuProviderBase } from './NovuProviderBase';
+
+export function NovuSMSProviderModal({ onClose }: { onClose: () => void }) {
```
p-fernandez, line 4:
```suggestion
export function NovuSmsProviderModal({ onClose }: { onClose: () => void }) {
```
Reminder as one of the first comments in the PR.

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -26,14 +30,28 @@ export class GetNovuIntegration {
...
+    ) {
+      return;
+    }
+
+    if (
+      channelType === ChannelTypeEnum.SMS &&
+      !process.env.NOVU_SMS_INTEGRATION_ACCOUNT_SID &&
+      !process.env.NOVU_SMS_INTEGRATION_TOKEN &&
+      !process.env.NOVU_SMS_INTEGRATION_SENDER
+    ) {
```
p-fernandez, line 47: This validation checks are used in 2 different use cases. We need to find a way to isolate them in one single place and reuse them.

p-fernandez (approved):
🌟

PR #3484, opened 2023-05-23
feat(web): templates store - fe caching for blueprints

LetItRock:

Added the FE caching logic for the blueprints, they will be cached for 1 week in the `indexedDB`.

Templates Store PRD.

Screenshot 2023-05-23 at 14 05 16


djabarovgeorge (approved):
If it works as expected then it looks good to me 🚀

small note not related to this pr, if I will have spare time I was thinking to get the same hardcoded popular templates (from @shared) to the client, if we will add this logic I think we can definitely reduce the client cache to 1 day, I really afraid we will abuse this client caching really fast 😅

PR #3488, opened 2023-05-24
withinPortal in PopoverNotificationCenter to be true by default

donnguyen:

Add `withinPortal` to the PopoverNotificationCenter.

Unlike Mantine 4, where the `withinPortal` prop was true by default and made the popover attach to the body element, Mantine 5 has it set to false by default. This means the popover is rendered inline with the parent element. We want to avoid positioning issues, especially when we have a notification bell/button in the header or sidebar (most of the cases we will do this), then set `withinPortal` to true and make the popover appear on top of everything else.


p-fernandez (commented):
Would it be possible to see different screenshots of how this change looks like in the UI? 🙏🏻

donnguyen:
@p-fernandez sure thing.

Say we have a sidebar like this:
image
Without the change, it would look like this when we click on Notification menu item:
image
You can see the overflow + position issue, because the popover is rendered inline with the parent element.

With the change, it would look like this:
image
Because the popover is now rendered inside the <body> tag.

p-fernandez (approved):
🌟

PR #3489, opened 2023-05-24
feat(one-signal): add override to use include_external_user_ids options

### LINKED ISSUE #3479: 🚀 Feature: OneSignal provider overrides to send push notification with external_user_id
### 🔖 Feature description

Hi, I have seen the onesignal feature provided in latest `0.15.0` release but it seems for now novu is only using `player_id` to target the push notification. I think the `external_user_id` could be easier to use in some scenario (OneSignal handle the device and user relationship instead of Novu). Would it be better if there's some override setting specific for OneSignal?

In my use-case there's still some limitation on how to manage `deviceTokens` with Novu `setCredentials` method so it would be easier if I can use overrides value to target the push notification with `external_user_id` instead of `player_id`

A new overrides key specific to OneSignal provider should be used and if that key is available, do not use `include_player_ids` args in OneSignal [createNotification](https://documentation.onesignal.com/reference/create-notification) method and use `include_external_user_ids` and `channel_for_external_user_ids='push'` instead

### 🔄️ Additional Information

_No response_

im-perativa:

Add new override key specific to OneSignal provider to use `include_external_user_ids` instead of `include_player_ids` in OneSignal [createNotification](https://documentation.onesignal.com/reference/create-notification) method 

Closes #3479


packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -64,6 +64,9 @@ export interface IPushOptions {
       };
     };
     fcmOptions?: { analyticsLabel?: string };
+    'one-signal'?: {
```
p-fernandez, line 67: Either this or the `ITriggerOverrides` in `events.interface` is wrong.
We should add tests to the provider to double check the right implementation.

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -64,6 +64,9 @@ export interface IPushOptions {
...
+      externalIds?: string[];
```
p-fernandez, line 68:
```suggestion
      externalUserIds?: string[];
```

p-fernandez (changes requested):
Thanks for your contribution!

packages/stateless/src/lib/provider/provider.interface.ts
im-perativa: Thanks for the feedback. Forgot about that one. Will try to fix and include the docs tomorrow

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -155,7 +155,7 @@ export class SendMessagePush extends SendMessageBase {
         continue;
       }
 
-      const overrides = command.overrides[integration.providerId] || {};
+      const overrides = Object.assign(command.overrides, command.overrides[integration.providerId]);
```
p-fernandez, line 158: What's the reason behind this change? I am afraid that it can led to buggy behaviour.

providers/one-signal/src/lib/one-signal.provider.ts
```diff
@@ -29,7 +29,12 @@ export class OneSignalPushProvider implements IPushProvider {
     const { sound, badge, ...overrides } = options.overrides ?? {};
 
     const res = await this.oneSignal.createNotification({
-      include_player_ids: options.target,
+      ...(overrides['one-signal']?.externalUserIds !== undefined
+        ? {
+            include_external_user_ids: overrides['one-signal'].externalUserIds,
+            channel_for_external_user_ids: 'push',
+          }
+        : { include_player_ids: options.target }),
```
p-fernandez, line 37:
```suggestion
   const { externalUserIds } =  overrides['one-signal'] || {};
   const res = await this.oneSignal.createNotification({ 
      ...(externalUserIds !== undefined
        ? {
            include_external_user_ids: externalUserIds,
            channel_for_external_user_ids: 'push',
          }
        : { include_player_ids: options.target }),
```
We can reuse the extraction of the `externalUserIds`.
Any way of doing it should be enough:
```
   const externalUserIds =  overrides['one-signal']?.externalUserIds;
```

im-perativa:
> @im-perativa I saw you requested a re-review but there is an open pending question to address. 🙏🏻

Sorry I didn't know my previous comment on your question was still on pending state 😅

docs/docs/channels/push/one-signal.md
```diff
@@ -30,6 +30,8 @@ await novu.subscribers.setCredentials('subscriberId', PushProviderIdEnum.OneSign
 });
 ```
 
+If your user have multiple devices and you want to leave the device-to-user relationship to OneSignal, you can use `externalUserIds` override to send push notifications to target [external_user_ids](https://documentation.onesignal.com/docs/external-user-ids) and skip the registration of [player_id](https://documentation.onesignal.com/docs/users#player-id) as `deviceToken` in Novu. You must set `external_user_ids` with OneSignal [setExternalUserId](https://documentation.onesignal.com/docs/external-user-ids#setexternaluserid-method) method to associate multiple `player_id` with `external_user_id` in your client first.
+
```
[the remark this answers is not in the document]
im-perativa: @L-U-C-K-Y sure, I haven't got the time to read the full deprecation notice thoroughly but I think we should update this MR first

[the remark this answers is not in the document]
L-U-C-K-Y: @im-perativa I agree, let's upgrade to latest OneSignal SDK and rename to external_id, I think this PR in general would be very useful. Not sure if I can work on it this week, but for sure I can review.

scopsy:
@im-perativa we have recently upgraded to the latest one signal client. I'm closing this PR, as I think it will be easier to just recreate the change instead of merging conflicts. My apologies for this PR has gotten stale from our side 🙏

PR #3493, opened 2023-05-24
Nv 2231 Template Store - template store analytics

djabarovgeorge:

We need to be able to answer the following questions with the new events

Key Event: How many people create a workflow from a template store
[1] event on template creation from the template store.

What are the most popular templates in the store?
[1] event on template creation from the template store.
[2] event on the template click on the template store.

How many people opened the template store but haven’t used it?
[3] event on the template store opened.
[1] event on template creation from the template store.

Did the average number of templates for an account go higher after the release of the feature
[4] event on template creation (we probably already have this one)

NOTE: I added a couple of events we could add in order to answer those questions. I numbered them so we can see that there are the same events that we need for different questions.

[2] === [3]


apps/web/src/pages/templates/TemplatesListNoData.tsx
```diff
@@ -154,7 +168,14 @@ export const TemplatesListNoData = ({
...
                     data-test-id="second-workflow-tile"
                     disabled={readonly || isCreating}
-                    onClick={() => onTemplateClick(template)}
+                    onClick={() => {
+                      segment.track('[Template Store] Click Create Notification Template', {
+                        templateIdentifier: template?.triggers[0]?.identifier || '',
+                        location: TemplateCreationSourceEnum.EMPTY_STATE,
+                      });
+
+                      onTemplateClick(template);
```
LetItRock, line 177: we will have 2 analytic events: one from here and one triggered on the BE, is that what we want?

apps/web/src/api/hooks/notification-templates/useCreateTemplateFromBlueprint.ts
```diff
@@ -14,6 +14,7 @@ const mapBlueprintToTemplate = (blueprint: INotificationTemplate): ICreateNotifi
   critical: blueprint.critical,
   preferenceSettings: blueprint.preferenceSettings,
   blueprintId: blueprint._id,
+  __source: blueprint.__source,
```
LetItRock, line 17: The `__source` is not a part of the `blueprint/template` it's something additional - metadata, then I feel like we shouldn't send it in the body, but in the query params instead.

apps/web/src/components/quick-start/in-app-onboarding/TriggerNode.tsx
```diff
@@ -80,6 +81,7 @@ function TriggerButton({ setOpened }: { setOpened: (value: boolean) => void }) {
     async function createOnBoardingTemplate() {
       const payloadToCreate = {
         notificationGroupId: groups[0]._id,
+        isBlueprint: false,
```
LetItRock, line 84: do we have to set it?

djabarovgeorge: Added this flag to make sure that there won't be a case where this template creates by mistake which will result in a blueprint in the Template Store

apps/web/src/pages/templates/TemplatesListNoData.tsx
djabarovgeorge: That is true @LetItRock , was not sure if we want to separate those api-clients events or not. do you think we can drop this client event?

apps/web/src/api/hooks/notification-templates/useCreateTemplateFromBlueprint.ts
djabarovgeorge: Done :)

apps/web/src/pages/templates/TemplatesListNoData.tsx
djabarovgeorge: Will merge for now we always can remove it.

djabarovgeorge: Let's merge this for now, we can always remove it.

djabarovgeorge: Let's merge this for now, we can always remove it.
@LetItRock  @scopsy

LetItRock (approved):

PR #3496, opened 2023-05-24
feat: add framework select accordion

davidsoderberg:
Screenshot 2023-05-24 at 15 50 52

### DIFF
--- a/apps/api/src/app/integrations/usecases/get-In-app-activated/get-In-app-activated.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/get-In-app-activated/get-In-app-activated.usecase.ts
@@ -10,7 +10,7 @@ export class GetInAppActivated {
     const inAppSubscriberCount = await this.subscriberRepository.count({
       _organizationId: command.organizationId,
       _environmentId: command.environmentId,
-      isOnline: { $exists: true },
+      isOnline: true,
       subscriberId: /on-boarding-subscriber/i,
     });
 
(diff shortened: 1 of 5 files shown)


apps/api/src/app/integrations/usecases/get-In-app-activated/get-In-app-activated.usecase.ts
```diff
@@ -10,7 +10,7 @@ export class GetInAppActivated {
     const inAppSubscriberCount = await this.subscriberRepository.count({
       _organizationId: command.organizationId,
       _environmentId: command.environmentId,
-      isOnline: { $exists: true },
+      isOnline: true,
```
davidsoderberg, line 13: Is not this what we really like to check and not only that the isOnline field exists on the subscriber?

apps/api/src/app/integrations/usecases/get-In-app-activated/get-In-app-activated.usecase.ts
p-fernandez, line 1: Inherited but this filename casing could be problematic.

apps/web/src/pages/integrations/components/FrameworkDisplay.tsx
```diff
@@ -0,0 +1,128 @@
...
+  JsGradient,
+  CodeGradient,
+} from '../../../design-system/icons';
+import { useEnvController } from '../../../hooks';
+
+export const FrameworkDisplay = ({ setFramework }: { setFramework: (framework: string) => void }) => {
+  const { environment } = useEnvController();
+  const theme = useMantineTheme();
+  const clipboardEnvironmentIdentifier = useClipboard({ timeout: 1000 });
+  const environmentIdentifier = environment?.identifier ? environment.identifier : '';
```
p-fernandez, line 18:
```suggestion
  const environmentIdentifier = environment?.identifier ?? '';
```

p-fernandez (approved):

apps/api/src/app/integrations/usecases/get-In-app-activated/get-In-app-activated.usecase.ts
davidsoderberg: I guess you are thinking about the `I` that should be `i`?

PR #3499, opened 2023-05-24
Add Python to the list of available SDKs in the README

jensenbox:

A hyperlink to the Python SDK

There appears to be an SDK that supports Python. Initially I avoided using Novu because of the lack of Python support but now that I found it I will give it a try and do what I can to support the Python SDK.


scopsy (approved):
Thank you!

PR #3500, opened 2023-05-24
feat: load balance between environments

djabarovgeorge:

make sure that all environments that are not prod will point to dev api.

remove the load of the requests from prod api.

the things that are still missing:
* create cms novu org in dev and create blueprints.
* after the templates are created updated the const popular blueprints in the code.
* make sure we update dev environment variable: BLUEPRINT_CREATOR.


apps/web/src/config/index.ts
```diff
@@ -27,6 +22,13 @@ export const SENTRY_DSN = window._env_.REACT_APP_SENTRY_DSN || process.env.REACT
 
 export const ENV = window._env_.REACT_APP_ENVIRONMENT || process.env.REACT_APP_ENVIRONMENT;
 
+const blueprintApiUrlByEnv = ENV === 'production' || ENV === 'prod' ? 'https://api.novu.co' : 'https://dev.api.novu.co';
+
+export const BLUEPRINTS_API_URL =
+  window._env_.REACT_APP_BLUEPRINTS_API_URL || isCypress
+    ? window._env_.REACT_APP_BLUEPRINTS_API_URL || process.env.REACT_APP_BLUEPRINTS_API_URL || 'http://localhost:1336'
```
LetItRock, line 29: `window._env_.REACT_APP_BLUEPRINTS_API_URL` allows to override that URL, should we keep that?

apps/api/src/app/blueprint/usecases/get-grouped-blueprints/consts.ts
```diff
@@ -2,90 +2,189 @@
...
+        isBlueprint: true,
+      },
+    ] as INotificationTemplate[],
+  };
+
+  return { templates: POPULAR_TEMPLATES_GROUPED };
 };
+
+export const POPULAR_TEMPLATES_GROUPED =
+  process.env.NODE_ENV === 'production' ? production().templates : development().templates;
```
LetItRock, line 190: why do we want to differentiate?

djabarovgeorge: because the ids will be different in different environments.

apps/web/src/config/index.ts
djabarovgeorge: we definitely can remove it, I left it is just in case we would want to override it.

ainouzgali (approved):

PR #3502, opened 2023-05-25
Template Store - Update blueprint ids

djabarovgeorge:


apps/api/src/app/blueprint/usecases/get-grouped-blueprints/consts.ts
```diff
@@ -1,109 +1,2 @@
...
-          type: 'event',
-          identifier: 'template_identifier_3',
-        },
-      ],
-      isBlueprint: true,
-    },
-  ] as INotificationTemplate[],
-};
+export const POPULAR_GROUPED_NAME = 'Popular';
+export const POPULAR_TEMPLATES_ID_LIST = ['646c77cf693b8e668a900a73', '646f123c720b54f89ed2130a'];
```
djabarovgeorge, line 2: Due to the reason that we decided to not use static values at the moment in the client, we do not need the metadata, therefore I decided to simplify the data here to the ids.

apps/api/src/app/blueprint/e2e/get-grouped-blueprints.e2e.ts
```diff
@@ -99,53 +97,6 @@ describe('Get grouped notification template blueprints - /blueprints/group-by-ca
     expect(updatedGroupedBluePrints[1].name).to.equal(categoryName);
   });
 
-  it('should get the popular blueprints', async function () {
```
djabarovgeorge, line 102: We won't need the following tests because they checked mainly the static data.

apps/api/src/app/blueprint/usecases/get-grouped-blueprints/get-grouped-blueprints.usecase.ts
```diff
@@ -52,12 +52,12 @@ export class GetGroupedBlueprints {
...
-    const localPopularBlueprints = { ...POPULAR_TEMPLATES_GROUPED };
+    const localPopularIds = [...POPULAR_TEMPLATES_ID_LIST];
 
-    return localPopularBlueprints.blueprints.map((localBlueprint) => {
-      const storedBlueprint = storedBlueprints.find((blueprint) => blueprint._id === localBlueprint._id);
+    return localPopularIds.map((localBlueprintId) => {
+      const storedBlueprint = storedBlueprints.find((blueprint) => blueprint._id === localBlueprintId);
 
-      return (storedBlueprint ? storedBlueprint : localBlueprint) as INotificationTemplate;
+      return (storedBlueprint ? storedBlueprint : localBlueprintId) as INotificationTemplate;
```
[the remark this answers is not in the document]
djabarovgeorge: > wait, something is weird here, this `updatePopularBlueprints` should return the blueprints, but not their `id`'s

You are correctly right, but i refactored a bit the code to return only the ids because at the moment we are not using the objects. If we will add the data to client (through @Shared i will return the metadata.) [REF](https://github.com/novuhq/novu/pull/3502#discussion_r1205391918)



> also it should take into consideration the environment, where we are currently running, because id's from prod won't be the same as id's from dev

I made the adjustment on previous PR :) [REF](https://github.com/novuhq/novu/pull/3500#discussion_r1207004322)

ainouzgali, line 60: I don't follow on this one. It seems it either returns a blueprint (storedBlueprint which is type INotificationTemplate) or just an id (localBlueprintId which is string), but then cast to INotificationTemplate.
What am I missing?

djabarovgeorge: Totally right missed the update here, let me do quick fix here 🫠

apps/api/src/app/blueprint/usecases/get-grouped-blueprints/get-grouped-blueprints.usecase.ts
```diff
@@ -37,27 +37,27 @@ export class GetGroupedBlueprints {
...
-    return localPopularBlueprints.blueprints.map((localBlueprint) => {
-      const storedBlueprint = storedBlueprints.find((blueprint) => blueprint._id === localBlueprint._id);
+      if (!storedBlueprint) {
+        Logger.warn(
+          `Could not find stored popular blueprint id: ${localBlueprintId}, BLUEPRINT_CREATOR: ${NotificationTemplateRepository.getBlueprintOrganizationId()}`
+        );
+      }
 
-      return (storedBlueprint ? storedBlueprint : localBlueprint) as INotificationTemplate;
+      return storedBlueprint as INotificationTemplate;
```
ainouzgali, line 60: So an empty array is ok to return?

ainouzgali (approved):

apps/api/src/app/blueprint/usecases/get-grouped-blueprints/get-grouped-blueprints.usecase.ts
djabarovgeorge: It should not happen, but it is possible 😅

PR #3504, opened 2023-05-25
feat(web): implement is template store enabled feature flag in web

p-fernandez:

Implements Launch Darkly React client SDK and the feature flag needed for the Template Store feature.

We had the implementation in the backend but we needed the frontend implementation.


LetItRock (approved):
🎸 ⭐

PR #3506, opened 2023-05-25
fix: add user id in boot to remove error

jainpawan21:

Add user._id in boot method of intercom

To fix error in identity verification (suggested by intercom)


BiswaViraj (approved):
🌟

PR #3508, opened 2023-05-25
feat: Export additional types and enums from sdk

### LINKED ISSUE #3442: 🚀 Feature: Export Notification Templates interfaces/enums from SDK
### 🔖 Feature description

SDK types are not being exported for consumers, example: `IMessageFilter` interface can't be imported from `@novu/node`

I'm building a library that generates Novu resources/entities programmatically and there's the need to have type-safety but it is hard to import the types from `@novu/node` as enums such as `DigestUnitEnum`, `BuilderFieldType` etc or any other interface related to notification templates (which is my primary use case) are not available 

The workaround that I have so far is using typescript `Parameters``
```typescript
import type { Novu } from '@novu/node';

export type SdkTemplate = Parameters<
  Novu['notificationTemplates']['create']
>[0];
```

This is not ideal as for some enum fields, we're not even able to pass a literal value and i end up casting to `any`

If possible export the types/enums so that the SDK can import, something like

```typescript
import type { INotificationTemplateStep, INotificationTemplatePayload,  ChannelCTATypeEnum } from '@novu/node'
```

### 🔄️ Additional Information

This might apply to other resources, right now I'm only exploring templates 

underfisk:
This PR exports additional enums and types that are important when building programmatic resource generation as these types facilitate the type-safety.

I'm open to suggestions if this doesn't feel like the right thing to do but I would like to access as much as possible the underlying types references

Closes #3442

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -1,4 +1,17 @@
-export { ChatProviderIdEnum, PushProviderIdEnum } from '@novu/shared';
+export {
+  ChatProviderIdEnum,
+  PushProviderIdEnum,
+  ChannelCTATypeEnum,
+  TemplateVariableTypeEnum,
+  IMessageTemplate,
+  SystemAvatarIconEnum,
+  INotificationTemplate,
+  INotificationTemplateStep,
+  ITemplateVariable,
+  IEmailBlock,
+  TextAlignEnum,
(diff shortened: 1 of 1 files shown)


jainpawan21 (approved):
Thanks 🙏🏻

PR #3510, opened 2023-05-26
feat: add check for at signs in redirect url

davidsoderberg:
Fixing the redirect vulnerability in github oauth endpoint.

### DIFF
--- a/apps/api/src/app/auth/auth.controller.ts
+++ b/apps/api/src/app/auth/auth.controller.ts
@@ -91,12 +91,7 @@ export class AuthController {
      * Make sure we only allow localhost redirects for CLI use and our own success route
      * https://github.com/novuhq/novu/security/code-scanning/3
      */
-    if (
-      redirectUrl &&
-      redirectUrl.startsWith('http://localhost:') &&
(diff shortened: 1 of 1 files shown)


scopsy (approved):

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -91,7 +91,7 @@ export class AuthController {
      * Make sure we only allow localhost redirects for CLI use and our own success route
      * https://github.com/novuhq/novu/security/code-scanning/3
      */
-    if (redirectUrl && redirectUrl.startsWith('http://localhost:')) {
+    if (redirectUrl && redirectUrl.startsWith('http://localhost:') && !redirectUrl.includes('@')) {
```
p-fernandez, line 94: Are we missing the check for the question mark? Based on the PR title.

davidsoderberg: Me and Dima discussed and said it was not needed after I had created the pr :)

scopsy: No, it was not needed in the end :) I see David updated the title

PR #3513, opened 2023-05-26
chore(web): templates store e2e tests

LetItRock:

Templates Store Cypress E2E tests.

Templates Store PRD.

Screenshot 2023-05-26 at 11 42 23


apps/api/src/.env.test
```diff
@@ -2,7 +2,7 @@ GOOGLE_OAUTH_CLIENT_ID=11
 GOOGLE_OAUTH_CLIENT_SECRET=11
 GOOGLE_OAUTH_REDIRECT=http://localhost:3000/v1/auth/google/callback
 STORE_ENCRYPTION_KEY="<ENCRYPTION_KEY_MUST_BE_32_LONG>"
-BLUEPRINT_CREATOR="<BLUEPRINT_CREATOR>"
+BLUEPRINT_CREATOR=645b648b36dd6d25f8650d37
```
LetItRock, line 5: This is the part of the trick... The `BLUEPRINT_CREATOR` is the `organizationId` of the org that has the blueprints (aka. notification-templates) on the prod env. In order for blueprints API to work in the testing environment I've set this hardcoded `organizationId` value for API and Cypress config as well. Now the API knows where it should find the blueprints. And in Cypress, I've prepared the plugin code that creates the org with that `id`, prod env, and blueprints, so the API can return something for tests.

apps/web/cypress.config.ts
```diff
@@ -43,6 +43,7 @@ export default defineConfig({
     apiUrl: 'http://localhost:1336',
     GITHUB_USER_EMAIL: '',
     GITHUB_USER_PASSWORD: '',
+    BLUEPRINT_CREATOR: '645b648b36dd6d25f8650d37',
```
LetItRock, line 46: same `id` used in the Cypress config

apps/web/cypress/global.d.ts
```diff
@@ -47,6 +47,8 @@ declare namespace Cypress {
 
     loginWithGitHub(): Chainable<any>;
 
+    makeBlueprints(): Chainable<any>;
```
LetItRock, line 50: the new command interface

apps/web/cypress/plugins/index.ts
```diff
@@ -138,5 +140,62 @@ module.exports = (on, config) => {
         templates,
       };
     },
+    async makeBlueprints() {
```
LetItRock, line 143: the plugin code that creates blueprints for the org

apps/web/cypress/plugins/index.ts
```diff
@@ -138,5 +140,62 @@ module.exports = (on, config) => {
...
+
+      const notificationGroupRepository = new NotificationGroupRepository();
+      const organizationService = new OrganizationService();
+      const environmentService = new EnvironmentService();
+
+      let organization = await organizationService.getOrganization(config.env.BLUEPRINT_CREATOR);
+      if (!organization) {
+        organization = await organizationService.createOrganization({ _id: config.env.BLUEPRINT_CREATOR });
+        await environmentService.createEnvironment(organization._id, 'Production');
+      }
```
LetItRock, line 158: I've noticed that for me locally the DB is not cleared after all tests (guess the same for you), so the organization with `BLUEPRINT_CREATOR` id stays there, that's why these checks are needed.

apps/web/cypress/plugins/index.ts
```diff
@@ -138,5 +140,62 @@ module.exports = (on, config) => {
...
+            isBlueprint: true,
+          }),
+          notificationTemplateService.createTemplate({
+            noFeedId: true,
+            noLayoutId: true,
+            name: ':fa-solid fa-lock: Password reset',
+            isBlueprint: true,
+          }),
+        ]);
+      }
```
LetItRock, line 196: create blueprints

apps/web/cypress/support/commands.ts
```diff
@@ -86,6 +86,10 @@ Cypress.Commands.add(
   }
 );
 
+Cypress.Commands.add('makeBlueprints', () => {
+  return cy.task('makeBlueprints');
+});
```
LetItRock, line 91: the new command

apps/web/cypress/tests/changes.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('Changes Screen', function () {
     switchEnvironment('Production');
     cy.location('pathname').should('equal', ROUTES.WORKFLOWS);
 
-    cy.getByTestId('create-template-btn').get('button').should('be.disabled');
+    cy.getByTestId('create-workflow-btn').get('button').should('be.disabled');
```
LetItRock, line 24: renamed

apps/web/cypress/tests/templates-store.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+      name = templateName.replace(regexResult[0], '').trim();
+      iconName = regexResult[0].replace(/:/g, '').trim();
+    }
+
+    return { name, iconName: iconName };
+  };
+
+  beforeEach(function () {
+    cy.initializeSession({ noTemplates: true }).as('session');
+    indexedDB.deleteDatabase('localforage');
```
LetItRock, line 16: initialize the session without the templates and clear the indexedDB (blueprints cache)

apps/web/cypress/tests/templates-store.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+      const secondPopular = popularBlueprints[0];
+      const thirdPopular = popularBlueprints[0];
+      const { name: firstName } = getTemplateDetails(firstPopular.name);
+      const { name: secondName } = getTemplateDetails(secondPopular.name);
+      const { name: thirdName } = getTemplateDetails(thirdPopular.name);
+
+      cy.getByTestId('create-template-dropdown-item').should('have.length', 3);
+      cy.getByTestId('create-template-dropdown-item').contains(firstName);
+      cy.getByTestId('create-template-dropdown-item').contains(secondName);
+      cy.getByTestId('create-template-dropdown-item').contains(thirdName);
```
LetItRock, line 98: get the blueprints and check if the they are shown under the `create workflow` dropdown

apps/web/cypress/plugins/index.ts
djabarovgeorge: I think it is not cleaning the DB on cases where the test run was forced to close. it is a good addon due to the hardcoded id. 🙃

apps/web/cypress/tests/templates-store.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+    cy.getByTestId('create-workflow-all-templates').should('be.visible');
+
+    cy.request({
+      method: 'GET',
+      url: `${Cypress.env('apiUrl')}/v1/blueprints/group-by-category`,
+    }).then(({ body }) => {
+      const { blueprints: popularBlueprints } = body.data.popular;
+      const firstPopular = popularBlueprints[0];
+      const secondPopular = popularBlueprints[0];
+      const thirdPopular = popularBlueprints[0];
```
djabarovgeorge, line 90: Probably a small typo on the index part.

apps/web/cypress/tests/templates-store.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+    cy.visit('/workflows');
+    cy.wait('@getTemplates');
+
+    cy.getByTestId('no-workflow-templates-placeholder').should('be.visible');
+    cy.getByTestId('create-workflow-tile').should('exist');
+    cy.getByTestId('all-workflow-tile').should('exist').should('be.disabled');
+
+    cy.wait('@getBlueprints');
+    cy.getByTestId('all-workflow-tile').should('exist').should('not.be.disabled');
+    cy.getByTestId('popular-workflow-tile').should('have.length', 3);
```
djabarovgeorge, line 43: This one is correct but at the moment we have only 2 popular templates, therefore I will update all the tests to check length of 2.

apps/web/cypress/plugins/index.ts
```diff
@@ -138,5 +140,62 @@ module.exports = (on, config) => {
...
+            name: ':fa-solid fa-star: Super cool workflow',
+            isBlueprint: true,
+          }),
+          notificationTemplateService.createTemplate({
+            noFeedId: true,
+            noLayoutId: true,
+            name: ':fa-solid fa-lock: Password reset',
+            isBlueprint: true,
+          }),
+        ]);
```
djabarovgeorge, line 195: In the current PR state, this part could cause flakiness where we create the second one before the first one, and the test could fail.
It will be resolved on the next commit to this branch. 

p.s the solution will be a side effect 😅
p.s.s the solution is that we will add here _id and it will be in the same order the API will have, therefore the list order will be always the same.

djabarovgeorge (approved):

PR #3514, opened 2023-05-26
feat: add new isDefined filter

### LINKED ISSUE #3421: [NV-1627] New "is defined" filter for payload checks
## Why?

We have **Equal** and **Not Equal** filters for payload, where equality check of variable value in payload will take place. **IsDefined** filter is slightly different, as it will check whether variable value is present in payload (in case this variable is not required)

## How?

Adding another filter type to check for "truthy" values in path

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-1627](https://linear.app/novu/issue/NV-1627/new-is-defined-filter-for-payload-checks)</sub>

jainpawan21:

Add isDefined filter

closes #3421


apps/worker/src/app/workflow/services/metric-queue.service.ts
```diff
@@ -124,11 +124,11 @@ export class MetricQueueService extends QueueService<Record<string, never>> {
...
-              Logger.log(`MetricQueueService/${queueService.name}/completed`, JSON.stringify(successMetric));
-              Logger.log(`MetricQueueService/${queueService.name}/failed`, JSON.stringify(failMetric));
-              Logger.log(`MetricQueueService/${queueService.name}/waiting`, waitCount);
-              Logger.log(`MetricQueueService/${queueService.name}/delayed`, delayedCount);
-              Logger.log(`MetricQueueService/${queueService.name}/active`, activeCount);
+              Logger.debug(`MetricQueueService/${queueService.name}/completed`, JSON.stringify(successMetric));
+              Logger.debug(`MetricQueueService/${queueService.name}/failed`, JSON.stringify(failMetric));
+              Logger.debug(`MetricQueueService/${queueService.name}/waiting`, waitCount);
+              Logger.debug(`MetricQueueService/${queueService.name}/delayed`, delayedCount);
+              Logger.debug(`MetricQueueService/${queueService.name}/active`, activeCount);
```
BiswaViraj, line 131: Do we need this change?

jainpawan21: Yes, because metric queue is logging too much logs

Cliftonz: I agree

apps/web/src/pages/templates/filter/FilterModal.tsx
```diff
@@ -312,22 +317,25 @@ function EqualityForm({
...
+            name={`steps.${stepIndex}.filters.0.children.${index}.value`}
+            defaultValue=""
+            render={({ field, fieldState }) => {
+              return (
+                <Input
+                  {...field}
+                  error={fieldState.error?.message}
+                  placeholder="Value"
+                  data-test-id="filter-value-input"
+                />
```
LetItRock, line 334: what should be provided here? because I don't see anywhere the logic that will use that value

BiswaViraj, line 334: @jainpawan21 Like we discussed previously, how is the value field is being handled if it contained a different value before selecting the IS_DEFINED operator.

apps/web/src/pages/templates/filter/FilterModal.tsx
```diff
@@ -304,30 +314,38 @@ function EqualityForm({
                   { value: 'SMALLER_EQUAL', label: 'Smaller or equal' },
                   { value: 'IN', label: 'Contains' },
                   { value: 'NOT_IN', label: 'Not contains' },
+                  { value: 'IS_DEFINED', label: 'isDefined' },
```
scopsy, line 317:
```suggestion
                  { value: 'IS_DEFINED', label: 'Is Defined' },
```

Is this the value that is displayed to the user? Better to not do a camelCase :)

apps/web/src/pages/templates/filter/FilterModal.tsx
```diff
@@ -304,30 +314,38 @@ function EqualityForm({
...
                 ]}
                 {...field}
                 data-test-id="filter-operator-dropdown"
+                onChange={(value) => {
+                  field.onChange(value);
+                  value === 'IS_DEFINED' && setValue(`steps.${stepIndex}.filters.0.children.${index}.value`, '');
```
scopsy, line 323:
```suggestion
              if (value === 'IS_DEFINED') {
             
                 setValue(`steps.${stepIndex}.filters.0.children.${index}.value`, '');
                 }
                 
```

Suggest avoiding this pattern and just using an if statement for better readability

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -406,7 +406,9 @@ export class MessageMatcher {
     if (fieldFilter.operator === 'IN') {
       result = actualValue.includes(filterValue);
     }
-
+    if (fieldFilter.operator === 'IS_DEFINED') {
+      result = actualValue !== undefined;
+    }
```
scopsy, line 411: Could we add a test case for this in one of the existing filters test on the API ?

apps/web/src/pages/templates/filter/FilterModal.tsx
jainpawan21: Added code to make it ""

jainpawan21: Added code to make it "" when filter is IS_DEFINED

jainpawan21:
@scopsy

scopsy (approved):

PR #3516, opened 2023-05-26
fix: add delete provider credentials endpoint

jainpawan21:

1. Add endpoint to delete subscriber provider credentials
2. Fix few things in node sdk

To remove provider credentials entirely.
I have few integrations like 3 discord integrations in this screenshot. But there is no any way to delete these
Screenshot 2023-05-26 at 6 47 25 PM


packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -66,10 +67,16 @@ export class Subscribers extends WithHttp implements ISubscribers {
     });
   }
 
+  async deleteCredentials(subscriberId: string, providerId: string) {
+    return await this.http.delete(
+      `/subscribers/${subscriberId}/credentials/${providerId}`
+    );
+  }
+
   async unsetCredentials(subscriberId: string, providerId: string) {
```
jainpawan21, line 76: Shall we remove this `unsetCredentials` method after this new `deleteCredentials` method?

LetItRock: maybe its better to deprecate

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await updateSubscriberChannelUsecase.execute(
+      UpdateSubscriberChannelCommand.create({
+        organizationId: subscriber._organizationId,
+        subscriberId: subscriber.subscriberId,
+        environmentId: session.environment._id,
+        providerId: subscriberChannel.providerId,
+        credentials: subscriberChannel.credentials,
+        oauthHandler: OAuthHandlerEnum.NOVU,
+      })
+    );
```
LetItRock, line 48: I would say that we have to add any other additional channel credentials, and at the end of the test check if it exists... like this way we will be sure that the others wouldn't be deleted

apps/api/src/app/subscribers/subscribers.controller.ts
```diff
@@ -223,6 +228,29 @@ export class SubscribersController {
...
   }
 
+  @Delete('/:subscriberId/credentials/:providerId')
+  @ExternalApiAccessible()
+  @UseGuards(JwtAuthGuard)
+  @ApiResponse(SubscriberResponseDto)
+  @ApiOperation({
+    summary: 'Delete subscriber credentials by providerId',
+    description: 'Delete subscriber credentials such as slack and expo tokens.',
+  })
```
LetItRock, line 238: usually delete endpoints just return the 200 or 204
like: `@HttpCode(HttpStatus.OK) or @HttpCode(HttpStatus.NO_CONTENT)`

p-fernandez: I personally push to always return the 204 and empty body as you should not need anything else besides the confirmation the operation was right.

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+import { expect } from 'chai';
+import { SharedModule } from '../../../shared/shared.module';
+import { ChatProviderIdEnum } from '@novu/shared';
+import { DeleteSubscriberCredentials } from './delete-subscriber-credentials.usecase';
+import { DeleteSubscriberCredentialsCommand } from './delete-subscriber-credentials.command';
+import { UpdateSubscriberChannel } from '../update-subscriber-channel/update-subscriber-channel.usecase';
+import { UpdateSubscriberChannelCommand } from '../update-subscriber-channel/update-subscriber-channel.command';
+import { OAuthHandlerEnum } from '../../types';
+
+describe('Update subscriber provider credentials', function () {
```
p-fernandez, line 13:
```suggestion
describe('Delete subscriber provider credentials', function () {
```

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.spec.ts
p-fernandez: I do agree. A second test for that would be very helpful.

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.usecase.ts
```diff
@@ -0,0 +1,68 @@
...
+    this.analyticsService.track('Delete Subscriber Credentials - [Subscribers]', command.organizationId, {
+      providerId: command.providerId,
+      _organization: command.organizationId,
+      _subscriberId: foundSubscriber._id,
+    });
+
+    return (await this.subscriberRepository.findBySubscriberId(
+      command.environmentId,
+      command.subscriberId
+    )) as SubscriberEntity;
```
p-fernandez, line 49: I do think this is not needed for the reasons stated in the status code to return in the controller.

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.usecase.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  async execute(command: DeleteSubscriberCredentialsCommand) {
+    const foundSubscriber = await this.subscriberRepository.findBySubscriberId(
+      command.environmentId,
+      command.subscriberId
+    );
+
+    if (!foundSubscriber) {
+      throw new ApiException(`SubscriberId: ${command.subscriberId} not found`);
+    }
```
p-fernandez, line 25: If we replace this bit of the code with the GetSubscriber use case we benefit of 2 things:
- not duplicating code and in case of modification to be reused here
- take advantage of the read cache of the subscriber

libs/shared/src/dto/subscriber/subscriber.dto.ts
```diff
@@ -36,7 +36,6 @@ export interface ISubscriberPayload {
   avatar?: string;
   locale?: string;
   data?: SubscriberCustomData;
-  [key: string]: string | string[] | boolean | number | SubscriberCustomData | undefined;
```
p-fernandez, line 39: Is this safe to do?

packages/node/src/lib/subscribers/subscriber.interface.ts
```diff
@@ -8,7 +8,7 @@ import {
 export { ISubscriberPayload };
 
 export interface ISubscribers {
-  list(page: number);
+  list(page: number, limit: number);
```
p-fernandez, line 11: I think this change doesn't belong to this PR. Does it come from an unmerged one?

libs/shared/src/dto/subscriber/subscriber.dto.ts
jainpawan21: Yes, we support custom property in the data field only.

`data?: SubscriberCustomData`

packages/node/src/lib/subscribers/subscriber.interface.ts
jainpawan21: Yes, I added this in subscriber doc PR but removed it later because it requires a new version to be released for `@node/novu` 
I will update the subscriber doc also after this PR merge

LetItRock:
hey @jainpawan21, let's make here a tiny change and return the 204 on that endpoint as Pablo and I suggested, ok? sorry for the hassle :) then I'll approve

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.usecase.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  async execute(command: DeleteSubscriberCredentialsCommand) {
+    const foundSubscriber = await this.getSubscriberUseCase.execute(
+      GetSubscriberCommand.create({
+        ...command,
+      })
+    );
+
+    if (!foundSubscriber) {
+      throw new ApiException(`SubscriberId: ${command.subscriberId} not found`);
```
djabarovgeorge, line 27:
```suggestion
      throw new NotFoundException(`SubscriberId: ${command.subscriberId} not found`);
```

djabarovgeorge (commented):
Left some small notes, i am wondering is your screenshot from the PR description is from one subscriber entity?

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.usecase.ts
```diff
@@ -0,0 +1,76 @@
...
+import { GetSubscriberCommand, GetSubscriber } from '../get-subscriber';
+
+@Injectable()
+export class DeleteSubscriberCredentials {
+  constructor(
+    private invalidateCache: InvalidateCacheService,
+    private subscriberRepository: SubscriberRepository,
+    private integrationRepository: IntegrationRepository,
+    private analyticsService: AnalyticsService,
+    private getSubscriberUseCase: GetSubscriber
```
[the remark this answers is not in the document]
p-fernandez: Outside of the scope of this PR: I discussed this with Dima time ago. I think the english name is `use case` and not `usecase` therefore the capitalisation should be `UseCase` and the hyphenation `use-case`. Pawan probably followed bits of my code where I did that. But we have mixed grammar all over the place. Dima said to leave it for a tech debt task after we reach a consensus.

apps/api/src/app/subscribers/usecases/delete-subscriber-credentials/delete-subscriber-credentials.usecase.ts
```diff
@@ -0,0 +1,65 @@
...
+  async execute(command: DeleteSubscriberCredentialsCommand): Promise<void> {
+    const foundSubscriber = await this.getSubscriberUseCase.execute(
+      GetSubscriberCommand.create({
+        ...command,
+      })
+    );
+
+    if (!foundSubscriber) {
+      throw new NotFoundException(`SubscriberId: ${command.subscriberId} not found`);
+    }
```
p-fernandez, line 27: This check is not needed as `GetSubscriberUseCase` already throws a `NotFoundException` when this case happens.
https://github.com/novuhq/novu/blob/4b71ba025c44a422f6640da3a371a5fdb6470e2d/apps/api/src/app/subscribers/usecases/get-subscriber/get-subscriber.usecase.ts#L14

p-fernandez (approved):
Just a quick thing to change but all good! 🌟

Issue #3431, opened 2023-05-16
[NV-2342] 🐛 Bug Report: Details Body on Unexpected Provider Error is not shown on Push Notifications


christos-deretzis:
### 📜 Description

Whenever we trigger a push notification and something goes wrong (e.g. a wrong device token or wrong credentials for push providers), we get an Unexpected Provider Error on the execution details.As a result, we can't see the body for this specific step whenever we press the button. Instead, we get an empty body ({})

### 👟 Reproduction steps

1. Login to Novu web and Navigate to the **Activity Feed** from the Left Menu
2. Click on a failed Push Notification 
3. On the last step (**Unexpected Provider Error**) click on the button next to it (**Show Details**)
4. The details body is empty ({})

### 👍 Expected behavior

It should have a non empty body with some kind of message, explaining what went wrong 

### 👎 Actual Behavior with Screenshots

![Screenshot 2023-05-16 at 4 22 56 PM](https://github.com/novuhq/novu/assets/124055228/5a031e13-fb5b-44fd-a63c-f15824699bfe)

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-2342](https://linear.app/novu/issue/NV-2342/🐛-bug-report-details-body-on-unexpected-provider-error-is-not-shown)</sub>

jainpawan21:
Sure @gitstart Assigning this to you!

gitstart:
Hi @christos-deretzis, I am currently facing a blocker while trying to reproduce this issue. I am having this blocker when trying to run a Push notification trigger. I have activated OneSignal and FCM separately but the error is persistent.
I will appreciate any form of help to unblock this issue. Cc @scopsy @jainpawan21

![Screenshot 2023-05-25 at 10 41 54](https://github.com/novuhq/novu/assets/1501599/698c06d6-345a-4522-bffe-0130ba7f4868)

Loom Video: https://www.loom.com/share/6f8654cc7fa544bcb6c9a49e392db4f6

christos-deretzis:
@gitstart what you have to do it to subscribe the device token that is generated by the push provider to the subscriber id, so that the subscriber id will have a device token for a given provider. The request to achieve that is:

[...]

PR #3519, opened 2023-05-26
fix: throw error in push only when success count is zero

jainpawan21:

fix: empty body in unexpected provider error in push channel for fcm

closes: #3431

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
@@ -269,7 +269,7 @@ export class SendMessagePush extends SendMessageBase {
           status: ExecutionDetailsStatusEnum.FAILED,
           isTest: false,
           isRetry: false,
-          raw: JSON.stringify(e),
+          raw: JSON.stringify(e) !== JSON.stringify({}) ? JSON.stringify(e) : JSON.stringify(e.message),
         })
       );
 
--- a/providers/fcm/src/lib/fcm.provider.ts
+++ b/providers/fcm/src/lib/fcm.provider.ts
@@ -89,7 +89,7 @@ export class FcmPushProvider implements IPushProvider {
       });
     }
 
-    if (res.failureCount > 0) {
+    if (res.successCount === 0) {
       throw new Error(
         `Sending message failed due to "${
           res.responses.find((i) => i.success === false).error.message
@@ -101,7 +101,9 @@ export class FcmPushProvider implements IPushProvider {
     await deleteApp(app);
 
     return {
-      ids: res?.responses?.map((response) => response.messageId),
+      ids: res?.responses?.map((response) =>
+        response.success ? response.messageId : response.error.message
+      ),
       date: new Date().toISOString(),
     };
   }


apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -269,7 +269,7 @@ export class SendMessagePush extends SendMessageBase {
           status: ExecutionDetailsStatusEnum.FAILED,
           isTest: false,
           isRetry: false,
-          raw: JSON.stringify(e),
+          raw: JSON.stringify(e) !== JSON.stringify({}) ? JSON.stringify(e) : JSON.stringify(e.message),
```
djabarovgeorge, line 272: did you or the issue reported reproduce this error? because I am not sure this will work 🤔

i wonder if this is the expected response
https://firebase.google.com/docs/reference/fcm/rest/v1/FcmError

providers/fcm/src/lib/fcm.provider.ts
```diff
@@ -101,7 +101,9 @@ export class FcmPushProvider implements IPushProvider {
     await deleteApp(app);
 
     return {
-      ids: res?.responses?.map((response) => response.messageId),
+      ids: res?.responses?.map((response) =>
+        response.success ? response.messageId : response.error.message
```
p-fernandez, line 105: Does `response.error.message` return an id? How would we know to which id belongs this error?

providers/fcm/src/lib/fcm.provider.ts
```diff
@@ -101,7 +101,9 @@ export class FcmPushProvider implements IPushProvider {
     await deleteApp(app);
 
     return {
-      ids: res?.responses?.map((response) => response.messageId),
+      ids: res?.responses?.map((response) =>
```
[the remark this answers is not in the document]
jainpawan21: Yes, messageId is missing in case of failure

providers/fcm/src/lib/fcm.provider.ts
p-fernandez: That was part of my point. Think of the scenario of 3 token, 2 of them are invalid. Does the error message provides the token identifier that error is related to or is just a generic error message? If it is the second case makes things more difficult to user because would need to debug in trial and error mode. Let me know if I didn't explain my worry clearly.

jainpawan21: In case of 3 invalid tokens -> ["invalid_token", "valid_token", "invalid_token"]
 
 execution log will be -> "ids": ["the registration token is not a valid FCM registration token", "projects/project-name/mesages/{messageId}", "the registration token is not a valid FCM registration token"]
 
 log will be in sync with token state
 Shall I add invalid token also in log?

p-fernandez: That's what I was thinking but was wondering if there was any way to get something like:
```
const tokens = ["invalid_token_1", "valid_token", "invalid_token_2"];

...

{
...
  "ids": ["the registration token invalid_token_1 is not a valid FCM registration token", "projects/project-name/mesages/{messageId}", "the registration token invalid_token_2 is not a valid FCM registration token"]
 ...
 }
 ```
Just to make life easier to the user. Not a big deal to not approve the PR on my side but if we could somehow show to the user which ones are invalid would be top.

p-fernandez (approved):
🌟

PR #3523, opened 2023-05-29
feat: add remove all messages endpoint in widget

jainpawan21:

### DIFF
--- a/apps/api/src/app/widgets/usecases/remove-messages/remove-messages.usecase.ts
+++ b/apps/api/src/app/widgets/usecases/remove-messages/remove-messages.usecase.ts
@@ -53,7 +53,7 @@ export class RemoveMessages {
 
     let deletedMessages;
     let feed;
-    if (command.feedId !== '' || command.feedId !== null || command.feedId !== undefined) {
+    if (command.feedId !== '' && command.feedId !== null && command.feedId !== undefined) {
       feed = await this.feedRepository.findById(command.feedId);
       if (!feed) {
         throw new NotFoundException(`Feed with ${command.feedId} not found`);
@@ -64,6 +64,7 @@ export class RemoveMessages {
         _environmentId: command.environmentId,
         _organizationId: command.organizationId,
         _subscriberId: command.subscriberId,
+        channel: ChannelTypeEnum.IN_APP,
       };
 
       if (feed) {
--- a/apps/api/src/app/widgets/usecases/remove-messages/remove-messages.usecase.ts
+++ b/apps/api/src/app/widgets/usecases/remove-messages/remove-messages.usecase.ts
@@ -1,0 +1,83 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+import {
+  MessageEntity,
+  DalException,
+  MessageRepository,
+  SubscriberRepository,
+  SubscriberEntity,
+  MemberRepository,
+  FeedRepository,
+} from '@novu/dal';
+import { ChannelTypeEnum } from '@novu/shared';
+import {
+  WsQueueService,
+  AnalyticsService,
+  InvalidateCacheService,
+  buildFeedKey,
+  buildMessageCountKey,
+} from '@novu/application-generic';
+
+import { RemoveMessagesCommand } from './remove-messages.command';
+import { ApiException } from '../../../shared/exceptions/api.exception';
+import { MarkEnum } from '../mark-message-as/mark-message-as.command';
+
+@Injectable()
+export class RemoveMessages {
+  constructor(
+    private invalidateCache: InvalidateCacheService,
+    private messageRepository: MessageRepository,
+    private wsQueueService: WsQueueService,
+    private analyticsService: AnalyticsService,
+    private subscriberRepository: SubscriberRepository,
+    private memberRepository: MemberRepository,
+    private feedRepository: FeedRepository
+  ) {}
+
+  async execute(command: RemoveMessagesCommand): Promise<MessageEntity[]> {
+    await this.invalidateCache.invalidateQuery({
+      key: buildFeedKey().invalidate({
+        subscriberId: command.subscriberId,
+        _environmentId: command.environmentId,
+      }),
+    });
+
+    await this.invalidateCache.invalidateQuery({
+      key: buildMessageCountKey().invalidate({
+        subscriberId: command.subscriberId,
+        _environmentId: command.environmentId,
+      }),
+    });
+
+    const subscriber = await this.subscriberRepository.findBySubscriberId(command.environmentId, command.subscriberId);
+    if (!subscriber) throw new NotFoundException(`Subscriber ${command.subscriberId} not found`);
+
+    let deletedMessages;
+    let feed;
+    if (command.feedId !== '' || command.feedId !== null || command.feedId !== undefined) {
+      feed = await this.feedRepository.findById(command.feedId);
+      if (!feed) {
+        throw new NotFoundException(`Feed with ${command.feedId} not found`);
+      }
+    }
+    try {
+      const deleteMessageQuery: Partial<MessageEntity> = {
+        _environmentId: command.environmentId,
+        _organizationId: command.organizationId,
+        _subscriberId: command.subscriberId,
+      };
+
+      if (feed) {
+        deleteMessageQuery._feedId = feed._id;
+      }
+
+      deletedMessages = await this.messageRepository.deleteMany(deleteMessageQuery);
+    } catch (e) {
+      if (e instanceof DalException) {
+        throw new ApiException(e.message);
+      }
+      throw e;
+    }
+
+    return deletedMessages;
+  }
+}
--- a/apps/api/src/app/widgets/widgets.controller.ts
+++ b/apps/api/src/app/widgets/widgets.controller.ts
@@ -50,6 +50,8 @@ import { MarkAllMessagesAsCommand } from './usecases/mark-all-messages-as/mark-a
 import { MarkAllMessagesAs } from './usecases/mark-all-messages-as/mark-all-messages-as.usecase';
 import { GetNotificationsFeedDto } from './dtos/get-notifications-feed-request.dto';
 import { LimitPipe } from './pipes/limit-pipe/limit-pipe';
+import { RemoveMessagesCommand } from './usecases/remove-messages/remove-messages.command';
+import { RemoveMessages } from './usecases/remove-messages/remove-messages.usecase';
 
 @Controller('/widgets')
 @ApiExcludeController()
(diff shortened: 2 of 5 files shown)


apps/api/src/app/widgets/usecases/remove-messages/remove-messages.usecase.ts
```diff
@@ -0,0 +1,115 @@
...
+      const admin = await this.memberRepository.getOrganizationAdminAccount(command.organizationId);
+      if (admin) {
+        this.analyticsService.track(`Removed Message - [Notification Center]`, admin._userId, {
+          _subscriber: subscriber._id,
+          _organization: command.organizationId,
+          _environment: command.environmentId,
+        });
+      }
+    } catch (e) {
+      if (e instanceof DalException) {
```
ainouzgali, line 89: To have a DalException here, you need to throw a DalException in the repository if delete did not succeed

ainouzgali (commented):
Great work @jainpawan21 🤩
Could we also add a few test cases for this?

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,137 @@
...
+        _environmentId: command.environmentId,
+      }),
+    });
+
+    const subscriber = await this.subscriberRepository.findBySubscriberId(command.environmentId, command.subscriberId);
+    if (!subscriber) throw new NotFoundException(`Subscriber ${command.subscriberId} not found`);
+
+    let deletedMessages;
+    let feed;
+    if (command.feedId !== '' && command.feedId !== null && command.feedId !== undefined) {
```
LetItRock, line 56: you can have the ` @IsMongoId()` annotation on the command, then you will do:
```suggestion
    if (command.feedId) {
```

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,137 @@
...
+      }),
+    });
+
+    await this.invalidateCache.invalidateQuery({
+      key: buildMessageCountKey().invalidate({
+        subscriberId: command.subscriberId,
+        _environmentId: command.environmentId,
+      }),
+    });
+
```
LetItRock, line 50: cache invalidation should be the last step after all the successful operations

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,137 @@
...
+    const admin = await this.memberRepository.getOrganizationAdminAccount(command.organizationId);
+    const count = await this.messageRepository.getCount(
+      command.environmentId,
+      subscriber._id,
+      ChannelTypeEnum.IN_APP,
+      {
+        [marked]: false,
+      },
+      { limit: 1000 }
+    );
```
LetItRock, line 107: the count will be always here zero if there was no feedId provided, so in that case, we don't need to do this operation...

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,137 @@
...
+      const deleteMessageQuery: Partial<MessageEntity> = {
+        _environmentId: command.environmentId,
+        _organizationId: command.organizationId,
+        _subscriberId: subscriber._id,
+        channel: ChannelTypeEnum.IN_APP,
+      };
+
+      if (feed) {
+        deleteMessageQuery._feedId = feed._id;
+      }
```
LetItRock, line 72:
```suggestion
      const deleteMessageQuery: Partial<MessageEntity> = {
        _environmentId: command.environmentId,
        _organizationId: command.organizationId,
        _subscriberId: subscriber._id,
        channel: ChannelTypeEnum.IN_APP,
        _feedId: feed._id,
      };
```
should also work

apps/api/src/app/widgets/widgets.controller.ts
```diff
@@ -228,6 +231,25 @@ export class WidgetsController {
...
 
+  @ApiOperation({
+    summary: `Remove a subscriber's feed messages`,
+  })
+  @UseGuards(AuthGuard('subscriberJwt'))
+  @Delete('/messages')
+  async removeAllMessages(
+    @SubscriberSession() subscriberSession: SubscriberEntity,
+    @Query('feedId') feedId?: string
+  ): Promise<number> {
```
LetItRock, line 242: do we need to return how many were deleted?

jainpawan21: no not necessary, Shall I return boolean?

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
jainpawan21: user can delete messages without feedId also

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
jainpawan21: shall I write if(feedId) condition where I am calling this function?

djabarovgeorge: > the count will be always here zero if there was no feedId provided, so in that case, we don't need to do this operation...

why? if the user won't provide the feedId won't the query delete all the available messages?

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,137 @@
...
+      { limit: 1000 }
+    );
+
+    this.updateSocketCount(subscriber, count, marked);
+
+    if (admin) {
+      this.analyticsService.track(`Removed All Messages - [Notification Center]`, admin._userId, {
+        _subscriber: subscriber._id,
+        _organization: command.organizationId,
+        _environment: command.environmentId,
```
djabarovgeorge, line 115: Same as above maybe we can add the feed id as well

libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -245,6 +245,10 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
     await this.message.delete({ _id: message._id, _environmentId: message._environmentId });
   }
 
+  async deleteMany(query: MessageQuery) {
+    return await this.message.delete({ ...query, deleted: false });
```
djabarovgeorge, line 249: not sure if `deleted: false` is needed here.

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
LetItRock: `_feedId: feed?._id,` should work

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
LetItRock: > shall I write if(feedId) condition where I am calling this function?

I think that you can pass the feed here and check if it exists and if yes then get the count. And by default set the count to zero, because we should remove all the messages for that case, meaning count will be always zero.

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
jainpawan21: in that case _feedId will be undefined and incorrect values will be deleted

libs/dal/src/repositories/message/message.repository.ts
jainpawan21: It will delete only those values which are not yet deleted

jainpawan21:
@djabarovgeorge @LetItRock @ainouzgali 
Please rereview this

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      await this.messageRepository.deleteMany(deleteMessageQuery);
+
+      await this.updateServices(command, subscriber, MarkEnum.SEEN);
+      await this.updateServices(command, subscriber, MarkEnum.READ);
+
+      const admin = await this.memberRepository.getOrganizationAdminAccount(command.organizationId);
+      if (admin) {
+        this.analyticsService.track(`Removed Message - [Notification Center]`, admin._userId, {
```
ainouzgali, line 68:
```suggestion
        this.analyticsService.track(`Removed All Feed Messages - [Notification Center]`, admin._userId, {
```

apps/api/src/app/widgets/e2e/remove-all-messages.e2e.ts
```diff
@@ -0,0 +1,71 @@
...
+      })
+      .expect(201);
+
+    const { token, profile } = body.data;
+
+    subscriberToken = token;
+    subscriberProfile = profile;
+  });
+
+  it('should remove all messages', async function () {
```
ainouzgali, line 43: What do you think about also adding a test case for the delete of a specific feed?

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
```diff
@@ -0,0 +1,130 @@
...
+      });
+    } catch (e) {
+      if (e instanceof DalException) {
+        throw new ApiException(e.message);
+      }
+      throw e;
+    }
+  }
+
+  private async updateServices(command: RemoveAllMessagesCommand, subscriber, marked: string, feed?: FeedEntity) {
```
ainouzgali, line 97: Where are you passing the feed param? I don't see where you use it with the feed

ainouzgali (commented):
@jainpawan21 just a few more questions :)

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
jainpawan21: Ohhh! I forgot to pass the feed in line no. 63 and 64

apps/api/src/app/widgets/e2e/remove-all-messages.e2e.ts
jainpawan21: Yeah. I agree. It will increase coverage. I will add

jainpawan21: Added

apps/api/src/app/widgets/usecases/remove-messages/remove-all-messages.usecase.ts
LetItRock: from what I remember it will skip that `_feedId` in the query if it's undefined... but better to verify

LetItRock (approved):
@jainpawan21 looks good to me!🥇

PR #3524, opened 2023-05-29
Nv 2215 delete messagetemplates of deleted templates

BiswaViraj:

- Delete the message templates and add them to the changes upon deletion of the Notification template

### DIFF
--- a/apps/api/src/app/message-template/usecases/delete-message-template/delete-message-template.usecase.ts
+++ b/apps/api/src/app/message-template/usecases/delete-message-template/delete-message-template.usecase.ts
@@ -1,0 +1,45 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { DalException, MessageTemplateRepository } from '@novu/dal';
+import { ChangeEntityTypeEnum } from '@novu/shared';
+
+import { CreateChange, CreateChangeCommand } from '../../../change/usecases';
+import { ApiException } from '../../../shared/exceptions/api.exception';
+import { DeleteMessageTemplateCommand } from './delete-message-template.command';
+
+@Injectable()
+export class DeleteMessageTemplate {
+  constructor(private messageTemplateRepository: MessageTemplateRepository, private createChange: CreateChange) {}
+
+  async execute(command: DeleteMessageTemplateCommand): Promise<boolean> {
+    try {
+      const template = await this.messageTemplateRepository.findById(command.messageTemplateId, command.environmentId);
+      if (!template) {
+        throw new NotFoundException(`Message Template with id ${command.messageTemplateId} not found`);
+      }
+
+      await this.messageTemplateRepository.delete({
+        _environmentId: command.environmentId,
+        _id: command.messageTemplateId,
+      });
+
+      await this.createChange.execute(
+        CreateChangeCommand.create({
+          organizationId: command.organizationId,
+          environmentId: command.environmentId,
+          userId: command.userId,
+          item: template,
+          type: ChangeEntityTypeEnum.MESSAGE_TEMPLATE,
+          parentChangeId: command.parentChangeId,
+          changeId: MessageTemplateRepository.createObjectId(),
+        })
+      );
+
+      return true;
+    } catch (error) {
+      if (error instanceof DalException) {
+        throw new ApiException(error.message);
+      }
+      throw error;
+    }
+  }
+}
--- a/apps/api/src/app/message-template/usecases/delete-message-template/delete-message-template.usecase.ts
+++ b/apps/api/src/app/message-template/usecases/delete-message-template/delete-message-template.usecase.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@nestjs/common';
-import { ChangeRepository, DalException, MessageTemplateRepository } from '@novu/dal';
+import { ChangeRepository, DalException, MessageTemplateEntity, MessageTemplateRepository } from '@novu/dal';
 import { ChangeEntityTypeEnum } from '@novu/shared';
 
 import { CreateChange, CreateChangeCommand } from '../../../change/usecases';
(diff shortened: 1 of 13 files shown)


apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
```diff
@@ -192,7 +192,10 @@ const TemplateEditorFormProvider = ({ children }) => {
           },
         });
         setTrigger(response.triggers[0]);
-        reset(form);
+        reset({
+          ...form,
+          steps: response.steps,
+        });
```
BiswaViraj, line 198: This fixes the issue where multiple duplicate message templates were being created while updating the Notification template from the UI

Steps to reproduce(without this code change):
- Create a new workflow
- add an in-app step, fill the details
-  within the in-app editor screen, click update button
- go back, and add another in-app step
- keep the second in-app editor screen open, click update
- do some minor changes in the content, click update
- Check DB, there will be multiple copies of the 2nd step.

@LetItRock You added this for some digest related issue. can you please confirm if this change works for that issue?

apps/api/src/app/notification-template/e2e/delete-notification-template.e2e.ts
```diff
@@ -139,6 +145,47 @@ describe('Delete notification template by id - /notification-templates/:template
 
     expect(response.body.message).to.contains('Could not find workflow with id');
   });
+
+  it('should delete the notification template along with the message templates', async function () {
```
ainouzgali, line 149: What do you think about also adding a test of deleting a notification template that has a feed assigned to it and then see that the feed can be deleted?

apps/api/src/app/message-template/usecases/delete-message-template/delete-message-template.usecase.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      await this.createChange.execute(
+        CreateChangeCommand.create({
+          organizationId: command.organizationId,
+          environmentId: command.environmentId,
+          userId: command.userId,
+          item: template,
+          type: ChangeEntityTypeEnum.MESSAGE_TEMPLATE,
+          parentChangeId: command.parentChangeId,
+          changeId: MessageTemplateRepository.createObjectId(),
```
ainouzgali, line 33: I believe here changeId should get the value from `changeRepository.getChangeId` because then it will aggregate all current changes made to that message template into one.

ainouzgali (commented):
Great job @BiswaViraj🤩
I just have a few concerns: 
1. Let's do a soft delete of the message template like we do for the other repositories. 
2. In promoting the notification template change it calls the [promote message template](https://github.com/novuhq/novu/blob/f50ac13c95d696281a8a6df8ec28b8d8329b12f4/apps/api/src/app/change/usecases/promote-message-template-change/promote-message-template-change.ts), which currently has no delete behavior. 
3. When [removing a step](https://github.com/novuhq/novu/blob/78a4901c7399703aca346d61f97cefce61924df4/apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx#L221) - not in the delete template - what do you think of also deleting the message template? I believe we are just updating the notification template now. 
4. Lets add a couple of tests for the promotion of notification template promotes deletion of the message templates + able to delete a feed.

apps/api/src/app/notification-template/e2e/delete-notification-template.e2e.ts
BiswaViraj: Agree, that will be a good addition, will add it.

apps/api/src/app/message-template/usecases/delete-message-template/delete-message-template.usecase.ts
```diff
@@ -0,0 +1,45 @@
...
+@Injectable()
+export class DeleteMessageTemplate {
+  constructor(private messageTemplateRepository: MessageTemplateRepository, private createChange: CreateChange) {}
+
+  async execute(command: DeleteMessageTemplateCommand): Promise<boolean> {
+    try {
+      const template = await this.messageTemplateRepository.findById(command.messageTemplateId, command.environmentId);
+      if (!template) {
+        throw new NotFoundException(`Message Template with id ${command.messageTemplateId} not found`);
+      }
```
p-fernandez, line 18: As a tech debt we should create an use case to find the message template to handle this logic as it can be reused in `UpdateMessageTemplate`.

apps/api/src/app/notification-template/e2e/delete-notification-template.e2e.ts
```diff
@@ -139,6 +145,47 @@ describe('Delete notification template by id - /notification-templates/:template
...
+    const deletedNotificationTemplate = await notificationTemplateRepository.findOne({
+      _environmentId: session.environment._id,
+      _id: template._id,
+    });
+
+    expect(deletedNotificationTemplate).to.equal(null);
+
+    const deletedIntegration = (
+      await notificationTemplateRepository.findDeleted({ _environmentId: session.environment._id, _id: template._id })
+    )[0];
```
p-fernandez, line 177:
```suggestion
    const [deletedIntegration] = await notificationTemplateRepository.findDeleted({ _environmentId: session.environment._id, _id: template._id });
```
Same vibes. 🙂 
Not to address it as it is stylistic and we have no defined preference for any.

apps/api/src/app/notification-template/usecases/delete-notification-template/delete-notification-template.usecase.ts
```diff
@@ -65,6 +68,18 @@ export class DeleteNotificationTemplate {
...
       );
+
+      for (const step of item.steps) {
+        await this.deleteMessageTemplate.execute(
+          DeleteMessageTemplateCommand.create({
+            organizationId: command.organizationId,
+            environmentId: command.environmentId,
+            userId: command.userId,
+            messageTemplateId: step._templateId,
+            parentChangeId: parentChangeId,
```
p-fernandez, line 79: Do we want to stop the execution of the notification template deletion if any of the message template deletions fail?

p-fernandez (approved):
🌟

PR #3525, opened 2023-05-29
feat: remove deprecated endpoints

davidsoderberg:
Removes deprecated endpoints

Around 7-8 months have passed since we deprecated these endpoints so should be safe to remove them now.


ainouzgali:
@davidsoderberg should also be removed from the [client package](https://github.com/novuhq/novu/blob/a0dbc5251036c13fb5599bbef186085ec700b8ee/packages/client/src/api/api.service.ts#L57). :)

p-fernandez (approved):
🌟

PR #3530, opened 2023-05-29
feat(web): launch darkly load the sdk in async

LetItRock:

Load the Launch Darkly SDK async and then render the application. Otherwise, if we do render the app first and then load the SDK then the feature flags are not initialized at the beginning and do fallback to the fallback values, which creates the blink on the UI when the flag fallback is `false` and the remote fetched flag value is `true`.


djabarovgeorge (approved):
I guess that withLDProvider and wrapping with LDProvider are equivalent?

LetItRock:
> I guess that withLDProvider and wrapping with LDProvider are equivalent?

yes it provider the same provider component

PR #3537, opened 2023-05-30
fix: `_feedId` bug in `novu/node` sdk

### LINKED ISSUE #3505: 🐛 Bug Report: Notification template _feedId is never set when given using SDK
### 📜 Description

If the developer passes `_feedId` field populated, the response comes with `_feedId: null` which is incorrect 

### 👟 Reproduction steps

1. Install the Nodejs SDK and create/update a notification template
2. Below there's a payload example

```json
{
  "active": true,
  "draft": false,
  "description": "My first notification",
  "name": "notification-one",
  "notificationGroupId": "646e30c38a6362190025b33e",
  "preferenceSettings": {
    "in_app": true
  },
  "steps": [
    {
      "name": "the_name_of_my_step",
      "template": {
        "_feedId": "646f716b0663687df4da4672",
        "content": "Something very cool",
        "type": "in_app",
        "variables": [
          {
            "type": "String",
            "name": "organizationId",
            "required": true
          }
        ],
        "cta": {
          "type": "redirect",
          "data": {
            "url": "http://localhost"
          }
        }
      }
    }
  ]
}
```

### 👍 Expected behavior

It should populate `_feedId` when the developer wants to generate resources programmatically a 

There's not much to show, by passing the `_feedId` you'll be able to observe the bug

_No response_

None

codingmickey:

Change the `_feedId` to `feedId` (also `_layoutId` to `layoutId`) in the shared entities to make sure to get updated in all the sdks 👍🏽 

Closes #3505 

Additional unit test added for this fix and all tests are passing 🚀 
image


codingmickey:
> What do you think about maybe extending `INotificationTemplateStep` in the [notification template interface](https://github.com/novuhq/novu/blob/baad09d42dc20d144cf4ac2beae4a2a20c825b7c/packages/node/src/lib/notification-template/notification-template.interface.ts) with feedId and layoutId ?

That would work. But wouldn't this be a problem for the other packages or the services in which they are exposed? Or this would be just internally called via the web/api/widget so wouldn't cause an issue?

ainouzgali:
> Btw, how do you guys test such integration changes? Like going into each module and then checking or waiting till you make a PR and then checking which of them failed

What do you mean? If you run everything locally it would show errors by itself. But for changes that you couldn't think to check, we have an action that runs all tests of packages that might have been affected when opening a pr

codingmickey:
> What do you mean? If you run everything locally it would show errors by itself. But for changes that you couldn't think to check, we have an action that runs all tests of packages that might have been affected when opening a pr

Yea got it, but I didn't run the tests for web/api from jarvis before raising the PR

ainouzgali:
@codingmickey I thought about it again, and there might be issues with what I suggested 🤔 🤦🏻‍♀️
Sorry for the delay, I would just like to get my colleagues opinion about this .@p-fernandez , What do you think a better way to handle this will be?

p-fernandez:
> @codingmickey I thought about it again, and there might be issues with what I suggested 🤔 🤦🏻‍♀️ Sorry for the delay, I would just like to get my colleagues opinion about this .@p-fernandez , What do you think a better way to handle this will be?

I think the best way to handle this is to do something like it was done in the Layouts feature and other places: https://github.com/novuhq/novu/blob/next/apps/api/src/app/layouts/usecases/create-layout/create-layout.use-case.ts#L83
https://github.com/novuhq/novu/blob/635cf96b737075c80b0bde9330a5eb6847e000bd/apps/api/src/app/topics/use-cases/get-topic/get-topic.use-case.ts#L25

That is creating mapper functions inside the use case to handle the transformation of the values from the entity interface to the DTO that is being used in the API responses, and therefore in all the SDKs.

The problem here is that there is a bit of tech debt in our monorepo from the confusion of that for DTOs we should expose the string type of the equivalent of the Mongo DB Ids, that shouldn't be exposed like that. That and the duplication of underscore and non underscore indexes due to that typing mixing and problems generates this kind of problems we need to slowly solve.  So in my opinion we need to handle this inside the API in as many places we can and not to push the fixes to the SDKs. Probably this will mean to have to do many changes in different places but while the API DTO structure is not changed neither in shape nor type we should be ok regarding retrocompatibility, while fixing this problem.

codingmickey:
Oh okay, got it.
- So for now this change will be only required in the SDK right?
- Also, the mapper is just what I've implemented in terms of the inheritance is the same thing right which was expected?

p-fernandez:
> Oh okay, got it.
> 
> * So for now this change will be only required in the SDK right?
> * Also, the mapper is just what I've implemented in terms of the inheritance is the same thing right which was expected?

No, the changes should be done in the API. SDK should remain untouched as the SDK just get the response provided by the API endpoint. Also the testing should be done in the API to double check we are not creating breaking changes in the API.
When I have a bit of free time I might dig up a bit more on the specific changes we need to do.

codingmickey:
> > Oh okay, got it.
> > 
> > * So for now this change will be only required in the SDK right?
> > * Also, the mapper is just what I've implemented in terms of the inheritance is the same thing right which was expected?
> 
> No, the changes should be done in the API. SDK should remain untouched as the SDK just get the response provided by the API endpoint. Also the testing should be done in the API to double check we are not creating breaking changes in the API. When I have a bit of free time I might dig up a bit more on the specific changes we need to do.

Ok sure I'll wait till then

Khongchai:
Hello, folks! Is this still being worked on? I stumbled upon this as well. If this is not being worked on anymore, or if help is needed, I'd be happy to lend a hand.

scopsy:
Will be closing this as in the inbox component we have moved towards tags based query for tabs vs the existing feeds flow

Issue #1959, opened 2022-11-03
🚀 Feature: Update to current active version of Node (18)


Cliftonz:
### 🔖 Feature description

This Issue is to discuss and track the progress of updating to the next LTS version of node.

### 🎤 Why is this feature needed ?

This will allow the platform to gain from underlying improvements to the ecosystem on top of being in support for 2+ years.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

gitstart:
@scopsy I would like to pick this up

PR #3546, opened 2023-05-31
NV-1959 - 🚀 Feature: Update to current active version of Node (18)

gitstart:
Updates to current active version of Node (18)

closes https://github.com/novuhq/novu/issues/1959

### Other information
[Ticket link](https://developers.gitstart.com/client/novu-oss/ticket/NV-1959)

---


scopsy:
There are quite a lot of other places this PR misses, including docker images and etc... This change could lead to some unwanted behaviour, so will close this PR for now so we can have a more dedicated and deep review of all the required places that need to change

PR #3549, opened 2023-06-01
fix noNewNotification field on russian locale

PidginEnemy:

This PR fixes ru locale

Because now it's have a mistake in `noNewNotification` field, it's writtend on bulgarian instead of russian


p-fernandez (approved):
🌟 
Thank you for helping us to fix this.

PR #3550, opened 2023-06-02
docs(run-locally.md): duplicate REDIS_DB_INDEX key

origranot:

There are two ```REDIS_DB_INDEX``` keys

I guess it should be only one key..

### DIFF
--- a/docs/docs/community/run-locally.md
+++ b/docs/docs/community/run-locally.md
@@ -65,6 +65,7 @@ However, if you want to test certain parts of Novu or run it in production mode,
 - `REDIS_DB_INDEX`<br />The Redis database index
 - `REDIS_CACHE_SERVICE_HOST`<br />The domain / IP of your redis instance for caching
 - `REDIS_CACHE_SERVICE_PORT`<br />The port of your redis instance for caching
+- `REDIS_CACHE_DB_INDEX`<br />The Redis cache database index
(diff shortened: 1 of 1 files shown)


docs/docs/community/run-locally.md
```diff
@@ -65,7 +65,6 @@ However, if you want to test certain parts of Novu or run it in production mode,
 - `REDIS_DB_INDEX`<br />The Redis database index
 - `REDIS_CACHE_SERVICE_HOST`<br />The domain / IP of your redis instance for caching
 - `REDIS_CACHE_SERVICE_PORT`<br />The port of your redis instance for caching
-- `REDIS_DB_INDEX`<br />The Redis cache database index
```
p-fernandez, line 68: This needs to be added but the name was wrong. Should be `REDIS_CACHE_DB_INDEX`.

p-fernandez (changes requested):
Indeed you have spotted something wrong but you focused in the wrong thing. The environment variable name is wrong.

docs/docs/community/run-locally.md
origranot: I had a guess it might also be the case, let me change it real quick 💯

p-fernandez (approved):
🌟 
Thanks for your contribution!

Issue #3142, opened 2023-04-04
[NV-2057] 🚀 Feature: Get topics of subscriber


szbartnik:
### 🔖 Feature description

Hi! In our use case it would be helpful to be able to list all subscribers added to a specific topic.

### 🎤 Why is this feature needed ?

Currently there is no way to do it. I believe there might be more people who may find it helpful to get such data efficiently.

Another reason is that it will be useful in development. Currently no UI exists where topics are visible so even when you have only 10 topics and each of them has 10 added subscribers it's not very easy to immediately make sure where a specific subscriber belongs.

### ✌️ How do you aim to achieve this?

Currently the `GET /v1/topics` endpoint has 3 optional query parameters:
- `page`
- `pageSize`
- `key` (topic key)

I was thinking about adding another optional query parameter: `subscriberId` which will narrow the results to include only topics which have a provided subscriber added.

### 🔄️ Additional Information

Workarounds:
- Listing all topics using `GET /v1/topics` endpoint & aggregate the results which is inefficient
- Implementing custom topic abstraction logic & do not use the one Novu provides

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

<sub>[NV-2057](https://linear.app/novu/issue/NV-2057/🚀-feature-get-topics-of-subscriber)</sub>

gitstart:
@davidsoderberg I would like to pick this up cc @scopsy

kgrimsby:
Could I also add that adding a wildcard to the `key` param would help?

We have topics like this: `itemA:signalA`, `itemA:signalB` and so on. If we could query `key=itemA:*` and `subscriberId=xxx` that would be really helpfully.

gitstart:
> Hi @gitstart Are you still interested to work on this?

Sure @jainpawan21

shafikmevrik:
Is this feature is done?? or still pending?

gitstart:
> Is this feature is done?? or still pending?

There is a PR up for it here

https://github.com/novuhq/novu/pull/3551

bartelemi:
[...]

PR #3551, opened 2023-06-02
NV-2057 - 🚀 Feature: Get topics of subscriber

gitstart:
This feature enables user to filter topics by subscriber ID.

closes https://github.com/novuhq/novu/issues/3142

### Other information

---


gitstart:
@scopsy this is ready for review.

apps/api/src/app/topics/use-cases/filter-topics/filter-topics.use-case.ts
```diff
@@ -21,21 +21,19 @@ export class FilterTopicsUseCase {
...
-    const totalCount = await this.topicRepository.count(query);
-
     const skipTimes = page <= 0 ? 0 : page;
     const pagination = {
       limit: pageSize,
       skip: skipTimes * pageSize,
     };
 
-    const filteredTopics = await this.topicRepository.filterTopics(query, pagination);
+    const { data, totalCount } = await this.topicRepository.filterTopics(query, pagination, command.subscriberId);
```
p-fernandez, line 30: SubscriberId should be passed in the query object. Adding it as an extra param complicates things a bit in a way I find unnecessary when we are already passing an object with the information for the DAL query.

libs/dal/src/repositories/topic/topic.repository.ts
```diff
@@ -54,31 +54,40 @@ export class TopicRepository extends BaseRepository<TopicDBModel, TopicEntity, E
...
-      {
-        $limit: pagination.limit,
+    const subscriberFilter = subscriberId && {
+      $match: {
+        $expr: {
+          $in: [subscriberId, '$topicSubscribers.externalSubscriberId'],
+        },
       },
+    };
+    const commonAggQuery = [{ $match: parsedQuery }, lookup, ...(subscriberFilter ? [subscriberFilter] : [])];
```
p-fernandez, line 74: Here we are splitting the matching step in 2 different places and I do not think it is optimal. I also understand the way of fetching the data due the lookup is not a straight forward implementation.
Thus my original idea to implement this functionality was to potentially implement an endpoint:
``` GET /topic-subscribers?subscriberId=:subscriberId``` or
``` GET /topics/subscribers/:subscriberId``` to handle this functionality to simplify implementation.
Therefore playing more with the search in the subresources (topic subscribers) than in the main resource (topics).

p-fernandez (changes requested):

libs/dal/src/repositories/topic/topic.repository.ts
gitstart: Thanks for the feedback @p-fernandez .

Would you prefer that we revert the entire implementation and implement a new endpoint or you would like to keep both?

gitstart: @p-fernandez gentle reminder on this

gitstart:
@p-fernandez I've updated the PR with the new endpoint. Although I didn't discard the old implementation. Would you prefer that it be discarded?

libs/dal/src/repositories/topic/topic.repository.ts
scopsy: @p-fernandez I think that if we are returning the `topics` resource here, the `subscriberId` is only a filter query to filter the topics resource. So I would stick with just leaving the query param. 
The `/topics/subscribers/:subscriberId` suggest we are returning a `subscriber` entity in the context of a topic, which is not really the case here. For example `/posts/:postId/comments` will return a list of comments related to the post vs a list of posts for a particular comment.

Wdyt?

p-fernandez: I agree that `/topics/subscribers/:subscriberId` could be misleading on what's being returned. Some APIs would clarify that by enabling URLs like this `/topics/all/subscribers/:subscriberId`. That will clarify what's being returned but being the topic external identifier some free choice text for our users we could hit some bug and we would need to start to forbid, thing that I am not keen to.
That's why I suggested the `GET /topic-subscribers?subscriberId=:subscriberId` option, also to not overcrowd the other option that was to implement it in the Subscribers API through `GET /subscribers/:subscriberId/topics`. But thinking in every module as it would have been a dedicated µservice I thought allow the resource `topic-subscribers` as maybe the best option.
Probably `/topics-subscribers` could be the resource proper resource.

Another worry is the usage of this endpoint. If our users would want to also return all the topic information that would need to filtered where @gitstart was doing it. But we have a problem with the API as in `TopicDto` the property subscribers is mandatory, but wouldn't make sense for this functionality, so we are between a rock and a hard place with a potential breaking change or a non performant endpoint when filtered by `suscriberId` trying to bring the relationships of the subscribers with the topics filtered.

Issue #3473, opened 2023-05-22
🚀 Feature: API endpoint to fetch execution details by transactionId


chauandrew:
### 🔖 Feature description

Create a new api endpoint that returns the execution details for events related to a given `transactionId`. 

### 🎤 Why is this feature needed ?

Currently, we support fetching execution details by `notificationId` and `subscriberId`. However, when we trigger a notification to a topic, we cannot easily get the execution details for all dependencies of that action. 

eg: we trigger a notification to a topic with 100 subscribers. It is difficult to get the execution details for all 100 subscribers.

In this case, it would be helpful to query the execution details by `transactionId`.

### ✌️ How do you aim to achieve this?

Create a new api endpoint (`/execution-details/transaction/:transactionId`) that returns the execution details for a given transaction. 

Since a transaction can have many notifications/subscribers attached to it, I think it makes sense to group each set of execution-details by notificationId/subscriberId.

I would also like to add the ability to filter by status (Pending, Success, Failure). 

The query will also need to be paginated, so the json response should include `totalCount`, `pageSize`, and `page`. 

### 🔄️ Additional Information

Here's a [thread](https://discord.com/channels/895029566685462578/1107698920643108964) in the discord with some more discussion on the topic. 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

NK8916:
Hey,can i pick this up?

jainpawan21:
Sure @NK8916 

Could you please share your approach, how you will add this feature?

gitstart:
@jainpawan21 I would like to pick this up if it's still available, here is my approach. The aim is to create an endpoint that filters execution details by `transactionId` and other stated parameters.

1. Implement a repository function at the execution-details level that will allow filtering by transaction and other stated params
2. This repository can then be injected into services that will expose the feature to the controller
3. The query written will be optimized with params such as counts, page size, and current page to improve performance

PR #3567, opened 2023-06-07
NV-2381 - 🚀 Feature: API endpoint to fetch execution details by transactionId

gitstart:
This feature enables users to filter execution details by transactionId

closes https://github.com/novuhq/novu/issues/3473

### Other information

Happy hacking! 🎉

---


gitstart:
@scopsy This is ready for review.

apps/api/src/app/execution-details/execution-details.controller.ts
```diff
@@ -36,4 +46,27 @@ export class ExecutionDetailsController {
...
     );
   }
+
+  @Get('/transaction/:transactionId')
+  @ApiResponse(ExecutionDetailsFilterResponseDto)
+  @ApiOperation({
+    summary: 'Get execution details by transaction id',
+  })
+  @ExternalApiAccessible()
+  getNotificationTemplates(
```
ainouzgali, line 56: Copy Paste🤪🙃

apps/api/src/app/execution-details/execution-details.controller.ts
```diff
@@ -36,4 +46,27 @@ export class ExecutionDetailsController {
...
+    @Query() query: GetExecutionDetailsByTransactionRequestDto
+  ): Promise<ExecutionDetailsFilterResponseDto> {
+    return this.getExecutionDetailsByTransactionId.execute(
+      GetExecutionDetailsByTransactionIdCommand.create({
+        transactionId,
+        organizationId: user.organizationId,
+        userId: user._id,
+        environmentId: user.environmentId,
+        page: query.page ? query.page : 0,
+        limit: query.limit ? query.limit : 1,
```
ainouzgali, line 68:
```suggestion
        limit: query.limit ? query.limit : 10,
```
I feel 10 is a reasonable default limit

ainouzgali, line 68:
```suggestion
        page: query.page ??  0,
        limit: query.limit ??  10,
```
Up to you, this syntax is my preference  :)

apps/api/src/app/execution-details/execution-details.controller.ts
```diff
@@ -36,4 +46,27 @@ export class ExecutionDetailsController {
       })
     );
   }
+
+  @Get('/transaction/:transactionId')
```
ainouzgali, line 50: I am not sure about the structure of this endpoint. The beginning with  '/transaction' I mean. I wonder if we should maybe do transactionId as a parameter  or maybe only '/:transactionId' 
@djabarovgeorge do you have a suggestion?

ainouzgali (commented):
Hi @gitstart ! Thank you for all your thoughtful contributions!! Much appreciated. 

Added some small comments and suggestions, let me know what you think

apps/api/package.json
```diff
@@ -118,4 +118,4 @@
       "eslint"
     ]
   }
-}
```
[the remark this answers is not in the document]
ainouzgali: oops, sorry. started the review before your recent changes :)

apps/api/src/app/execution-details/execution-details.controller.ts
gitstart: @ainouzgali Great suggestion. Looking at it, that makes more sense.

Looking at the base get endpoint, the main difference between this new one and that one is that we're filtering by `transactionId` so `/:transactionId` seems like a good structure and we've already implemented it that way in the latest commit.

p-fernandez: The path should be pluralised as per REST design.
```/transactions/:transactionId```.

File: libs/dal/src/repositories/execution-details/execution-details.repository.ts [hunk 5]
```diff
@@ -44,4 +44,26 @@ export class ExecutionDetailsRepository extends BaseRepository<
...
+   */
+  public async findAllNotificationExecutionsByTransactionId(
+    transactionId: string,
+    environmentId: string,
+    skip = 0,
+    limit = 10
+  ) {
+    const requestQuery = {
+      _environmentId: environmentId,
+      transactionId: transactionId,
```
djabarovgeorge, line 59: transactionId is not indexed on our end, therefore we will need to add it so we won't scan all the collection, we are using Compound Indexes so we should add it to the 
`executionDetailsSchema.index({
  _environmentId: 1,
});`
as a second index or create a new one for it, I am not entirely sure yet what will be the most optimized one.

gitstart:
@ainouzgali @p-fernandez @djabarovgeorge your feedback comments have been implemented.

scopsy:
@p-fernandez jumping in here, shouldn't we just have the transactionId as a query filter on the existing GET / method?

PR #3584, opened 2023-06-13
feat: add log when segment is throwing error

djabarovgeorge:

Add catch and log the error we get from segment.

In order to prevent job failure.

### DIFF
--- a/packages/application-generic/src/services/analytics.service.ts
+++ b/packages/application-generic/src/services/analytics.service.ts
@@ -19,6 +19,8 @@ interface IUser {
   createdAt?: string | null;
 }
 
+const LOG_CONTEXT = 'AnalyticsService';
+
 export class AnalyticsService {
   private segment: Analytics;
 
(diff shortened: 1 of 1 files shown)


packages/application-generic/src/services/analytics.service.ts
```diff
@@ -102,11 +104,22 @@ export class AnalyticsService {
...
+          userId: userId,
+          event: name,
+          properties: data,
+        });
+      } catch (error: any) {
+        Logger.error(
+          JSON.stringify({
+            name: name,
+            usedId: userId,
+            message: error.message,
```
djabarovgeorge, line 118: I am wondering if we could add more data here, like the transaction id etc..
one way of doing it is passing it as part of the data to track function. and then add it here 
`transactionId: data. transactionId,`

packages/application-generic/src/services/analytics.service.ts
```diff
@@ -19,6 +19,8 @@ interface IUser {
   createdAt?: string | null;
 }
 
+const LOG_CONTEXT = 'AnalyticsService';
```
p-fernandez, line 22: 👏🏻

packages/application-generic/src/services/analytics.service.ts
p-fernandez: I think it is a good idea. Also we then so can start to design the interface of the `data` passed. 👍🏻

p-fernandez (approved):
🌟

packages/application-generic/src/services/analytics.service.ts
scopsy: I think that pino supporting objects, so we don't need to pass stringify there. We can have an error name, and pass the whole object.

In most cases where transacitonId is available, we automatically add it to the logger context. So we don't need to manually specify it here. It will be automatically added :)

PR #3588, opened 2023-06-14
fix(infra): remove node package strange dependencies and move to shared

p-fernandez:

Removes strange Node package dependencies from most of the apps and libs that are dependencies for either our infra or setup.

I notice about this while trying to solve the failing tests of another PR. Part is inherited of a bad implementation of the Topics feature and other were inherit from previous refactors.
Note that we are missing to totally remove a dependency of the Node package from API app as all the invite system is depending on it (not ideal).

### DIFF
--- a/apps/api/src/app/events/dtos/trigger-event-request.dto.ts
+++ b/apps/api/src/app/events/dtos/trigger-event-request.dto.ts
@@ -1,7 +1,6 @@
 import { ArrayMaxSize, ArrayNotEmpty, IsArray, IsDefined, IsObject, IsOptional, IsString } from 'class-validator';
 import { ApiExtraModels, ApiProperty, ApiPropertyOptional, getSchemaPath } from '@nestjs/swagger';
-import { TriggerRecipientSubscriber, TriggerRecipients } from '@novu/node';
-import { TopicKey, TriggerRecipientsTypeEnum } from '@novu/shared';
+import { TopicKey, TriggerRecipientSubscriber, TriggerRecipients, TriggerRecipientsTypeEnum } from '@novu/shared';
 import { CreateSubscriberRequestDto } from '../../subscribers/dtos/create-subscriber-request.dto';
 
 export class SubscriberPayloadDto extends CreateSubscriberRequestDto {}
--- a/apps/worker/package.json
+++ b/apps/worker/package.json
@@ -32,7 +32,6 @@
     "@nestjs/terminus": "9.2.1",
     "@novu/application-generic": "^0.16.0",
     "@novu/dal": "^0.16.0",
-    "@novu/node": "^0.16.0",
     "@novu/shared": "^0.16.0",
     "@novu/stateless": "^0.16.0",
     "@novu/testing": "^0.16.0",
--- a/libs/shared/src/dto/events/event.interface.ts
+++ b/libs/shared/src/dto/events/event.interface.ts
@@ -1,4 +1,9 @@
-export enum TriggerRecipientsTypeEnum {
-  SUBSCRIBER = 'Subscriber',
-  TOPIC = 'Topic',
-}
+import { ISubscribersDefine, ITopic, TopicKey } from '../../types';
+
+export type TriggerRecipientSubscriber = string | ISubscribersDefine;
+
+export type TriggerRecipient = TriggerRecipientSubscriber | ITopic;
+
+export type TriggerRecipients = TriggerRecipient[];
+
+export type TriggerRecipientsPayload = TriggerRecipientSubscriber | TriggerRecipients;
--- a/libs/shared/src/dto/subscriber/subscriber.dto.ts
+++ b/libs/shared/src/dto/subscriber/subscriber.dto.ts
@@ -1,6 +1,5 @@
 import { ChatProviderIdEnum, PushProviderIdEnum } from '../../consts';
-import { SubscriberCustomData } from '../../entities/subscriber';
-import { EnvironmentId, ExternalSubscriberId, OrganizationId } from '../../types';
+import { EnvironmentId, ExternalSubscriberId, OrganizationId, SubscriberCustomData } from '../../types';
 
 interface IChannelCredentials {
   webhookUrl?: string;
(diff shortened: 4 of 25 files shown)


apps/api/src/app/events/dtos/trigger-event-request.dto.ts
```diff
@@ -1,7 +1,6 @@
 import { ArrayMaxSize, ArrayNotEmpty, IsArray, IsDefined, IsObject, IsOptional, IsString } from 'class-validator';
 import { ApiExtraModels, ApiProperty, ApiPropertyOptional, getSchemaPath } from '@nestjs/swagger';
-import { TriggerRecipientSubscriber, TriggerRecipients } from '@novu/node';
```
p-fernandez, line 3: We had a lot of dependencies of types from the Node package so I moved them to the Shared package as it would be expected.

apps/worker/package.json
```diff
@@ -32,7 +32,6 @@
     "@nestjs/terminus": "9.2.1",
     "@novu/application-generic": "^0.16.0",
     "@novu/dal": "^0.16.0",
-    "@novu/node": "^0.16.0",
```
p-fernandez, line 35: Worker app shouldn't have this dependency at all (plus not being used).

libs/shared/src/dto/events/event.interface.ts
```diff
@@ -1,4 +1,9 @@
-export enum TriggerRecipientsTypeEnum {
-  SUBSCRIBER = 'Subscriber',
-  TOPIC = 'Topic',
-}
```
p-fernandez, line 4: Wrongly located as this is not a DTO.

libs/shared/src/dto/subscriber/subscriber.dto.ts
```diff
@@ -27,18 +26,3 @@ export class SubscriberDto {
...
-  phone?: string;
-  avatar?: string;
-  locale?: string;
-  data?: SubscriberCustomData;
-  [key: string]: string | string[] | boolean | number | SubscriberCustomData | undefined;
-}
-
-export interface ISubscribersDefine extends ISubscriberPayload {
-  subscriberId: string;
-}
```
p-fernandez, line 44: More interfaces than DTOs.

libs/shared/src/entities/subscriber/subscriber.interface.ts
```diff
@@ -29,5 +30,3 @@ export interface IChannelCredentials {
   webhookUrl?: string;
   deviceTokens?: string[];
 }
-
-export type SubscriberCustomData = { [key: string]: string | string[] | boolean | number | undefined };
```
p-fernandez, line 33: More a type interface than a entity.

libs/testing/package.json
```diff
@@ -23,7 +23,6 @@
   "dependencies": {
     "@faker-js/faker": "^6.0.0",
     "@novu/dal": "^0.16.0",
-    "@novu/node": "^0.16.0",
```
p-fernandez, line 26: Testing lib shouldn't have this dependency at all.

libs/testing/src/user.session.ts
```diff
@@ -137,10 +141,6 @@ export class UserSession {
       this.subscriberToken = token;
       this.subscriberProfile = profile;
     }
-
-    this.serverSdk = new Novu(this.apiKey, {
```
p-fernandez, line 141: Not used at all.

packages/application-generic/package.json
```diff
@@ -68,7 +68,6 @@
     "@novu/mattermost": "^0.16.0",
     "@novu/ms-teams": "^0.16.0",
     "@novu/netcore": "^0.16.0",
-    "@novu/node": "^0.16.0",
```
p-fernandez, line 71: Application  generic shouldn't have this dependency at all (plus not being used).

packages/node/src/lib/topics/topic.interface.ts
```diff
@@ -1,15 +1,11 @@
 import {
   ExternalSubscriberId,
+  ITopic,
   TopicKey,
   TopicName,
   TriggerRecipientsTypeEnum,
 } from '@novu/shared';
 
-export interface ITopic {
```
p-fernandez, line 8: Should live in Shared as it is reused.

packages/node/src/lib/novu.interface.ts
```diff
@@ -1,17 +1,9 @@
 import { AxiosInstance } from 'axios';
-import { ChannelTypeEnum } from '@novu/shared';
 
 export interface INovuConfiguration {
   backendUrl?: string;
 }
 
-export interface IAttachmentOptions {
```
p-fernandez, line 8: Should live in Shared as it is reused.

packages/node/src/lib/events/events.interface.ts
```diff
@@ -1,19 +1,10 @@
...
-export type TriggerRecipientTopics = ITopic[];
-
-export type TriggerRecipient = TriggerRecipientSubscriber | ITopic;
-
-export type TriggerRecipients = TriggerRecipient[];
-
-// string | ISubscribersDefine | (string | ISubscribersDefine | ITopic)[]
-export type TriggerRecipientsPayload =
-  | TriggerRecipientSubscriber
-  | TriggerRecipients;
```
p-fernandez, line 16: Should live in Shared as it is reused.

packages/application-generic/src/factories/mail/interfaces/send.handler.interface.ts
```diff
@@ -1,5 +1,5 @@
+import { IEmailOptions } from '@novu/shared';
 import {
-  IEmailOptions,
```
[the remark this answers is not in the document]
ainouzgali: Could you please elaborate a bit what you mean?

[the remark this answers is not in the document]
p-fernandez: @ainouzgali Application generic are a set of use cases and services that are used mostly by API and Worker apps. As far as I understand, Stateless is a simplified API layer to be able to execute the providers added to Novu. So I can't figure out why either API or Worker should have a dependency on Stateless package.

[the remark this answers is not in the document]
ainouzgali: Only for the types it seems. I believe you can just create those types in Shared as well. As stateless is deprecated as far as I know. There should not be future changes in stateless.

ainouzgali (approved):
Looks good, couldn't find anything missing

PR #3593, opened 2023-06-14
fix(web): show add step side menu after delete of step

ainouzgali:
Fixes the  following:

Select the step stat opens the step side settings like: digest, delay, sms, push, chat.
Try to delete the step, then the step side settings disappear and the default channels panel is not shown.

Fixed by navigating back to the basePath.


p-fernandez (approved):
🌟

PR #3595, opened 2023-06-14
Generalize jitter function from cache and add a Delay function

Cliftonz:

These functions are to help with the implementation of https://linear.app/novu/issue/NV-2408/job-backoff-retry-policy for jobs. These functions can be used in many places and are moved out from cache to a resiliency package.

I have added 3 functions:
- addjitter which takes in some time and adds jitter based on the variance percentage passed in
- getDelay which get the delay amount based on the number retries
- getDelayWithJitter does the same as above but applies the jitter after getting the delay


packages/application-generic/src/services/cache/cache.service.ts
```diff
@@ -157,13 +158,7 @@ export class CacheService implements ICacheService {
...
     const seconds = options?.ttl || this.cacheTtl;
 
-    return this.ttlVariant(seconds);
-  }
-
-  private ttlVariant(num): number {
-    const variant = this.TTL_VARIANT_PERCENTAGE * num * Math.random();
-
-    return Math.floor(num - (this.TTL_VARIANT_PERCENTAGE * num) / 2 + variant);
+    return addJitter(seconds, this.TTL_VARIANT_PERCENTAGE);
```
djabarovgeorge, line 161: Is adding Jitter the term of what we want to add here? 
I am asking because it is a new term for me 😅

djabarovgeorge (approved):
🚀

packages/application-generic/src/services/cache/cache.service.ts
Cliftonz: Its a common term over the statistical one, variance. 
When I abstracted this jitter was the better choice.

p-fernandez: AWS uses `jitter` to define the variable time to wait between retries, so it is rightly used.

Issue #3548, opened 2023-06-01
🐛 Bug Report: Can't embed hyperlink in Slack message 


phlrss:
### 📜 Description

According to [slack docs](https://api.slack.com/reference/surfaces/formatting#linking-urls) the way to insert a hyperlink in a Slack message is the following:
`<http://www.example.com|This message is a link>`
When i try using this, anything with `< >` gets auto-removed.

### 👟 Reproduction steps

1. Add a Chat step to a template
2. try adding a hyperlink to your Slack message like this: `<http://www.example.com|This message is a link>`
3. When you click the "Update" button the entire text with `<>` gets automatically removed

### 👍 Expected behavior

it should allow me to embed hyperlinks in my Slack message

### 👎 Actual Behavior with Screenshots

it auto-removes the hyperlink

https://github.com/novuhq/novu/assets/46248889/ccc06d2e-a78d-4cf5-a7e3-e945ece9e31f

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

scopsy:
This was resolved with Echo: https://docs.novu.co/echo/quickstart

PR #3610, opened 2023-06-16
NV-3548 - 🐛 Bug Report: Can't embed hyperlink in Slack message

gitstart:
Allows users to embed links in the message.

Closes: https://github.com/novuhq/novu/issues/3548

### Other information

 [Loom Video](https://www.loom.com/share/83cec8292cdb452eaaceb03bb453dc14?sid=33b2553c-a208-43d9-a7aa-790d7af983e3)

---


gitstart:

gitstart:

apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
```diff
@@ -42,7 +42,7 @@ const makeStep = (channelType: StepTypeEnum, id: string): IFormStep => {
       subject: '',
       type: channelType,
       content: channelType === StepTypeEnum.EMAIL ? defaultEmailBlocks : '',
-      contentType: 'editor',
+      contentType: 'customHtml',
```
scopsy, line 45: @gitstart this change will modify all existing channel which is something we don't want. This fix should only apply to the "Chat" channel and not on all available channels 🙏

scopsy (changes requested):

apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
gitstart: @scopsy sorry for the delay on this. The latest updates resolves this. Here is how it renders on slack:

image

scopsy:
@gitstart looks better! Is this change backwards. compatible if the user had content on the step previously in the editor type?

gitstart:
@scopsy the change only affects new ones. So no. It's not backwards compatible in that case.

gitstart:
@scopsy How do you want to proceed with this?

scopsy:
@gitstart we would need to validate that old workflows with chat content, are still rendered after this change in the `customHtml` editor. I believe this should work, but will need to validate this.

PR #3614, opened 2023-06-18
🐛 CreateNotificationJobCommand fix - update templateProviderIds Map to Record.

djabarovgeorge:

Changed the templateProviderIds Map to Record.

Looks like the `plainToInstance` (from 'class-transformer') in the BaseCommand is mapping Map values to 'undefined'.

https://github.com/typestack/class-transformer/issues/1256


scopsy:
@djabarovgeorge do we have any way to create a test case for this value? I think it's the second time it's doing problems :(

djabarovgeorge:
> @djabarovgeorge do we have any way to create a test case for this value? I think it's the second time it's doing problems :(

Do you mean that `templateProviderIds` is doing problems?

scopsy:
@djabarovgeorge yes, it caused some issue previously with undefined values for providerId when the job is created. It was fixed later, but now seems there is another issue in that area.

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -241,14 +242,14 @@ export class TriggerEvent {
               channelType
             );
             if (provider) {
-              providers.set(channelType, provider);
+              providers[channelType] = provider;
```
p-fernandez, line 245: Isn't this to cause problems for example in the email channel with the Novu provider plus the email provider a user might enable?

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -241,14 +242,14 @@ export class TriggerEvent {
...
             }
           }
         }
       }
     }
 
-    return providers;
+    return providers as Record<ChannelTypeEnum, ProvidersIdEnum>;
```
LetItRock, line 252: this is a bad thing, you say it's partial and then cast it to the record at the end... please define it as record

packages/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -87,9 +87,10 @@ export class CreateNotificationJobs {
       if (!step.template)
         throw new PlatformException('Step template was not found');
 
-      const providerId = command.templateProviderIds.get(
-        STEP_TYPE_TO_CHANNEL_TYPE.get(step.template.type) as ChannelTypeEnum
-      );
+      const channel = STEP_TYPE_TO_CHANNEL_TYPE.get(
+        step.template.type
+      ) as ChannelTypeEnum;
```
LetItRock, line 92: I feel we should change the `STEP_TYPE_TO_CHANNEL_TYPE` to object as we now need to cast the value always

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
LetItRock: and won't return the novu email integration?

packages/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
djabarovgeorge: Looks like the cast was not needed at all so i removed it.

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -196,6 +196,7 @@ export class SendMessageInApp extends SendMessageBase {
         transactionId: command.transactionId,
         content: this.storeContent() ? content : null,
         payload: messagePayload,
+        providerId: integration.providerId,
```
djabarovgeorge, line 199: funny thing, we did not save the provider on the message, which also means that the previous query of `oldMessage` always failed.

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -257,7 +258,7 @@ export class SendMessageInApp extends SendMessageBase {
       CreateExecutionDetailsCommand.create({
         ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
         messageId: message._id,
-        providerId: InAppProviderIdEnum.Novu,
+        providerId: integration.providerId,
```
djabarovgeorge, line 261: added dynamic provider id

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -296,7 +297,7 @@ export class SendMessageInApp extends SendMessageBase {
...
```
djabarovgeorge, line 300: same

LetItRock (approved):

PR #3620, opened 2023-06-19
NV-2457 - Create Tenant Repository

djabarovgeorge:

The tenants entity. The tenants for the current version should include the following information:

* Unique Identifier (key/tenantId) - A user-provided id that will be used to retrieve and manipulate tenants.
* Name - The name of the tenant
* (Optional) Custom Data - An object provided by the user to contain custom tenant data. 

Tenants are attached to the environment level of the system.

So that our users will be able to manage and create tenants in our system.

### DIFF
--- a/libs/dal/src/repositories/tenant/tenant.entity.ts
+++ b/libs/dal/src/repositories/tenant/tenant.entity.ts
@@ -1,0 +1,22 @@
+import { TenantCustomData } from '@novu/shared';
+import { TenantId } from './types';
+import { EnvironmentId } from '../environment';
+import { ChangePropsValueType } from '../../types/helpers';
+
+export class TenantEntity {
+  _id: TenantId;
+
+  identifier: string;
+
+  name?: string;
+
+  deleted?: boolean;
+
+  createdAt: string;
+
+  data?: TenantCustomData;
+
+  _environmentId: EnvironmentId;
+}
+
+export type TenantDBModel = ChangePropsValueType<TenantEntity, '_environmentId'>;
--- a/libs/dal/src/repositories/tenant/tenant.schema.ts
+++ b/libs/dal/src/repositories/tenant/tenant.schema.ts
@@ -1,0 +1,28 @@
+import * as mongoose from 'mongoose';
+import { Schema } from 'mongoose';
+
+import { schemaOptions } from '../schema-default.options';
+import { TenantDBModel } from './tenant.entity';
+
+const tenantSchema = new Schema<TenantDBModel>(
+  {
+    identifier: Schema.Types.String,
+    name: Schema.Types.String,
+    deleted: Schema.Types.Boolean,
+    createdAt: Schema.Types.String,
+    data: Schema.Types.Mixed,
+    _environmentId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Environment',
+    },
+  },
+  schemaOptions
+);
+
+tenantSchema.index({
+  _environmentId: 1,
+});
+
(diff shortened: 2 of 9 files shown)


libs/dal/src/repositories/tenant/tenant.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  deleted?: boolean;
+
+  createdAt: string;
+
+  data?: TenantCustomData;
+
+  _environmentId: EnvironmentId;
+}
+
```
djabarovgeorge, line 21: Not sure if an organization id needed here, and if this entity is one to many with an organization.
meaning one tenant related to two organizations.

raised a question on the linear task as well.

libs/dal/src/repositories/tenant/tenant.schema.ts
```diff
@@ -0,0 +1,28 @@
...
+      type: Schema.Types.ObjectId,
+      ref: 'Environment',
+    },
+  },
+  schemaOptions
+);
+
+tenantSchema.index({
+  _environmentId: 1,
+});
```
djabarovgeorge, line 24: the initial index will probably evolve over the iteration in this cycle.

libs/dal/src/repositories/tenant/tenant.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+  identifier: string;
+
+  name?: string;
+
+  deleted?: boolean;
+
+  createdAt: string;
+
+  data?: TenantCustomData;
```
djabarovgeorge, line 17: Additional custom data that the user can add.

libs/dal/src/repositories/tenant/tenant.entity.ts
p-fernandez: I think the hierarchy should be:
organisation > environment > tenant > workflow
so I think it could be added without harm.
But not sure if @scopsy and @LetItRock could think in a different way.

libs/dal/src/repositories/tenant/tenant.entity.ts
p-fernandez: Wondering if we should define the expected structure even if we have not a clear idea of what to allow. For example right now with the custom data type we are allowing array of strings but not array of numbers. And no nested objects.
Another point is if the needs for the subscriber custom data might be different for the tenant. I guess something to discuss soon in the future.

p-fernandez (approved):

libs/dal/src/repositories/tenant/tenant.entity.ts
djabarovgeorge: I believe `unknown` could do the job, but IMO it depends on how we will use it internally, if it will be used in the template-editor/integration-condition/template-filter, etc because if we, for example, won't support their nested object but will support in this object it could confuse the user. in one place he can store data but he won't be able to use it across the app.

meanwhile, i added the TenantCustomData in order to maintain/reuse the current BL :)

libs/dal/src/repositories/tenant/tenant.entity.ts
djabarovgeorge: Thinking about it again now i think we can add the organizationId as well, and not mix the tenants across organizations by one to many relation as i said before 😅





> wouldn't this be a tenant per env? I feel like we should have here just the `_organizationId` and no `_environmentId`, wdyt?

i am not sure because we would want to separate the tenant entity on the environment level so for example `production` level won't be affected by the `test` environment therefore more secure in a way.

Issue #3581, opened 2023-06-11
[NV-2409] 🐛 Bug Report: Querying by non-existent `subscriberId` returns all messages


thenbe:
### Good example

```bash
URL="https://api.novu.co/v1/messages"
SUBSCRIBER_ID="1"
http GET ${URL} subscriberId=="${SUBSCRIBER_ID}"
```

| Header | Header |
|--------|--------|
| Expected | Subscriber 1's messages returned |
| Actual | Subscriber 1's messages returned | 

---

### Bugged example

However, when we use a non-existent subscriber ID:

```bash
URL="https://api.novu.co/v1/messages"
SUBSCRIBER_ID="NON_EXISTENT"
http GET ${URL} subscriberId=="${SUBSCRIBER_ID}"
```
| Header | Header |
|--------|--------|
| Expected | No messages returned|
| Actual | *All* messages returned | 

Related: https://github.com/novuhq/novu/blob/5ef36d0e3fca483605ad99ae26b7f04a2656ecd0/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts#L30

### 👟 Reproduction steps

.

### 👍 Expected behavior

.

### 👎 Actual Behavior with Screenshots

.

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-2409](https://linear.app/novu/issue/NV-2409/🐛-bug-report-querying-by-non-existent-subscriberid-returns-all)</sub>

ainouzgali:
Hi @thenbe! Thanks for reporting :)
I agree, no messages should be the expected behavior.
In my opinion, should throw an error if a subscriberId is sent (command.subscriberId) to the query, but no such subscriber exists. If no subscriberId is sent to the query, should keep the current behavior. 
What do you think? @thenbe @jainpawan21

thenbe:
@ainouzgali Agreed. A `404` seems appropriate here.

gitstart:
@ainouzgali @thenbe I would like to pick this up

jainpawan21:
[...]

PR #3622, opened 2023-06-19
NV-2409 - 🐛 Bug Report: Querying by non-existent subscriberId returns all messages

gitstart:
This fix prevents returning all messages when querying with non-existent subscriberId

closes https://github.com/novuhq/novu/issues/3581

### Other information

Happy hacking! 🎉

---

### DIFF
--- a/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
+++ b/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
@@ -1,4 +1,4 @@
-import { BadRequestException, Injectable } from '@nestjs/common';
+import { BadRequestException, Injectable, NotFoundException } from '@nestjs/common';
 import { MessageEntity, MessageRepository, SubscriberRepository, SubscriberEntity } from '@novu/dal';
 import { CachedEntity, buildSubscriberKey } from '@novu/application-generic';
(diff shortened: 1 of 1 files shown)


gitstart:

apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
p-fernandez, line 1: I think it is better to replace the line 25 call to the method `this.fetchSubscriber` for the call to the use case `GetSubscriber` (https://github.com/novuhq/novu/blob/next/apps/api/src/app/subscribers/usecases/get-subscriber/get-subscriber.usecase.ts) as it handles all the logic we need for this bit:
1. Subscriber fetching
2. NotFound exception thrown if not found
3. Cache logic.

Also we should add a test case in https://github.com/novuhq/novu/blob/next/apps/api/src/app/messages/e2e/get-messages.e2e.ts for this scenario to validate the implementation.

p-fernandez (changes requested):

jainpawan21:
Hi @gitstart 
I have fixed this bug in this PR
https://github.com/novuhq/novu/pull/3619

gitstart:
Closing this since the issue has been resolved

PR #3623, opened 2023-06-20
Nv 2458 - Add Tenant Create Usecase

djabarovgeorge:

adds the create use case, in order to create new tenant entities in the system.
because it is the first tenant use case this PR will contain the initial tenant module as well.

in order to create new tenant entities.

### DIFF
--- a/apps/api/src/app/tenant/tenant.controller.ts
+++ b/apps/api/src/app/tenant/tenant.controller.ts
@@ -22,12 +22,12 @@ export class TenantController {
 
   @Post('/')
   @ExternalApiAccessible()
-  @ApiResponse(CreateTenantRequestDto)
+  @ApiResponse(CreateTenantResponseDto)
   @ApiOperation({
     summary: 'Create tenant',
     description: 'Create tenant under the current environment',
   })
-  async createOrganization(
+  async createTenant(
     @UserSession() user: IJwtPayload,
     @Body() body: CreateTenantRequestDto
   ): Promise<CreateTenantResponseDto> {
--- a/apps/api/src/app/tenant/tenant.controller.ts
+++ b/apps/api/src/app/tenant/tenant.controller.ts
@@ -1,0 +1,44 @@
+import { Body, ClassSerializerInterceptor, Controller, Post, UseGuards, UseInterceptors } from '@nestjs/common';
+import { ApiExcludeController, ApiOperation, ApiTags } from '@nestjs/swagger';
+
+import { IJwtPayload } from '@novu/shared';
+
+import { JwtAuthGuard } from '../auth/framework/auth.guard';
+import { UserSession } from '../shared/framework/user.decorator';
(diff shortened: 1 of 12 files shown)


apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+@Controller('/tenants')
+@ApiTags('Tenants')
+@UseInterceptors(ClassSerializerInterceptor)
+@UseGuards(JwtAuthGuard)
+@ApiExcludeController()
+export class TenantController {
+  constructor(private CreateTenantUsecase: CreateTenant) {}
+
+  @Post('/')
+  @ExternalApiAccessible()
```
djabarovgeorge, line 24: Exposed the endpoint to the API key accese.

apps/api/src/app/tenant/usecases/create-tenant/create-tenant.usecase.ts
```diff
@@ -0,0 +1,28 @@
...
+  constructor(private tenantRepository: TenantRepository) {}
+
+  async execute(command: CreateTenantCommand) {
+    const tenantExist = await this.tenantRepository.findOne({
+      _environmentId: command.environmentId,
+      identifier: command.identifier,
+    });
+
+    if (tenantExist) {
+      throw new BadRequestException(
```
djabarovgeorge, line 16: As far as I know, we do not want to make duplication between the environments.

libs/dal/src/repositories/tenant/tenant.schema.ts
```diff
@@ -23,6 +23,21 @@ tenantSchema.index({
...
+ * Context: execute()
+ * Query: findOne({
+ *    _environmentId: command.environmentId,
+ *    identifier: command.identifier,
+ *  });
+ */
+tenantSchema.index({
+  identifier: 1,
+  _environmentId: 1,
+});
```
djabarovgeorge, line 39: Created to the new use case.

michaldziuba03: I would suggest to make this compound index unique if this combination is supposed to be unique (if I understand correctly).

Even with `if` statement in create usecase, some weird race conditions are possible.

djabarovgeorge: Thanks, love the suggestion, I think it would be a very edge case but still, there is a possibility it could happen. I will take a look at it.

apps/api/src/app/tenant/usecases/create-tenant/create-tenant.usecase.ts
ainouzgali: Maybe throw `ConflictException`?

ainouzgali (approved):
Saw you made some related changes in the next pr. so looks good to me

PR #3625, opened 2023-06-20
feat(dal): update integration entity schema for multi provider

p-fernandez:

Update integration entity schema...

...to support multi provider configuration.

### DIFF
--- a/libs/dal/src/repositories/integration/integration.entity.ts
+++ b/libs/dal/src/repositories/integration/integration.entity.ts
@@ -1,7 +1,5 @@
 import { ChannelTypeEnum } from '@novu/shared';
 
-import { KindId, ProviderKindEnum } from './types';
-
 import type { EnvironmentId } from '../environment';
 import type { OrganizationId } from '../organization';
 import { ChangePropsValueType } from '../../types/helpers';
@@ -53,10 +51,6 @@ export class IntegrationEntity {
   name?: string;
 
   identifier?: string;
-
-  kind?: ProviderKindEnum;
-
-  kindId?: KindId;
 
   deleted: boolean;
 
--- a/libs/dal/src/repositories/integration/integration.entity.ts
+++ b/libs/dal/src/repositories/integration/integration.entity.ts
@@ -1,5 +1,7 @@
 import { ChannelTypeEnum } from '@novu/shared';
 
+import { KindId, ProviderKindEnum } from './types';
+
 import type { EnvironmentId } from '../environment';
 import type { OrganizationId } from '../organization';
 import { ChangePropsValueType } from '../../types/helpers';
@@ -47,6 +49,14 @@ export class IntegrationEntity {
   credentials: ICredentials;
 
   active: boolean;
+
+  name?: string;
+
+  identifier?: string;
+
+  kind?: ProviderKindEnum;
+
+  kindId?: KindId;
 
   deleted: boolean;
 
--- a/libs/dal/src/repositories/integration/integration.schema.ts
+++ b/libs/dal/src/repositories/integration/integration.schema.ts
@@ -45,11 +45,14 @@ const integrationSchema = new Schema<IntegrationDBModel>(
       redirectUrl: Schema.Types.String,
       hmac: Schema.Types.Boolean,
     },
-
     active: {
       type: Schema.Types.Boolean,
       default: false,
     },
+    name: Schema.Types.String,
+    identifier: Schema.Types.String,
+    kind: Schema.Types.String,
+    kindId: Schema.Types.ObjectId,
   },
   schemaOptions
 );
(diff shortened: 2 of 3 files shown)


libs/dal/src/repositories/integration/integration.entity.ts
```diff
@@ -48,6 +50,14 @@ export class IntegrationEntity {
 
   active: boolean;
 
+  name?: string;
+
+  identifier?: string;
+
+  kind?: ProviderKindEnum;
```
p-fernandez, line 57: Typing as `ProviderKindEnum` in case in the future we have `kinds` in other entities to help the readability and searcheability.

libs/dal/src/repositories/integration/types.ts
```diff
@@ -0,0 +1,7 @@
+export type KindId = string;
+
+export enum ProviderKindEnum {
+  ENVIRONMENT = 'environment',
+  TENANT = 'tenant',
+  WORKFLOW = 'workflow',
+}
```
p-fernandez, line 7: So far we will only use the environment kind but ready for the future.

libs/dal/src/repositories/integration/integration.schema.ts
```diff
@@ -45,11 +45,14 @@ const integrationSchema = new Schema<IntegrationDBModel>(
...
     },
-
     active: {
       type: Schema.Types.Boolean,
       default: false,
     },
+    name: Schema.Types.String,
+    identifier: Schema.Types.String,
+    kind: Schema.Types.String,
+    kindId: Schema.Types.ObjectId,
```
ainouzgali, line 55: Should we already add conditions? But as an empty array for now? 
To make it easier for next phase

libs/dal/src/repositories/integration/types.ts
ainouzgali: I'm probably jumping ahead - but maybe already add this type to shared ? As we would use it everywhere for sure

libs/dal/src/repositories/integration/integration.entity.ts
ainouzgali: I am not sure about 'kind'. What do you think about 'category' ? Both don't feel completely right🤷🏼‍♀️

ainouzgali (approved):
⭐️

libs/dal/src/repositories/integration/integration.entity.ts
scopsy: A small question regarding the `kind` property, if we are doing the "conditions" array, doesn't it mean that the kinds can be combined in a more dynamic way inside the the `conditions` array. Meaning that you can have the following dynamic conditions:
- TenantId = 123
- AND
- Environment = production
- AND
- Workflow = 123

So in this case it's either the "kind" or "conditions" approach, not the two of them combined. What are your thoughts on this, maybe I'm missing something

djabarovgeorge:
Not sure i followed how is the new `kind` param will be used :(

libs/dal/src/repositories/integration/integration.entity.ts
p-fernandez: @ainouzgali `kind` and `category` are synonims in a way. In the technical design I chose `kind` as it is used to define entities in places like GCP and I come from working a lot with GCP.

p-fernandez: > A small question regarding the `kind` property, if we are doing the "conditions" array, doesn't it mean that the kinds can be combined in a more dynamic way inside the the `conditions` array. Meaning that you can have the following dynamic conditions:
> 
> * TenantId = 123
> * AND
> * Environment = production
> * AND
> * Workflow = 123
> 
> So in this case it's either the "kind" or "conditions" approach, not the two of them combined. What are your thoughts on this, maybe I'm missing something

Ok, I wasn't aware that we totally ditched the `kind` property. Makes no sense if we are going to go `conditions` approach.

libs/dal/src/repositories/integration/types.ts
p-fernandez: Good point. But going to delete it as we do not need it until we create the `conditions` in the future.

libs/dal/src/repositories/integration/integration.schema.ts
p-fernandez: Better wait for now. If we then change our mind we might forget to clean it.

ainouzgali (approved):
🚀

p-fernandez:
> Oh @p-fernandez , a question. We now don't have a field to save the environment the user chooses for the instance. Screen Shot 2023-06-21 at 16 00 12

But that looks like a presentational thing. At the end we are storing the environmentId in the provider. I think.

PR #3629, opened 2023-06-20
NV 2459 - Get Tenant By ID Endpoint

djabarovgeorge:

Created a get endpoint + usecase

In order to fetch tenant data by id.

### DIFF
--- a/apps/api/src/app/tenant/e2e/get-tenant.e2e.ts
+++ b/apps/api/src/app/tenant/e2e/get-tenant.e2e.ts
@@ -1,0 +1,52 @@
+import { expect } from 'chai';
+import axios, { AxiosResponse } from 'axios';
+
+import { UserSession } from '@novu/testing';
+import { createTenant } from './create-tenant.e2e';
+
+describe('Get Tenant - /:tenantId (GET)', function () {
+  let session: UserSession;
+
+  beforeEach(async () => {
+    session = new UserSession();
+    await session.initialize();
+  });
+
+  it('should get a newly created tenant', async function () {
+    const res = await createTenant({
+      session,
+      identifier: 'identifier_123',
+      name: 'name_123',
+      data: { test1: 'test value1', test2: 'test value2' },
+    });
+
+    const getTenantResult = await getTenant({ session, id: res.data._id });
+
+    expect(getTenantResult.data.identifier).to.equal('identifier_123');
+    expect(getTenantResult.data.namd).to.equal('name_123');
+    expect(getTenantResult.data.data).to.deep.equal({ test1: 'test value1', test2: 'test value2' });
+  });
+
+  it('should throw exception if tenant does not existing', async function () {
+    const incorrectId = session.environment._id;
+    try {
+      await getTenant({ session, id: incorrectId });
+
+      throw new Error('');
+    } catch (e) {
+      expect(e.response.data.message).to.contains(
+        `Tenant with id: ${incorrectId} does not exist under environment ${session.environment._id}`
+      );
+    }
+  });
+});
+
+async function getTenant({ session, id }: { session; id: string }): Promise<AxiosResponse> {
+  const axiosInstance = axios.create();
+
+  return await axiosInstance.get(`${session.serverUrl}/v1/tenants/${id}`, {
+    headers: {
+      authorization: `ApiKey ${session.apiKey}`,
+    },
+  });
+}
--- a/apps/api/src/app/tenant/tenant.controller.ts
+++ b/apps/api/src/app/tenant/tenant.controller.ts
@@ -1,4 +1,13 @@
-import { Body, ClassSerializerInterceptor, Controller, Post, UseGuards, UseInterceptors } from '@nestjs/common';
+import {
+  Body,
+  ClassSerializerInterceptor,
+  Controller,
+  Get,
+  Param,
+  Post,
+  UseGuards,
+  UseInterceptors,
+} from '@nestjs/common';
 import { ApiExcludeController, ApiOperation, ApiTags } from '@nestjs/swagger';
 
 import { IJwtPayload } from '@novu/shared';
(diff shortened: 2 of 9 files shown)


apps/api/src/app/tenant/dtos/create-tenant-response.dto.ts
```diff
@@ -19,4 +19,7 @@ export class CreateTenantResponseDto {
 
   @ApiProperty()
   createdAt: string;
+
```
djabarovgeorge, line 22: Added missing attribute.

libs/dal/src/repositories/tenant/tenant.schema.ts
```diff
@@ -9,7 +9,6 @@ const tenantSchema = new Schema<TenantDBModel>(
     identifier: Schema.Types.String,
     name: Schema.Types.String,
     deleted: Schema.Types.Boolean,
-    createdAt: Schema.Types.String,
```
djabarovgeorge, line 12: Removed createdAt that was mistakenly added on the previous branch.

libs/dal/src/repositories/tenant/tenant.entity.ts
```diff
@@ -14,6 +14,8 @@ export class TenantEntity {
 
   createdAt: string;
 
+  updatedAt: string;
```
djabarovgeorge, line 17: Added missing attribute.

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -11,14 +20,34 @@ import { CreateTenantCommand } from './usecases/create-tenant/create-tenant.comm
...
+  constructor(private CreateTenantUsecase: CreateTenant, private getTenantUsecase: GetTenant) {}
+
+  @Get('/:tenantId')
+  @ApiResponse(GetTenantResponseDto)
+  @ApiOperation({
+    summary: 'Get tenant',
+    description: `Get tenant by your internal id used to identify the tenant`,
+  })
+  @ExternalApiAccessible()
+  getTenantById(@UserSession() user: IJwtPayload, @Param('tenantId') tenantId: string): Promise<GetTenantResponseDto> {
```
jainpawan21, line 42: @djabarovgeorge I have a doubt. here`tenantId` param can be any string or MongoDB id?

djabarovgeorge: This one should be Novu internal id so MongoDB Id. If the user will provide not MongoDB id. 
MongoDB will throw an exception. do you think we should add the validation on the controller level? (actually, I did not think about it because I have not seen yet that we make such validation on query params)

apps/api/src/app/tenant/usecases/get-tenant/get-tenant.command.ts
```diff
@@ -0,0 +1,9 @@
+import { EnvironmentCommand } from '@novu/application-generic';
+import { IsMongoId, IsNotEmpty, IsString } from 'class-validator';
+
+export class GetTenantCommand extends EnvironmentCommand {
+  @IsString()
+  @IsNotEmpty()
+  @IsMongoId()
+  id: string;
```
ainouzgali, line 8: What do you think about calling it tenantId here? So it remains clear when we use command.tenantId vs command.id in future uscases

ainouzgali (commented):
Looks great❤️ Left a few suggestions, let me know what you think

apps/api/src/app/tenant/e2e/get-tenant.e2e.ts
```diff
@@ -0,0 +1,52 @@
...
+    expect(getTenantResult.data.name).to.equal('name_123');
+    expect(getTenantResult.data.data).to.deep.equal({ test1: 'test value1', test2: 'test value2' });
+  });
+
+  it('should throw exception if tenant does not existing', async function () {
+    const incorrectId = session.environment._id;
+    try {
+      await getTenant({ session, id: incorrectId });
+
+      throw new Error('');
```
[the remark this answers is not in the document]
djabarovgeorge: I like the idea of not throwing an unnecessary exception(which is the worst), but this way the console won't print the issue/test that failed, which could make the debugging a bit more difficult, wdyt about it?

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -11,14 +20,34 @@ import { CreateTenantCommand } from './usecases/create-tenant/create-tenant.comm
...
+import { GetTenantCommand } from './usecases/get-tenant/get-tenant.command';
 
 @Controller('/tenants')
 @ApiTags('Tenants')
 @UseInterceptors(ClassSerializerInterceptor)
 @UseGuards(JwtAuthGuard)
 @ApiExcludeController()
 export class TenantController {
-  constructor(private CreateTenantUsecase: CreateTenant) {}
+  constructor(private CreateTenantUsecase: CreateTenant, private getTenantUsecase: GetTenant) {}
```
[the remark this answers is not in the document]
djabarovgeorge: good catch!
copy pasta mess 🧹

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -11,14 +20,34 @@ import { CreateTenantCommand } from './usecases/create-tenant/create-tenant.comm
...
+
+  @Get('/:tenantId')
+  @ApiResponse(GetTenantResponseDto)
+  @ApiOperation({
+    summary: 'Get tenant',
+    description: `Get tenant by your internal id used to identify the tenant`,
```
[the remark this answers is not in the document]
djabarovgeorge: You're right I already noticed this mess, I will refactor it to the external id instead.

apps/api/src/app/tenant/tenant.controller.ts
jainpawan21: yes we can create a dto for this and in dto use `@IsMongo()` decorator for this field

djabarovgeorge: After some discovery, we decided to use here the tenant identifier, so the user won't need to manage/work with Novu's internal ids.
So I will refactor it to the tenant identifier which is a string.

PR #3631, opened 2023-06-21
feat: change limit to 100 from 1000

davidsoderberg:
Change limit from 100 to 1000 for notification center usage

To reduce load added from the notification center requests


apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -80,23 +80,15 @@ export class GetNotificationsFeed {
...
-
     return {
       data: feed || [],
-      totalCount: totalCount || 0,
-      hasMore,
+      hasMore: feed.length < totalCount,
       pageSize: command.limit,
       page: command.page,
+      skip: command.page * command.limit,
+      limit: command.limit + 1,
```
djabarovgeorge, line 91: Those two should be in the `getCount` query options. and not in the return value.

Regarding the totalCount removal, I could love to do quick sync :)

apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
```diff
@@ -47,26 +47,15 @@ export class GetMessages {
...
 
     return {
       page: command.page,
-      totalCount,
-      hasMore,
+      hasMore: data.length < totalCount,
       pageSize: LIMIT,
       data,
+      skip: command.page * LIMIT,
+      limit: LIMIT + 1,
```
djabarovgeorge, line 58: Those two should be in the getCount query options. and not in the return value.

Regarding the totalCount removal, I could love to do quick sync :)

djabarovgeorge (commented):
Would love to sync about it as well, tried the coffee time so we get Dima as well but i am not sure if will make it 😅

apps/api/src/app/shared/dtos/pagination-response.ts
```diff
@@ -8,8 +8,9 @@ export class PaginatedResponseDto<T> {
 
   @ApiProperty({
     description: 'Total count of items matching the query',
+    required: false,
   })
-  totalCount: number;
+  totalCount?: number;
```
p-fernandez, line 13: What's the reason behind making it optional? Being an interface for paginated responses I do think that `totalCount` should be mandatory. Same for `skip` and `limit`.
I can see that the use case modified has a different type of response that includes `hasMore`.  Not sure when it was introduced but we might need to create a different type for `PaginatedResponseWithHasMore` if they have to share some of the functionalities.
Am I missing anything from the context?

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
djabarovgeorge, line 1: Looks great here :)

apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts
```diff
@@ -47,17 +47,7 @@ export class GetMessages {
...
-      COUNT_LIMIT
-      /*
-       * todo NV-2161 in version 0.16:
-       *  update option as below,
-       *  update below:  hasMore = feed.length < totalCount
-       *  remove totalCount
-       * { skip: command.page * LIMIT, limit: LIMIT + 1 }
-       */
-    );
+    const totalCount = await this.messageRepository.count(query, COUNT_LIMIT);
```
djabarovgeorge, line 50: I wonder if we should remove the count limit here because it is not used by the notification-center

apps/api/src/app/subscribers/subscribers.controller.ts
```diff
@@ -385,8 +385,7 @@ export class SubscribersController {
     @Query('feedIdentifier') feedId: string[] | string,
     @Query('seen') seen: boolean,
     @Param('subscriberId') subscriberId: string,
-    // todo NV-2161 update DefaultValuePipe to 100 in version 0.16
-    @Query('limit', new DefaultValuePipe(1000), new LimitPipe(1, 1000, true)) limit: number
+    @Query('limit', new DefaultValuePipe(100), new LimitPipe(1, 100, true)) limit: number
```
djabarovgeorge, line 388: I am thinking about maybe it will be a good idea to remove all those LimitPipes on the count endpoint (unsee/unread etc as well) so that the user will have the ability to query the count if needed.

I saw that the notification center is using '/widgets/notifications/unseen' endpoint in useUnseenCount, so maybe we could still use the default for the users that use custom UI but i am not sure about the max limit there anymore.

wdyt?

djabarovgeorge (commented):
Looks good left a couple of comments there.

libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -91,7 +91,7 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
     subscriberId: string,
     channel: ChannelTypeEnum,
     query: { feedId?: string[]; seen?: boolean; read?: boolean } = {},
-    options: { limit: number } = { limit: 1000 } // todo NV-2161 update to 100 in version 0.16
+    options: { limit: number; skip?: number } = { limit: 100, skip: 0 }
```
p-fernandez, line 94: We need to start to centralise these in shared constants so we avoid to miss them when updating. One for the tech debt.

p-fernandez (approved):

apps/api/src/app/subscribers/subscribers.controller.ts
davidsoderberg: I agree with you :)

davidsoderberg: I think this is a matter of the endpoint being copied from the widgets controller

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -79,22 +81,13 @@ export class GetNotificationsFeed {
...
       },
-      { limit: COUNT_LIMIT }
-      /*
-       * todo NV-2161 in version 0.16
-       *  update option as below,
-       *  update below:  hasMore = feed.length < totalCount
-       *  remove totalCount
-       * { skip: command.page * command.limit, limit: command.limit + 1 }
-       */
+      { limit: COUNT_LIMIT, skip }
```
djabarovgeorge, line 84: small question, I think we have here a small loophole,
if the user will provide a `limit` bigger than 100 for example 1000 (which is allowed, I don't fount any limitation to it) then:
* The total count won't make sense if the feed length is 900 but the total count is 100.
* The hasMore will be falsy, if the feed length will be 1000 but the total count will be 100, the hasMore will be false but the user could have more notifications. 

Let me know what you think, maybe a missed something.

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -79,22 +81,13 @@ export class GetNotificationsFeed {
...
     );
 
-    const hasMore = this.getHasMore(command.page, command.limit, feed, totalCount);
-
     return {
       data: feed || [],
-      totalCount: totalCount || 0,
-      hasMore,
+      totalCount: Math.min(skip + totalCount, 100),
+      hasMore: feed.length < totalCount,
```
djabarovgeorge, line 90: Could we get here `feed=null`?

Maybe we can add else on 
`    if (feed.length) `
and do return there, one request less to the db 🫣

djabarovgeorge (commented):
Yeah looks good, left a couple of comments on there regarding things i was not sure about.

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
davidsoderberg: Ah yes, should we limit the limit to 100 or do you have a suggestion on how we could solve this?

djabarovgeorge: It could be a possible solution i am just not sure regarding the product part of this change, `should we support a higher limit or not`.
Maybe we could do it optional to update to 100/1000 (but so they both limit and count limit would overlap)

davidsoderberg: so use `command.limit` instead of `COUNT_LIMIT`?

djabarovgeorge: Generally yes 🙃, but I wonder if it is ok, product-wise. wdyt?

davidsoderberg: as long as we have skip as well in the query I do not see any problem with it...

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -70,31 +68,27 @@ export class GetNotificationsFeed {
...
-       *  update below:  hasMore = feed.length < totalCount
-       *  remove totalCount
-       * { skip: command.page * command.limit, limit: command.limit + 1 }
-       */
-    );
+    const skip = command.page * command.limit;
+    let totalCount = 0;
 
-    const hasMore = this.getHasMore(command.page, command.limit, feed, totalCount);
+    if (feed.length) {
```
djabarovgeorge, line 74:
```suggestion

    if (feed.length) {
      totalCount = await this.messageRepository.getCount(
        command.environmentId,
        subscriber._id,
        ChannelTypeEnum.IN_APP,
        {
          feedId: command.feedId,
          seen: command.query.seen,
          read: command.query.read,
        },
        { limit: command.limit + 1, skip }
      );
    }

    const hasMore = feed.length < totalCount;
    totalCount = Math.min(totalCount, command.limit);

    return {
      data: feed || [],
      totalCount: totalCount,
      hasMore: hasMore,
      pageSize: command.limit,
      page: command.page,
    };
```

i think this could help to fixing the hasMore flag.
@davidsoderberg

apps/api/src/app/widgets/usecases/get-feed-count/get-feed-count.command.ts
```diff
@@ -16,8 +16,7 @@ export class GetFeedCountCommand extends EnvironmentWithSubscriber {
   @IsOptional()
   @Transform(({ value }) => {
     if (isNaN(value) || value == null) {
-      // todo NV-2161 update the limit default to 100 to in version 0.16
-      return 1000;
+      return 100;
```
p-fernandez, line 19: We should centralize the selected values and reduce the hardcodes. 🙏🏻

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -70,31 +68,51 @@ export class GetNotificationsFeed {
...
+        subscriber._id,
+        ChannelTypeEnum.IN_APP,
+        {
+          feedId: command.feedId,
+          seen: command.query.seen,
+          read: command.query.read,
+        },
+        { limit: command.limit, skip }
+      );
+    }
```
p-fernandez, line 110: This code doesn't look right.

p-fernandez (changes requested):
Just the bit of code that looks off. The limit for a tech debt.

p-fernandez (approved):
🌟

PR #3635, opened 2023-06-21
docs: Add Termii provider documentation

peoray:
Add Termii provider documentation


p-fernandez (approved):
Brilliant as always. 🌟 
I am assuming the information in the screenshots are API keys from a throwaway account or a free one you are not using.

PR #3638, opened 2023-06-21
Nv 2460 - Update Tenant Endpoint

djabarovgeorge:

Adds updated tenant metadata.

So the user's tenants will be updated.

### DIFF
--- a/apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
+++ b/apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
@@ -1,13 +1,19 @@
-import { ApiProperty } from '@nestjs/swagger';
+import { ApiPropertyOptional } from '@nestjs/swagger';
 import { TenantCustomData } from '@novu/shared';
+import { IsOptional, IsString } from 'class-validator';
 
 export class UpdateTenantRequestDto {
-  @ApiProperty()
+  @IsOptional()
+  @IsString()
+  @ApiPropertyOptional({ type: String })
   identifier?: string;
 
-  @ApiProperty()
+  @IsOptional()
+  @IsString()
+  @ApiPropertyOptional({ type: String })
   name?: string;
 
-  @ApiProperty()
+  @IsOptional()
+  @ApiPropertyOptional()
   data?: TenantCustomData;
 }
--- a/apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
+++ b/apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
@@ -1,0 +1,13 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { TenantCustomData } from '@novu/shared';
+
+export class UpdateTenantRequestDto {
+  @ApiProperty()
+  identifier?: string;
+
+  @ApiProperty()
+  name?: string;
+
+  @ApiProperty()
+  data?: TenantCustomData;
+}
--- a/apps/api/src/app/tenant/dtos/update-tenant-response.dto.ts
+++ b/apps/api/src/app/tenant/dtos/update-tenant-response.dto.ts
@@ -1,0 +1,25 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { TenantCustomData } from '@novu/shared';
+
+export class UpdateTenantResponseDto {
+  @ApiProperty()
+  _id: string;
+
+  @ApiProperty()
+  identifier: string;
+
+  @ApiProperty()
+  name?: string;
+
+  @ApiProperty()
+  data?: TenantCustomData;
+
+  @ApiProperty()
+  _environmentId: string;
+
+  @ApiProperty()
+  createdAt: string;
+
+  @ApiProperty()
+  updatedAt: string;
+}
--- a/apps/api/src/app/tenant/dtos/update-tenant-response.dto.ts
+++ b/apps/api/src/app/tenant/dtos/update-tenant-response.dto.ts
@@ -1,25 +1,32 @@
-import { ApiProperty } from '@nestjs/swagger';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { TenantCustomData } from '@novu/shared';
+import { IsString } from 'class-validator';
 
 export class UpdateTenantResponseDto {
-  @ApiProperty()
+  @ApiProperty({ type: String })
+  @IsString()
   _id: string;
 
-  @ApiProperty()
+  @ApiProperty({ type: String })
+  @IsString()
   identifier: string;
 
-  @ApiProperty()
+  @ApiPropertyOptional({ type: String })
+  @IsString()
   name?: string;
 
-  @ApiProperty()
+  @ApiPropertyOptional()
   data?: TenantCustomData;
 
-  @ApiProperty()
+  @ApiProperty({ type: String })
+  @IsString()
   _environmentId: string;
 
-  @ApiProperty()
+  @ApiProperty({ type: String })
+  @IsString()
   createdAt: string;
 
-  @ApiProperty()
+  @ApiProperty({ type: String })
+  @IsString()
   updatedAt: string;
 }
--- a/apps/api/src/app/tenant/tenant.controller.ts
+++ b/apps/api/src/app/tenant/tenant.controller.ts
@@ -5,6 +5,7 @@ import {
   Get,
   Param,
   Post,
+  Put,
   UseGuards,
   UseInterceptors,
 } from '@nestjs/common';
@@ -23,6 +24,10 @@ import { ApiResponse } from '../shared/framework/response.decorator';
 import { GetTenantResponseDto } from './dtos/get-tenant.response';
 import { GetTenant } from './usecases/get-tenant/get-tenant.usecase';
 import { GetTenantCommand } from './usecases/get-tenant/get-tenant.command';
+import { UpdateTenantResponseDto } from './dtos/update-tenant-response.dto';
+import { UpdateTenantRequestDto } from './dtos/update-tenant-request.dto';
+import { UpdateTenant } from './usecases/update-tenant/update-tenant.usecase';
+import { UpdateTenantCommand } from './usecases/update-tenant/update-tenant.command';
 
 @Controller('/tenants')
 @ApiTags('Tenants')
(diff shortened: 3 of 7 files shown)


apps/api/src/app/tenant/usecases/update-tenant/update-tenant.command.ts
```diff
@@ -0,0 +1,20 @@
...
+import { IsNotEmpty, IsOptional, IsString } from 'class-validator';
+
+export class UpdateTenantCommand extends EnvironmentCommand {
+  @IsString()
+  @IsNotEmpty()
+  identifier: string;
+
+  @IsString()
+  @IsOptional()
+  newIdentifier?: string;
```
djabarovgeorge, line 12: not sure what will be the best name here, but I added the ability to update the identifier from the start because we all know that it is popular to be a `formally 'old company name'` / `previously known as 'old company name'`.
if someone can think of a drawback Speak Now or Forever Hold Your Peace (till the refactor)

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (command.name != null) {
+      updatePayload.name = command.name;
+    }
+
+    if (command.data != null) {
+      updatePayload.data = command.data;
+    }
+
+    if (command.newIdentifier != null && command.newIdentifier !== command.identifier) {
+      await this.validateIdentifierDuplication({
```
djabarovgeorge, line 33: Because it is an identifier that can not be duplicated we throw an error if the user present duplication.

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+        identifier: command.identifier,
+        _id: tenant._id,
+      },
+      {
+        $set: updatePayload,
+      }
+    );
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return (await this.tenantRepository.findOne({
```
djabarovgeorge, line 54: Can not be null because we fetched it in the beginning.

apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export class UpdateTenantRequestDto {
+  @ApiProperty()
+  identifier?: string;
+
+  @ApiProperty()
+  name?: string;
+
+  @ApiProperty()
+  data?: TenantCustomData;
```
LetItRock, line 12: all these are optional? if yes then use `@ApiPropertyOptional`

apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
```diff
@@ -0,0 +1,13 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { TenantCustomData } from '@novu/shared';
+
+export class UpdateTenantRequestDto {
+  @ApiProperty()
+  identifier?: string;
+
+  @ApiProperty()
+  name?: string;
```
LetItRock, line 9: if these are strings please use ` @IsString()`

apps/api/src/app/tenant/dtos/update-tenant-response.dto.ts
```diff
@@ -0,0 +1,25 @@
...
+  data?: TenantCustomData;
+
+  @ApiProperty()
+  _environmentId: string;
+
+  @ApiProperty()
+  createdAt: string;
+
+  @ApiProperty()
+  updatedAt: string;
```
LetItRock, line 24: please provide the types on the `@ApiProperty` annotation

apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
LetItRock: and please provide the types on the `@ApiPropertyOptional`, check the `apps/api/src/app/topics/dtos/filter-topics.dto.ts` as an example

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  async execute(command: UpdateTenantCommand): Promise<TenantEntity> {
+    const tenant = await this.tenantRepository.findOne({
+      _environmentId: command.environmentId,
+      identifier: command.identifier,
+    });
+
+    if (!tenant) {
+      throw new BadRequestException(
+        `Tenant with identifier: ${command.identifier} is not exists under environment ${command.environmentId}`
```
LetItRock, line 18:
```suggestion
        `Tenant with identifier: ${command.identifier} doesn't exist under environment ${command.environmentId}`
```

apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
LetItRock: if `name` cannot be empty string please use `@IsNotEmpty`

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const updatePayload: Partial<TenantEntity> = {};
+
+    if (command.name != null) {
+      updatePayload.name = command.name;
+    }
+
+    if (command.data != null) {
+      updatePayload.data = command.data;
+    }
```
LetItRock, line 30:
```suggestion
    if (command.name) {
      updatePayload.name = command.name;
    }

    if (command.data) {
      updatePayload.data = command.data;
    }
```

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    if (command.newIdentifier != null && command.newIdentifier !== command.identifier) {
```
LetItRock, line 32:
```suggestion
    if (command?.newIdentifier !== command.identifier) {
```

apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
djabarovgeorge: > 

Yes the user can decide what he wants to update,

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
djabarovgeorge: here we need to make sure that `command.newIdentifier` is not undefined as well, and then check if they are not equal, because of the following case:
`command?.newIdentifier !== command.identifier` could be  `undefine !== 'current_identifier'` and the if will the true but i am not sure we will want to update it with `undefined`

apps/api/src/app/tenant/dtos/update-tenant-request.dto.ts
djabarovgeorge: > and please provide the types on the `@ApiPropertyOptional`, check the `apps/api/src/app/topics/dtos/filter-topics.dto.ts` as an example

I was not sure how to type `data: TenantCustomData` i thought maybe converting it to class or about something like:
  {
    type: Record,
    additionalProperties: {
      oneOf: [
        // { type: 'string' },
        { type: 'number' },
        { type: 'boolean' },
        { type: 'undefined' },
        {
          type: 'array',
          items: {
            type: 'string',
          },
        },
      ],
    },
  }
  
  are you aware of any good ways to define it?

LetItRock: we don't require any specific shape restrictions for this field apart from it should be an object, so I would say that what you have rn is ok

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.command.ts
LetItRock: We don't have to require sending the old `identifier` and the `newIdentifier`, but just one field. 
Also, the validation logic on this field in the use-case should be fixed, because it's possible even to skip the validation if I'll send the same value in both fields.

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
LetItRock: you should compare the `indentifier` from the database with a new value from command

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -73,4 +82,28 @@ export class TenantController {
...
+  @Put('/:tenantId')
+  @ExternalApiAccessible()
+  @ApiResponse(UpdateTenantResponseDto)
+  @ApiOperation({
+    summary: 'Update tenant',
+    description: 'Update tenant under the current environment',
+  })
+  async updateTenant(
+    @UserSession() user: IJwtPayload,
+    @Param('tenantId') tenantId: string,
```
LetItRock, line 95: it should be `identifier`

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.command.ts
djabarovgeorge: I presented both of them, in order to provide the ability to update identifiers on our end, how can we update the identifier without both of them? perhaps i messed up something?

It is expected to skip the validation because if the user has `identifier : evil-corp` and the new update payload has the `identifier : evil-corp`  which is the same identifier, i don't see a reason to update it or make an unnecessary request to the DB.

apps/api/src/app/tenant/tenant.controller.ts
djabarovgeorge: Totally agree already update in the next PR 🙏

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
djabarovgeorge: It should be the same if it was found in the query but would be better anyway. 👍

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.command.ts
LetItRock: yeah, I forgot that we are updating with the tenant identifier in the URL and looking for that tenant in db by identifier

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
LetItRock: This TS statement `command?.newIdentifier !== command.identifier` is compiled to JS like `command?.newIdentifier && command.newIdentifier !== command.identifier`.
So for the `undefined` case it will be like `undefined && command.newIdentifier !== command.identifier`, in the result `false`.

djabarovgeorge: Added a dedicated test to check the null/undefined updates just in case.

PR #3639, opened 2023-06-21
NV 2461 - Delete Tenant Endpoint

djabarovgeorge:

Adds tenant delete use case.

In order to provide the ability to delete redundant tenants from the platform.


apps/api/src/app/tenant/dtos/delete-tenant.response.dto.ts
djabarovgeorge, line 1: Duplicated this one from another class, i wonder if we need to create a more common one for all the deletes or to create a shared interface for those DTOs.

apps/api/src/app/tenant/e2e/get-tenant.e2e.ts
```diff
@@ -4,7 +4,7 @@ import axios, { AxiosResponse } from 'axios';
 import { UserSession } from '@novu/testing';
 import { TenantRepository } from '@novu/dal';
 
-describe('Get Tenant - /:tenantId (GET)', function () {
+describe('Get Tenant - /tenants/:identifier (GET)', function () {
```
djabarovgeorge, line 7: typo

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -83,21 +88,21 @@ export class TenantController {
     );
   }
 
-  @Put('/:tenantId')
+  @Put('/:identifier')
```
djabarovgeorge, line 91: I think that 'identifier' would be more intuitive in this context.

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -106,4 +111,25 @@ export class TenantController {
...
+
+  @Delete('/:identifier')
+  @ExternalApiAccessible()
+  @UseGuards(JwtAuthGuard)
+  @ApiResponse(DeleteTenantResponseDto)
+  @ApiOperation({
+    summary: 'Delete tenant',
+    description: 'Deletes a tenant entity from the Novu platform',
+  })
+  async removeTenant(
```
LetItRock, line 123: Let's be more consistent and make this endpoint similar to what we have in the `topics` controller, I mean the response and the status.

djabarovgeorge: thanks love the suggestion, will add it for sure.

LetItRock: you are still returning the response but the status is "no content" ;)

djabarovgeorge: actually the `@HttpCode(HttpStatus.NO_CONTENT)` do the heavy lifting and removed the data from the response but there were definitely leftovers on the flow, remove the unnecessary remaining code.

LetItRock (approved):

PR #3640, opened 2023-06-21
fix(api): expose api keys only for current environment

### LINKED ISSUE #3490: [NV-2380] bug: user can access prod environment using dev api key
### Discussed in https://github.com/orgs/novuhq/discussions/3483

<div type='discussions-op-text'>

<sup>Originally posted by **leonardbinet** May 23, 2023</sup>
Hi there 👋,

Given 2 environments, `dev` and `production` (default), each environment has its dedicated api key.

I've noticed that by calling: (https://docs.novu.co/api/get-environments/)
```
curl -X GET \
 -H "Authorization: ApiKey <dev api key>" \
https://api.novu.co/v1/environments
```

I would get all environments info, with related api keys (both `production` and `dev`) 🙀 

Long story short, with `dev` api key I get access to `production` environment.

It seems to me like a serious security flaw, why would you want api keys to be returned in this case?</div>

<sub>[NV-2380](https://linear.app/novu/issue/NV-2380/bug-user-can-access-prod-environment-using-dev-api-key)</sub>

michaldziuba03:
This PR introduce security fix, because currently our API expose production keys in `/v1/environments` when using development API key.

Challenging part was making it without breaking SDKs and clients -  my goal is to fix this problem without changing response schema :) To be honest if it could be possible I would remove from response this field (`apiKeys`) at all.

The way how this endpoint works after fix: 
- it still returns array of environments
- each environment still has `apiKeys` array
- if environment is current env, then `apiKeys` contains data, otherwise `apiKeys` returns empty array

To improve security.

Closes #3490

Related discussion: https://github.com/novuhq/novu/discussions/3483


apps/api/src/app/environments/dtos/environment-response.dto.ts
```diff
@@ -1,7 +1,6 @@
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
-import { IApiKey, IWidgetSettings } from '@novu/dal';
+import { IApiKey } from '@novu/dal';
 import { ApiKey } from '../../shared/dtos/api-key';
-import { WidgetSettings } from '../../shared/dtos/widget-settings';
```
michaldziuba03, line 4: I just removed some old unused imports

apps/api/src/app/environments/environments.controller.ts
```diff
@@ -88,6 +88,7 @@ export class EnvironmentsController {
   async getMyEnvironments(@UserSession() user: IJwtPayload): Promise<EnvironmentResponseDto[]> {
     return await this.getMyEnvironmentsUsecase.execute(
       GetMyEnvironmentsCommand.create({
+        environmentId: user.environmentId,
```
michaldziuba03, line 91: we need access to environmentId of current environment

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -11,6 +11,17 @@ export class GetMyEnvironments {
...
+    const environments = await this.environmentRepository.findOrganizationEnvironments(command.organizationId);
+
+    return environments.map((environment) => {
+      if (environment._id === command.environmentId) {
+        return environment;
+      }
+
+      return {
+        ...environment,
+        apiKeys: [],
```
michaldziuba03, line 23: we don't want to expose api keys from other environments :)

djabarovgeorge (commented):
Looks good, could we add small tests in order to validate that we won't expose api keys from other environments in the future?

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -18,10 +18,9 @@ export class GetMyEnvironments {
         return environment;
       }
 
-      return {
-        ...environment,
-        apiKeys: [],
-      };
+      environment.apiKeys = [];
```
michaldziuba03, line 21: I made change to code to make sure the class instance is preserved.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.e2e.ts
```diff
@@ -15,6 +15,12 @@ describe('Get My Environments - /environments (GET)', async () => {
     expect(body.data.length).to.be.greaterThanOrEqual(2);
     for (const elem of body.data) {
       expect(elem._organizationId).to.eq(session.organization._id);
+
+      if (elem._id !== session.environment._id) {
```
michaldziuba03, line 19: @djabarovgeorge I added two assertions in test for use case. I think it's enough as test itself is simple.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -11,6 +11,16 @@ export class GetMyEnvironments {
   async execute(command: GetMyEnvironmentsCommand): Promise<EnvironmentEntity[]> {
     Logger.verbose('Getting Environments');
 
-    return await this.environmentRepository.findOrganizationEnvironments(command.organizationId);
+    const environments = await this.environmentRepository.findOrganizationEnvironments(command.organizationId);
+
+    return environments.map((environment) => {
+      if (environment._id === command.environmentId) {
+        return environment;
+      }
```
djabarovgeorge, line 19: Sorry for raising it now and not before but if we already have the environment id here maybe we could do `environmentRepository.findById` this way we won't need to iterate over a list. (reduce some of the complexity), once again apologies for raising it now and not before 🙏

djabarovgeorge (commented):
Looks great to me, I small question there, other than that I would love another team member to look at this PR and see if we did not miss anything.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
michaldziuba03: @djabarovgeorge we shouldn't because this endpoint is still supposed to return ALL environments for specific organization, not just one with specific id. This `.map()` just transforms other environments than "current" to hide api keys field.

djabarovgeorge: Oh right right, sorry missed this one now. 😅

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.command.ts
```diff
@@ -1,3 +1,8 @@
+import { IsDefined, IsMongoId } from 'class-validator';
 import { OrganizationCommand } from '../../../shared/commands/organization.command';
 
-export class GetMyEnvironmentsCommand extends OrganizationCommand {}
+export class GetMyEnvironmentsCommand extends OrganizationCommand {
```
BiswaViraj, line 4: I think you can use the existing `EnvironmentWithUserCommand` here instead of declaring `environmentId` manually here.

michaldziuba03: I agree, check my last change

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.command.ts
```diff
@@ -1,8 +1,3 @@
-import { IsDefined, IsMongoId } from 'class-validator';
-import { OrganizationCommand } from '../../../shared/commands/organization.command';
+import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
 
-export class GetMyEnvironmentsCommand extends OrganizationCommand {
-  @IsDefined()
-  @IsMongoId()
-  environmentId: string;
-}
+export class GetMyEnvironmentsCommand extends EnvironmentWithUserCommand {}
```
michaldziuba03, line 3: extend `EnvironmentWithUserCommand` as @BiswaViraj suggested in review

BiswaViraj (approved):

PR #3641, opened 2023-06-21
NV 2466 - Get Tenant List

djabarovgeorge:

Adds get tenants lists use case.

So the user could browse the data.

### DIFF
--- a/apps/api/src/app/tenant/e2e/get-tenants.e2e.ts
+++ b/apps/api/src/app/tenant/e2e/get-tenants.e2e.ts
@@ -1,0 +1,81 @@
+import { expect } from 'chai';
+import axios, { AxiosResponse } from 'axios';
+
+import { UserSession } from '@novu/testing';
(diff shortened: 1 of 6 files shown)


apps/api/src/app/tenant/e2e/get-tenants.e2e.ts
```diff
@@ -0,0 +1,81 @@
...
+    const { data } = getTenantResult;
+
+    expect(data.page).to.equal(1);
+    expect(data.pageSize).to.equal(5);
+    expect(data.totalCount).to.equal(9);
+    expect(data.data.length).to.equal(4);
+    expect(data.data[0].identifier).to.equal('identifier_5');
+    expect(data.data[3].identifier).to.equal('identifier_8');
+  });
+});
```
LetItRock, line 59: we should have tests that do validate the pagination properties like page number, limit

LetItRock (approved):

PR #3642, opened 2023-06-21
fix: notification template dual state

### LINKED ISSUE #3386: [NV-2292] 🐛 Bug Report: dual active states in UI, if template is created with api

michaldziuba03:
It fixes bug where it was possible to create notification template with both `enabled` and `disabled` states at the same time.

screenshot

Something like this is only possible when template is created using API or SDK.

Bugs happen when `active` and `draft` are equal, so I decided to deprecate `draft` field in DTO. After my changes `draft` is always set as opposite of `active` value provided by the user (this is exactly what frontend does but SDK users can set `draft` value however they want). I also fixed this problem on frontend side where I removed `draft` from render conditions.

Closes #3386


apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -3,7 +3,7 @@ import * as Sentry from '@sentry/node';
 import * as hat from 'hat';
 import { merge } from 'lodash';
 import { v4 as uuidv4 } from 'uuid';
-import { AnalyticsService, Instrument, InstrumentUsecase } from '@novu/application-generic';
```
michaldziuba03, line 6: some unused service

apps/api/src/app/notification-template/dto/create-notification-template.request.dto.ts
```diff
@@ -42,7 +42,7 @@ export class CreateNotificationTemplateRequestDto implements ICreateNotification
   @IsOptional()
   active?: boolean;
 
-  @ApiPropertyOptional()
+  @ApiPropertyOptional({ deprecated: true })
```
michaldziuba03, line 45: mark as deprecated in swagger

apps/api/src/app/notification-template/notification-template.controller.ts
```diff
@@ -174,7 +174,7 @@ export class NotificationTemplateController {
         steps: body.steps,
         notificationGroupId: body.notificationGroupId,
         active: body.active ?? false,
-        draft: body.draft ?? true,
+        draft: !body.active,
```
michaldziuba03, line 177: always set `draft` as opposite of `value`. Ignore deprecated `body.draft`.

This is exactly what frontend does - but SDK users can set draft/active however they want what may lead them to bugs.

apps/api/src/app/notification-template/usecases/change-template-active-status/change-template-active-status.usecase.ts
```diff
@@ -55,7 +55,7 @@ export class ChangeTemplateActiveStatus {
       {
         $set: {
           active: command.active,
-          draft: command.active === false,
+          draft: !command.active,
```
michaldziuba03, line 58: just shorter version suggested by editor

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -56,7 +55,7 @@ export class ParseEventRequest {
       throw new UnprocessableEntityException('workflow_not_found');
     }
 
-    if (!template.active || template.draft) {
```
michaldziuba03, line 59: after change users may still have those "broken" templates with active=true and draft=true.

UI after my changes will show that template is active, but API will be unable to trigger event becasue `draft` is true.

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -71,7 +71,7 @@ const columns: IExtendedColumn<INotificationTemplateExtended>[] = [
     maxWidth: 125,
     Cell: withCellLoading(({ row: { original } }) => (
       <>
-        {original.draft ? (
+        {!original.active ? (
```
michaldziuba03, line 74: after change users may still have those "broken" templates with active=true and draft=true.

so without this change on frontned they could still see dual state :)

apps/api/src/app/notification-template/dto/update-notification-template-request.dto.ts
```diff
@@ -37,9 +37,6 @@ export class UpdateNotificationTemplateRequestDto implements ICreateNotification
   @IsMongoId()
   notificationGroupId: string;
 
-  @ApiPropertyOptional()
-  active?: boolean;
```
michaldziuba03, line 41: this property from DTO is not even passed to the usecase in controller.

Application has seperated endpoint to update status

scopsy (approved):
Fantastic work @michaldziuba03!

PR #3643, opened 2023-06-22
feat: add removeAllMessages function in react nc

jainpawan21:
- Add `useRemoveAllNotification` hook
- Add `removeAllMessages` function in `useNotifications` hook
- Update API reference docs for above

### DIFF
--- a/docs/docs/notification-center/react/api-reference.md
+++ b/docs/docs/notification-center/react/api-reference.md
@@ -388,7 +388,7 @@ interface INotificationsContext {
 | markAllNotificationsAsRead     | function                                                    | The function allowing you to mark all notifications as read.                                                                  |
 | markAllNotificationsAsSeen     | function                                                    | The function allowing you to mark notifications as seen.                                                                      |
 | removeMessage                  | function                                                    | The function allowing you to delete a notification.                                                                           |
-| removeAllMessages              | function                                                    | The function allowing you to delete all notifications or a specific feed's notifications at once.                      |
+| removeAllMessages              | function                                                    | The function allowing you to delete all notifications or a specific feed's notifications at once.                             |
 
 You can find more information about the `IMessage` interface [here](./api-reference#the-notification-imessage-model).
 
--- a/docs/docs/notification-center/react/api-reference.md
+++ b/docs/docs/notification-center/react/api-reference.md
@@ -388,7 +388,7 @@ interface INotificationsContext {
 | markAllNotificationsAsRead     | function                                                    | The function allowing you to mark all notifications as read.                                                                  |
 | markAllNotificationsAsSeen     | function                                                    | The function allowing you to mark notifications as seen.                                                                      |
 | removeMessage                  | function                                                    | The function allowing you to delete a notification.                                                                           |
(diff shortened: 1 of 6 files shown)


packages/notification-center/src/hooks/useRemoveAllNotifications.ts
```diff
@@ -0,0 +1,35 @@
+import { useMutation, useQueryClient, UseMutationOptions } from '@tanstack/react-query';
+import type { IMessage } from '@novu/shared';
+
+import { useNovuContext } from './useNovuContext';
+import { useFetchNotificationsQueryKey } from './useFetchNotificationsQueryKey';
+
+interface IRemoveAllNotificationsVariables {
```
p-fernandez, line 7:
```suggestion
interface IRemoveAllNotificationsPayload {
```

packages/notification-center/src/hooks/useRemoveAllNotifications.ts
```diff
@@ -0,0 +1,35 @@
...
+  feedId?: string;
```
p-fernandez, line 8: If we make this optional then the mutation code should be prepared for it, checking if the `feedId` is being set and executing the mutation if so (line 23).

packages/client/src/api/api.service.ts
```diff
@@ -70,6 +70,14 @@ export class ApiService {
     return await this.httpClient.delete(`/widgets/messages/${messageId}`, {});
   }
 
+  async removeAllMessages(feedId?: string): Promise<any> {
+    const url = feedId
+      ? `/widgets/messages?feedId=${feedId}`
+      : `/widgets/messages`;
```
p-fernandez, line 76: This is dangerous, to allow to delete all messages if a feedId is missed for any reason.

davidsoderberg (approved):
Looks good to me :)

p-fernandez (commented):
I am missing the context of what needs to be done and why to be able to have a full input on this as at first sight there are 2 functionalities implemented: removing all notifications and removing all notifications from a feed. Could you provide more context? 🙏🏻

packages/client/src/api/api.service.ts
jainpawan21: feedId is optional, user can remove either feed specific message or all messages

docs/docs/notification-center/react/api-reference.md
```diff
@@ -387,6 +388,7 @@ interface INotificationsContext {
 | markAllNotificationsAsRead     | function                                                    | The function allowing you to mark all notifications as read.                                                                  |
 | markAllNotificationsAsSeen     | function                                                    | The function allowing you to mark notifications as seen.                                                                      |
 | removeMessage                  | function                                                    | The function allowing you to delete a notification.                                                                           |
+| removeAllMessages              | function                                                    | The function allowing you to delete all feeds notifications or specific feed's all notification at once.                      |
```
p-fernandez, line 391:
```suggestion
| removeAllMessages              | function                                                    | The function allowing you to delete all notifications or a specific feed's notifications at once.                      |
```

p-fernandez (approved):

PR #3647, opened 2023-06-22
Add IP Pool Name In Integration

djabarovgeorge:

Add ip pool name into the send grid integration.

in order to provide the user the ability to add/update it.

i refactored(moved to shared) the ICredentials by mistake we had too many duplications of it mostly mine fult.
i created an ICredentials in shared.
deleted some of the duplications.
currently created `type` of the interface because they were the same, it this type will need to change we change to update it to interface and `implement` the shared ICredentials.

### DIFF
--- a/apps/api/src/app/integrations/dtos/credentials.dto.ts
+++ b/apps/api/src/app/integrations/dtos/credentials.dto.ts
@@ -123,7 +123,7 @@ export class CredentialsDto implements ICredentials {
   redirectUrl?: string;
 
   @ApiPropertyOptional()
-  @IsString()
+  @IsBoolean()
   @IsOptional()
   hmac?: boolean;
 
--- a/apps/api/src/app/integrations/dtos/credentials.dto.ts
+++ b/apps/api/src/app/integrations/dtos/credentials.dto.ts
@@ -1,8 +1,9 @@
 import { ApiPropertyOptional } from '@nestjs/swagger';
 import { IsBoolean, IsObject, IsOptional, IsString } from 'class-validator';
 import { TransformToBoolean } from '../../shared/transformers/to-boolean';
+import { ICredentials } from '@novu/shared';
 
-export class CredentialsDto {
+export class CredentialsDto implements ICredentials {
   @ApiPropertyOptional()
   @IsString()
   @IsOptional()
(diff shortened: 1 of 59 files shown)


apps/api/src/app/integrations/dtos/credentials.dto.ts
```diff
@@ -115,4 +116,24 @@ export class CredentialsDto {
   @IsString()
   @IsOptional()
   webhookUrl?: string;
+
```
djabarovgeorge, line 119: missing attributes that i located due the `implements ICredentials` (in a hacky way)

apps/web/src/pages/integrations/IntegrationsStoreModal.tsx
```diff
@@ -328,32 +329,6 @@ export interface IIntegratedProvider {
   novu?: boolean;
 }
 
-export interface ICredentials {
```
djabarovgeorge, line 331: 🧹

apps/web/src/pages/integrations/IntegrationsStorePage.tsx
```diff
@@ -152,34 +153,6 @@ export interface IIntegratedProvider {
   novu?: boolean;
 }
 
-export interface ICredentials {
```
djabarovgeorge, line 155: 🧹

libs/shared/src/dto/integration/construct-integration.interface.ts
```diff
@@ -1,30 +1,6 @@
-export interface ICredentialsDto {
```
djabarovgeorge, line 1: 🧹

libs/dal/src/repositories/integration/integration.entity.ts
```diff
@@ -59,4 +30,6 @@ export class IntegrationEntity {
   deletedBy: string;
 }
 
+export type ICredentialsEntity = ICredentials;
```
djabarovgeorge, line 33: was not sure what to call this one :/ 
because it is basically a type of interface, i left the I for now.
any thought about it?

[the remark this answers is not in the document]
djabarovgeorge: Decided to keep the "Entity" separation so we will have it in code. because one can change while the other one may not.

packages/application-generic/src/factories/chat/handlers/discord.handler.ts
```diff
@@ -1,4 +1,4 @@
-import { ICredentials } from '@novu/dal';
+import { ICredentials } from '@novu/shared';
```
djabarovgeorge, line 1: I thought that the shared one will do the work in the provider's packages.

p-fernandez (approved):
The naming and location of some of the interfaces can be done in the future. This PR is big enough already.
🌟

PR #3650, opened 2023-06-22
feat(api): update get integrations endpoints

p-fernandez:

Makes the retrieval of the integrations API to be based on organization rather than environment. This is a breaking change.
It also updates documentation and the tests accordingly.

The multi provider feature will implement UI changes that need the integrations API to change the way the integrations are returned.

### DIFF
--- a/apps/api/src/app/events/e2e/trigger-event.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event.e2e.ts
@@ -639,11 +639,17 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
       _environmentId: session.environment._id,
     });
 
-    await integrationRepository.delete({
-      _id: { $in: existingIntegrations.map((integration) => integration._id) },
+    const integrationIdsToDelete = existingIntegrations.flatMap((integration) =>
+      integration._environmentId === session.environment._id ? [integration._id] : []
+    );
+
+    const deletedIntegrations = await integrationRepository.deleteMany({
+      _id: { $in: integrationIdsToDelete },
       _organizationId: session.organization._id,
       _environmentId: session.environment._id,
     });
+
+    expect(deletedIntegrations.modifiedCount).to.eql(integrationIdsToDelete.length);
 
     const newSubscriberIdInAppNotification = SubscriberRepository.createObjectId();
     const channelType = ChannelTypeEnum.EMAIL;
@@ -728,7 +734,7 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
     expect(messages[0].providerId).to.be.equal(EmailProviderIdEnum.SendGrid);
 
     const payload = {
-      providerId: 'mailgun',
+      providerId: EmailProviderIdEnum.Mailgun,
       channel: 'email',
       credentials: { apiKey: '123', secretKey: 'abc' },
       active: true,
--- a/apps/api/src/app/integrations/e2e/create-integration.e2e.ts
+++ b/apps/api/src/app/integrations/e2e/create-integration.e2e.ts
@@ -1,5 +1,6 @@
 import { IntegrationRepository, IntegrationEntity } from '@novu/dal';
 import { UserSession } from '@novu/testing';
+import { ChannelTypeEnum, EmailProviderIdEnum, SmsProviderIdEnum } from '@novu/shared';
 import { expect } from 'chai';
 
 describe('Create Integration - /integration (POST)', function () {
@@ -12,39 +13,39 @@ describe('Create Integration - /integration (POST)', function () {
   });
 
   it('should create the email integration successfully', async function () {
-    const integration = (await session.testAgent.get(`/v1/integrations`)).body.data.find(
-      (searchIntegration) => searchIntegration.channel === 'email'
+    const integrations = (await session.testAgent.get(`/v1/integrations`)).body.data;
+
+    const emailIntegrations = integrations.filter(
+      (searchIntegration) => searchIntegration.channel === ChannelTypeEnum.EMAIL
     );
 
-    expect(integration.providerId).to.equal('sendgrid');
-    expect(integration.channel).to.equal('email');
-    expect(integration.credentials.apiKey).to.equal('SG.123');
-    expect(integration.credentials.secretKey).to.equal('abc');
-    expect(integration.active).to.equal(true);
+    expect(emailIntegrations.length).to.eql(2);
+
+    for (const emailIntegration of emailIntegrations) {
+      expect(emailIntegration.providerId).to.equal(EmailProviderIdEnum.SendGrid);
+      expect(emailIntegration.channel).to.equal(ChannelTypeEnum.EMAIL);
+      expect(emailIntegration.credentials.apiKey).to.equal('SG.123');
+      expect(emailIntegration.credentials.secretKey).to.equal('abc');
+      expect(emailIntegration.active).to.equal(true);
+    }
   });
 
   it('should create the sms integration successfully', async function () {
-    const integration = (await session.testAgent.get(`/v1/integrations`)).body.data.find(
-      (searchIntegration) => searchIntegration.channel === 'sms'
-    );
-
-    expect(integration.providerId).to.equal('twilio');
-    expect(integration.channel).to.equal('sms');
-    expect(integration.credentials.accountSid).to.equal('AC123');
-    expect(integration.credentials.token).to.equal('123');
-    expect(integration.active).to.equal(true);
-  });
-
-  it('should create the sms and email integration successfully', async function () {
     const integrations = (await session.testAgent.get(`/v1/integrations`)).body.data;
 
-    expect(integrations.length).to.equal(6);
+    const smsIntegrations = integrations.filter(
+      (searchIntegration) => searchIntegration.channel === ChannelTypeEnum.SMS
+    );
 
-    const emailIntegration = integrations.find((i) => i.channel.toString() === 'email');
-    const smsIntegration = integrations.find((i) => i.channel.toString() === 'sms');
+    expect(smsIntegrations.length).to.eql(2);
 
-    expect(emailIntegration).to.be.ok;
-    expect(smsIntegration).to.be.ok;
+    for (const smsIntegration of smsIntegrations) {
+      expect(smsIntegration.providerId).to.equal(SmsProviderIdEnum.Twilio);
+      expect(smsIntegration.channel).to.equal(ChannelTypeEnum.SMS);
+      expect(smsIntegration.credentials.accountSid).to.equal('AC123');
+      expect(smsIntegration.credentials.token).to.equal('123');
+      expect(smsIntegration.active).to.equal(true);
+    }
   });
 
   it('should return error on creation of same provider on same environment twice', async function () {
@@ -89,17 +90,21 @@ describe('Create Integration - /integration (POST)', function () {
 
     const integrations = await integrationRepository.findByEnvironmentId(environmentId);
 
-    const firstIntegration = integrations.find((i) => i.providerId.toString() === 'sendgrid');
-    const secondIntegration = integrations.find((i) => i.providerId.toString() === 'mailgun');
+    const firstIntegration = integrations.find(
+      (i) => i.providerId.toString() === EmailProviderIdEnum.SendGrid && i._environmentId === session.environment._id
+    );
+    const secondIntegration = integrations.find(
+      (i) => i.providerId.toString() === EmailProviderIdEnum.Mailgun && i._environmentId === session.environment._id
+    );
 
-    expect(firstIntegration.active).to.equal(false);
-    expect(secondIntegration.active).to.equal(true);
+    expect(firstIntegration?.active).to.equal(false);
+    expect(secondIntegration?.active).to.equal(true);
   });
 
   it('should create custom SMTP integration with TLS options successfully', async function () {
     const payload = {
-      providerId: 'nodemailer',
-      channel: 'email',
+      providerId: EmailProviderIdEnum.CustomSMTP,
+      channel: ChannelTypeEnum.EMAIL,
       credentials: {
         host: 'smtp.example.com',
         port: '587',
@@ -117,8 +122,12 @@ describe('Create Integration - /integration (POST)', function () {
 
     const integrations = await integrationRepository.findByEnvironmentId(environmentId);
 
-    const sendGridIntegration = integrations.find((i) => i.providerId.toString() === 'sendgrid');
-    const nodeMailerIntegration = integrations.find((i) => i.providerId.toString() === 'nodemailer');
+    const sendGridIntegration = integrations.find(
+      (i) => i.providerId.toString() === EmailProviderIdEnum.SendGrid && i._environmentId === session.environment._id
+    );
+    const nodeMailerIntegration = integrations.find(
+      (i) => i.providerId.toString() === EmailProviderIdEnum.CustomSMTP && i._environmentId === session.environment._id
+    );
 
     expect(sendGridIntegration?.active).to.equal(false);
     expect(nodeMailerIntegration?.credentials?.host).to.equal(payload.credentials.host);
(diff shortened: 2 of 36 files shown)


apps/api/src/app/integrations/e2e/get-integration.e2e.ts
```diff
@@ -10,24 +10,35 @@ describe('Get Integrations - /integrations (GET)', function () {
     await session.initialize();
   });
 
-  it('should get newly created integration', async function () {
+  it('should retrieve all the integrations of all environments from an organization from the prefilled test data', async () => {
```
p-fernandez, line 13: I am replacing this existing test that just checked one integration from the prefilled test data for this one that does a deeper check and serves to prove the functionality of the endpoint even if we test again the test data.

apps/api/src/app/integrations/usecases/remove-integration/remove-integration.usecase.ts
```diff
@@ -13,6 +13,7 @@ export class RemoveIntegration {
 
   async execute(command: RemoveIntegrationCommand) {
     try {
+      // TODO: We should check first if the Integration exists in the database
```
p-fernandez, line 16: Something I found.

docs/docs/platform/integrations.md
```diff
@@ -28,3 +30,9 @@ For email providers, you will be required to add the following data:
 ## Missing a provider?
 
 Novu is an open-source platform, meaning that if you are missing a particular provider you can create one using the guide specified [here](/community/create-provider).
+
+### Changelog
+
+:::warning
+In version 0.17.0 we introduced a breaking change in the Integrations API as the GET endpoints passed to bring the integrations filtered by environment identifier to bring the integrations filtered by organization.
+:::warning
```
p-fernandez, line 38: Is this a right place?

libs/testing/src/integration.service.ts
```diff
@@ -58,5 +58,18 @@ export class IntegrationService {
     };
 
     await this.integrationRepository.create(pushFcmPayload);
+
+    const inAppPayload = {
```
p-fernandez, line 62: This was moved from here as it didn't make sense to me to have it isolated (and also thus we create the integration also for any environment we create during tests).
https://github.com/novuhq/novu/blob/b23c6c56c55ab9950495bde79caccecd75e46d4f/libs/testing/src/user.session.ts#L149

libs/testing/src/user.session.ts
```diff
@@ -44,12 +36,13 @@ const EMAIL_BLOCK: IEmailBlock[] = [
 ];
 
 export class UserSession {
-  private environmentRepository = new EnvironmentRepository();
```
p-fernandez, line 47: We have the services with the dedicated functionalities to prepare the session so makes sense to use them rather than the DAL directly.

libs/testing/src/user.session.ts
```diff
@@ -113,19 +106,9 @@ export class UserSession {
...
-
-        if (!options?.noIntegrations) {
-          await this.createIntegration();
-        }
-        await this.createFeed();
-        await this.createFeed('New');
-      }
+    if (!options.noOrganization && !options?.noEnvironment) {
+      const shouldCreateIntegrations = !options?.noIntegrations;
+      await this.createEnvironmentsAndFeeds(shouldCreateIntegrations);
```
p-fernandez, line 111: Grouped all this code into one single method.

libs/testing/src/user.session.ts
```diff
@@ -146,17 +129,6 @@ export class UserSession {
   private async initializeWidgetSession() {
     this.subscriberId = SubscriberRepository.createObjectId();
 
-    await this.integrationRepository.create({
```
p-fernandez, line 149: Reference: https://github.com/novuhq/novu/pull/3650/files#r1238768923

libs/testing/src/user.session.ts
```diff
@@ -248,24 +228,16 @@ export class UserSession {
     this.notificationGroups = groupsResponse.body.data;
   }
 
-  async addEnvironment() {
```
p-fernandez, line 251: Unused.

packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -18,8 +22,8 @@ export class GetDecryptedIntegrations {
   async execute(
     command: GetDecryptedIntegrationsCommand
   ): Promise<IntegrationEntity[]> {
-    const query: Partial<IntegrationEntity> & { _environmentId: string } = {
-      _environmentId: command.environmentId,
+    const query: Partial<IntegrationEntity> & { _organizationId: string } = {
```
p-fernandez, line 25: Actually this is the task definition of done. Funnily to be able to do this we needed to do a ton of things before.

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -597,12 +597,18 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
       _environmentId: session.environment._id,
     });
 
-    await integrationRepository.delete({
-      _id: { $in: existingIntegrations.map((integration) => integration._id) },
+    const integrationIdsToDelete = existingIntegrations.flatMap((integration) =>
+      integration._environmentId === session.environment._id ? [integration._id] : []
+    );
+
+    const deletedIntegrations = await integrationRepository.deleteMany({
```
p-fernandez, line 604: Had to create a deleteMany method as the delete one has a safeguard of looking for one integration. With the change of paradigm and after adding test data for production environment too can't be relied on for this task.

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -12,39 +13,39 @@ describe('Create Integration - /integration (POST)', function () {
...
-    expect(integration.credentials.token).to.equal('123');
-    expect(integration.active).to.equal(true);
+    expect(emailIntegrations.length).to.eql(2);
+
+    for (const emailIntegration of emailIntegrations) {
+      expect(emailIntegration.providerId).to.equal(EmailProviderIdEnum.SendGrid);
+      expect(emailIntegration.channel).to.equal(ChannelTypeEnum.EMAIL);
+      expect(emailIntegration.credentials.apiKey).to.equal('SG.123');
+      expect(emailIntegration.credentials.secretKey).to.equal('abc');
+      expect(emailIntegration.active).to.equal(true);
```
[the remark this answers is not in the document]
p-fernandez: This was a quite dumb test testing test data so I just double check that for both environments the integration is created. Wouldn't make much difference if checking those I reckon.

apps/api/src/app/integrations/e2e/get-integration.e2e.ts
```diff
@@ -39,21 +50,29 @@ describe('Get Integrations - /integrations (GET)', function () {
...
 
-    // create nodemailer integration
+    // create nodemailer integration added to the existing sendgrid integration
     await session.testAgent.post('/v1/integrations').send(nodeMailerProviderPayload);
 
     const activeIntegrations = (await session.testAgent.get(`/v1/integrations/active`)).body.data;
 
-    const activeEmailIntegration = activeIntegrations.find(
-      (integration) => integration.channel == ChannelTypeEnum.EMAIL
+    expect(activeIntegrations.length).to.eq(12);
```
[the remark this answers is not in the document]
p-fernandez: We would need to do the same all testing with the FF activated... 😕

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
```diff
@@ -48,16 +49,16 @@ export abstract class SendMessageBase extends SendMessageType {
   }
 
   @CachedQuery({
-    builder: ({ environmentId, ...command }: GetDecryptedIntegrationsCommand) =>
+    builder: ({ environmentId, ...command }: GetEnvironmentDecryptedIntegrationsCommand) =>
```
p-fernandez, line 52: In the worker, the functionalities regarding the providers should still be tackled by environment, while the functionalities in the API for the general endpoints (get integrations and get active integrations) need to be done at organization level. This caused to have to duplicate the commands used for getting the decrypted integrations to satisfy the conditions without having to create an extra use case like `GetEnvironmentDecryptedIntegrationsUseCase`. That could be an option we can follow if preferred.

docs/docs/platform/integrations.md
p-fernandez: I would agree, but I think a lot of people don't read the changelogs based on the support threads, so I was thinking the most the better. Trying to show the information by chance probability. 😄

libs/testing/src/user.session.ts
```diff
@@ -84,7 +77,6 @@ export class UserSession {
     options: {
       noOrganization?: boolean;
       noEnvironment?: boolean;
-      noIntegrations?: boolean;
```
p-fernandez, line 87: I removed this. It was only used in 2 tests. In one of the tests didn't make sense at all to not have default integrations and the other one was doing something that could be checked with the existing test data.

packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -18,10 +18,14 @@ export class GetDecryptedIntegrations {
...
+      _organizationId: command.organizationId,
     };
 
+    if (command.environmentId) {
+      query._environmentId = command.environmentId;
+    }
```
p-fernandez, line 27: We give the option to decrypt the integrations:
- all of the organization at once
- all of the specific environment of an organization with this condition

docs/docs/platform/integrations.md
LetItRock: we should bring the changelog closer to the API docs

packages/application-generic/src/commands/project.command.ts
djabarovgeorge, line 1: Should we add @IsMongoId() on those fields?

packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.command.ts
djabarovgeorge, line 1: Missing here @IsBoolea() and @IsEnum() 👀

apps/api/src/app/integrations/usecases/remove-integration/remove-integration.usecase.ts
djabarovgeorge: I am not entirely sure for 2 reasons.
1. due to the reason that the integration in the cache is stored as a query, even if we don't find integration by `command.integrationId` we still need to invalidate because it was not stored by `id` :(
2. regarding this one I am not sure, we could delete it without checking, and if it doesn't exist we will throw an exception from the DAL layer. that will basically mean 1 request to the DB and not 2.

libs/dal/src/repositories/integration/integration.repository.ts
```diff
@@ -43,13 +44,36 @@ export class IntegrationRepository extends BaseRepository<IntegrationDBModel, In
...
+    if (!_environmentId || !_organizationId) {
+      throw new DalException(
+        'Deletion operation blocked for missing any of these properties: [_environmentId, _organizationId]. We are avoiding a potential unexpected multiple deletion'
+      );
+    }
+
+    const { acknowledged, modifiedCount, matchedCount } = await this.integration.delete(query);
+
+    if (matchedCount === 0 || modifiedCount === 0) {
+      throw new DalException(
```
djabarovgeorge, line 66: Would not it be enough to return the actual result of matchedCount = 0 modifiedCount = 0?

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -686,7 +692,7 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
     expect(messages[0].providerId).to.be.equal(EmailProviderIdEnum.SendGrid);
 
     const payload = {
-      providerId: 'mailgun',
+      providerId: EmailProviderIdEnum.Mailgun,
```
djabarovgeorge, line 695: 🎨

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
LetItRock: I feel like we don't need to have two commands, just make the env property optional... we don't use anywhere that approach with two commands that might be provided to the use-case, this is confusing...

p-fernandez: TypeScript complains if you provide `environmentId` as optional because it is needed for the `buildIntegrationKey` functionality. 😕 
Therefore the reason for duplication. I am just proposing 2 different ways. At first went for the lazier one.

libs/dal/src/repositories/integration/integration.repository.ts
p-fernandez: I would expect that if trying to delete entries and there are not matched or any of the matched updated that means the query is like a 404 for the first case and for the second an error has happened below the DAL level, therefore is important to flag that it happened.
Do you think returning kinda like a go through in this method serves better purposes? Also that for me would mean the DAL responsibility of the management of what is happening is pushed in other layers. 😕

apps/api/src/app/integrations/usecases/remove-integration/remove-integration.usecase.ts
p-fernandez: Regarding both points I am afraid that now with multiple integrations we would need to change that caching approach, as we will face situations of integrations for the same provider so the only way of differentiate them is by their id. 😕

djabarovgeorge: Will love to sync about it because i am not i am following the whole matter here

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
djabarovgeorge: this decorator will be removed after Pawels & Biswa PRs merged, and we will cache the `select integration usecase`.
will it solve the command duplication here?

djabarovgeorge: this decorator will be removed after Pawels & Biswa PRs merged, and we will cache the select integration usecase.
will it solve the duplication here?

libs/dal/src/repositories/integration/integration.repository.ts
djabarovgeorge: tend to agree maybe we could add it in the DAL layer (most definitely not in this PR)

libs/testing/src/environment.service.ts
```diff
@@ -30,7 +43,7 @@ export class EnvironmentService {
     return environment;
   }
 
-  async getProductionEnvironment(organizationId: any) {
+  async getProductionEnvironment(organizationId: string) {
```
djabarovgeorge, line 46: 🔥

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
p-fernandez: Yeah, it could be. But this PR might need to go before so then this is temporary and then deleted on their work.

packages/application-generic/src/commands/project.command.ts
djabarovgeorge: in their repo it says that it: `Checks if the string is a valid hex-encoded representation of a MongoDB ObjectId.`

apps/web/cypress/tests/templates-store.spec.ts
```diff
@@ -12,7 +12,6 @@ describe('Templates Store', function () {
   };
 
   beforeEach(function () {
-    cy.clearDatabase();
```
p-fernandez: My experience is this was showing up some false positives. I can still replicate the same failures in the local environment even with its removal.

LetItRock (approved):
⭐

p-fernandez:
We still need to review the Cypress tests properly. Some stuff looks off. 😞

PR #3654, opened 2023-06-23
chore(web): use is_multi_provider_configuration_enabled feature flag

LetItRock:

Use the `IS_MULTI_PROVIDER_CONFIGURATION_ENABLED` feature flag to toggle between the Integrations List Page / Integrations Store Tiles.

For the e2e tests approach I think that we have to run for both pages, that's why I invented a way to achieve it with the global variable `_cypress`, check the code and comments.

Screenshot 2023-06-23 at 11 32 14
Screenshot 2023-06-23 at 11 32 23


apps/web/cypress/tests/integration-store.spec.ts
```diff
@@ -1,6 +1,14 @@
+// @ts-nocheck
 import { ChannelTypeEnum, InAppProviderIdEnum } from '@novu/shared';
 
-describe.skip('Integration store page', function () {
+Cypress.on('window:before:load', (win) => {
+  win._cypress = {
+    ...win._cypress,
+    IS_INTEGRATIONS_LIST_PAGE_ENABLED: 'false',
+  };
+});
```
LetItRock, line 9: Set the `_cypress.IS_INTEGRATIONS_LIST_PAGE_ENABLED` flag to false, it is then used in the configuration file where we do set up the env variables. 
I wasn't able to use the `window._env_` property as it overrides somewhere during the application initialization all the values.

apps/web/cypress/tests/integrations-list-page.spec.ts
```diff
@@ -1,3 +1,12 @@
+// @ts-nocheck
+
+Cypress.on('window:before:load', (win) => {
+  win._cypress = {
+    ...win._cypress,
+    IS_INTEGRATIONS_LIST_PAGE_ENABLED: 'true',
+  };
+});
```
LetItRock, line 8: similar approach with flag `true`, this way we can run both e2e tests for these pages

apps/web/src/config/index.ts
```diff
@@ -61,7 +61,8 @@ export const IS_TEMPLATE_STORE_ENABLED = isCypress
   : window._env_.IS_TEMPLATE_STORE_ENABLED || process.env.IS_TEMPLATE_STORE_ENABLED || 'false';
 
 export const IS_MULTI_PROVIDER_CONFIGURATION_ENABLED = isCypress
-  ? window._env_.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED ||
+  ? window._cypress.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED ||
```
LetItRock, line 64: use the flag

p-fernandez (approved):
🌟

apps/web/src/config/index.ts
djabarovgeorge: Those config conditions are becoming more and more complex to understand at first glance, maybe we could do another approach in the near future. (not related to this PR :) )

apps/web/src/pages/integrations/IntegrationsListPage.tsx
```diff
@@ -104,5 +105,7 @@ const IntegrationsList = () => {
 };
 
 export const IntegrationsListPage = () => {
-  return <IntegrationsList />;
+  const isIntegrationsListPageEnabled = useIsMultiProviderConfigurationEnabled();
+
+  return isIntegrationsListPageEnabled ? <IntegrationsList /> : <IntegrationsStore />;
```
djabarovgeorge, line 110: 👌

djabarovgeorge (approved):
🚀

PR #3660, opened 2023-06-26
feat: select integration usecase

BiswaViraj:

- Adds the select integration usecase.
- Will be updated once conditions are added


packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -0,0 +1,51 @@
...
+  ) {}
+
+  async execute(
+    command: SelectIntegrationCommand
+  ): Promise<IntegrationEntity | undefined> {
+    const query: Partial<IntegrationEntity> & { _organizationId: string } = {
+      _organizationId: command.organizationId,
+      _environmentId: command.environmentId,
+      channel: command.channelType,
+      providerId: command.providerId,
```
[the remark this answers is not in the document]
LetItRock: noticed the same and fixed ;)

packages/application-generic/src/usecases/select-integration/select-integration.spec.ts
```diff
@@ -0,0 +1,93 @@
+import { Test } from '@nestjs/testing';
+import { ChannelTypeEnum, EmailProviderIdEnum } from '@novu/shared';
+import { IntegrationEntity, IntegrationRepository } from '@novu/dal';
+import * as sinon from 'sinon';
+import { UserSession } from '@novu/testing';
+import { SelectIntegration } from './select-integration.usecase';
+import { SelectIntegrationCommand } from './select-integration.command';
+
+describe('select integration', function () {
```
LetItRock, line 9: the tests are not working if I do run them manually with `node 'node_modules/jest/bin/jest.js' './src/usecases/select-integration/select-integration.spec.ts' -c './jest.config.js' -t 'select integration'`

BiswaViraj: All the tests in the `application-generic` package are failing with the errors with the assertion methods
`Property 'toContain' does not exist on type 'Assertion'. Did you mean 'contain'?`

LetItRock: yes, but let's make sure that this one works, so someone who will be fixing them will have less work to do

djabarovgeorge:
Are the tests that were mentioned [here](https://github.com/novuhq/novu/pull/3660#discussion_r1243290095) fine right now?

BiswaViraj:
> Are the tests that were mentioned [here](https://github.com/novuhq/novu/pull/3660#discussion_r1243290095) fine right now?

@djabarovgeorge Yes, the errors are fixed,
Although the tests won't work as there is no test setup with testServer and MongoDB connection inside `application-generic` package

packages/application-generic/src/usecases/select-integration/select-integration.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      },
+      providerId: 'test-provider-id',
+      deleted: false,
+      identifier: 'test-integration-identifier',
+      name: 'test-integration-name',
+      deletedAt: null,
+      deletedBy: null,
+    };
+
+    sinon.stub(integrationRepository, 'findOne').resolves(testIntegration);
```
p-fernandez: Is the data stubbed because the problems are you finding running the tests in Application Generic?

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -0,0 +1,52 @@
...
+      _environmentId: command.environmentId,
+      channel: command.channelType,
+      ...(command.providerId ? { providerId: command.providerId } : {}),
+      active: true,
+    };
+
+    const integration = await this.integrationRepository.findOne(
+      query,
+      undefined,
+      { query: { sort: { createdAt: -1 } } }
```
p-fernandez, line 32: Not in the requirements but as a team we might discuss if we want to favour to also sort them based in if they are active or not. Like:
First: newest and active
Second: newest and disabled
Third: oldest and disabled

LetItRock (approved):

packages/application-generic/src/usecases/select-integration/select-integration.spec.ts
LetItRock: I've updated the tests, they are passing right now

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
LetItRock: I'm not sure I do understand what you are saying. This use-case will be used in the send phase and it will look for the latest active integration, we might want to return in the future an array of active ones, but not sure why we want to have also inactive?

p-fernandez: Argh I mixed that because I thought was going to reuse this use case like the one we replaced (GetDecryptedIntegrations). Ignore me, please. 🙏🏻

Issue #3311, opened 2023-05-01
🚀 Feature: Ability to discard a Change


phlrss:
### 🔖 Feature description

It would be greatly beneficial if we had the ability to "Discard" a Change rather than only "Promote" it. If we make changes to a template in Development, and then want to fine-tune or even revert the change there is no way to "Discard" it so it remains in the Change table indefinitely.

### 🎤 Why is this feature needed ?

Without it, old no longer relevant changes will remain trailing in the Change table unnecessarily. 

### ✌️ How do you aim to achieve this?

Simply add a "Discard" button beside the "Promote" button with a dialog "Are you sure you want to discard this change?" to avoid any accidental discards. And perform a delete by ID in the `changes` table in the database

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

cedric-fauth:
Also it would be great to improve the changes check to see if there are really any changes made or if nothing has changed.
When calling the api and updating a workflow twice with the same data there are 2 changes displayed even if nothing has changed since the first update. You should consider checking for changes based on actual data changes inside the workflows and displaying a diff so that everyone knows what has changed.

scopsy:
@gitstart posting my Pr comment here as well:
```
 As It might affect some other underlying changes once the entity will be removed. We will need to do some deeper investigation before we can go and implement something like this. My suggestion is to close this for the time being until we finalize a proper flow for this.
 ```

One of the challenges is how to deal with the already made changes on dev if we discard the change from the list, subsequent save can just return the unexpected change as a diff is calculated for each time a workflow is updated

bartelemi:
To add even more context to this issue, there is one more situation where this feature would be helpful. Currently we got stuck in a situation where we made some changes to a workflow that was subsequently removed, but the "change" was never promoted.

Trying to promote such change just errors with a message: "_Notification Group Id 111a11dbb1a1aa111a1a1111 not found, Notification Template: Mention in a comment_" (ID redacted) and is kind of stuck in a limbo.

scopsy:
We are moving towards publish and sync based approach in v2, so closing this ticket as it not something we are going to support

PR #3661, opened 2023-06-26
NV-2194 - 🚀 Feature: Ability to discard a Change

gitstart:
This feature gives users the ability to delete changes from the changes table. 

Closes: https://github.com/novuhq/novu/issues/3311

### Other information

[Loom Video](https://www.loom.com/share/92da2ab901c441038f967d537ea640a7?sid=d284e1b9-0a67-4286-8e2a-bf7d51cb8a1a)

---

### DIFF
--- a/apps/api/src/app/change/changes.controller.ts
+++ b/apps/api/src/app/change/changes.controller.ts
@@ -8,6 +8,7 @@ import {
   Query,
   UseGuards,
   UseInterceptors,
+  Delete,
 } from '@nestjs/common';
 import { IJwtPayload } from '@novu/shared';
 import { UserSession } from '../shared/framework/user.decorator';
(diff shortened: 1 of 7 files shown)


michaldziuba03:
@gitstart I think you referenced wrong issue. 

I guess it shoud be: https://github.com/novuhq/novu/issues/3311

jainpawan21:
> @gitstart I think you referenced wrong issue.
> 
> I guess it should be: #3311

Updated

gitstart:

ainouzgali (commented):
@gitstart thanks for your hard work. I'm afraid it is more complicated than that. 
While you are deleting the change itself, the content of the change(for example, changes to a workflow) will not be discarded and still exist in dev env. So any other future change in that workflow will be inconsistent as it will build on current content. 

Another issue is that you are deleting one change id, but that represents and depends on other parent changes. So if you delete the one from the table, I believe it would just delete one, and still show the children changes. 
 
Also, some changes are actually aggregated, and includes multiple separate entities changes. For example - creating a notification group that is immediately used in a workflow - will have one row change for both. 

In my opinion, this should be carefully planned before executing. Maybe the right thing is to wait for the team and add to the roadmap. @scopsy what do you think?

scopsy:
@ainouzgali I totally aggree here that there are just too many moving items for simply deleting the change. As It might affect some other underlying changes once the entity will be removed. We will need to do some deeper investigation before we can go and implement something like this. My suggestion is to close this for the time being until we finalize a proper flow for this. 🙏

ainouzgali:
@gitstart closing this for now. Again, thank you! We appreciate it

PR #3662, opened 2023-06-26
chore(api,worker): use select integration usecase before we do a send

LetItRock:

Refactor: use the new use-case `SelectIntegration` in the places in the code where we do a message send. The use-case right now picks the last created and active integration for a channel, but in the future, it's functionality will be more complex.

Multi Channel Providers Configuration PRD.


apps/api/src/app/integrations/usecases/index.ts
```diff
@@ -17,6 +22,7 @@ export const USE_CASES = [
   CreateIntegration,
   GetIntegrations,
   GetActiveIntegrations,
+  SelectIntegration,
```
LetItRock, line 25: the new use-case is added to the integrations module use-cases array

apps/api/src/app/widgets/usecases/initialize-session/initialize-session.usecase.ts
```diff
@@ -35,10 +35,8 @@ export class InitializeSession {
       throw new ApiException('Please provide a valid app identifier');
     }
 
-    const inAppIntegration = await this.getDecryptedIntegrations.execute(
-      GetDecryptedIntegrationsCommand.create({
-        findOne: true,
-        active: true,
+    const inAppIntegration = await this.selectIntegration.execute(
```
LetItRock, line 38: use the new use-case to check if the `in-app` integration is created... in the future, we might allow having many in-app integrations, that's why we've also introduced it here...

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -44,15 +43,9 @@ export class SendMessageChat extends SendMessageBase {
     protected createLogUsecase: CreateLog,
     protected createExecutionDetails: CreateExecutionDetails,
     private compileTemplate: CompileTemplate,
-    protected getDecryptedIntegrationsUsecase: GetDecryptedIntegrations
+    protected selectIntegration: SelectIntegration
```
LetItRock, line 46: the new integration use-case is passed to the base `SendMessageBase`

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
```diff
@@ -54,12 +55,10 @@ export abstract class SendMessageBase extends SendMessageType {
...
-  protected async getIntegration(getDecryptedIntegrationsCommand: GetDecryptedIntegrationsCommand) {
-    return (
-      await this.getDecryptedIntegrationsUsecase.execute(
-        GetDecryptedIntegrationsCommand.create(getDecryptedIntegrationsCommand)
-      )
-    )[0];
+  protected async getIntegration(
+    selectIntegrationCommand: SelectIntegrationCommand
+  ): Promise<IntegrationEntity | undefined> {
+    return this.selectIntegration.execute(SelectIntegrationCommand.create(selectIntegrationCommand));
```
LetItRock, line 61: in the upcoming PR with the feature flag I'll bring the code back :P

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
djabarovgeorge, line 1: One thing that came to my mind right now and apologies for raising it only now is that we need to do more discovery on this channel (maybe push as well) because the `Chat` `Integration Entity` credentials and the `Subscriber Entity` credentials are coupled.

Or we could notify the user regarding the restriction.

background: we have integration credentials that we use in order to integrate Novu user's Slack application, and we use it in order to get access to a webhook in order to send notifications on behave of this application. which is happening in  chat-oauth.usecase (i think we need to update the query there as well).

The possible problem could be if a user have 2 Slack integration with different application credentials. a message by the wrong  `Slack application` may be sent.

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
```diff
@@ -54,12 +55,10 @@ export abstract class SendMessageBase extends SendMessageType {
...
       }),
   })
-  protected async getIntegration(getDecryptedIntegrationsCommand: GetDecryptedIntegrationsCommand) {
-    return (
-      await this.getDecryptedIntegrationsUsecase.execute(
-        GetDecryptedIntegrationsCommand.create(getDecryptedIntegrationsCommand)
-      )
-    )[0];
+  protected async getIntegration(
+    selectIntegrationCommand: SelectIntegrationCommand
```
p-fernandez, line 59: In the child use cases you are not creating the commands, just passing an object. Even if they are compatible what are the reasons to do it like that?

djabarovgeorge: Do you mean to call the parent method in the children's classes?

If so then this is basically Legacy code where we encapsulated the functionality with @Cached decorator. I agree that this one looks like overkill at the moment but once we add the Feature Flag it will make more sense IMO.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
djabarovgeorge: It is a bit difficult case and I am not sure I managed to explain it properly, let me know if you want to sync about it.

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
p-fernandez: No, that we are not using <X>Command to create the objects in the use cases to pass the information to the base use case.

p-fernandez (approved):
I like the new implementation as it is cleaner and simpler. Wish we could just have to send the channel type to the base having a clever way to set up the organization, environment and user. That would be chef's kiss.
One thing is we are not using a `<X>Command.create` to pass the information so it is a bit different that we usually do.
And last I have now a concern with all the Cache changes if we are not missing the invalidation cache functionalities, as I do not see changes in the invalidation and I would expect them to be updated.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
LetItRock: It will pick the latest Slack integration created. The important thing here is the `chatWebhookUrl` is the URL where we do send the message, and this information is stored on the subscriber. So as a result no matter what the integration is, we will always send the message to the correct `chat`.

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
LetItRock: In the children's classes we are not using the use-cases, that's the reason. We do call this parent function `getIntegration` and it calls the use-case and uses the object and creates the command.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
LetItRock: I'm not aware of how it works, but seems like that part should also use the new use-case?

PR #3664, opened 2023-06-26
Add better Uptime Annoucement for System issues

Cliftonz:

This pr adds an integration for the web deployment to have announcements from better uptime be propagated to any user using Novu cloud.

This change allows us to communicate to users in the event of a failure.

image


BiswaViraj (approved):

jainpawan21:
@Cliftonz is .idea directory files needed for this change?

Cliftonz:
Yes, this tells Jetbrains to download this library for IntelliSense.

Cliftonz:
> I am not familiar with uptime announcements, will it send toast announcements as in your screenshot? Additionally what announcements it includes? Will the different environments will be separated, meaning US users won't get EU announcements and announcements and vice versa?

Yes, but it will only do it if I manually put it in better uptime. This would not happen automatically.

No, it will be a catch all. Thus if we are having an issue for US we will need to make sure that is put in the announcement.

image

apps/web/public/index.html
```diff
@@ -40,5 +40,11 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
+    <script
```
scopsy, line 43: @Cliftonz this should only be included for cloud based hosting, not self hosted

apps/web/public/index.html
```diff
@@ -23,28 +23,21 @@
...
-      To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    -->
-    <script
-      src="https://uptime.betterstack.com/widgets/announcement.js"
-      data-id="144175"
-      async="async"
-      type="text/javascript"
-    ></script>
+    <% if ( process.env.REACT_APP_NOVU_MANAGED_SERVICE === 'true' ) { %>
```
djabarovgeorge, line 33: @Cliftonz I know you already merged the PR but maybe we can reuse REACT_APP_DOCKER_HOSTED_ENV here so we won't have 2 environment variables that are similar to manage.

PR #3666, opened 2023-06-27
NV 2418 - Integration Selection Caching

djabarovgeorge:

We would like to apply the caching mechanism over the Integration Selection use-case. The rationale is that the integration configurations should not be changing really often.

### DIFF
--- a/apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
@@ -49,7 +49,7 @@ export class CreateIntegration {
 
       await this.invalidateCache.invalidateQuery({
         key: buildIntegrationKey().invalidate({
-          _environmentId: command.environmentId,
+          _organizationId: command.organizationId,
         }),
       });
 
--- a/packages/application-generic/src/services/cache/key-builders/queries.ts
+++ b/packages/application-generic/src/services/cache/key-builders/queries.ts
@@ -84,22 +84,26 @@ const buildMessageCountKey = () => {
 
 const buildIntegrationKey = () => {
   const cache = (
-    command: Record<string, unknown> & { _environmentId: string }
+    command: Record<string, unknown> & { _organizationId: string }
   ): string =>
-    buildQueryByEnvironmentKey({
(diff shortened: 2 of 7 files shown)


apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -49,7 +49,7 @@ export class CreateIntegration {
 
       await this.invalidateCache.invalidateQuery({
         key: buildIntegrationKey().invalidate({
-          _environmentId: command.environmentId,
+          _organizationId: command.organizationId,
```
djabarovgeorge, line 52: Here we specify the cache key and the level changed from environment to organization.

packages/application-generic/src/services/cache/key-builders/queries.ts
```diff
@@ -134,38 +138,40 @@ export const buildQueryKey = ({
...
 }): string =>
-  `${buildKeyByEnvironment({
+  `${buildKeyByOrganization({
     type,
     keyEntity,
-    environmentIdPrefix,
-    environmentId,
+    organizationIdPrefix,
+    organizationId,
   })}:${QUERY_PREFIX}=${JSON.stringify(query)}`;
```
djabarovgeorge, line 159: Here we can see that we specify the cache key where the key changed from environment to organization and that the environment is now part of the query key.

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -15,6 +16,13 @@ export class SelectIntegration {
     private getNovuIntegration: GetNovuIntegration
   ) {}
 
+  @CachedQuery({
+    builder: ({ organizationId, ...command }: SelectIntegrationCommand) =>
+      buildIntegrationKey().cache({
+        _organizationId: organizationId,
+        ...command,
+      }),
+  })
```
djabarovgeorge, line 25: Meaning we will cache the new usecase.

LetItRock (commented):
Looks good to me! ⭐

LetItRock (approved):

Issue #3627, opened 2023-06-20
[NV-2469] 🚀 Feature: Rename Sendinblue to Brevo


p-fernandez:
### 🔖 Feature description

Sendinblue is rebranding into Brevo.
Reference: https://www.brevo.com/blog/becoming-brevo/

New Brevo users might not know they could use Sendinblue provider in Novu.
Also depending in how they do the rebranding a migration might be needed.
We need to implement the changes needed to support this rebranding for existing and future users.

### 🎤 Why is this feature needed ?

We need to implement the changes needed to support this rebranding for existing and future users.

### ✌️ How do you aim to achieve this?

* Check if rebranding implies a migration
* Update the UI to reflect the rebranding and the potentially co-living of both brands.
* Update logos, links and documentation references.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-2469](https://linear.app/novu/issue/NV-2469/🚀-feature-rename-sendinblue-to-brevo)</sub>

gitstart:
@p-fernandez @scopsy I would like to pick this up

ankur0904:
Can I work on this issue?

PR #3667, opened 2023-06-27
NV-2469 - 🚀 Feature: Rename Sendinblue to Brevo

gitstart:
This PR implement the changes needed to support the rebranding of Sendinblue to Brevo, this includes to the UI or both the web app and the documentation

Closes: https://github.com/novuhq/novu/issues/3627

### Other information

[Loom Video](https://www.loom.com/share/0251e1100cb04f33be3f8f613fabcf8d?sid=adc5781c-02c5-4358-a81f-97aa1db2b974)

---


gitstart:
@p-fernandez this PR is ready for review

libs/shared/src/consts/providers/channels/email.ts
```diff
@@ -62,12 +62,12 @@ export const emailProviders: IProviderConfig[] = [
     logoFileName: { light: 'sendgrid.png', dark: 'sendgrid.png' },
   },
   {
-    id: EmailProviderIdEnum.Sendinblue,
-    displayName: 'Sendinblue',
+    id: EmailProviderIdEnum.Brevo,
```
scopsy, line 65: Unfortuently this change is not backwards compatible for scheduled jobs using `sendinblue` and they will just fail. My suggestion is to create backwards compatible approach where both enums still exist, and during the handler selection we process both sendinblue and brevo using the same handler.

scopsy (changes requested):

libs/shared/src/consts/providers/channels/email.ts
gitstart: Your feedback has been implemented @scopsy, thanks for the review.

scopsy:
Unfortunately there are still a few problems here :( Will try to comment on the relevant places a bit later 🙏

gitstart:
Noted. Eagerly waiting for your review 🙏

PR #3668, opened 2023-06-27
feat: add integration scheme migration script

djabarovgeorge:

integration migration script.

migration script in order to update the old document into the new scheme structure


apps/api/migrations/integration-scheme-update/add-integration-identifier-migration.ts
```diff
@@ -0,0 +1,95 @@
...
+  if (!integration.identifier) {
+    updatePayload.identifier = await getIntegrationIdentifier(environmentRepo, integration);
+  }
+
+  return updatePayload;
+}
+
+export async function getIntegrationIdentifier(environmentRepo, integration) {
+  const environment = await environmentRepo.findOne({ _id: integration._environmentId });
+  const environmentName = ENVIRONMENT_NAME_TO_SHORT_NAME[environment?.name || 'undefined'];
```
p-fernandez, line 87:
```suggestion
  const environmentName = ENVIRONMENT_NAME_TO_SHORT_NAME[environment?.name] || '';
```
And in the object remove the entry for `undefined`. I think it can be simplified this logic like this.

djabarovgeorge: OMG, I forgot to comment there apologies! I added it because of TS, but it could not happen in real life IMO added the empty string backoff as an unnecessary over-engineering guard.

apps/api/migrations/integration-scheme-update/add-integration-identifier-migration.ts
```diff
@@ -0,0 +1,95 @@
...
+async function getUpdatePayload(integration: IntegrationEntity, environmentRepo: EnvironmentRepository) {
+  const updatePayload: Partial<IntegrationEntity> = {};
+
+  if (!integration.name) {
+    updatePayload.name = getDisplayName(integration.providerId);
+  }
+
+  if (!integration.identifier) {
+    updatePayload.identifier = await getIntegrationIdentifier(environmentRepo, integration);
+  }
```
LetItRock, line 80: This is how I'm going to generate these for the create endpoint, maybe you could do it similarly:
```
const providerIdCapitalized = `${command.providerId.charAt(0).toUpperCase()}${command.providerId.slice(1)}`;
const defaultName = providers.find((provider) => provider.id === command.providerId)?.displayName ?? providerIdCapitalized;
const name = command.name ?? defaultName;
const identifier = command.identifier ?? `${slugify(name, { lower: true, strict: true })}-${shortid.generate()}`;
```

p-fernandez: It would be great if both places (migration and use case) share the same function, in my opinion. 👍🏻

apps/api/migrations/integration-scheme-update/add-integration-identifier-migration.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      channel: ChannelTypeEnum.EMAIL,
+      providerId: EmailProviderIdEnum.SendGrid,
+      active: true,
+    });
+
+    await integrationRepository.create({
+      _environmentId: session.environment._id,
+      _organizationId: session.organization._id,
+      channel: ChannelTypeEnum.SMS,
+      providerId: SmsProviderIdEnum.SNS,
```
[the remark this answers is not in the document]
djabarovgeorge: No reason, do you prefer another one here?

apps/api/migrations/integration-scheme-update/add-integration-identifier-migration.ts
```diff
@@ -0,0 +1,95 @@
+import { EnvironmentRepository, IntegrationEntity, IntegrationRepository } from '@novu/dal';
+import { providers } from '@novu/shared';
+
+export const ENVIRONMENT_NAME_TO_SHORT_NAME = { ['Development']: 'dev', ['Production']: 'prod', ['undefined']: '' };
+
+export async function addIntegrationIdentifierMigrationBatched() {
```
djabarovgeorge, line 6: Created the batched version so we could do it with less load on the DB.

apps/api/migrations/integration-scheme-update/add-integration-identifier-migration.spec.ts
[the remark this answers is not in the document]
p-fernandez: Sorry I didn't pay attention well and this is a test. No preference. All good. 😓

apps/api/migrations/integration-scheme-update/add-integration-identifier-migration.ts
djabarovgeorge: > This is how I'm going to generate these for the create endpoint, maybe you could do it similarly:
> 
> ```
> const providerIdCapitalized = `${command.providerId.charAt(0).toUpperCase()}${command.providerId.slice(1)}`;
> const defaultName = providers.find((provider) => provider.id === command.providerId)?.displayName ?? providerIdCapitalized;
> const name = command.name ?? defaultName;
> const identifier = command.identifier ?? `${slugify(name, { lower: true, strict: true })}-${shortid.generate()}`;
> ```

Is this logic already merged with next? If not can you pin here the relevant PR id so I will know where it is, once it is merged i will reuse the same function logic here

LetItRock: the PR: https://github.com/novuhq/novu/pull/3674

djabarovgeorge: in order to not wait for the merge to next i committed duplication of this logic for now.

LetItRock (approved):

PR #3669, opened 2023-06-27
fix(ci): debugging cypress

p-fernandez:


apps/api/package.json
```diff
@@ -48,7 +48,6 @@
     "@sentry/hub": "^7.40.0",
     "@sentry/node": "^7.40.0",
     "@sentry/tracing": "^7.40.0",
-    "@types/handlebars": "^4.1.0",
```
p-fernandez, line 51: PNPM install script said were not needed.

libs/dal/package.json
```diff
@@ -49,21 +49,19 @@
     "rimraf": "^3.0.2",
     "superagent-defaults": "^0.1.14",
     "supertest": "^5.0.0",
-    "twilio": "^3.62.0",
```
p-fernandez, line 52: Unused (and has no sense).

scopsy (approved):

PR #3674, opened 2023-06-28
feat(api): update the post, put integration endpoints to allow passing the name, identifier, skip passing credentials

LetItRock:

Updated the `POST /v1/integrations` endpoint: 
- to allow passing individually `name, identifier, environmentId`
- skipping `credentials`
- inactive by default
- the only required fields are: `providerId, channel`

Updated the `PUT /v1/integrations` endpoint: 
- to allow passing individually `name, identifier, environmentId`
- skipping `credentials`

Added e2e tests that are covering the current state.

### DIFF
--- a/apps/api/src/app/integrations/dtos/create-integration-request.dto.ts
+++ b/apps/api/src/app/integrations/dtos/create-integration-request.dto.ts
@@ -1,11 +1,27 @@
-import { IsBoolean, IsDefined, IsEnum, IsString, ValidateNested } from 'class-validator';
-import { ChannelTypeEnum, ICreateIntegrationBodyDto } from '@novu/shared';
-import { CredentialsDto } from './credentials.dto';
-import { ApiProperty } from '@nestjs/swagger';
+import { IsBoolean, IsDefined, IsEnum, IsMongoId, IsOptional, IsString, ValidateNested } from 'class-validator';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
+import { ChannelTypeEnum, ICreateIntegrationBodyDto } from '@novu/shared';
+
+import { CredentialsDto } from './credentials.dto';
 
 export class CreateIntegrationRequestDto implements ICreateIntegrationBodyDto {
-  @ApiProperty()
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsString()
+  name?: string;
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsString()
+  identifier?: string;
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsMongoId()
+  _environmentId?: string;
+
+  @ApiProperty({ type: String })
   @IsDefined()
   @IsString()
   providerId: string;
@@ -17,21 +33,21 @@ export class CreateIntegrationRequestDto implements ICreateIntegrationBodyDto {
   @IsEnum(ChannelTypeEnum)
   channel: ChannelTypeEnum;
 
-  @ApiProperty({
+  @ApiPropertyOptional({
     type: CredentialsDto,
   })
-  @IsDefined()
+  @IsOptional()
   @Type(() => CredentialsDto)
   @ValidateNested()
-  credentials: CredentialsDto;
+  credentials?: CredentialsDto;
 
-  @ApiProperty()
+  @ApiPropertyOptional({ type: Boolean })
+  @IsOptional()
+  @IsBoolean()
+  active?: boolean;
+
+  @ApiPropertyOptional({ type: Boolean })
   @IsDefined()
   @IsBoolean()
-  active: boolean;
-
-  @ApiProperty()
-  @IsDefined()
-  @IsBoolean()
-  check: boolean;
+  check?: boolean;
 }
--- a/apps/api/src/app/integrations/dtos/update-integration.dto.ts
+++ b/apps/api/src/app/integrations/dtos/update-integration.dto.ts
@@ -1,25 +1,40 @@
-import { ApiProperty } from '@nestjs/swagger';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { IUpdateIntegrationBodyDto } from '@novu/shared';
-import { IsBoolean, IsDefined, ValidateNested } from 'class-validator';
+import { IsBoolean, IsDefined, IsMongoId, IsOptional, IsString, ValidateNested } from 'class-validator';
 import { CredentialsDto } from './credentials.dto';
 import { Type } from 'class-transformer';
 
 export class UpdateIntegrationRequestDto implements IUpdateIntegrationBodyDto {
-  @ApiProperty()
-  @IsDefined()
-  @IsBoolean()
-  active: boolean;
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsString()
+  name?: string;
 
-  @ApiProperty({
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsString()
+  identifier?: string;
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsMongoId()
+  _environmentId?: string;
+
+  @ApiPropertyOptional({ type: Boolean })
+  @IsOptional()
+  @IsBoolean()
+  active?: boolean;
+
+  @ApiPropertyOptional({
     type: CredentialsDto,
   })
-  @IsDefined()
+  @IsOptional()
   @Type(() => CredentialsDto)
   @ValidateNested()
-  credentials: CredentialsDto;
+  credentials?: CredentialsDto;
 
-  @ApiProperty()
+  @ApiPropertyOptional({ type: Boolean })
   @IsDefined()
   @IsBoolean()
-  check: boolean;
+  check?: boolean;
 }
--- a/apps/api/src/app/integrations/e2e/create-integration.e2e.ts
+++ b/apps/api/src/app/integrations/e2e/create-integration.e2e.ts
@@ -1,4 +1,4 @@
-import { IntegrationRepository, IntegrationEntity } from '@novu/dal';
+import { IntegrationRepository, EnvironmentRepository } from '@novu/dal';
 import { UserSession } from '@novu/testing';
 import { ChannelTypeEnum, EmailProviderIdEnum, SmsProviderIdEnum } from '@novu/shared';
 import { expect } from 'chai';
@@ -6,13 +6,14 @@ import { expect } from 'chai';
 describe('Create Integration - /integration (POST)', function () {
   let session: UserSession;
   const integrationRepository = new IntegrationRepository();
+  const envRepository = new EnvironmentRepository();
 
   beforeEach(async () => {
     session = new UserSession();
     await session.initialize();
   });
 
-  it('should create the email integration successfully', async function () {
+  it('should get the email integration successfully', async function () {
     const integrations = (await session.testAgent.get(`/v1/integrations`)).body.data;
 
     const emailIntegrations = integrations.filter(
@@ -30,7 +31,7 @@ describe('Create Integration - /integration (POST)', function () {
     }
   });
 
-  it('should create the sms integration successfully', async function () {
+  it('should get the sms integration successfully', async function () {
     const integrations = (await session.testAgent.get(`/v1/integrations`)).body.data;
 
     const smsIntegrations = integrations.filter(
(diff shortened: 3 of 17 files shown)


apps/api/src/app/integrations/dtos/create-integration-request.dto.ts
```diff
@@ -1,11 +1,27 @@
...
+import { IsBoolean, IsDefined, IsEnum, IsMongoId, IsOptional, IsString, ValidateNested } from 'class-validator';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { Type } from 'class-transformer';
 import { ChannelTypeEnum, ICreateIntegrationBodyDto } from '@novu/shared';
+
 import { CredentialsDto } from './credentials.dto';
-import { ApiProperty } from '@nestjs/swagger';
-import { Type } from 'class-transformer';
 
 export class CreateIntegrationRequestDto implements ICreateIntegrationBodyDto {
```
LetItRock, line 8: The only required fields now are: `providerId, channel`. 
The default values for the boolean flags are `active: false, check: true`.

apps/api/src/app/integrations/dtos/create-integration-request.dto.ts
```diff
@@ -1,11 +1,27 @@
...
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsString()
+  identifier?: string;
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsMongoId()
+  _environmentId?: string;
```
LetItRock, line 22: The new optional fields: `name, identifier, _environmentId`. The `name` and `identifier` will be auto-generated if not provided. The `_environmentId` if not provided will be picked from the user token.

apps/api/src/app/integrations/dtos/create-integration-request.dto.ts
```diff
@@ -17,21 +33,21 @@ export class CreateIntegrationRequestDto implements ICreateIntegrationBodyDto {
...
-  @ApiProperty({
+  @ApiPropertyOptional({
     type: CredentialsDto,
   })
-  @IsDefined()
+  @IsOptional()
   @Type(() => CredentialsDto)
   @ValidateNested()
-  credentials: CredentialsDto;
+  credentials?: CredentialsDto;
```
LetItRock, line 42: do not require credentials

apps/api/src/app/integrations/dtos/update-integration.dto.ts
```diff
@@ -1,25 +1,40 @@
...
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsString()
+  identifier?: string;
+
+  @ApiPropertyOptional({ type: String })
+  @IsOptional()
+  @IsMongoId()
+  _environmentId?: string;
```
LetItRock, line 21: The new optional fields: `name, identifier, _environmentId` are optional in the update request, so we can send them individually if we want.

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -26,7 +28,20 @@ export class CreateIntegration {
...
+    if (command.identifier) {
+      const existingIntegrationWithIdentifier = await this.integrationRepository.findOne({
+        _organizationId: command.organizationId,
+        identifier: command.identifier,
+      });
+
+      if (existingIntegrationWithIdentifier) {
+        throw new BadRequestException('Integration with identifier already exists');
+      }
+    }
```
LetItRock, line 40: check if the integration with `identifier` already exists for the organization, if yes throw an error

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -26,7 +28,20 @@ export class CreateIntegration {
...
+
+    if (command.active && Object.keys(command.credentials ?? {}).length === 0) {
+      throw new BadRequestException('The credentials are required to activate the integration');
+    }
```
LetItRock, line 44: If someone tries to activate the integration without the credentials throw an error. I know this check is silly but I don't know the better way and also we have the email test check below implemented too.

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -53,31 +68,29 @@ export class CreateIntegration {
         }),
       });
 
-      response = await this.integrationRepository.create({
+      const providerIdCapitalized = `${command.providerId.charAt(0).toUpperCase()}${command.providerId.slice(1)}`;
+      const defaultName =
+        providers.find((provider) => provider.id === command.providerId)?.displayName ?? providerIdCapitalized;
+      const name = command.name ?? defaultName;
+      const identifier = command.identifier ?? `${slugify(name, { lower: true, strict: true })}-${shortid.generate()}`;
```
LetItRock, line 75: generate the `name, identifier` if not provided

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -32,6 +32,22 @@ export class UpdateIntegration {
...
+    if (identifierHasChanged) {
+      const existingIntegrationWithIdentifier = await this.integrationRepository.findOne({
+        _organizationId: command.organizationId,
+        identifier: command.identifier,
+      });
+
+      if (existingIntegrationWithIdentifier) {
+        throw new BadRequestException('Integration with identifier already exists');
+      }
+    }
```
LetItRock, line 45: update request: when identifier has changed, check if we don't have another similar in the DB, and if we have throw an error

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
ainouzgali: wdyt about throwing ConflictException here?

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -53,31 +68,29 @@ export class CreateIntegration {
...
 
-      if (command.active && ![ChannelTypeEnum.CHAT, ChannelTypeEnum.PUSH].includes(command.channel)) {
-        await this.deactivateSimilarChannelIntegrations.execute({
-          environmentId: command.environmentId,
-          organizationId: command.organizationId,
-          integrationId: response._id,
-          channel: command.channel,
-          userId: command.userId,
-        });
-      }
```
ainouzgali, line 73: We need to remember to keep it depending on the feature flag, right? Or am I missing something?

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -83,7 +113,7 @@ export class UpdateIntegration {
 
     if (command.active && ![ChannelTypeEnum.CHAT, ChannelTypeEnum.PUSH].includes(existingIntegration.channel)) {
       await this.deactivateSimilarChannelIntegrations.execute({
-        environmentId: command.environmentId,
+        environmentId,
         organizationId: command.organizationId,
         integrationId: command.integrationId,
         channel: existingIntegration.channel,
```
ainouzgali, line 119: Shouldn't it also be removed(depending on feature falg), as you did in create-integration?

libs/dal/src/repositories/integration/integration.repository.ts
```diff
@@ -33,15 +33,6 @@ export class IntegrationRepository extends BaseRepository<IntegrationDBModel, In
...
-    const existingIntegration = await this.findOne({
-      _environmentId: data._environmentId,
-      providerId: data.providerId,
-      channel: data.channel,
-    });
-    if (existingIntegration) {
-      throw new DalException('Duplicate key - One environment may not have two providers of the same channel type');
-    }
-
     return await super.create(data);
```
ainouzgali, line 36: Could this be removed completely? It is supposed to go the the create of the super if create is not defined in the repository?

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -48,57 +49,120 @@ describe('Create Integration - /integration (POST)', function () {
...
   });
 
-  it('should deactivated old providers', async function () {
+  it('should not allow to activate the integration without the credentials', async function () {
     const payload = {
-      providerId: 'mailgun',
-      channel: 'email',
-      credentials: { apiKey: '123', secretKey: 'abc' },
+      providerId: EmailProviderIdEnum.SendGrid,
+      channel: ChannelTypeEnum.EMAIL,
```
p-fernandez, line 118: Here we have a problem, at least of coordination. We are stating credentials are optional in the creation of the request but if active is true then they are mandatory.
As to make the validation very complicated and to show it automatically with Swagger even more, I think we need to clarify this in the description of the endpoint and the property validation so it is documented in the API.

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -48,57 +49,120 @@ describe('Create Integration - /integration (POST)', function () {
...
+    expect(data.name).to.equal('SendGrid');
+    expect(data.identifier).to.exist;
+    expect(data.providerId).to.equal(EmailProviderIdEnum.SendGrid);
+    expect(data.channel).to.equal(ChannelTypeEnum.EMAIL);
+    expect(data.active).to.equal(false);
+  });
 
-    const integrations = await integrationRepository.findByEnvironmentId(environmentId);
+  it('should allow creating the integration for different environment', async function () {
+    const prodEnv = await envRepository.findOne({ name: 'Production', _organizationId: session.organization._id });
```
p-fernandez, line 148: I think this test misses to show that we already have the same entry but for development environment. We could even compare that the data in both are the same but the _id and the identifiers.

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
```diff
@@ -36,48 +37,126 @@ describe('Update Integration - /integrations/:integrationId (PUT)', function ()
...
+      _environmentId: session.environment._id,
+    });
+    const payload = {
+      identifier: identifier2,
+      check: false,
+    };
+
+    const { body } = await session.testAgent.put(`/v1/integrations/${integrationOne._id}`).send(payload);
+
+    expect(body.statusCode).to.equal(400);
```
p-fernandez, line 67: As the same reasoning as in my previous comment before this should be a 409 Conflict.

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
```diff
@@ -36,48 +37,126 @@ describe('Update Integration - /integrations/:integrationId (PUT)', function ()
...
+      providerId: EmailProviderIdEnum.SendGrid,
       channel: ChannelTypeEnum.EMAIL,
-      credentials: { apiKey: '123', secretKey: 'abc' },
       active: false,
+      _organizationId: session.organization._id,
+      _environmentId: session.environment._id,
+    });
+
+    const payload = {
+      identifier: 'identifier2',
```
p-fernandez, line 105: Based on the test name shouldn't we remove the property `check`? Or is it needed?

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
```diff
@@ -36,48 +37,126 @@ describe('Update Integration - /integrations/:integrationId (PUT)', function ()
...
-      (i) => i.providerId.toString() === 'sendgrid' && i.channel.toString() === ChannelTypeEnum.SMS
-    );
+    const {
+      body: { data },
+    } = await session.testAgent.put(`/v1/integrations/${integrationOne._id}`).send(payload);
+
+    expect(data.name).to.eq(payload.name);
+  });
+
+  it('should allow updating the integration with just environment', async function () {
```
p-fernandez, line 139: Can we add a quick test for when updating identifier and environment while in the other environment uses that identifier? Individually both behaviours are covered with this and a previous test, but (I haven't checked the internals yet) would be a nasty surprise if when together the checks wouldn't work or the order of the checks allowed to a database error. 😄

apps/api/src/app/integrations/usecases/check-integration/check-integration-email.usecase.ts
```diff
@@ -10,10 +10,10 @@ export class CheckIntegrationEMail {
     const mailHandler = mailFactory.getHandler(
       {
         channel: command.channel,
-        credentials: command.credentials,
+        credentials: command.credentials ?? {},
         providerId: command.providerId,
       },
-      command.credentials.from
+      command.credentials?.from
```
p-fernandez, line 16: What about rather than do safechecks and fallbacks we don't throw an error when trying to get this method executed while missing the credentials? Might be something to implement in a different task though. 👍🏻

apps/api/src/app/integrations/usecases/check-integration/check-integration.command.ts
```diff
@@ -6,8 +6,10 @@ export class CheckIntegrationCommand extends EnvironmentCommand {
   @IsDefined()
   @IsString()
   providerId: string;
+
   @IsDefined()
   channel: ChannelTypeEnum;
+
   @IsDefined()
-  credentials: ICredentials;
+  credentials?: ICredentials;
```
p-fernandez, line 14: Another idea. Maybe we shouldn't allow in the UI to check integration if the credentials are not set for the integration.

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
p-fernandez: This is enough for now. Should be on the validation layer but there the complication is bigger to make it work properly.

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -59,7 +77,19 @@ export class UpdateIntegration {
 
     const updatePayload: Partial<IntegrationEntity> = {};
 
-    if (command.active || command.active === false) {
+    if (command.name) {
+      updatePayload.name = command.name;
+    }
+
+    if (command.identifier) {
```
p-fernandez, line 84:
```suggestion
    if (identifierHasChanged) {
```
I'd reuse it as we don't need to update the field if it is the same.

libs/testing/src/notification-template.service.ts
```diff
@@ -31,6 +31,8 @@ export class NotificationTemplateService {
       _environmentId: this.environmentId,
     });
 
+    console.log({ groups, feeds, layouts });
```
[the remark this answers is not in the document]
p-fernandez: This is mine for sure. 🤦🏻

apps/api/src/app/integrations/usecases/update-integration/update-integration.command.ts
```diff
@@ -1,19 +1,29 @@
...
+  @IsOptional()
+  credentials?: ICredentialsDto;
 
-  @IsDefined()
-  active: boolean;
+  @IsOptional()
+  active?: boolean;
 
   @IsDefined()
   check: boolean;
```
ainouzgali, line 29: Should check be optional? Or are we keeping it mandatory?

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
p-fernandez: No, I mean in the `CreateIntegrationRequestDto` when we validate using `class-validator`.

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -48,57 +49,120 @@ describe('Create Integration - /integration (POST)', function () {
...
+    );
+
+    expect(sendgridIntegrations.length).to.eql(2);
+
+    for (const integration of sendgridIntegrations) {
+      expect(integration.name).to.equal(payload.name);
+      expect(integration.identifier).to.exist;
+      expect(integration.providerId).to.equal(EmailProviderIdEnum.SendGrid);
+      expect(integration.channel).to.equal(ChannelTypeEnum.EMAIL);
+      expect(integration.credentials.apiKey).to.equal(payload.credentials.apiKey);
```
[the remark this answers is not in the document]
p-fernandez: Isn't channel in Chat part of the credentials? I am talking about our credentials property in the integration, just to clarify. 😅

apps/api/src/app/integrations/e2e/get-decrypted-integrations.e2e.ts
```diff
@@ -3,13 +3,20 @@ import { expect } from 'chai';
 import { ChannelTypeEnum, EmailProviderIdEnum } from '@novu/shared';
 import { IntegrationRepository } from '@novu/dal';
 
+const IS_MULTI_PROVIDER_CONFIGURATION_ENABLED = process.env.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED;
```
p-fernandez, line 6:
```suggestion
const ORIGINAL_IS_MULTI_PROVIDER_CONFIGURATION_ENABLED = process.env.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED;
```
Suggestion. I use this _trick_ to clearly state why I am doing this with the environment variables in my test suites.

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
p-fernandez: Then I would rename the test to `should allow creating the integration in the chosen environment`. Current name makes me think an integration is created in one environment already (dev, considering the usage of `prodEnv`) and we are just checking we can create the same integration in a different one.

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
```diff
@@ -36,48 +44,126 @@ describe('Update Integration - /integrations/:integrationId (PUT)', function ()
     expect(integration.credentials.secretKey).to.equal(payload.credentials.secretKey);
   });
 
-  it('should deactivate other providers on the same channel', async function () {
```
p-fernandez, line 39: 👋🏻

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
```diff
@@ -36,48 +44,126 @@ describe('Update Integration - /integrations/:integrationId (PUT)', function ()
...
+    });
+    const payload = {
+      identifier: identifier2,
+      check: false,
+    };
+
+    const { body } = await session.testAgent.put(`/v1/integrations/${integrationOne._id}`).send(payload);
+
+    expect(body.statusCode).to.equal(409);
+    expect(body.message).to.equal('Integration with identifier already exists');
```
p-fernandez, line 75: 👏🏻

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
p-fernandez: We might need to change then the default logic. As it is, seems if we don't pass anything the validation of credentials gets executed. Might be a breaking change? 😓

p-fernandez (approved):

apps/api/src/app/integrations/e2e/update-integration.e2e.ts
LetItRock: as far as I remember we want to do the validation by default

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
LetItRock: ok, got it, not sure where is the best place for that more complex logic, but I would say probably in the use-case.

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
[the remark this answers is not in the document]
LetItRock: by `channel` I meant the `webhook` - Slack channel

PR #3676, opened 2023-06-28
Reuse existing env variables that denotates cloud hosted instead of a new one

Cliftonz:

Changed as recommended in #3664


djabarovgeorge (approved):
🚀

PR #3677, opened 2023-06-28
chore(api,worker): use feature flag when selecting the integration during the send phase

LetItRock:

In this PR:
- moved the feature flag services and use-case to `application-generic`

The `SelectIntegration` use-case:
- use the `GetFeatureFlag` use-case and the `isMultiProviderConfigurationEnabled`
- added the logic to select the integration "old way" using the `GetDecryptedIntegrations`

Created also the infra PR to pass the launch darkly SDK key.


packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -26,6 +33,28 @@ export class SelectIntegration {
...
+        GetDecryptedIntegrationsCommand.create({
+          organizationId: command.organizationId,
+          environmentId: command.environmentId,
+          channelType: command.channelType,
+          findOne: true,
+          active: true,
+          userId: command.userId,
+        })
+      )[0];
+    }
```
LetItRock, line 56: `isMultiProviderConfigurationEnabled` is not enabled use the "old way" of selecting the integration with `GetDecryptedIntegrationsUsecase`

ainouzgali (approved):
Looks good

PR #3681, opened 2023-06-29
NV-2420 - Update the active integrations endpoint

djabarovgeorge:

Update the active integrations endpoint (/integrations/active) to select at least one "Environment" level connected integrations.

This endpoint used to indicate user facing widgets that he needs to connect a provider for a channel in the workflow editor. In case the user writes an e-mail but he hasn't connected any email provider to the system.


apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,21 +1,36 @@
...
-      GetDecryptedIntegrationsCommand.create({
-        environmentId: command.environmentId,
-        organizationId: command.organizationId,
-        active: true,
-        userId: command.userId,
+    const integrations = await Promise.all(
+      channelTypes.map(async (channelType) => {
+        return await this.selectIntegration.execute(
+          SelectIntegrationCommand.create({
+            environmentId: command.environmentId,
```
djabarovgeorge, line 20: we query by the environment, hope i got it right.

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,21 +1,36 @@
...
       })
     );
+
+    return integrations.filter(notEmpty);
   }
 }
+
+export function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
+  return value !== null && value !== undefined;
+}
```
djabarovgeorge, line 36: Added util to help TS understand that the type is not undefined.
https://github.com/microsoft/TypeScript/issues/45097

Maybe we can move it to shared later.

djabarovgeorge: `value is TValue` - do most of the magic there

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,21 +1,36 @@
...
+            organizationId: command.organizationId,
+            userId: command.userId,
+            channelType: channelType,
+            providerId: command.providerId,
+          })
+        );
       })
     );
+
+    return integrations.filter(notEmpty);
```
djabarovgeorge, line 30: If we are missing channel integration we wont send it.

apps/api/src/app/integrations/integrations.controller.ts
```diff
@@ -71,13 +72,11 @@ export class IntegrationsController {
...
   @ApiResponse(IntegrationResponseDto, 200, true)
-  @ApiOperation({
-    summary: 'Get active integrations',
-  })
+  @ApiOperation({ summary: 'Get active integrations', description: 'Get active selected integration per channel' })
   @ExternalApiAccessible()
   async getActiveIntegrations(@UserSession() user: IJwtPayload): Promise<IntegrationResponseDto[]> {
     return await this.getActiveIntegrationsUsecase.execute(
-      GetIntegrationsCommand.create({
+      GetActiveIntegrationsCommand.create({
```
djabarovgeorge, line 79: fixing my mistakes from the past 🤦‍♂️

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
djabarovgeorge, line 1: added a small test to the existing functuanality

djabarovgeorge, line 1: added a small test to the existing functionality

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,21 +1,36 @@
...
+      channelTypes.map(async (channelType) => {
+        return await this.selectIntegration.execute(
+          SelectIntegrationCommand.create({
+            environmentId: command.environmentId,
+            organizationId: command.organizationId,
+            userId: command.userId,
+            channelType: channelType,
+            providerId: command.providerId,
+          })
+        );
```
LetItRock, line 26: @djabarovgeorge we want to return all active integrations, and then having those active we want to mark the ones that will be picked when sending the message, so it should be like this:
```
[
{ name: 'SendGrid Old', channel: 'email', active: true, picked: false }, 
{ name: 'SendGrid New', channel: 'email', active: true, picked: true },
{ name: 'Mailgun', channel: 'email', active: true, picked: false },
{ name: 'Nexmo', channel: 'sms', active: true, picked: false },
{ name: 'Twilio', channel: 'sms', active: true, picked: true }
]
```
So like only one will be picked per channel from active integrations. Please find a better name than `picked`.

djabarovgeorge: got you, i thought only the active integration will do the work, my mistake.

do we need the `active:true` in the response object isn't it obvious because it is a get active integration usecaase

LetItRock: I think that we should return the same interface as before + additional flag, if we will return the minimized interface then it will be more breaking change

djabarovgeorge: awesome that what i made as the initial state will merge shortly

p-fernandez: One question this PR is part of the transition to adapt the current system or is the final multi provider version? Considering the API mentioned here with the active/picked is the first one but to double check.

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
```diff
@@ -0,0 +1,87 @@
...
+    expect(pushIntegration.length).to.equal(1);
+    expect(chatIntegration.length).to.equal(2);
+
+    expect(inAppIntegration[0].selected).to.equal(true);
+    expect(emailIntegration[0].selected).to.equal(true);
+    expect(smsIntegration[0].selected).to.equal(true);
+    expect(pushIntegration[0].selected).to.equal(true);
+
+    const oneSelected = chatIntegration.some((integration) => integration.selected === true);
+    const oneNotSelected = chatIntegration.some((integration) => integration.selected === false);
```
p-fernandez, line 33: Should we just count the ocurrences of the condition checked so in the future if we improve the test for multiple providers with some selected and some not it will be easier to update the test?

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
```diff
@@ -0,0 +1,87 @@
...
+    expect(oneSelected).to.equal(true);
+    expect(oneNotSelected).to.equal(true);
+  });
+
+  it('should have thrown an exception if no active integration are exist', async function () {
+    const activeIntegrations = (await session.testAgent.get(`/v1/integrations/active`)).body.data;
+    await deleteAll(activeIntegrations, integrationRepository, session);
+    const response = await session.testAgent.get(`/v1/integrations/active`);
+
+    expect(response.body.statusCode).to.equal(404);
```
p-fernandez, line 44: I think we shouldn't send a 404 as we are trying to retrieve a list of resources that match certain conditions instead of checking a certain resource.
I don't recall in our API to have any get endpoint that for lack of results in a group of resources `/resources` to return a 404 and I haven't experienced an API designed like that so I am more keen in return a 200 with an empty array as we already do.
I do understand the intention but in this case I think 404 is not the best response to send even if at the end is _we haven't found the resources we were looking for_.

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
```diff
@@ -0,0 +1,87 @@
...
+    expect(oneNotSelected).to.equal(true);
+  });
+});
+
+function splitByChannels(activeIntegrations) {
+  const inAppIntegration = activeIntegrations.filter((integration) => integration.channel === ChannelTypeEnum.IN_APP);
+  const emailIntegration = activeIntegrations.filter((integration) => integration.channel === ChannelTypeEnum.EMAIL);
+  const smsIntegration = activeIntegrations.filter((integration) => integration.channel === ChannelTypeEnum.SMS);
+  const chatIntegration = activeIntegrations.filter((integration) => integration.channel === ChannelTypeEnum.CHAT);
+  const pushIntegration = activeIntegrations.filter((integration) => integration.channel === ChannelTypeEnum.PUSH);
```
p-fernandez, line 75: Nothing against this, very clear. Though a `reduce` will merge all loops into one.

apps/api/src/app/integrations/integrations.controller.ts
```diff
@@ -77,17 +79,17 @@ export class IntegrationsController {
   @Get('/active')
   @ApiOkResponse({
     type: IntegrationResponseDto,
-    description: 'The list of active integrations belonging to the organization that are successfully returned.',
+    description: 'The list of active integrations belonging to the environment that are successfully returned.',
```
p-fernandez, line 82: I was expecting that this endpoint had to be based on organization and not in environment. Maybe we should clarify that because API can be confusing now as the generic GET /integrations endpoint is based on organization thus the breaking change.

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,20 +1,61 @@
...
+              providerId: command.providerId,
+            })
+          );
+        })
+      )
+    ).filter(notEmpty);
+  }
+}
+
+export function notEmpty<TValue>(value: TValue | null | undefined): value is TValue {
```
p-fernandez, line 59: Not particularly fond of the name: [] is empty, we have empty string.
The behaviour we try to implement is similar to Nullish Coalescing, so I'd suggest to rename it to `notNullish`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

p-fernandez (changes requested):

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
djabarovgeorge: hard to say if it is the final stage, yes we will reuse the SelectIntegration Use-case and it will work with the current logic, but IMO we will need to revisit this one in the next iterations.

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
djabarovgeorge: I was not aware that in the future we will support 'some selected' i thought there will be only one selected, in order to add such tests I will need more information regarding the future implementation, regardless I think this test will be enough for this stage.

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
djabarovgeorge: good catch! actually i was sure i saw it somewhere in the app we do, but don't find any place we throw exceptions, will update to an empty array.

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
djabarovgeorge: i think it will be a bit less readable but let's make it a bit more optimized 😉

apps/api/src/app/integrations/integrations.controller.ts
djabarovgeorge: 100% agree, did it like this because in the DOD the task was written it should be on the environment level. [REF](https://linear.app/novu/issue/NV-2420/update-the-active-integrations-endpoint)

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
djabarovgeorge: tend to agree, I am good with notNullish :)

apps/api/src/app/integrations/integrations.controller.ts
djabarovgeorge: refactored to the organization level 🫣

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(pushIntegration.length).to.equal(2);
+    expect(chatIntegration.length).to.equal(4);
+
+    expect(inAppIntegration[0].selected).to.equal(true);
+    expect(emailIntegration[0].selected).to.equal(true);
+    expect(smsIntegration[0].selected).to.equal(true);
+    expect(pushIntegration[0].selected).to.equal(true);
+
+    const oneSelected = chatIntegration.some((integration) => integration.selected === true);
+    const oneNotSelected = chatIntegration.some((integration) => integration.selected === false);
```
LetItRock, line 33:
```suggestion
    const selected = chatIntegration.filter((integration) => integration.selected);
    const notSelected = chatIntegration.filter((integration) => !integration.selected);
    
    expect(selected.length).to.equal(1);
    expect(notSelected.length).to.equal(?);
```

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(oneNotSelected).to.equal(true);
+  });
+
+  it('should have thrown an exception if no active integration are exist', async function () {
+    const activeIntegrations = (await session.testAgent.get(`/v1/integrations/active`)).body.data;
+    await deleteAll(activeIntegrations, integrationRepository, session);
+    const response = await session.testAgent.get(`/v1/integrations/active`);
+
+    expect(response.body.data).to.deep.equal([]);
+  });
```
LetItRock, line 45: this test doesn't check the exception thrown, but an empty array returned

apps/api/src/app/integrations/e2e/get-active-integration.e2e.ts
```diff
@@ -0,0 +1,107 @@
...
+    const activeIntegrations = (await session.testAgent.get(`/v1/integrations/active`)).body.data;
+    const { emailIntegration } = splitByChannels(activeIntegrations);
+
+    expect(emailIntegration.length).to.equal(3);
+
+    const oneSelected = emailIntegration.some((integration) => integration.selected === true);
+    const oneNotSelected = emailIntegration.some((integration) => integration.selected === false);
+
+    expect(oneSelected).to.equal(true);
+    expect(oneNotSelected).to.equal(true);
```
LetItRock, line 65: similar

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,20 +1,108 @@
...
+
+    const activeIntegrationChannelTypes = this.getDistinctChannelTypes(activeIntegration);
+    const selectedIntegrations = await this.getSelectedIntegrations(command, activeIntegrationChannelTypes);
+
+    return this.mapBySelectedIntegration(activeIntegration, selectedIntegrations);
+  }
+
+  private getDistinctChannelTypes(activeIntegration: IntegrationEntity[]): ChannelTypeEnum[] {
+    return activeIntegration.map((integration) => integration.channel).filter(this.distinct);
+  }
```
LetItRock, line 45:
```suggestion
 private getDistinctChannelTypes(activeIntegration: IntegrationEntity[]): ChannelTypeEnum[] {
    return Array.from(
      activeIntegration.reduce((acc, integration) => {
        acc.add(integration.channel);

        return acc;
      }, new Set<ChannelTypeEnum>())
    );
  }
```

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -1,20 +1,108 @@
...
+      return await this.selectIntegration.execute(
+        SelectIntegrationCommand.create({
+          environmentId: environmentId,
+          organizationId: command.organizationId,
+          userId: command.userId,
+          channelType: channelType as ChannelTypeEnum,
+          providerId: command.providerId,
+        })
+      );
+    });
```
LetItRock, line 102: I would combine all these in one function, it's harder to read if they are that split granually

LetItRock, line 102:
```suggestion
    return activeIntegrationChannelTypes.map((channelType) => this.selectIntegration.execute(
        SelectIntegrationCommand.create({
          environmentId: environmentId,
          organizationId: command.organizationId,
          userId: command.userId,
          channelType: channelType as ChannelTypeEnum,
          providerId: command.providerId,
        })
      ));
```

p-fernandez: I kindly disagree with this because from a perspective of functional programming. Maybe one for a tech summit to discuss style guide and agreements we would like to have regarding style. What do you think?

djabarovgeorge: initially, it was one function while selectIntegrationByEnvironment was one inline nested functions but I was not sure it will be readable for someone seeing it for the first time.

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
djabarovgeorge: Is this one will be more readable in your opinion?
i wonder if we could create a util that we could reuse in order to get a distinct array, IMO this way one filter can do the job and should be readable.

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -18,26 +20,61 @@ export class GetActiveIntegrations {
...
+            GetDecryptedIntegrationsCommand.create({
+              environmentId: environment._id,
+              organizationId: command.organizationId,
+              userId: command.userId,
+              active: true,
+            })
+          );
+        });
+
+    return (await Promise.all(activeIntegrationPromises)).flat();
```
p-fernandez: Why don't we for any case we bring all the integrations by organizationId and for `isMultiProviderConfigurationEnabled=false` we filter the integrations by environmentId?
If worried about decryption cost there is always the chance of take advantage of the caching.

p-fernandez: What logic we would need to add to `getDecryptedIntegrations`?

djabarovgeorge: I hope I am not missing something but let's break it to steps:

> @djabarovgeorge Why don't we for any case we bring all the integrations by organizationId 

This one is manageable when isMultiProviderConfigurationEnabled=true.

> and for `isMultiProviderConfigurationEnabled=false` we filter the integrations by environmentId? 

But this one will not work because if we pass envorinmentId and isMultiProviderConfigurationEnabled=false then the use-case will fetch by the environment. (this is why I iterate through the environment ids in my logic)

The other way around will be adding changes to getDecryptedIntegrations use case,
In order to get all by Organization we will need to not pass environment id, and if we won't pass environment id we would need to add a logit that will avoid extracting the feature flag (because environment will be missing.), maybe add other logic as well, but this one feels even hackier.

p-fernandez: Not sure if I am missing anything but considering that to the endpoint we always pass `organizationId` and `environmentId` in `GetActiveIntegrationsCommand` as they are injected by JWT in the endpoint, if in the use case we can do:
```
let organizationActiveIntegrations = await this.getDecryptedIntegrationsUsecase.execute(
    GetDecryptedIntegrationsCommand.create({
        organizationId: command.organizationId, // I modified the code in previous PR to by default reach by org
        userId: command.userId,
        active: true,
    })
);

const isMultiProviderConfigurationEnabled = await this.getFeatureFlag.isMultiProviderConfigurationEnabled(
      FeatureFlagCommand.create({
        userId: command.userId,
      })
    );
    
 if (!isMultiProviderConfigurationEnabled) {
    organizationActiveIntegrations = organizationActiveIntegrations.filter(integration => integration.environmentId === command.environmentId);
 }
```
So I think it is viable to do unless I am missing any nuance, that could be possible. 😕

p-fernandez: We discussed it privately and we will follow @djabarovgeorge approach until we can untangle some changes that were implemented that block a simpler approach.

PR #3682, opened 2023-06-29
NV2402 - [STACKED PR's] Tenancy CRUD API

djabarovgeorge:

This PR is a Stacked PRs - meaning that all of the changes in here were separately reviewed in the previous PRs. 

Add a CRUD endpoint for the tenants entity. The tenants for the current version should include the following information:

* Unique Identifier (key/tenantId) - A user-provided id that will be used to retrieve and manipulate tenants.
* Name - The name of the tenant

(Optional) Custom Data - An object provided by the user to contain custom tenant data. 

Tenants are attached to the environment level of the system.

So that our users will be able to manage and create tenants in our system. 

### Definition of Done
* Create tenant endpoint
* Delete tenant endpoint
* Update tenant endpoint
* Get tenant by id endpoint
* List and paginate tenants
* E2E tests for the new endpoints


apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  @Post('/')
+  @ExternalApiAccessible()
+  @ApiResponse(CreateTenantResponseDto)
+  @ApiOperation({
+    summary: 'Create tenant',
+    description: 'Create tenant under the current environment',
+  })
+  @ApiConflictResponse({
```
p-fernandez, line 113: 👏🏻 
(I don't recall seeing a test for this though)

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -0,0 +1,178 @@
...
+        environmentId: user.environmentId,
+        organizationId: user.organizationId,
+        identifier: body.identifier,
+        name: body.name,
+        data: body.data,
+      })
+    );
+  }
+
+  @Put('/:identifier')
```
p-fernandez, line 131: Considering the definition of the `UpdateTenantRequestDto` this should be a Patch. If we want a PUT we would need to send all the properties of the resource.

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -0,0 +1,178 @@
...
+  @ExternalApiAccessible()
+  @ApiResponse(UpdateTenantResponseDto)
```
p-fernandez, line 133: Missing the definition for `ApiNotFound`.

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -0,0 +1,178 @@
...
+        organizationId: user.organizationId,
+        environmentId: user.environmentId,
+        page: query.page ? Number(query.page) : 0,
+        limit: query.limit ? Number(query.limit) : 10,
+      })
+    );
+  }
+
+  @Get('/:identifier')
+  @ApiResponse(GetTenantResponseDto)
```
p-fernandez, line 87: Missing `@ApiNotFound`.

apps/api/src/app/tenant/usecases/delete-tenant/delete-tenant.usecase.ts
```diff
@@ -0,0 +1,38 @@
...
+      environmentId: command.environmentId,
+      organizationId: command.organizationId,
+      identifier: command.identifier,
+    });
+
+    if (!tenant) {
+      throw new NotFoundException(
+        `Tenant with identifier: ${command.identifier} is not exists under environment ${command.environmentId}`
+      );
+    }
```
p-fernandez, line 23: Reuse the `get-tenant` use case to avoid to have to reimplement this.

apps/api/src/app/tenant/usecases/get-tenants/get-tenants.usecase.ts
```diff
@@ -0,0 +1,34 @@
...
+      {
+        _environmentId: command.environmentId,
+        _organizationId: command.organizationId,
+      },
+      '',
+      {
+        limit: command.limit,
+        skip: command.page * command.limit,
+      }
+    );
```
p-fernandez, line 25: Both can be executed in parallel.

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+    const tenant = await this.tenantRepository.findOne({
+      _environmentId: command.environmentId,
+      identifier: command.identifier,
+    });
+
+    if (!tenant) {
+      throw new NotFoundException(
+        `Tenant with identifier: ${command.identifier} doesn't exist under environment ${command.environmentId}`
+      );
+    }
```
p-fernandez, line 20: Reuse the GetTenant use case, please.

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+    environmentId: string;
+    identifier: string;
+  }) {
+    const tenantExist = await this.tenantRepository.findOne({
+      _environmentId: environmentId,
+      identifier: identifier,
+    });
+
+    if (tenantExist) {
+      throw new BadRequestException(
```
p-fernandez, line 74: This should be a ConflictException.

p-fernandez (changes requested):

apps/api/src/app/tenant/dtos/get-tenant.response.ts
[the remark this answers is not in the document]
djabarovgeorge: what an eye! thanks!

apps/api/src/app/tenant/e2e/create-tenant.e2e.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it('should throw error if a missing tenant identifier', async function () {
+    try {
+      await createTenant({
+        session,
+      });
+
+      throw new Error('');
+    } catch (e) {
+      expect(e.response.data.message).to.be.an('array').that.includes('identifier should not be empty');
```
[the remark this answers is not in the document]
djabarovgeorge: 409 added a check for it.

apps/api/src/app/tenant/e2e/get-tenant.e2e.ts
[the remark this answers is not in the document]
djabarovgeorge: adjusted there as well. 👍

apps/api/src/app/tenant/tenant.controller.ts
djabarovgeorge: true, fixed existing one while i wend over your comments 🔥

djabarovgeorge: true, fixed an existing one while i went over your comments 🔥

apps/api/src/app/tenant/tenant.controller.ts
djabarovgeorge: what should be added to `UpdateTenantRequestDto `  if we want to make sure it is PUT? i am asking because the other params are immutable imo or am I missing something? 

Tenant Entity REF:

```
export class TenantEntity {
  _id: TenantId;
  identifier: string;
  name?: string;
  deleted?: boolean;
  createdAt: string;
  updatedAt: string;
  data?: TenantCustomData;
  _environmentId: EnvironmentId;
}
```

apps/api/src/app/tenant/tenant.controller.ts
djabarovgeorge: added.
i wonder if we can make those more auto-generated or make it mandatory on the compilation time.

apps/api/src/app/tenant/usecases/delete-tenant/delete-tenant.usecase.ts
djabarovgeorge: good idea!

apps/api/src/app/tenant/usecases/update-tenant/update-tenant.usecase.ts
djabarovgeorge: Agree already updated this one.

apps/api/src/app/tenant/tenant.controller.ts
p-fernandez: We would need to send all the entity information but the timestamps. Therefore makes the fields in the request DTO validation mandatory.
Basically the flow for a PUT is having done a GET of the resource(s) and keeping that state in the client and then modify just the values chosen and sending the full entity modified to the PUT endpoint.

p-fernandez (approved):
What a shiny API. 🌟

djabarovgeorge:
> What a shiny API. 🌟

It became shinier with your help ✨

PR #3692, opened 2023-06-30
feat: add sort function for providers

davidsoderberg:

Add sort order for providers.

### Other information
See Linear issue for order.


p-fernandez (approved):
🌟

PR #3701, opened 2023-07-02
feat: added new workflows controller

BiswaViraj:

- Created the new `workflows` controller.
- Copy pasted the notification-template controller and renamed it.
This PR is a part of stacked PRs
- https://github.com/novuhq/novu/pull/3704
- https://github.com/novuhq/novu/pull/3703
- https://github.com/novuhq/novu/pull/3702
- https://github.com/novuhq/novu/pull/3701

### DIFF
--- a/apps/api/src/app.module.ts
+++ b/apps/api/src/app.module.ts
@@ -30,6 +30,7 @@ import { PartnerIntegrationsModule } from './app/partner-integrations/partner-in
 import { TopicsModule } from './app/topics/topics.module';
 import { InboundParseModule } from './app/inbound-parse/inbound-parse.module';
 import { BlueprintModule } from './app/blueprint/blueprint.module';
+import { WorkflowsModule } from './app/workflows/workflows.module';
 
 const modules: Array<Type | DynamicModule | Promise<DynamicModule> | ForwardReference> = [
   InboundParseModule,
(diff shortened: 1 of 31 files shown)


djabarovgeorge (commented):
Is this one is a finished state?
i am asking because i am wondering if we need to update all the usecases command names as well here? or will it be as a separate task?
Additionally, i think we need to recreate the e2e test as well, so we will make sure that this module is working as expected. 

another small point i am not sure about, if we create such copy i think we should make sure that we talk about if we will maintain both copies and if not we need to alert the users we won't.

ainouzgali:
> Is this one is a finished state? i am asking because i am wondering if we need to update all the usecases command names as well here? or will it be as a separate task? Additionally, i think we need to recreate the e2e test as well, so we will make sure that this module is working as expected.
> 
> another small point i am not sure about, if we create such copy i think we should make sure that we talk about if we will maintain both copies and if not we need to alert the users we won't.

It is in another pr, there are 4 stacked

djabarovgeorge:
> > Is this one is a finished state? i am asking because i am wondering if we need to update all the usecases command names as well here? or will it be as a separate task? Additionally, i think we need to recreate the e2e test as well, so we will make sure that this module is working as expected.
> > another small point i am not sure about, if we create such copy i think we should make sure that we talk about if we will maintain both copies and if not we need to alert the users we won't.
> 
> It is in another pr, there are 4 stacked

yeah sorry just saw it as subtasks in linear.

PR #3706, opened 2023-07-03
feat(worker): store only minimal job info instead of full job

p-fernandez:

Stores in BullMQ only the minimal information of the job.

After investigations on the outage of 23rd June one of the measures to tackle the cluttering of Redis instance is to only store the minimal information (jobId, environmentId, organizationId, userId) taking in count than when pulling the job from the queue and running it we retrieve the job information again from the database already, therefore not need to store it in the Redis instance. This should help to slow down the potential filling of the Redis instance until moving to a MemoryDB instance.


apps/worker/src/app/workflow/services/workflow-queue.service.ts
```diff
@@ -99,30 +99,44 @@ export class WorkflowQueueService extends QueueService<IJobData> implements INov
...
-          _jobId: job.data._id,
-          organizationId: job.data._organizationId,
+          environmentId,
+          jobId,
+          userId,
         })
       );
     } catch (error) {
-      Logger.error('Failed to set job as completed', LOG_CONTEXT, error);
+      Logger.error(`Failed to set job ${jobId} as completed`, error, LOG_CONTEXT);
```
p-fernandez, line 118: I realised this is the right order when adding context to error loggings. Need to do a task to fix it all across the codebase.

apps/worker/src/app/workflow/services/workflow-queue.service.ts
```diff
@@ -133,43 +147,48 @@ export class WorkflowQueueService extends QueueService<IJobData> implements INov
...
-      await this.queueNextJob.execute(
-        QueueNextJobCommand.create({
-          parentId: job?.data._id,
-          environmentId: job?.data._environmentId,
-          organizationId: job?.data._organizationId,
-          userId: job?.data._userId,
-        })
-      );
-    }
+    /*
```
p-fernandez, line 165: Have to see how to solve this... it is a problem.

apps/worker/src/app/workflow/usecases/update-job-status/set-job-as-completed.usecase.ts
```diff
@@ -13,8 +13,7 @@ export class SetJobAsCompleted {
     await this.updateJobStatus.execute(
       UpdateJobStatusCommand.create({
         environmentId: command.environmentId,
-        _jobId: command._jobId,
-        organizationId: command.organizationId,
```
p-fernandez, line 17: SetJobAsCompleted doesn't need `organizationId`.

apps/worker/src/app/workflow/usecases/update-job-status/set-job-as.command.ts
```diff
@@ -1,9 +1,12 @@
 import { IsDefined } from 'class-validator';
-import { EnvironmentCommand } from '@novu/application-generic';
+import { EnvironmentLevelWithUserCommand } from '@novu/application-generic';
 
-export class SetJobAsCommand extends EnvironmentCommand {
+export class SetJobAsCommand extends EnvironmentLevelWithUserCommand {
   @IsDefined()
-  _jobId: string;
```
p-fernandez, line 6: Changing to non underscore as it is the _standard_ in the commands.

libs/shared/src/types/jobs/index.ts
```diff
@@ -0,0 +1,12 @@
...
+import { UserId } from '../user';
+
+export type JobId = string;
+
+export interface IJobData {
+  _id: JobId;
+  _environmentId: EnvironmentId;
+  _organizationId: OrganizationId;
+  _userId: UserId;
+}
```
p-fernandez, line 12: Minimal information we need to store in BullMQ to be able to run the jobs. Unfortunately for certain dependencies with certain features we can't only store the `JobId`.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -110,9 +110,17 @@ export class AddJob {
       };
       options.attempts = this.queueService.DEFAULT_ATTEMPTS;
     }
+
+    const jobData = {
+      _environmentId: job._environmentId,
+      _id: job._id,
+      _organizationId: job._organizationId,
+      _userId: job._userId,
+    };
```
p-fernandez, line 119: Minimal information we need to store in BullMQ to be able to run the jobs. Unfortunately for certain dependencies with certain features we can't only store the JobId.

apps/worker/src/app/workflow/usecases/handle-last-failed-job/handle-last-failed-job.usecase.ts
```diff
@@ -0,0 +1,71 @@
...
+  @InstrumentUsecase()
+  public async execute(command: HandleLastFailedJobCommand): Promise<void> {
+    const { jobId, error } = command;
+
+    const hasToBackoff = this.runJob.shouldBackoff(error);
+
+    if (hasToBackoff) {
+      const job = await this.jobRepository.findById(jobId);
+      if (!job) {
+        throw new Error();
```
LetItRock, line 44: `PlatformException` with the error message

apps/worker/src/app/workflow/services/workflow-queue.service.ts
```diff
@@ -99,76 +89,59 @@ export class WorkflowQueueService extends QueueService<IJobData> implements INov
...
+
+      await this.setJobAsFailed.execute(
+        SetJobAsFailedCommand.create({
+          environmentId,
+          jobId,
+          organizationId,
+          userId,
+        }),
+        error
+      );
```
LetItRock, line 129: Before we were not setting the job as failed when `hasToBackoff` flag was `true`. What is the reason that we've changed it?

apps/worker/src/app/workflow/usecases/handle-last-failed-job/handle-last-failed-job.usecase.ts
```diff
@@ -0,0 +1,71 @@
...
+  ) {}
+
+  /**
+   * TODO: IMPORTANT
+   * We are going to assume that is not very likely that a job has to be retried the maximum default attempts.
+   * That means that we could do a database call here to retrieve the job information for the ExecutionDetails
+   * as it will be only happening for a small percentage of failed jobs.
+   * This consideration is sensible as we are afraid of having to execute many database calls every time the
+   * WorkflowQueueService pulls a job to run or queue again.
+   */
```
LetItRock, line 34: I don't know if we need this TODO here as it's not accurate because we do run many database calls when the job is executed. Or maybe you know something and can share more context?

LetItRock (approved):
⭐

apps/worker/src/app/workflow/usecases/handle-last-failed-job/handle-last-failed-job.usecase.ts
p-fernandez: Maybe I should rephrase it in the way that this is trying to avoid an extra overload of database calls by not requesting the Job entity every time a job is pulled in by the worker and it fails and just when has to back off because is when we set the execution details log about it and we try to queue the next one if existing.
I wanted to have a balance and not to overload MongoDB to alleviate the Redis instance, so for this extra Mongo DB call that we would need to operate this part of the flow, might not be even executed because the case scenario of jobs reaching the maximum attempts and that have to back off should be reduced compared to other outcomes of the flow.

apps/worker/src/app/workflow/usecases/handle-last-failed-job/handle-last-failed-job.usecase.ts
p-fernandez: I forgot about setting this properly. Good point highlighting it.

apps/worker/src/app/workflow/services/workflow-queue.service.ts
p-fernandez: I understood that we were always setting it as failed even later in `handleLastFailedWebhookFilter` but I forgot to take mentally in count that is for the maximum amount of attempts. I need to fix this. Thanks for noticing it!

LetItRock (approved):
looks good 🚀

PR #3711, opened 2023-07-04
Adding inkeep widget

nick-inkeep:

Three major things:
#### Move to Docusorous V3
This introduces support for React 18 and MDX V2. 

Current release is [alpha](https://github.com/facebook/docusaurus/releases/tag/v3.0.0-alpha.0)

Few key things that I did in order to support this:
- upgraded dependencies to use the alpha version, new mdx-js version, and react 18.
- renamed all docs files to "mdx"
- addressed certain invalid mdx formatting issues. For example, all JSX must be wrapped with { } now. Links, like <https://mydomain.com> that aren't HTML tags must either be ticked marked `` or use []().

#### Addressed previously existing yarn 'check' issues
As part of checking whether there were any issues introduced by V3, I did `yarn check` and addressed many of the issues. Here were the original ones:

Add the Inkeep search widget. It only supports React 18 currently, hence the upgrade to V3.

image

image


p-fernandez:
Thanks for your contribution.
Could you please elaborate more why this PR is needed?

unicodeveloper:
@p-fernandez We have a contract with inKeep. So now we are trying to add it to the docs to test how it works and also monitor how valuable it is for us.

PR #3713, opened 2023-07-04
feat: search empty state

ainouzgali:

Display interactions with the search UI
- don't display tabs for a channel if no provider related to the channel founded
- display empty state when there are no providers matching the search criteria

Also refactored the naming of the list so it will be easier to iterate over the providers list by channel type. Will also keep the appearance order the same in multiple places.

Screen Shot 2023-07-04 at 14 50 01
Screen Shot 2023-07-04 at 14 50 35


apps/web/src/pages/integrations/components/multi-provider/SelectProviderSidebar.tsx
```diff
@@ -1,7 +1,7 @@
 import { useMemo, useState } from 'react';
```
LetItRock, line 1: could you please export the image `empty-provider-search.png` as the SVG, so it will weight less and scale better?

apps/web/src/pages/integrations/components/multi-provider/SelectProviderSidebar.tsx
```diff
@@ -126,36 +129,20 @@ export function SelectProviderSidebar() {
...
+          {CHANNELS_ORDER.map((channelType) => {
+            return (
+              <ListProviders
+                key={channelType}
+                selectedProvider={selectedProvider}
+                onProviderClick={onProviderClick}
+                channelProviders={providersList[channelType]}
+                channelType={channelType}
+              />
+            );
```
LetItRock, line 144: when no providers, we don't have to render these at all... rn they are hidden, but we do waste this way react time and memory resources

LetItRock (approved):

PR #3714, opened 2023-07-04
feat: add name field to update sidebar

ainouzgali:

UX consistency in the product. We display the Input "Name" of a Workflow in the WF settings. 
Input "Name" is displayed in the update integration sidebar

Screen Shot 2023-07-04 at 14 59 31

### DIFF
--- a/apps/web/src/pages/integrations/UpdateProviderPage.tsx
+++ b/apps/web/src/pages/integrations/UpdateProviderPage.tsx
@@ -227,7 +227,13 @@ export function UpdateProviderPage() {
             }}
             render={({ field }) => (
               <InputWrapper>
-                <Input {...field} required label="Name" error={errors.name?.message} />
+                <Input
+                  {...field}
+                  value={field.value ? field.value : selectedProvider?.displayName}
+                  required
+                  label="Name"
+                  error={errors.name?.message}
+                />
               </InputWrapper>
             )}
           />
--- a/apps/web/src/pages/integrations/UpdateProviderPage.tsx
+++ b/apps/web/src/pages/integrations/UpdateProviderPage.tsx
@@ -220,6 +220,19 @@ export function UpdateProviderPage() {
           />
           <Controller
             control={control}
+            name="name"
+            defaultValue={''}
+            rules={{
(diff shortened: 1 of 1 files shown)


LetItRock (approved):

apps/web/src/pages/integrations/UpdateProviderPage.tsx
```diff
@@ -218,6 +218,19 @@ export function UpdateProviderPage() {
...
+          <Controller
+            control={control}
+            name="name"
+            defaultValue={''}
+            rules={{
+              required: 'Required - Instance name',
+            }}
+            render={({ field }) => (
+              <InputWrapper>
+                <Input {...field} required label="Name" error={errors.name?.message} />
```
djabarovgeorge, line 230: i think we are missing the fallback for the name, in case the migration is not run yet.

djabarovgeorge (commented):
Looks great but i am not sure i see if the name is stay updated after the update click, (same for the identifier). 
Is it meant to be this way?

ainouzgali:
> i am not sure i see if the name is stay updated after the update click, (same for the identifier).
> Is it meant to be this way?

What do you mean by 'stay updated after the update click' ?

djabarovgeorge:
> > i am not sure i see if the name is stay updated after the update click, (same for the identifier).
> > Is it meant to be this way?
> 
> What do you mean by 'stay updated after the update click' ? @djabarovgeorge

i run it locally and the name was not updating after the click, you have not seen this behavior in your machine?וןן

apps/web/src/pages/integrations/UpdateProviderPage.tsx
ainouzgali: Right! Good catch

ainouzgali:
> > > i am not sure i see if the name is stay updated after the update click, (same for the identifier).
> > > Is it meant to be this way?
> > 
> > 
> > What do you mean by 'stay updated after the update click' ? @djabarovgeorge
> 
> i run it locally and the name was not updating after the click, you have not seen this behavior in your machine?

No, doesn't happen to me😬

PR #3716, opened 2023-07-04
[NV-2437] bug - Union received device tokens

djabarovgeorge:

If user store tokens using set credentials ["token_1", "token_1", "token_2", "token_2"]

we store it as it is ["token_1", "token_1", "token_2", "token_2"]

But these should be stored as ["token_1", "token_2" ]

Having duplicate tokens in subscriber credentials results in duplicate push notifications to the same device.


apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.usecase.ts
```diff
@@ -118,25 +118,55 @@ export class UpdateSubscriberChannel {
...
+    return {
+      _integrationId: updatePayload._integrationId || existingChannel._integrationId,
+      providerId: updatePayload.providerId || existingChannel.providerId,
+      credentials: { ...existingChannel.credentials, ...updatePayload.credentials, ...{ deviceTokens } },
+    };
+  }
+
+  private unionDeviceTokens(existingDeviceTokens: string[], updateDeviceTokens: string[]): string[] {
+    // in order to not have breaking change we will support [] update
+    if (updateDeviceTokens?.length === 0) return [];
```
ainouzgali, line 163: What if updateDeviceTokens is 0 but he does have existingDeviceTokens. Wouldn't this override the existing ones when used in the mapping?

apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.usecase.ts
```diff
@@ -118,25 +118,55 @@ export class UpdateSubscriberChannel {
...
 
+  private mapChannel(
+    updatePayload: Partial<IChannelSettings>,
+    existingChannel: IChannelSettings,
+    deviceTokens: string[]
+  ): IChannelSettings {
+    return {
+      _integrationId: updatePayload._integrationId || existingChannel._integrationId,
+      providerId: updatePayload.providerId || existingChannel.providerId,
+      credentials: { ...existingChannel.credentials, ...updatePayload.credentials, ...{ deviceTokens } },
```
ainouzgali, line 157:
```suggestion
      credentials: { ...existingChannel.credentials, ...updatePayload.credentials, deviceTokens },
```
I believe that should have the same result

apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.usecase.ts
```diff
@@ -118,25 +118,55 @@ export class UpdateSubscriberChannel {
...
+    };
+  }
+
+  private unionDeviceTokens(existingDeviceTokens: string[], updateDeviceTokens: string[]): string[] {
+    // in order to not have breaking change we will support [] update
+    if (updateDeviceTokens?.length === 0) return [];
+
+    const merged = [...existingDeviceTokens, ...updateDeviceTokens];
+
+    return union(merged);
```
ainouzgali, line 167: Maybe instead of union we could use the set trick
` [...new Set([...existingDeviceTokens, ...updateDeviceTokens])]`

Not sure what has better performance

djabarovgeorge: actually, it should have better performance, i was not sure regarding the readability. but i tend to think that this one is more common therefore more readable, thanks for the comment ill update it :)

apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.usecase.ts
djabarovgeorge: TRUE :)

apps/api/src/app/subscribers/usecases/update-subscriber-channel/update-subscriber-channel.usecase.ts
djabarovgeorge: @ainouzgali 
do you mean it the array is empty it will erase the tokens?

ainouzgali: Yes, because if the update array is empty, you return an empty array into deviceTokens. But in the mapChannel function you use deviceToken after the spreading of the other credentials. So it will go through with deviceToken: [] even if existing deviceTokens has tokens

djabarovgeorge: yes you got it right, at the moment the API is supporting it and i am not sure if someone is using this 'hack' and in order to not break existing functionality i added this one.
once we will add a delete endpoint for the credentials we could remove this if. already added this note to task 2372.

ainouzgali: ohhhh its to update for delete !!! I missed the purpose 😂 I thought it was just for a case of empty update 
Sorry !

ainouzgali (approved):
❤️

PR #3717, opened 2023-07-04
feat: add unique handlebar helper

jainpawan21:

1. Add `unique` handlebar helper and its test.

To remove duplicates in array iteration

Usage Example:- 

Custom code
```
{{#each (unique names "name")}}
--<b>{{this}}</b>--- 
{{/each}}
```

Input:- 

```
{
	"names": [{
		"name": "Name 1",
		"age": "30"
	}, {
		"name": "Name 2"
	}, {
		"name": "Name 1"
	}]
}
```

HTML output:- 

` --Name 1----Name 2--`

### DIFF
--- a/packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
+++ b/packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
@@ -50,6 +50,21 @@ Handlebars.registerHelper(
   }
 );
 
+Handlebars.registerHelper(
+  HandlebarHelpersEnum.UNIQUE,
+  function (array, property) {
+    if (!Array.isArray(array)) return '';
+
+    return array
+      .map((item) => {
+        if (item[property]) {
+          return item[property];
+        }
+      })
+      .filter((value, index, self) => self.indexOf(value) === index);
+  }
+);
+
 @Injectable()
 export class CompileTemplate {
   async execute(command: CompileTemplateCommand): Promise<string> {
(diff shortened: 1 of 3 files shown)


LetItRock (approved):
⭐

packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
```diff
@@ -50,6 +50,21 @@ Handlebars.registerHelper(
...
 );
 
+Handlebars.registerHelper(
+  HandlebarHelpersEnum.UNIQUE,
+  function (array, property) {
+    if (!Array.isArray(array)) return '';
+
+    return array
+      .map((item) => {
+        if (item[property]) {
```
ainouzgali, line 60: @jainpawan21 wdyt of also making it work for unique values of an array of strings?

ainouzgali, line 60: If the object doesn't have that property we should handle it somehow

packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
```diff
@@ -50,6 +50,21 @@ Handlebars.registerHelper(
...
+          return item[property];
+        }
+      })
+      .filter((value, index, self) => self.indexOf(value) === index);
```
ainouzgali, line 64: If I am getting it right, we will return an array of only the property we filtered the distinct of. I feel we should return the whole object. 
For example:
`array: [{name:1,var:1},{name:1,var:1},{name:2,var:1}]` 
should return `
`[{name:1,var:1},{name:2,var:1}]`
instead of ['1','2']

what do you think?

PR #3718, opened 2023-07-04
v0.16.3

LetItRock:


jainpawan21 (approved):

LetItRock:
this PR: https://github.com/novuhq/novu/pull/3743 should be also included in the patch release

LetItRock:
the #3743 included

packages/notification-center-angular/projects/notification-center-angular/src/lib/notification-center.component.ts
```diff
@@ -58,7 +58,7 @@ export class NotificationCenterComponent {
   @Input() tabs: NotificationCenterComponentProps['tabs'];
   @Input() showUserPreferences: NotificationCenterComponentProps['showUserPreferences'];
   @Input() allowedNotificationActions: NotificationCenterComponentProps['allowedNotificationActions'];
-  @Input() popover: NotificationCenterComponentProps['popover'];
+  @Input() popoverConfig: NotificationCenterComponentProps['popoverConfig'];
```
scopsy, line 61: @jainpawan21 @LetItRock just noticed this as a breaking change, so we can't release this as part of a patch :( How complex will it be to still have the old input along side the new one and mark it as deprecated? Soo all the logic stays the same we just have a duplicate property named `popover`?

LetItRock: @scopsy This is a breaking change for the Web Component, Angular and Vue... 
The thing is that `popover` prop is a new attribute that can be added on any dom element: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover
And because of that our Web Component `popover` prop is not working, that's why we needed to change it to `popoverConfig`. Also, Angular/Vue does use underneath the Web Component, so we need to change it there too.

scopsy: Yeah that totally makes sense, my suggestion for the previous working browsers that some people might already use the web component for the `popover` will continue to behave as previously, and we will recommend all users to move to the new `popoverConfig` prop until we officially deprecate the old `popover` field. Hope this better explains it

jainpawan21: added popover and marked it as deprecated
https://github.com/novuhq/novu/pull/3752

PR #3722, opened 2023-07-05
fix: aggregation of events by digest key, fix nested digest key

ainouzgali:

1. **Fix - digestKey events don't aggregate the digested payloads**:
The digested events were added only for the first digestKey runs. All other digestKeys had the first event payload but marked all the others as digested.

For example: 
digestKey: 1 , trigger 3 times. get one email with the payload of all 3 triggers (as expected) 
digestKey: 2 , trigger 3 times. get only one email with the payload of only the first trigger

2. **Fix - no digestKey functionality for a nested digestKey path**
Giving a nested path for the digestKey value resulted in the normal digest behaviour, ignores the digest key values

Fix bug of digestKey

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import { JobStatusEnum } from '@novu/dal';
-import { StepTypeEnum } from '@novu/shared';
-import { InstrumentUsecase } from '@novu/application-generic';
+import { IDigestRegularMetadata, StepTypeEnum } from '@novu/shared';
+import { DigestFilterSteps, InstrumentUsecase } from '@novu/application-generic';
 
 import { SendMessageCommand } from '../send-message.command';
 import { GetDigestEvents } from './get-digest-events.usecase';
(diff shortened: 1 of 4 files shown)


libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -101,6 +108,7 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
       type: StepTypeEnum.TRIGGER,
       _environmentId: environmentId,
       _subscriberId: subscriberId,
+      ...(digestKey && { [`payload.${digestKey}`]: digestValue }),
```
ainouzgali, line 111: The issue was that the jobs that should have been digested in the second digest key were retrieved here, for the first digest key. So, when the run for the second digestKey was getting the events, they were not included for him.

packages/application-generic/src/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -74,11 +74,11 @@ export class DigestFilterSteps {
     }
 
     try {
-      let result;
+      let result = payload;
       const keys = path.split('.');
 
       for (const key of keys) {
-        result = payload[key];
+        result = result[key];
```
ainouzgali, line 81: Fix the nested

apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
```diff
@@ -22,6 +26,7 @@ export class GetDigestEventsBackoff extends GetDigestEvents {
       status: JobStatusEnum.COMPLETED,
       type: StepTypeEnum.TRIGGER,
       _environmentId: command.environmentId,
+      ...(digestKey && { [`payload.${digestKey}`]: digestValue }),
```
ainouzgali, line 29: Same fix for the backoff digest

LetItRock (approved):
💎

PR #3723, opened 2023-07-05
feat: add update sidebar for novu providers

davidsoderberg:

Add info for Novu providers in update provider sidebar

Screenshot 2023-07-05 at 13 43 17


apps/web/src/pages/integrations/UpdateProviderPage.tsx
```diff
@@ -29,10 +37,22 @@ interface IProviderForm {
   identifier: string;
 }
 
+const Info = ({ provider, environments }) => (
```
davidsoderberg, line 40: Maybe this should be in its own file even if it is only used here and it feels like this is the only place it will be used it?

apps/web/src/pages/integrations/UpdateProviderPage.tsx
```diff
@@ -152,6 +172,45 @@ export function UpdateProviderPage() {
     return null;
   }
 
+  if (
+    SmsProviderIdEnum.Novu === selectedProvider.providerId ||
+    EmailProviderIdEnum.Novu === selectedProvider.providerId
+  ) {
+    return (
```
davidsoderberg, line 179: Tried to reuse as many components as possible

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -134,7 +137,19 @@ export class GetNovuIntegration {
       from: process.env.NOVU_SMS_INTEGRATION_SENDER,
     };
 
-    return item;
+    return this.mapNovuItem(item, command);
+  }
+
+  private mapNovuItem(
```
davidsoderberg, line 143: Adding these as values until the exists in the database.

davidsoderberg:
The images for novu sms provider did not exists before so created them now :)

packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -50,19 +56,58 @@ export class GetDecryptedIntegrations {
...
+          environmentId: command.environmentId,
+          userId: command.userId,
+        })
+      );
+
+      return novuIntegration ? [novuIntegration] : [];
     }
 
-    const novuIntegration = await this.getNovuIntegration.execute(
+    const novuSmsIntegration = await this.getNovuIntegration.execute(
```
davidsoderberg, line 85: This replaces the hardcoded providers in frontend

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.command.ts
```diff
@@ -7,4 +7,8 @@ export class GetNovuIntegrationCommand extends EnvironmentWithUserCommand {
   @IsOptional()
   @IsEnum(ChannelTypeEnum)
   channelType?: ChannelTypeEnum;
+
+  @IsOptional()
+  @IsBoolean()
+  ignoreActiveCount? = false;
```
p-fernandez, line 13:
```suggestion
  ignoreActiveCount = false;
```
Not sure if this can be done here but if setting the default value you don't need to add the question mark.

apps/web/src/pages/integrations/useProviders.ts
```diff
@@ -72,17 +76,17 @@ function initializeProvidersByIntegration(integrations: IntegrationEntity[]): II
     fcmFallback(integrationItem, clonedCredentials);
 
     return {
-      providerId: providerItem.id,
+      providerId: integrationItem.providerId || providerItem?.id,
```
p-fernandez, line 79: I do not understand why in this return some of the fallbacks are prioritised by `integrationItem` and others `providerItem`. Could you share the reasoning of it? I am confused, sorry. 😞

apps/web/src/pages/integrations/UpdateProviderPage.tsx
p-fernandez: There is another PR where I think @LetItRock implements something similar. 
https://github.com/novuhq/novu/blob/e97f83e0482882d1b218c9d6cce71e66ff5571b1/apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx#L41

https://github.com/novuhq/novu/pull/3740

So I agree with you and I push for it to be moved it to its own component so it can be reused for you both. 👏🏻

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -23,13 +23,13 @@ describe('Create Integration - /integration (POST)', function () {
...
-    const emailIntegrations = integrations.filter(
+    const emailIntegrations: any[] = integrations.filter(
       (searchIntegration) => searchIntegration.channel === ChannelTypeEnum.EMAIL
     );
 
-    expect(emailIntegrations.length).to.eql(2);
+    expect(emailIntegrations.length).to.eql(3);
 
-    for (const emailIntegration of emailIntegrations) {
+    for (const emailIntegration of emailIntegrations.splice(1, 1)) {
```
p-fernandez, line 32: I am sorry, can't understand what you try to do here. 😞

davidsoderberg: We do it because the last one will be Novu free provider and we do not like to check it here.

apps/web/src/pages/integrations/useProviders.ts
davidsoderberg: Because Novus free provider is not a part of the novu providers list and then we need to use it for the integration item we get from the backend.

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.command.ts
davidsoderberg: It needs to be there otherwise typescript thinks it is always needed to be provided.

apps/api/src/app/integrations/e2e/create-integration.e2e.ts
davidsoderberg: I changed it to make more sense :)

apps/api/src/app/integrations/usecases/get-integrations/get-integrations.usecase.ts
```diff
@@ -13,6 +13,7 @@ export class GetIntegrations {
       GetDecryptedIntegrationsCommand.create({
         organizationId: command.organizationId,
         userId: command.userId,
+        environmentId: command.environmentId,
```
ainouzgali, line 16: Why did you add the environment? As I understand it we do want all integrations, from both environments

davidsoderberg: because the feature flag usecase requires it :)

ainouzgali: Oops. Sorry ;)

p-fernandez: @davidsoderberg mind to leave a TODO comment to remind us to remove once we sunset the pre feature functionality?

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.command.ts
p-fernandez: I am so disappointed it behaves differently as the arguments in a function. 😞

p-fernandez (approved):

PR #3724, opened 2023-07-05
feat: add actor access to notification templates

### LINKED ISSUE #3583: 🚀 Feature: Access actor information from notification templates
### 🔖 Feature description

Allow notification templates to access information provided by the actor's subscriber object.

I want to add a simple connection request notification - "John Smith would like to connect". At present, I'm sending { firstName, lastName } as part of the payload, but this is duplicate data as this is also available via the actor (I'm sending { actor: { ... } } alongside the payload: { ... }. It doesn't appear in the default variables when creating a notification template, but it seems like an obvious thing to be able to use, as {{ actor.firstName }}, for example.

Currently, actor information isn't passed into the template builders, it should be a straightforward change to make this available.

### 🔄️ Additional Information

I tried various variable accessors in the template builder UI but none of them worked. On submitting a query to Discord, I was asked to create an issue for this use case.

richardbutler:

When using the Novu console, a user may access the actor using `actor.`-scoped variables, e.g. `actor.firstName`. Before this change, this information was not available to the template.

`actor` is also added to the System Variables whitelist and treated as a special case, similar to `subscriber`.

Closes #3583

### Other information

I'm not 100% sure I've managed to successfully run all tests related to this change as the development workflow is a little confusing and I wasn't able to setup/run all test scripts successfully.


packages/application-generic/src/services/cache/key-builders/entities.ts
```diff
@@ -23,6 +23,21 @@ const buildSubscriberKey = ({
...
+  _environmentId: string;
+}): string =>
+  buildCommonKey({
+    type: CacheKeyTypeEnum.ENTITY,
+    keyEntity: CacheKeyPrefixEnum.SUBSCRIBER,
+    environmentId: _environmentId,
+    identifierPrefix: IdentifierPrefixEnum.ID,
+    identifier: _id,
+  });
+
```
ainouzgali, line 40: I understand why you took this approach :) But adding another caching function is not preferred for this particular issue. Handling cache is delicate, we would need to add invalidation at places, and also it will not use the already cached data we have by the subscriberId. So, as it is done only for the actor (the only place its cached by _id) it will probably not have significant affect. Yet, using the one with subscriberId will have so much affect, as we know the actor has already been fetched once in trigger-event.
Let's go the other way, and add the external subscriberId to the job entity. I think we should call it actorSubscriberId. 
it will be the equivalent of :
 _id -> _actorId and subscriberId -> actorSubscriberId
 Let me know if that makes sense to you

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -161,7 +161,8 @@ export class TriggerEvent {
             to: subscriber,
             transactionId: command.transactionId,
             userId,
-            ...(actor && actorProcessed && { actor: actorProcessed }),
+            ...(actor &&
+              actorProcessed && { actor, actorSubscriber: actorProcessed }),
```
ainouzgali, line 165: Here, revert back please to:
`...(actor && actorProcessed && { actor: actorProcessed }),`

The actorProcessed already has the subscriberId (which is the only thing in the actor object)

packages/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -71,6 +71,8 @@ export class CreateNotificationJobs {
       payload: command.payload,
       expireAt: this.calculateExpireAt(command),
       channels,
+      ...(command.actorSubscriber && { _actorId: command.actorSubscriber._id }),
+      ...(command.actor && { actor: command.actor }),
```
ainouzgali, line 75: Let's remove that fro notification

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -165,6 +176,11 @@ export class SendMessageEmail extends SendMessageBase {
       overrides,
       templateIdentifier: command.identifier,
       _jobId: command.jobId,
+      ...(actor &&
+        actor.type !== ActorTypeEnum.NONE && {
+          actor,
+          _actorId: command.job?._actorId,
+        }),
```
ainouzgali, line 183: No need to be saved on message

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -90,6 +99,7 @@ export class SendMessageEmail extends SendMessageBase {
     if (!emailChannel.template) throw new PlatformException('Email channel template not found');
 
     const email = command.payload.email || subscriber.email;
+    const { actor } = emailChannel.template;
```
ainouzgali, line 102: not needed :)

ainouzgali (commented):
@richardbutler thank you so much! You did amazing work. Sorry for my delay. I left you some questions and suggestions. 
The main things for me is the reason for saving the actor on the notification. I don't see why you needed it except for populating for the feed, which could be done from the job. 
Another thing, is that I would like to take the approach of adding the actor subscriberId to the job entity instead of creating a new caching function. 
Also, some tests will need to be added - but let's first concentrate on the implementation :)
Let me know if you have any questions or need any assistance⭐️

packages/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -109,7 +111,9 @@ export class CreateNotificationJobs {
         type: step.template.type,
         providerId: providerId,
         expireAt: notification.expireAt,
-        ...(command.actor && { _actorId: command.actor?._id }),
+        ...(command.actorSubscriber && {
```
ainouzgali, line 114: My intention is , just to have an example, is to be used like this in the job: 
```
_actorId: command.actor._id,
actorSubscriberId: command.actor.subscriberId 
```

scopsy:
Closing as stale for now 🙏

PR #3725, opened 2023-07-05
Refactor update provider page to drawer

djabarovgeorge:

The sidebar length is the length of the integrations list. If the page is long the form will be long. 

Why? (Context)

UX debt, consistency. All right panels in Novu should behave same.

### Definition of Done

* The eight should be max height of the viewport minus paddings. 
* Panel should stick to the top while the list is scrolling. 
* UX should be the same as the integrations store modal.

https://www.loom.com/share/e0644f7bdd974e24ba7a2b7f2a99297c
https://www.loom.com/share/ffeb4d39f2e94fc0a1ad620548b1049d


ainouzgali:
First, this is a css miracle. Great job🤩
About [this loom](https://www.loom.com/share/e0644f7bdd974e24ba7a2b7f2a99297c) I think the meaning was that it will be the size if the page container, but nit sure how it will be best to deal with that. Mayne we should - in general, all across the project - have the pages at the same consistent height and not be dependent on amount of content. 

I would consult @nknyshov about how to solve the loom I linked.

ainouzgali:
Also, this Drawer should apply to SelectProviderSidebar and CreateProviderPage :)

LetItRock:
> Also, this Drawer should apply to SelectProviderSidebar and CreateProviderPage :)

hey, I'll consult how the sidebar should behave with Nikolay and solve this one in my next PR :)

PR #3730, opened 2023-07-06
feat: Helm Repo

aexei:

Use the branch charts as a helm repo.

This is how to install novu via helmet without cloning the github repo

Please create an empty branch "charts".
The branch could additionally be published via github pages.


.github/workflows/release-helm.yml
```diff
@@ -0,0 +1,18 @@
...
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: stefanprodan/helm-gh-pages@v1.7.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          charts_dir: "docker/kubernetes"
+          branch: charts
```
p-fernandez, line 18: Does the branch has to be called `charts`? If the name could be chosen I'd go for `helm`.

docker/kubernetes/helm/templates/api/deployment.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: {{ printf "%s-%s" .Release.Name "api"  }}
   namespace: {{ include "common.names.namespace" . | quote }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
-    app.kubernetes.io/component: novu-api
+    app.kubernetes.io/component: {{ printf "%s-%s" .Release.Name "api"  }}
```
p-fernandez, line 7: These changes belongs to other PR.

p-fernandez (changes requested):
I really liked you separated all the changes in different PRs (even if minimal changes) but helps to people like me less familiarised with Helm. Though this PR includes 2 different PRs.
Also I have a question regarding the branch suggested to create.
One thing regarding that is that we would need to add a workflow in the pipeline to keep `helm` aligned with either `main` or `next` branch, depending on what are the expectations to have with Helm.

.github/workflows/release-helm.yml
aexei: yes that doesn't matter, most use charts

docker/kubernetes/helm/templates/api/deployment.yaml
aexei: oh crap that was not intentional, apparently I made a mistake. Should I open a new PR or can you pick it up only in part?

p-fernandez: Fixed. 😄

p-fernandez (approved):

PR #3731, opened 2023-07-06
fix: k8s component labels hard coded

aexei:

In the k8s chart, the labels "app.kubernetes.io/component" are now generated, taking into account the ".Release.Name". 

If the release name did not correspond to "novu", the selector labels of the services were wrong.


p-fernandez (approved):
Looks good but I am wondering if you should also update the `tls-secret.yaml` files to match the component labels also. What do you think?

aexei:
> Looks good but I am wondering if you should also update the `tls-secret.yaml` files to match the component labels also. What do you think?

I think the labels fit there, since we only changed the labels of the pods and not the servic

PR #3732, opened 2023-07-06
fix: helm values.yaml missing web endpoints

aexei:

In the values.yaml, the configuration for the points that are used by the web deployment is missing.

Due to its absence, the API cannot always be accessed from the web component.


p-fernandez (approved):
LGTM! Thanks for the contribution.
🌟

PR #3733, opened 2023-07-06
fix: web deployment using wrong NODE_ENV

aexei:
The web deployment accesses the node_env variable of the api and not its own.


p-fernandez (approved):
Good catch! 🌟

PR #3734, opened 2023-07-06
fix: digest and delay values are reset to default values on update

ainouzgali:

fix - update resets to default values of digest or delay metadata.

Reproducing steps: 
1. add digest step
2. change any value of regular digest
3. click update 

https://www.loom.com/share/038f03d3f7ab4695b70636e69c5be858

The problem was that we reset the steps in the form from the response we got from the update. We have different field names for the digest metadata in the backend vs the frontend.

### DIFF
--- a/apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
+++ b/apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
@@ -192,10 +192,8 @@ const TemplateEditorFormProvider = ({ children }) => {
           },
         });
         setTrigger(response.triggers[0]);
-        reset({
-          ...form,
-          steps: response.steps,
-        });
+        reset(mapNotificationTemplateToForm(response));
+
         if (showMessage) {
           successMessage('Trigger code is updated successfully', 'workflowSaved');
         }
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
```diff
@@ -192,10 +192,8 @@ const TemplateEditorFormProvider = ({ children }) => {
           },
         });
         setTrigger(response.triggers[0]);
-        reset({
-          ...form,
-          steps: response.steps,
-        });
+        reset(mapNotificationTemplateToForm(response));
```
ainouzgali, line 195: @BiswaViraj could you please make sure the bug you described [here](https://github.com/novuhq/novu/pull/3524/files#r1208925796) is still resolved? 
I was not able to reproduce what you described

LetItRock (approved):

apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
LetItRock: I've also checked and wasn't able to reproduce it.

PR #3736, opened 2023-07-06
fix(providers): Sparkpost provider bugs

### LINKED ISSUE #3670: [NV-2500] 🐛 Bug Report: `credentials.region must be a string` error in SparkPost email integration
### 📜 Description

In SparkPost provider integration for email, when we try to add our credentials, and then click on connect, it throws `credentials.region must be a string` error

### 👟 Reproduction steps

1. Go to Integration Store
2. Scroll to SparkPost provider in the Email channel
3. Click on connect
4. Add all Credentials
5. Region toggle can be either EU or Non-EU
6. Click on connect button of modal
7. There is an error

### 👍 Expected behavior

It should not throw an error

https://github.com/novuhq/novu/assets/39362422/13eb8979-8745-4e8c-a69f-380a1e4e13b3

_No response_

None

<sub>[NV-2500](https://linear.app/novu/issue/NV-2500/🐛-bug-report-credentialsregion-must-be-a-string-error-in-sparkpost)</sub>

michaldziuba03:

This PR introduce some **SparkPost** provider fix, because currently this provider is broken. It fixes 2 bugs:

#### 1. Credentials.eu must be a string error
When you try to set EU to true (enable switch in UI) and save configuration - you will get an error:

sparkpost-eu-switch

sparkpost-eu-error

It happens because switch in UI indicates boolean value but string is expected (by validator).

What I decided to do is add new credentials field: `eu` of boolean type. I wanted to reuse existing any field but I really cannot find any that would make sense.

I cannot change `credentials.region` to boolean because it is used by other integrations as string value. 
I cannot transform boolean to string on frontend because it may break a flow defined in [docs](https://docs.novu.co/community/create-provider/).

Let me know if new `credentials.eu` is fine or I should reuse `credentials.region` at all costs.

#### 2. SparkPost checks always fail

It happens because SparkPost provider tries to iterate over `attachements` array but it can be `undefined`. What I did is basically iterate only when attachements are defined.

Closes #3670


apps/api/src/app/integrations/usecases/check-integration/check-integration.usecase.ts
```diff
@@ -1,7 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import { CheckIntegrationEMail } from './check-integration-email.usecase';
 import { CheckIntegrationCommand } from './check-integration.command';
-import { ChannelTypeEnum, ICredentialsDto } from '@novu/shared';
```
michaldziuba03, line 4: remove unused import

providers/sparkpost/src/lib/sparkpost.provider.ts
```diff
@@ -41,7 +41,7 @@ export class SparkPostEmailProvider implements IEmailProvider {
 
     const files: Array<{ name: string; type: string; data: string }> = [];
 
-    attachments.forEach((attachment) => {
+    attachments?.forEach((attachment) => {
```
michaldziuba03, line 44: `attachements` can be undefined - when Novu checks integration it ALWAYS fails

providers/sparkpost/src/lib/sparkpost.provider.ts
```diff
@@ -62,6 +62,7 @@ export class SparkPostEmailProvider implements IEmailProvider {
 
     return {
       id: sent.results.id,
+      date: new Date().toISOString(),
```
michaldziuba03, line 65: I see other mail providers returns current date, so I decided to do it here as well.

apps/api/src/app/integrations/dtos/credentials.dto.ts
```diff
@@ -50,6 +50,12 @@ export class CredentialsDto implements ICredentials {
   @IsOptional()
   region?: string;
 
+  @ApiPropertyOptional()
+  @TransformToBoolean()
+  @IsBoolean()
+  @IsOptional()
+  eu?: boolean;
```
p-fernandez, line 57: Indeed adding the boolean of this region is a short term fix that will easily become a problem if Sparkpost for example would enable more regions.
I know it is painful and a bit tricky but I think we should handle this in the following way:
- in the Web app we should have a dropdown selector with both allowed regions (Default, EU) and reuse the region property.
- in the provider check which region is being set and depending on that set the EU URL if that region is selected.

Our design system has already the Dropdown component:
https://github.com/novuhq/novu/blob/8e71c9b09144b5f950c6d8317e2339964051bd93/apps/web/src/design-system/dropdown/Dropdown.tsx#L26

What it might be a bit tricky is in the rendering of the providers is looped so we might need to see how we could identify for the case of Sparkpost that for the Region and type string to render a Dropdown instead an input. We can help you to find the best solution.
Also worth to mention that we are in the process of revamping the integration store so it could happen this change could have conflicts in a near future. But again we can assist with that if needed.

p-fernandez (changes requested):

providers/sparkpost/src/lib/sparkpost.provider.ts
michaldziuba03: @p-fernandez  SparkPost doesn't return timestamp

michaldziuba03:
@p-fernandez I made changes for backend. Now it's usable.

integration-config

Provider works, checks work, next step is making it dropdown for better UX.

If fix will be needed as soon as possible, we can split it to 2 PRs.

providers/sparkpost/src/lib/sparkpost.provider.ts
```diff
@@ -90,4 +92,13 @@ export class SparkPostEmailProvider implements IEmailProvider {
...
     }
   }
+
+  private getEndpoint(region: string) {
+    switch (region) {
+      case 'eu':
+        return 'https://api.eu.sparkpost.com:443';
+      default:
+        return;
+    }
```
p-fernandez, line 102: Good to prepare for more regions if Sparkpost adds them.

libs/shared/src/consts/providers/credentials/provider-credentials.ts
```diff
@@ -229,9 +229,9 @@ export const sparkpostConfig: IConfigCredentials[] = [
   },
   {
     key: CredentialsKeyEnum.Region,
-    displayName: 'EU',
+    displayName: 'Region',
     description: 'Use `eu` if your account is registered to SparkPost EU',
-    type: 'boolean',
+    type: 'string',
```
p-fernandez, line 234: This is the point that I mention it was going to be tricky. But I didn't explain the things properly. I will follow up in the thread of the comment.

p-fernandez (changes requested):

apps/api/src/app/integrations/dtos/credentials.dto.ts
p-fernandez: Now with the fixes I realised I forgot to explain one important detail.
We should use Region as the field and keep it as string but we also need to accommodate the transition as right now our users have `true` and `false` stored as strings and with the changes their integrations will break.
That means that in the Integrations use cases we would need to transform the boolean as strings values to the proper string to show `EU` if a `true` value is provided. At the very same time that transformation will help that if updating the integration we can slowly replace those `true` strings for the proper `eu` value that should be expected.
In the provider we could keep the boolean check as it would work (for default configuration an empty string or undefined would be stored).
```
      /**
       * Both true and `eu` string would set the first option and false and '' or another non defined value would set the other option
       */
      endpoint: config.eu ? 'https://api.eu.sparkpost.com:443' : undefined, 
``
If this explanation is not very clear I can try to help you to implement that when I have a bit of free time.

Comment (michaldziuba03) [reply apps/api/src/app/integrations/dtos/credentials.dto.ts]: @p-fernandez I suggest to just write some method like:
```ts
transformRegion(region: string) {
  // ...add some comment to explain why this method exists
  if (region === 'true' || region === true) return 'eu'

  return region;
}

getEndpont(_region: string) {
  const region = this.transformRegion(_region);
  
  switch(region) {
    case 'eu':
      return '...'
    default:
      return;
  }
}
```

I guess some migration script is needed as well.

apps/web/src/pages/integrations/components/IntegrationInput.tsx
```diff
@@ -67,6 +67,23 @@ export function IntegrationInput({
...
+        description={credential.description}
+        data-test-id={credential.key}
+        error={errors[credential.key]?.message}
+        data={credential.dropdown?.map((item) => ({
+          label: item.name,
+          value: item.value,
+        }))}
+      />
+    );
+  }
```
michaldziuba03, line 85: I made generic way to render dropdown

screenshot

apps/api/src/app/integrations/dtos/credentials.dto.ts
michaldziuba03: > Now with the fixes I realised I forgot to explain one important detail. We should use Region as the field and keep it as string but we also need to accommodate the transition as right now our users have `true` and `false` stored as strings and with the changes their integrations will break. That means that in the Integrations use cases we would need to transform the boolean as strings values to the proper string to show `EU` if a `true` value is provided. At the very same time that transformation will help that if updating the integration we can slowly replace those `true` strings for the proper `eu` value that should be expected. In the provider we could keep the boolean check as it would work (for default configuration an empty string or undefined would be stored).
> 
> ```
>       /**
>        * Both true and `eu` string would set the first option and false and '' or another non defined value would set the other option
>        */
>       endpoint: config.eu ? 'https://api.eu.sparkpost.com:443' : undefined, 
> ```
> 
> If this explanation is not very clear I can try to help you to implement that when I have a bit of free time.


Actually in `sparkpost.handler.ts` `credentials.region` seems to be string and expected value is `"eu"`, so I decided to leave as is.

https://github.com/novuhq/novu/blob/977b6058d866645faf3c25776f3e1d22a623bdab/packages/application-generic/src/factories/mail/handlers/sparkpost.handler.ts#L18C5-L18C5

michaldziuba03:
@p-fernandez @scopsy

libs/shared/src/consts/providers/provider.interface.ts
```diff
@@ -24,6 +24,10 @@ export interface IConfigCredentials {
     text: string;
     when?: boolean;
   };
+  dropdown?: Array<{
+    name: string;
+    value: string | null;
+  }>;
```
michaldziuba03, line 30: new interface element used only for credential of `dropdown` type

scopsy (approved):
👌 Great work, really loved the addition of the select dropdown

scopsy (changes requested):
Following up @p-fernandez comment, the issue is that we already have some users who have the credentials europe stored in the database as boolean.

Introducing this change will create a breaking change for them. In this case we need to still be able to handle boolean value like it was before, vs the new "eu" value in the current dropdown situation. 

I think we can in troduce a check there regarding that. Wdyt?

michaldziuba03:
@scopsy check new changes, this PR is quite urgent

providers/sparkpost/src/lib/sparkpost.provider.ts
```diff
@@ -90,4 +92,21 @@ export class SparkPostEmailProvider implements IEmailProvider {
...
+  private getEndpoint(_region: string) {
+    const region = this.transformLegacyRegion(_region);
+
+    switch (region) {
+      case 'eu':
+        return 'https://api.eu.sparkpost.com:443';
+      default:
+        return;
+    }
+  }
```
michaldziuba03, line 111: legacy region values will be transformed to `eu` string.

scopsy:
@LetItRock could you please help here?

scopsy (approved):
Fantastic work!

PR #3743, opened 2023-07-07
fix(worker): fixed the fcm data message issue with payload messed with additional data

LetItRock:

The FCM data message `data` field should be a type of `Record<string, string>` otherwise the FCM platform will raise an error.

FCM docs:
https://firebase.google.com/docs/cloud-messaging/concept-options

https://firebase.google.com/docs/reference/admin/node/firebase-admin.messaging.basemessage.md#basemessage_interface

Screenshot 2023-07-08 at 00 31 19
Screenshot 2023-07-08 at 00 31 08

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
@@ -24,6 +24,7 @@ import {
   CompileTemplateCommand,
   PushFactory,
 } from '@novu/application-generic';
+import type { IPushOptions } from '@novu/stateless';
 
 import { CreateLog } from '../../../shared/logs';
 import { SendMessageCommand } from './send-message.command';
@@ -59,13 +60,14 @@ export class SendMessagePush extends SendMessageBase {
 
     const pushChannel: NotificationStepEntity = command.step;
 
+    const stepData: IPushOptions['step'] = {
+      digest: !!command.events?.length,
+      events: command.events,
+      total_count: command.events?.length,
+    };
     const data = {
       subscriber: subscriber,
-      step: {
-        digest: !!command.events?.length,
-        events: command.events,
-        total_count: command.events?.length,
-      },
+      step: stepData,
       ...command.payload,
     };
     let content = '';
@@ -147,7 +149,17 @@ export class SendMessagePush extends SendMessageBase {
 
       const overrides = command.overrides[integration.providerId] || {};
 
-      await this.sendMessage(integration, channel.credentials.deviceTokens, title, content, command, data, overrides);
+      await this.sendMessage(
+        subscriber,
+        integration,
+        channel.credentials.deviceTokens,
+        title,
+        content,
+        command,
+        command.payload,
+        overrides,
+        stepData
+      );
     }
   }
 
@@ -165,13 +177,15 @@ export class SendMessagePush extends SendMessageBase {
   }
 
   private async sendMessage(
+    subscriber: IPushOptions['subscriber'],
     integration: IntegrationEntity,
     target: string[],
     title: string,
     content: string,
     command: SendMessageCommand,
     payload: object,
-    overrides: object
+    overrides: object,
+    step: IPushOptions['step']
   ) {
     const message: MessageEntity = await this.messageRepository.create({
       _notificationId: command.notificationId,
@@ -217,6 +231,8 @@ export class SendMessagePush extends SendMessageBase {
         content,
         payload,
         overrides,
+        subscriber,
+        step,
       });
 
       await this.createExecutionDetails.execute(
--- a/packages/stateless/src/lib/provider/provider.interface.ts
+++ b/packages/stateless/src/lib/provider/provider.interface.ts
@@ -68,6 +68,12 @@ export interface IPushOptions {
     };
     fcmOptions?: { analyticsLabel?: string };
   };
+  subscriber: object;
+  step: {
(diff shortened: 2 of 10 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -147,7 +149,17 @@ export class SendMessagePush extends SendMessageBase {
...
 
-      await this.sendMessage(integration, channel.credentials.deviceTokens, title, content, command, data, overrides);
+      await this.sendMessage(
+        subscriber,
+        integration,
+        channel.credentials.deviceTokens,
+        title,
+        content,
+        command,
+        command.payload,
```
LetItRock, line 159: this line basically is the fix... the payload was containing additional data that shouldn't be sent with the FCM data message

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -217,6 +231,8 @@ export class SendMessagePush extends SendMessageBase {
         content,
         payload,
         overrides,
+        subscriber,
+        step,
```
LetItRock, line 235: this data is needed only for the `webhook` provider, so I'm sending them as a separate fields

providers/fcm/src/lib/fcm.provider.spec.ts
```diff
@@ -257,3 +277,49 @@ test('should trigger fcm data for android with priority option', async () => {
     },
   });
 });
+
+test('should clean the payload for the FCM data message', async () => {
```
LetItRock, line 281: test that makes sure that the FCM data message data field is `Record<string, string>`

providers/fcm/src/lib/fcm.provider.ts
```diff
@@ -38,17 +38,18 @@ export class FcmPushProvider implements IPushProvider {
...
   }
 
   async sendMessage(
-    options: IPushOptions
+    options: IPushOptions,
   ): Promise<ISendMessageSuccessResponse> {
     delete (options.overrides as { deviceTokens?: string[] })?.deviceTokens;
 
     const overridesData = options.overrides || ({} as any);
+    const payload = this.cleanPayload(options.payload);
```
LetItRock, line 52: clean the payload, like transform it to `Record<string, string>`

providers/push-webhook/src/lib/push-webhook.provider.ts
```diff
@@ -15,13 +15,20 @@ export class PushWebhookPushProvider implements IPushProvider {
...
+    options: IPushOptions,
   ): Promise<ISendMessageSuccessResponse> {
-    const bodyData = this.createBody(options);
+    const bodyData = this.createBody({
+      ...options,
+      payload: {
+        ...options.payload,
+        subscriber: options.subscriber,
+        step: options.step,
+      },
```
LetItRock, line 30: for the `webhook` provider send `subscriber, step` as the payload, like we did before

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -59,13 +60,14 @@ export class SendMessagePush extends SendMessageBase {
 
     const pushChannel: NotificationStepEntity = command.step;
 
+    const stepData: IPushOptions['step'] = {
```
p-fernandez, line 63: I like this separation. You could also call the object `step` already.

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -68,6 +68,12 @@ export interface IPushOptions {
     };
     fcmOptions?: { analyticsLabel?: string };
   };
+  subscriber: object;
+  step: {
+    digest: boolean;
+    events: object[] | undefined;
+    total_count: number | undefined;
+  };
```
p-fernandez, line 76: Can't here we reuse `IPushOptions['step']`? If it is needed let's move it to `@novu/shared`.

providers/fcm/src/lib/fcm.provider.ts
```diff
@@ -93,7 +94,7 @@ export class FcmPushProvider implements IPushProvider {
       throw new Error(
         `Sending message failed due to "${
           res.responses.find((i) => i.success === false).error.message
-        }"`
+        }"`,
```
p-fernandez, line 97: I think you might have a change of configuration in your prettifying.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
LetItRock: I didn't want to confuse it with a real `step`

packages/stateless/src/lib/provider/provider.interface.ts
LetItRock: this is the `IPushOptions` interface :P

providers/fcm/src/lib/fcm.provider.ts
p-fernandez: We need to disable it. But I have no access to it.

packages/stateless/src/lib/provider/provider.interface.ts
p-fernandez: You got my point 😂

p-fernandez (approved):
🌟

PR #3768, opened 2023-07-11
Remove attachments on last job execution only

scopsy:

Only remove attachments on the last job execution phase, or when a failure happened without retry.

In case of multiple nodes where the email is the last node, it would remove the attachments before the email will be sent.


jainpawan21 (approved):

ainouzgali:
Why @scopsy ? Just wondering

scopsy:
@ainouzgali added in the description phase, that basically what would happen is that the email node will fail because the attachments have been removed from S3 by then. So we should only remove attachments on the last node of execution basically.

PR #3772, opened 2023-07-11
feat(infra): consolidate usage of redis instance by bull-mq

p-fernandez:

Consolidates the usage and instantiation of the Redis instance used by BullMQ for the queues and workers of Novu.

It is needed in preparation of the move in the Cloud offering to support MemoryDB for our solution.

Haven't yet touched:
- `libs/testing/src/queue.service.ts`: as it is used by `@novu/application-generic` and we should think if we want to reuse the Redis instance / MemoryDB instance for the testing Queue. In Test environment we use the same DB index for the testing Queue service and for the apps.
- `apps/ws/src/shared/framework/redis.adapter.ts`: it shares the configuration of the Redis instance used by BullMQ, even using the same DB, so I am tempted (I think we should) to implement it there.

### DIFF
--- a/apps/inbound-mail/src/server/queue-service.ts
+++ b/apps/inbound-mail/src/server/queue-service.ts
@@ -9,14 +9,8 @@ export class QueueService {
   private bullConfig: QueueBaseOptions = {
     connection: {
       db: Number(process.env.REDIS_DB_INDEX) ?? 2,
-      host: process.env.REDIS_HOST,
       port: Number(process.env.REDIS_PORT) ?? 6379,
-      password: process.env.REDIS_PASSWORD,
-      connectTimeout: 50000,
-      keepAlive: 30000,
-      family: 4,
       keyPrefix: process.env.REDIS_PREFIX ?? '',
-      tls: process.env.REDIS_TLS as ConnectionOptions,
     },
   };
   public readonly bullMqService: BullMqService;
--- a/apps/inbound-mail/src/server/queue-service.ts
+++ b/apps/inbound-mail/src/server/queue-service.ts
@@ -1,6 +1,5 @@
 import { BullMqService } from '@novu/application-generic';
-import { Queue, QueueBaseOptions } from 'bullmq';
-import { ConnectionOptions } from 'tls';
+import { QueueBaseOptions } from 'bullmq';
 
 export class QueueService {
   readonly DEFAULT_ATTEMPTS = 5;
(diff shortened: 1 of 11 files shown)


apps/inbound-mail/src/server/queue-service.ts
```diff
@@ -9,14 +9,8 @@ export class QueueService {
   private bullConfig: QueueBaseOptions = {
     connection: {
       db: Number(process.env.REDIS_DB_INDEX) ?? 2,
```
p-fernandez, line 11: I do think we use the same values also for the Inbound Mail Queue Service but at first touch I was conservative and left this same injection and just moved the shared values with the other apps implementations.

packages/application-generic/src/services/bull-mq.service.ts
```diff
@@ -62,7 +81,19 @@ export class BullMqService {
     };
   }
 
-  public createQueue(name: string, config: QueueOptions) {
+  public createQueue(name: string, queueOptions: QueueOptions) {
+    const config = {
```
p-fernandez, line 85: Here and in `createWorker` we will do the discrimination of using either the Redis instance or the MemoryDB (in the next tasks). Yet think how to do it carefully as we will need to have the worker pulling from the old instance and in the new one for our cloud solution.

apps/inbound-mail/src/server/queue-service.ts
```diff
@@ -9,14 +9,8 @@ export class QueueService {
...
-      host: process.env.REDIS_HOST,
       port: Number(process.env.REDIS_PORT) ?? 6379,
-      password: process.env.REDIS_PASSWORD,
-      connectTimeout: 50000,
-      keepAlive: 30000,
-      family: 4,
       keyPrefix: process.env.REDIS_PREFIX ?? '',
-      tls: process.env.REDIS_TLS as ConnectionOptions,
```
LetItRock, line 19: please clean up the unused imports

apps/worker/src/app/workflow/services/workflow-queue.service.spec.ts
```diff
@@ -92,7 +92,7 @@ describe('Workflow Queue service', () => {
 
   it('should be initialised properly', async () => {
     expect(queueService).to.be.ok;
-    expect(queueService).to.have.all.keys('DEFAULT_ATTEMPTS', 'bullConfig', 'bullMqService', 'name');
+    expect(queueService).to.have.all.keys('DEFAULT_ATTEMPTS', 'bullMqService', 'name');
```
LetItRock, line 95: this is my favorite test :D

LetItRock (approved):
you nailed it 💅

PR #3774, opened 2023-07-11
fix(web,shared): added missing novu providers to the providers array

LetItRock:

The integrations endpoint now returns additionally the Novu integrations when the Multi Provider feature flag is on.
This change breaks the Workflow Settings --> Providers tab where these integrations are rendered. 

This PR fixes:
- the icon issue for Novu Email/SMS integration
- the React key warning in the Providers tab
- excludes Novu providers from the "providers" array which is used in the Providers tab, when the ff is off


libs/shared/src/consts/providers/channels/email.ts
```diff
@@ -21,6 +21,14 @@ import { EmailProviderIdEnum } from '../provider.enum';
...
 
 export const emailProviders: IProviderConfig[] = [
+  {
+    id: EmailProviderIdEnum.Novu,
+    displayName: 'Novu Email',
+    channel: ChannelTypeEnum.EMAIL,
+    credentials: [],
+    docReference: 'https://docs.novu.co/channels/email/novu-email',
+    logoFileName: { light: 'novu.png', dark: 'novu.png' },
+  },
```
LetItRock, line 31: add Novu Email provider to the email providers array...
this fixes the issue for icon for Novu Email/SMS integration when multi providers ff is on

libs/shared/src/consts/providers/channels/sms.ts
```diff
@@ -23,6 +23,14 @@ import { SmsProviderIdEnum } from '../provider.enum';
...
 
 export const smsProviders: IProviderConfig[] = [
+  {
+    id: SmsProviderIdEnum.Novu,
+    displayName: 'Novu SMS',
+    channel: ChannelTypeEnum.SMS,
+    credentials: [],
+    docReference: 'https://docs.novu.co/channels/sms/novu-sms',
+    logoFileName: { light: 'novu.png', dark: 'novu.png' },
+  },
```
LetItRock, line 33: add Novu SMS provider to sms providers array

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -65,6 +66,7 @@ export const ListProviders = ({
         .map((provider) => {
           return (
             <UnstyledButton
+              key={provider.identifier ?? provider.providerId}
```
LetItRock, line 69: the fix for the React key warning

apps/web/src/pages/integrations/useProviders.ts
```diff
@@ -23,36 +34,38 @@ function fcmFallback(integration: IntegrationEntity | undefined, clonedCredentia
 }
 
 function initializeProviders(integrations: IntegrationEntity[]): IIntegratedProvider[] {
-  return providers.map((providerItem) => {
-    const integration = integrations.find((integrationItem) => integrationItem.providerId === providerItem.id);
+  return providers
```
ainouzgali, line 37: The filtering of novu providers could also be done here,  but not significant

libs/shared/src/consts/providers/channels/sms.ts
```diff
@@ -23,6 +23,14 @@ import { SmsProviderIdEnum } from '../provider.enum';
 import { ChannelTypeEnum } from '../../../types';
 
 export const smsProviders: IProviderConfig[] = [
+  {
+    id: SmsProviderIdEnum.Novu,
+    displayName: 'Novu SMS',
+    channel: ChannelTypeEnum.SMS,
+    credentials: [],
+    docReference: 'https://docs.novu.co/channels/sms/novu-sms',
```
ainouzgali, line 31: We son't have those docs. It will be done next cycle, right? Maybe for now use a link that we do have, in case its released before that ?

ainouzgali (approved):

apps/web/src/pages/integrations/useProviders.ts
LetItRock: yeah, better to move to here, will do

jacobwgillespie:
Hey @LetItRock, we got an alert at Depot about the build that's running here - the build is indefinitely stuck as it's waiting for the input that `pnpm pnpm-context` is supposed to be sending:

https://github.com/novuhq/novu/actions/runs/5545386319/jobs/10124296523#step:5:346

The container build is waiting for stdin data, but isn't getting any. Most likely, something in the `pnpm-context` script is broken, but it's not showing up in the logs because of the `--silent` flag passed to pnpm.

You will likely want to cancel the current container build and determine why the pnpm script isn't generating input for it.

LetItRock:
hey @jacobwgillespie 👋 thanks for keeping the finger on a pulse :) 
I'll re-run the pipeline again and hope that it's a temporal glitch :)

PR #3778, opened 2023-07-12
Nv 2569 update provider instances display in the provider tab of the

davidsoderberg:

Info about integration provider fro multi provider feature

Screenshot 2023-07-12 at 06 15 43


apps/web/src/pages/templates/components/ListProviders.tsx
davidsoderberg, line 1: In this file the main change is made

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -79,20 +84,60 @@ export const ListProviders = ({
...
+                    style={{
+                      height: '24px',
+                      maxWidth: '140px',
+                      opacity: provider.active ? 1 : colorScheme === 'dark' ? 0.4 : 1,
+                    }}
+                  />
+
+                  <Stack
+                    sx={{
+                      width: isIntegrationsListPageEnabled ? '132px' : undefined,
```
davidsoderberg, line 101: This was the only way I could get the provider name truncated. Not 100% happy with it so if someone else have a better way of solving it, please tell me :)

apps/web/src/pages/integrations/IntegrationsStoreModal.tsx
```diff
@@ -314,22 +312,6 @@ const useDrawerStyles = createStyles((theme: MantineTheme) => {
   };
 });
 
-export interface IIntegratedProvider {
```
davidsoderberg, line 317: I think it was @p-fernandez that noticed this in one of my previous PRs and now I am handling this technical debt.

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -17,6 +21,8 @@ export const ListProviders = ({
   setProvider: (provider: IIntegratedProvider) => void;
 }) => {
   const { colorScheme } = useMantineColorScheme();
+  const isIntegrationsListPageEnabled = !useIsMultiProviderConfigurationEnabled();
```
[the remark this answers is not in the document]
ainouzgali: Is the ! a leftover from testing?

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -79,20 +85,58 @@ export const ListProviders = ({
...
+                    }}
+                  />
+
+                  <Stack
+                    sx={{
+                      width: isIntegrationsListPageEnabled ? '117px' : undefined,
+                    }}
+                    spacing={0}
+                  >
+                    <Tooltip label={provider.displayName} opened={provider.name ? undefined : false}>
```
ainouzgali, line 106:
```suggestion
                    <Tooltip label={provider.displayName} opened={isIntegrationsListPageEnabled}>
```
I think after the migration, all will have a name, also those that don't have the feature. You could also use the hidden property hidden={!isIntegrationsListPageEnabled} instead of open, to avoid the undefined

ainouzgali (commented):
@davidsoderberg sorry if I just missed it, but where are you filtering to show only the current env?

apps/web/src/pages/templates/components/ListProviders.tsx
ainouzgali: Could <Text row={1} ..> work here? Text from our design system

apps/web/src/pages/templates/components/ListProviders.tsx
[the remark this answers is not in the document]
davidsoderberg: yes will remove it and rename the var :)

apps/web/src/pages/templates/components/ListProviders.tsx
davidsoderberg: I tried it and the problem is that it does not create two rows instead it breaks between the text and integration environment pill.

davidsoderberg:
> @davidsoderberg sorry if I just missed it, but where are you filtering to show only the current env?


The filtering is not included in the issue, I guess they did go with adding this info instead.

davidsoderberg:
> Also - maybe we should add a scroll, or set a consistent height, as it shows a bigger page container only for the provider tab.

You mean like we did for the provider list on create integration?

ainouzgali:
> > Also - maybe we should add a scroll, or set a consistent height, as it shows a bigger page container only for the provider tab.
> 
> You mean like we did for the provider list on create integration?

Or maybe simpler. Just that if they have a lot of instances, the page will be very very long

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -17,6 +21,9 @@ export const ListProviders = ({
   setProvider: (provider: IIntegratedProvider) => void;
 }) => {
   const { colorScheme } = useMantineColorScheme();
+  const isMultiProviderConfigurationEnabled = !useIsMultiProviderConfigurationEnabled();
+  const { environments } = useFetchEnvironments();
+  const { environment: currentEnvironment } = useEnvController();
```
ainouzgali, line 26: Can we use only the useEnvController now?

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -79,20 +86,61 @@ export const ListProviders = ({
...
+                        Key: {provider.identifier}
+                      </Text>
+                    </When>
+                  </Stack>
+                </Group>
+                <Group spacing={16} position="apart">
+                  <When truthy={isMultiProviderConfigurationEnabled}>
+                    <IntegrationEnvironmentPill
+                      name={
+                        environments?.find((environment) => environment._id === provider?.environmentId)?.name ||
```
ainouzgali, line 131: we can just take currentEnv name here

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -17,6 +21,9 @@ export const ListProviders = ({
   setProvider: (provider: IIntegratedProvider) => void;
 }) => {
   const { colorScheme } = useMantineColorScheme();
+  const isMultiProviderConfigurationEnabled = !useIsMultiProviderConfigurationEnabled();
```
ainouzgali, line 24: the ! ;)

ainouzgali (approved):

apps/web/src/pages/templates/components/ListProviders.tsx
davidsoderberg: No funny enough useEnvController only return current env

PR #3780, opened 2023-07-12
Content is changed in two sms steps with a delay in between.

djabarovgeorge:

Steps to reproduce:

1. Create a workflow like this:- trigger → sms → delay → sms
2. Click on the first sms, and a content window will open on the right side
3. Add some content
4. Click on the second sms, the first sms content will be there for the second sms as well.
5. Try to change the content in the second sms, it will change for the first sms as well

Now click on either delay or trigger and then repeat step steps 3, 4, and 5. In this case, it will not change for both


ainouzgali (approved):
Whoa. Great job. Lets also add a test for that?❤️

scopsy (approved):
Would be great to add a test case for this aswell for the future 🎉

PR #3781, opened 2023-07-12
fix: node uses non existing endpoint

ainouzgali:
@novu/node markMessageSeen uses endpoint that no longer exists.

```novu.subscribers.markMessageSeen(..)```

Directs to the deprecated endpoint - which has been removed completely last version. 

```  
async markMessageSeen(subscriberId: string, messageId: string) {
    return await this.http.post(
      `/subscribers/${subscriberId}/messages/${messageId}/seen`
    );
  } 

```

Should use the ```'/:subscriberId/messages/markAs'``` endpoint.


LetItRock (approved):
you have the eagle eye 🦅

PR #3782, opened 2023-07-12
fix: webhook enabled if app is not self-hosted

djabarovgeorge:

Will provide a webhook if it is available.

Only the cloud version supports it at the moment.


jainpawan21:
@scopsy Can we please include this in the upcoming patch release?

jainpawan21 (approved):

p-fernandez (approved):

PR #3785, opened 2023-07-12
feat(infra): create feature flag in launch darkly for is_in_memory_cluster_mode_enabled

p-fernandez:

Creates a feature flag in Launch Darkly for IS_IN_MEMORY_CLUSTER_MODE_ENABLED. It also changes the way the InMemoryProviderService is initialized and instantiated.
Also the environment flag ruling this is renamed from IN_MEMORY_CLUSTER_MODE_ENABLED to IS_IN_MEMORY_CLUSTER_MODE_ENABLED to follow current _unofficial_ standard.
It also creates the feature to be able to request global feature flags that can be used by the whole monorepo (anonymous context; that means not need to pass neither a targeted context to get the feature flag value; useful for system based feature flags like this one from the infrastructure).

There are several reasons for this. One is solve the problem that we were causing as side effect to our self hosted users that were enabling not on purpose the cluster mode when deploying Docker instances with NODE_ENV=test. The other one is to prepare the code for a rework needed in InMemoryProviderService to be able to connect optionally the a MemoryDB cluster instance to BullMQ instead of the current Redis instance. For our Cloud service we need to connect to it while leaving the connection to the current Redis instance for the self hosted users. That work will come in future PRs.

Things to do in infra:
* Review that LAUNCH_DARKLY_API_KEY is properly set to Test environment value for API and Worker in GitHub Actions (CI/CD).
* Set IS_IN_MEMORY_CLUSTER_MODE_ENABLED in Launch Darkly only for DEV environment as `true` in AWS for API and Worker.
* Remove existing IN_MEMORY_CLUSTER_MODE_ENABLED as it is deprecated by the new environment variable. This is a potential breaking change but I highly doubt any of our users will be using the cluster mode. Reminding to warn in release notes.


packages/application-generic/src/services/feature-flags.service.ts
```diff
@@ -76,6 +76,25 @@ export class FeatureFlagsService {
     )) satisfies T;
   }
 
+  /**
+   * When we want to retrieve a global feature flag that shouldn't be dependant on any context
+   * we will use this functionality. Helpful for setting feature flags that discriminate
+   * the Novu Cloud service offerings with the self hosted users.
+   */
+  public async getGlobal<T>(
```
p-fernandez, line 84: Functionality to use when the feature flag to apply is general and doesn't need targeting by context (like infra needs).

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -37,13 +38,18 @@ export class InMemoryProviderService {
   public inMemoryProviderClient: InMemoryProviderClient;
   public inMemoryProviderConfig: InMemoryProviderConfig;
 
-  constructor(private enableAutoPipelining?: boolean) {
+  constructor(
+    private getFeatureFlag: GetFeatureFlag,
```
p-fernandez, line 42: This aches my brain for 2 different reason: I am plugging a use case in a service to avoid having to build a middle man service `NovuFeatureFlagsService` and also because the pattern we chose for instantiation forces me to have to inject in every use of `InMemoryProviderService` when this dependency could be self contained and instantiated inside of the constructor saving hassles. 🤷🏻

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -80,12 +86,21 @@ export class InMemoryProviderService {
     return this.getStatus() === CLIENT_READY;
   }
 
-  public isClusterMode(): boolean {
-    return process.env.IN_MEMORY_CLUSTER_MODE_ENABLED === 'true';
+  public async isClusterMode(): Promise<boolean> {
+    const isClusterModeEnabled =
```
p-fernandez, line 90: Now we can rule this by the feature flags service. For our self hosted users that would mean that if they do not plug a feature flags service they will fallback to the environment variable set (in the default .env files provided would be `false`) and if not setting the environment variable files, would fallback to `false` as the hardcoded value.

packages/application-generic/src/services/launch-darkly.service.ts
```diff
@@ -40,6 +40,19 @@ export class LaunchDarklyService implements IFeatureFlagsService {
...
 
+  public async getWithAnonymousContext<T>(
+    key: FeatureFlagsKeysEnum,
+    defaultValue: T
+  ): Promise<T> {
+    const anonymousUserContext = {
+      key,
+      kind: 'user',
+      anonymous: true,
+    };
```
p-fernandez, line 51: Just following Launch Darkly documentation.

packages/application-generic/src/usecases/get-feature-flag/get-feature-flag.use-case.ts
```diff
@@ -24,6 +25,26 @@ export class GetFeatureFlag {
     return await this.featureFlagsService.get(key, defaultValue, context);
   }
 
+  async getGlobal<T>(command: GetGlobalFeatureFlagCommand<T>): Promise<T> {
```
p-fernandez, line 28: In the use case we can differentiate now between functionalities that need context discrimination (getWithContext) and the ones that not through `getGlobal`. Happy to rename to something more suitable. But didn't want to call it `getAnonymous` or similar as that's a naming choice of Launch Darkly and I'd like to leave the FeatureFlagsService agnostic to provider. Also considering that feature flags targeted by environmentId or organizationId fall in the _anonymous_ definition (no userId provided). 🤔

apps/api/src/app/shared/shared.module.ts
```diff
@@ -82,37 +82,74 @@ function getStorageServiceClass() {
...
 
     return new DistributedLockService(factoryInMemoryProviderService);
   },
 };
 
 const PROVIDERS = [
+  launchDarklyService,
+  featureFlagsService,
+  getFeatureFlagUseCase,
+  inMemoryProviderService,
```
p-fernandez, line 150: Reordered in their dependency order.

packages/application-generic/src/usecases/get-feature-flag/get-feature-flag.test.ts
```diff
@@ -85,7 +101,7 @@ describe('Get Feature Flag', () => {
       });
     });
 
-    describe.skip('SDK key environment variable is set', () => {
+    describe('SDK key environment variable is set', () => {
```
p-fernandez, line 104: Now they shouldn't be flakey anymore.

apps/api/src/app/widgets/e2e/initialize-widget-session.e2e.ts
```diff
@@ -64,48 +84,64 @@ describe('Initialize Session - /widgets/session/initialize (POST)', async () =>
...
+    providerId: InAppProviderIdEnum.Novu,
+    channel: ChannelTypeEnum.IN_APP,
+  };
+
+  await invalidateCache.invalidateByKey({
+    key: buildIntegrationKey().cache({
+      _organizationId: session.organization._id,
+      ...command,
+    }),
+  });
```
p-fernandez, line 142: Without this cache wasn't fully erased, returning in SelectIntegration all the time the cached initial integration (the one created by UserSession) and not the updated one with the values changed from line 144.

apps/worker/src/app/shared/shared.module.ts
```diff
@@ -118,29 +118,34 @@ const getFeatureFlagUseCase = {
 
 const inMemoryProviderService = {
   provide: InMemoryProviderService,
-  useFactory: (enableAutoPipelining?: boolean): InMemoryProviderService => {
-    const inMemoryProvider = new InMemoryProviderService(enableAutoPipelining);
-    inMemoryProvider.initialize();
+  useFactory: async (enableAutoPipelining?: boolean): Promise<InMemoryProviderService> => {
+    const getFeatureFlag = await getFeatureFlagUseCase.useFactory();
+    const inMemoryProvider = new InMemoryProviderService(getFeatureFlag, enableAutoPipelining);
+    await inMemoryProvider.initialize();
```
LetItRock, line 124: same comments that I left above

packages/application-generic/src/health/in-memory.health-indicator.ts
```diff
@@ -15,6 +15,10 @@ export class InMemoryProviderServiceHealthIndicator extends HealthIndicator {
     super();
   }
 
+  /**
+   * TODO: Think how to check FeatureFlagsService dependency
+   * without blocking self hosted users
```
LetItRock, line 20: sorry, but I do not fully understand what you mean... like do you mean to check if the feature flag service is up?

packages/application-generic/src/services/cache/cache.service.ts
```diff
@@ -95,28 +103,32 @@ export class CacheService implements ICacheService {
...
+      return await this.capturedExec(
+        pipeline,
+        CacheServiceActionsEnum.SET_QUERY,
+        key
+      );
     }
   }
 
-  public async keys(pattern?: string) {
+  public async keys(pattern?: string): Promise<string[]> {
```
LetItRock, line 114: the types are deferred... same above and below

packages/application-generic/src/services/feature-flags.service.ts
LetItRock: Do you mean kill system switch?

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
LetItRock: First I feel like the `GetFeatureFlag` should be a service, not a use-case, because it exposes many other public functions through its API, and it's against the idea of use-case.
Second, you can also pass the `isClusterMode` flag to the constructor.

.github/workflows/reusable-web-e2e.yml
```diff
@@ -61,13 +62,14 @@ jobs:
         with:
           cypress: true
 
+      - uses: ./.github/actions/setup-redis-cluster
+      - uses: ./.github/actions/start-localstack
```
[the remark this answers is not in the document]
p-fernandez: I do agree. I'd like us to review the CI pipelines and try to reuse as much common things as possible.

apps/api/src/app/shared/shared.module.ts
```diff
@@ -114,29 +114,34 @@ const getFeatureFlagUseCase = {
 
 const inMemoryProviderService = {
   provide: InMemoryProviderService,
-  useFactory: (enableAutoPipelining?: boolean): InMemoryProviderService => {
-    const inMemoryProvider = new InMemoryProviderService(enableAutoPipelining);
-    inMemoryProvider.initialize();
+  useFactory: async (enableAutoPipelining?: boolean): Promise<InMemoryProviderService> => {
+    const getFeatureFlag = await getFeatureFlagUseCase.useFactory();
```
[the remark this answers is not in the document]
p-fernandez: Sorry, I do not understand what do you mean.

apps/api/src/app/widgets/e2e/get-count.e2e.ts
```diff
@@ -21,13 +23,23 @@ describe('Count - GET /widget/notifications/count', function () {
...
+  let inMemoryProviderService: InMemoryProviderService;
+
+  before(async () => {
+    const featureFlagsService = new FeatureFlagsService();
+    const getFeatureFlag = new GetFeatureFlag(featureFlagsService);
+    inMemoryProviderService = new InMemoryProviderService(getFeatureFlag);
+    await inMemoryProviderService.initialize();
+    const cacheService = new CacheService(inMemoryProviderService);
+    await cacheService.initialize();
+    invalidateCache = new InvalidateCacheService(cacheService);
```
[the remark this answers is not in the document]
p-fernandez: Yeah, I was considering to add it to the testing library as initialization of dependencies for E2E tests. Cooldown or tech debt?

apps/api/src/app/widgets/e2e/initialize-widget-session.e2e.ts
p-fernandez: Not sure about that but that won't make any reader to lose knowledge of what's going on in the tests?
CC: @djabarovgeorge how is your vision on this for cache?

packages/application-generic/src/services/cache/cache.service.ts
p-fernandez: I guess you mean inferred.
Personally, I'd rather define the contract for the public API of any class explicit to try to catch the errors as soon as possible. Also it helps me to understand what's going on. Thus I favour verbosity and explicit typing in returns.

packages/application-generic/src/services/feature-flags.service.ts
p-fernandez: Not sure what you mean, but for features that for any usage they just need to be on or off no matter of the user, org and env. Like our current feature flag system works.

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
p-fernandez: The second suggestion would probably disappear that need with the forthcoming refactor InMemoryProviderService needs.

About the first I agree and disagree at the very same time.
I do agree that `getWithContext` and `getGlobal` methods from the use case should live in the FeatureFlagsService, but still don't solve the problem.
The use case is where the business rules should live in the domain, that in this case are the default and fallback values of every feature flag plus the key they are related. If I converted this into a service we would expose that information and that's something I consider we shouldn't do it.
I think the problem that I am in would need another layer that we still haven't defined in our architecture. Also I don't recall us having this situation before of 2 interdependent services.
Should we bring this situation to the tech summit to debate how we want to face it as a standard?

apps/api/src/app/shared/shared.module.ts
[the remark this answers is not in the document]
LetItRock: basically you can inject the feature flag use-case to the `useFactory` function: https://docs.nestjs.com/fundamentals/custom-providers#factory-providers-usefactory

packages/application-generic/src/health/in-memory.health-indicator.ts
LetItRock: maybe better to have a ticket for that, than leaving the todo?

packages/application-generic/src/services/cache/cache.service.ts
LetItRock: The downside IMO is that you have to maintain the types anytime anything inside changes. Matt Pocock recommends doing that only when the function has multiple branches or when the return type is complex:
https://www.youtube.com/watch?v=nwSe95uFN8E

packages/application-generic/src/services/feature-flags.service.ts
LetItRock: maybe kill system switch is not a good example, but I got what you mean

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
LetItRock: `The use case is where the business rules should live in the domain` - the same you can say about the service, but the differentiation is that the use-case is more granular, and it encapsulates "one business rule, one use-case".

`If I converted this into a service we would expose that information and that's something I consider we shouldn't do it.` - what about right now, isn't it already exposed?

```
await this.getFeatureFlag.isMultiProviderConfigurationEnabled(...)
```

I think that we can move the `getWithContext` and `getGlobal` to the `FeatureFlagsService` as you mentioned. The rest functions should be defined as separate use-cases or in a service that will expose them.

p-fernandez: > `If I converted this into a service we would expose that information and that's something I consider we shouldn't do it.` - what about right now, isn't it already exposed?
> 
> ```
> await this.getFeatureFlag.isMultiProviderConfigurationEnabled(...)
> ```

What I meant with expose the information is to have in an infrastructure service like FeatureFlagsService to have the domain information of each feature flag default and fallback value. And take the risk of that in some way to be spread around the apps when implementing their usage (like someone deciding to have to pass a new fallback or default) because that would lead to inconsistencies.

If the main problem is that I called use case to a group of individual feature flags implementations inside one class, I can split them in multiple classes with the command pattern implementation we use. But it disappoints me that this implementation was ok until few days before the release giving me less margin to work in them.

I have other disagreements with some of the architecture related statements of your message but this is not the place to discuss them. Worth for an architecture discussion in a tech summit better. 🙂

.github/workflows/reusable-web-e2e.yml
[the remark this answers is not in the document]
LetItRock: added ticket: https://linear.app/novu/issue/NV-2604/github-actions-move-the-localstack-and-the-redis-cluster-actions-to

LetItRock (approved):
to me it looks much more clear than it was 🥇

PR #3787, opened 2023-07-12
[New Email Provider]: Add Plunk email provider

toyosi12:

This PR adds a new email provider to the existing ones. 

Email Provider: [Plunk](https://docs.useplunk.com)

Plunk is very easy to use and integrate. The more excellent providers, the better.

Screenshot 2023-07-12 at 19 23 14
Screenshot 2023-07-12 at 19 23 28
Screenshot 2023-07-12 at 19 24 08


providers/plunk/src/lib/plunk.provider.ts
```diff
@@ -0,0 +1,68 @@
...
+  ): Promise<ISendMessageSuccessResponse> {
+    const response: IPlunkResponse = await this.plunk.emails.send({
+      from: options.from,
+      to: options.to,
+      subject: options.subject,
+      body: options.html || options.text,
+    });
+
+    return {
+      id: response.contact,
```
michaldziuba03, line 64: I cannot find any reference to response shape in the provider docs.

What is `response.contact`? Is this id returned by provider?

driaug: Founder of Plunk here. First of all really happy to see the open-source power. I went ahead and updated the docs because return shape was indeed missing for this feature. Plunk by default returns an array of contact IDs (https://docs.useplunk.com/api-docs/transactional#returns) because it is possible to send the same email to up to 5 contacts at once (https://docs.useplunk.com/api-docs/transactional#sending-to-multiple-recipients). 

A Plunk contact is an email that is shared over automations, transactional emails and campaigns. We return unique identifiers so that our users can set up triggers, workflows etc in their own applications across different types of emails.

providers/plunk/src/lib/plunk.provider.ts
```diff
@@ -0,0 +1,68 @@
...
+      date: new Date().toISOString(),
```
michaldziuba03, line 65: according to updated provider docs I see there is `response.timestamp`. 
We use `new Date().toISOString()` only as fallback when provider does not return any timestamp.

so I suggest to use  `date: response.timestamp`

toyosi12: Yeah. Got that. I'm making the update already

providers/plunk/src/lib/plunk.provider.ts
toyosi12: @michaldziuba03 The provider has updated their docs. I have also updated my PR to return the ID

providers/plunk/src/lib/plunk.provider.ts
toyosi12: @michaldziuba03 I have updated the response to return the timestamp from the provider

providers/plunk/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+A Plunk email provider library for [@novu/node](https://github.com/novuhq/novu)
+
+## Usage
+
+```javascript
+    import { PlunkEmailProvider } from '@novu/plunk';
+
+    const provider = new PlunkEmailProvider({
+    apiKey: process.env.PLUNK_API_KEY
```
p-fernandez, line 11:
```suggestion
      apiKey: process.env.PLUNK_API_KEY
```
Just indenting.

providers/plunk/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@novu/plunk",
+  "version": "0.16.0",
```
p-fernandez, line 3: We are in version 0.16.3 right now: https://github.com/novuhq/novu/blob/next/apps/api/package.json
```suggestion
  "version": "0.16.3",
```
Not your fault as the script to generate the skeleton of the provider was set to 0.16.0.

providers/plunk/package.json
```diff
@@ -0,0 +1,83 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "engines": {
+    "node": ">=13.0.0 <17.0.0",
+    "pnpm": "^7.26.0"
+  },
+  "dependencies": {
+    "@novu/stateless": "0.16.0",
```
p-fernandez, line 36:
```suggestion
    "@novu/stateless": "0.16.3",
```

providers/plunk/src/lib/plunk.provider.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  constructor(
+    private config: {
+      apiKey: string;
+    }
+  ) {
+    this.plunk = new Plunk(this.config.apiKey);
+  }
+  getMessageId?: (body: any) => string[];
+  parseEventBody?: (body: any, identifier: string) => IEmailEventBody;
```
p-fernandez, line 28: Seem unnecessary. 🤔

packages/application-generic/src/factories/mail/handlers/plunk.handler.ts
```diff
@@ -0,0 +1,18 @@
...
+    super('plunk', ChannelTypeEnum.EMAIL);
+  }
+
+  buildProvider(credentials: ICredentials, from?: string) {
+    const config: { apiKey: string; from: string } = {
+      apiKey: credentials.apiKey,
+      from,
+    };
+
+    this.provider = new PlunkEmailProvider(config);
```
p-fernandez, line 16: `from` is not being used later in the provider. It is being taken from `options`. Does Plunk do not need to set a `from` email?
Also sender name that is mandatory for the email config base is not being passed in any way.

p-fernandez (changes requested):

packages/application-generic/src/factories/mail/handlers/plunk.handler.ts
driaug: The `from` and `name` parameter are optional with Plunk. Our users verify their domain in the dashboard and select a default from address. The `from` parameter in our API can be used to overwrite the default value.

providers/plunk/README.md
toyosi12: Done. Thanks

providers/plunk/package.json
toyosi12: That's okay. I've updated it

providers/plunk/package.json
toyosi12: Updated

providers/plunk/src/lib/plunk.provider.ts
toyosi12: I think this was generated in the boilerplate. I've updated it any ways.

toyosi12:
@p-fernandez Kindly check the changes

jainpawan21:
@toyosi12 
Thanks for your contributions
Please add documentation for this provider 
You can take this as an example:- https://docs.novu.co/channels/email/sendgrid

toyosi12:
> @toyosi12 Thanks for your contributions Please add documentation for this provider You can take this as an example:- https://docs.novu.co/channels/email/sendgrid

Thanks for the feedback @jainpawan21. I have added the documentation

packages/application-generic/src/factories/mail/handlers/plunk.handler.ts
p-fernandez: > The `from` and `name` parameter are optional with Plunk. Our users verify their domain in the dashboard and select a default from address. The `from` parameter in our API can be used to overwrite the default value.

Considering in our system `from` and `senderName` are mandatory parameters to add in our integration our users are going to expect that those values are going to be the used ones if no overrides are placed. So from a point of view of having less cognitive burden I'd pass them in the constructor of the provider and do something like we do in the Sendgrid provider and specify the fallbacks in the usage:
https://github.com/novuhq/novu/blob/a4a0b9681ff89b7d1e16d51b7c9cb773c65d573e/providers/sendgrid/src/lib/sendgrid.provider.ts#L68
But after reviewing our SendMessageEmail use case I reckon we sort that out already so we could leave the provider like this. We can fix it up if any weird behaviour happens in the future. 👍🏻

p-fernandez (approved):
Thanks for your contribution. 🌟

sumitsaurabh927 (approved):

PR #3789, opened 2023-07-13
feat: override integration instance on trigger

ainouzgali:

Allow the API user to override the integration config for a channel using the trigger endpoint and overrides property. 
For email and sms. 

```
await novu.trigger('<TRIGGER_NAME>', {
  to: [
    {
      ...subscriber
    },
  ],
  payload: {
    ...data
  },
  overrides: {
     email: {
       integrationIdentifier: 'instance-identifier'
     },
    sms: {
       integrationIdentifier: 'instance-identifier'
     }
  }
});
```


apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -94,6 +96,23 @@ export class SendMessageEmail extends SendMessageBase {
...
+            isTest: false,
+            isRetry: false,
+            raw: JSON.stringify({
+              integrationIdentifier: overrideSelectedIntegration,
+            }),
+          })
+        );
+
+        return;
+      }
```
ainouzgali, line 115: Show on activity feed that failed, because of a wrong integrationIdentifier

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -66,6 +66,14 @@ export class SelectIntegration {
...
     };
 
+    if (command.identifier) {
+      query = {
+        _organizationId: command.organizationId,
+        channel: command.channelType,
+        identifier: command.identifier,
+      };
+    }
+
```
ainouzgali, line 76: For now, allow also disabled integrations. Until we allow more than one active instance per environment.

ainouzgali: I see now that we do allow 2 active for same env. I will check again

ainouzgali: Changed it to allow to override only with active instances

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -1553,41 +1357,313 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
...
+      const newSubscriberId = SubscriberRepository.createObjectId();
+      const channelType = ChannelTypeEnum.EMAIL;
+
+      template = await createTemplate(session, channelType);
+
+      await sendTrigger(session, template, newSubscriberId);
+
+      await session.awaitRunningJobs(template._id);
+
+      const createdSubscriber = await subscriberRepository.findBySubscriberId(session.environment._id, newSubscriberId);
```
ainouzgali, line 1620: @LetItRock There is the test, sorry, I should have highlighted it with a comment.

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -66,6 +66,15 @@ export class SelectIntegration {
...
     };
 
+    if (command.identifier) {
+      query = {
+        _organizationId: command.organizationId,
+        channel: command.channelType,
+        identifier: command.identifier,
+        active: true,
+      };
+    }
```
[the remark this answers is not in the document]
ainouzgali: Sorry, there is a test, should have wrote a comment on it 🤦🏻‍♀️ Tagged you there.

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
ainouzgali: I see your point. But I'm not sure about it being understandable enough if he just made a mistake with the identifier. Also, in general, I feel  SUBSCRIBER_NO_ACTIVE_INTEGRATION doesn't really fit the reason we give the error here. 
But for now, you are right it might be better. I will change it.

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -60,12 +60,14 @@ export class SendMessageEmail extends SendMessageBase {
...
     let integration: IntegrationEntity | undefined = undefined;
 
+    const overrideSelectedIntegration = command.overrides?.email?.integrationIdentifier;
     try {
       integration = await this.getIntegration({
         organizationId: command.organizationId,
         environmentId: command.environmentId,
         channelType: ChannelTypeEnum.EMAIL,
         userId: command.userId,
+        identifier: overrideSelectedIntegration as string,
```
LetItRock, line 70: as string is not needed

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
```diff
@@ -48,11 +48,14 @@ export class SendMessageSms extends SendMessageBase {
...
     if (!subscriber) throw new PlatformException('Subscriber not found');
 
+    const overrideSelectedIntegration = command.overrides?.sms?.integrationIdentifier;
+
     const integration = await this.getIntegration({
       organizationId: command.organizationId,
       environmentId: command.environmentId,
       channelType: ChannelTypeEnum.SMS,
       userId: command.userId,
+      identifier: overrideSelectedIntegration as string,
```
LetItRock, line 58: same

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -66,6 +66,15 @@ export class SelectIntegration {
       active: true,
     };
 
+    if (command.identifier) {
+      query = {
+        _organizationId: command.organizationId,
+        channel: command.channelType,
+        identifier: command.identifier,
+        active: true,
+      };
```
LetItRock, line 75: Let's specify the `_environment` here too. 

Also, I feel that it would be easier if you do, than having the separate if, otherwise if there are any changes to the above we would have to remember to update also what we have in the if:
```
const query: Partial<IntegrationEntity> & { _organizationId: string } = {
      ...(command.id ? { id: command.id } : {}),
      channel: command.channelType,
      ...(command.providerId ? { providerId: command.providerId } : {}),
      active: true,
      ...(command.identifier ? { identifier: command.identifier } : {}),
    };
```

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
ainouzgali: Because it is coming from the overrides, its type is unknown. If you have a trick for me to use, let me know🤩

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
LetItRock: ok, that makes sense :) we should state that in the docs in the future :) 
what about active then, we can also allow inactive ones then?

ainouzgali: I discussed with Dima, and decided to only allow active ones

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
LetItRock: ohh, forgot that we have it not typed at all, I think that we have to improve it :) like it can be optional string there

LetItRock (approved):

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
ainouzgali: it is defined as an optional string in the command, but it is 'unknown' from the overrides

PR #3793, opened 2023-07-13
fix: export missing enums in node sdk for create template

jainpawan21:
1. export missing types
2. fix getAll params

### DIFF
--- a/libs/shared/src/entities/notification-template/notification-template.interface.ts
+++ b/libs/shared/src/entities/notification-template/notification-template.interface.ts
@@ -63,7 +63,7 @@ export interface INotificationTemplateStep {
 
 export interface IMessageFilter {
   isNegated?: boolean;
-  type: BuilderFieldType;
+  type?: BuilderFieldType;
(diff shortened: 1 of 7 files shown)


libs/shared/src/entities/notification-template/notification-template.interface.ts
```diff
@@ -63,7 +63,7 @@ export interface INotificationTemplateStep {
 
 export interface IMessageFilter {
   isNegated?: boolean;
-  type: BuilderFieldType;
+  type?: BuilderFieldType;
```
p-fernandez, line 66: What's the reason behind making this field optional?

p-fernandez (approved):
Just a small question. The rest looks good!
🌟

libs/shared/src/entities/notification-template/notification-template.interface.ts
jainpawan21: When we fetch a template and create a template
this field is not needed to apply filter.
type-steps

PR #3811, opened 2023-07-17
feat(app-generic): add cluster mode e2e tests for cache service

p-fernandez:

Adds E2E tests for the Cache service in cluster and non cluster mode. Also fixes instantiation of FeatureFlagsService.

Wanted to split part of the work done in this PR: https://github.com/novuhq/novu/pull/3785.
Should facilitate reviews.
Also to validate the right behaviour of the CacheService in both modes.


packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -37,7 +37,9 @@ export class InMemoryProviderService {
   public inMemoryProviderClient: InMemoryProviderClient;
   public inMemoryProviderConfig: InMemoryProviderConfig;
 
-  constructor(private enableAutoPipelining?: boolean) {
+  constructor(private enableAutoPipelining?: boolean) {}
+
+  public initialize(): void {
```
p-fernandez, line 42: Preparation to make this async when we need to plug in the GetFeatureFlag use case to retrieve the global feature flag to check if the cluster mode is enabled.

packages/application-generic/src/services/launch-darkly.service.ts
```diff
@@ -20,18 +20,35 @@ export class LaunchDarklyService implements IFeatureFlagsService {
   private client: LDClient;
   public isEnabled: boolean;
 
-  constructor() {
-    Logger.verbose('Launch Darkly service initialized', LOG_CONTEXT);
+  public async initialize(): Promise<void> {
```
p-fernandez, line 23: We move to async initialization to avoid race conditions in tests so Launch Darkly is plugged before executing the tests. Also for deployment of the apps that rely on feature flags through the SDK.

ainouzgali (approved):

PR #3814, opened 2023-07-18
fix: getContextPath did not work correctly when called in a browser

DaveAtKeelvar:
The function `getContextPath` does not work correctly when called from inside a web browser. `process.env` is only available when the function is called within node.js. When called inside a web browser, we need to instead use `window._env_` to get access to the different context path variables.

To support custom context paths for Novu web


DaveAtKeelvar:
Feel free to tidy up the TypeScript code, I haven't really used TypeScript before, and I'm sure it could be neater.

vfxGer:
Good catch

libs/shared/src/config/contextPath.ts
```diff
@@ -8,29 +8,41 @@ export enum NovuComponentEnum {
 export function getContextPath(component: NovuComponentEnum) {
   let contextPath = '';
 
-  if (process.env.GLOBAL_CONTEXT_PATH) {
-    contextPath += process.env.GLOBAL_CONTEXT_PATH + '/';
+  // Determine if we are running in the browser or in node.js. If we are
+  // running in node.js, we will have access to the process.env object,
+  // otherwise we will have access to the window._env_ object to get the
+  // environment variables.
```
scopsy, line 14: Seems like there is some eslint error here, i believe you would need to convert this to a multiline comment `/**`

DaveAtKeelvar: Thanks. I updated the code.

DaveAtKeelvar:
@scopsy Is there any hope in moving this forward?

scopsy:
@DaveAtKeelvar yes! Sorry about that, everytime I approve the  test runs and then while waiting forget to come backto it to merge. Added an alarm reminder 😅

PR #3822, opened 2023-07-19
fix: variable manager throwing error for array vars

ainouzgali:

Variable manager was throwing error for using array vars for indexes if not all preceding indexes were mentioned as well. 
For example: array.org.[2] will throw error if  array.org.[0] and array.org.[1] are not included explicitly in content. 

https://github.com/novuhq/novu/assets/7778680/801095cc-614a-4828-8116-e8dc22d5b7ed


LetItRock (approved):
🥇

Issue #3460, opened 2023-05-19
[NV-2582] TypeError: oldMessages.map is not a function


yngfoxx:
## 📜 Description
I get this error when I mark a notification as seen with the headless service library
https://github.com/novuhq/novu/blame/e726694c4a8d2a9af17f7a24bd6f8bf7a06a2223/packages/headless/src/lib/headless.service.ts#L581
![image](https://github.com/novuhq/novu/assets/47150226/b3517bb7-16ce-4817-b04a-13da96ef479e)

## Reproduction
I use Novu's cloud service with HMAC enabled
I am trying to create a simple custom notification centre for React Native + Expo (TypeScript)
```
headlessService.markNotificationsAsSeen({
    messageId: ['message-id-0', 'message-id-1'],
    listener: ({status, isLoading}) => {
        if (isLoading) return;
        if (!isLoading && status !== 'success') return Prompt({
            status: 'error',
            content: 'An error occured while trying to mark a notification as seen.',
            closeText: ''
        });
    },
    onError: console.error
})
```
Note: I used Brave browser 

<sub>[NV-2582](https://linear.app/novu/issue/NV-2582/typeerror-oldmessagesmap-is-not-a-function)</sub>

akashvivek:
I am interested in solving this issue.

jainpawan21:
@BiswaViraj

PR #3824, opened 2023-07-19
fix: updated the types and onsuccess method

BiswaViraj:
fix: #3460 

- The data stored in the `NOTIFICATIONS_QUERY_KEY` was not `IMessage[]`, it was supposed to be `IPaginatedResponse<IMessage>>`, So I updated the onSuccess method to use the correct types.


packages/headless/src/lib/headless.service.ts
```diff
@@ -690,16 +698,20 @@ export class HeadlessService {
...
+              ...(oldMessages?.data && {
+                data: oldMessages?.data?.map((message) => {
+                  if (message._id === data._id) {
+                    return data;
+                  }
+
+                  return message;
+                }),
+              }),
+            })
```
p-fernandez, line 714: Great find, but could you please extract the content of `onSucess` in a private function in the class so it is reused in the 3 places we are using it? 🙏🏻

BiswaViraj: Good idea, will do that

p-fernandez (approved):
🌟

PR #3827, opened 2023-07-20
refactor: bootstrap worker as part of e2e setup

scopsy:

Instead of running the worker in a separate process, we can reuse the same running process of the tests for both service. 

When modifying the feature flags in the API tests, we are not able to modify the feature flags on the worker service.

### DIFF
--- a/apps/api/e2e/setup.ts
+++ b/apps/api/e2e/setup.ts
@@ -3,10 +3,15 @@ import { testServer } from '@novu/testing';
 import * as sinon from 'sinon';
 
 import { bootstrap } from '../src/bootstrap';
+import { bootstrap as bootstrapWorker } from '../../worker/src/bootstrap';
 
 const dalService = new DalService();
 
 before(async () => {
+  process.env.PORT = '1342';
+  await bootstrapWorker();
+
+  process.env.PORT = '1337';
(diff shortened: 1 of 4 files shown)


apps/api/e2e/setup.ts
```diff
@@ -3,10 +3,15 @@ import { testServer } from '@novu/testing';
 import * as sinon from 'sinon';
 
 import { bootstrap } from '../src/bootstrap';
+import { bootstrap as bootstrapWorker } from '../../worker/src/bootstrap';
 
 const dalService = new DalService();
 
 before(async () => {
+  process.env.PORT = '1342';
```
scopsy, line 11: This happens because the API overrides the port for the worker (.env behaviour). I don't really like this approach, we can maybe get value from the .env.test respectively.

Or find a way to run then in isolation, but still have a way to modify the env variables from a single run time for feature flags.

p-fernandez: For me this would be the way to go. In testing library to set up a Worker test bootstrap so we have a similar structure and we can leave the test suites as it is now, so the FF can be applied once and affect the relevant apps. I think that is also crucial for Cypress web tests.

PR #3836, opened 2023-07-21
chore(web): fixing eslint warnings for the react hooks rules - part 2

LetItRock:

This is part 2 of fixing ESLint warnings for the react hook rules. Around ~51 left.
I've verified all the places where I touched the code, so it should be fine 😄 

https://github.com/novuhq/novu/assets/2607232/43d471f8-1595-42d1-92a7-15e459dd9f86

https://github.com/novuhq/novu/assets/2607232/ce56423f-07f8-4179-880f-2e50513130b3

https://github.com/novuhq/novu/assets/2607232/2bac8bdf-7383-4512-b463-8e0c16e0a6c8

https://github.com/novuhq/novu/assets/2607232/ffc10388-dc03-4913-849a-39c40eba1e3d

https://github.com/novuhq/novu/assets/2607232/5e31ae78-db02-471f-b936-371e70fedcef

https://github.com/novuhq/novu/assets/2607232/d9f6e801-fe5e-494a-a956-dc96f732fa20


apps/web/src/components/layout/components/SideNav.tsx
```diff
@@ -70,7 +69,7 @@ export function SideNav({}: Props) {
         },
       },
     ]);
-  }, [environment]);
+  }, [environment, addItem, removeItems, setEnvironment]);
```
LetItRock, line 72: the spotlight item to toggle the environment

apps/web/src/components/layout/components/SideNav.tsx
```diff
@@ -15,11 +15,9 @@ import { colors, NavMenu, SegmentedControl, shadows } from '../../../design-syst
 import { Activity, Bolt, Box, Settings, Team, Repeat, CheckCircleOutlined, Brand } from '../../../design-system/icons';
 import { ChangesCountBadge } from './ChangesCountBadge';
 import { useEnvController } from '../../../hooks';
-import { useAuthContext } from '../../providers/AuthProvider';
 import OrganizationSelect from './OrganizationSelect';
 import { useSpotlightContext } from '../../providers/SpotlightProvider';
 import { HEADER_HEIGHT } from '../constants';
-import { LimitBar } from '../../../pages/integrations/components/LimitBar';
```
LetItRock, line 22: through some files, I've removed the unused imports

apps/web/src/components/quick-start/digest-demo-flow/DigestDemoFlow.tsx
```diff
@@ -35,7 +32,7 @@ export function DigestDemoFlow({
 
   useEffect(() => {
     reactFlowInstance.fitView({ minZoom: 1, maxZoom: 1 });
-  }, [rect.width, rect.height]);
+  }, [reactFlowInstance, rect.width, rect.height]);
```
LetItRock, line 35: reactFlowInstance is not changing between the re-renders

apps/web/src/components/quick-start/digest-demo-flow/DigestNode.tsx
```diff
@@ -51,17 +52,19 @@ export function DigestNode({ data, id }: { data: any; id: string }) {
...
+      intervalObject.start();
     } else {
-      interval.stop();
+      intervalObject.stop();
     }
 
-    return interval.stop;
-  }, [isRunningDigest]);
+    return intervalObject.stop;
+  }, [isRunningDigest, intervalRef]);
```
LetItRock, line 67: to make this change safe I do use ref to hold the `interval, digestInterval` values

apps/web/src/components/quick-start/digest-demo-flow/NodeStepWithPopover.tsx
```diff
@@ -70,7 +70,7 @@ export function NodeStepWithPopover({
 
   useEffect(() => {
     setSequence(popoverData.sequence[counter.toString()] as IBeat);
-  }, [counter]);
+  }, [counter, popoverData]);
```
LetItRock, line 73: popover data is const object

apps/web/src/components/quick-start/in-app-onboarding/TriggerNode.tsx
```diff
@@ -76,39 +76,35 @@ function TriggerButton({ setOpened }: { setOpened: (value: boolean) => void }) {
...
 
-    if (!templatesLoading && !notificationGroupLoading && !createTemplateLoading && !onboardingNotificationTemplate) {
-      createOnBoardingTemplate();
-    }
-  }, [templates, onboardingNotificationTemplate]);
+    await createNotificationTemplate({
+      template: payloadToCreate as unknown as ICreateNotificationTemplateDto,
+      params: { __source: TemplateCreationSourceEnum.ONBOARDING_IN_APP },
+    });
+  }, hasToCreateOnboardingTemplate);
```
LetItRock, line 107: this was the `useEffect` before, but the intention was to run the code only once when the condition is true, that's why I've changed it to `useEffectOnce`

apps/web/src/design-system/table/Table.tsx
```diff
@@ -91,10 +93,8 @@ export function Table<T extends object>({
   ) as unknown as UseTableProps<T>;
 
   useEffect(() => {
-    if (onPageChange) {
-      onPageChange(pageIndex);
-    }
-  }, [pageIndex]);
+    onPageChangeRef.current?.(pageIndex);
```
LetItRock, line 96: the `onPageChange` property changes on every render, so to not break this `useEffect` I do pass it through the ref

apps/web/src/hooks/useBlueprint.ts
```diff
@@ -1,33 +1,33 @@
...
+import { useNavigate, useLocation, useSearchParams } from 'react-router-dom';
 import { useEffect } from 'react';
+
 import { getToken } from './useAuthController';
 import { useSegment } from '../components/providers/SegmentProvider';
 import { ROUTES } from '../constants/routes.enum';
 
 export const useBlueprint = () => {
-  const searchParams = useSearchParams();
+  const [params] = useSearchParams();
```
LetItRock, line 9: use `useSearchParams` from the `react-router-dom` instead of our own

apps/web/src/hooks/useBlueprint.ts
```diff
@@ -1,33 +1,33 @@
...
+  const blueprintId = params.get('blueprintId');
   const navigate = useNavigate();
   const { pathname } = useLocation();
   const segment = useSegment();
+  const id = localStorage.getItem('blueprintId');
+  const token = getToken();
```
LetItRock, line 15: moved from the deps array of the `useEffect`

apps/web/src/hooks/useDebounce.ts
```diff
@@ -1,13 +1,15 @@
...
+import { useDataRef } from './useDataRef';
+
 export const useDebounce = <Arguments = unknown | unknown[]>(callback: (args: Arguments) => void, ms = 0) => {
-  const callbackRef = useRef(callback);
-  callbackRef.current = callback;
+  const callbackRef = useDataRef(callback);
 
-  const debouncedCallback = useCallback(debounce(callbackRef.current, ms), [ms]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const debouncedCallback = useCallback(debounce(callbackRef.current, ms), [callbackRef, ms]);
```
LetItRock, line 10: unfortunately, I had to use this here, because we don't have the types for the `debounce` and it's not a simple thing, it's basically the function that could be called but it has also a cancel method, so it's not also easy to type is by hands...

apps/web/src/hooks/useEnvController.ts
```diff
@@ -40,33 +41,36 @@ export const useEnvController = (): EnvironmentContext => {
     setReadonly(environment?._parentId !== undefined);
   }, [environment]);
 
-  async function setEnvironment(environmentName: string) {
-    if (isLoading || isLoadingMyEnvironments || isLoadingCurrentEnvironment) {
-      return;
-    }
+  const setEnvironmentCallback = useCallback(
```
LetItRock, line 44: wrapped in a callback

apps/web/src/components/layout/components/SideNav.tsx
```diff
@@ -45,22 +43,23 @@ type Props = {};
...
+  const { addItem, removeItems } = useSpotlightContext();
   const { classes } = usePopoverStyles();
 
   useEffect(() => {
     setOpened(readonly);
     if (readonly && location.pathname === ROUTES.CHANGES) {
       navigate(ROUTES.HOME);
     }
-  }, [readonly]);
+  }, [readonly, navigate, location.pathname]);
```
djabarovgeorge, line 58: I think we could remove this effect altogether, IMO it is duplicated because we already have navigated to HOME in useEnvController().setEnvironment, which should be triggered every time the env changes similar to `readonly` change

apps/web/src/hooks/useEnvController.ts
```diff
@@ -40,33 +41,36 @@ export const useEnvController = (): EnvironmentContext => {
...
+      setIsLoading(false);
+      if (!tokenResponse.token) {
+        return;
+      }
+      setToken(tokenResponse.token);
 
-    await navigate('/');
-    await queryClient.invalidateQueries();
-  }
+      await navigate('/');
```
djabarovgeorge, line 63:
```suggestion
      await navigate(ROUTES.HOME);
```

apps/web/src/components/layout/components/SideNav.tsx
djabarovgeorge: 🧹

apps/web/src/components/quick-start/in-app-onboarding/TriggerNode.tsx
djabarovgeorge: Love it :)
small note, i know the code was like this before but maybe we could extract the payloadToCreate outside so it will be easier to read, and I think we could remove the `await` for the creation because useEffectOnce is not async by design.

apps/web/src/design-system/table/Table.tsx
```diff
@@ -61,9 +62,10 @@ export function Table<T extends object>({
   ...props
 }: ITableProps<T>) {
   const { pageSize, total, onPageChange, current } = pagination;
-  const columns = React.useMemo(() => userColumns?.map((col) => ({ ...col })), [userColumns]);
-  const data = React.useMemo(() => (userData || [])?.map((row) => ({ ...row })), [userData]);
-  const fakeData = React.useMemo(() => Array.from({ length: loadingItems }).map((_, index) => ({ index })), []);
+  const columns = useMemo(() => userColumns?.map((col) => ({ ...col })), [userColumns]);
+  const data = useMemo(() => (userData || [])?.map((row) => ({ ...row })), [userData]);
+  const fakeData = useMemo(() => Array.from({ length: loadingItems }).map((_, index) => ({ index })), [loadingItems]);
```
djabarovgeorge, line 67: 🧹👌

apps/web/src/design-system/table/Table.tsx
djabarovgeorge: Would not it be easier to use useCallback?

djabarovgeorge (approved):
Looks great to me left a couple of comments on places i was not sure about 🚀

apps/web/src/components/layout/components/SideNav.tsx
LetItRock: You've probably missed that the navigation happens only when the route is `ROUTES.CHANGES`, so like this `useEffect` is guarding that on the prod env no one can navigate to the Changes page. 
But I think that this logic should be in the `PromoteChangesPage` component, so I'll move it there.

apps/web/src/components/quick-start/in-app-onboarding/TriggerNode.tsx
LetItRock: There is no sense to extract the payload, otherwise, it will be created on every render call. But agree we don't need async/await here.

apps/web/src/design-system/table/Table.tsx
LetItRock: no, because it will be re-created on every render, because `onPageChange` prop changes every render... but I'll think about this as I think we should be able to remove `useEffect` here at all

PR #3838, opened 2023-07-23
refactor: remove duplicate code from renaming to workflow

ainouzgali:

As part of the renaming of notification template to workflow(in API) - reuse the current usecases for notification template Instead of duplicate all. 

What was done in this pr:  
- deleted the duplicated workflows folder 
- renamed current notification-templates folder to workflows
- renamed the templates dtos to workflows 
- use 1 module - workflows - with both controllers 
- refactor all existing api tests to use the workflows path


apps/api/src/bootstrap.ts
```diff
@@ -103,6 +103,7 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
     .addTag('Integrations')
     .addTag('Layouts')
     .addTag('Workflows')
+    .addTag('Notification Templates')
```
ainouzgali, line 106: We still want to show the notification templates in our api reference, but all endpoints will have the deprecated tag.

apps/api/src/bootstrap.ts
```diff
@@ -21,7 +21,7 @@ import { validateEnv } from './config/env-validator';
 
 import * as packageJson from '../package.json';
 
-const extendedBodySizeRoutes = ['/v1/events', '/v1/notification-templates', '/v1/layouts'];
+const extendedBodySizeRoutes = ['/v1/events', '/v1/notification-templates', '/v1/workflows', '/v1/layouts'];
```
ainouzgali, line 24: Added the workflows to the exceptions of a large payload

apps/api/src/app/workflows/dto/update-workflow-request.dto.ts
```diff
@@ -37,9 +37,6 @@ export class UpdateWorkflowRequestDto implements ICreateNotificationTemplateDto
   @IsMongoId()
   notificationGroupId: string;
 
-  @ApiPropertyOptional()
-  active?: boolean;
-
```
ainouzgali, line 42: This is an example of why this refactor was needed- a change/fix for the notification templates needed to be duplicated in the workflows folder.

apps/api/src/app.module.ts
```diff
@@ -42,7 +41,7 @@ const modules: Array<Type | DynamicModule | Promise<DynamicModule> | ForwardRefe
   HealthModule,
   EnvironmentsModule,
   ExecutionDetailsModule,
-  NotificationTemplateModule,
+  WorkflowModule,
```
LetItRock, line 44: we should probably use the plural names when calling all these modules... wdyt?

PR #3839, opened 2023-07-23
fix: error handling in nodejs sdk

jainpawan21:
1. Add Missing methods
2. fix params where needed
3. Error handling


packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -110,27 +129,53 @@ export class Subscribers extends WithHttp implements ISubscribers {
...
+   */
   async markMessageRead(subscriberId: string, messageId: string) {
-    return await this.http.post(
+    return await this.novu.post(
       `/subscribers/${subscriberId}/messages/markAs`,
       { messageId, mark: { read: true } }
     );
   }
 
+  async markInAppMessageAs(
```
ainouzgali, line 152: I feel markMessageAs would fit better with our overall use. What do you think? Is there a specific reason you want to call it like that?

packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -110,27 +129,53 @@ export class Subscribers extends WithHttp implements ISubscribers {
...
+    subscriberId: string,
+    messageId: string,
+    mark: IMarkFields
+  ) {
+    if (!mark.read && !mark.seen) {
+      throw new Error('Either seen or read must be set');
+    }
```
ainouzgali, line 159: If we are going to add the mark as unread / unseen soon, this will not be a good enough check. Maybe you could check if mark is empty for now. Until we have logic for what makes sense for them to put in the object (like cases where they set both read and seen that conflicts with each other)

apps/api/src/app/feeds/feeds.controller.ts
```diff
@@ -68,6 +69,7 @@ export class FeedsController {
     );
   }
 
+  // TODO: change feedId type to mongodbId
```
ainouzgali, line 72: I think that instead of adding all those TODOs, we should add tickets for those tasks.

packages/node/src/lib/novu.ts
```diff
@@ -39,37 +58,139 @@ export class Novu extends EventEmitter {
...
-    this.broadcast = this.events.broadcast;
-    this.changes = new Changes(this.http);
-    this.layouts = new Layouts(this.http);
-    this.notificationGroups = new NotificationGroups(this.http);
-    this.notificationTemplates = new NotificationTemplates(this.http);
-    this.feeds = new Feeds(this.http);
-    this.topics = new Topics(this.http);
-    this.integrations = new Integrations(this.http);
-    this.messages = new Messages(this.http);
+    this.subscribers = new Subscribers(this);
```
djabarovgeorge, line 61: i am not sure i am a bit fan of passing 'this' to all of the properties only to use the http methods. this way we will pass this ref to all of the `Novu` properties with no actual reason unless i am missing something. 

potentially we could add the http methods and handleError to the WithHttp class.

anyways would love to hear other's opinions on the matter.

apps/api/src/app/change/dtos/change-request.dto.ts
```diff
@@ -2,6 +2,7 @@ import { ApiProperty } from '@nestjs/swagger';
 import { IsDefined, IsString } from 'class-validator';
 import { PaginationRequestDto } from '../../shared/dtos/pagination-request';
 
+// TODO: change promoted type to boolean
```
djabarovgeorge, line 5: I think we could either implement the todos or create tasks for them. if we will create the tasks it will be easier to prioritize them, then having them as todos in the code. wdyt?

djabarovgeorge (commented):
Looks great for now i left a couple of small comments 
unfortunately, i did not have the time to go over all the pr, I hope I'll have some time a bit later or tomorrow morning 🙏

packages/node/src/lib/workflows/workflow.interface.ts
ainouzgali, line 1: I prefer you wouldn't duplicate the workflow code (create the new workflow folder) from the notification templates folder. We could change the behind implementation inside the notification templates (like the paths). 
Also, I feel this should be a separate task, as we are doing everything gradually.

packages/node/src/lib/novu.ts
```diff
@@ -39,37 +58,139 @@ export class Novu extends EventEmitter {
...
+      if (params) {
+        return await this.http.get(path, params);
+      } else {
+        return await this.http.get(path);
+      }
+    } catch (error: unknown) {
+      this.handleError(error);
+      throw error;
+    }
+  }
```
ainouzgali, line 112: I am not sure I like that it is here. Why are we changing it? Because of the errors?

packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -110,27 +129,53 @@ export class Subscribers extends WithHttp implements ISubscribers {
...
+      { messageId, mark }
+    );
+  }
+
   async markMessageActionSeen(
     subscriberId: string,
     messageId: string,
-    type: string
+    type: ButtonTypeEnum,
+    data: IMarkMessageActionFields
```
ainouzgali, line 171: Isn't this a breaking change? It is not optional... What is behind this?

ainouzgali (commented):
Hi @jainpawan21 ⭐️ 
left you some questions :)

packages/node/src/lib/novu.ts
jainpawan21: Yes, added custom axios methods to return only specific error information
Earlier our SDK was returning axios error. It was hard to debug error information from that error.

packages/node/src/lib/workflows/workflow.interface.ts
jainpawan21: I think I have copied everything now.
Shall I revert back?

apps/api/src/app/feeds/feeds.controller.ts
jainpawan21: yeah sure I will create a task in linear with all those apis

packages/node/src/lib/subscribers/subscribers.ts
jainpawan21: having InApp in method name signifies that this method can mark only in-app messages.
Some users got confused that it can mark other channel messages as well.

apps/api/src/app/change/dtos/change-request.dto.ts
jainpawan21: Sure I will create linear ticket

packages/node/src/lib/novu.ts
jainpawan21: @djabarovgeorge 
I reverted back to WithHttp in `feeds`, `subscribers` and `changes` and added axios custom request getRequest, postRequest etc 

What do you think about this approach?

packages/node/src/lib/subscribers/subscribers.ts
jainpawan21: Yes a breaking change, even though this method implementation was wrong earlier.
User could send any string as type and no status

apps/api/src/app/feeds/feeds.controller.ts
jainpawan21: https://linear.app/novu/issue/NV-2597

apps/api/src/app/change/dtos/change-request.dto.ts
jainpawan21: https://linear.app/novu/issue/NV-2597

packages/node/src/lib/subscribers/subscribers.ts
jainpawan21: @ainouzgali 
if the user sends `{ seen: false }` in this method, the message will be marked as unseen?
or a new ` unseen : boolean ` option will be added in future?

packages/node/src/lib/novu.ts
djabarovgeorge: any reason we can't use WithHttp in all of the classes?

packages/node/src/lib/subscribers/subscribers.ts
djabarovgeorge: This method still can mark other channels as well, so it could confuse the user even more. 
This method is not intended to be only for in-app it should allow potentially mark other channels as well. so users should be aware of the message ids and not mark other channels. 
Although we could allow passing channel type to help users to guard the mark.

packages/node/src/lib/subscribers/subscribers.ts
djabarovgeorge: Yes, the API support { seen: false }.
IMO markInAppMessageAs name could confuse the usee because it can be interpreted as it could mark as true/false.

packages/node/src/lib/novu.ts
jainpawan21: Not a specific reason, We can go ahead in other classes as well. if you like this approach

packages/node/src/lib/subscribers/subscribers.ts
jainpawan21: Shall we update the API for this?

packages/node/src/lib/workflows/workflow.interface.ts
ainouzgali: Yes, IMO lets do the workflow renaming in a different pr

packages/node/src/lib/subscribers/subscribers.ts
jainpawan21: rename this method to `markMessageAs`

packages/node/src/lib/workflows/workflow.interface.ts
jainpawan21: removed workflows folder and class

apps/api/src/app/widgets/widgets.controller.ts
```diff
@@ -56,7 +56,8 @@ import { RemoveAllMessages } from './usecases/remove-messages/remove-all-message
 import { RemoveAllMessagesDto } from './dtos/remove-all-messages.dto';
 
 @Controller('/widgets')
-@ApiExcludeController()
+// @ApiExcludeController()
+@ApiTags('Widgets')
```
ainouzgali, line 60: It shouldn't be externally accessible  - why do you need that? :)

jainpawan21: Sorry,
It was mistake :)
Fixed ✅

packages/node/src/lib/novu.interface.ts
```diff
@@ -10,4 +18,103 @@ export class WithHttp {
...
+      }
+    } catch (error: unknown) {
+      this.handleError(error);
+      throw error;
+    }
+  }
+
+  async postRequest(path: string, data?: unknown): Promise<AxiosResponse> {
+    try {
+      if (data) {
```
LetItRock, line 40: this if statement is not necessary

packages/node/src/lib/novu.interface.ts
```diff
@@ -10,4 +18,103 @@ export class WithHttp {
...
+        case 404: {
+          const { message } = data;
+          throw new NotFoundException(message);
+        }
+        default: {
+          throw new ServerException(status, data.message);
+        }
+      }
+    }
+  }
```
LetItRock, line 119: The error handling will be breaking for the existing clients that do base their error handling logic on the Axios error object.

packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -1,20 +1,20 @@
...
-interface IChannelCredentials {
-  webhookUrl?: string;
-  deviceTokens?: string[];
-}
+import { WithHttp } from '../novu.interface';
 
 export class Subscribers extends WithHttp implements ISubscribers {
+  create(subscriberId: string, data: ISubscriberPayload) {
+    throw new Error('Method not implemented.');
+  }
```
LetItRock, line 17: Maybe let's remove this for now until we implement it? Like the library users might be super angry to see this error in the runtime. Imagine yourself using the library and seeing the method you need and then you think that you're good and you implement some code, but then get this error during the development. Wtf is the thing that first comes from your mouth? I would be super angry.

packages/node/src/lib/novu.ts
jainpawan21: Removed error handling as per Pawel's suggestion
Reverted back these changes

apps/api/src/app/change/dtos/change-request.dto.ts
jainpawan21: remove all comments

packages/node/src/lib/novu.ts
jainpawan21: Removed Error handling as per Pawel's suggestion
Reverted back these changes

apps/api/src/app/feeds/feeds.controller.ts
```diff
@@ -52,6 +52,7 @@ export class FeedsController {
     );
   }
 
+  // TODO: add pagination options
```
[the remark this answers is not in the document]
jainpawan21: done

PR #3843, opened 2023-07-24
feat: create novu providers when organization is created

davidsoderberg:

Add use case to create novu providers and is used from create organization usecase

### DIFF
--- a/apps/api/src/app/integrations/usecases/create-novu-integrations/create-novu-integrations.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/create-novu-integrations/create-novu-integrations.usecase.ts
@@ -1,0 +1,75 @@
+import { Injectable } from '@nestjs/common';
+import { IntegrationEntity, IntegrationRepository } from '@novu/dal';
+
+import { CreateNovuIntegrationsCommand } from './create-novu-integrations.command';
+import { CreateIntegration } from '../create-integration/create-integration.usecase';
+import { CreateIntegrationCommand } from '../create-integration/create-integration.command';
+import { GetNovuIntegration } from '@novu/application-generic';
+
+@Injectable()
+export class CreateNovuIntegrations {
+  constructor(private createIntegration: CreateIntegration, private integrationRepository: IntegrationRepository) {}
+
+  async execute(command: CreateNovuIntegrationsCommand): Promise<IntegrationEntity[]> {
+    const integrations: IntegrationEntity[] = [];
+
+    if (
+      !process.env.NOVU_EMAIL_INTEGRATION_API_KEY ||
+      !process.env.NOVU_SMS_INTEGRATION_ACCOUNT_SID ||
+      !process.env.NOVU_SMS_INTEGRATION_TOKEN ||
+      !process.env.NOVU_SMS_INTEGRATION_SENDER
+    ) {
+      return [];
+    }
+
+    const emailProvider = GetNovuIntegration.createNovuEmailIntegration(null, command.environmentId, true);
+    let count = await this.integrationRepository.count({
+      channel: emailProvider.channel,
+      providerId: emailProvider.providerId,
+      _organizationId: command.organizationId,
+      _environmentId: command.environmentId,
+    });
+
+    if (count === 0) {
+      integrations.push(
+        await this.createIntegration.execute(
+          CreateIntegrationCommand.create({
+            providerId: emailProvider.providerId,
+            channel: emailProvider.channel,
+            active: emailProvider.active,
+            check: false,
+            userId: command.userId,
+            environmentId: command.environmentId,
+            organizationId: command.organizationId,
+          })
+        )
+      );
+    }
+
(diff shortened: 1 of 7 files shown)


apps/api/src/app/integrations/usecases/create-novu-integrations/create-novu-integrations.usecase.ts
```diff
@@ -0,0 +1,75 @@
...
+
+@Injectable()
+export class CreateNovuIntegrations {
+  constructor(private createIntegration: CreateIntegration, private integrationRepository: IntegrationRepository) {}
+
+  async execute(command: CreateNovuIntegrationsCommand): Promise<IntegrationEntity[]> {
+    const integrations: IntegrationEntity[] = [];
+
+    if (
+      !process.env.NOVU_EMAIL_INTEGRATION_API_KEY ||
```
davidsoderberg, line 17: we should not create any integrations if the credentials for them does not exists

davidsoderberg: This also protects self hosted versions from getting these

apps/api/src/app/organization/usecases/create-organization/create-organization.usecase.ts
```diff
@@ -53,7 +56,15 @@ export class CreateOrganization {
       })
     );
 
-    await this.createEnvironmentUsecase.execute(
+    await this.createNovuIntegrations.execute(
```
davidsoderberg, line 59: Creating for dev and prod separate to make it easier make it into events in the future.

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -98,17 +98,24 @@ export class GetNovuIntegration {
...
+          command.environmentId,
+          active
+        );
       default:
         return undefined;
     }
   }
 
-  private createNovuEmailIntegration(
+  public static createNovuEmailIntegration(
```
davidsoderberg, line 116: Refactor these two to use them in create Novu integrations.

apps/api/src/app/integrations/usecases/create-novu-integrations/create-novu-integrations.usecase.ts
```diff
@@ -0,0 +1,75 @@
...
+      !process.env.NOVU_SMS_INTEGRATION_ACCOUNT_SID ||
+      !process.env.NOVU_SMS_INTEGRATION_TOKEN ||
+      !process.env.NOVU_SMS_INTEGRATION_SENDER
+    ) {
+      return [];
+    }
+
+    const emailProvider = GetNovuIntegration.createNovuEmailIntegration(null, command.environmentId, true);
+    let count = await this.integrationRepository.count({
```
scopsy, line 26: To avoid ambiguity, would suggest renaming this to `emailProviderCount` and `smsProviderCount` below, so we won't have to override the count variables

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -98,17 +98,24 @@ export class GetNovuIntegration {
 
     switch (command.channelType) {
       case ChannelTypeEnum.EMAIL:
-        return this.createNovuEmailIntegration(organization, command, active);
+        return GetNovuIntegration.createNovuEmailIntegration(
+          organization,
+          command.environmentId,
+          active
+        );
```
scopsy, line 105: A small question, why do we need to create integrations when we using the get novu integration method? Shouldn't they be created once when the organization is created?

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.usecase.ts
```diff
@@ -143,14 +150,14 @@ export class GetNovuIntegration {
...
   }
 
-  private mapNovuItem(
+  private static mapNovuItem(
     entity: IntegrationEntity,
-    command: GetNovuIntegrationCommand
+    environmentId: string
   ): IntegrationEntity {
-    entity._environmentId = command.environmentId;
+    entity._environmentId = environmentId;
```
scopsy, line 160: Love the more definitive parameter passing here

scopsy (commented):
A few small questions 🙏

packages/application-generic/src/usecases/get-novu-integration/get-novu-integration.usecase.ts
davidsoderberg: yeah the name on that method in GetNovuIntegration is connected to that we "created" them on the fly before. I will rename them so it is clearer now.

scopsy: Ohh okey, makes sense. Thanks for the explanation.

scopsy (approved):

PR #3845, opened 2023-07-24
feat(app-generic): refactor in-memory-provider service

p-fernandez:

Allows to provide the provider choose for any specific InMemoryProviderService instance.
Also it fallbacks to Redis instance if no provider chosen or the configuration validation for the provider chosen fails.

Right now we only have Elasticache as provider implemented with Redis Cluster compatibility. In order to support further Redis Cluster compatible providers we need to refactor InMemoryProviderService.

### DIFF
--- a/packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
+++ b/packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
@@ -5,6 +5,7 @@ import {
   getElasticacheCluster,
   getElasticacheClusterProviderConfig,
   IElasticacheClusterProviderConfig,
+  validateElasticacheClusterProviderConfig,
 } from './elasticache-cluster-provider';
 import {
   CLIENT_READY,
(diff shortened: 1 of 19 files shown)


packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -121,43 +133,60 @@ export class InMemoryProviderService {
...
 
-  private isElasticacheEnabled(): boolean {
-    return (
-      !!process.env.ELASTICACHE_CLUSTER_SERVICE_HOST &&
-      !!process.env.ELASTICACHE_CLUSTER_SERVICE_PORT
-    );
-  }
-
-  private getClientAndConfigForCluster(): {
+  public getClientAndConfigForCluster(providerId: InMemoryProviderEnum): {
```
p-fernandez, line 136: Wanted to unit test it so changed the privacy scope. Can be temporary.

LetItRock: btw you can hack it in the tests, like have this one still private and in the tests access it like this:
```
(obj as any).functionName();
```
the "private" scope is a TS thing, but the code transpiles to JS at the end ;)

LetItRock (approved):
🎸 ⭐

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
p-fernandez: I am sorry but that's not the best advice as TS already has the bracket notation accessor as safe escape while keeping the typing. https://github.com/microsoft/TypeScript/issues/19335
At the end I moved in a different PR this method as generic function in a private Node.js module, so that settles my discomfort about this change. 😅

PR #3846, opened 2023-07-24
NV-2601 - Add support in the client for tenant feature

djabarovgeorge:

Feature flag to hide Tenant functionality

We introducing new Tenant functionality that needs to be hidden before the general auditory release. We are planning to roll out that functionality to some early birds before the GA release.

### Definition of Done

* A feature flag is created in Lanch Darkly
* Created related feature flag hook


apps/web/src/config/index.ts
```diff
@@ -68,3 +68,7 @@ export const IS_MULTI_PROVIDER_CONFIGURATION_ENABLED = isCypress
   : window._env_.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED ||
     process.env.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED ||
     'false';
+
+export const IS_MULTI_TENANCY_ENABLED = isCypress
+  ? window._env_.IS_MULTI_TENANCY_ENABLED || process.env.IS_MULTI_TENANCY_ENABLED || 'true'
```
djabarovgeorge, line 73: Was not sure if to call it multi-tenancy or just tenant/tenants, naming 🤦‍♂️

ainouzgali (approved):
🚀

apps/web/src/config/index.ts
ainouzgali: I think it sounds good

PR #3852, opened 2023-07-25
feat: remove get novu integration usecase

davidsoderberg:

We will not need the usecase when we store the Novu providers in database.


packages/application-generic/src/factories/mail/mail.factory.ts
```diff
@@ -48,7 +49,12 @@ export class MailFactory {
   ): IMailHandler {
     const handler =
       this.handlers.find((handlerItem) =>
-        handlerItem.canHandle(integration.providerId, integration.channel)
+        handlerItem.canHandle(
```
davidsoderberg, line 52: I think this is the base place to handle it now days. It was not here before because then we used the stateless library instead and that one was much more used by others.

packages/application-generic/src/factories/sms/sms.factory.ts
```diff
@@ -44,7 +45,12 @@ export class SmsFactory implements ISmsFactory {
   getHandler(integration: IntegrationEntity) {
     const handler =
       this.handlers.find((handlerItem) =>
-        handlerItem.canHandle(integration.providerId, integration.channel)
+        handlerItem.canHandle(
```
davidsoderberg, line 48: I think this is the best place to handle it now days. It was not here before because then we used the stateless library instead and that one was much more used by others.

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -53,6 +53,20 @@ export class CreateIntegration {
       }
     }
 
+    if (command.providerId === SmsProviderIdEnum.Novu || command.providerId === EmailProviderIdEnum.Novu) {
```
davidsoderberg, line 56: We should only allow having one novu provider of each type

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -120,6 +134,20 @@ export class CreateIntegration {
         });
       }
 
+      if (command.channel === ChannelTypeEnum.EMAIL && command.providerId !== EmailProviderIdEnum.Novu) {
```
davidsoderberg, line 137: we need to deactivate npvu provider if they create another type of integration

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -656,6 +657,19 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
...
+        {
+          _organizationId: session.organization._id,
+          _environmentId: session.environment._id,
+          active: false,
+        },
+        {
+          $set: {
+            active: true,
+          },
+        }
```
p-fernandez, line 670: I do not understand this addition. Why do you need to set all the integrations to active? In the creation of the organization Novu providers should be enabled, right? What I am missing?

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
p-fernandez: Besides covering the case that @ainouzgali raised that is very important, what do you think about creating an use case to isolate this logic, temptively: `DisableNovuIntegration` where we could you pass in the command the `channel` and the `environmentId`? Based on the channel we would know which `providerId` to use.

libs/testing/src/integration.service.ts
```diff
@@ -80,6 +80,28 @@ export class IntegrationService {
...
+    const novuSmsPayload = {
+      _environmentId: environmentId,
+      _organizationId: organizationId,
+      providerId: SmsProviderIdEnum.Novu,
+      channel: ChannelTypeEnum.SMS,
+      credentials: {},
+      active: false,
+    };
+
+    await this.integrationRepository.create(novuSmsPayload);
```
p-fernandez, line 103: They can be executed in parallel (we are not doing any kind of error management so it would work the same if any failure).

packages/application-generic/src/factories/mail/mail.factory.ts
p-fernandez: Have you considered creating a NovuEmailHandler to avoid the condition and where we can store the provider used under the hood?
Also with that we might be able to clean up a bit how we avoid the self hosted users to not have it.

packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -94,33 +48,16 @@ export class GetDecryptedIntegrations {
...
+      .map((integration: IntegrationEntity) =>
+        GetDecryptedIntegrations.decryptCredentials(integration)
       );
+  }
 
-      return novuIntegration ? [novuIntegration] : [];
-    }
-
-    const novuIntegrations = await this.getNovuIntegrations(command);
+  public static decryptCredentials(integration: IntegrationEntity) {
```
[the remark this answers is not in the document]
p-fernandez: `getDecryptedCredentials` to round @ainouzgali suggestion. 🙂

libs/testing/src/integration.service.ts
davidsoderberg: This is in test that why I did not do any error management.

apps/api/src/app/events/e2e/trigger-event.e2e.ts
davidsoderberg: Because we handle everything from repository level so we need to activate Novu providers here to test them.

packages/application-generic/src/factories/mail/mail.factory.ts
davidsoderberg: No but I like your suggestion so will do that :)

apps/api/src/app/integrations/usecases/disable-novu-integration/disable-novu-integration.command.ts
```diff
@@ -0,0 +1,13 @@
...
+import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
+
+export class DisableNovuIntegrationCommand extends EnvironmentWithUserCommand {
+  @IsDefined()
+  @IsEnum(ChannelTypeEnum)
+  channel: ChannelTypeEnum;
+
+  @IsDefined()
+  @IsString()
+  providerId: string;
```
davidsoderberg, line 12: This is needed so we can know that they do not create a Novu provider

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -120,6 +136,16 @@ export class CreateIntegration {
         });
       }
 
+      if (integrationEntity.active) {
```
davidsoderberg, line 139: Should I move this to the disable Novu integration usecase?

p-fernandez: I think not because this condition is related to the integration being created, not the Novu integration we want to disable.

packages/application-generic/src/factories/sms/sms.factory.ts
```diff
@@ -19,6 +19,8 @@ import {
   AfricasTalkingSmsHandler,
   SendchampSmsHandler,
 } from './handlers';
+import { SmsProviderIdEnum } from '@novu/shared';
+import { NovuSmsHandler } from './handlers/novu.handler';
```
p-fernandez, line 23: You can move the `NovuSmsHandler` to the `./handlers` import of line 21 and above.

p-fernandez (approved):
Slick! 🌟

PR #3853, opened 2023-07-25
refactor: update asset

BiswaViraj:

Updates the no new notification image inside `notification-center` and `widget`

image


packages/notification-center/src/images/no-new-notifications.svg
ainouzgali, line 1: This does not take into account the light vs dark scheme. IMO you should wrap it as a react component (like we do for icons) and pass props and use "currentColor" where needed

scopsy (approved):

PR #3855, opened 2023-07-25
feat: add initial tenants ui

djabarovgeorge:

The new Tenants page will show the list of tenants.

The API endpoint to get the tenants list was created in the previous cycle.

We need a page to manage tenants

* The title "Tenants" is displayed
* Add Tenant button is displayed
* Display tenants as a list with the columns:
* Name – Tenant Name
* Tenant identifier – Tenant Identifier
* Created at – creation time in the format DD/MM/YYYY  HH: MM
* Updated at – creation time in the format DD/MM/YYYY  HH: MM
* Display a skeleton when loading list of the tenants

image

https://www.loom.com/share/6bb92713c4714538a1e853dcba0f93af


apps/web/src/pages/tenants/components/list/columns.tsx
djabarovgeorge, line 1: was not sure if `withCellLoading` is needed here or not.

apps/web/src/pages/tenants/components/list/ToolBar.tsx
djabarovgeorge, line 1: basically the same as in the integration list page was not sure if creating a reusable component for both or not. due to the reason that this component is small and used in 2 places, I decided to duplicate the code for now. lmkwyt.

apps/web/src/App.tsx
```diff
@@ -201,7 +201,7 @@ function App() {
                     <Route path=":channel/:stepUuid" element={<TemplateEditor />} />
                   </Route>
                   <Route path={ROUTES.WORKFLOWS} element={<WorkflowListPage />} />
-                  <Route path={ROUTES.TENANTS} element={<Tenants />} />
+                  <Route path={ROUTES.TENANTS} element={<TenantsPage />} />
```
djabarovgeorge, line 204: added the `page` part to make sure it is straightforward.

apps/web/src/pages/tenants/TenantsPage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import PageContainer from '../../components/layout/components/PageContainer';
+import PageHeader from '../../components/layout/components/PageHeader';
+import { TenantsList } from './components/list/TenantsList';
+
+export function TenantsPage() {
+  const navigate = useNavigate();
+
+  const onAddTenantClickCallback = useCallback(() => {
+    // navigate();
```
djabarovgeorge, line 12: we will navigate hit to the side bar that Gali is working on.

apps/web/src/pages/tenants/components/list/columns.tsx
```diff
@@ -0,0 +1,42 @@
...
+  },
+  {
+    accessor: 'createdAt',
+    Header: 'Create at',
+    maxWidth: maxWidth,
+    Cell: ({ row: { original }, isLoading }) => {
+      return isLoading ? (
+        <Skeleton width={100} height={20} />
+      ) : (
+        format(new Date(original.createdAt), 'dd/MM/yyyy HH:mm')
```
djabarovgeorge, line 26: requested in the ticked by the design

apps/web/src/pages/tenants/components/list/TenantsList.tsx
```diff
@@ -1,3 +1,31 @@
...
+import { Table } from '../../../../design-system';
+import PageContainer from '../../../../components/layout/components/PageContainer';
+import { When } from '../../../../components/utils/When';
+import { columns } from './columns';
+import { useTenants } from '../../../../hooks/useTenants';
+import { Toolbar } from './ToolBar';
+
+export const TenantsList = ({ onAddTenantClick }: { onAddTenantClick: React.MouseEventHandler<HTMLButtonElement> }) => {
+  const { tenants, loading: tenantLoading } = useTenants();
+  const isLoading = tenantLoading;
```
ainouzgali, line 12:
```suggestion
  const { tenants, loading: tenantsLoading } = useTenants();
```

unnecessary isLoading param

apps/web/src/pages/tenants/components/list/columns.tsx
ainouzgali: I believe it is :) It is now used in all tables

apps/web/src/pages/tenants/components/list/columns.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const columns: IExtendedColumn<ITenantEntity>[] = [
+  {
+    accessor: 'name',
+    Header: 'Name',
+  },
+  {
+    accessor: 'identifier',
+    Header: 'Tenant identifier',
+  },
```
ainouzgali, line 17: Let's add the withCellLoading and also an ellipsis cut for long names/identifiers with a tooltip for the full name? wdyt?

I mean like this: 
```
 Cell: withCellLoading(({ row: { original } }) => (
      <Tooltip label={original.name}>
        <Text rows={1}>{original.name}</Text>
      </Tooltip>
    )),
```

apps/web/src/pages/tenants/components/list/columns.tsx
```diff
@@ -0,0 +1,42 @@
...
+  {
+    accessor: 'createdAt',
+    Header: 'Create at',
```
ainouzgali, line 20:
```suggestion
    Header: 'Created At',
```

apps/web/src/pages/tenants/components/list/columns.tsx
```diff
@@ -0,0 +1,42 @@
...
+  },
+  {
+    accessor: 'updatedAt',
+    Header: 'Updated at',
+    maxWidth: maxWidth,
+    Cell: ({ row: { original }, isLoading }) => {
+      return isLoading ? (
+        <Skeleton width={100} height={20} />
+      ) : (
+        format(new Date(original.createdAt), 'dd/MM/yyyy HH:mm')
```
ainouzgali, line 38:
```suggestion
        format(new Date(original.updatedAt), 'dd/MM/yyyy HH:mm')
```

libs/shared/src/entities/tenant/tenant.interface.ts
```diff
@@ -0,0 +1,21 @@
...
+  createdAt: string;
+
+  updatedAt: string;
+
+  data?: TenantCustomData;
+
+  _environmentId: EnvironmentId;
+}
+
+export type TenantId = string;
```
ainouzgali, line 21: You could add that TenantId type to the types folder as well as you already have one for `TenantCustomData`

apps/web/src/pages/tenants/components/list/columns.tsx
```diff
@@ -0,0 +1,42 @@
...
+      );
+    },
+  },
+  {
+    accessor: 'updatedAt',
+    Header: 'Updated at',
```
ainouzgali, line 32:
```suggestion
    Header: 'Updated At',
```

apps/web/src/pages/tenants/components/list/ToolBar.tsx
ainouzgali, line 1: What do you think about extracting this to a ListToolbar component? As it is pretty identical to the one we have for integrations. And will probably be the standard from now on to all tables

ainouzgali (approved):
🤩💃🏼💙

apps/web/src/pages/tenants/components/list/columns.tsx
djabarovgeorge: Make it like this because of the design 🙃

ainouzgali: Then just create -> created

apps/web/src/pages/tenants/components/list/columns.tsx
djabarovgeorge: i wonder what is the main use of it here because i still get the skeletons etc in the view as expected.

apps/web/src/pages/tenants/components/list/columns.tsx
djabarovgeorge: Oh sorry you're right! 
it should be `Created at` missed the past part

apps/web/src/pages/tenants/components/list/columns.tsx
djabarovgeorge: love it! thanks for the suggetion

libs/shared/src/entities/tenant/tenant.interface.ts
djabarovgeorge: Good point, done :)

apps/web/src/pages/tenants/components/list/columns.tsx
djabarovgeorge: never mind i see why now, it is much cleaner with it. 😅

apps/web/src/pages/tenants/components/list/ToolBar.tsx
djabarovgeorge: due the reason this component pretty small i thought about refactoring on the next time we will need to reuse it 
https://github.com/novuhq/novu/pull/3855#discussion_r1273826869

PR #3860, opened 2023-07-26
NV 2602 - Hide Tenant functionality under the feature flag

djabarovgeorge:

Feature flag to hide Tenant functionality.

We introducing new Tenant functionality that needs to be hidden before the general auditory release. We are planning to roll out that functionality to some early birds before the GA release.

### Definition of Done
* Tenant functionality is hidden under the feature flag

### DIFF
--- a/apps/web/src/components/layout/components/SideNav.tsx
+++ b/apps/web/src/components/layout/components/SideNav.tsx
@@ -81,7 +81,7 @@ export function SideNav({}: Props) {
   const lastStep = currentOnboardingStep().get();
   const getStartedRoute = lastStep === ROUTES.GET_STARTED_PREVIEW ? ROUTES.GET_STARTED : lastStep;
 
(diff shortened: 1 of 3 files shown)


apps/web/src/components/layout/components/SideNav.tsx
```diff
@@ -121,6 +122,10 @@ export function SideNav({}: Props) {
     },
   ];
 
+  if (!isMultiTenancyEnabled) {
```
ainouzgali, line 125: You can use condition property of the menu item - like in Changes. 

```
{
 icon:<Building />,
.....,
condition: isMultiTenancyEnabled
```

djabarovgeorge: i was not aware of the condition param 😅
much better, thanks!

p-fernandez (approved):
🌟

PR #3864, opened 2023-07-27
feat: add delete item and name fields for novu providers

davidsoderberg:
Change to the normal update header for provider and add name + identifier fields.

Screenshot 2023-07-27 at 07 51 35

### DIFF
--- a/apps/web/src/pages/integrations/components/UpdateIntegrationCommonFields.tsx
+++ b/apps/web/src/pages/integrations/components/UpdateIntegrationCommonFields.tsx
@@ -5,6 +5,7 @@ import { useClipboard } from '@mantine/hooks';
 import { Input, Switch } from '../../../design-system';
 import { Check, Copy } from '../../../design-system/icons';
 import type { IIntegratedProvider } from '../types';
+import { When } from '../../../components/utils/When';
 
 const CopyWrapper = styled.div`
   cursor: pointer;
@@ -13,7 +14,13 @@ const CopyWrapper = styled.div`
   }
 `;
 
-export const UpdateIntegrationCommonFields = ({ provider }: { provider: IIntegratedProvider | null }) => {
+export const UpdateIntegrationCommonFields = ({
+  provider,
+  showActive = true,
+}: {
+  provider: IIntegratedProvider | null;
+  showActive?: boolean;
+}) => {
   const {
     control,
     formState: { errors },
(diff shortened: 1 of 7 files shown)


apps/web/src/pages/integrations/components/UpdateIntegrationCommonFields.tsx
```diff
@@ -24,19 +31,21 @@ export const UpdateIntegrationCommonFields = ({ provider }: { provider: IIntegra
...
-        render={({ field }) => (
-          <Switch
-            checked={field.value}
-            label={field.value ? 'Active' : 'Disabled'}
-            data-test-id="is_active_id"
-            {...field}
-          />
-        )}
-      />
+      <When truthy={showActive}>
```
davidsoderberg, line 34: Add this because we should not allow user to activate and deactivate a novu provider

apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
```diff
@@ -15,9 +15,11 @@ import { ProviderInfo } from './multi-provider/ProviderInfo';
 export const UpdateIntegrationSidebarHeader = ({
   provider,
   onSuccessDelete,
+  children = null,
 }: {
   provider: IIntegratedProvider | null;
   onSuccessDelete: () => void;
+  children?: ReactNode | null;
```
davidsoderberg, line 22: To be able to show the Free tag for Novu provider

apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
```diff
@@ -159,23 +157,38 @@ export function UpdateProviderSidebar({
...
-            <Free>🎉 Free</Free>
-          </Group>
-        }
-        data-test-id="update-provider-sidebar-novu"
-      >
-        <ProviderInfo provider={selectedProvider} environments={environments} />
-
-        <NovuProviderSidebarContent provider={selectedProvider} />
-      </Sidebar>
+      <FormProvider {...methods}>
```
davidsoderberg, line 160: I keept adding stuff up here because I did not like to make the part lower down even more complex.

apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
```diff
@@ -184,15 +197,15 @@ export function UpdateProviderSidebar({
       <Sidebar
         isOpened={isSidebarOpened}
         isLoading={areProvidersLoading || areEnvironmentsLoading}
-        isExpanded={sidebarState === 'expanded'}
+        isExpanded={sidebarState === SidebarStateEnum.EXPANDED}
```
davidsoderberg, line 200: And some small refactoring

apps/web/src/pages/integrations/components/UpdateIntegrationCommonFields.tsx
djabarovgeorge: I guess it is by design, but i am not sure it is the best way going forward, for a couple of reasons:
* there could be a case where the user won't want to use `Novu provider` due user's internal business interests (the easy way to handle it is to deactivate it).
* especially in the integration conditions that we will add there could be a case where the user won't want to have Novu as an optional provider as a backoff opportunity (the easy way to handle it is to deactivate it).
* use a specific provider (and not Novu's) even if it is failing.
* still having if statements in the code related to the Novu provider. 😅

davidsoderberg: We only allow using the Novu provider if they do not have any other provider for the same channel so this should not be a case.

djabarovgeorge: Actually, this is the problem i had in my mind but it is the current behavior and not part of this scope, them never mind for not :)

djabarovgeorge (approved):
Looks good to me :)

PR #3865, opened 2023-07-27
fix(app-generic): adding context and information for logs for debugging

p-fernandez:

Improving context.

To help debug prod issues.

### DIFF
--- a/packages/application-generic/src/usecases/add-job/add-job.usecase.ts
+++ b/packages/application-generic/src/usecases/add-job/add-job.usecase.ts
@@ -24,6 +24,8 @@ export enum BackoffStrategiesEnum {
   WEBHOOK_FILTER_BACKOFF = 'webhookFilterBackoff',
 }
 
+const LOG_CONTEXT = 'AddJob';
+
 @Injectable()
 export class AddJob {
   constructor(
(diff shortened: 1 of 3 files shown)


packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -37,40 +39,47 @@ export class AddJob {
...
+      );
 
       return;
     }
 
     if (digestAmount === undefined && delayAmount === undefined) {
       Logger.verbose(
-        'updating status as digestAmount and delayAmount is undefined'
+        `Updating status as digestAmount and delayAmount is undefined for job ${job._id}`,
+        LOG_CONTEXT
```
ainouzgali, line 82: I feel this relays an incorrect state - if I understand correctly, here it realizes it is not a delay or a digest, so its a normal step and he updates its status to Queued like it should. 
Maybe we could add to the message that the 'Updating status' is to queued

ainouzgali (approved):

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
p-fernandez: Yeah, we can simplifiy the message.

PR #3867, opened 2023-07-27
feat: remove novu providers for select provider

davidsoderberg:

Remove Novu providers from select provider sidebar and from use providers if self hosted.


davidsoderberg:
@scopsy and I decided that we should not allow to create new Novu providers even if the old one is deleted by them.

p-fernandez (approved):
🌟

PR #3868, opened 2023-07-27
chore(*): remove `gitHead` field from all packages

marvinjude:

`gitHead` is a temporary property only meant to exist during lerna publish when it is constructing the package tarballs. AFAIK, you probably have this as a result of some failed publish and Ideally, you'd want to remove it. 

This PR removes all `gitHead` fields in `package.json`

#### References

https://github.com/mui/mui-x/pull/669  - *all my package.json's now contain gitHead key. What to do about it?*
https://github.com/gatsbyjs/gatsby/pull/12584   -  *remove all `gitHead` from package.json*
https://github.com/lerna/lerna/issues/1880 - *Remove gitHead*


p-fernandez (approved):
🌟

PR #3871, opened 2023-07-28
fix: so activating a provider deactivates novu provider

davidsoderberg:

To disable Novu provider if another provider is activated


p-fernandez (approved):
🌟 
Now the usefulness of having the separate use case shows up. Well done.

PR #3873, opened 2023-07-28
[WIP] 🚧 feat(api): set integration as primary and priority system

LetItRock:

Set integration as "primary" endpoint and priority system logic:
1. One integration per channel could be marked as primary.
2. Only Email and SMS integrations could be marked as primary and will apply priority to the integrations. For the Chat and Push integrations messages are delivered for each channel, potentially different integrations, meaning that we can't have a "primary" integration concept or priority system. For In-App we only have one provider rn, but in future, we might have many per application, so there also we can't apply these concepts.
3. When we do create or update the first active integration (excluding Novu providers) for the Email or SMS channel, that integration will become automatically primary.
4. I will describe all the other logic in comments on the code.

### DIFF
--- a/apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.command.ts
+++ b/apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.command.ts
@@ -1,0 +1,13 @@
+import { IsDefined, IsMongoId } from 'class-validator';
+
+import { OrganizationCommand } from '../../../shared/commands/organization.command';
+
+export class SetIntegrationAsPrimaryCommand extends OrganizationCommand {
+  @IsDefined()
+  @IsMongoId()
+  integrationId: string;
+
+  @IsDefined()
+  @IsMongoId()
+  environmentId: string;
+}
--- a/apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.usecase.ts
@@ -1,0 +1,109 @@
+import { Injectable, NotFoundException, Logger, BadRequestException } from '@nestjs/common';
+import { IntegrationEntity, IntegrationRepository } from '@novu/dal';
+import { CHANNELS_WITH_PRIMARY } from '@novu/shared';
+import {
+  AnalyticsService,
+  buildIntegrationKey,
+  InvalidateCacheService,
+  GetFeatureFlag,
+  FeatureFlagCommand,
+} from '@novu/application-generic';
+
+import { SetIntegrationAsPrimaryCommand } from './set-integration-as-primary.command';
+
+@Injectable()
+export class SetIntegrationAsPrimary {
+  constructor(
+    private invalidateCache: InvalidateCacheService,
+    private integrationRepository: IntegrationRepository,
+    private analyticsService: AnalyticsService,
+    private getFeatureFlag: GetFeatureFlag
+  ) {}
+
+  private async updatePrimaryFlag({ existingIntegration }: { existingIntegration: IntegrationEntity }) {
+    await this.integrationRepository.update(
+      {
+        _organizationId: existingIntegration._organizationId,
+        _environmentId: existingIntegration._environmentId,
+        channel: existingIntegration.channel,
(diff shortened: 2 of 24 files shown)


apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.command.ts
```diff
@@ -0,0 +1,13 @@
...
+import { OrganizationCommand } from '../../../shared/commands/organization.command';
+
+export class SetIntegrationAsPrimaryCommand extends OrganizationCommand {
+  @IsDefined()
+  @IsMongoId()
+  integrationId: string;
+
+  @IsDefined()
+  @IsMongoId()
+  environmentId: string;
```
djabarovgeorge, line 12: should we reuse EnvironmentWithUserCommand as base here?

apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.usecase.ts
```diff
@@ -0,0 +1,109 @@
...
+      channel,
+      _organizationId,
+      _environmentId,
+    });
+
+    await this.invalidateCache.invalidateQuery({
+      key: buildIntegrationKey().invalidate({
+        _organizationId,
+      }),
+    });
```
djabarovgeorge, line 89: 🚀

apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.usecase.ts
```diff
@@ -0,0 +1,109 @@
...
+          primary: true,
+        },
+      }
+    );
+  }
+
+  async execute(command: SetIntegrationAsPrimaryCommand): Promise<IntegrationEntity> {
+    Logger.verbose('Executing Set Integration As Primary Usecase');
+
+    const existingIntegration = await this.integrationRepository.findById(command.integrationId);
```
djabarovgeorge, line 57: can we add here environment id as well as a guard just in case?

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -29,6 +29,88 @@ export class UpdateIntegration {
...
+        _environmentId: existingIntegration._environmentId,
+        channel: existingIntegration.channel,
+        exclude: true,
+      });
+
+      result = {
+        priority: 0,
+        primary: false,
+      };
+    }
```
djabarovgeorge, line 109: does this mean that if i update inactive integration it will update with a priority of 0 and primary false? 
if so it feels a bit like a side effect, would not we expect to get the same order we had before (if we deactivate and activate) so when we do recalculatePriorityForAllActive so the original order remains the same?

I hope I got it right. the side effect i think we have is that all the diactivated integration will go to the last priority right away but it could be an expected behavior

djabarovgeorge (commented):
Everything looks great! :) I just had a couple of questions sorry about it, if my concerns are by design I can approve it just let me know

another small thing i think create and update usecase shares some logic maybe we could extract it as well but totally could we for future refactoring.

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
djabarovgeorge: > means the user doesn't want to use it.
or diactivated for some period of time due to internal issues or even the user is playing with the dashboard.
but at this stage as you said we can leave it like this and then update **if needed**

djabarovgeorge (approved):
Yeah looks great to me awesome work :)

libs/dal/src/repositories/integration/integration.repository.ts
```diff
@@ -70,4 +104,49 @@ export class IntegrationRepository extends BaseRepository<IntegrationDBModel, In
...
+  }) {
+    const otherActiveIntegrations = await this.find(
+      {
+        _organizationId,
+        _environmentId,
+        channel,
+        active: true,
+        _id: {
+          $nin: [_id],
+        },
```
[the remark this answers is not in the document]
djabarovgeorge: actually, i looked at it and i think it should work, but not a priority for now.

PR #3875, opened 2023-07-29
chore(api): migration script for primary and priority integration fields

LetItRock:

The migration script to populate the primary and priority integration fields.

Screenshot 2023-07-30 at 00 28 12


apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
```diff
@@ -0,0 +1,139 @@
...
+      );
+      continue;
+    }
+
+    console.log('\n------------------------------------------');
+    console.log(
+      `Migrating integrations for the "${organization.name}" organization in the ${environment.name} environment:`
+    );
+
+    for (const channel of CHANNELS_WITH_PRIMARY) {
```
LetItRock, line 38: Only the `email` and `sms` channels do support the primary and priority fields

apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
```diff
@@ -0,0 +1,139 @@
...
+      const integrations = await integrationRepository.find(
+        {
+          _organizationId: organization._id,
+          _environmentId: environment._id,
+          channel,
+          active: true,
+        },
+        undefined,
+        { sort: { createdAt: -1 } }
+      );
```
LetItRock, line 48: find all active integrations for a channel sorted by `createdAt` date desc

apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
```diff
@@ -0,0 +1,139 @@
...
+        await integrationRepository.update(
+          {
+            _id: integration._id,
+            _environmentId: integration._environmentId,
+            _organizationId: integration._organizationId,
+          },
+          {
+            $set: payload,
+          }
+        );
```
LetItRock, line 68: update the integration: the last created is primary... the priority is set like the biggest number is the highest priority

apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
```
LetItRock, line 97: for all inactive integrations set primary to false and priority to 0

apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
```diff
@@ -0,0 +1,139 @@
...
```
LetItRock, line 126: We don't support primary and priority for the In-App, Push, Chat. Because one subscriber can have multiple channels assigned, ex. two Push channels for two devices that are both "active" - so "primary" and with the highest "priority".

djabarovgeorge (commented):
Looks amazing! Personally, i forgot about the last two parts of your script.

What do you think about adding gourds so we can run this script twice, in case we run this one and while the script is running users will create integrations? your code should work but i suspect it could cause unpredicted side effects. Or we won't run this script twice? 😅

LetItRock:
> Looks amazing! Personally, i forgot about the last two parts of your script.
> 
> What do you think about adding gourds so we can run this script twice, in case we run this one and while the script is running users will create integrations? your code should work but i suspect it could cause unpredicted side effects. Or we won't run this script twice? 😅

thanks, George for the suggestion, we will check the items in the DB after running this script, and if there would be any we might run the script again ;)

apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
```diff
@@ -0,0 +1,139 @@
...
+      const integrations = await integrationRepository.find({
+        _organizationId: organization._id,
+        _environmentId: environment._id,
+        channel,
+      });
+      console.log('------');
+      console.log(`Found ${integrations.length} integrations for the ${channel} channel`);
+
+      for (let i = 0; i < integrations.length; i++) {
+        const integration = integrations[i];
```
djabarovgeorge, line 110: what do you think about doing a small check if primary/priority already exists in the integration and if not skip the update?

djabarovgeorge: @LetItRock

LetItRock: @djabarovgeorge for the above channels there won't be primary integrations... but if you mean for the sms/email and you are worried that this migration might fail, I would say that we will run it in the beginning without the updates just to get some statistics and see if it will pass, and if everything will be ok then we will run it with updates... 

anyway, even if it will fail it's "safe" to run it a second time, because the integrations are updated by order of `createdAt` desc...

djabarovgeorge: I am not afraid of failure just for side effects for the user if it will time from the fist and second run. But if the difference will be small there shouldn't be issue.

djabarovgeorge (approved):
🔥

apps/api/migrations/integration-scheme-update/add-primary-priority-migration.ts
LetItRock: we will figure it out :)

PR #3876, opened 2023-07-30
fix: allow return 'data' as response prop

djabarovgeorge:

allow return 'data' as response prop

so we could create entities with a 'data' prop on the root layer and return them properly.


apps/api/src/app/shared/framework/response.interceptor.ts
```diff
@@ -17,7 +17,7 @@ export class ResponseInterceptor<T> implements NestInterceptor<T, Response<T>> {
     return next.handle().pipe(
       map((data) => {
         // For paginated results that already contain the data wrapper, return the whole object
-        if (data?.data) {
+        if (data?.data && !data?._id) {
```
scopsy, line 20: Amazing! Maybe worth adding a small comment here on why it's added like this here.

scopsy (approved):

scopsy (approved):

PR #3882, opened 2023-07-31
test: fix failing test because of a run on same day

ainouzgali:

This test would fail every time we run it on the 31th or the 1th of every month. 
Because today's value is already selected by default and in the test we click on the 31 and 1 -> so if the run is on that day we de-select them.


davidsoderberg (approved):
🥳

PR #3886, opened 2023-07-31
feat: add migration script to add novu providers to database

davidsoderberg:

Add migration script for Novu providers.

### DIFF
--- a/apps/api/migrations/novu-integrations/novu-integrations.migration.ts
+++ b/apps/api/migrations/novu-integrations/novu-integrations.migration.ts
@@ -1,0 +1,85 @@
+import '../../src/config';
+import {
+  OrganizationRepository,
+  EnvironmentRepository,
+  IntegrationRepository,
+  ChannelTypeEnum,
+  EnvironmentEntity,
+} from '@novu/dal';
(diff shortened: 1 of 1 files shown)


apps/api/migrations/novu-integrations/novu-integrations.migration.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  if (count > 0) {
+    return;
+  }
+
+  const countChannelIntegrations = await integrationRepository.count({
+    _environmentId: environment._id,
+    _organizationId: environment._organizationId,
+    channel,
+  });
```
djabarovgeorge, line 38: Would we want to skip the creation if the user already has integration on the same channel?

djabarovgeorge (approved):
Looks great to me left a small question regarding the open question for me. :)

apps/api/migrations/novu-integrations/novu-integrations.migration.ts
davidsoderberg: No we just like to keep it disabled so the user that are used to see them still see them as usual even with the new list design

PR #3889, opened 2023-07-31
Stacked PR's Branch for Tenants Support

djabarovgeorge:


libs/shared/src/entities/tenant/tenant.interface.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  deleted?: boolean;
+
+  createdAt: string;
+
+  updatedAt: string;
+
+  data?: TenantCustomData;
+
+  _environmentId: EnvironmentId;
```
ainouzgali, line 18: Could you also add the _organizationId here?

ainouzgali (approved):

PR #3890, opened 2023-08-01
Allow storing custom metadata on the workflow entity

djabarovgeorge:

* Allow customers to add and update custom metadata information (JSON Object) and pass it to the workflow
* When fetching the preference API, attach the workflow information as a joined object with identifier, and custom metadata provided there
* Other place which populate the workflow objects such as notification feed, add the custom object metadata to it.

When displaying a custom UI for the user preference menu, customers would love to render those preference items with some extra information or filter our specific preferences based on this custom metadata.

### Definition of Done
* Allow storing custom data using the API on the workflow
* Populate workflow custom data on the user preference API response
* On the notification center react component, add a preferences filter function, that expects a return of a boolean to show or skip showing a particular preference. The filter function should pass the preference with the template populate custom fields. So the user can make his own custom logic to when display something. 
*(Optional) Allow viewing custom metadata on the user interface in the workflow UI

### Other information
Unfortunately, i make one PR containing all the logic due to time constrictions.


packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -249,5 +249,6 @@ function mapTemplateConfiguration(
     _id: template._id,
     name: template.name,
     critical: template.critical != null ? template.critical : true,
+    ...(template.data ? { data: template.data } : {}),
```
djabarovgeorge, line 252: made it like this so we won't store nullish param

packages/notification-center/src/components/notification-center/components/user-preference/SubscriberPreference.tsx
```diff
@@ -17,13 +17,19 @@ export function SubscriberPreference() {
...
   const [rootStyles, accordionItemStyles, accordionContentStyles, accordionControlStyles, accordionChevronStyles] =
     useStyles(['preferences.root', 'accordion.item', 'accordion.content', 'accordion.control', 'accordion.chevron']);
 
   const baseTheme = theme?.userPreferences;
-  const preferences = data
+  let preferences = data
     ?.filter((item) => !item.template.critical)
     ?.filter((pref) => Object.keys(pref.preference.channels).length > 0);
+
+  if (preferenceFilter && preferences) {
```
djabarovgeorge, line 29: here we use the filter function.
added it under the if statement and not under nested filtering above because in most cases we the user won't provide a filter function and the iteration will be refunded.

libs/dal/src/repositories/notification-template/notification-template.entity.ts
```diff
@@ -61,6 +61,8 @@ export class NotificationTemplateEntity {
   isBlueprint: boolean;
 
   blueprintId?: string;
+
+  data?: NotificationTemplateCustomData;
```
djabarovgeorge, line 65: created it under the same structure we have the subscriber data.

apps/widget/src/components/notification-center/NotificationCenterWidget.tsx
djabarovgeorge, line 1: added support in the widget as well so we will have cypress test to this functionality ✨

packages/notification-center/src/store/notification-center.context.ts
```diff
@@ -1,12 +1,17 @@
...
+
 import { INotificationCenterContext, ITab } from '../shared/interfaces';
 
 export const NotificationCenterContext = React.createContext<INotificationCenterContext>({
   onUrlChange: (url: string) => {},
   onNotificationClick: (notification: IMessage) => {},
   onUnseenCountChanged: (unseenCount: number) => {},
   onActionClick: (identifier: string, type: ButtonTypeEnum, message: IMessage) => {},
+  onTabClick: (tab: ITab) => {},
+  preferenceFilter: (userPreference: IUserPreferenceSettings) => {},
```
ainouzgali, line 14: I think you need to add it to component web, vue and angular components

libs/shared/src/types/notification-templates/index.ts
```diff
@@ -0,0 +1,3 @@
+import { CustomDataType } from '../shared';
+
+export type NotificationTemplateCustomData = CustomDataType;
```
ainouzgali, line 3:
```suggestion
export type WorkflowCustomData = CustomDataType;
```
To help with our renaming

libs/shared/src/dto/notification-templates/create-template.dto.ts
ainouzgali, line 1: I think it should be added in create/update-workflow.dto as well. Until we deprecate this one completely..

ainouzgali:
@djabarovgeorge let's add it to the node sdk [here](https://github.com/novuhq/novu/blob/e03a76cc30f52f2cc52f9be1eef055337015c0c7/packages/node/src/lib/notification-template/notification-template.interface.ts) as well, so we are consistent with the types.

libs/shared/src/types/notification-templates/index.ts
djabarovgeorge: i had it in a lot of places but was not sure what is our policy regarding new items.

ainouzgali: If it's new, I feel we should just create with the new naming. Wdyt?

libs/shared/src/dto/notification-templates/create-template.dto.ts
djabarovgeorge: the funny thing is that update-workflow.dto is not used 😅
will fix it.

libs/shared/src/types/notification-templates/index.ts
djabarovgeorge: thanks for the comment i thought it was autogenerated, i will look into it.

packages/notification-center/src/store/notification-center.context.ts
djabarovgeorge: thanks for the comment i thought it was autogenerated, i will look into it.

packages/notification-center/src/components/notification-center/NotificationCenter.tsx
```diff
@@ -12,14 +14,15 @@ export interface INotificationCenterProps {
   onUrlChange?: (url: string) => void;
   onNotificationClick?: (notification: IMessage) => void;
   onUnseenCountChanged?: (unseenCount: number) => void;
+  onActionClick?: (templateIdentifier: string, type: ButtonTypeEnum, message: IMessage) => void;
+  actionsResultBlock?: (templateIdentifier: string, messageAction: IMessageAction) => JSX.Element;
+  preferenceFilter?: (userPreference: IUserPreferenceSettings) => boolean;
```
LetItRock, line 19: please update also documentation

packages/notification-center/src/components/notification-center-web-component/notification-center-component.types.ts
```diff
@@ -25,6 +25,7 @@ interface PopoverAdditionalProps {
   unseenCountChanged?: IPopoverNotificationCenterProps['onUnseenCountChanged'];
   actionClicked?: IPopoverNotificationCenterProps['onActionClick'];
   tabClicked?: IPopoverNotificationCenterProps['onTabClick'];
+  preferenceFilter?: IPopoverNotificationCenterProps['preferenceFilter'];
```
ainouzgali, line 28: I feel this should be added to the pick in `SelectedPopoverProps`  (line 31) instead of here. As these are for the unique renaming of our functions

packages/notification-center/src/components/notification-center-web-component/notification-center-component.tsx
```diff
@@ -110,6 +111,7 @@ function PopoverWrapper({
   popoverConfig,
   unseenBadgeColor,
   unseenBadgeBackgroundColor,
+  preferenceFilter,
```
ainouzgali, line 114: This should also be passed to the NotificationCenterComponent props above (line 68)

ainouzgali (commented):
@djabarovgeorge there are still some missing additions inside the vue and angular packages. We can go over it if you'd like, or what I do is a search of tabClicked to find potential places ;)

packages/notification-center/src/components/notification-center-web-component/notification-center-component.types.ts
djabarovgeorge: looks like i missed this one thanks!

LetItRock:
if you need any help with testing this for angular, vue, web component let me know ;)

LetItRock (approved):
looks awesome! 🎉

PR #3899, opened 2023-08-02
Docs: Add Python quickstart

unicodeveloper:

### DIFF
--- a/docs/docs/overview/quickstart/get-started-with-python.md
+++ b/docs/docs/overview/quickstart/get-started-with-python.md
@@ -43,10 +43,9 @@ NovuConfig().configure("https://api.novu.co", "<YOUR_NOVU_API_KEY>")
 
 Replace the `<YOUR_NOVU_API_KEY>` value with the authentic key from the **API Key** section of your [Novu Dashboard](https://web.novu.co/settings).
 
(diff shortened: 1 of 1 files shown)


docs/docs/overview/quickstart/get-started-with-python.md
```diff
@@ -0,0 +1,278 @@
...
+
+# Define a subscriber instance
+subscriber = SubscriberDto(
+   subscriber_id=your_subscriber_id,
+   email="abc@gmail.com",
+   first_name="John",
+   last_name="Doe"
+)
+
+SubscriberApi().create(subscriber)
```
jainpawan21, line 129: I am not sure, it will take apiUrl from config or we need to put API URL like this
https://github.com/novuhq/novu-python#subscribers

unicodeveloper: It will. There are two ways to do it.

Set the apiUrl from config so it uses throughout all the API calls or put in the API url for each api method you call. I chose the former @jainpawan21

jainpawan21: Okay, Rest looks fine to me.

PR #3900, opened 2023-08-02
feat: make identifier labels consistent

davidsoderberg:
Change labels according to issue

Screenshot 2023-08-02 at 12 53 56
Screenshot 2023-08-02 at 12 54 02
Screenshot 2023-08-02 at 12 54 14
Screenshot 2023-08-02 at 12 55 58
Screenshot 2023-08-02 at 12 56 29
Screenshot 2023-08-02 at 12 56 42


ainouzgali (approved):
@davidsoderberg there is also the key: string in the integration list table, under the name of the integration. I believe there is also supposed to be a change :)

PR #3904, opened 2023-08-02
fix: add identifier in migration script for novu providers

davidsoderberg:

Add so an Indentifier is created in migration for Novu provider.

### DIFF
--- a/apps/api/migrations/novu-integrations/novu-integrations.migration.ts
+++ b/apps/api/migrations/novu-integrations/novu-integrations.migration.ts
@@ -9,6 +9,8 @@ import {
 import { EmailProviderIdEnum, SmsProviderIdEnum } from '@novu/shared';
 import { NestFactory } from '@nestjs/core';
 import { AppModule } from '../../src/app.module';
+import slugify from 'slugify';
+import * as shortid from 'shortid';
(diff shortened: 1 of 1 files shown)


apps/api/migrations/novu-integrations/novu-integrations.migration.ts
```diff
@@ -42,6 +44,7 @@ const createNovuIntegration = async (
     _organizationId: environment._organizationId,
     providerId,
     channel,
+    identifier: `${slugify(providerId, { lower: true, strict: true })}-${shortid.generate()}`,
```
djabarovgeorge, line 47: Good catch! is it the same way we do it on the client and api?

djabarovgeorge (approved):

apps/api/migrations/novu-integrations/novu-integrations.migration.ts
davidsoderberg: Not exactly we use the provider name that we have hard coded on the frontend maybe it should be here as well?

davidsoderberg: I added it :)

PR #3907, opened 2023-08-02
fix(app-generic): circular dependency on subscriber pref use cases

p-fernandez:

Fixes a circular dependency between 2 use case modules. Arranges also the location of the types.

### DIFF
--- a/packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
+++ b/packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
@@ -9,19 +9,22 @@ import {
 import {
   ChannelTypeEnum,
   IPreferenceChannels,
+  IPreferenceOverride,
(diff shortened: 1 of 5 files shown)


packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -156,21 +159,21 @@ export class GetSubscriberTemplatePreference {
 
 function updateOverrideReasons(
   channelName,
-  sourceName: string,
+  sourceName: PreferenceOverrideSourceEnum,
   index: number,
   overrideReasons: IPreferenceOverride[]
 ) {
-  const currOverride: IPreferenceOverride = {
+  const currentOverride: IPreferenceOverride = {
```
p-fernandez, line 166: So CSpell stops complaining about `curr`.

LetItRock (approved):
awesome 🎉

PR #3910, opened 2023-08-03
Nv 2588 multi provider multiple in app instances can be created

davidsoderberg:

Fix so not more then one in app provider can be created per environment

Screenshot 2023-08-03 at 08 40 55

### DIFF
--- a/apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
@@ -2,7 +2,7 @@ import { BadRequestException, ConflictException, Inject, Injectable } from '@nes
 import * as shortid from 'shortid';
 import slugify from 'slugify';
 import { IntegrationEntity, IntegrationRepository, DalException } from '@novu/dal';
-import { ChannelTypeEnum, providers } from '@novu/shared';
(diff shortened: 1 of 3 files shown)


apps/web/cypress/tests/integrations-list-page.spec.ts
```diff
@@ -291,63 +297,70 @@ describe('Integrations List Page', function () {
...
-    cy.get('@channelTabs').contains('Chat');
-    cy.get('@channelTabs').contains('Push');
-    cy.get('@channelTabs').contains('SMS');
-
-    cy.getByTestId('providers-group-in_app').contains('In-App').as('inAppGroup');
-    cy.getByTestId('providers-group-email').contains('Email').as('emailGroup');
-    cy.getByTestId('providers-group-chat').contains('Chat').as('chatGroup');
-    cy.getByTestId('providers-group-push').contains('Push').as('pushGroup');
-    cy.getByTestId('providers-group-sms').contains('SMS').as('smsGroup');
+    cy.task('deleteProvider', {
```
davidsoderberg, line 300: Need to delete in app provider here because we create both on init session so for that option to be visibile we need to delete them here.

LetItRock (approved):
awesome work 🎉

PR #3912, opened 2023-08-03
fix: fix update of digest jobs only for current subscriber

ainouzgali:

When concurrently triggering a workflow with a digest + digest key for multiple subscribers in same trigger, some events contained did not belong to the correct digest key. 

When getting the jobs to update the digest field with the events we save on job, we didn't filter it with the subscriberId (🤦‍♀️ ). Meaning that when we trigger for multiple subscribers ( they have the same transactionId ) we update all the "next jobs" - all the not completed jobs of all the subscribers with the same transaction -  with the digest: {events }. And by that we are overriding the other settings we had in that digest object, like the digest key. So, when later, in the execution of that job we accidentally updated he doesn't have the digestKey and so he aggregates all the events. As if there was no digestKey.

### DIFF
--- a/apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
@@ -17,6 +17,8 @@ import {
   TriggerRecipients,
   TriggerRecipientsTypeEnum,
   ExternalSubscriberId,
+  DigestUnitEnum,
+  DigestTypeEnum,
 } from '@novu/shared';
 import { SubscribersService, UserSession } from '@novu/testing';
 import axios from 'axios';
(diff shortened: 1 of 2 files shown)


scopsy:
@ainouzgali is it possible to add another e2e test for this?

p-fernandez:
> @ainouzgali is it possible to add another e2e test for this?

@scopsy test suites for digest has been heavily skipped and neglected by the team during all these months because they weren't passing when moving pieces around. So it is a bit unlikely that @ainouzgali can add a test safely without stepping inside on something in this potential minefield.
For this time I think it is sensible to suggest to merge the fix and try it manually and time forgiving try to add a test later. 👍🏻

scopsy:
@p-fernandez I was able to successfully run the concurrent tests with modification to have multiple subscribers yesterday. So I don't see why it wouldn't be possible to add that test now. @ainouzgali if you see it doesn't work for some reason let me know

p-fernandez:
> @p-fernandez I was able to successfully run the concurrent tests with modification to have multiple subscribers yesterday. So I don't see why it wouldn't be possible to add that test now. @ainouzgali if you see it doesn't work for some reason let me know

Then please unskip them so we have them available. 🙏🏻

scopsy (approved):

PR #3913, opened 2023-08-03
feat: render html in in app preview

davidsoderberg:

Render html in in app preview.


LetItRock (approved):
⭐

PR #3914, opened 2023-08-03
feat: Subscribers list UI

BiswaViraj:

- Updated the date format for created at

image

### DIFF
--- a/apps/web/src/design-system/table/Table.tsx
+++ b/apps/web/src/design-system/table/Table.tsx
@@ -183,7 +183,8 @@ export function Table<T extends object>({
       {!loading && pagination && pageSize > 1 && pagination?.minimalPagination && (
         <div
           style={{
-            marginTop: '10px',
+            marginTop: 'auto',
+            marginBottom: '10px',
             display: 'flex',
             width: '100%',
             justifyContent: 'center',
--- a/apps/web/src/design-system/table/Table.tsx
+++ b/apps/web/src/design-system/table/Table.tsx
@@ -183,8 +183,6 @@ export function Table<T extends object>({
       {!loading && pagination && pageSize > 1 && pagination?.minimalPagination && (
         <div
           style={{
-            marginTop: 'auto',
-            marginBottom: '10px',
             display: 'flex',
             width: '100%',
             justifyContent: 'center',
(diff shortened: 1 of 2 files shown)


apps/web/src/design-system/table/Table.tsx
```diff
@@ -116,7 +116,7 @@ export function Table<T extends object>({
   const noData = rows.length === 0;
 
   return (
-    <div style={{ position: 'relative', minHeight: 500, display: 'flex', flexDirection: 'column' }}>
+    <div style={{ position: 'relative', minHeight: 500, display: 'flex', flexDirection: 'column', height: '85%' }}>
```
BiswaViraj, line 119: added height: 85% for now to show the pagination at the bottom of screen.
once we implement infinite scroll we might need to adjust this value

apps/web/src/design-system/table/Table.tsx
```diff
@@ -183,7 +183,8 @@ export function Table<T extends object>({
       {!loading && pagination && pageSize > 1 && pagination?.minimalPagination && (
         <div
           style={{
-            marginTop: '10px',
+            marginTop: 'auto',
+            marginBottom: '10px',
```
LetItRock, line 187: maybe let's not tackle this problem right now, because it doesn't look well on the bigger screens, we would need to think about that with Nikolay

apps/web/src/pages/subscribers/SubscribersListPage.tsx
```diff
@@ -34,7 +34,7 @@ const columns: IExtendedColumn<ISubscriber>[] = [
   {
     accessor: 'createdAt',
     Header: 'Created At',
-    Cell: withCellLoading(({ row: { original } }) => format(new Date(original.createdAt), 'dd/MM/yyyy HH:mm')),
+    Cell: withCellLoading(({ row: { original } }) => format(new Date(original.createdAt), 'dd.MM.yyyy')),
```
LetItRock, line 37: not sure why you also changed this formatting, we do use everywhere with "/" as a divider

BiswaViraj: It was in the designs
will revert them

apps/web/src/design-system/table/Table.tsx
BiswaViraj: Will remove this for now

apps/web/src/pages/subscribers/SubscribersListPage.tsx
BiswaViraj: done

apps/web/src/design-system/table/Table.tsx
BiswaViraj: done

LetItRock (approved):

PR #3921, opened 2023-08-03
feat(ws): create unit test suite

p-fernandez:

Creates the unit test suite for WebSocket app.

For confidence in the migration of BullMqService to use InMemoryProviderService I need to be able to test the changes in all the apps that use any functionality of BullMq.


.github/workflows/reusable-ws-e2e.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install wait-on plugin
+        run: npm i -g wait-on
+
+        # Runs a single command using the runners shell
+      - name: Build WS
+        run: CI='' pnpm build:ws
+
+      - name: Run unit tests
+        env:
+          IN_MEMORY_CLUSTER_MODE_ENABLED: false
```
p-fernandez, line 50: Not needed to enable for now as this environment variable is not yet used in WS.

apps/ws/src/socket/socket.module.ts
```diff
@@ -11,7 +11,7 @@ import { SharedModule } from '../shared/shared.module';
   exports: [WSGateway],
 })
 export class SocketModule implements OnModuleInit {
-  private readonly bullMqService: BullMqService;
+  public readonly bullMqService: BullMqService;
```
p-fernandez, line 14: Is `readonly` and I am looking to validate the right instantiation and values of the connection to Redis for this app. Therefore the change of visibility.

libs/testing/src/test-server.service.ts
```diff
@@ -19,3 +19,27 @@ export class TestServer {
 }
 
 export const testServer = new TestServer();
+
+export class WsTestServer {
```
p-fernandez, line 23: Separate the instantiation of the different test apps.

packages/application-generic/src/services/in-memory-provider/redis-provider.ts
```diff
@@ -37,17 +39,19 @@ export interface IRedisProviderConfig {
 
 export const getRedisProviderConfig = (): IRedisProviderConfig => {
   const redisConfig: IRedisConfig = {
-    host: process.env.REDIS_CACHE_SERVICE_HOST,
-    port: process.env.REDIS_CACHE_SERVICE_PORT || String(6379),
+    db: process.env.REDIS_DB_INDEX,
+    host: process.env.REDIS_HOST || 'localhost',
+    port: process.env.REDIS_PORT || String(6379),
```
p-fernandez, line 44: Everything started here when I noticed that I had set for the Redis instance provider for InMemoryProviderService the wrong settings when trying to hook up the BullMqService to InMemoryProviderService in the WS app in the MemoryDB task.
Fixing it here already.

BiswaViraj (approved):
🚀

PR #3935, opened 2023-08-07
feat: add sort by handlebar helper

jainpawan21:
1. Add sort by helper


packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
```diff
@@ -50,21 +50,6 @@ Handlebars.registerHelper(
...
-    return array
-      .map((item) => {
-        if (item[property]) {
-          return item[property];
-        }
-      })
-      .filter((value, index, self) => self.indexOf(value) === index);
-  }
-);
-
```
BiswaViraj, line 67: Was this supposed to be removed?

BiswaViraj (approved):

PR #3936, opened 2023-08-07
feat(tests): add tests for worker and queue services

p-fernandez:

Adds unit tests for worker and queue services dependant of BullMqService.

Give extra confidence in the changes for MemoryDB and identify possible migration mistakes.


LetItRock (approved):
hope that you run manually tests with `jest` ;)

PR #3938, opened 2023-08-08
feat(wip): initial bulk subscriber create

ainouzgali:

For larger organizations allow creating subscribers in bulk.
Introduce a subscriber bulk API to create subscriber in batches of up to 500 entries.


libs/dal/src/repositories/base-repository.ts
```diff
@@ -163,8 +163,8 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
     return await Promise.all(promises);
   }
 
-  async bulkWrite(bulkOperations: any) {
-    await this.MongooseModel.bulkWrite(bulkOperations);
+  async bulkWrite(bulkOperations: any, ordered = false): Promise<any> {
```
ainouzgali, line 166: Mongoose doesn't export the Bulk types. Not sure how we can still type this

libs/dal/src/repositories/subscriber/subscriber.repository.ts
```diff
@@ -32,6 +33,68 @@ export class SubscriberRepository extends BaseRepository<SubscriberDBModel, Subs
     );
   }
 
+  async bulkCreateSubscribers(subscribers, environmentId: EnvironmentId, organizationId: OrganizationId) {
```
ainouzgali, line 36: Wondering what type I should use for subscribers, as they are actually of the dto type defined in the api / the ISubscriberDefine in shared

libs/dal/src/repositories/subscriber/subscriber.repository.ts
```diff
@@ -32,6 +33,68 @@ export class SubscriberRepository extends BaseRepository<SubscriberDBModel, Subs
...
+    });
+
+    let result;
+    try {
+      const response = await this.bulkWrite(bulkWriteOps);
+      result = response.result;
+    } catch (e) {
+      if (!e.writeErrors) {
+        throw new DalException(e.message);
+      }
```
ainouzgali, line 56: If the error is not from write errors related to the bulk execution, throw the error. 
Otherwise, we want to parse that into the response and not throw error

libs/dal/src/repositories/subscriber/subscriber.repository.ts
```diff
@@ -32,6 +33,68 @@ export class SubscriberRepository extends BaseRepository<SubscriberDBModel, Subs
...
+        return {
+          message: error.err.errmsg,
+          subscriberId: error.err.op?.subscriberId,
+        };
+      });
+    }
+
+    const updatedSubscribers = subscribers
+      .filter((subId, index) => !indexes.includes(index))
+      .map((subscriber) => {
```
ainouzgali, line 83: updated subscribers are those that didn't have errors and were not inserted

libs/dal/src/repositories/base-repository.ts
p-fernandez: Let's check if it is a version thing and we need to upgrade the Mongoose version. Found some comments mentioning that with the same problem.

libs/dal/src/repositories/subscriber/subscriber.repository.ts
p-fernandez: I think we should use the same types. This is one of the parts that I wanted to reuse from the trigger event subscriber on-the-fly creation to have one single source of truth.

apps/api/src/app/subscribers/e2e/bulk-create-subscribers.e2e.ts
```diff
@@ -0,0 +1,149 @@
...
+    expect(updatedSubscriber?.firstName).to.equal('update');
+    expect(createdSubscriber?.firstName).to.equal('John');
+    expect(createdSubscriber?.email).to.equal('john@doe.com');
+    expect(createdSubscriber?.phone).to.equal('+972523333333');
+    expect(createdSubscriber?.locale).to.equal('en');
+    expect(createdSubscriber?.data?.test1).to.equal('test value1');
+  });
+
+  it('should throw an error when sending more than 500 subscribers', async function () {
+    const payload = {
```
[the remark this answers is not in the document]
ainouzgali: I can't think of a way to fake a failed update during the write executions... If the user is not providing enough information, it will be caught by the nested validator

libs/dal/src/repositories/base-repository.ts
LetItRock: I see that they don't fix that also in the latest :( super weird...

apps/api/src/app/subscribers/e2e/bulk-create-subscribers.e2e.ts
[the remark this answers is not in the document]
p-fernandez: Ok, so assuming the nested validator is tested already to solve the other failure case we could provide proper uploads and fake the failure in the DAL service, for example, with Sinon. So in:
```
      bulkResponse = await this.bulkWrite(bulkWriteOps);
 ```
we would need to fake the response so for example from 3 subscribers, 2 go well and 1 fails. Therefore we can test the rest of the logic for the wrong case.
Anyway, if you need to move on other tasks let's not spend too much time on it. 👍🏻

File: apps/api/src/app/events/usecases/map-trigger-recipients/map-trigger-recipients.use-case.ts [hunk 6]
```diff
@@ -27,7 +27,7 @@ interface ILogTopicSubscribersPayload {
 }
 
 const isNotTopic = (recipient: TriggerRecipientSubscriber): recipient is TriggerRecipientSubscriber =>
-  typeof recipient === 'string' || recipient?.type !== TriggerRecipientsTypeEnum.TOPIC;
+  typeof recipient === 'string' || !recipient.hasOwnProperty('type');
```
p-fernandez, line 30: What's the reason behind changing this? We might not be using the type for a different situation of a Topic, but we wanted to leave that way open for future solutions. I can totally see us allowing to trigger an event to all the subscribers of a TenantId (becoming `TriggerRecipientsTypeEnum.TENANT`) as related to the work we are doing on that side.

p-fernandez (approved):
🌟 
Brilliant work. The endpoint is nicely implemented and the API response is well thought.
Just one question regarding one change that I am not sure about it.

apps/api/src/app/events/usecases/map-trigger-recipients/map-trigger-recipients.use-case.ts
ainouzgali: @p-fernandez Because I removed the unused ['record' fields](https://github.com/novuhq/novu/pull/3938/files#diff-9d8bab12d2fa35e0af04328affef14329e2db0a4715728bf2b955b25cac6ff89L17) from subscriber type (it was never used or saved anywhere), so in the context of this function - which subscriber is of type `TriggerRecipientSubscriber` - type script no longer accepts that there could be type property for subscriber. It wasn't exactly right even before, as it accepted it by 'mistake' 

We could refactor the function to be of different type/ different check when we need to and get to a specific case. Wdyt? makes sense?

PR #3939, opened 2023-08-08
feat(api): notifications feed filtering by partial payload object

LetItRock:

Allow filtering of the notifications feed by the custom data from the payload object.

The two endpoints `GET /subscribers/:subscriberId/notifications/feed` and `GET /widgets/notifications/feed` now can take the query param like `?payload=eyJmb28iOjEyM30=`. 

The value is a base64 encoded string of the partial payload JSON object. 

For example, if the payload passed to the trigger event was like `{ foo: 123, bar: 'bar' }`, then you filter notifications by `{ foo: 123 }` by doing `btoa(JSON.stringify({ foo: 123 }))` that results in base64 encoded string `eyJmb28iOjEyM30=` and passing it to the query param.

Allows building custom notification center experiences.

Screenshot 2023-08-08 at 18 33 01

Screenshot 2023-08-08 at 18 33 11

### DIFF
--- a/apps/api/src/app/subscribers/dtos/get-in-app-notification-feed-for-subscriber.dto.ts
+++ b/apps/api/src/app/subscribers/dtos/get-in-app-notification-feed-for-subscriber.dto.ts
@@ -26,4 +26,12 @@ export class GetInAppNotificationsFeedForSubscriberDto extends PaginationRequest
 
   @ApiPropertyOptional({ required: false, type: Boolean })
   seen: boolean;
+
+  @ApiPropertyOptional({
+    required: false,
+    type: 'string',
+    description: 'Base64 encoded string of the partial payload JSON object',
+    example: 'btoa(JSON.stringify({ foo: 123 })) results in base64 encoded string like eyJmb28iOjEyM30=',
+  })
+  payload?: string;
 }
--- a/apps/api/src/app/subscribers/e2e/get-notifications-feed.e2e.ts
+++ b/apps/api/src/app/subscribers/e2e/get-notifications-feed.e2e.ts
@@ -38,6 +38,53 @@ describe('Get Notifications feed - /:subscriberId/notifications/feed (GET)', fun
       );
     }
   });
+
+  it('should throw exception when invalid payload query param is passed', async function () {
+    await session.triggerEvent(template.triggers[0].identifier, subscriberId);
+
+    await session.awaitRunningJobs(template._id);
+
+    try {
+      await getNotificationsFeed(subscriberId, session.apiKey, { limit: 5, payload: 'invalid' });
+    } catch (err) {
+      expect(err.response.status).to.equals(400);
+      expect(err.response.data.message).to.eq(`Invalid payload, the JSON object should be encoded to base64 string.`);
+    }
+  });
+
+  it('should allow filtering by custom data from the payload', async function () {
+    const partialPayload = { foo: 123 };
+    const payload = { ...partialPayload, bar: 'bar' };
+
+    await session.triggerEvent(template.triggers[0].identifier, subscriberId);
+    await session.awaitRunningJobs(template._id);
+
+    await session.triggerEvent(template.triggers[0].identifier, subscriberId, payload);
+    await session.awaitRunningJobs(template._id);
+
+    const payloadQueryValue = Buffer.from(JSON.stringify(partialPayload)).toString('base64');
+    const { data } = await getNotificationsFeed(subscriberId, session.apiKey, { limit: 5, payload: payloadQueryValue });
+
+    expect(data.length).to.equal(1);
+    expect(data[0].payload).to.deep.equal(payload);
+  });
+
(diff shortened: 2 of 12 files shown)


apps/api/src/app/subscribers/dtos/get-in-app-notification-feed-for-subscriber.dto.ts
```diff
@@ -26,4 +26,12 @@ export class GetInAppNotificationsFeedForSubscriberDto extends PaginationRequest
...
   @ApiPropertyOptional({ required: false, type: Boolean })
   seen: boolean;
+
+  @ApiPropertyOptional({
+    required: false,
+    type: 'string',
+    description: 'Base64 encoded string of the partial payload JSON object',
+    example: 'btoa(JSON.stringify({ foo: 123 })) results in base64 encoded string like eyJmb28iOjEyM30=',
+  })
+  payload?: string;
```
LetItRock, line 36: The two feed endpoints now do take the payload query param which is the base64 encoded string of the partial payload JSON object. Check the description for more info.

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -21,6 +21,18 @@ export class GetNotificationsFeed {
...
+    if (!payload) {
+      return;
+    }
+
+    try {
+      return JSON.parse(Buffer.from(payload, 'base64').toString());
+    } catch (e) {
+      throw new BadRequestException('Invalid payload, the JSON object should be encoded to base64 string.');
+    }
+  }
```
LetItRock, line 34: convert the base64 string payload to the JS object

apps/api/src/app/widgets/usecases/get-notifications-feed/get-notifications-feed.usecase.ts
```diff
@@ -47,7 +61,7 @@ export class GetNotificationsFeed {
       command.environmentId,
       subscriber._id,
       ChannelTypeEnum.IN_APP,
-      { feedId: command.feedId, seen: command.query.seen, read: command.query.read },
+      { feedId: command.feedId, seen: command.query.seen, read: command.query.read, payload },
```
LetItRock, line 64: pass the payload to the repository

libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -62,14 +71,21 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
       requestQuery.read = { $in: [true, false] };
     }
 
+    if (query.payload) {
+      requestQuery = {
+        ...requestQuery,
+        ...getFlatObject({ payload: query.payload }),
+      };
+    }
```
LetItRock, line 79: If the payload object is passed, then prepare the flat object out of it, like if the payload is `{ foo: { bar: 123 } }`, then the flat object would be:
```
{
"payload.foo.bar": 123
}
```
Which is passed then to the find query.

libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -62,14 +71,21 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
       requestQuery.read = { $in: [true, false] };
     }
 
+    if (query.payload) {
+      requestQuery = {
+        ...requestQuery,
+        ...getFlatObject({ payload: query.payload }),
```
ainouzgali, line 77: what if `query.payload` is an empty object, `{ }`. Are we still covered with errors/will not throw errors? Also in the conversion of the payload in other places ?

ainouzgali (approved):
@LetItRock really awesome work ! 
As I mentioned before, I feel it will have a lot of impact to add it to the node sdk and headless. 
Also, probably not now, but it will be nice to add a function that does the encoding for the user, at least for the node sdk

libs/dal/src/repositories/message/message.repository.ts
LetItRock: if the `query.payload` is `{}` then the `getFlatObject({ payload: {} })` will return an empty object `{}` ;) 
So basically it won't return any keys flattened if the value is an empty object, including nested objects.

apps/api/src/app/subscribers/e2e/get-notifications-feed.e2e.ts
```diff
@@ -38,6 +38,53 @@ describe('Get Notifications feed - /:subscriberId/notifications/feed (GET)', fun
...
     }
   });
+
+  it('should throw exception when invalid payload query param is passed', async function () {
+    await session.triggerEvent(template.triggers[0].identifier, subscriberId);
+
+    await session.awaitRunningJobs(template._id);
+
+    try {
+      await getNotificationsFeed(subscriberId, session.apiKey, { limit: 5, payload: 'invalid' });
```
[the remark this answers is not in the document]
LetItRock: thanks, missed that :)

PR #3916, opened 2023-08-03
Nv 2652 implement inmemoryproviderservice into bullmqservice

p-fernandez:
### What change does this PR introduce?

Makes the changes needed for BullMqService to depend on the InMemoryProviderService. This is being tested agains a single Redis instance as our self hosted users will use Novu. Next steps is validate this implementation agains Redis CLuster and then implement MemoryDB.

### Why was this change needed?

Mid step for the MemoryDB implementation.
I had to rewrite the Queues and Workers due huge limitations of our codebase regarding circular dependencies in several parts of our code that lead to weird Dependency Injections using the old way. So the InMemoryProviderService has been encapsulated inside of BullMqService (was a particular preference in the past also this way). This lead me to rewrite the queues and workers separating both concerns as previously they were mixed and highly confused in the previous implementation. On my opinion harder to follow up so I tried to bring my vision for improvements in the readability and cognitive load going through those bits.
As an extra tried to clean up the dependencies in the modules and head to a newer approach of creating modules in the Application Generic for easier dependency injection of related functionalities that belong to a certain feature.

PR #3940, opened 2023-08-09
feat(app-generic): revome queue and worker hardcoded names

p-fernandez:

Move queue and worker names to enums in Shared. Also observability keys.

Reducing the size of MemoryDB PR while fixing it.


apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
```diff
@@ -1,27 +1,32 @@
...
+import { QueueNameEnum, WorkerNameEnum } from '@novu/shared';
 import { Injectable } from '@nestjs/common';
 
 import { InboundEmailParse } from '../usecases/inbound-email-parse/inbound-email-parse.usecase';
 import { InboundEmailParseCommand } from '../usecases/inbound-email-parse/inbound-email-parse.command';
 
 @Injectable()
 export class InboundParseQueueService {
-  readonly QUEUE_NAME = 'inbound-parse-mail';
+  readonly name = QueueNameEnum.INBOUND_PARSE_MAIL;
```
p-fernandez, line 11: We tend to follow this naming for the QueueServices in general.

libs/shared/src/config/trigger-workflow.ts
```diff
@@ -0,0 +1,20 @@
+export enum QueueNameEnum {
+  INBOUND_PARSE_MAIL = 'inbound-parse-mail',
+  METRICS = 'metric',
+  STANDARD = 'standard',
+  WEB_SOCKETS = 'ws_socket_queue',
+  WORKFLOW = 'trigger-handler',
```
p-fernandez, line 6: AS the trigger-handler lives inside the Workflow module made the decision for this key.

libs/shared/src/config/trigger-workflow.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export enum WorkerNameEnum {
+  INBOUND_PARSE_MAIL = 'inbound-parse-mail',
+  METRICS = 'metric',
+  STANDARD = 'standard',
+  WEB_SOCKETS = 'ws_socket_queue',
+  WORKFLOW = 'trigger-handler',
+}
```
p-fernandez, line 15: _Duplicating_ them as I want to split in the future the Queues and Workers to not be mixed up like it happens in some classes that extend from Queue services and then instantiate Workers.

LetItRock: I'm not sure about having the separate enums, because the worker takes the queue name when you create it.
Now I would have to remember to call the queue using one enum and to call workers using another enum that should have the exact same value. Why complicate our lives, what is the benefit?

libs/shared/src/config/trigger-workflow.ts
```diff
@@ -0,0 +1,20 @@
+export enum QueueNameEnum {
```
LetItRock, line 1: maybe it's better to define these in the application generic?

libs/shared/src/config/trigger-workflow.ts
p-fernandez: Separation of concerns is the benefit. It is an extra help to understand what's going on when using it.
If you are instantiating a worker using for example a reference to a queue name might mislead you to think that's a queue, while they are different things. For me is one of the problems we have had in our current codebase due to naming and mixing concerns and want to slowly fix it up.

libs/shared/src/config/trigger-workflow.ts
p-fernandez: That would generate a circular dependency between `@novu/testing` and `@novu/application-generic`.

libs/shared/src/config/trigger-workflow.ts
LetItRock: You know that they are "coupled" by the queue name. But if the reference name is a problem here and misleading let's maybe not use the enums but instead just the constants?

LetItRock (approved):

libs/shared/src/config/trigger-workflow.ts
p-fernandez: As I finally separated the Queues and Workers in the new implementation here: https://github.com/novuhq/novu/pull/3916
having the naming separated to distinguish doesn't bother me anymore and I will use a single reference. Just will rename it as `JobTopicNameEnum` as basically is what it refers, the name of the topic where the jobs are published and then consumed from. Topic is the name used across the terminology of majority of the message brokers. 👍🏻

PR #3941, opened 2023-08-09
feat(app-generic): add test for add job use case to check the injected queue service

p-fernandez:

Adds unit test for AddJob use case.

To validate the QueueService injection in the MemoryDB PR for validation.

### DIFF
--- a/packages/application-generic/src/usecases/add-job/add-job.usecase.spec.ts
+++ b/packages/application-generic/src/usecases/add-job/add-job.usecase.spec.ts
@@ -1,0 +1,76 @@
+import { Test } from '@nestjs/testing';
+import { ExecutionDetailsRepository, JobRepository } from '@novu/dal';
+
+import { AddJob, AddDigestJob, AddDelayJob } from './index';
+
+import { CreateExecutionDetails } from '../create-execution-details';
+
+import {
+  CalculateDelayService,
+  DistributedLockService,
+  EventsDistributedLockService,
+  InMemoryProviderService,
+  QueueService,
+} from '../../services';
+
+let inMemoryProviderService: InMemoryProviderService;
+
+describe('AddJob use case', function () {
+  let useCase: AddJob;
+
+  beforeEach(async () => {
+    const moduleRef = await Test.createTestingModule({
+      imports: [],
+      providers: [
+        {
+          provide: InMemoryProviderService,
+          useFactory: () => {
+            return new InMemoryProviderService();
+          },
+        },
+        CalculateDelayService,
+        DistributedLockService,
+        EventsDistributedLockService,
+        ExecutionDetailsRepository,
+        JobRepository,
+        CreateExecutionDetails,
+        AddDigestJob,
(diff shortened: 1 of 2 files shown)


packages/application-generic/src/usecases/add-job/add-job.usecase.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { Test } from '@nestjs/testing';
+import { ExecutionDetailsRepository, JobRepository } from '@novu/dal';
+import { UserSession } from '@novu/testing';
```
LetItRock, line 3: there are some unused imports and variables ;)

packages/application-generic/src/usecases/add-job/add-job.usecase.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    expect(queueService.name).toEqual('standard');
+    const { bullMqService } = queueService;
+    expect(bullMqService).toBeDefined();
+    expect(await bullMqService.getRunningStatus()).toEqual({
+      queueIsPaused: false,
+      queueName: 'standard',
+      workerIsRunning: undefined,
+      workerName: undefined,
+    });
+  });
```
LetItRock, line 76: You've created the test for the AddJob use-case but just checking if the queue is instantiated properly? 🤔 
is something missing here?

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -36,6 +36,10 @@ export class AddJob {
     private addDelayJob: AddDelayJob
   ) {}
 
+  public getQueueService(): QueueService {
+    return this.queueService;
+  }
```
LetItRock, line 41: you don't need to expose it this way, in the test you can do `(useCase as any).queueService`

packages/application-generic/src/usecases/add-job/add-job.usecase.spec.ts
p-fernandez: I am trying to validate that the queueName used by the BullMqService is the right one for the intentions of AddJob. I am facing problems with the DI on the MemoryDB PR and I am going back and forth from `next` to that PR to double check I am doing the changes properly.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
p-fernandez: Changed the visibility to public readonly as I consider bad advise with hacking the type away as I mentioned in previous PRs.

packages/application-generic/src/usecases/add-job/add-job.usecase.spec.ts
LetItRock: If you are facing problems with DI typically it's because few provider instances have the same interface. If that's still a problem maybe you should use the "named" injections.

LetItRock (approved):

PR #3942, opened 2023-08-09
NV-2665 - [Discovery] - check if we can use the connection pool in our redis queries.

djabarovgeorge:

In this ticket, we need to do a discovery of the connection pool on redis and test if the performance is better.

### DIFF
--- a/packages/application-generic/src/services/cache/cache.service.ts
+++ b/packages/application-generic/src/services/cache/cache.service.ts
@@ -1,4 +1,5 @@
 import { Logger } from '@nestjs/common';
+import { createPool, Pool } from 'generic-pool';
 
 import { QUERY_PREFIX } from './key-builders';
 import {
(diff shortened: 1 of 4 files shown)


packages/application-generic/src/services/cache/cache.service.ts
```diff
@@ -152,37 +241,56 @@ export class CacheService implements ICacheService {
...
-        stream.on('end', () => {
-          resolve(undefined);
-        });
-        stream.on('error', (err) => {
-          reject(err);
+  public async delByPattern(pattern: string) {
+    return this.usingClient(async (client) => {
+      if (client) {
+        return new Promise((resolve, reject) => {
+          const stream = this.inMemoryProviderService.inMemoryScan(pattern);
```
djabarovgeorge, line 248: Not sure about this part, but as i remember we don't/should not use delByPattern in the code anymore so maybe we can remove this method altogether.

PR #3944, opened 2023-08-10
feat(worker): remove performance svc as it was temporary

p-fernandez:

Remove the skeleton of the performance service.

It was introduced for some quick tests to help one of our users. We rely in New Relic for metrics and observability so this is not needed anymore. Cleaning time.


LetItRock (approved):

LetItRock:
🧹

PR #3951, opened 2023-08-11
feat(infra): implement memory db and dual reads from old database

p-fernandez:

Introduces a worker for the old instance of Redis so we can keep reading it until consumption.

For a smooth migration from the Redis instance to MemoryDB.


LetItRock (approved):
looks good to me :)

Issue #3854, opened 2023-07-25
[NV-2610] 🏗️ Provide the ability to rearrange steps in a workflow


scopsy:
**What?**

WF editor steps order

**Why?**

Currently, users won't be able to rearrange step orders, so if they make a mistake and put a step into the wrong order, they may just create a new step in the desired place of a workflow. But he needs to fill in all the metadata for that step all over again. We assume that UX where UI prompts users to do double work in case they made a mistake with a steps order might be confusing.

**How?**

Allow users to reorder steps by **drag and drop.**

**MECHANICS**

* User hover over a step
* UI display D&D hover icon near a step
* User clicks on the icon and drags it Up or Down
* When the step becomes closer than distance 8px to its neighbor, they switch places.
* The dragged step takes the place of its neighbor, the neighbor takes the place of a step that was dragged

Example

![](https://uploads.linear.app/f13f1996-c9b0-4fea-8ee7-2c3faf6a832d/5c2e111d-3b18-4e49-a81e-49418b9409c0/a7ceebbc-23a1-4008-94c3-69418aaca601?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiL2YxM2YxOTk2LWM5YjAtNGZlYS04ZWU3LTJjM2ZhZjZhODMyZC81YzJlMTExZC0zYjE4LTRlNDktYTgxZS00OTQxOGI5NDA5YzAvYTdjZWViYmMtMjNhMS00MDA4LTk0YzMtNjk0MThhYWNhNjAxIiwiaWF0IjoxNzAzMTIxMDc1LCJleHAiOjE3MDMyMDc0NzV9.gClE7c9e4wyKb6G4Ga_7wZtS_W3I5OWqS2--aFqYiVI)

[Re-order steps - Workflow create/edit (Figma)](https://www.figma.com/file/72BWZJMFf9aikgI7CPpohp/Workflow-create%2Fedit?type=design&node-id=274-53699&mode=design&t=G2ykbLhUeNYYGvOS-4)

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-2610](https://linear.app/novu/issue/NV-2610/🏗️-provide-the-ability-to-rearrange-steps-in-a-workflow)</sub>

feliciien:
can I work on this?

scopsy:
[...]

PR #3952, opened 2023-08-11
NV-2610 - [WF editor] Provide ability to rearrange steps

gitstart:
This PR enhances the user experience in the workflow area, allowing users to easily update the position of individual steps. Previously, users had to remove all steps below a wrongly placed step to reposition it. With this update, users can now directly drag and move a step to a different position, making the process more intuitive and efficient.

The following improvements were implemented to achieve a seamless repositioning experience:

- When a user starts dragging a step, it is detached from its parent and child nodes, ensuring that only the selected step is moved without affecting its descendants.
- To maintain clarity and prevent confusion, the descendants of the repositioned step remain at their original positions during the drag operation.
- The step position is now correctly updated in the form state, reflecting the changes made by the user during the repositioning.
- A special styling is applied to the step being repositioned, providing a visual cue that it can be placed in a new position. 

To allow users to change the position of steps without having to remove all steps down to the wrongly placed step.

### Other information

Closes https://github.com/novuhq/novu/issues/3854

Loom Video https://www.loom.com/share/ba3167669ba545bd996181466f22cd67

---


gitstart:

gitstart:
@scopsy This PR is ready for review

LetItRock (changes requested):
hey @gitstart 👋 first, thank you for you contribution! 🤝  

I've found a few things that we have to improve in this PR:

1. When hovering over the node we should show the "handle" and not the "mouse drag hand", the designs are trying explaining the drag mouse behaviour, but we don't need to show the "mouse drag hand".
Screenshot 2023-08-18 at 14 26 49

2. When clicking on the handle it should not take to the editor.
https://github.com/novuhq/novu/assets/2607232/b4dad738-1e6f-4518-bf1d-69e3d1534615

3. When the dragged node is moved around I see some flickering.

https://github.com/novuhq/novu/assets/2607232/8654b413-41f4-42fb-88da-b5e7711ff733

4. When in dragged mode no error popovers, node actions (delete) should be shown.

https://github.com/novuhq/novu/assets/2607232/e14c872f-639f-4c88-b107-e67d63bd404f

5. When the distance is too small it doesn't swap nodes.

https://github.com/novuhq/novu/assets/2607232/4db4249e-a4f3-4a74-a8fe-723e8a568c9a

I feel like if we change the approach a little we might be able to fix the flickering issue and improve the UI/UX here. My suggestion is like: when the element is dragged we shouldn't move the real node, but instead create it's copy that will sit all above the others (like we do with the actions in the sidebar on the right) and on the drop do the swapping.
So this way user will see the real position of dragged element and we hopefully won't any have flickering because the real node is not dragged.

Let me know your thoughts. Thanks a lot! 🙌

gitstart:
Thanks for the feedback @LetItRock, we have applied some fixes that should resolve the issues you mentioned above.

LetItRock:
hey @gitstart 👋 
Thanks, it looks much much better, but there are still a few problems that I noticed. 
I really appreciate your work, but if we can fix them it would be really awesome 🙌 

1. The connecting lines disappear when clicking on the node handle.

https://github.com/novuhq/novu/assets/2607232/038d8f8f-404c-4162-86f5-f7b4a6250a2c

2. The swapping between the first and the last node doesn't work properly. The expectation is that the nodes in the middle won't change, otherwise, users would have to update even more nodes.

https://github.com/novuhq/novu/assets/2607232/6e9e67a4-0799-41d6-958b-6ca34e426d11

3. The tree behaves weirdly when dragging nodes. I couldn't explain it in words but you can notice that on the video, like when you are moving one node, the others start also to move, even if they are parents.

https://github.com/novuhq/novu/assets/2607232/be484a65-499b-4594-8ee7-098716655b21

4. After the node is dragged, I can't move the tree nodes anymore.

https://github.com/novuhq/novu/assets/2607232/08f2108b-46a7-4d2a-97ba-1faae88aee5b

Let me know if anything needs clarification! Thank you! 🤝

gitstart:
@LetItRock Thanks. We appreciate the detailed feedback. Although, before we push the next update to address your comments, we would like to get some more clarification about a few things. Our questions on the issues are in this [loom video](https://www.loom.com/share/Clarification-on-Distance-React-Flow-Intersection-and-Flickering-564e05f880bf415082645b381476ff1c)

PR #3961, opened 2023-08-13
feat: add override layout in trigger functionality

ainouzgali:

allow to override the layoutIdentifier with overrides property 

Note - I am still working on adding tests, I am not sure yet how to check in a test which layout was actually executed as we don't save the resulting content.


LetItRock (approved):
🥇

PR #3964, opened 2023-08-13
feat(notification-center-vue): Make Popover.triggers a prop for notification-center-vue

danidee10:

This PR makes the `PopoverConfig.triggers` a prop. This allows the developer to control the "triggers" that are passed to the `VDropdown` component

This change was introduced to fix an annoying issue that I had with the triggers for the Internal dropdown. The default values of `['click', 'touch']` didn't work well on Mobile. It kept flickering between the visible and hidden state because, on mobile. This happened because, on mobile, the dropdown responds to both events (touch and click). _It's fine on desktops because there's no "touch" event._
More Information can be found here: https://github.com/Akryum/floating-vue/issues/504

The Proposed solution there was to change the triggers. It's currently Impossible to do this at the moment with `@novu/notification-center-vue` which is an Option that this PR wants to add.

None. You can view the short recording in the linked Github issue from `floating-vue`. It shows **exactly** what I experienced on Mobile.

Thanks


LetItRock (commented):
hey @danidee10 👋 thank you a lot for your contribution 🎉 

It would be nice if we can mention about this prop in the docs: https://github.com/novuhq/novu/blob/2181e9469f53470da7c6904739e4380343eafc8e/docs/docs/notification-center/vue-component.md?plain=1#L99

However as you can see before this component accepted the exact same props as the Web Component and now the docs should be mentioning additionally "Vue" component props on that page.

I would be really thankful if you can do that 🙌

jainpawan21:
@danidee10 Thanks for adding this config, Could you please create an issue on our docs repo so that we can update documentation for this?

cc:- @unicodeveloper

danidee10:
Sure will do

PR #3966, opened 2023-08-14
Nv 2517 display that workflow has nodes that will fail to send a notification

BiswaViraj:

- get active integrations status returns an object which contains 

 `{channels : {channelType: boolean} }` which describes if there's any active integrations for the particular channelType
 along with and `isActive: boolean` field which denotes if all the steps of a single workflow have active integrations or not

image


apps/api/src/app/workflows/usecases/get-active-integrations-status/get-active-integrations-status.usecase.ts
```diff
@@ -0,0 +1,96 @@
...
+        ...workflows,
+        activeIntegrationStatus: {
+          channels: activeChannelsStatus,
+          isActive,
+        },
+      };
+    }
+  }
+
+  private handleSteps(steps: NotificationStep[], activeChannelsStatus: { [key in ChannelTypeEnum]: boolean }) {
```
BiswaViraj, line 78: loops through the steps and updates the `isActive` flag if the step has a channelType which doesn't have any active integrations.

apps/api/src/app/workflows/dto/workflow-response.dto.ts
```diff
@@ -115,4 +116,13 @@ export class WorkflowResponse {
   @ApiPropertyOptional()
   @IsOptional()
   data?: NotificationTemplateCustomData;
+  activeIntegrationStatus?: ActiveIntegrationsChannelsStatus;
+
+  @ApiProperty()
+  isBlueprint: boolean;
```
davidsoderberg, line 122: Should we really expose these in the swagger json? right now this is only used internally for use?

apps/api/src/app/workflows/usecases/get-active-integrations-status/get-active-integrations-status.usecase.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const activeIntegrations = await this.getActiveIntegrationUsecase.execute(
+      GetActiveIntegrationsCommand.create({
+        organizationId: command.organizationId,
+        environmentId: command.environmentId,
+        userId: command.userId,
+      })
+    );
+
+    const activeIntegrationsByEnv = activeIntegrations.filter(
```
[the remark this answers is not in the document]
davidsoderberg: Why do we need to do this when we pass environment id here: https://github.com/novuhq/novu/pull/3966/files#diff-dcd9b18e319e2a1e95ddca88446db6525e43dea143d7aaca024d6e563cbafc3fR26

apps/api/src/app/workflows/dto/workflow-response.dto.ts
BiswaViraj: Agree, we shouldn't expose it, let me remove it

apps/api/src/app/workflows/usecases/get-active-integrations-status/get-active-integrations-status.usecase.ts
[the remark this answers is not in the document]
BiswaViraj: @davidsoderberg
We do not use environmentId if the multiProvider is enabled
https://github.com/novuhq/novu/blob/f86886414ebb8853e9f83c11e06afacb3aa52864/packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts#L31

davidsoderberg (approved):

PR #3971, opened 2023-08-14
Add Analytics call to Track Deletion of Templates

Cliftonz:
## What change does this PR introduce?

This PR introduces the addition of analytics tracking for the deletion of templates.

## Why was this change needed?
Allows us to track the analytics on deleting notfication templates so that:

1. User Behavior Insight: By tracking the deletion of templates, we gain deeper insights into user behavior and understand which templates are being frequently deleted and why.

2. Enhancing User Experience: By understanding the patterns around template deletions, we can provide more tailored recommendations or warnings to users, ensuring a smoother and more intuitive user experience.


apps/api/src/app/workflows/usecases/delete-notification-template/delete-notification-template.usecase.ts
```diff
@@ -94,6 +96,19 @@ export class DeleteNotificationTemplate {
...
       );
+
+      this.analyticsService.track(`Removed Notification Template`, command.userId, {
+        _organization: command.organizationId,
+        _environment: command.environmentId,
+        _templateId: command.templateId,
+        data: {
+          createdAt: item.createdAt,
+          lastUpdatedAt: item.updatedAt,
+          deletedAt: new Date(),
```
scopsy, line 107: No need to pass this, as the time is already created by segment and downstream providers

scopsy (approved):

PR #3975, opened 2023-08-15
feat(ci): reuse instructions to tidy up pipelines

p-fernandez:

Tidy up the CI pipelines.

### DIFF
--- a/.github/actions/setup-project/action.yml
+++ b/.github/actions/setup-project/action.yml
@@ -60,3 +60,7 @@ runs:
     - name: 🛟 Install dependencies
       shell: bash
       run: pnpm install --frozen-lockfile
+
+    - name: Install wait-on plugin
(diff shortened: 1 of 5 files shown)


.github/actions/setup-project/action.yml
```diff
@@ -60,3 +60,7 @@ runs:
     - name: 🛟 Install dependencies
       shell: bash
       run: pnpm install --frozen-lockfile
+
+    - name: Install wait-on plugin
```
p-fernandez, line 64: `setup-project` action is being used by all the workflows. Makes sense to install the dependency here and forget. Even if `wait-on` is not used in all the workflows the installation of it should be quick and is available as part of the tooling for the workflows.

scopsy (approved):
Bravo 👌

PR #3980, opened 2023-08-16
patch : remove rows only when steps are truthy

djabarovgeorge:

deleted removed steps only if new steps were provided, or empty array.

so the client of the api could provide params for update without the steps.


BiswaViraj (approved):

ainouzgali:
@djabarovgeorge could you please just make sure it works as expected in the promotion?

djabarovgeorge:
> @djabarovgeorge could you please just make sure it works as expected in the promotion?

it is already addressed here :)
https://github.com/novuhq/novu/pull/3524#pullrequestreview-1448910996

PR #3981, opened 2023-08-16
feat(ci): separate the swagger validation in own runner plus improvement

p-fernandez:

Separates the Swagger validation in its own runner to avoid interfering in E2E test runners. Also isolates some reusable actions.

Clarify the intention of the Swagger validation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   dependency-review:
+    name: Dependency review
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout Repository'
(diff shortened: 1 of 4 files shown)


davidsoderberg (approved):

.github/workflows/test.yml
```diff
@@ -268,3 +249,25 @@ jobs:
         with:
           targets: lint,build,test
           projects: ${{matrix.projectName}}
+
+  validate_swagger:
+    name: Validate Swagger
+    runs-on: ubuntu-latest
+    needs: [get-affected]
+    if: ${{ fromJson(needs.get-affected.outputs.test-unit)[0] }}
```
p-fernandez, line 257: We would expect to validate the Swagger when something affecting the unit tests for the API change.

PR #3982, opened 2023-08-16
optimization : merge two redis calls in api key authenticate

djabarovgeorge:

merge two redis calls in api key to authenticate

in order to reduce the redis call in trigger high load

### DIFF
--- a/apps/api/src/app/auth/services/auth.service.ts
+++ b/apps/api/src/app/auth/services/auth.service.ts
@@ -9,6 +9,8 @@ import {
   SubscriberRepository,
   UserEntity,
   UserRepository,
+  EnvironmentEntity,
+  IApiKey,
 } from '@novu/dal';
 import { AuthProviderEnum, IJwtPayload, ISubscriberJwt, MemberRoleEnum, SignUpOriginEnum } from '@novu/shared';
 import {
@@ -19,6 +21,7 @@ import {
   buildEnvironmentByApiKey,
   buildSubscriberKey,
   buildUserKey,
+  buildUserDataKey,
 } from '@novu/application-generic';
 
 import { CreateUserCommand } from '../../user/usecases/create-user/create-user.dto';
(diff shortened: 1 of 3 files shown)


packages/application-generic/src/services/cache/key-builders/shared.ts
```diff
@@ -32,6 +32,7 @@ export enum CacheKeyPrefixEnum {
   INTEGRATION = 'integration',
   ENVIRONMENT_BY_API_KEY = 'environment_by_api_key',
   GROUPED_BLUEPRINTS = 'grouped-blueprints',
+  USER_DATA = 'user_data_env_entity_user_entity',
```
p-fernandez, line 35:
```suggestion
  AUTH_SERVICE = 'auth_service',
```
Mainly because it is not our User data entity and we try to return multiple stuff, all related to a certain service.

djabarovgeorge: awesome suggestion IMO thanks!

p-fernandez (approved):
🌟 
Perf improvements always welcome.

apps/api/src/app/auth/services/auth.service.ts
```diff
@@ -335,4 +337,31 @@ export class AuthService {
   private async getSubscriber({ subscriberId, _environmentId }: { subscriberId: string; _environmentId: string }) {
     return await this.subscriberRepository.findBySubscriberId(_environmentId, subscriberId);
   }
+
+  @CachedEntity({
```
scopsy, line 341: @djabarovgeorge what is our current default TTL?

djabarovgeorge: @scopsy I kept it as it is now (environment and user), it is 2 hours.

LetItRock: We should clear it when the user `email` has changed, like in this use-case `UpdateProfileEmail`. Otherwise, the signed token might have an old email.

LetItRock (approved):
🥇

Issue #3905, opened 2023-08-02
[NV-2473] data field is not added in case of inline of trigger


jainpawan21:
We can create a new subscriber on the fly using inline of trigger

[https://docs.novu.co/platform/subscribers#2-inline-of-trigger](https://docs.novu.co/platform/subscribers#2-inline-of-trigger)

All attributes except `data` is updated/added in this case

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-2473](https://linear.app/novu/issue/NV-2473/data-field-is-not-added-in-case-of-inline-of-trigger)</sub>

jainpawan21:
@scopsy 

What is the correct flow of syncing linear issue to github

gitstart:
@jainpawan21 we would like to work on this

PR #3987, opened 2023-08-16
NV-2473 - data field is not added in case of inline of trigger

gitstart:
This fix ensures data field is added in case of inline of trigger

#### Description
We can create a new subscriber on the fly using inline of trigger

[https://docs.novu.co/platform/subscribers#2-inline-of-trigger](https://docs.novu.co/platform/subscribers#2-inline-of-trigger)

All attributes except data is updated/added in this case

closes https://github.com/novuhq/novu/issues/3905

---


gitstart:
@jainpawan21 this PR is ready for review

packages/node/src/lib/events/events.ts
```diff
@@ -17,6 +17,7 @@ export class Events extends WithHttp implements IEvents {
       transactionId: data.transactionId,
       overrides: data.overrides || {},
       ...(data.actor && { actor: data.actor }),
+      data,
```
ainouzgali, line 20: @gitstart this is not the right place for it. The data of the subscriber will be already included inside the data.to. 
The change you need to make will be inside the process-subscriber usecase. I believe it is not passed to the create subscriber there.

ainouzgali (commented):
@gitstart could you also please add a test after you make the changes? :)

packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -48,6 +48,12 @@ export class Subscribers extends WithHttp implements ISubscribers {
     });
   }
 
+  async create(subscriberId: string, data: ISubscriberPayload) {
```
ainouzgali, line 51: @gitstart this is not needed. There is the identify function that does that. 
No changes are needed in the node sdk.

ainouzgali (commented):
@gitstart thanks. Adding a test for this should be done in this [test](https://github.com/novuhq/novu/blob/6de3cf8f0ab1434a203866f4fda49a69336c7331/apps/api/src/app/events/e2e/trigger-event.e2e.ts#L258)

Please remove all changes in node sdk, they are not needed :)

packages/node/src/lib/subscribers/subscribers.ts
gitstart: The function has been removed @ainouzgali, thanks for the feedback.

ainouzgali (approved):
Thanks @gitstart !

PR #3989, opened 2023-08-16
[POC] -  offload execution logs to queue

djabarovgeorge:

Offload execution logs to a queue. This PR is a POC, only to show how it could look like. In the first stage, we could use just the queue, this way we offload the DB query, and the response time of the trigger flow will be faster. 
Due to the reason that the execution details are not time sensitive and are meant only for debugging the possible latency is bearable.

Currently, we create a lot of execution detail requests in our trigger flow and they will probably increase in the future, which could cause a load in the DB.

Addition things that needed to be done + next steps:
* initialize Date time of the queued execution detail to the command and store it as createdAt. 
* refactor all of the execution detail requests to use the queue.
* refactor the worker to WorkerPro in order to get the execution details in batches and reduce the number of requests to the DB.  REF https://docs.bullmq.io/bullmq-pro/batches

### DIFF
--- a/apps/worker/src/app/workflow/services/execution-details-archive/execution-details-archive-consumer.service.ts
+++ b/apps/worker/src/app/workflow/services/execution-details-archive/execution-details-archive-consumer.service.ts
@@ -1,0 +1,38 @@
+import { Injectable, Logger } from '@nestjs/common';
+
+import { BullMqService, CreateExecutionDetails, CreateExecutionDetailsCommand } from '@novu/application-generic';
+import { JobTopicNameEnum } from '@novu/shared';
+
+@Injectable()
+export class ExecutionDetailsArchiveConsumer {
+  private readonly LOG_CONTEXT = 'ExecutionDetailsArchiveConsumerService';
+  private readonly name = JobTopicNameEnum.EXECUTION_DETAIL_ARCHIVE;
+
+  constructor(private createExecutionDetails: CreateExecutionDetails, private bullMqService: BullMqService) {
+    this.bullMqService.createWorker(this.name, this.getWorkerProcessor(), this.getWorkerOpts());
+  }
+
+  private getWorkerProcessor() {
+    return async ({ data }) => {
+      try {
+        await this.createExecutionDetails.execute(CreateExecutionDetailsCommand.create({ ...data }));
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        Logger.error(
+          'Unexpected exception occurred while execution details archive consumer services route ',
+          e,
+          this.LOG_CONTEXT
+        );
+
+        throw e;
+      }
+    };
+  }
+
+  private getWorkerOpts() {
+    return {
+      lockDuration: 90000,
+      concurrency: 5,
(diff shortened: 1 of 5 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -65,6 +67,19 @@ export class SendMessageInApp extends SendMessageBase {
 
   @InstrumentUsecase()
   public async execute(command: SendMessageCommand) {
+    await this.executionDetailsArchiveProducer.add(
```
djabarovgeorge, line 70: image

apps/worker/src/app/workflow/services/execution-details-archive/execution-details-archive-consumer.service.ts
djabarovgeorge, line 1: still not sure how we want to call out queue services, therefore added those tmp ones.

p-fernandez: Once this is done: https://github.com/novuhq/novu/pull/3866, half of the work you are doing will be simplified to the minimum.
We can even consider moving all the execution details jobs to a dedicated instance (chosen by provider in InMemoryProviderService).
Regarding naming we either keep consistency with [Queue|Worker]Service or we can think on moving to Publisher/Subscriber.

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
djabarovgeorge: definitely for now i created a minimalistic solution only for us to see how it could look.

apps/worker/src/app/workflow/services/execution-details-archive/execution-details-archive-consumer.service.ts
djabarovgeorge: yeah, the refactoring could help a lot simplify the flow.

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -65,6 +67,19 @@ export class SendMessageInApp extends SendMessageBase {
...
+      command.transactionId,
+      CreateExecutionDetailsCommand.create({
+        ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
```
LetItRock, line 73: Maybe in the future, we could define some property, let's call it for now `type`, and store only the minimalistic amount of data on the job, and then in the consumer based on the type we will do some logic, for ex. retrieve the job details. WDYT?

djabarovgeorge: I think it is a great idea and I believe we could add some strategies in order to build our execution details.

djabarovgeorge:
Will close this one because we won't implement this flow at the moment.

PR #3991, opened 2023-08-17
feat(ci): reusable action to run backend (api, worker) for clients

p-fernandez:

Creates an action to run API and Worker apps to be used in the reusable workflows of Web and Worker.

Tidying up the pipelines.

### DIFF
--- a/.github/workflows/reusable-web-e2e.yml
+++ b/.github/workflows/reusable-web-e2e.yml
@@ -68,6 +68,10 @@ jobs:
 
       - uses: ./.github/actions/start-localstack
 
+      - uses: ./.github/actions/run-backend
+        with:
+          launch_darkly_sdk_key: ${{ secrets.LAUNCH_DARKLY_SDK_KEY }}
+
       - name: Start Client
         working-directory: apps/web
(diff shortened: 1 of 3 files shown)


scopsy (approved):
💪

.github/workflows/reusable-web-e2e.yml
```diff
@@ -78,31 +82,11 @@ jobs:
         run: pnpm start:static:build &
 
       - name: Start WS
-        env:
-          IN_MEMORY_CLUSTER_MODE_ENABLED: false
```
p-fernandez, line 82: Not in use in WS.

.github/workflows/reusable-widget-e2e.yml
```diff
@@ -69,6 +69,10 @@ jobs:
           targets: build
           projects: '@novu/widget,@novu/embed,@novu/api,@novu/worker,@novu/ws'
 
+      - uses: ./.github/actions/run-backend
```
p-fernandez, line 72: Backend should be available before we try to connect the clients.

PR #3994, opened 2023-08-17
Nv 2403 tenant id trigger engine integration

ainouzgali:

Allow to override tenant identifier in the context of a trigger. 

I still have a lot of refactoring to do - opened this PR to allow discussion about this direction. 

When using tenant in an editor(will also have autocomplete for tenant):
Screen Shot 2023-08-17 at 16 35 53

Trigger snippet will recognize that a tenant object needs to be included in the trigger event:
Screen Shot 2023-08-21 at 13 32 37
Screen Shot 2023-08-21 at 13 32 46

The tenant object in the trigger will add the tenant to the context of this run. Also allows to create/update the tenant, similar to what we do for subscribers. 

Execution detail example:
Screen Shot 2023-08-21 at 13 33 09
Screen Shot 2023-08-21 at 13 33 22
Screen Shot 2023-08-21 at 13 33 46


libs/shared/src/entities/notification-template/notification-template.interface.ts
```diff
@@ -36,6 +36,17 @@ export interface INotificationTrigger {
   identifier: string;
   variables: INotificationTriggerVariable[];
   subscriberVariables?: INotificationTriggerVariable[];
+  snippetVariables?: ITriggerSnippetVariable[];
+}
+
+export enum TriggerContextTypeEnum {
+  TENANT = 'tenant',
+  ACTOR = 'actor',
+}
```
ainouzgali, line 45: I will remove the actor behviour from this PR, but just wanted to show the direction I intend to take here. It will make adding new sections of trigger objects easier and more immediate

libs/shared/src/entities/message-template/message-template.interface.ts
```diff
@@ -59,4 +60,17 @@ export const SystemVariablesWithTypes = {
...
+  },
+};
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const TriggerReservedVariables = ['tenant', 'actor'];
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const ReservedVariablesMap = {
+  [TriggerContextTypeEnum.TENANT]: [{ name: 'identifier', type: TemplateVariableTypeEnum.STRING }],
+  [TriggerContextTypeEnum.ACTOR]: [{ name: 'subscriberId', type: TemplateVariableTypeEnum.STRING }],
```
ainouzgali, line 75: The variables we need to be included for our 'reserved' variables to work in the context of a trigger.

libs/dal/src/repositories/job/job.schema.ts
```diff
@@ -116,6 +116,13 @@ const jobSchema = new Schema<JobDBModel>(
       type: Schema.Types.ObjectId,
       ref: 'Subscriber',
     },
+    _tenantId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Tenant',
+    },
+    tenantIdentifier: {
+      type: Schema.Types.String,
+    },
```
ainouzgali, line 125: I will remove the _tenantId and leave only the tenantIdentifier.
I feel it does make sense to have the tenantIdentifier as part of the job entity, also for future use when we incorporate more of the tenants logic in the future.

libs/shared/src/dto/events/event.interface.ts
```diff
@@ -7,3 +7,5 @@ export type TriggerRecipient = TriggerRecipientSubscriber | ITopic;
 export type TriggerRecipients = TriggerRecipient[];
 
 export type TriggerRecipientsPayload = TriggerRecipientSubscriber | TriggerRecipients;
+
+export type TriggerTenantContext = string | ITenantDefine;
```
ainouzgali, line 11: Name help? ;)

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
```diff
@@ -70,6 +73,34 @@ export class SendMessageSms extends SendMessageBase {
...
+              tenantIdentifier: tenantIdentifier,
+            }),
+          })
+        );
+
+        return;
+      }
+      await this.sendSelectedTenantExecution(command.job, tenant);
+    }
+
```
ainouzgali, line 103: I will extract this duplication and add to all the other channels

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -142,14 +154,43 @@ export class ParseEventRequest {
...
+        invalidKeys.push(`${reservedVariableType} object`);
+      } else {
+        const requiredVariablesContextNames = ReservedVariablesMap[reservedVariableType].map(
+          (variable) => variable.name
+        );
+        for (const variableName of requiredVariablesContextNames) {
+          const variableNameExists = typeof payload === 'string' ? payload : payload[variableName];
+          if (!variableNameExists) {
+            invalidKeys.push(`${variableName} property of ${reservedVariableType}`);
+          }
```
ainouzgali, line 184: On trigger, if missing the tenant object in trigger or missing the mandatory property 'identifier' will throw error

libs/dal/src/repositories/job/job.entity.ts
```diff
@@ -16,6 +16,7 @@ export class JobEntity {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   overrides: Record<string, Record<string, unknown>>;
   step: NotificationStepEntity;
+  tenant?: ITenantDefine;
```
ainouzgali, line 19: Added so we can show it in the snippet in activity feed -  so we can show the user a correct and accurate presentation of the snippet he used  + from here we take the identifier for the messages compilation

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -98,6 +107,28 @@ export class TriggerEvent {
       template
     );
 
+    if (tenant) {
```
davidsoderberg, line 110: Could this be a function?

packages/application-generic/src/usecases/process-tenant/process-tenant.usecase.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    try {
+      tenantEntity = await this.getTenant(
+        environmentId,
+        organizationId,
+        userId,
+        tenant
+      );
+    } catch (e) {
+      tenantEntity = null;
```
davidsoderberg, line 35: Why do we do null here? will getTenant in some cases return null otherwise we could reduce this code with just do the catch with no code and skip the null check on line 38?

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
```diff
@@ -112,4 +120,58 @@ export abstract class SendMessageBase extends SendMessageType {
...
+  protected async handleTenantExecution(job: JobEntity): Promise<TenantEntity | null> {
+    const tenantIdentifier = job.tenant?.identifier;
+
+    let tenant: TenantEntity | null = null;
+    if (tenantIdentifier) {
+      tenant = await this.tenantRepository.findOne({
+        _environmentId: job._environmentId,
+        identifier: tenantIdentifier,
+      });
+      if (!tenant) {
```
davidsoderberg, line 155: Could this be a part of sendSelectedTenantExecution?

apps/web/src/pages/templates/components/TriggerSnippetTabs.tsx
```diff
@@ -19,14 +20,20 @@ export function TriggerSnippetTabs({ trigger }: { trigger: INotificationTrigger
   const toValue = getSubscriberValue(subscriberVariables, (variable) => variable.value || '<REPLACE_WITH_DATA>');
   const payloadValue = getPayloadValue(trigger.variables);
 
+  const reservedValue = triggerSnippetVariables.reduce((acc, variable) => {
```
davidsoderberg, line 23: This will not create a lot of rerenders?

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -142,14 +154,43 @@ export class ParseEventRequest {
...
+    command: ParseEventRequestCommand,
+    reservedVariablesTypes: TriggerContextTypeEnum[]
+  ): void {
+    const invalidKeys: string[] = [];
+
+    for (const reservedVariableType of reservedVariablesTypes) {
+      const payload = command[reservedVariableType];
+      if (!payload) {
+        invalidKeys.push(`${reservedVariableType} object`);
+      } else {
```
davidsoderberg, line 176: could we do a continue instead of else to make it less nested?

apps/web/src/pages/templates/components/TriggerSnippetTabs.tsx
LetItRock: The `Tabs` component is not memoized, so it will re-render it and all the children when `TriggerSnippetTabs` will render, like it is right now. 😉

The thing that we can improve is memoize the calculations for this variable, so React won't need to do it every time on a new re-render. Actually same for tabs.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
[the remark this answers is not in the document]
ainouzgali: Whoa. You are right. Didn't think about it. Good catch🙏🏻

apps/api/src/app/events/events.controller.ts
```diff
@@ -63,6 +69,7 @@ export class EventsController {
         overrides: body.overrides || {},
         to: body.to,
         actor: body.actor,
+        tenant: body.tenant,
```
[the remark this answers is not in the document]
ainouzgali: I map it inside the usecase, similar to the actor

LetItRock (approved):
🚀

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -59,6 +62,7 @@ import {
   WebhookFilterBackoffStrategy,
 } from './usecases';
 import { MetricQueueService } from './services/metric-queue.service';
+import { ProcessTenant } from '@novu/application-generic/build/main/usecases/process-tenant';
```
p-fernandez, line 65: @ainouzgali this import slipped through the cracks. Could you please fix it in `next`? 🙏🏻

PR #3996, opened 2023-08-18
feat(types): create enum for the web socket events

p-fernandez:

### DIFF
--- a/apps/api/src/app/widgets/usecases/mark-all-messages-as/mark-all-messages-as.usecase.ts
+++ b/apps/api/src/app/widgets/usecases/mark-all-messages-as/mark-all-messages-as.usecase.ts
@@ -7,7 +7,7 @@ import {
   buildFeedKey,
   buildMessageCountKey,
 } from '@novu/application-generic';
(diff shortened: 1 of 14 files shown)


LetItRock (approved):
finally someone did that! thank you! 🎉

djabarovgeorge:
Amazing! do you think we should update [here](https://github.com/novuhq/novu/blob/b8b50f2b41dff4d96521c0b3191d108d0e78748a/packages/notification-center-vue/src/lib/notification-center-component/NotificationCenterComponent.vue#L82) as well?

p-fernandez:
> Amazing! do you think we should update [here](https://github.com/novuhq/novu/blob/b8b50f2b41dff4d96521c0b3191d108d0e78748a/packages/notification-center-vue/src/lib/notification-center-component/NotificationCenterComponent.vue#L82) as well?

I didn't do it because we are not importing the Shared package in the notification centers. And we probably wouldn't want to leave them as smaller as possible. We can always think the needs and leave local types for the notification center package.

PR #3999, opened 2023-08-18
Fix: Add Redis password into docker-compose and sample env

harrisyn:

Adds the REDIS_PASSWORD env to the sample env and docker-compose for local deployment

Addresses and instance where a redis password is used, this env is already being referenced in the redis.adapter file but hasn't been passed into the container.


scopsy (approved):

scopsy:
Thank you for your contribution @harrisyn 🙏

PR #4011, opened 2023-08-22
feat: refactor add job logs

djabarovgeorge:

refactor add job usecase logs.
 
* some of the logs were printed in debug but were not supposed to.
* added return if the step is delayed and not delay amount exists.
* add more indication either it is a delay or a digest step.

hope the change make sense for the flow.


packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -102,12 +114,7 @@ export class AddJob {
       })
     );
 
-    if (delay === null) {
-      Logger.warn(
-        'Variable delay is null which is not apart of the definition',
-        LOG_CONTEXT
-      );
```
djabarovgeorge, line 109: this should not be a possible step because we return on each `amount` === undefined.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -102,12 +114,7 @@ export class AddJob {
       })
     );
 
-    if (delay === null) {
-      Logger.warn(
-        'Variable delay is null which is not apart of the definition',
```
djabarovgeorge, line 107: this should not be a possible step because we return on each amount === undefined.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -102,12 +114,7 @@ export class AddJob {
...
-        LOG_CONTEXT
-      );
-    }
```
djabarovgeorge, line 110: this should not be a possible step because we return on each amount === undefined.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -59,17 +59,32 @@ export class AddJob {
...
         ? await this.addDelayJob.execute(command)
         : undefined;
-    Logger.debug(`DelayAmount is: ${delayAmount}`, LOG_CONTEXT);
 
-    if (job.type === StepTypeEnum.DIGEST && digestAmount === undefined) {
+    if (job.type === StepTypeEnum.DELAY) {
+      Logger.debug(`Delay Amount is: ${delayAmount}`, LOG_CONTEXT);
+    }
+
+    if (job.type === StepTypeEnum.DELAY && delayAmount === undefined) {
```
djabarovgeorge, line 85: added log and return on delay step + no amount data.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -136,14 +143,21 @@ export class AddJob {
...
-      Logger.verbose(
-        'Delay is active, Creating execution details',
-        LOG_CONTEXT
-      );
+      const logMessage =
+        job.type === StepTypeEnum.DELAY
+          ? 'Delay is active, Creating execution details'
+          : job.type === StepTypeEnum.DIGEST
+          ? 'Digest is active, Creating execution details'
+          : 'Unexpected job type, Creating execution details';
```
djabarovgeorge, line 151: made it like this and not with template string so it will be easier to discover this log.

djabarovgeorge: the unexpected backoff should not happen added it just in case for the future. we can remove it if it looks complex.

packages/application-generic/src/usecases/create-execution-details/types/index.ts
```diff
@@ -8,6 +8,7 @@ export enum DetailEnum {
   STEP_CREATED = 'Step created',
   STEP_QUEUED = 'Step queued',
   STEP_DELAYED = 'Step delayed',
+  STEP_DIGESTED = 'Step digested',
```
djabarovgeorge, line 11: Added so it will be more informative because we could have a delay and digest delay. let me know if it makes sense.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -59,17 +59,32 @@ export class AddJob {
...
+    }
+
+    if (job.type === StepTypeEnum.DIGEST && digestAmount === undefined) {
+      Logger.warn(
+        `Digest Amount does not exist on a digest job ${job._id}`,
+        LOG_CONTEXT
+      );
+
+      return;
+    }
```
scopsy, line 74: Amazing! had a lot of duplicate logs that are not needed

scopsy (approved):

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
djabarovgeorge: This PR started because of them 😅

PR #4015, opened 2023-08-22
[patch] - nv 2701 - bug report cannot create custom smtp integration

### LINKED ISSUE #3976: 🐛 Bug Report: Cannot create custom SMTP integration
### 📜 Description

I cannot add new custom SMTP integrations after providing all required fields.
It errors on the TLS options and complains that it needs to be an object.

By the looks of it, it is always serialized as a string.

image

### 👟 Reproduction steps

1. Follow guide to add new SMTP connection:
   https://docs.novu.co/channels/email/custom-smtp/
2. Provide all required fields
3. Click on Update

BOOM, error.

### 👍 Expected behavior

Custom SMTP provider should be created.

Step fails completely, leaving the new provider in a default state.

### Novu version

Novu SaaS

### npm version

_No response_

### node version

_No response_

_No response_

None

djabarovgeorge:

Closes #3976

### DIFF
--- a/apps/api/src/app/integrations/usecases/check-integration/check-integration.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/check-integration/check-integration.usecase.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, NotFoundException } from '@nestjs/common';
 import { CheckIntegrationEMail } from './check-integration-email.usecase';
 import { CheckIntegrationCommand } from './check-integration.command';
 import { ChannelTypeEnum, ICredentialsDto } from '@novu/shared';
@@ -8,9 +8,19 @@ export class CheckIntegration {
   constructor(private checkIntegrationEmail: CheckIntegrationEMail) {}
 
   public async execute(command: CheckIntegrationCommand) {
-    switch (command.channel) {
-      case ChannelTypeEnum.EMAIL:
-        return await this.checkIntegrationEmail.execute(command);
+    try {
+      switch (command.channel) {
+        case ChannelTypeEnum.EMAIL:
+          return await this.checkIntegrationEmail.execute(command);
+      }
+    } catch (e) {
+      if (e.message?.includes('getaddrinfo ENOTFOUND')) {
+        throw new NotFoundException(
+          `Provider gateway can't resolve the host with the given hostname ${command.credentials?.host || ''}`
+        );
+      }
+
+      throw e;
     }
   }
 }
--- a/apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
+++ b/apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
@@ -291,7 +291,7 @@ export function UpdateProviderSidebar({
               <Controller
                 name={`credentials.${credential.key}`}
                 control={control}
-                defaultValue={credential.type === 'boolean' || credential.type === 'switch' ? false : ''}
+                {...(credential.type === 'boolean' || credential.type === 'switch' ? { defaultValue: false } : {})}
                 rules={{
                   required: credential.required ? `Please enter a ${credential.displayName.toLowerCase()}` : undefined,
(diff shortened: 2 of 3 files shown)


apps/api/src/app/integrations/usecases/check-integration/check-integration.usecase.ts
```diff
@@ -8,9 +8,19 @@ export class CheckIntegration {
...
+      switch (command.channel) {
+        case ChannelTypeEnum.EMAIL:
+          return await this.checkIntegrationEmail.execute(command);
+      }
+    } catch (e) {
+      if (e.message?.includes('getaddrinfo ENOTFOUND')) {
+        throw new NotFoundException(
+          `Provider gateway can't resolve the host with the given hostname ${command.credentials?.host || ''}`
+        );
+      }
```
djabarovgeorge, line 21: Added additional catch so we could provide more information error to the user. i was not sure what status to add because this one is not found gateway (502) but not in novu platform.

apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
```diff
@@ -291,7 +291,7 @@ export function UpdateProviderSidebar({
               <Controller
                 name={`credentials.${credential.key}`}
                 control={control}
-                defaultValue={credential.type === 'boolean' || credential.type === 'switch' ? false : ''}
+                {...(credential.type === 'boolean' || credential.type === 'switch' ? { defaultValue: false } : {})}
```
djabarovgeorge, line 294: this one makes the issue IMO we could remove the default for boolean and switches as well but i decided to keep it as it was already there, on other params i don't see any reason to store any default in the db.

apps/web/src/pages/integrations/useProviders.ts
```diff
@@ -41,7 +41,10 @@ function initializeProviders(integrations: IntegrationEntity[]): IIntegratedProv
       if (integration?.credentials && Object.keys(clonedCredentials).length !== 0) {
         clonedCredentials.forEach((credential) => {
           // eslint-disable-next-line no-param-reassign
-          credential.value = integration.credentials[credential.key]?.toString();
+          credential.value =
+            credential.type === 'boolean' || credential.type === 'switch'
+              ? integration.credentials[credential.key]
+              : integration.credentials[credential.key]?.toString();
```
djabarovgeorge, line 47: found this issue in the way, we stringify boolean params which is wrong, switch component's in the form will think that "false" is truthy.
it should be in prod as well IMO but i could not reproduce it, could be related to the FF's

davidsoderberg: yeah this one have been hitting me before a tricky one and does not seems to happen in all places.

davidsoderberg (approved):

apps/web/src/pages/integrations/useProviders.ts
djabarovgeorge: Very strange it should happen where we initializeProvidersByIntegration which should happen when isMultiProviderConfigurationEnabled.

apps/api/src/app/integrations/usecases/check-integration/check-integration.usecase.ts
LetItRock: If I would change the host to something correct I won't get this error, right? if yes, I feel it should be more bad request error

apps/web/src/pages/integrations/useProviders.ts
LetItRock: `initializeProviders` is called when the `isMultiProviderConfigurationEnabled: false`, we have a similar one `initializeProvidersByIntegration` but it's fixed there

djabarovgeorge: > `initializeProviders` is called when the `isMultiProviderConfigurationEnabled: false`, we have a similar one `initializeProvidersByIntegration` but it's fixed there

totally right i mismatched the function names 😔

PR #4017, opened 2023-08-22
fix(web): delete primary integration - show the select primary modal when there are active integrations

LetItRock:

**NOTE: I do plan to cover all the functionality with e2e tests during the cooldown**

Bug: the empty Select Primary Integration modal is shown when the user deletes a primary integration.

Fix: show the Select Primary Integration modal only when there are active integrations.

https://github.com/novuhq/novu/assets/2607232/3c9b37b7-61b8-43bb-a5fd-76e6283e75a8


apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
```diff
@@ -36,7 +36,7 @@ export const UpdateIntegrationSidebarHeader = ({
 
     const { channel: selectedChannel, environmentId, integrationId, primary } = provider;
     const hasSameChannelActiveIntegration = !!providers
-      .filter((el) => !NOVU_PROVIDERS.includes(el.providerId) && el.integrationId !== integrationId)
```
LetItRock, line 39: This was missed when we moved the Novu integrations to DB. We do allow selecting Novu integrations as a primary integration.

apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
```diff
@@ -59,7 +59,7 @@ export const UpdateIntegrationSidebarHeader = ({
       return;
     }
 
-    if (provider.primary) {
+    if (shouldSetNewPrimary) {
```
LetItRock, line 62: this flag consists of: 
- if the deleted integration is the primary
- and there are some other active integrations from the same channel and env

and if both are true we will show the Select Primary Integration modal

ainouzgali (approved):

PR #4023, opened 2023-08-23
refactor: move subscriber mapping to worker

djabarovgeorge:
please dont review i wanted to trigger the ci.

refactor - more the subscriber mapping.

there is no reason to wait for the mapping on the api.


packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -91,22 +95,6 @@ export class TriggerEvent {
...
-    /*
-     * Makes no sense to execute anything if template doesn't exist
-     * TODO: Send a 404?
-     */
-    if (!template) {
-      const message = 'Notification template could not be found';
-      const error = new ApiException(message);
-      Logger.error(error, message, LOG_CONTEXT);
-      throw error;
-    }
```
djabarovgeorge, line 103: this part was redundant because we check it in ParseEventRequest use case and throw an exception

djabarovgeorge:
Close this one as i recreated this ticket in https://github.com/novuhq/novu/pull/4205

PR #4025, opened 2023-08-24
fix: cypress json error

davidsoderberg:

Fix the failing cypress runs
https://github.com/cypress-io/cypress/issues/27555

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -138,7 +138,7 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@types/react": "^17.0.0",
     "@types/testing-library__jest-dom": "^5.14.5",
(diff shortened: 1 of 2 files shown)


djabarovgeorge:
Interesting I wonder if it will fix the tests, i tried in the past to rollback to our previous version which was 12.3 i think and it did not help

djabarovgeorge (approved):
🥳

djabarovgeorge:
i saw that all tests passed but one which was in progress, not sure what happened in the end there 😅

scopsy:
@djabarovgeorge @davidsoderberg there is a suggestion that it was fixed in 13.2.0 from someone on GitHub. Worth trying

PR #4027, opened 2023-08-24
feat: move tabs state to routes

davidsoderberg:

Fix so settings are using routes instead of tabs internal state.


LetItRock (approved):
amazing! 🎉

PR #4028, opened 2023-08-24
Fix cypress json error

davidsoderberg:


apps/web/package.json
```diff
@@ -138,7 +138,7 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@types/react": "^17.0.0",
     "@types/testing-library__jest-dom": "^5.14.5",
-    "cypress": "^12.17.2",
+    "cypress": "^12.14.0",
```
scopsy, line 141: If that worked maybe we can upgrade to latest again?

.github/workflows/reusable-web-e2e.yml
```diff
@@ -116,7 +113,7 @@ jobs:
           working-directory: apps/web
           browser: "${{ env.BROWSER_PATH }}"
           record: true
-          parallel: true
+          parallel: false
```
djabarovgeorge, line 116: does that mean that the ci will be much slower? 🫣

scopsy: Unfortunately yes, but now it's still slower because it's failing every second time :(

scopsy (approved):

PR #4038, opened 2023-08-28
refactor: so filter logic from message matcher to filter abstract class

davidsoderberg:
I move some of the methods that will be used for conditions as well from message matcher so we can reuse it. I was not able to move as much as I would like to lets see if we can move some more when we are doing the conditions equal usecase.

### DIFF
--- a/apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.spec.ts
+++ b/apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.spec.ts
@@ -1053,7 +1053,7 @@ describe('Message filter matcher', function () {
     it('should add a passed condition', () => {
       const result = MessageMatcher.sumFilters(
         {
-          stepFilters: [],
+          filters: [],
           failedFilters: [],
           passedFilters: ['payload'],
         },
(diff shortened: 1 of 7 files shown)


apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.spec.ts
```diff
@@ -1053,7 +1053,7 @@ describe('Message filter matcher', function () {
     it('should add a passed condition', () => {
       const result = MessageMatcher.sumFilters(
         {
-          stepFilters: [],
```
davidsoderberg, line 1056: Rename this to filters so it is not workflow specific.

packages/application-generic/src/utils/filter-processing-details.ts
```diff
@@ -1,7 +1,11 @@
-import { StepFilter } from '@novu/dal';
```
davidsoderberg, line 1: Moved from api to application-generic.

packages/application-generic/src/utils/filter.ts
```diff
@@ -0,0 +1,139 @@
+import * as _ from 'lodash';
```
davidsoderberg, line 1: Reefactored and moved from Message matcher

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -153,19 +159,7 @@ export class MessageMatcher {
       type = 'online';
```
davidsoderberg, line 159: This is step specific so I keep it in the class and just pass it on tho the Filter class.

ainouzgali (approved):
looks good :)

packages/application-generic/src/utils/filter-processing-details.ts
```diff
@@ -1,7 +1,11 @@
...
-import { ICondition } from '@novu/shared';
+import { StepFilter, SubscriberEntity } from '@novu/dal';
+import { ICondition, ITriggerPayload } from '@novu/shared';
 
-import { IFilterVariables } from './types';
+export interface IFilterVariables {
```
ainouzgali, line 4: Might need to add tenant here as well. Like when we have the whole subscriber entity, we might need the tenant entity as well.

davidsoderberg: I do that in my next pr

PR #4044, opened 2023-08-29
Nv 2702 selectintegration use case update the logic for tenants

davidsoderberg:


packages/application-generic/src/usecases/conditions-filter/conditions-filter.command.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export interface IFilter {
+  isNegated: boolean;
+
+  type: BuilderFieldType;
+
+  value: BuilderGroupValues;
+
+  children: FilterParts[];
+}
```
ainouzgali, line 17: can we reuse an existing filter type?

packages/application-generic/src/utils/filter-processing-details.ts
```diff
@@ -1,10 +1,11 @@
 import { StepFilter, SubscriberEntity } from '@novu/dal';
-import { ICondition, ITriggerPayload } from '@novu/shared';
+import { ICondition, ITenantDefine, ITriggerPayload } from '@novu/shared';
 
 export interface IFilterVariables {
   payload?: ITriggerPayload;
   subscriber?: SubscriberEntity;
   webhook?: Record<string, unknown>;
+  tenant?: ITenantDefine;
```
ainouzgali, line 8: this should be TenantEntity IMO. as we have a filter with the thenat name, and in the ITenantDefine we only force to have an identifier. 

Should be similar to the subscriber

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -52,6 +55,51 @@ export class SelectIntegration {
       return integrations[0];
     }
 
+    let integration: IntegrationEntity | null = null;
+
+    if (command.filterData.tenant) {
```
ainouzgali, line 60: if we have an integrationIdentifier in the overrides, I think it should go by  that one and not do it  by the tenant filters

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -0,0 +1,145 @@
...
+  private async handleOrFilters(
+    filter: IFilter,
+    variables: IFilterVariables,
+    filterProcessingDetails: FilterProcessingDetails
+  ): Promise<boolean> {
+    const foundFilter = await this.findAsync(filter.children, (i) =>
+      this.processFilter(variables, i, filterProcessingDetails)
+    );
+
+    return foundFilter ? true : false;
```
ainouzgali, line 125:
```suggestion
    return !!foundFilter ;
```

libs/dal/src/repositories/integration/integration.entity.ts
```diff
@@ -32,6 +32,16 @@ export class IntegrationEntity {
...
+
+  conditions?: {
+    isNegated: boolean;
+
+    type: BuilderFieldType;
+
+    value: BuilderGroupValues;
+
+    children: FilterParts[];
+  }[];
```
ainouzgali, line 44: Maybe lets use a type here

apps/api/src/app/shared/dtos/step-filter.ts
```diff
@@ -102,7 +109,8 @@ type FilterParts =
   | WebhookFilterPart
   | RealtimeOnlineFilterPart
   | OnlineInLastFilterPart
-  | PreviousStepFilterPart;
+  | PreviousStepFilterPart
+  | ITenantFilterPart;
```
ainouzgali, line 113: All the others are a class with the api properties, and are not imported from shared. Maybe we should stay consistent and add a TenantFilterPart here as well

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -57,7 +63,7 @@ export class SelectIntegration {
 
     let integration: IntegrationEntity | null = null;
 
-    if (command.filterData.tenant) {
+    if (command.identifier) {
```
ainouzgali, line 66: @davidsoderberg I think here it should still be the check that tenant is included in the command. 
What I meant in my previous comment about the integrationIdentifier override  (which is command.identifier here) is that I think it should take precedence over the tenant filters. I mean that if someone used the integration identifier override - that one should be used and not by the tenant. 
Right now, you are doing that check in the `getPrimaryIntegration` and only if the filters for tenant didn't find an integration

apps/api/src/app/shared/dtos/step-filter.ts
davidsoderberg: This will be solved in my next PR: https://github.com/novuhq/novu/pull/4048

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -52,6 +61,58 @@ export class SelectIntegration {
       return integrations[0];
     }
 
+    let integration: IntegrationEntity | null =
+      await this.getPrimaryIntegration(command);
```
ainouzgali, line 65: I'm concerned about getting the primary unnecessarily ... but lets get back to it

ainouzgali (approved):

PR #4048, opened 2023-08-30
feat: add api to add conditions on create and update integration

davidsoderberg:

Tests will be done in https://linear.app/novu/issue/NV-2771/🎛️-api-e2e-tests-for-the-integration-conditions.
Also solves https://linear.app/novu/issue/NV-2683/🎛️-apply-conditions-when-the-integration-is-updated.


apps/api/src/app/shared/dtos/step-filter.ts
```diff
@@ -104,13 +97,21 @@ class PreviousStepFilterPart extends BaseFilterPart {
   stepType: PreviousStepTypeEnum;
 }
 
+class TenantFilterPart extends BaseFieldFilterPart {
+  @ApiProperty({
+    enum: [FilterPartTypeEnum.TENANT],
+    description: 'Only on integrations right now',
```
davidsoderberg, line 103: What do you guys think about this? because we are not having tenant on step filters yet.

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -201,7 +203,7 @@ export class UpdateIntegration {
         active: !!command.active,
       });
 
-      updatePayload.primary = primary;
+      updatePayload.primary = command.conditions && command.conditions.length > 0 ? false : primary;
```
ainouzgali, line 206: What about a case that the primary received true (from the calculate func) => as I understand it primary will also always have the highest priority. 
So could that state happen here? that we are setting primary to false but still the priority here will be set as the highest

ainouzgali (commented):
@davidsoderberg we will also need to change the node sdk to accept the conditions. Do you want to open a ticket for it or do it here?

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
davidsoderberg: I guess so not sure how we should handle it thu... @LetItRock ?

ainouzgali: maybe in the calculate function itself we can check if its now primary and conditions were added in the current update - and if so, recalculate all priorities ? maybe similar to what we do if the active is changed to inactive

apps/api/src/app/integrations/dtos/create-integration-request.dto.ts
```diff
@@ -53,4 +54,9 @@ export class CreateIntegrationRequestDto implements ICreateIntegrationBodyDto {
   @IsOptional()
   @IsBoolean()
   check?: boolean;
+
+  @ApiPropertyOptional({
+    type: [StepFilter],
+  })
+  conditions?: StepFilter[];
```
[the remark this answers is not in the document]
LetItRock: `@ValidateNested({ each: true })`... but the thing is that the `StepFilter` doesn't have any validation annotation on the class... I see that in the `notification-template.controller` we use `MessageFilter` class

apps/api/src/app/integrations/dtos/integration-response.dto.ts
```diff
@@ -45,4 +47,10 @@ export class IntegrationResponseDto {
 
   @ApiProperty()
   primary: boolean;
+
+  @ApiPropertyOptional({
+    type: [StepFilter],
+  })
+  @ValidateNested({ each: true })
```
LetItRock, line 54: on the response, we don't need validation ;) probably you mixed it up with request dto: `CreateIntegrationRequestDto`

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
LetItRock: @davidsoderberg 
This is not the right place, because this if statement will only execute when the "active" flag changes. 


But what we actually want to do is independent of the "active" flag, same as @ainouzgali is suggesting DoD: 
`if existing integration is marked as primary and conditions are updated (exists) we should remove that primary flag and update the priority for all active integrations`.

apps/api/src/app/shared/dtos/step-filter.ts
LetItRock: I would say, that we are actually allowing to do that from the API, and the API docs are actually for those who use it, so I don't think we need to mention that it's only for integrations, because we can set it now on the steps right? or we don't have the logic yet?

ainouzgali: We don't have that logic yet for steps. It is still the filters structure from before. And as far as I know its not a part of the cycle right now...

LetItRock (approved):
🎉

PR #4050, opened 2023-08-30
feat: fix so conditions are emptied when integration are set as primary

davidsoderberg:
Not sure it's supposed to be this easy but. it should empty conditions when an integration is set as primary.


LetItRock (approved):
I love this one :D ❤️ can we have a quick e2e test to confirm that we are doing removal?

PR #4053, opened 2023-08-30
fix: critical vulnerability in vm2 imported by inline css

scopsy:
Following the advice at https://github.com/TooTallNate/superagent-proxy/issues/48 to override the vulnerable vm2 version due to lack of support from superagent proxy library and the parent inline-css library


p-fernandez (approved):
🌟

PR #4060, opened 2023-08-30
Make the Agents Metrics Accurate Again

Cliftonz:

This pr corrects the times and granularity for metrics with bullmq.
Why was this change needed?

This pr will allow use to accurately autoscale our workers by sending the historical job processing data to a 3rd party New relic.


apps/worker/src/app/shared/utils/cron-health.ts
```diff
@@ -0,0 +1,14 @@
+import { Logger } from '@nestjs/common';
+
+const url = process.env.CRON_CHECKIN_URL
+  ? process.env.CRON_CHECKIN_URL
+  : 'https://uptime.betterstack.com/api/v1/heartbeat/';
```
p-fernandez, line 5: Are we sending a beat to an external service after each job is completed?
Are we going to individually do that for every queue? If it is to provide uptime stats I think how is it implemented only sends the liveness of the job completed metrics queue but not the actual queues that process jobs. Am I missing anything?

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -0,0 +1,115 @@
...
+        for (const jobElement of job) {
+          if (jobElement.id === METRIC_JOB_ID) {
+            exists = true;
+          }
+        }
+
+        return exists;
+      })
+    )
+      .then(async (exists): Promise<void> => {
```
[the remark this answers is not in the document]
Cliftonz: Close, we inject a repeatable job into the queue (where the queue on redis is the scheduler)  to get the stats of the queue not the processor of the worker. The worker stats are already covered by ECS at this moment.

apps/worker/src/app/shared/utils/cron-health.ts
Cliftonz: Yes, we are sending it to better uptime.

You are correct this is just for the metric queue. These queues are critical for auto-scaling the workers to ensure that there are enough.

p-fernandez: Oh I get it now. So it is to based on active jobs and completed jobs understand the pressure of the system. 👍🏻

apps/worker/src/app/workflow/services/metric-queue-active.service.ts
```diff
@@ -0,0 +1,115 @@
...
+      await checkinForCronJob(process.env.ACTIVE_CRON_ID);
+      Logger.verbose('Metric job Completed', job.id, LOG_CONTEXT);
+    });
+
+    this.bullMqService.worker.on('failed', async (job, error) => {
+      Logger.verbose('Metric job failed', LOG_CONTEXT, error);
+    });
+
+    this.addToQueueIfMetricJobExists();
+  }
```
p-fernandez, line 28:
```suggestion
export class MetricQueueActiveService extends QueueService<Record<string, never>> {
  constructor(@Inject('BULLMQ_LIST') private token_list: QueueService[]) {
    super(JobTopicNameEnum.METRICS_ACTIVE);

    if (process.env.NOVU_MANAGED_SERVICE === 'true' && process.env.NEW_RELIC_LICENSE_KEY) {
      this.bullMqService.createWorker(this.name, this.getWorkerProcessor(), this.getWorkerOpts());

      this.bullMqService.worker.on('completed', async (job) => {
        await checkinForCronJob(process.env.ACTIVE_CRON_ID);
        Logger.verbose('Metric job Completed', job.id, LOG_CONTEXT);
      });

      this.bullMqService.worker.on('failed', async (job, error) => {
        Logger.verbose('Metric job failed', LOG_CONTEXT, error);
      });

      this.addToQueueIfMetricJobExists();
    }
  }
```
This is my suggestion for this comment: https://github.com/novuhq/novu/pull/4060/files#r1311743616

Thus for self hosted users we do not initialise the worker as they are not going to use it or need it. We might think in the future how to redo this if users request a feature to have a similar uptime implementation like us.

apps/worker/src/app/shared/utils/cron-health.ts
Cliftonz: Correct, as that is all we need. If processing stopped tomorrow because there was no use then we would not need to alert just check the pressure and scale down.

p-fernandez:
Moving to this PR: https://github.com/novuhq/novu/pull/4099

PR #4064, opened 2023-08-31
Nv 2781 move auth logic to application generic

djabarovgeorge:

extract auth logic to application generic.

So we could reuse it in other packages as well.


packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -90,7 +117,36 @@ export class AuthService {
...
+          if (!user) throw new ApiException('User not found');
+        }
+      }
+
+      if (authProvider === AuthProviderEnum.GOOGLE) {
+        const withoutUsername = user.tokens.find(
+          (token) =>
+            token.provider === AuthProviderEnum.GOOGLE &&
+            !token.username &&
+            String(token.providerId) === String(profile.id)
```
LetItRock, line 149: This logic and the below code is duplicated between Google and GitHub. Can you please move it to a function?

LetItRock (approved):

PR #4067, opened 2023-08-31
feat: add external oauth module

djabarovgeorge:

### DIFF
--- a/apps/api/src/app.module.ts
+++ b/apps/api/src/app.module.ts
@@ -4,6 +4,8 @@ import { APP_INTERCEPTOR } from '@nestjs/core';
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { ForwardReference } from '@nestjs/common/interfaces/modules/forward-reference.interface';
 
+import { EEAuthModule } from '@novu/ee-auth';
+
 import { SharedModule } from './app/shared/shared.module';
 import { UserModule } from './app/user/user.module';
 import { AuthModule } from './app/auth/auth.module';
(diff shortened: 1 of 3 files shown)


scopsy (approved):

LetItRock (approved):

apps/api/src/app.module.ts
```diff
@@ -4,6 +4,8 @@ import { APP_INTERCEPTOR } from '@nestjs/core';
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { ForwardReference } from '@nestjs/common/interfaces/modules/forward-reference.interface';
 
+import { EEAuthModule } from '@novu/ee-auth';
```
LetItRock, line 7: Ahh, this might not work for contributors that won't have access to the ee packages repository, like they won't be able to build it and will get an error when accessing this package.
We need to do the trick with `require` as we have for the `bullmq-pro` package. Something like this, but you have to check:

```
const getExternalModules = () => {
   const modules = [];

   const eeAuth = require('@novu/ee-auth');
   if (eeAuth) {
      modules.push(eeAuth.EEAuthModule);
   }

   return modules;
}

const externalModules: Array<Type | DynamicModule | Promise<DynamicModule> | ForwardReference> = getExternalModules();
```

LetItRock (changes requested):

apps/api/src/app.module.ts
djabarovgeorge: Yeah you are totally right and I am aware of this issue, i am testing currently what would be the cleanest solution for us.

apps/api/src/app.module.ts
```diff
@@ -61,6 +69,12 @@ const modules: Array<Type | DynamicModule | Promise<DynamicModule> | ForwardRefe
   TenantModule,
 ];
 
+const eeAuthModule = importWhen({ path: '@novu/ee-auth', condition: process.env.NOVU_MANAGED_SERVICE === 'true' });
```
scopsy, line 72: Would suggest maybe considering even a higher level function to encapsulat this logic, for example `enterpriseImport('@novu/ee-auth')`

PR #4072, opened 2023-09-01
feat: add logic for select primary integration modal

davidsoderberg:
Not sure the logic is right so please double-check it otherwise we should be good.

### DIFF
--- a/apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
+++ b/apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
@@ -64,11 +64,7 @@ export const UpdateIntegrationSidebarHeader = ({
         environmentId: provider.environmentId,
         channelType: provider.channel,
         exclude: (el: IntegrationEntity) => {
-          if (el._id === provider.integrationId || (el.conditions && el.conditions.length > 0)) {
-            return true;
-          }
-
-          return false;
+          return el._id === provider.integrationId;
         },
         onClose: () => {
           deleteIntegration({
--- a/apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
+++ b/apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
@@ -6,7 +6,7 @@ import { CHANNELS_WITH_PRIMARY, NOVU_PROVIDERS } from '@novu/shared';
 import { Button, colors, Dropdown, Modal, NameInput, Text, Title } from '../../../design-system';
 import { useFetchEnvironments } from '../../../hooks/useFetchEnvironments';
 import { ProviderImage } from './multi-provider/SelectProviderSidebar';
-import type { IIntegratedProvider } from '../types';
+import type { IIntegratedProvider, IntegrationEntity } from '../types';
 import { useProviders } from '../useProviders';
 import { useDeleteIntegration } from '../../../api/hooks';
 import { errorMessage, successMessage } from '../../../utils/notifications';
@@ -63,7 +63,13 @@ export const UpdateIntegrationSidebarHeader = ({
       openModal({
         environmentId: provider.environmentId,
         channelType: provider.channel,
-        exclude: [provider.integrationId],
+        exclude: (el: IntegrationEntity) => {
+          if (el._id === provider.integrationId || (el.conditions && el.conditions.length > 0)) {
+            return true;
+          }
+
+          return false;
+        },
         onClose: () => {
           deleteIntegration({
             id: provider.integrationId,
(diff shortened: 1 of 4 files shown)


apps/web/src/pages/integrations/types.ts
```diff
@@ -40,6 +40,7 @@ export interface IIntegratedProvider {
   name?: string;
   identifier?: string;
   primary: boolean;
+  conditions?: any[];
```
ainouzgali, line 43: Could we type those?

apps/web/src/pages/integrations/components/multi-provider/SelectPrimaryIntegrationModal.tsx
```diff
@@ -152,7 +155,7 @@ export const SelectPrimaryIntegrationModal = ({
       return el.channel === channelType && isNotExcluded;
     });
 
-    return filteredIntegrations.map((el) => mapToTableIntegration(el, environments));
+    return filteredIntegrations.filter((el) => el.active).map((el) => mapToTableIntegration(el, environments));
```
ainouzgali, line 158: Shouldn't we allow to see the not active ones? Because if they select them as primary we also change them to active.

apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
ainouzgali, line 1: Should line 31 - canMarkAsPrimary - also take into account if there are conditions?

apps/web/src/pages/integrations/components/multi-provider/SelectPrimaryIntegrationModal.tsx
davidsoderberg: This was a requirement in the issue... @LetItRock maybe you have more info here.

apps/web/src/pages/integrations/components/UpdateIntegrationSidebarHeader.tsx
davidsoderberg: oh good question, what do you think @LetItRock ?

apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
```diff
@@ -167,22 +167,26 @@ export function UpdateProviderSidebar({
...
       .find((el) => el.active && el.channel === selectedChannel && el.environmentId === environmentId);
     const isChannelSupportPrimary = CHANNELS_WITH_PRIMARY.includes(selectedChannel);
+    const hasAddedCondition = primary && conditions && conditions.length > 0; // show modal
+    const hasNotAddedConditionOnActive = !(isActive && conditions && conditions.length > 0); // show modal
 
     if (
       isActiveFieldChanged &&
       isChannelSupportPrimary &&
+      hasAddedCondition &&
+      hasNotAddedConditionOnActive &&
```
LetItRock, line 183: I feel like this should be updated. Basically, you want to show the modal only when the current integration is primary and has conditions applied (no matter the active flag). So it will be like:
```
const isChangedToActive =  isActiveFieldChanged && isChannelSupportPrimary && isActive && hasSameChannelActiveIntegration;

const isChangedToInactiveAndIsPrimary = isActiveFieldChanged && isChannelSupportPrimary && !isActive && primary && hasSameChannelActiveIntegration;

const isPrimaryAndHasConditions = primary && conditions && conditions.length > 0;

if (isChangedToActive || isChangedToInactiveAndIsPrimary || isPrimaryAndHasConditions)
```

apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
```diff
@@ -167,22 +167,28 @@ export function UpdateProviderSidebar({
...
-      isChannelSupportPrimary &&
-      ((isActive && hasSameChannelActiveIntegration) || (!isActive && primary && hasSameChannelActiveIntegration))
-    ) {
+    const isChangedToActive =
+      isActiveFieldChanged && isChannelSupportPrimary && isActive && hasSameChannelActiveIntegration;
+
+    const isChangedToInactiveAndIsPrimary =
+      isActiveFieldChanged && isChannelSupportPrimary && !isActive && primary && hasSameChannelActiveIntegration;
+
+    const isPrimaryAndHasConditionsApplied = primary && conditions && conditions.length > 0;
```
LetItRock, line 183:
```suggestion
    const isPrimaryAndHasConditionsApplied = primary && conditions && conditions.length > 0 && hasSameChannelActiveIntegration;
```
forgot to mention that there is no sense to show empty modal if no integrations ;)

ainouzgali (approved):

PR #4073, opened 2023-09-01
feat(dal): Configure minimum connection pool size by env var

dmgarland:
Running Novu self-hosted on Mongo Atlas, we quickly hit problems with the 500 connection limit. The DAL service configures the connection pool based on `MONGO_MAX_POOL_SIZE` for the upper limit, and a hard-coded value based on `NODE_ENV` for the lower limit. For both our development and production boxes, we want to run the environment in production to benefit from CORS protection and other optimisations, so that entails that the minimum connection pool size is being set to 200. Given that the API, worker and web processes setup a mongo connection (I believe), this equates to 3 x 200 minimum connections and 3 x 500 maximum. This triggers alerts on Atlas and sometimes prevents our Novu service to come up, if say we restart the API process but the workers / web are hogging all the connections.

A minimum of 200 also means that we cannot set the `MONGO_MAX_POOL_SIZE` to be less than 200, so to workaround this we're setting our `MONGO_MAX_POOL_SIZE` to be 200 as well to reduce the likelihood of hitting these limits.

This issue affects all shared Mongo Atlas instances up to M5, so requiring an M10 instance to run Novu seems overkill.

To allow us to have a more modest connection pool for our dev instances, this PR adds a `MONGO_MIN_POOL_SIZE` so that we can adjust this for our setup to keep within the 500 limit. We'd probably have the following settings to keep within 500 connections:

MONGO_MIN_POOL_SIZE = 50
MONGO_MAX_POOL_SIZE = 166


p-fernandez:
@novuhq/novu-devs prior merging this we need to set up in the environment variables of production this:
```MONGO_MIN_POOL_SIZE = 200```

File: libs/dal/src/dal.service.ts [hunk 1]
```diff
@@ -7,7 +7,7 @@ export class DalService {
   async connect(url: string, config: ConnectOptions = {}) {
     const baseConfig: ConnectOptions = {
       maxPoolSize: process.env.MONGO_MAX_POOL_SIZE || 500,
-      minPoolSize: process.env.NODE_ENV === 'production' ? 200 : 10,
+      minPoolSize: process.env.MONGO_MIN_POOL_SIZE || 10,
```
p-fernandez, line 10: With this change we need to add:
`MONGO_MIN_POOL_SIZE = 200`

to the following environment variable files to guarantee retrocompatibility for our users:
- `/apps/api/src/.env.production`
- `/apps/webhook/src/.env.production`
- `/apps/worker/src/.env.production`
- `/apps/ws/src/.env.production`

p-fernandez (approved):
🌟

scopsy:
Unfortuently not, as 0.19.0 is already finalized and release notes are written right now. This should be available at 0.20.0 or the "next" tag

scopsy:
@dmgarland a small reminder, I would it do it myself, but have no push access to your fork. Once you can pull latest from next, we can run the test pipeline and merge 🙏

toddb:
Just a comment to anyone reading. This also affects Atlas Serverless and is made worse if you are running each of the novu services as a cluster.

Thanks for the change.

p-fernandez:
> @dmgarland a small reminder, I would it do it myself, but have no push access to your fork. Once you can pull latest from next, we can run the test pipeline and merge 🙏

@dmgarland could you provide @scopsy or me write rights to your repo so we can fix the failing test and merge your PR. 🙏🏻

dmgarland:
@p-fernandez @scopsy done!

p-fernandez:
Thanks for the contribution again and for having patience with us.

PR #4085, opened 2023-09-03
refactor: replace node-apns library with parse implementation

scopsy:

Replace the `node-apns` package with the fork from `@parse/node-apn`.

 
The original package is not maintained anymore and contains a few high vulnerabilities in sub-packages.


providers/apns/src/lib/apns.provider.ts
```diff
@@ -55,6 +55,8 @@ export class APNSPushProvider implements IPushProvider {
       );
     }
 
+    this.provider.shutdown();
```
scopsy, line 58: This was suggested by the library when creating Provider instances

p-fernandez (approved):
🌟

PR #4100, opened 2023-09-05
feat: allow novu providers to be activated/deactivated and primary

BiswaViraj:

- Allows novu providers to be activated/deactivated
- Ability choose novu providers to be primary

https://github.com/novuhq/novu/assets/29251776/7a5e11db-d059-43a3-9cc6-c99c3b7c0741


apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.usecase.ts
```diff
@@ -16,6 +16,7 @@ export class GetActiveIntegrations {
         organizationId: command.organizationId,
         environmentId: command.environmentId,
         userId: command.userId,
+        filterByEnvironment: command.filterByEnvironment,
```
BiswaViraj, line 19: Currently, the usecase used to return active active integrations of all environments
this will allow to filter and get active integrations for the current env

packages/application-generic/src/usecases/get-decrypted-integrations/get-decrypted-integrations.usecase.ts
```diff
@@ -31,7 +31,10 @@ export class GetDecryptedIntegrations {
       _organizationId: command.organizationId,
     };
 
-    if (command.environmentId && !isMultiProviderConfigurationEnabled) {
+    if (
+      (command.filterByEnvironment && command.environmentId) ||
+      (command.environmentId && !isMultiProviderConfigurationEnabled)
+    ) {
```
BiswaViraj, line 37: Same, allows filtering active integrations by env

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -179,15 +183,8 @@ export class CreateIntegration {
...
 
       const isActiveAndChannelSupportsPrimary = command.active && CHANNELS_WITH_PRIMARY.includes(command.channel);
-      if (isMultiProviderConfigurationEnabled && isActiveAndChannelSupportsPrimary) {
-        await this.disableNovuIntegration.execute({
-          channel: command.channel,
-          providerId: command.providerId,
-          environmentId: command.environmentId,
-          organizationId: command.organizationId,
-          userId: command.userId,
-        });
```
BiswaViraj, line 189: not needed, as we dont need to disable novu integrations

apps/web/src/pages/integrations/components/multi-provider/UpdateProviderSidebar.tsx
```diff
@@ -237,8 +237,9 @@ export function UpdateProviderSidebar({
           }
         >
           <NovuProviderSidebarContent provider={selectedProvider} />
-          <UpdateIntegrationCommonFields provider={selectedProvider} showActive={false} />
+          <UpdateIntegrationCommonFields provider={selectedProvider} showActive={true} />
```
LetItRock, line 240: we don't need now the `showActive` flag

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.command.ts
```diff
@@ -1,3 +1,8 @@
+import { IsBoolean, IsOptional } from 'class-validator';
 import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
 
-export class GetActiveIntegrationsCommand extends EnvironmentWithUserCommand {}
+export class GetActiveIntegrationsCommand extends EnvironmentWithUserCommand {
+  @IsBoolean()
+  @IsOptional()
+  filterByEnvironment?: boolean;
```
BiswaViraj, line 7: @LetItRock Should we revert this addition? it is not used in this PR but can be a little nice feature to have

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
```diff
@@ -69,7 +67,7 @@ export class CreateIntegration {
       result.priority = highestPriorityIntegration ? highestPriorityIntegration.priority + 1 : 1;
     }
 
-    const activeIntegrationsCount = await this.integrationRepository.countActiveExcludingNovu({
+    const activeIntegrationsCount = await this.integrationRepository.countActiveIncludingNovu({
```
LetItRock, line 70: I'm not sure about this change as it will be out of sync with the UI behavior.
 
The idea before was like that:
1. The user starts by default with Novu integrations, Novu Email is primary.
2. He creates the Mailjet integration.
3. Then, he activates the Mailjet integration. The UI doesn't show the "Select primary provider" modal. It becomes primary as it is the first integration activated excluding Novu integrations.

With the current code in step 3, the integration won't be primary, nor will the "Select primary provider" modal be shown, which is inconsistent with if the user creates another integration and activates it (the modal will be shown).

LetItRock: Let's have it like it was before, and we can think about it again.

apps/api/src/app/integrations/usecases/get-active-integration/get-active-integration.command.ts
LetItRock: I would say let's remove it, less code, less thinking, maintaining etc

apps/api/src/app/integrations/usecases/create-integration/create-integration.usecase.ts
BiswaViraj: Can you check again?
For me it does shows the select primary modal upon activating mailjet integration.
image

LetItRock: ahh ok, that's why I've added to the DoD this:
```
remove the code on the backend in the create/update integration use-case
- that is counting active integrations and set the primary flag when count is 0
```
ok, it makes sense now if we remove the above code too... 

Like I feel that the user should decide about the primary integration, and we shouldn't do it on his behalf... that's why we will always show the "Select primary provider" if the integration is activated.

WDYT?

BiswaViraj: Yes, in that case I agree with you, it will be better to let users choose
I will remove that part

LetItRock: thanks! sorry about the confusion :)

LetItRock (approved):
thanks! 🎉

PR #4101, opened 2023-09-05
feat: add interface for conditions in sdk

davidsoderberg:


packages/node/src/lib/integrations/integrations.interface.ts
```diff
@@ -1,5 +1,17 @@
-import { ChannelTypeEnum, ICredentialsDto } from '@novu/shared';
+import {
+  BuilderFieldType,
+  BuilderGroupValues,
+  ChannelTypeEnum,
+  FilterParts,
+  ICredentialsDto,
+} from '@novu/shared';
 
+interface ICondition {
```
ainouzgali, line 9:
```suggestion
export interface IConditions {
```
add an s - To be consistent with the name in the fe 
+ export the type

ainouzgali (approved):
lets also add a small test ?

PR #4105, opened 2023-09-05
refactor: conditions component

ainouzgali:

Refactor Conditions component


apps/web/src/components/conditions/Conditions.tsx
```diff
@@ -284,6 +213,93 @@ export function Conditions({
   );
 }
 
+function EqualityForm({ control, index }: { control: Control<IConditionsForm>; index: number }) {
```
davidsoderberg, line 216: this seems long enough to be in its own file?

PR #4107, opened 2023-09-05
Stacked PRs - add ee-auth package

djabarovgeorge:


LetItRock (approved):

.github/workflows/test.yml
```diff
@@ -224,6 +227,54 @@ jobs:
...
+      - uses: mansagroup/nrwl-nx-action@v3
+        name: Lint and build
+        with:
+          targets: lint,build
+          projects: ${{matrix.projectName}}
+
+      - uses: ./.github/actions/start-localstack
+
+      - uses: ./.github/actions/run-worker
+        if: ${{matrix.projectName == '@novu/api' }}
```
p-fernandez, line 262: Is this test suite planned to be executed for multiple apps? That means that the EE tests will live in API, Worker, WS, etc?

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -95,24 +97,39 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
...
+              my={30}
+              variant="white"
+              fullWidth
+              radius="md"
+              leftIcon={<Google />}
+              data-test-id="google-button"
+              sx={{ color: colors.B40, fontSize: '16px', fontWeight: 700, height: '50px', marginLeft: 10 }}
+            >
+              Sign In with Google
+            </GitHubButton>
```
p-fernandez, line 128: It doesn't make much sense to me that the component rendering the SSO for Google is called `GitHubButton`.
I probably would expect this component to look like:
```
<OAuth>
  <GitHubButton link={gitHubLink} />
  <GoogleButton token={invitationToken} />
 </OAuth>
```
With some of the logic encapsulated in the different button components.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -12,48 +19,65 @@ import {
...
+} from '../cache';
+import { normalizeEmail } from '../../utils/email-normalization';
 
 @Injectable()
 export class AuthService {
   constructor(
     private logger: PinoLogger,
     private userRepository: UserRepository,
     private subscriberRepository: SubscriberRepository,
+    @Inject(forwardRef(() => CreateUser))
```
p-fernandez, line 57: If moving this to App Generic creates a circular dependency we are doing something wrong. I can help you to find out the guilty relationship.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -66,8 +90,12 @@ export class AuthService {
...
           picture: profile.avatar_url,
           email,
-          lastName: profile.name ? profile.name.split(' ').slice(-1).join(' ') : null,
-          firstName: profile.name ? profile.name.split(' ').slice(0, -1).join(' ') : profile.login,
+          lastName: profile.name
+            ? profile.name.split(' ').slice(-1).join(' ')
+            : null,
+          firstName: profile.name
+            ? profile.name.split(' ').slice(0, -1).join(' ')
+            : profile.login,
```
p-fernandez, line 98: I'd isolate the first and last name capturing in own functions.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -124,6 +136,44 @@ export class AuthService {
...
+          'tokens.providerId': profile.id,
+        },
+        {
+          $set: {
+            'tokens.$.username': profile.login,
+          },
+        }
+      );
+
+      user = await this.userRepository.findById(user._id);
```
p-fernandez, line 170: This flow is a bit confusing even if I understand what it is doing.
I guess that the exception should be thrown if when updating the user is not found.
Also there must be an option to retrieve the updated entity in the return of the MongoDB functionality, saving us from doing the fetch later. I think we could review that part. Highly likely we would need to change our base repository class for that.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -201,20 +266,26 @@ export class AuthService {
...
-      );
+      const userActiveProjects =
+        await this.environmentRepository.findOrganizationEnvironments(
+          organizationToSwitch._id
+        );
       let environmentToSwitch = userActiveProjects[0];
 
-      const reduceEnvsToOnlyDevelopment = (prev, current) => (current.name === 'Development' ? current : prev);
+      const reduceEnvsToOnlyDevelopment = (prev, current) =>
+        current.name === 'Development' ? current : prev;
```
p-fernandez, line 282: Is there any possibility to have an enum with environment names or this forces you to do so many changes in the codebase?

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -334,8 +412,17 @@ export class AuthService {
...
   })
-  private async getSubscriber({ subscriberId, _environmentId }: { subscriberId: string; _environmentId: string }) {
-    return await this.subscriberRepository.findBySubscriberId(_environmentId, subscriberId);
+  private async getSubscriber({
+    subscriberId,
+    _environmentId,
+  }: {
+    subscriberId: string;
+    _environmentId: string;
+  }) {
```
p-fernandez, line 421: Is this return type right?

packages/application-generic/src/services/auth/shared.ts
```diff
@@ -0,0 +1,54 @@
+export const buildOauthRedirectUrl = (request): string => {
+  if (!request.user || !request.user.token) {
+    return `${
+      process.env.FRONT_BASE_URL + '/auth/login'
+    }?error=AuthenticationError`;
+  }
+
+  let url = process.env.FRONT_BASE_URL + '/auth/login';
```
p-fernandez, line 8:
```suggestion
  let url = `${process.env.FRONT_BASE_URL}/auth/login`;
  
  if (!request.user || !request.user.token) {
    return `${url}?error=AuthenticationError`;
  }
```

packages/application-generic/src/services/auth/shared.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const configurationId = JSON.parse(request.query.state).configurationId;
+  if (configurationId) {
+    url += `&configurationId=${configurationId}`;
+  }
+
+  const invitationToken = JSON.parse(request.query.state).invitationToken;
+  if (invitationToken) {
+    url += `&invitationToken=${invitationToken}`;
+  }
```
p-fernandez, line 51:
```suggestion
  const parsedQueryState = JSON.parse(request.query.state);
  const { configurationId, invitationToken, next, partnerCode } = parsedQueryState;
  if (partnerCode) {
    url += `&code=${partnerCode}`;
  }

  if (next) {
    url += `&next=${next}`;
  }

  if (configurationId) {
    url += `&configurationId=${configurationId}`;
  }

  if (invitationToken) {
    url += `&invitationToken=${invitationToken}`;
  }
```
And this could be isolated in a function that handles all these conditions to build app to the URL.

apps/web/src/pages/auth/components/LoginForm.tsx
djabarovgeorge: Totally write i probably missed this one

.github/workflows/test.yml
djabarovgeorge: For now, they will live in the consumer app, in this case it will be the API

packages/application-generic/src/services/auth/auth.service.ts
djabarovgeorge: Actually, I am not sure why this one happened it did not have circular dependency until I merged from next yesterday. which is very weird.

PR #4110, opened 2023-09-06
Add ES i18n translations

luisdnodrizat:

Add ES i18n messages

There is no traduction for some ES messages


ainouzgali (approved):
Awesome! thank you @luisdnodrizat🚀

PR #4117, opened 2023-09-06
feat(infra): implement memory db connection to inbound mail parse

p-fernandez:

Makes Inbound Mail app to use MemoryDB.
Adds test suite.

It was noticed by @djabarovgeorge that it was missing from the migration.


packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -31,14 +31,15 @@ import { PinoLogger } from '../../logging';
 import { AnalyticsService } from '../analytics.service';
 import { ApiException } from '../../utils/exceptions';
 import { Instrument } from '../../instrumentation';
+import { CreateUser, CreateUserCommand } from '../../usecases/create-user';
```
p-fernandez, line 34: We can't rely in the index exports of the folders because due our usage of App generic is what causes circular dependencies.

apps/inbound-mail/src/server/index.ts
```diff
@@ -51,16 +53,10 @@ class Mailin extends events.EventEmitter {
...
-  public start(options: any, callback: (err?) => void) {
+  public start(options: object, callback: (err?) => void) {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     const _this = this;
 
-    options = options || {};
-    if (_.isFunction(options)) {
-      callback = options;
-      options = {};
-    }
```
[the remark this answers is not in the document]
djabarovgeorge: Hard to say `we did not implement it` it is an npm package reused 😅.

packages/application-generic/src/services/in-memory-provider/providers/elasticache-cluster-provider.ts
```diff
@@ -52,7 +52,9 @@ export const getElasticacheClusterProviderConfig =
     };
 
     const host = redisClusterConfig.host;
-    const port = Number(redisClusterConfig.port);
+    const port = redisClusterConfig.port
```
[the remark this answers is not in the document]
djabarovgeorge: should we add a default port there or do we want it to fail if not provided?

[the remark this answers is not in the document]
p-fernandez: If no right configuration provided it will fallback to Redis Cluster configuration. If that configuration is not right nothing won't work and we report it in the logging.
Considering the Elasticache is our own implementation and the previous point I think we can not have a default. But open to suggestions.

apps/inbound-mail/src/.env.test
```diff
@@ -1,11 +1,23 @@
 NODE_ENV=test
 
-POST="25"
+PORT="2500"
```
djabarovgeorge, line 3: very strange that it fails on ports under 1024

p-fernandez: No, it is not strange. It is a Unix security feature for non-root processes.

.github/workflows/test.yml
```diff
@@ -296,7 +295,7 @@ jobs:
       - uses: ./.github/actions/setup-project
         with:
           # Don't run redis and etc... for other unit tests
-          slim: ${{ !contains(matrix.projectName, '@novu/api') && !contains(matrix.projectName, '@novu/worker') && !contains(matrix.projectName, '@novu/ws') }}
+          slim: ${{ !contains(matrix.projectName, '@novu/api') && !contains(matrix.projectName, '@novu/worker') && !contains(matrix.projectName, '@novu/ws') && !contains(matrix.projectName, '@novu/inbound-mail')}}
```
djabarovgeorge, line 298: ⚡️

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -19,8 +19,18 @@ export class QueueBaseService {
...
 
-  public createQueue(): void {
-    this.queue = this.instance.createQueue(this.topic, this.getQueueOptions());
+  public createQueue(overrideOptions?: QueueOptions): void {
+    const options = {
+      ...this.getQueueOptions(),
+      ...(overrideOptions && {
+        defaultJobOptions: {
+          ...this.getQueueOptions().defaultJobOptions,
+          ...overrideOptions.defaultJobOptions,
```
djabarovgeorge, line 28: Will this line will executed if overrideOptions is truthy?

djabarovgeorge (approved):
🚀

packages/application-generic/src/services/queues/queue-base.service.ts
p-fernandez: Yes, if `overrideOptions` is set it will be executed. Even if it is set as `{}` but with the spread we are safe nothing will override it.

.github/workflows/test.yml
p-fernandez: This was the reason of Redis not running. God bless whoever wrote the comment. 🙏🏻

PR #4120, opened 2023-09-07
Nv 2771 api e2e tests for the integration conditions

davidsoderberg:


apps/api/src/app/integrations/e2e/create-integration.e2e.ts
```diff
@@ -104,6 +104,21 @@ describe('Create Integration - /integration (POST)', function () {
...
   });
 
+  it('should create integration with conditions', async function () {
+    const payload = {
+      providerId: EmailProviderIdEnum.SendGrid,
+      channel: ChannelTypeEnum.EMAIL,
+      identifier: 'identifier-conditions',
+      active: false,
+      check: false,
+      conditions: [{}],
```
LetItRock, line 114: interesting that in this case, it doesn't validate the condition object

davidsoderberg: Yes it seems to ignore empty objects because we have the forbidUnknownValues set to false.

LetItRock (approved):
looks good! thanks! 🎉

PR #4121, opened 2023-09-07
Build submodule package by branch

djabarovgeorge:


.github/workflows/test.yml
```diff
@@ -248,6 +248,10 @@ jobs:
         with:
           submodules: true
           token: ${{ secrets.SUBMODULES_TOKEN }}
+        name: Checkout main on all submodules
+        run: |
+          git submodule foreach git checkout main
+          git submodule foreach git pull origin main
```
p-fernandez, line 254: Should we create a dedicated action to checkout with the submodules for all the workflows? Or will just be needed for the E2E enterprise tests? I feel like we might need to check it out for all of them but might miss any nuance that is not needed all the time.

djabarovgeorge: for not i am only testing here, and yes we will need it in couple of places

.github/workflows/reusable-api-e2e.yml
```diff
@@ -15,6 +15,11 @@ on:
         required: false
         default: false
         type: boolean
+      submodule_branch:
+        description: 'Submodule branch to checkout to'
+        required: false
+        default: 'main'
```
djabarovgeorge, line 21: * I could not create `submodule_branch` as required because `submodules` is not required.
* i added submodule_branch default as a guard but it should not be used IMO.

.github/workflows/reusable-inbound-mail-e2e.yml
```diff
@@ -33,14 +38,11 @@ jobs:
...
-        if: steps.setup.outputs.has_token == 'true'
+      - uses: ./.github/actions/checkout-submodules
         with:
-          submodules: ${{ inputs.submodules }}
-          token: ${{ secrets.SUBMODULES_TOKEN }}
-
-      # checkout without submodules
-      - uses: actions/checkout@v3
-        if: steps.setup.outputs.has_token != 'true'
+          run: ${{ steps.setup.outputs.has_token == 'true' }}
```
djabarovgeorge, line 43: Not sure why we use `has_token` if we have here input.submodule

.github/workflows/prod-deploy-ws.yml
djabarovgeorge, line 1: I added it because it was missing, Does anyone know if there were any reason for it?

.github/workflows/prod-deploy-inbound-mail.yml
djabarovgeorge, line 1: I added it because it was missing, Does anyone know if there were any reason for it?

.github/workflows/dev-deploy-api.yml
```diff
@@ -45,9 +46,11 @@ jobs:
         name: ['novu/api-ee', 'novu/api']
     steps:
       - uses: actions/checkout@v3
+      - uses: ./.github/actions/checkout-submodules
         with:
-          submodules: ${{ contains (matrix.name,'ee') }}
-          token: ${{ secrets.SUBMODULES_TOKEN }}
+          run: ${{ contains (matrix.name,'ee') }}
```
scopsy, line 51:
```suggestion
          enabled: ${{ contains (matrix.name,'ee') }}
```

Another option, but current also make sense!

.github/workflows/dev-deploy-inbound-mail.yml
```diff
@@ -26,6 +26,7 @@ jobs:
     with:
       ee: ${{ contains (matrix.name,'ee') }}
       submodules: ${{ contains (matrix.name,'ee') }}
+      submodule_branch: "next"
```
scopsy, line 29: Should we maybe use the `current` branch here in case we do a dev deploy from a branch in the future?

.github/workflows/prod-deploy-inbound-mail.yml
scopsy: Not that i'm aware of

.github/workflows/test.yml
```diff
@@ -245,9 +245,10 @@ jobs:
     steps:
       - run: echo ${{ matrix.projectName }}
       - uses: actions/checkout@v3
+      - uses: ./.github/actions/checkout-submodules
         with:
-          submodules: true
-          token: ${{ secrets.SUBMODULES_TOKEN }}
+          submodule_token: ${{ secrets.SUBMODULES_TOKEN }}
+          submodule_branch: "next"
```
scopsy, line 251: Should we try using the current branch here aswell?

scopsy (approved):

.github/workflows/dev-deploy-api.yml
djabarovgeorge: Actually i wanted to find another name because 'run' felt like a 'reserved word' for github actions.

.github/workflows/test.yml
djabarovgeorge: For the test i am not sure, because the name should be the exact name as in the enterprise repo which could be hard to maintain, and in addition It will break in the following case, if we have one PR in main repo and multiply branches in the enterprise repo (due the possible complexity)

.github/workflows/dev-deploy-inbound-mail.yml
djabarovgeorge: It makes sense here i will look into it.

djabarovgeorge: Now what i thought about it i am not sure, couple you provide a use case for such a case?

PR #4126, opened 2023-09-08
feat(ws): reduce the ws payload size

p-fernandez:

Only send the messageId in the payload of the event `RECEIVED` for the WebSocket queue.

The in-memory database gets fulled with messages of almost 1MB per message collapsing the database. We can leverage that by sending the messageId and just retrieve the full message if needed to pass on the WebSocket.


apps/ws/package.json
```diff
@@ -16,7 +16,7 @@
     "start:debug": "nest start --debug --watch",
     "start:prod": "node dist/src/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
-    "test": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true mocha --timeout 10000 --require ts-node/register --exit --file e2e/setup.ts src/**/**/*.spec.ts"
+    "test": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true mocha --timeout 10000 --require ts-node/register --exit --file e2e/setup.ts './src/**/*.spec.ts'"
```
p-fernandez, line 19: Tests beyond the second level of the folder were missed and not executed. The single quotes is the key here.

scopsy:
@p-fernandez not sure what happened here, but seems like there are more changes than should be here

p-fernandez:
> @p-fernandez not sure what happened here, but seems like there are more changes than should be here

I forgot to rebase from `next` the feature branch with both tasks of WebSockets improvements. Fixed.

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
```diff
@@ -28,7 +30,25 @@ export class ExternalServicesRoute {
       return;
     }
 
-    await this.wsGateway.sendMessage(command.userId, command.event, command.payload);
+    if (command.event === WebSocketEventEnum.RECEIVED) {
+      // TODO: Retro-compatibility for a bit just in case stalled messages
+      if (command.payload?.message) {
```
LetItRock, line 35: ❤️

LetItRock (approved):
🥇

PR #4130, opened 2023-09-10
fix: feed duplication

### LINKED ISSUE #4026: [NV-2593] Feed identifier is same when feed name is same
### What?

We have a create new feed API which has one payload `name` [https://docs.novu.co/api/create-feed/](https://docs.novu.co/api/create-feed/)

when we hit this API with same name it creates new feed with same identifier

### Why? (Context)

Since two feeds have same identifier, there will be an issue in inconsistent results items in-app messages in tab layout 

[https://docs.novu.co/notification-center/react/react-components#multiple-tab-layout](https://docs.novu.co/notification-center/react/react-components#multiple-tab-layout)

### Definition of Done

feedIdentifier should always be unique

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-2593](https://linear.app/novu/issue/NV-2593/feed-identifier-is-same-when-feed-name-is-same)</sub>

sr-26:

If two feeds have same identifier, there will be an issue in inconsistent results items in-app messages in tab layout.
Closes #4026


apps/api/src/app/feeds/usecases/create-feed/create-feed.usecase.ts
```diff
@@ -9,6 +9,15 @@ export class CreateFeed {
   constructor(private feedRepository: FeedRepository, private createChange: CreateChange) {}
 
   async execute(command: CreateFeedCommand): Promise<FeedEntity> {
+    const feedExist = await this.feedRepository.findOne({
+      _organizationId: command.organizationId,
```
scopsy, line 13: Feeds are per enviroment, sou would need to add here _environmentId instead of organizationId

scopsy (approved):

scopsy:
Thank you for your contribution @sr-26 !

PR #4140, opened 2023-09-11
Nv 2826 investigate failing digest tests

davidsoderberg:


packages/application-generic/src/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
```diff
@@ -69,6 +69,9 @@ export class DigestFilterSteps {
       const keys = path.split('.');
 
       for (const key of keys) {
+        if (result === undefined) {
```
davidsoderberg, line 72: Fix so if you have a nested digestkey but do not pass a nested value it will throw "can not find x on undefined"

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -756,10 +557,16 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
     });
 
     expect(digests.length).to.equal(5);
-    expect(digests[0].payload.postId).to.equal(undefined);
-    expect(digests[1].payload.postId).not.to.equal(digests[2].payload.postId);
-    expect(digests[3].payload.postId).to.equal(undefined);
-    expect(digests[1].payload.postId).to.equal(digests[4].payload.postId);
+    const noPostIdJobs = digests.filter((job) => !job.payload.postId);
```
scopsy, line 560: Not to rely on hardcoded values based on index which can be flaky

apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
```diff
@@ -495,7 +495,7 @@ describe('Topic Trigger Event', () => {
       }
     });
 
-    it('should not contain events from a different digestKey ', async () => {
+    it.only('should not contain events from a different digestKey ', async () => {
```
scopsy, line 498:
```suggestion
    it('should not contain events from a different digestKey ', async () => {
```

cc. @davidsoderberg

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -213,4 +213,40 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
...
+        },
+      }
+    );
+
+    if (childJob) {
+      updatedJobs.push(childJob);
+    }
+
+    while (childJob) {
+      childJob = await this.MongooseModel.findOneAndUpdate<JobEntity>(
```
scopsy, line 237: @davidsoderberg we should push those jobs to updatedJobs similarly to line 232-234

packages/application-generic/src/usecases/add-job/add-digest-job.usecase.ts
```diff
@@ -72,6 +73,23 @@ export class AddDigestJob {
...
+          $set: {
+            status: JobStatusEnum.MERGED,
+          },
+        }
+      );
+
+      await this.jobRepository.updateAllChildJobStatus(
+        job,
+        JobStatusEnum.MERGED
+      );
```
p-fernandez, line 91: This should happen before `digestMergedExecutionDetails` as it would be misleading that in Execution Details we have that update and for any reason these 2 calls failed.
Also have you considered keep these 2 updates together in a single functionality? Where we could have another functionality that would need update job status from a parent job, for a status different to merged?

packages/application-generic/src/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
p-fernandez: Great find!

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -858,7 +678,6 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
     expect(jobCount).to.equal(15);
   });
 
-  // TODO: Review backoff individually
   it.skip('should create multiple digest based on different digestKeys with backoff', async function () {
```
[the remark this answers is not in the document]
scopsy: We identified the race conditions on the backoff part functionality, due to the size of the change will address it later/

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -335,10 +333,10 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
     const firstDigestKeyBatch = messages.filter((message) => (message.content as string).includes('Hello world 2'));
     const secondDigestKeyBatch = messages.filter((message) => (message.content as string).includes('Hello world 1'));
 
-    expect(firstDigestKeyBatch.length).to.eql(2);
+    expect(firstDigestKeyBatch.length).to.eql(1);
```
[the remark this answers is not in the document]
scopsy: Yes, i think it's also the expected behaviour. Since only one message should be sent to the end user. Wdyt?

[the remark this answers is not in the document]
p-fernandez: I am just thinking in a potential situation where the digested messages content are different. In this test we are assuming that the content is the same but I think we would need to think in covering the scenario that I mention. Even if right now might not be totally possible as the content of the message is quite _static_.

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -102,7 +102,7 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
 
     expect(delayedJob).to.be.ok;
 
-    await session.awaitRunningJobs(template?._id, false, 1);
+    await session.awaitRunningJobs(template?._id, false, 0);
```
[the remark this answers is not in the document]
davidsoderberg: yes and now when they are changed to merged it should be 0 as changed to here.

p-fernandez (approved):

PR #4141, opened 2023-09-11
feat: skip ee tests if no token

djabarovgeorge:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -246,7 +246,7 @@ jobs:
     name: Test E2E EE
     runs-on: ubuntu-latest
     needs: [get-affected, check_submodule_token]
-    if:  ${{ fromJson(needs.get-affected.outputs.test-e2e-ee)[0] }} && needs.check_submodule_token.outputs.has_token == 'true'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -227,11 +227,26 @@ jobs:
...
+            echo "::set-output name=has_token::false"
+          fi
+  
   test_e2e_ee:
     name: Test E2E EE
     runs-on: ubuntu-latest
-    needs: [get-affected]
-    if:  ${{ fromJson(needs.get-affected.outputs.test-e2e-ee)[0] }}
+    needs: [get-affected, check_submodule_token]
+    if:  ${{ fromJson(needs.get-affected.outputs.test-e2e-ee)[0] }} && needs.check_submodule_token.outputs.has_token == 'true'
```
scopsy, line 249:
```suggestion
    if:  ${{ fromJson(needs.get-affected.outputs.test-e2e-ee)[0] && needs.check_submodule_token.outputs.has_token == 'true' }} 
```

If it's not working, you might need to try something like this

scopsy (approved):

PR #4142, opened 2023-09-11
feat(ws): performance improvements

p-fernandez:

Integrates two tasks to improve the performance of the WebSockets.

Performance improvements identified while migrating MemoryDB.

- [x] [nv-2823-ws-reduce-the-payload-size-of-the-jobs](https://github.com/novuhq/novu/pull/4126)
- [x] [nv-2824-websockets-simplify-the-job-events-created-when-sending-an](https://github.com/novuhq/novu/pull/4152)


LetItRock (commented):
do we need this one if you did changes here? 
https://github.com/novuhq/novu/pull/4126/files#r1319969641

p-fernandez:
> do we need this one if you did changes here? https://github.com/novuhq/novu/pull/4126/files#r1319969641

A change from `next` was needed to create the PR.

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
```diff
@@ -1,34 +1,52 @@
...
+        await this.processReceivedEvent(command);
+      }
+
+      if (command.event === WebSocketEventEnum.UNSEEN) {
+        await this.sendUnseenCountChange(command);
+      }
+
+      if (command.event === WebSocketEventEnum.UNREAD) {
+        await this.sendUnreadCountChange(command);
+      }
```
scopsy, line 29: Is this still needed or it's for backwards compatible version of a worker? Since if we now only send the `RECEIVED` those will never run?

p-fernandez: I mentioned it here: https://github.com/novuhq/novu/pull/4152#discussion_r1323359256
Headless and Notification Center still use those events. @LetItRock confirmed.

scopsy (approved):
Just a small question about the retro-competitive part, if it is for that than looks good🙏

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
scopsy: Very good Catch! Totally forgot about it 💪

PR #4144, opened 2023-09-12
Update the notification-template schema

djabarovgeorge:

We would like to introduce the new field in the NotificationStepEntity which will be a variants.

NotificationStepEntity {
 ...
  variants: Omit<NotificationStepEntity,'variants'>[]
}

### Definition of Done
* update the type
* update the schema


ainouzgali (approved):
❤️

libs/dal/src/repositories/notification-template/notification-template.schema.ts
```diff
@@ -158,6 +158,97 @@ const notificationTemplateSchema = new Schema<NotificationTemplateDBModel>(
             },
           },
         },
+        variants: [
+          {
```
LetItRock, line 162: can we have this object as a const and reuse it for step and variant?

LetItRock (approved):
please update the branch because there are some other changes here :)

PR #4147, opened 2023-09-12
feat: add custom data in email overrides

jainpawan21:
Added an option to send custom data in email overrides.
Currently added support for SendGrid only, it can be added on other providers if provider supports sending custom data


scopsy:
@djabarovgeorge could you assist? It's due to the removal of the enterprise ee in the lock file

LetItRock (approved):

PR #4152, opened 2023-09-12
feat(ws): simplify events for web sockets queue for performance

p-fernandez:

From the Worker we will only send one event instead of 3 and we will process it executing the functions that before were triggered by 3 different events, as they are sent at the same time.
We also only process the event and trigger the messages to the WebSocket if the user is online as if not it is a waste of resources.

It is a performance improvement identified during the MemoryDB migration.

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
@@ -247,7 +247,7 @@ export class SendMessageInApp extends SendMessageBase {
     );
 
     await this.webSocketsQueueService.bullMqService.add(
-      `sendMessage-received-${message._id}`,
+      message._id,
       {
         event: WebSocketEventEnum.RECEIVED,
         userId: command._subscriberId,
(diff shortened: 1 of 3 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -247,7 +247,7 @@ export class SendMessageInApp extends SendMessageBase {
     );
 
     await this.webSocketsQueueService.bullMqService.add(
-      `sendMessage-received-${message._id}`,
+      message._id,
```
p-fernandez, line 250: We will use the `messageId` as the `jobId` now we will have an unique event sent, therefore one unique job created.

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
```diff
@@ -12,42 +11,41 @@ const LOG_CONTEXT = 'ExternalServicesRoute';
 
 @Injectable()
 export class ExternalServicesRoute {
-  public readonly bullMqService: BullMqService;
+  constructor(private wsGateway: WSGateway, private messageRepository: MessageRepository) {}
 
-  constructor(private wsGateway: WSGateway, private messageRepository: MessageRepository) {
-    this.bullMqService = new BullMqService();
```
p-fernandez, line 18: This wasn't used. The dependency with BullMQ is attached in the WebSocketWorker in this same app.

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
```diff
@@ -12,42 +11,41 @@ const LOG_CONTEXT = 'ExternalServicesRoute';
...
-      if (command.payload?.message) {
-        Logger.verbose('Sending full message in the payload', LOG_CONTEXT);
-        await this.wsGateway.sendMessage(command.userId, command.event, command.payload);
+      if (command.event === WebSocketEventEnum.UNSEEN) {
+        await this.sendUnseenCountChange(command);
+      }
 
-        return;
+      if (command.event === WebSocketEventEnum.UNREAD) {
+        await this.sendUnreadCountChange(command);
```
p-fernandez, line 28: Headless sends this events to request the count changes, so keeping them.

LetItRock: The `headless` and `notification-center` actually are listening for these events. The events are triggered when the messages are marked as read/seen through the API or the `notification-center`.

LetItRock (approved):
an awesome improvement, thanks! 🙌

PR #4153, opened 2023-09-12
feat(provider): new provider zenvia

Fernando-Erd:

This PR adding a new provider called Zenvia. This provider is very popular in Brazil, Argentina and Mexico

New provider


Fernando-Erd:
@michaldziuba03 @jainpawan21 This PR is ready for review.

providers/zenvia/src/lib/zenvia.provider.ts
```diff
@@ -0,0 +1,157 @@
...
+  ): Promise<ISendMessageSuccessResponse> {
+    let response = {};
+
+    const contentParse = JSON.parse(options.content);
+
+    switch (contentParse.type_provider) {
+      case 'SMS':
+        response = this.sendSMS(options);
+        break;
+      case 'WHATSAPP':
```
michaldziuba03, line 34: To be honest for me it doesn't feel right - it kinda breaks responsibility of SMS channel.

packages/stateless/src/lib/provider/provider.interface.ts
```diff
@@ -28,7 +28,7 @@ export interface IEmailOptions {
 
 export interface ISmsOptions {
   to: string;
-  content: string;
+  content: any;
```
michaldziuba03, line 31: I don't agree with that change - it should be a string and is expected to be string by other providers.

providers/zenvia/src/lib/zenvia.provider.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  private async sendSMS(
+    options: ISmsOptions
+  ): Promise<ISendMessageSuccessResponse> {
+    const BASE_URL = 'https://api.zenvia.com/v2/channels/sms/messages';
+
+    const contentParse = JSON.parse(options.content);
+
+    const data: ZenviaParams = {
+      from: contentParse.from,
```
michaldziuba03, line 137: I think `from` should be configurable from provider's `options` - it what other providers do.

Example here: https://github.com/novuhq/novu/blob/f2b8089c84ff567d694934253152378a6731c61c/providers/burst-sms/src/lib/burst-sms.provider.ts#L35

providers/zenvia/src/lib/zenvia.provider.ts
```diff
@@ -0,0 +1,157 @@
...
+      date: new Date().toISOString(),
+    };
+  }
+
+  private async sendSMS(
+    options: ISmsOptions
+  ): Promise<ISendMessageSuccessResponse> {
+    const BASE_URL = 'https://api.zenvia.com/v2/channels/sms/messages';
+
+    const contentParse = JSON.parse(options.content);
```
michaldziuba03, line 134: `options.content` I believe should be a simple string, not parsable string JSON.

Provider should be configurable by `options`

packages/stateless/src/lib/provider/provider.interface.ts
michaldziuba03, line 31: I disagree with that change - content should be always a string and is expected to be a string by other providers.

michaldziuba03 (commented):
In my opinion you should change way how you configure behavior of provider - you should use provider's `options` and `options.content` should be a simple string instead parsable stringified JSON.

providers/zenvia/src/lib/zenvia.provider.ts
Fernando-Erd: So should I put WhatsApp sending as a chat category?

scopsy:
Closing for now due to inactivity 🙏

PR #4156, opened 2023-09-14
Nv 2788 allow applying filters for the digest delay steps

davidsoderberg:
Screenshot 2023-09-14 at 10 13 31
Screenshot 2023-09-14 at 10 14 10
Screenshot 2023-09-14 at 10 19 46
Screenshot 2023-09-14 at 10 22 59


packages/application-generic/src/usecases/add-job/add-job.command.ts
```diff
@@ -7,6 +7,9 @@ export class AddJobCommand extends EnvironmentWithUserCommand {
   @IsDefined()
   jobId: string;
 
+  @IsDefined()
```
davidsoderberg, line 10: this was always defined so changed to to be so in the command as well.

packages/application-generic/src/usecases/add-job/add-job.command.ts
```diff
@@ -7,6 +7,9 @@ export class AddJobCommand extends EnvironmentWithUserCommand {
...
+  job: JobEntity;
+
   @IsOptional()
-  job?: JobEntity;
+  filtered?: boolean;
```
davidsoderberg, line 14: This is the property used to set delay for digest and delay to zero. Name suggestions is more then welcome here :)

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -114,7 +113,7 @@ export class AddJob {
       })
     );
 
-    const delay = digestAmount ?? delayAmount;
+    const delay = command.filtered ? 0 : digestAmount ?? delayAmount;
```
davidsoderberg, line 116: If delay or digest step is filtered we should not delay the step but execute it directly.

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -59,8 +66,9 @@ export class MessageMatcher {
   ) {}
 
   public async filter(
-    command: SendMessageCommand,
-    variables: IFilterVariables
+    command: MessageMatcherCommand,
+    variables: IFilterVariables,
+    prefiltering = false
```
davidsoderberg, line 71: Should we call this `skipExecutionDetails` instead?

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
LetItRock, line 116: forgot to mention that we have to update some logs here in this file, like to provide the correct information about the delay

LetItRock (approved):
awesome! ⭐

PR #4158, opened 2023-09-14
refactor: refactored add digest job usecase

scopsy:


.cspell.json
```diff
@@ -503,6 +503,7 @@
     "springboot",
     "errmsg",
     "shelljs",
+    "compatability",
```
p-fernandez, line 506: It is better to fix it where it is happening:
https://github.com/novuhq/novu/blob/e54550062cc5fe2b48f5ba353ff7f2298a69174a/apps/api/e2e/compile-email-template.e2e.ts#L143
https://github.com/novuhq/novu/blob/e54550062cc5fe2b48f5ba353ff7f2298a69174a/apps/widget/public/iframeResizer.contentWindow.js#L872

apps/api/src/app/events/e2e/digest-events.e2e.ts
```diff
@@ -455,8 +456,7 @@ describe('Trigger event - Digest triggered events - /v1/events/trigger (POST)',
           metadata: {
             unit: DigestUnitEnum.SECONDS,
             amount: 1,
-            type: DigestTypeEnum.REGULAR,
-            backoff: true,
```
p-fernandez, line 459: Simplifying this is a great move.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -77,6 +77,8 @@ export class RunJob {
           job,
         })
       );
+
+      await this.jobRepository.updateStatus(job._environmentId, job._id, JobStatusEnum.COMPLETED);
```
p-fernandez, line 81: Why not using the use case `SetJobAsCompleted`?
```
await this.setJobAsCompleted.execute(
        SetJobAsCommand.create({
          environmentId,
          jobId,
          userId,
        })
      );
```
I get the reason to move to the the use case the completion flow instead of in the worker but if we are afraid of potential race conditions and want to directly hit the DAL we might just be masking a bit still the concurrency problem.
Unless you have discovered also that there is some flow in the `on('completed', ...)` path of BullMQ that is delaying executions.

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -174,27 +175,42 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
...
-    digestValue?: string | number
-  ): Promise<{ matched: number; modified: number }> {
-    const execution = {
-      matched: 0,
-      modified: 0,
-    };
+    digestValue?: string | number,
+    digestMeta?: IDigestRegularMetadata
+  ): Promise<DigestCreationResultEnum> {
+    const isBackoff = job.digest?.type === DigestTypeEnum.BACKOFF || (job.digest as IDigestRegularMetadata)?.backoff;
```
p-fernandez, line 181: We are keeping the fallback of the shortcut to handle the existing jobs in Redis when deploying this change? If so, well thought. 👏🏻

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -174,27 +175,42 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
...
-      status: JobStatusEnum.DELAYED,
-      type: StepTypeEnum.DIGEST,
-      _templateId: job._templateId,
-      _environmentId: this.convertStringToObjectId(job._environmentId),
-      _subscriberId: this.convertStringToObjectId(job._subscriberId),
-      ...(digestKey && { [`payload.${digestKey}`]: digestValue }),
-    });
+      /**
+       * In case of 2 triggers happened concurrently,
+       * we want ony one of those jobs to be skipped, while the second to be creating a digest.
```
p-fernandez, line 190:
```suggestion
       * we want only one of those jobs to be skipped, while the second to be creating a digest.
```

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -54,22 +62,47 @@ export class AddJob {
...
-    const digestAmount =
-      job.type === StepTypeEnum.DIGEST
-        ? await this.addDigestJob.execute(AddDigestJobCommand.create({ job }))
-        : undefined;
-
+    let digestAmount: number | undefined;
+    let digestCreationResult: DigestCreationResultEnum | undefined;
     if (job.type === StepTypeEnum.DIGEST) {
+      validateDigest(job);
+      digestAmount = this.calculateDelayService.calculateDelay({
```
p-fernandez, line 69: I consider this should live inside `MergeOrCreateDigestUsecase` and the output of that use case to be an object with:
- digestAmount
- digestCreationResult

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
```diff
@@ -54,22 +62,47 @@ export class AddJob {
...
+        await this.execute({
+          userId: job._userId,
+          environmentId: job._environmentId,
+          organizationId: command.organizationId,
+          jobId: nextJobToSchedule._id,
+          job: nextJobToSchedule,
+        });
+
+        return;
+      }
```
p-fernandez, line 105: Is this recursive iteration to try to jump all the extra time spent in different calculations in the flow to minimise the race conditions for jobs that are meant to be skipped because they are digested?
I find this bit of flow easy to miss conceptually and not sure how it would interact with `QueueNextJob` that is executed in `RunJob` and at the end of its execution calls to `AddJob`.

packages/application-generic/src/usecases/add-job/merge-or-create-digest.usecase.ts
```diff
@@ -0,0 +1,183 @@
...
+    const shouldDelayDigestJobOrMerge = async () =>
+      this.jobRepository.shouldDelayDigestJobOrMerge(
+        job,
+        digestKey,
+        digestValue,
+        digestMeta
+      );
+
+    const result =
+      await this.eventsDistributedLockService.applyLock<DigestCreationResultEnum>(
```
p-fernandez, line 160: Not sure if with the new logic flow we still need the Distributed Lock as we are handling the concurrency at DAL level by pessimistic locking. It would be great if we could finally get rid of this.

davidsoderberg: Me neither but let's remove it in another PR. So we can keep this to fix the tests we discussed in the planning meeting.

davidsoderberg: https://linear.app/novu/issue/NV-2870/investigate-if-distributed-lock-is-still-needed

.cspell.json
davidsoderberg: Yeah this fix came from it complaining about a comment that now is removed thu.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
davidsoderberg: I think we just missed that usecase, @scopsy do you remember if there was another reason for it?

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
davidsoderberg: Yes that is the reason. If I remember correctly me and @scopsy had a discussion about how it would interact with `QueueNextJob` and ended up with that it would be okay.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
davidsoderberg: I think MergeOrCreateDigestUsecase should only return the status for the step as it does now because it is now very easy to follow on a high level of what we are doing with a digest step.

scopsy: Yeah, that was also our consideration during the implementation. But it felt that the function was doing 2 distinguished actions. One, is to `MergeOrCreateDigestUsecase` which is as expected, and the second is to calculate the time. 

I can see a reason to return an object instead. Giving it a second thought, it might be better to return object and incapsulate the logic of the creation of digest to be responsible for the "How long" we should delay it too. 

@davidsoderberg I think the previous issue was that the response type was very confusing. And we acted on "null" or number for multiple states. Having an object return and an optional key for the digestAmount will handle it nicely.

wdyt?

p-fernandez (approved):
This a massive tech debt fix. And now Digest is fully reliable.
Brilliant work.
🌟

scopsy:
@davidsoderberg I'm a bit concerned that the 1-second digest window can cause flakiness in cases where the third job is executed in 1.1 seconds.

packages/application-generic/src/usecases/add-job/add-job.usecase.ts
davidsoderberg: I still think there is a reason to keep them separate to make it easy to understand what happens on add job

davidsoderberg:
> @davidsoderberg I'm a bit concerned that the 1-second digest window can cause flakiness in cases where the third job is executed in 1.1 seconds.

Should we raise the amount to 2-5 seconds?

scopsy:
@scopsy I say let's indeed do something like 2 seconds. And then let's create a ticket on linear to:

Ideally I would want to set the digest to something like 1 minute, and then be able to process it on demand immediately. Meaning that after we will trigger 5 triggers, I want to get the pending digest in bullmq, and process it right now instead of delaying it.

davidsoderberg:
@scopsy still one test that seems flakey.

scopsy:
I see those tests running in average 3-5 seconds. Do we want to try to leave it at 5? Until we will be doing the time manipulation

scopsy:
@davidsoderberg nope. Still same, let's do another session on this one to try and see the reasons for it to happen

apps/worker/src/app/workflow/usecases/send-message/digest/digest.usecase.ts
```diff
@@ -67,19 +67,29 @@ export class Digest extends SendMessageType {
...
+        _environmentId: currentJob._environmentId,
+        _subscriberId: command._subscriberId,
+      },
+      'payload'
+    );
+
+    return [currentJob.payload, ...jobs.map((job) => job.payload)];
+  }
+
+  private async getEvents(command: SendMessageCommand) {
```
scopsy, line 87: Would suggest to call this one:

```suggestion
  private async backwardCompatibleGetEvents(command: SendMessageCommand) {
```

And add a comment, and for the new one just calling getEvents

packages/application-generic/src/usecases/get-feature-flag/get-use-merged-digest-id.use-case.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@nestjs/common';
+import { FeatureFlagsKeysEnum } from '@novu/shared';
+import { FeatureFlagCommand } from './get-feature-flag.command';
+import { GetFeatureFlag } from './get-feature-flag.use-case';
+
+@Injectable()
+export class GetUseMergedDigestId extends GetFeatureFlag {
```
davidsoderberg, line 7: Not 100% happy with the naming here so suggestions is warmly welcome :)

PR #4160, opened 2023-09-15
Nv 2744 remove messages by transactionid

BiswaViraj:
- adds a new endpoint which allows user to delete messages using transactionId.

### DIFF
--- a/apps/api/src/app/messages/e2e/remove-messages-by-transactionId.e2e.ts
+++ b/apps/api/src/app/messages/e2e/remove-messages-by-transactionId.e2e.ts
@@ -1,0 +1,137 @@
+import { SubscribersService, UserSession } from '@novu/testing';
+import { MessageRepository, NotificationTemplateEntity, SubscriberEntity } from '@novu/dal';
+import { expect } from 'chai';
+import axios from 'axios';
+import { ChannelTypeEnum } from '@novu/shared';
+
+const axiosInstance = axios.create();
+
+describe('Delete Messages By TransactionId - /messages/?transactionId= (DELETE)', function () {
+  let session: UserSession;
+  const messageRepository = new MessageRepository();
+  let template: NotificationTemplateEntity;
+  let subscriber: SubscriberEntity;
+  let subscriberService: SubscribersService;
+
+  beforeEach(async () => {
+    session = new UserSession();
+    await session.initialize();
+    template = await session.createTemplate();
+    subscriberService = new SubscribersService(session.organization._id, session.environment._id);
+    subscriber = await subscriberService.createSubscriber();
+  });
+
+  it('should fail to delete non existing message', async function () {
+    const response = await session.testAgent.delete(`/v1/messages?transactionId=abc-1234`);
+
+    expect(response.statusCode).to.equal(404);
+    expect(response.body.error).to.equal('Not Found');
+  });
+
+  it('should delete messages by transactionId', async function () {
+    await axiosInstance.post(
+      `${session.serverUrl}/v1/subscribers`,
+      {
+        subscriberId: '123456',
+        name: 'broadcast subscriber',
+      },
+      {
+        headers: {
+          authorization: `ApiKey ${session.apiKey}`,
+        },
+      }
+    );
+
+    const response = await axiosInstance.post(
+      `${session.serverUrl}/v1/events/trigger/broadcast`,
+      {
+        name: template.triggers[0].identifier,
+        payload: {
+          email: 'new-test-email@gmail.com',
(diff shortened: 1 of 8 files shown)


apps/api/src/app/messages/messages.controller.ts
```diff
@@ -71,4 +78,26 @@ export class MessagesController {
       })
     );
   }
+
+  @Delete('')
```
scopsy, line 82:
```suggestion
  @Delete('/transaction/:transactionId')
```

Not sure regarding the current implementation, we either need to have a general remove endpoint with an optional filter, or a more specific endpoint for this.

@LetItRock wdyt here?

apps/api/src/app/messages/usecases/remove-messages-by-transactionId/remove-messages-by-transactionId.usecase.ts
```diff
@@ -0,0 +1,59 @@
...
+    );
+
+    if (subscriberIds.length === 0) {
+      throw new NotFoundException('Invalid transactionId or channel');
+    }
+
+    for (const subscriberId of subscriberIds) {
+      await this.invalidateCache.invalidateQuery({
+        key: buildFeedKey().invalidate({
+          subscriberId: subscriberId._subscriberId,
```
djabarovgeorge, line 29: I am not sure this one will work because we store by subscriberId and not _subscriberId, we would need either to fetch the subscriberId here or find another way to find it.

apps/api/src/app/messages/usecases/remove-messages-by-transactionId/remove-messages-by-transactionId.usecase.ts
```diff
@@ -0,0 +1,59 @@
...
+        ...(command.channel && { channel: command.channel }),
+      },
+      '_subscriberId'
+    );
+
+    if (subscriberIds.length === 0) {
+      throw new NotFoundException('Invalid transactionId or channel');
+    }
+
+    for (const subscriberId of subscriberIds) {
```
djabarovgeorge, line 26: Maybe it will be worth checking creating a pipe to redis in order to reduce the number of calls to redis here, in case the subscriber count is big.

BiswaViraj: Created a separate [ticket](https://linear.app/novu/issue/NV-2879/add-pipeline-support-to-handle-large-numbers-of-cache-invalidation) to handle large numbers of cache invalidation.

apps/api/src/app/messages/e2e/remove-messages-by-transactionId.e2e.ts
```diff
@@ -0,0 +1,137 @@
...
+    );
+
+    const result = await messageRepository.find({
+      transactionId: transactionId,
+      _environmentId: session.environment._id,
+      _organizationId: session.organization._id,
+      channel: ChannelTypeEnum.EMAIL,
+    });
+
+    expect(result.length).to.equal(0);
```
LetItRock, line 135: could you please update this test with a scenario:
- call a trigger for template with in-app and email
- check that in-app and email messages were sent
- delete by transaction and email channel
- check that the only email messages are deleted but in-app still present

Thank you! 🙌

apps/api/src/app/messages/usecases/remove-messages-by-transactionId/remove-messages-by-transactionId.usecase.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (command.channel) {
+      deleteQuery.channel = command.channel;
+    }
+
+    await this.messageRepository.deleteMany(deleteQuery);
+
+    return {
+      acknowledged: true,
+      status: 'deleted',
+    };
```
LetItRock, line 56: we should just return the 204

pnpm-lock.yaml
```diff
@@ -1162,84 +1162,6 @@ importers:
       stylelint-scss: 4.6.0_stylelint@15.10.1
       typescript: 4.9.5
 
-  enterprise/packages/auth:
```
LetItRock, line 1165: did you pull the changes from the `submodules` because it seems like you didn't and that's why this lock file was updated, [how to do ref](https://www.notion.so/novuhq/Git-Submodules-243ac16a181c4bb49d51241c63bc9226?pvs=4#0bc316fbe3474e53a9fa780af9feaa90)

LetItRock (approved):

PR #4166, opened 2023-09-18
chore: upgrade storybook

scopsy:

Upgrade storybook to latest version

Due to a vulnerable sub-dependency in the dependency tree.


scopsy:
@LetItRock although we not currently use it, there was a vulnerable sub dependency. So used the upgrade utility they provide to he help with the upgrade.

LetItRock (approved):
looks good! 🎉

PR #4168, opened 2023-09-19
feat: add support for sms overrides

### LINKED ISSUE #4030: [NV-2730] Add field overrides in SMS channel like we have in the Email
## What?

We can override almost all fields in the email channel

ref:- [https://docs.novu.co/channels/email/#sending-email-overrides](https://docs.novu.co/channels/email/#sending-email-overrides)

I see we are using `from` field from credentials here, we cannot override in SMS similar to email

[https://github.com/novuhq/novu/blob/next/apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts#L265C16-L265C16](https://github.com/novuhq/novu/blob/next/apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts#L265C16-L265C16)

## Definition of done:

1. Allow all field overrides in the SMS channel similar to email
2. Add types in node sdk
3. Check if Swagger API is reflecting the change

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-2730](https://linear.app/novu/issue/NV-2730/add-field-overrides-in-sms-channel-like-we-have-in-the-email)</sub>

mahendraHegde:

added support for overriding SMS

- Changed worker/send-sms-usecase to support overrides
- added types to node SDK
- updated relevant docs

closes #4030 

Screenshot 2023-09-18 at 9 01 22 PM


apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
```diff
@@ -272,9 +272,9 @@ export class SendMessageSms extends SendMessageBase {
       }
 
       const result = await smsHandler.send({
-        to: phone,
+        to: overrides.to || phone,
         from: integration.credentials.from,
```
scopsy, line 276: I think we can also allow overriding the from here

mahendraHegde: makes sense, added it

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
```diff
@@ -137,7 +137,7 @@ export class SendMessageSms extends SendMessageBase {
 
     await this.sendSelectedIntegrationExecution(command.job, integration);
 
-    const overrides = command.overrides[integration?.providerId] || {};
+    const overrides = command.overrides[integration?.channel] || {};
```
BiswaViraj, line 140: Why was this changed from `providerId` to `channel`?

mahendraHegde: we dont pass providerId in the overrides, we pass channel
ex:
```
{
 overrides:{sms:{...data}}
}
```
not
```
{
 overrides:{twillio:{...data}}
}
```
i'm not sure why it was written like that, but it was not being used until now, so it its safe to change it.

BiswaViraj: https://github.com/novuhq/novu/blob/e54550062cc5fe2b48f5ba353ff7f2298a69174a/packages/node/src/lib/events/events.interface.ts#L43

https://github.com/novuhq/novu/blob/e54550062cc5fe2b48f5ba353ff7f2298a69174a/apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts#L147

@mahendraHegde We do support providerId in the overrides for email, it would be great if we can enable the same for sms

mahendraHegde: @BiswaViraj  thanks for pointing that out, updated the PR

mahendraHegde:
@BiswaViraj  i have fixed the review comments

scopsy (approved):

PR #4175, opened 2023-09-19
feat: implement the logic that will determine which variant should be

djabarovgeorge:

We are introducing the variants and conditions to the workflow steps. Now we have to implement the logic that during the send phase will execute the variant conditions and will determine based on the trigger context provided which variant and content should be sent to the user as a notification.

Use-case: clients want to send different message contents to their tenant users.

#### Definition of Done

* during the send phase write the logic that will determine which variant of the content to send to the user
* e2e tests
* Send execution details with the chosen variant if possible.

### DIFF
--- a/apps/api/src/app/events/e2e/trigger-event.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event.e2e.ts
@@ -14,6 +14,7 @@ import {
   IntegrationRepository,
   ExecutionDetailsRepository,
   EnvironmentRepository,
+  TenantRepository,
 } from '@novu/dal';
 import { UserSession, SubscribersService } from '@novu/testing';
 import {
@@ -56,6 +57,7 @@ describe(`Trigger event - ${eventTriggerPath} (POST)`, function () {
   const jobRepository = new JobRepository();
   const executionDetailsRepository = new ExecutionDetailsRepository();
   const environmentRepository = new EnvironmentRepository();
+  const tenantRepository = new TenantRepository();
 
   describe(`Trigger Event - ${eventTriggerPath} (POST)`, function () {
     beforeEach(async () => {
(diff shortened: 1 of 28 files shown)


apps/api/src/app/events/e2e/trigger-event.e2e.ts
djabarovgeorge, line 1: Created small e2e just to make sure the logic is passing, in addition, we have usecase e2e tests.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -71,11 +74,17 @@ export class SendMessageChat extends SendMessageBase {
     Sentry.addBreadcrumb({
       message: 'Sending Chat',
     });
-    const chatChannel: NotificationStepEntity = command.step;
-    if (!chatChannel?.template) throw new PlatformException('Chat channel template not found');
+    const step: NotificationStepEntity = command.step;
```
djabarovgeorge, line 77: was not sure why to call it `chatChannel`

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -71,11 +74,17 @@ export class SendMessageChat extends SendMessageBase {
...
+    if (!step?.template) throw new PlatformException('Chat channel template not found');
 
     const tenant = await this.handleTenantExecution(command.job);
 
+    const template = await this.processVariants(command, tenant, subscriber, command.payload);
+
+    if (template) {
+      step.template = template;
+    }
```
djabarovgeorge, line 86: here we check if there is a variant, if it exists we will override the template.

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -141,22 +145,28 @@ export class SendMessageEmail extends SendMessageBase {
...
+    let subject = step?.template?.subject || '';
     let content;
 
     const payload = {
-      subject: emailChannel.template.subject || '',
-      preheader: emailChannel.template.preheader,
-      content: emailChannel.template.content,
-      layoutId: overrideLayoutId ?? emailChannel.template._layoutId,
-      contentType: emailChannel.template.contentType ? emailChannel.template.contentType : 'editor',
+      subject,
```
djabarovgeorge, line 165: Looks like we were saving an empty string as a subject, not sure if there was a reason for that.

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
djabarovgeorge, line 1: if we will get variants we log them in the execution details and return the template.

libs/testing/src/notification-template.service.ts
```diff
@@ -79,25 +79,60 @@ export class NotificationTemplateService {
...
-      const saved = await this.messageTemplateRepository.create({
+      const savedMessageTemplate = await this.messageTemplateRepository.create({
         type: message.type,
         cta: message.cta,
         variables: message.variables,
         content: message.content,
         subject: message.subject,
         title: message.title,
         name: message.name,
+        preheader: message.preheader,
```
djabarovgeorge, line 90: was missing

libs/testing/src/notification-template.service.ts
```diff
@@ -79,25 +79,60 @@ export class NotificationTemplateService {
...
         _feedId: override.noFeedId ? undefined : feeds[0]._id,
         _layoutId: override.noLayoutId ? undefined : layouts[0]._id,
         _creatorId: this.userId,
         _organizationId: this.organizationId,
         _environmentId: this.environmentId,
       });
 
-      if (saved?._id) {
+      const variantSteps: NotificationStepEntity[] = [];
```
djabarovgeorge, line 98: added logic to store variants as well.

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
djabarovgeorge, line 1: Only added here function in order to build a query and reuse it.

packages/application-generic/src/usecases/select-variant/select-variant.usecase.ts
```diff
@@ -0,0 +1,89 @@
...
+    messageTemplate: MessageTemplateEntity;
+    conditions?: ICondition[];
+  }> {
+    if (!command.step.variants?.length) {
+      return { messageTemplate: command.step.template };
+    }
+
+    if (!this.isFilterDataExist(command.filterData)) {
+      return { messageTemplate: command.step.template };
+    }
```
djabarovgeorge, line 31: If there are no variants, return the step template as chosen.

djabarovgeorge: same for the return value if all of the filters skiped

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
djabarovgeorge, line 1: here i basically copied the logic from MessageMatcher, i created a ticket in order to deprecate MessageMatcher so we will use only the new ConditionsFilter from application-generic. so we won't have duplicated code.

packages/application-generic/src/usecases/select-variant/select-variant.usecase.ts
```diff
@@ -0,0 +1,89 @@
...
+    return { messageTemplate: command.step.template };
+  }
+
+  private isFilterDataExist(filterData: IFilterVariables) {
+    return (
+      !!filterData.tenant ||
+      !!filterData.payload ||
+      !!filterData.subscriber ||
+      !!filterData.webhook
+    );
```
ainouzgali, line 87: But what if the condition is using tenant fields but this return true because it has subscriber data?

packages/application-generic/src/usecases/select-variant/select-variant.usecase.ts
ainouzgali, line 1: We also need to check that the variant is active as we can disable variants.

ainouzgali: Also, keep in mind that we need to do the same for digest and delay - and they are using the metadata which is not inside the template object

packages/application-generic/src/usecases/select-variant/select-variant.usecase.ts
djabarovgeorge: Should be totally fine that means that the filter will fail, but we could make this check more strict with some mapping i am just afraid it will make it more complex.

This one is a minor check in order to test if any data exists, initialy i created it when i thought it would be using only tenant data, so we could remove it as well.

packages/application-generic/src/usecases/select-variant/select-variant.usecase.ts
djabarovgeorge: > We also need to check that the variant is active as we can disable variants.

Good point i will look into it!


> Also, keep in mind that we need to do the same for digest and delay - and they are using the metadata that is not inside the template object

Are action nodes part of this scope?

ainouzgali:
@djabarovgeorge you would probably need to a[dd tenant here ](https://github.com/novuhq/novu/blob/7e418edf20388fc5b508d65d07678e137f6980df/apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts#L172)as well

PR #4178, opened 2023-09-20
Nv 2745 global channel subscriber preferences api only

BiswaViraj:

-  added subscribers global preference usecase
- added new API endpoints to subscribers and widgets controller to update global preferences
- Updated send message filter to consider global prefernces
- added hook to update global preferences to the NC
- added  method to the headless
- added method to the api-reference of headless docs
- added method to the node SDK
- added method example to node SDK documentation
- added e2e tests


docs/docs/notification-center/headless/api-reference.md
```diff
@@ -283,6 +283,56 @@ interface IUpdateUserPreferences {
...
+    result: UpdateResult<
+      IUserGlobalPreferenceSettings,
+      unknown,
+      IUpdateUserGlobalPreferencesVariables
+    >
+  ) => void;
+  onSuccess?: (settings: IUserGlobalPreferenceSettings) => void;
+  onError?: (error: unknown) => void;
+}
+```
```
LetItRock, line 334: we should probably include this in the new docs

packages/notification-center/src/hooks/useUpdateUserGlobalPreferences.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { useFetchUserGlobalPreferencesQueryKey } from './useFetchUserGlobalPreferencesQueryKey';
+import { useNovuContext } from './useNovuContext';
+
+interface IUpdateUserGlobalPreferencesVariables {
+  preferences: { channelType: string; enabled: boolean }[];
+  enabled?: boolean;
+}
+
+export const useUpdateUserGlobalPreferences = ({
```
LetItRock, line 13: I'm missing the hook and the controller code that allows to fetch global user preferences. I feel like we should return the global preferences together with template preferences, but for now, let's not show them in the NC. 
Then the `PreferenceLevelEnum.TEMPLATE` enum will be useful for filtering.

pnpm-lock.yaml
```diff
@@ -1162,84 +1162,6 @@ importers:
       stylelint-scss: 4.6.0_stylelint@15.10.1
       typescript: 4.9.5
 
-  enterprise/packages/auth:
```
LetItRock, line 1165: please update the lock file, this shouldn't be removed

apps/api/src/app/subscribers/usecases/update-subscriber-global-preferences/update-subscriber-global-preferences.usecase.ts
```diff
@@ -0,0 +1,105 @@
...
+    await this.subscriberPreferenceRepository.create({
+      _environmentId: command.environmentId,
+      _organizationId: command.organizationId,
+      _subscriberId: subscriber._id,
+      /*
+       * Unless explicitly set to false when creating a user preference we want it to be enabled
+       * even if not passing at first enabled to true.
+       */
+      enabled: command.enabled !== false,
+      channels: command.preferences && command.preferences.length > 0 ? channelObj : null,
```
djabarovgeorge, line 70:
```suggestion
      channels: channelObj && Object.keys(channelObj).length > 0 ? channelObj : null,
```

packages/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
```diff
@@ -0,0 +1,68 @@
...
+        _subscriberId: subscriber._id,
+        level: PreferenceLevelEnum.GLOBAL,
+      });
+
+    const subscriberChannelPreference = subscriberPreference?.channels;
+
+    return {
+      preference: {
+        enabled: subscriberPreference?.enabled ?? true,
+        channels: subscriberChannelPreference ?? {},
```
djabarovgeorge, line 44: I wonder if we could add here a default value like { email : true, ...} this way all of this use-case clients could consume this default and don't need to generate it themselves. (for example the addition logic like we have in SendMessage.stepPreferred)

libs/dal/src/repositories/subscriber-preference/subscriber-preference.entity.ts
```diff
@@ -13,14 +13,21 @@ export class SubscriberPreferenceEntity {
...
 }
 
 export type SubscriberPreferenceDBModel = ChangePropsValueType<
   SubscriberPreferenceEntity,
   '_environmentId' | '_organizationId' | '_subscriberId' | '_templateId'
 >;
+
+export enum PreferenceLevelEnum {
+  GLOBAL = 'GLOBAL',
+  TEMPLATE = 'TEMPLATE',
```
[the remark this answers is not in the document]
djabarovgeorge: Tend to agree, because today we could say if level is undefined it is most definitely PreferenceLevelEnum.TEMPLATE and hack it, but if we add in the future another level we will never know what the undefined means.

apps/api/migrations/subscriber-preferences-level/subscriber-preferences-level.migration.ts
```diff
@@ -0,0 +1,33 @@
...
+  const subscriberPreferenceRepository = app.get(SubscriberPreferenceRepository);
+  // eslint-disable-next-line no-console
+  console.log('add level: PreferenceLevelEnum.TEMPLATE to all subscriber preferences without level property');
+
+  await subscriberPreferenceRepository._model.collection.updateMany(
+    { level: { $exists: false }, _templateId: { $exists: true } },
+    {
+      $set: { level: PreferenceLevelEnum.TEMPLATE },
+    }
+  );
```
BiswaViraj, line 24: currently all the preference have templateId and will be converted to level: template

LetItRock (approved):

PR #4180, opened 2023-09-20
Add inbound mail Queue to metrics

Cliftonz:
This adds reply to emails to the queue metrics, it will also allow us to see how often the inbound mail is being used.

### DIFF
--- a/packages/application-generic/src/services/queues/index.ts
+++ b/packages/application-generic/src/services/queues/index.ts
@@ -13,7 +13,7 @@ export {
   QueueBaseService,
   ActiveJobsMetricQueueService,
   CompletedJobsMetricQueueService,
-  InboundParseQueueService as InboundParseQueue,
+  InboundParseQueueService,
   StandardQueueService,
   WebSocketsQueueService,
   WorkflowQueueService,
--- a/apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
+++ b/apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
@@ -1,6 +1,6 @@
 import {
   getInboundParseMailWorkerOptions,
-  InboundParseQueue,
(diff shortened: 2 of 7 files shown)


packages/application-generic/src/services/queues/index.ts
```diff
@@ -15,4 +15,5 @@ export {
   StandardQueueService,
   WebSocketsQueueService,
   WorkflowQueueService,
+  InboundParseQueueService,
```
p-fernandez, line 18: Better name it in the import than export it twice with different names.

p-fernandez (changes requested):

packages/application-generic/src/services/queues/index.ts
Cliftonz: What do you mean?

p-fernandez: With your addition in `[packages/application-generic/src/services/queues/index.ts](https://github.com/novuhq/novu/pull/4180/files/195c10a8a44705b827d0b6d130cad297d6c54b2e#diff-47bdb9862f6d33518064bef8765388edcd2d7182861a8f09a0edc6dba3b3edbd)` we will be exporting twice the same class, as `InboundParseQueueService` and as `InboundParseQueue` (this was done like this because some name clashing in the Inbound Parse Mail app.

So that's why I said that in `packages/application-generic/src/custom-providers/index.ts` when importing in line 13 you should do:
```
InboundParseQueue as InboundParseQueueService
```.

Another options would be to export the original name (what you tried) and rewrite the import in Inbound Parse Mail app. Any direction but export twice the same class with 2 different names.

p-fernandez:
I see the re-review request but I can't see the changes I requested. Could it be you forget to push?

scopsy:
Closing for now due to inactivity.

p-fernandez (approved):

Issue #3957, opened 2023-08-13
[NV-2697] senderName field is missing in resend api and nodejs sdk


jainpawan21:
### What?

Novu allows using sender name field in email editor and via overrides. If sender name is Pawan Jain and sender from email is pawan@novu.co, in email clients shows like this `Pawan Jain <pawan@novu.co>`

### Why? (Context)

In the case of resend provider, it does not have a separate sender name field.

Ref API:- [https://resend.com/docs/api-reference/emails/send-email](https://resend.com/docs/api-reference/emails/send-email)

currently, there are two ways to handle this:- 

1. use **from** field like this in integration modal

   
![image](https://github.com/novuhq/novu/assets/39362422/9de9ac85-cd81-4248-9227-bdbe883b03b7)

2. or use overrides like this 
       `"overrides": {`
   `        "email":{`
   `            "from": "Pawan Jain Novu <pawan@novu.co>"`
   `        }`
   `    }`

Users get confused about why the sender's name is not showing up correctly.

### Definition of Done

1. Novu should automatically handle this at provider implementation
2. Update docs accordingly

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-2697](https://linear.app/novu/issue/NV-2697/sendername-field-is-missing-in-resend-api-and-nodejs-sdk)</sub>

Sangam-ghimire:
Hey @jainpawan21 @scopsy I would love to help here...seems pretty interesting

scopsy:
@Sangam-ghimire assigned you 🙏

Sangam-ghimire:
@scopsy  @jainpawan21 I am having trouble setting up the project locally, I have asked it in the discord channel as well.
![image](https://github.com/novuhq/novu/assets/90495874/0cd67144-cbc4-40ee-9a1b-b8d7741e4648)
I am stuck here, and the error is not getting resolved for now.

scopsy:
@Sangam-ghimire could you try running `npm run setup:project` from project root?

Sangam-ghimire:
> @Sangam-ghimire could you try running `npm run setup:project` from project root?

I have done that and whats next? is it optional to setup enviroment from npm start  after npm run project:setup?

Sangam-ghimire:
> > @Sangam-ghimire could you try running `npm run setup:project` from project root?
> 
> I have done that and whats next? is it optional to setup enviroment from npm start  after npm run project:setup?

[...]

PR #4186, opened 2023-09-21
NV-2697 - senderName field is missing in resend api and nodejs sdk

gitstart:
This fix ensures field senderName is part of the payload for the resendEmail API

### Description

Novu allows the use of the sender name field in the email editor and via overrides. If the sender name is Pawan Jain and the sender from email is [pawan@novu.co](mailto:pawan@novu.co), in email clients shows like this `Pawan Jain <pawan@novu.co>`

Resend API Doc: https://resend.com/docs/api-reference/emails/send-email

closes https://github.com/novuhq/novu/issues/3957

### Other information

---

### DIFF
--- a/providers/resend/src/lib/resend.provider.ts
+++ b/providers/resend/src/lib/resend.provider.ts
@@ -17,6 +17,7 @@ export class ResendEmailProvider implements IEmailProvider {
     private config: {
       apiKey: string;
       from: string;
+      senderName: string;
     }
   ) {
     this.resendClient = new Resend(this.config.apiKey);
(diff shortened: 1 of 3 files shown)


gitstart:
@scopsy @jainpawan21 This PR is ready for review.

providers/resend/src/lib/resend.provider.ts
```diff
@@ -17,6 +17,7 @@ export class ResendEmailProvider implements IEmailProvider {
     private config: {
       apiKey: string;
       from: string;
+      senderName: string;
```
p-fernandez, line 20: Sender name is an optional property of CredentialsDTO. In the implementation provided that is not taken in account and it is assumed it will always be present. And that's not right.

p-fernandez (changes requested):

scopsy:
Hi @gitstart ! 👋 thanks for the contribution 🤝
For now, I'm closing it.

PR #4184, opened 2023-09-21
feat: allow variables in sender name field of email editor

mahendraHegde:
### What change does this PR introduce?

- pass senderName variable to compile template to parse the variable
- use parsed sender name as sender name to email provider

### Why was this change needed?

closes #3958 

### Other information (Screenshots)
Screenshot 2023-09-20 at 10 44 44 PM

Screenshot 2023-09-20 at 10 50 14 PM
Screenshot 2023-09-20 at 10 51 05 PM

PR #4187, opened 2023-09-21
Stacked PR - new variants & condition support

djabarovgeorge:


packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -1,27 +1,77 @@
...
+import { CachedEntity } from '../../services/cache/interceptors/cached-entity.interceptor';
+import { buildSubscriberKey } from '../../services/cache/key-builders/entities';
 
 @Injectable()
 export class ConditionsFilter extends Filter {
-  public async filter(
-    command: ConditionsFilterCommand,
-    variables: IFilterVariables
-  ): Promise<{
+  constructor(
```
p-fernandez, line 60: This is one of my main concerns and what I tried to tackle in my PR (https://github.com/novuhq/novu/pull/4702).
Most of the logic here inside should be a service and try to make it reusable. This neither follows the Execution pattern we have implemented in other use cases.

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -136,49 +131,33 @@ export class SendMessageEmail extends SendMessageBase {
...
       {},
       command.overrides.email || {},
       command.overrides[integration?.providerId] || {}
     );
 
     let html;
-    let subject = '';
+    let subject = step?.template?.subject || '';
     let content;
-    let senderName = overrides?.senderName || emailChannel.template.senderName;
```
ainouzgali, line 162: [The sender name allowing variables](https://github.com/novuhq/novu/pull/4184) has somehow been removed in this file. Probably a mix up when merging conflicts

LetItRock: @ainouzgali the variable was removed, but instead `overrides?.senderName || emailChannel.template.senderName` is passed to the `sendMessage` function in the line 288

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -105,42 +504,197 @@ export class ConditionsFilter extends Filter {
...
+
+  private async fetchTenantIfMissing(
+    command: ConditionsFilterCommand,
+    filters: IMessageFilter[]
+  ): Promise<TenantEntity | undefined> {
+    if (command.variables?.tenant) {
+      return command.variables.tenant;
+    }
+
+    const tenantIdentifier = command.job?.tenant?.identifier;
```
djabarovgeorge, line 617: can we add here condition in case `command.job?.tenant` is the identifier? 

```suggestion
    const tenantIdentifier = typeof command.job?.tenant === "string" ? command.job?.tenant : command.job?.tenant?.identifier;
```

PR #4188, opened 2023-09-21
Variants polishing

djabarovgeorge:

### DIFF
--- a/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
+++ b/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
@@ -43,6 +43,9 @@ export class CreateNotificationTemplate {
   async execute(usecaseCommand: CreateNotificationTemplateCommand) {
     const blueprintCommand = await this.processBlueprint(usecaseCommand);
     const command = blueprintCommand ?? usecaseCommand;
+
+    this.validatePayload(command);
+
     const triggerIdentifier = `${slugify(command.name, {
       lower: true,
       strict: true,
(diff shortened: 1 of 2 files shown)


apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
```diff
@@ -61,6 +64,16 @@ export class CreateNotificationTemplate {
     return storedWorkflow;
   }
 
+  private validatePayload(command: CreateNotificationTemplateCommand) {
+    const variants = command.steps ? command.steps?.flatMap((step) => step.variants || []) : [];
+
+    for (const variant of variants) {
+      if (!variant.filters?.length) {
+        throw new ApiException(`Variant filters are required, variant name ${variant.name} id ${variant._id}`);
```
ainouzgali, line 72:
```suggestion
        throw new ApiException(`Variant conditions are required, variant name ${variant.name} id ${variant._id}`);
```

ainouzgali (commented):
I wonder if we have a way to make filters mandatory for variants already in the [command level](https://github.com/novuhq/novu/blob/60288060e631c7a220c6dccf049432390b63cead/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts#L121)

djabarovgeorge:
> I wonder if we have a way to make filters mandatory for variants already in the [command level](https://github.com/novuhq/novu/blob/60288060e631c7a220c6dccf049432390b63cead/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts#L121)

It is a great idea but unfortunately, the scope of the request will be too big for this PR, in order to implement it we will need to do the following: Create a custom validation decorator, fix the nested validation that does not work at the moment, and the hardest part which will be to return the nested errors that we will get, therefore I will postpone it to another dedicated PR.

djabarovgeorge:
> > I wonder if we have a way to make filters mandatory for variants already in the [command level](https://github.com/novuhq/novu/blob/60288060e631c7a220c6dccf049432390b63cead/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts#L121)
> 
> It is a great idea but unfortunately, the scope of the request will be too big for this PR, in order to implement it we will need to do the following: Create a custom validation decorator, fix the nested validation that does not work at the moment, and the hardest part which will be to return the nested errors that we will get, therefore I will postpone it to another dedicated PR.

Created https://github.com/novuhq/novu/pull/4306

PR #4193, opened 2023-09-22
fix(notification-center): when the ws unread event comes in refetch the unseen messages count

LetItRock:

Issue: when all messages are marked as a read we do send the `unread_count_changed` WS event but the NC doesn't react to it.

Now: I've added the hook `useUnreadCount` that is listening on the `unread_count_changed` WS event and refreshes the notifications, unseen count, and feeds unseen count.

Covered also with tests.

https://github.com/novuhq/novu/assets/2607232/115f4654-b840-4043-9948-0addcde95ede

### DIFF
--- a/packages/notification-center/src/hooks/useUnreadCount.ts
+++ b/packages/notification-center/src/hooks/useUnreadCount.ts
@@ -10,9 +10,10 @@ import { useFetchNotificationsQueryKey } from './useFetchNotificationsQueryKey';
 import { useUnseenCountQueryKey } from './useUnseenCountQueryKey';
 import { useDataRef } from './useDataRef';
 import { FEED_UNSEEN_COUNT_QUERY_KEY } from './queryKeys';
+import { useUnreadCountQueryKey } from './useUnreadCountQueryKey';
 
-const dispatchUnseenCountEvent = (count: number) => {
-  document.dispatchEvent(new CustomEvent('novu:unseen_count_changed', { detail: count }));
+const dispatchUnreadCountEvent = (count: number) => {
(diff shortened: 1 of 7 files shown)


packages/notification-center/src/hooks/useUnreadCount.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * TODO: This is a temporary fix
+ * Cypress is slow, so we need to increase the debounce time
+ * This is happening in a very high thruput of updates in testing env.
+ *
+ * Can also happen in real scenarios, so we need to review how we handle concurrency in the future
+ */
+const DEBOUNCE_TIME = typeof window !== 'undefined' && (window as any)?.Cypress ? 1000 : 100;
+
+export const useUnreadCount = ({ onSuccess, ...restOptions }: UseQueryOptions<ICountData, Error, ICountData> = {}) => {
```
LetItRock, line 28: this hook is basically the copy of the `useUnseen` hook, but with few differences:
- listens to the `unread_count_changed` ws event
- when event comes in it refetches also the unseen count

packages/notification-center/src/hooks/useUnseenCount.ts
```diff
@@ -46,9 +46,6 @@ export const useUnseenCount = ({ onSuccess, ...restOptions }: UseQueryOptions<IC
             count: data?.unseenCount ?? oldData.count,
           }));
 
-          queryClient.refetchQueries(queryKeysRef.current.unseenCountQueryKey, {
-            exact: false,
-          });
```
LetItRock, line 51: the query is updated above, no need to do the refetch

BiswaViraj (approved):

PR #4195, opened 2023-09-24
refactor: use conditions filter

djabarovgeorge:

deprecated the use of MessageMatcher usecase and switch to ConditionsFilter usecase

In the past we used MessageMatcher in the worked however today we ConditionsFilter in application-generic, initially made for integration but today the code is pretty much the same and they bear the same responsibility.

IMO there is no reason to have them both.


apps/api/src/app/events/e2e/trigger-event.e2e.ts
djabarovgeorge, line 1: Now we have the execution detail of the filter action.

apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts
djabarovgeorge, line 1: On create workflow we use Partial<INotificationTemplateStep>  where we have IMessageFilter with optional `isNegated` and `type` therefore I aligned those types to be optional as well.

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.spec.ts
djabarovgeorge, line 1: refactored to use Conflict Filter use case, we still need to change this spec name and delete the MessageMatcher use-case. 

This spec will remain in the worker project because we still do not run the application generic tests.

libs/dal/src/repositories/notification-template/notification-template.entity.ts
djabarovgeorge, line 1: Same as on the comment above:

On create workflow we use Partial where we have IMessageFilter with optional `isNegated` and `type` therefore I aligned those types to be optional as well.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.command.ts
```diff
@@ -1,13 +1,18 @@
...
+
 export class ConditionsFilterCommand extends EnvironmentWithUserCommand {
   @IsDefined()
   filters: StepFilter[];
 
   job?: IJob;
 
   step?: INotificationTemplateStep;
+
+  variables?: IFilterVariables;
```
djabarovgeorge, line 17: We allow the passing of the data if exists. if not, we will fetch it in the use case if needed.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -37,31 +39,38 @@ import {
...
     conditions: ICondition[];
+    variables: IFilterVariables;
   }> {
-    const { filters } = command;
+    const variables = await this.normalizeVariablesData(command);
+    const filters = command.filters?.length
+      ? command.filters
+      : command.step?.filters?.length
+      ? command.step.filters
+      : [];
```
djabarovgeorge, line 68: Not all of the clients have the job entity with filters (like select integrations), therefore we allow passing filters separately with priority.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -214,9 +244,8 @@ export class ConditionsFilter extends Filter {
     command: ConditionsFilterCommand,
     filterProcessingDetails: FilterProcessingDetails
   ): Promise<boolean> {
-    const subscriber = await this.subscriberRepository.findOne({
-      _id: command.job.subscriberId,
```
djabarovgeorge, line 218: IMO this one is a bug, because Subscriber._id is not equal to job.subscriberId.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -114,10 +123,31 @@ export class ConditionsFilter extends Filter {
 
     return {
       passed: !!foundFilter,
-      conditions: conditions,
+      conditions,
+      variables,
     };
   }
 
+  public static sumFilters(
```
djabarovgeorge, line 131: copied from message matcher for our use.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -448,11 +477,32 @@ export class ConditionsFilter extends Filter {
...
+  private splitFilters(filter: StepFilter) {
+    const webhookFilters = filter.children.filter(
+      (childFilter) => childFilter.on === 'webhook'
+    );
+
+    const otherFilters = filter.children.filter(
+      (childFilter) => childFilter.on !== 'webhook'
+    );
+
+    return { webhookFilters, otherFilters };
```
djabarovgeorge, line 505: was missing from the message matcher usecase

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -461,11 +511,98 @@ export class ConditionsFilter extends Filter {
...
+    }
 
-    return !!foundFilter;
+    return !!(await this.findAsync(webhookFilters, (i) =>
+      this.processFilter(variables, i, command, filterProcessingDetails)
+    ));
+  }
+
+  @Instrument()
+  private async normalizeVariablesData(command: ConditionsFilterCommand) {
```
djabarovgeorge, line 529: here we fetch the variable is they are needed and missing from the command

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
djabarovgeorge, line 1: refactor to use the updated command

packages/application-generic/src/usecases/select-variant/select-variant.usecase.ts
djabarovgeorge, line 1: refactor to use the updated command

apps/worker/src/app/workflow/specs/conditions-filter.usecase.spec.ts
djabarovgeorge, line 1: was not sure how to call this folder 😅

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -1,525 +1,522 @@
...
-        const noRules = !children || (Array.isArray(children) && children.length === 0);
-        if (noRules) {
-          return true;
-        }
-
-        const singleRule = !children || (Array.isArray(children) && children.length === 1);
-        if (singleRule) {
-          const result = await this.processFilter(variables, children[0], command, filterProcessingDetails);
-          if (!prefiltering) {
-            await this.createExecutionDetails.execute(
```
davidsoderberg, line 102: I can not find this detail somewhere in the new code.

djabarovgeorge: Are you referring to the createExecutionDetails execution or `if (singleRule)` block?

davidsoderberg: to the `createExecutionDetails `

djabarovgeorge: we will still log the execution detail in `send-message.usecase.ts` if shouldRun.passed is false, and we won't print in queue-next-job.usecase.ts the same as it was before because of the flag. or did u miss something? 

for the moment i remove the responsibility of the createExecutionDetails execution from the filter, do you think we should return it? 

@davidsoderberg

davidsoderberg: But the one of this type is not used anymore?

djabarovgeorge: Not sure I understood what you mean here

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -461,11 +511,98 @@ export class ConditionsFilter extends Filter {
...
+  private async fetchSubscriberIfMissing(
+    command: ConditionsFilterCommand
+  ): Promise<SubscriberEntity | undefined> {
+    if (command.variables?.subscriber) {
+      return command.variables.subscriber;
+    }
+
+    const subscriberFilterExist = command.step?.filters?.find((filter) => {
+      return filter?.children?.find((item) => item?.on === 'subscriber');
+    });
```
ainouzgali, line 550: @djabarovgeorge what would happen here if there is a condition on subscriber in a variant? Or in the future with integration?
Could we make this more general?

djabarovgeorge: This state is not possible because we made select variant command with a mandatory variable, however, due to the reasoning we could have other clients of condition filters in the future i added the addition logit in order to fetch the possible condition from variants.

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
djabarovgeorge: @davidsoderberg i will merge this one for now, please let me know if i missed something here 🙏

PR #4200, opened 2023-09-25
feat(worker): refactor get digest events flow

p-fernandez:

Refactors the logic flow in getting the digest events for message of type digest.

During the works to provide stability to the Digest feature performed by @scopsy and @davidsoderberg it was noticed that we needed improvements in the send message codebase.
I noticed in the Digest Events flow that unneeded calls to the database were made and we could avoid that with a simple refactor, besides reusing functionality for regular and backoff digest cases.

This PR shouldn't conflict with https://github.com/novuhq/novu/pull/4158 but let's keep an eye in both and prioritise the linked one as it is more crucial for our needs.

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
@@ -1,22 +1,18 @@
 import { Injectable } from '@nestjs/common';
 import { JobStatusEnum } from '@novu/dal';
-import { IDigestRegularMetadata, StepTypeEnum } from '@novu/shared';
-import { DigestFilterSteps, InstrumentUsecase } from '@novu/application-generic';
+import { StepTypeEnum } from '@novu/shared';
+import { InstrumentUsecase } from '@novu/application-generic';
 
-import { SendMessageCommand } from '../send-message.command';
+import { DigestEventsCommand } from './digest-events.command';
 import { GetDigestEvents } from './get-digest-events.usecase';
-import { PlatformException } from '../../../../shared/utils';
 
 @Injectable()
 export class GetDigestEventsBackoff extends GetDigestEvents {
   @InstrumentUsecase()
-  public async execute(command: SendMessageCommand) {
-    const currentJob = await this.jobRepository.findOne({ _environmentId: command.environmentId, _id: command.jobId });
-    if (!currentJob) throw new PlatformException('Digest job is not found');
+  public async execute(command: DigestEventsCommand) {
+    const currentJob = command.currentJob;
 
-    const digestMeta = currentJob.digest as IDigestRegularMetadata | undefined;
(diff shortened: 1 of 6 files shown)


apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
```diff
@@ -1,22 +1,18 @@
...
+import { DigestEventsCommand } from './digest-events.command';
 import { GetDigestEvents } from './get-digest-events.usecase';
-import { PlatformException } from '../../../../shared/utils';
 
 @Injectable()
 export class GetDigestEventsBackoff extends GetDigestEvents {
   @InstrumentUsecase()
-  public async execute(command: SendMessageCommand) {
-    const currentJob = await this.jobRepository.findOne({ _environmentId: command.environmentId, _id: command.jobId });
-    if (!currentJob) throw new PlatformException('Digest job is not found');
```
p-fernandez, line 15: We were making the same call twice in a row and in both digest types so this is clearly unneeded.

apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-regular.usecase.ts
```diff
@@ -1,47 +1,42 @@
...
+        }
+      : {};
+    const earliest = sub(new Date(currentJob.createdAt), subtractedTime);
 
     const jobs = await this.jobRepository.findJobsToDigest(
       earliest,
       currentJob._templateId,
-      command.environmentId,
-      // backward compatibility - ternary needed to be removed once the queue renewed
-      command._subscriberId ? command._subscriberId : command.subscriberId,
```
p-fernandez, line 40: With the refactor we are able to move this backward compatibility check to the main use case of Digest and just do it once, decreasing the chances of missing it as it was spread.

apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
```diff
@@ -25,12 +21,11 @@ export class GetDigestEventsBackoff extends GetDigestEvents {
       _templateId: currentJob._templateId,
       status: JobStatusEnum.COMPLETED,
       type: StepTypeEnum.TRIGGER,
-      _environmentId: command.environmentId,
+      _environmentId: currentJob._environmentId,
       ...(digestKey && { [`payload.${digestKey}`]: digestValue }),
-      // backward compatibility - ternary needed to be removed once the queue renewed
```
scopsy, line 30: 👌

apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events-backoff.usecase.ts
scopsy: Bravo!

scopsy (approved):
Amazing work, as always! 🎉

PR #4202, opened 2023-09-25
feat(api): create trigger event status enum

p-fernandez:

Creates an enum to gather all the potential status of the result of executing the trigger event endpoints.

### DIFF
--- a/apps/api/src/app/events/dtos/trigger-event-response.dto.ts
+++ b/apps/api/src/app/events/dtos/trigger-event-response.dto.ts
@@ -1,4 +1,5 @@
 import { IsBoolean, IsDefined, IsString } from 'class-validator';
+import { TriggerEventStatusEnum } from '@novu/shared';
 import { ApiProperty } from '@nestjs/swagger';
 
 export class TriggerEventResponseDto {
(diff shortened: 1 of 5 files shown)


apps/api/src/app/events/dtos/trigger-event-response.dto.ts
```diff
@@ -11,7 +12,7 @@ export class TriggerEventResponseDto {
 
   @ApiProperty({
     description: 'Status for trigger',
-    enum: ['processed', 'trigger_not_active', 'subscriber_id_missing', 'error'],
+    enum: TriggerEventStatusEnum,
   })
   @IsString()
   @IsDefined()
```
BiswaViraj, line 18: should we also update the `status: string;` to `status: TriggerEventStatusEnum;` and remove the `@IsString()`?

BiswaViraj (approved):
🚀

apps/api/src/app/events/dtos/trigger-event-response.dto.ts
p-fernandez: Good catch!

PR #4205, opened 2023-09-26
refactor: map subscribers in worker part

djabarovgeorge:

In case we have a big amount of subscribers/big topics and we loop (and await) through all of them it can take a lot of time.

remove redundant mapping in the trigger flow, and move it to the worker part

recreation of https://github.com/novuhq/novu/pull/4023


packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -273,29 +292,49 @@ export class TriggerEvent {
...
-            );
-            if (provider) {
-              providers[channelType] = provider;
-            }
-          }
+      if (!type) continue;
+
+      const channelType = STEP_TYPE_TO_CHANNEL_TYPE.get(type);
+      if (!channelType) continue;
+      if (providers[channelType]) continue;
```
p-fernandez, line 299:
```suggestion
      if (!channelType || providers[channelType]) continue;
```
Good approach that we can even simplify it.

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -63,6 +67,16 @@ import { SharedModule } from '../shared/shared.module';
...
+export const getIsTopicNotificationEnabled = {
+  provide: GetIsTopicNotificationEnabled,
+  useFactory: async (featureFlagsServiceItem: FeatureFlagsService): Promise<GetIsTopicNotificationEnabled> => {
+    const useCase = new GetIsTopicNotificationEnabled(featureFlagsServiceItem);
+
+    return useCase;
+  },
+  inject: [FeatureFlagsService],
+};
+
```
p-fernandez, line 79: You don't need to do this as you can just import the class reference `GetIsTopicNotificationEnabled`. The instantiation is already done inside custom providers in application generic.

libs/shared/src/dto/topic/topic-subscriber.interface.ts
```diff
@@ -0,0 +1,15 @@
+import { EnvironmentId, ExternalSubscriberId, OrganizationId, SubscriberId, TopicId, TopicKey } from '../../types';
+
+export interface ITopicSubscriber {
```
p-fernandez, line 3: I couldn't guess what was the reason behind you creating this interface that is only used in the TopicSubscriberDTO. What it was? Do you plan to reuse it somewhere and we would need this interface for another use case?

djabarovgeorge: Please note that it is used in 
https://github.com/novuhq/novu/blob/517b4fa662a8b5a72d62a1518bff9d68c21f5909/packages/application-generic/src/usecases/get-topic-subscribers/get-topic-subscribers.use-case.ts#L48
as well.
Due to the reason, we do the get-topic-subscribers mapping in the application generic now, it does not make sense to use API TopicSubscriberDto
there.

scopsy (approved):

PR #4209, opened 2023-09-27
feat: add templateid support for sendgrid

jainpawan21:
Added `templateId` and `dynamicTemplateData` fields to support the use of a pre-built SendGrid template.


p-fernandez (approved):
🌟

PR #4212, opened 2023-09-27
fix(app-generic): proper naming for the standard worker and worker svc

p-fernandez:

### DIFF
--- a/packages/application-generic/src/services/workers/old-instance-standard-worker.service.ts
+++ b/packages/application-generic/src/services/workers/old-instance-standard-worker.service.ts
@@ -9,14 +9,14 @@ import {
   WorkerOptions,
 } from '../bull-mq';
 
-const LOG_CONTEXT = 'OldInstanceWorkerService';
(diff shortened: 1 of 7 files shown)


packages/application-generic/src/services/workers/old-instance-standard-worker.service.ts
```diff
@@ -9,14 +9,14 @@ import {
...
-const LOG_CONTEXT = 'OldInstanceWorkerService';
+const LOG_CONTEXT = 'OldInstanceStandardWorkerService';
 
 type WorkerProcessor = string | Processor<any, unknown, string> | undefined;
 
 /**
  * TODO: Temporary for migration to MemoryDB
  */
-export class OldInstanceWorkflowWorkerService {
+export class OldInstanceStandardWorkerService {
```
p-fernandez, line 19: Line 26 is the reason for all these changes.

scopsy (approved):

PR #4215, opened 2023-09-27
chore: remove unused docs project

scopsy:

Removes the docs project from the monorepo.

We have moved the docs project to a new repository and removed the implementation to point to mintlify


djabarovgeorge (approved):
Could not find any other references 💪

PR #4031, opened 2023-08-24
feat: fan out process to separate queue

djabarovgeorge:
please dont review i wanted to trigger the ci.

PR #4225, opened 2023-09-28
feat: fan-out subscriber processing

djabarovgeorge:

Recreation of https://github.com/novuhq/novu/pull/4031


packages/application-generic/src/modules/queues.module.ts
djabarovgeorge, line 1: All those dependencies are because of SubscriberJobBoundUsecase, i wonder if we should add SharedModule instead.

libs/shared/src/config/job-queue.ts
```diff
@@ -5,10 +5,12 @@ export enum JobTopicNameEnum {
   STANDARD = 'standard',
   WEB_SOCKETS = 'ws_socket_queue',
   WORKFLOW = 'trigger-handler',
+  SUBSCRIBER_PROCESS = 'subscriber-process',
```
p-fernandez, line 8:
```suggestion
  PROCESS_SUBSCRIBER = 'process-subscriber',
```
Just a matter of meaning.

libs/testing/src/jobs.service.ts
```diff
@@ -11,26 +11,20 @@ export class JobsService {
 
   public standardQueue: Queue;
   public workflowQueue: Queue;
+  public subscriberProcessQueue: Queue;
```
p-fernandez, line 14: I think that if you want this structure we should go for `subscriberProcessingQueue` naming.

packages/application-generic/src/modules/queues.module.ts
p-fernandez: No, these providers and use cases you added doesn't belong to this module and shouldn't. Only Queue/Workers class definitions and related. If any of them has dependencies they should be provided when used in their appropriate modules or shared module.
I am being quite cutting because doing it this way we would repeat the problem we have in other modules of not being able to manage the dependencies and that I wanted to solve by grouping services and use cases in domain modules. Think for example a HealthModule with all the code needed for that domain.
I am happy to help you with this bit as I didn't write any document to explain this direction and only left the explanation in my PRs. Also because you might have a problem with the dependency injection that I am missing. But overall the goal of this module was to them be used to be imported in other modules as clean as possible.

packages/application-generic/src/services/workers/subscriber-process-worker.service.ts
```diff
@@ -0,0 +1,72 @@
...
+  SubscriberJobBoundUsecase,
+} from '../../usecases';
+import { PinoLogger, storage, Store } from '../../logging';
+
+@Injectable()
+export class SubscriberProcessWorkerService
+  extends WorkerBaseService
+  implements INovuWorker
+{
+  public readonly name = JobTopicNameEnum.SUBSCRIBER_PROCESS;
```
p-fernandez, line 21: This is not used here.

packages/application-generic/src/services/workers/subscriber-process-worker.service.ts
```diff
@@ -0,0 +1,72 @@
...
+                  transaction.end();
+                });
+            });
+          }
+        );
+      });
+    };
+  }
+
+  private getWorkerOpts() {
```
p-fernandez, line 66: This shouldn't live in this class.
This would need to be added to a SubscriberProcessWorker in the Worker app that would extend from SubscriberProcessWorkerService, if you want to follow the generic implementations I did for workers.

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -57,7 +58,8 @@ export class TriggerEvent {
     private processTenant: ProcessTenant,
     private logger: PinoLogger,
     private analyticsService: AnalyticsService,
-    private mapTriggerRecipients: MapTriggerRecipients
+    private mapTriggerRecipients: MapTriggerRecipients,
+    private subscriberProcessQueueService: SubscriberProcessQueueService
```
p-fernandez, line 62: Adding this lines 52-54 are not needed anymore. Those dependencies are moved to SubscriberJobBound use case, right?

packages/application-generic/src/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
```diff
@@ -0,0 +1,200 @@
...
+        )} of organization ${command.organizationId} in transaction ${
+          command.transactionId
+        } was not processed. No jobs are created.`,
+        LOG_CONTEXT
+      );
+
+      return;
+    }
+
+    const notificationJobs = await this.createNotificationJobs.execute(
```
p-fernandez, line 113: Now I understand the naming a bit but we might consider if the main goal of the use case is to create the jobs needed and processing the subscriber is the dependency as we need their information to proper create the jobs. Mention that as it seems by the naming that the main goal of the new queue is to process subscribers but at the end what it does is to create all the jobs needed for them through AddJob to add them to a different queue.

p-fernandez (changes requested):

packages/application-generic/src/services/bull-mq/bull-mq.service.ts
```diff
@@ -18,7 +18,7 @@ import {
   InMemoryProviderEnum,
   InMemoryProviderService,
 } from '../in-memory-provider';
-import { validateMemoryDbClusterProviderConfig } from '../in-memory-provider/providers/memory-db-cluster-provider';
+import { BulkJobOptions } from 'bullmq/dist/esm/interfaces';
```
p-fernandez, line 21: The import should be done from the package. Try to import it in the line 13.

packages/application-generic/src/services/bull-mq/bull-mq.service.ts
```diff
@@ -229,6 +230,38 @@ export class BullMqService {
...
   }
 
+  public async addBulk(
+    data: [
+      {
+        name: string;
+        data: BullMqJobData;
+        options: BulkJobOptions;
+        groupId?: string;
+      }
```
p-fernandez, line 240: Shouldn't this type be:
```
 {
        name: string;
        data: BullMqJobData;
        options: BulkJobOptions;
        groupId?: string;
      }[]
```
If we are just passing a single element array, I think we should change the declaration of the arguments in the function to like the `add` method in line 214.

packages/application-generic/src/services/bull-mq/bull-mq.service.ts
djabarovgeorge: Good catch! missed this one!

packages/application-generic/src/services/bull-mq/bull-mq.service.ts
djabarovgeorge: Not sure i am following here, i would still love it to be an object tho

p-fernandez (approved):
🌟

PR #4237, opened 2023-09-30
fix: Fixed send the sender name in email when using resend email provider

### LINKED ISSUE #3957: [NV-2697] senderName field is missing in resend api and nodejs sdk

piyushgarg-dev:
This PR fixes the sender name being not sent in email when using Resend as an email provider.

The change was to add senderName before the from email address. Example: `Piyush Garg <piyushgarg.dev@gmail.com>`.

Fixes #3957


Dyr0tH (commented):
It looks like the problem is solved and the sender name is now present via 'Resend' mail service.

scopsy:
@piyushgarg-dev could you please comment on the issue so we can assign you to it?

piyushgarg-dev:
Done @scopsy

providers/resend/src/lib/resend.provider.ts
```diff
@@ -26,7 +27,9 @@ export class ResendEmailProvider implements IEmailProvider {
     options: IEmailOptions
   ): Promise<ISendMessageSuccessResponse> {
     const response: any = await this.resendClient.sendEmail({
-      from: options.from || this.config.from,
+      from: this.config?.senderName
+        ? `${this.config?.senderName} <${options.from || this.config.from}>`
+        : options.from || this.config.from,
```
scopsy, line 32: Do you think for better visibility maybe we can extract some of it to variables? For example:


const fromAddress = options.from || this.config.from;

....
from: this.config?.senderName
        ? `${this.config?.senderName} <${fromAddress}>`
        : fromAddress,
....

scopsy (approved):

providers/resend/src/lib/resend.provider.ts
piyushgarg-dev: Sure @scopsy Please check commit https://github.com/novuhq/novu/pull/4237/commits/da058e81b25bc944c8ac3a1ae8f783d1678b1b41

piyushgarg-dev:
Hi @rifont, I have added the test script 😄
https://github.com/novuhq/novu/pull/4237/files#diff-934b8378daa66d6a98de9a371c69ecf99f320ea36a3c5debaa995247b7e10184R43-R67

providers/resend/src/lib/resend.provider.spec.ts
```diff
@@ -39,3 +39,29 @@ test('should trigger resend library correctly', async () => {
...
+
+test('should trigger resend email with From Name', async () => {
+  const mockConfigWithSenderName = {
+    ...mockConfig,
+    senderName: 'Test User',
+  };
+
+  const provider = new ResendEmailProvider(mockConfigWithSenderName);
+  const spy = jest
+    .spyOn(provider, 'sendMessage')
```
rifont, line 51: The tests are currently failing (https://github.com/novuhq/novu/actions/runs/6394994960/job/17358214122?pr=4237#step:5:86) because the `sendMessage` method on the provider is being mocked, instead of the client's send method.

Instead, we should mock the `resendClient`'s `sendEmail` method to execute the tested code. For an example, take a look at the Vonage client tests: https://github.com/novuhq/novu/blob/2f47f7b0add731870bef1672cc928246a654279b/providers/nexmo/src/lib/nexmo.provider.spec.ts#L13

Let me know if you have any trouble with the example and mocking 😀

rifont (changes requested):

providers/resend/src/lib/resend.provider.spec.ts
piyushgarg-dev: Hi @rifont, Thanks a lot for the example.
I have checkout the implementation of SendGrid and mocked it that way 😄. Tests are now passing on my local machine.

Thanks a lot 🙌🏻

rifont (approved):
Thanks for your contribution 🙌

Issue #4234, opened 2023-09-29
[NV-2931] [NV-2000] Display Novu Version to the WEB UI for self-hosted users


scopsy:
## What?

Display the current version of the Novu product. 

**Note:** 

* The potential solution could be to introduce the env variable that will be set during the build in the Webpack configuration, then use that variable inside the Web app

## Why?

So users can know what current version they are running when self-hosting. This can help us identify support issues related to old versions and etc…

## Definition of done

* The Novu version in the dropdown menu when clicking on a user avatar is displayed
* The version is displayed for self-hosted users only

[NOVU Versioning for selfhosted users - Crafting updates (Figma)](https://www.figma.com/file/mxViLXxtC0PnB0ehXDcafw/Crafting-updates?type=design&node-id=831%3A23443&mode=design&t=smxZcujJc1Arklf4-1)

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-2000](https://linear.app/novu/issue/NV-2000/display-novu-version-to-the-web-ui-for-self-hosted-users)</sub>

<sub>[NV-2931](https://linear.app/novu/issue/NV-2931/[nv-2000]-display-novu-version-to-the-web-ui-for-self-hosted-users)</sub>

pedrobruneli:
Hey @scopsy how do i differ a self-hosted person to your cloud?

pedrobruneli:
@scopsy i wanna do this one

PR #4241, opened 2023-09-30
Display Novu Version to the WEB UI for self-hosted users

pedrobruneli:

- The Novu version in the dropdown menu when clicking on a user avatar is displayed
- The version is displayed for self-hosted users only

So users can know what current version they are running when self-hosting. This can help us identify support issues related to old versions and etc…

Closes: #4234 

image


LetItRock (approved):
hey @pedrobruneli! 👋 
thanks a lot it works perfectly! 🚀

PR #4272, opened 2023-10-01
Fix/resend email sender

mohit-100:
This PR Fix : The sender name can't send a Email when we used Resend as email provider;

 The channge is add senderName Before a Email address 

Fixes : #3957


scopsy:
Hi @mohit-100 ! 👋 thanks for the contribution 🤝
For now, I'm closing it.

PR #4278, opened 2023-10-01
feat: add actor to system variables

ainouzgali:

Add actor to system variables, allows the use of the actor's subscriber data in editor + allows creation and update of the subscriber. 

I still need to add tests.

Screen Shot 2023-10-01 at 17 00 18
Screen Shot 2023-10-01 at 17 00 38
Screen Shot 2023-10-01 at 17 00 48
Screen Shot 2023-10-01 at 17 01 06
Screen Shot 2023-10-01 at 17 01 22


scopsy (approved):
🎉

PR #4282, opened 2023-10-01
fix: tenant fetch by id

djabarovgeorge:

### DIFF
--- a/libs/dal/src/repositories/base-repository.ts
+++ b/libs/dal/src/repositories/base-repository.ts
@@ -9,7 +9,7 @@ import {
 import { Model, Types, ProjectionType, FilterQuery, UpdateQuery, QueryOptions } from 'mongoose';
 import { DalException } from '../shared';
 
-export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
+export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement> {
   public _model: Model<T_DBModel>;
 
   constructor(protected MongooseModel: Model<T_DBModel>, protected entity: ClassConstructor<T_MappedEntity>) {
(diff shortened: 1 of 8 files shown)


libs/dal/src/repositories/base-repository.ts
```diff
@@ -9,7 +9,7 @@ import {
 import { Model, Types, ProjectionType, FilterQuery, UpdateQuery, QueryOptions } from 'mongoose';
 import { DalException } from '../shared';
 
-export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement = object> {
+export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement> {
```
djabarovgeorge, line 12: removed the default value from the base, if any of the child classes are already on the env level they can pass empty object. this way we will make sure other classes are enforced to pass enforcement type

libs/dal/src/repositories/tenant/tenant.repository.ts
```diff
@@ -3,8 +3,9 @@ import { SoftDeleteModel } from 'mongoose-delete';
 import { BaseRepository } from '../base-repository';
 import { TenantDBModel, TenantEntity } from './tenant.entity';
 import { Tenant } from './tenant.schema';
+import { EnforceEnvId, EnforceEnvOrOrgIds } from '../../types';
 
-export class TenantRepository extends BaseRepository<TenantDBModel, TenantEntity> {
+export class TenantRepository extends BaseRepository<TenantDBModel, TenantEntity, EnforceEnvId> {
```
djabarovgeorge, line 8: added EnforceEnvId because there could be the same identifier ids on 2 organizations.

scopsy (approved):

PR #4283, opened 2023-10-01
fix: Fixed the email sender name problem in resend email provider

### LINKED ISSUE #3957: [NV-2697] senderName field is missing in resend api and nodejs sdk

priyanshusinha11:

This PR fixes the sender name being not sent in email when using Resend as an email provider.

The change was to add senderName before the from email address. Example: `Priyanshu Sinha <priyanshusinha.2807@gmail.com>`
Fixes #3957

### DIFF
--- a/packages/application-generic/src/factories/mail/handlers/resend.handler.ts
+++ b/packages/application-generic/src/factories/mail/handlers/resend.handler.ts
@@ -6,8 +6,8 @@ export class ResendHandler extends BaseHandler {
   constructor() {
     super('resend', ChannelTypeEnum.EMAIL);
   }
-  buildProvider(credentials: ICredentials, from?: string) {
-    const config: { apiKey: string; from: string } = {
+  buildProvider(credentials: ICredentials, from?: string, senderName?: string) {
+    const config: { apiKey: string; from: string; senderName?: string } = {
       from: from as string,
       apiKey: credentials.apiKey as string,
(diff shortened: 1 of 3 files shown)


LetItRock:
hi @priyanshusinha11 ! 👋 thanks for the contribution 🤝
For now, I'm closing it.

PR #4286, opened 2023-10-01
An AsyncGenerator object cannot be iterated at 'yield*'

sachin-nanayakkara:
The error you're encountering, "An AsyncGenerator object cannot be iterated at 'yield*'," is because you are trying to yield from an async generator using the yield* syntax, and it's not allowed in JavaScript. The yield* syntax is used for delegating to another generator, but in the context of an async generator, you cannot directly delegate to an async generator using yield*.


scopsy:
@sachin-nanayakkara what issue is this addressing? I have never encountered this error before.

PR #4300, opened 2023-10-02
feat: add newrelic custom attributes

scopsy:

For more context on APM transactions and breaking down transactions by org and job type (digest, delay, etc...)

### DIFF
--- a/apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
@@ -34,19 +34,18 @@ export class RunJob {
     const job = await this.jobRepository.findById(command.jobId);
     if (!job) throw new PlatformException(`Job with id ${command.jobId} not found`);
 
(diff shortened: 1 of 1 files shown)


apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -32,6 +34,12 @@ export class RunJob {
     const job = await this.jobRepository.findById(command.jobId);
     if (!job) throw new PlatformException(`Job with id ${command.jobId} not found`);
 
+    nr.addCustomAttributes({
+      jobType: job.type,
+      organizationId: job._organizationId,
+      transactionId: job.transactionId,
+    });
```
p-fernandez, line 41: I think we should merge this and the line 44 as we are adding 3/4 of the same information. And I think both are doing the same.

Also this would only add the attributes to the scope of this use case, nothing else, if I am not wrong. Is that your goal?

p-fernandez (approved):
🌟

Issue #4294, opened 2023-10-01
[NV-2933] Fix the kebab-case for css properties warning


LetItRock:
## Reproduction Steps
1. Navigate to the Get Started page
2. Navigate to the Workflows page
3. The warning: `Using kebab-case for css properties in objects is not supported. Did you mean WebkitTextFillColor?` is shown in the console.

![image](https://github.com/novuhq/novu/assets/2607232/c1432f95-2709-4349-98a9-0e2d3c14c812)

## Expected Behaviour
- provide the fix so the warning is not printed in the console

sagnik3788:
i want to work on this issue please assign me @LetItRock

LetItRock:
> i want to work on this issue please assign me @LetItRock

@sagnik3788 sure! thanks and good luck 🤞

gouravkundu7370:
i want to work on this issue please assign me @LetItRock

LetItRock:
> i want to work on this issue please assign me @LetItRock

hey @gouravkundu7370! 👋 the issue is already assigned, please take a look into others we have :)

shakthi2003-ice:
where is the get started page?

Debangan202Ok:
I will work on it

LetItRock:
hey @shakthi2003-ice @Debangan202Ok! 👋 
Thanks for your willingness to help us improve the Novu :) but we have an assigned person to this ticket and there is already an opened PR.

pragyamishra56:
could you please assign me it would be my first contribution?

soumik-saha:
Hey @LetItRock, I am not getting this error in my console. I think it is a system specific issue. Please tell me what I should do for resolving it?

LetItRock:
> Hey @LetItRock, I am not getting this error in my console. I think it is a system specific issue. Please tell me what I should do for resolving it?

hey @soumik-saha! thanks for your willingness to help, but it's already assigned to @sagnik3788 who has already opened a PR with the fix :) please take a look into some other tickets we have ;)

sagnik3788:
hey @LetItRock can you check my pr

LetItRock:
> hey @LetItRock can you check my pr

sure, we will do that soon ;)

shaurya-jain23:
can you please assign it to me, it would be my first contribution

PR #4301, opened 2023-10-02
Fix the kebab-case for css properties warning

sagnik3788:
solve the kebab-case for css properties warning

solve issue #4294 @LetItRock


apps/web/src/design-system/button/Button.styles.ts
```diff
@@ -66,7 +66,7 @@ export const getSubtleStyles = (theme) => {
           backgroundColor: dark ? theme.white : 'transparent',
           backgroundClip: 'none !important',
           color: 'transparent',
-          '-webkit-text-fill-color': 'initial !important',
+          WebkitTextFill: 'initial !important',
```
scopsy, line 69: @LetItRock I think it should be: `WebkitTextFillColor` ?

sagnik3788: is there anything i change @scopsy @LetItRock

davidsoderberg: Hi @sagnik3788 if the error that is shown in the issue is not showing anymore I see this as the solution to the issue and nothing else needed to change 😄

sagnik3788:
if everything is okay can you merge my pr @davidsoderberg

davidsoderberg:
Hi @sagnik3788, I think it should be `WebkitTextFillColor` or have you checked so the css is applied with your solution?

sagnik3788:
Yes, you are correct  it should be WebkitTextFillColor due to the camelCase property naming convention . i correct this

apps/web/src/design-system/button/Button.styles.ts
```diff
@@ -66,7 +66,7 @@ export const getSubtleStyles = (theme) => {
           backgroundColor: dark ? theme.white : 'transparent',
           backgroundClip: 'none !important',
           color: 'transparent',
-          '-webkit-text-fill-color': 'initial !important',
+        WebkitTextFillColor: 'initial !important',
```
ainouzgali, line 69: This line should have the same indentation as the others.

ainouzgali (commented):
@sagnik3788 please run the prettier and I'll approve this. Thanks!

LetItRock (approved):

PR #4306, opened 2023-10-02
WIP - Fix Nested Validation

djabarovgeorge:

The PR introduces WIP of what we need to do in order to fix the nested validation that currently is broken. In order to add nested validation to the variants we will need to do the following: 
* Create a custom validation decorator. 
* Fix the nested validation that does not work at the moment. In order to fix it we need to do the following add ValidateNested decorator, if it is an array we should add each as well (`@ValidateNested({ each: true })`), then add `@Type(() => NotificationStepVariant)` decorator from `class-transformer` in order to map the type implicitly.
* The hardest part IMO will be to return the nested errors that we will get, at the moment we throw an exception on the first layer, but it an error is not thrown we do flat on the first layer on BaseComman which won't return us the nested `constrains`.

So we could validate nested objects and return a valid error if validations had errors.

### Other information

In this PR I am adding nested validation only for variants as POC, once we merge this PR we most definitely need to validate all of the rest `ValidateNested` decorators.

### DOD
* Add unit tests for the extractConstraints function.
* Add tests of command in order to check that the new logic works with it as expected.
* Add max validation depth in order to not over-compute extractConstraints if the input is unexpectedly long or infinite.

### DIFF
--- a/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts
+++ b/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts
@@ -21,6 +21,8 @@ import {
 
 import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
 import { MessageTemplate } from '../../../shared/dtos/message-template';
+import { Type } from 'class-transformer';
+import { VariantWithFilter } from '@novu/application-generic';
 
 /**
  * DEPRECATED:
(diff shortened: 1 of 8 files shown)


apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.command.ts
```diff
@@ -46,7 +48,8 @@ export class CreateNotificationTemplateCommand extends EnvironmentWithUserComman
 
   @IsDefined()
   @IsArray()
-  @ValidateNested()
+  @ValidateNested({ each: true })
+  @Type(() => NotificationStep)
   steps: NotificationStep[];
```
ainouzgali, line 53: I believe this should be added in update-notification-template.command as well

ainouzgali (approved):
@djabarovgeorge thanks for doing this!! Left a couple of comments

Issue #4244, opened 2023-09-30
[NV-2570] Update email and In-app editor's height to sync it with the viewport height


LetItRock:
## What?

The email and In-app editor's height

![image](https://github.com/novuhq/novu/assets/2607232/74106af6-67f2-48fe-97b5-16f50ab77c52)

![image](https://github.com/novuhq/novu/assets/2607232/64af1e14-b568-4e55-bd1f-c7d9259b1fdf)

## Why? 

Currently, the email and In-app editor's height is not adjusted based on the available viewport height.

## Definition of done:
- The email editor's height is adjusted based on the available viewport height.
- The email custom code height is adjusted based on the available viewport height.
- The In-app editor's height is adjusted based on the available viewport height
- The bottom margin is 24px

![image](https://github.com/novuhq/novu/assets/2607232/71a657ec-633c-40e1-91d0-842002dffc44)

![image](https://github.com/novuhq/novu/assets/2607232/a9edf979-9169-49bf-a901-6be95d1e7ffb)

![image](https://github.com/novuhq/novu/assets/2607232/42977436-2494-4534-9518-1029b19919a2)

sarmadhamdani02:
Hi there 👋
can I get assigned this issue?

soubhik-111:
Can you assign me

ainouzgali:
Thank you @soubhik-111 ! assigning you

microsoftwala:
can you please assign me

SmitDasadia:
Hi,

I'm following up on Issue #4244 regarding "Update email and In-app editor's height to sync it with the viewport height" that was opened for [Repository Name]. Could you please provide an update on its status?

Thank you.

SmitDasadia:
Hi,

I'm following up on Issue #4244 regarding "Update email and In-app editor's height to sync it with the viewport height". Could you please provide an update on its status weahter is open or not?

Thank you.

SmitDasadia:
Hi,

I'm following up on Issue #4244 regarding "Update email and In-app editor's height to sync it with the viewport height". Could you please provide an update on its status weather is open or not?

Thank you.

sarthaksharma27:
Working on it

scopsy:
This issue was already assigned to @soubhik-111, please let us know if you would love to pass this issue to @SmitDasadia in case you can't continue with it 🙏

soubhik-111:
Okay @scopsy  Pass it to @SmitDasadia

[...]

PR #4313, opened 2023-10-02
Update email and In-app editor's height to sync it with the viewport height

SmitDasadia:
This PR update Email and In-app editor's height to sync it with the viewport height

The change was to add div with  sx={styledContainer} and contains height, overflow and marginBottom for email editor's, email custom code and In-app editor's height.

Fixes: #4244


apps/web/src/pages/templates/components/email-editor/EmailCustomCodeEditor.tsx
```diff
@@ -68,9 +69,17 @@ export function EmailCustomCodeEditor({
         }}
       />
     </Card>
+    </div>
   );
 }
 
+const styledContainer = () => ({
+  height: '80vh', // Adjust the percentage as needed
```
ainouzgali, line 77: Setting a wrapper to be 80vh will not do the job, as the card itself -> the ace editor is set to a specific height (line 55). So what you did will  simply result in the editor height being as it is now, unchanged, but will add a lot of empty page + scroll

ainouzgali (commented):
Hi @SmitDasadia , thanks for your contribution! Currently, it doesn't compile and does not solve the issue as it doesn't fix the actual height. 
Please take another look :)

Also, if you could, its always helpful to add a screenshot of your result. We would appreciate it :)

scopsy:
Closing due to inactivity

PR #4329, opened 2023-10-02
fix: Fixed senders name in email while using resend email provider

mahimavarma:
This PR will fix the name not being displayed in the email when the user is using Resend as an email provider.
The change was to add the senderName before the from email address 
Fixes:#3957

image
image

### DIFF
--- a/packages/application-generic/src/factories/mail/handlers/resend.handler.ts
+++ b/packages/application-generic/src/factories/mail/handlers/resend.handler.ts
@@ -6,10 +6,11 @@ export class ResendHandler extends BaseHandler {
   constructor() {
     super('resend', ChannelTypeEnum.EMAIL);
   }
-  buildProvider(credentials: ICredentials, from?: string) {
-    const config: { apiKey: string; from: string } = {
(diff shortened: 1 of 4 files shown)


packages/application-generic/src/factories/mail/handlers/resend.handler.ts
```diff
@@ -6,10 +6,11 @@ export class ResendHandler extends BaseHandler {
   constructor() {
     super('resend', ChannelTypeEnum.EMAIL);
   }
-  buildProvider(credentials: ICredentials, from?: string) {
-    const config: { apiKey: string; from: string } = {
+  buildProvider(credentials: ICredentials, from?: string, senderName?: string) {
```
BiswaViraj, line 9: `credentials: ICredentials` already includes the senderName.
So we don't need to pass senderName as an extra param. Also in the current implementation. In the `mail.factory.ts` file, only 2 params are being passed for all the providers.

BiswaViraj (changes requested):

LetItRock:
hi @mahimavarma! 👋 thanks for the contribution 🤝
For now, I'm closing it.

PR #4351, opened 2023-10-03
Failing Unit test for Providers - nodemailer

mainadennis:

This change addresses failing unit tests from providers as a result of the newly added "name" property on nodemailer SMTPTransport options

Failing Unit test for Providers

image


scopsy:
I believe it's already was resolved on next. Closing this now.

PR #4376, opened 2023-10-03
Update HomePage.tsx Navigation Issue in HomePage Component

GAUTAMMANAK1:
Navigation Issue in HomePage Component

Issue Description:
In the HomePage component, there is a potential issue with the usage of the Navigate component from react-router-dom. The Navigate component is being used to redirect users to the ROUTES.WORKFLOWS route, but there may be an issue with the replace prop that could lead to unexpected behavior.

Steps to Reproduce:

Load the application containing the HomePage component.
Observe the behavior of the navigation when the HomePage component is rendered.
Expected Behavior:
The Navigate component should redirect users to the ROUTES.WORKFLOWS route without any unexpected side effects.

Actual Behavior:
The replace prop is used with the Navigate component, but its behavior can vary depending on the router version and configuration. Depending on the router's version and settings, using replace may cause unexpected behavior, such as:

Navigating to the ROUTES.WORKFLOWS route without adding an entry to the navigation history stack, preventing users from going back to the previous page.
Unexpected URL changes or incorrect navigation behavior.
Additional Information:

The behavior of the replace prop can depend on the specific version and configuration of the router library being used in the project.
Consider whether using replace is necessary for your intended navigation behavior. It is typically used when you want to replace the current URL in the history stack instead of adding a new entry. Ensure that this behavior aligns with your application's navigation requirements.
Environment:

React Router Version: [Specify the version of react-router-dom being used, e.g., v6]
Router Configuration: [Provide any relevant details about the router configuration, if applicable]


scopsy:
Please create an issue before submitting changes using a PR. Closing this for now, feel free to reopen once we have discussed the intended behaviour in an issue 🙏

PR #4399, opened 2023-10-04
feat: remove variants from the workflow list response

BiswaViraj:
- removes variant data from the workflow list response.
- e2e test for the same


davidsoderberg (approved):
🥳

PR #4404, opened 2023-10-04
fix: Fixed Empty Style Element Problem

### LINKED ISSUE #4257: [NV-2790] Error: empty style element while compiling email template
## Reproduction steps
1. Create a new workflow.
2. Add the email step to the workflow.
3. Navigate to the Custom Code editor.
4. Paste the below template
5. Trigger the workflow from the "Get Snippet" button

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Some Title</title>
    <style type="text/css"></style>
  </head>
  <body>
    Test
  </body>
</html>
```

Then after the workflow is triggered this error will be shown in the Execution Details:
image

## Definition of done
- the workflow should be triggered without the errors

anshul-kh:

This PR fixes the empty style element error when running trigger of  email workflow trigger.

This change was to resolve the error when custom code is entered in email workflow trigger with empty style element.
The  style element cannot be empty by convention.

Fixes #4257

I have created a check to alert the user that Style tag cannot be empty whenever it detects one.

Screenshot_20231004_195914

### DIFF
--- a/apps/web/src/pages/templates/components/email-editor/EmailCustomCodeEditor.tsx
+++ b/apps/web/src/pages/templates/components/email-editor/EmailCustomCodeEditor.tsx
@@ -1,47 +1,61 @@
 import AceEditor from 'react-ace';
 import 'ace-builds/src-noconflict/mode-handlebars';
 import 'ace-builds/src-noconflict/theme-monokai';
+import 'ace-builds/src-noconflict/mode-css';
+import 'ace-builds/src-noconflict/worker-css';
 import { addCompleter } from 'ace-builds/src-noconflict/ext-language_tools';
 import { Card } from '@mantine/core';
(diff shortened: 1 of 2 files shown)


ainouzgali:
Hi @anshul-kh, thanks for your work! 
The desired outcome is that sending the email will not fail when there is an empty style tag. 
With your solution, as I can see, you only show an alert. The user could still save and send it, and the execution will still fail. Or am I missing something?

anshul-kh:
Hi @ainouzgali  ! Thanks for your review . You are absolutely correct . I got your point and will work on it . Can you please suggest me some basic approach to achieve the desired solution . It'll be helpful for me.

scopsy:
@anshul-kh closing this for now, so we can continue conversation in the issue page 🙏

PR #4418, opened 2023-10-05
feat: add script create release with submodules

djabarovgeorge:

* Create version update in the monorepo and submodule.
* Updated the submodule repo with commit and tag. 
* Updated the main repo with commit and tag.

So we could create a version in one flow.

Addition point to check:
* We need to validate that there is no uncommit files in a submodule, if there is we need to throw an error (in the main repo lerna takes care of it)


scripts/release-with-submodule.ts
```diff
@@ -0,0 +1,42 @@
...
+  console.log(`Running command: ${command}${directory ? `directory:${directory}` : '.'}`);
+  execSync(command, {
+    stdio: 'inherit',
+    cwd: directory,
+  });
+}
+
+try {
+  console.log('Step 1: Run lerna version in monorepo without committing or creating a release');
+  runCommand('pnpm lerna version patch --yes --no-git-tag-version --no-push --force-publish');
```
djabarovgeorge, line 17:
```suggestion
  runCommand('pnpm lerna version patch --yes --no-git-tag-version --no-push ');
```

needed to be removed, i added it only for local testing.

scripts/release-with-submodule.ts
```diff
@@ -0,0 +1,79 @@
...
+    runCommand('git commit -m "chore: update versions"', submoduleDir);
+    runCommand('git push', submoduleDir);
+
+    console.log('Step 6: Create release tag for submodule');
+    runCommand(`git tag v${newVersion}`, submoduleDir);
+    runCommand('git push --tags', submoduleDir);
+
+    console.log('Step 7: Commit version changes and submodule updated hash main repository');
+    runCommand('git add .');
+    runCommand('git commit -m "chore: update versions"');
```
djabarovgeorge, line 70: @scopsydo Do you know if we have some commit convention, that will require an update in the commit messages?

LetItRock: @djabarovgeorge Usually the message is like this

Screenshot 2023-10-18 at 12 51 02

LetItRock (approved):

scripts/release-with-submodule.ts
djabarovgeorge: totally right that's why I asked, maybe now we would like to customize it abit.

djabarovgeorge: will update to 
`chore(release): publish`

LetItRock: shouldn't we keep `- ci skip`?

djabarovgeorge: If you say so, i was not sure what was it purpose 😕

PR #4429, opened 2023-10-05
feat(api): Add Access-Control-Max-Age response header

rifont:

Adds an Access-Control-Max-Age response header to API responses, with cache duration of 86400 (maximum allowable on browsers, 1 day).

By specifying a Max-Age CORS header, we can provide a more responsive browser experience by eliminating a round-trip to the API in event of a preflight request cache-hit, and significantly reduce API server load.

After a successful preflight request against a URI, subsequent requests against the same URI do not require a new preflight request.
image


davidsoderberg (approved):
👍

PR #4440, opened 2023-10-06
Nv 2893 endpoint to create the new translation group

davidsoderberg:

Moving code to support translation module in enterprise package.

### DIFF
--- a/packages/application-generic/src/decorators/external-api.decorator.ts
+++ b/packages/application-generic/src/decorators/external-api.decorator.ts
@@ -1,0 +1,5 @@
+import { SetMetadata } from '@nestjs/common';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const ExternalApiAccessible = () =>
+  SetMetadata('external_api_accessible', true);
--- a/packages/application-generic/src/decorators/user-session.decorator.ts
+++ b/packages/application-generic/src/decorators/user-session.decorator.ts
@@ -1,11 +1,8 @@
 import { createParamDecorator, UnauthorizedException } from '@nestjs/common';
 import * as jwt from 'jsonwebtoken';
-import { UserSession } from '@novu/application-generic';
-
-export { UserSession };
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export const SubscriberSession = createParamDecorator((data, ctx) => {
+export const UserSession = createParamDecorator((data, ctx) => {
   let req;
   if (ctx.getType() === 'graphql') {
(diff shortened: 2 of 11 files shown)


LetItRock (approved):

packages/application-generic/src/decorators/external-api.decorator.ts
```diff
@@ -0,0 +1,5 @@
+import { SetMetadata } from '@nestjs/common';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const ExternalApiAccessible = () =>
```
p-fernandez, line 4: This makes the endpoints where it is applied `public` for our users or just our UI?

packages/application-generic/src/decorators/user-session.decorator.ts
```diff
@@ -0,0 +1,38 @@
+import { createParamDecorator, UnauthorizedException } from '@nestjs/common';
+import * as jwt from 'jsonwebtoken';
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export const UserSession = createParamDecorator((data, ctx) => {
+  let req;
+  if (ctx.getType() === 'graphql') {
```
p-fernandez, line 7: It is tech debt but I wonder why we keep having this.

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -0,0 +1,49 @@
...
+import { AuthService } from './auth.service';
+import { PinoLogger } from '../../logging';
+import { Instrument } from '../../instrumentation';
+
+@Injectable()
+export class JwtAuthGuard extends AuthGuard('jwt') {
+  constructor(
+    @Inject(forwardRef(() => AuthService)) private authService: AuthService,
+    private readonly reflector: Reflector,
+    private readonly logger: PinoLogger
```
p-fernandez, line 19: I can't see this being used (tech debt again I think)

p-fernandez (commented):
I find the PR name very confusing. This PR is moving the middleware decorators to Application Generic package.

packages/application-generic/src/decorators/external-api.decorator.ts
davidsoderberg: It makes it public

packages/application-generic/src/decorators/user-session.decorator.ts
davidsoderberg: Good question

p-fernandez (approved):
🌟

PR #4459, opened 2023-10-08
Fix send message uuuid

scopsy:


apps/api/src/app/widgets/usecases/mark-all-messages-as/mark-all-messages-as.usecase.ts
```diff
@@ -66,13 +67,12 @@ export class MarkAllMessagesAs {
...
     this.webSocketsQueueService.add(
-      'sendMessage',
+      uuid(),
       {
         event: isUnreadCountChanged ? WebSocketEventEnum.UNREAD : WebSocketEventEnum.UNSEEN,
         userId: subscriber._id,
         _environmentId: command.environmentId,
       },
-      undefined,
       subscriber._organizationId
```
scopsy, line 76: The group id should be the 3rd parameter not the last.

djabarovgeorge: True, good find !

djabarovgeorge (approved):
🚀

PR #4462, opened 2023-10-09
feat(node): add exponential retry mechanism with idempotency headers

### LINKED ISSUE #4452: [NV-3009] 🚀 Feature: Typescript SDK - Add Exponential Retry Mechanism with Idempotency Headers
In order to enhance the resilience and reliability of our SDK, we would like to introduce an Exponential Retry mechanism for retrying failed requests. Additionally, to ensure the idempotent processing of requests, it's vital to incorporate support for providing an Idempotency Key as per the draft specified in the [HTTP Idempotency Key Header Field](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/).

The key requirements for this implementation include:

1. **Exponential Retry Mechanism**:
   * The SDK should retry failed requests following an exponential backoff strategy to minimize the contention and impact on the systems involved.
   * The SDK should ensure that the retry mechanism is configurable (e.g., max retries, initial delay, maximum delay).
2. **Idempotency Key Provisioning**:
   * The SDK should allow for either automatic or manual provisioning of an Idempotency Key for each request.
   * The Idempotency Key should conform to either CUID, ULID, or UUID formats as specified in the draft.
   * The Idempotency Key should be included in the HTTP Header as `Idempotency-Key` and following the standards outlined in the draft.
3. **Configuration and Documentation**:
   * The SDK should provide configuration options for enabling/disabling the Exponential Retry mechanism and Idempotency Key provisioning.
   * Comprehensive documentation should be provided explaining the configuration options, operational behavior, and the benefits of using the Exponential Retry mechanism along with Idempotency Keys.

Acceptance Criteria:

* [ ] Implementation of the Exponential Retry mechanism with configurable parameters.
* [ ] Provisioning of Idempotency Keys, either automatically or manually, conforming to specified formats (CUID, ULID, or UUID).
* [ ] Adequate unit and integration testing to ensure the robustness and reliability of the implemented features.
* [ ] Comprehensive documentation on the usage and configuration of the Exponential Retry mechanism and Idempotency Key provisioning.
* [ ] Adherence to the specifications outlined in the [HTTP Idempotency Key Header Field draft](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/).

Please refer to the draft for further details on the HTTP Idempotency Key Header Field and ensure adherence to the specified standards while implementing this feature in the SDK.

**Update**: You can reference the go-lang library to keep the method signature and configuration the same.
[https://github.com/novuhq/go-novu/pull/62](https://github.com/novuhq/go-novu/pull/62)

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-3009](https://linear.app/novu/issue/NV-3009/🚀-feature-typescript-sdk-add-exponential-retry-mechanism-with)</sub>

michaldziuba03:

This PR introduces exponential retry mechanism with idempotency headers for Node SDK.

PR contains:
- adding idempotency key header for each `POST` and `PATCH` request
- configurable exponential retry mechanism
- tests

Closes #4452


packages/node/src/lib/novu.ts
```diff
@@ -34,13 +36,53 @@ export class Novu extends EventEmitter {
...
+      if (idempotencyKey) {
+        return axiosConfig;
+      }
+
+      axiosConfig.headers['Idempotency-Key'] = uuid();
+
+      return axiosConfig;
+    });
+
+    const retryConfig = config?.retryConfig || {};
```
dammy001, line 58: shouldn't this retry be opt-in? I mean there should be a check if retryConfig exists instead of making use of default values?

michaldziuba03: There is 0 retries by default.

michaldziuba03:
@Cliftonz I'm wondering about the retry condition. I think it should be clearly defined in requirements.

#### What I plan to do:

for **idempotent** http methods SDK will retry on 50x errors + network errors without response. I will not apply `Idempotency-Key` header for these idempotent  methods.

for **non-idempotent** http methods SDK will retry only on network errors without response / maybe on `409 Conflict`. I think most 50x errors will be replayed from the cache, so request with 50x status shouldn't be retried.

dammy001:
> @Cliftonz I'm wondering about the retry condition. I think it should be clearly defined in requirements.
> 
> 
> 
> #### What I plan to do:
> 
> 
> 
> for **idempotent** http methods SDK will retry on 50x errors + network errors without response. I will not apply `Idempotency-Key` header for these idempotent  methods.
> 
> 
> 
> for **non-idempotent** http methods SDK will retry only on network errors without response / maybe on `409 Conflict`. I think most 50x errors will be replayed from the cache anyway, so request errored with 50x status shouldn't be retried.
> 
>  

And also, there should be an option for setting retryCondition callback should from userland.

Cliftonz:
> @Cliftonz I'm wondering about the retry condition. I think it should be clearly defined in requirements.
> #### What I plan to do:
> 
> for **idempotent** http methods SDK will retry on 50x errors + network errors without response. I will not apply `Idempotency-Key` header for these idempotent methods.
> 
> for **non-idempotent** http methods SDK will retry only on network errors without response / maybe on `409 Conflict`. I think most 50x errors will be replayed from the cache anyway, so request errored with 50x status shouldn't be retried.

Just to make sure we are on the same page, these should be configurable by the client.
By default we want RetryMax, WaitMin, WaitMax , and InitialDelay to be resonable values like 
RetryMax: 0
Waitmin: 1
Waitmax: 2
InitialDelay: 0.5

As for applying the the idempotency header, you should only do it to POST and PATCH methods however we can add it to all calls its just not preferred.

As for which errors that should be retried, not all errors are considered equal and here are the Errors that should be considered for retrying:
    - 408 Request Timeout
    - 429 Too Many Requests
    - 500 Internal Server Error
    - 502 Bad Gateway
    - 503 Service Unavailable
    - 504 Gateway Timeout

Errors that generally should not be retried without addressing the cause:
    - 400 Bad Request
    - 401 Unauthorized
    - 403 Forbidden
    - 404 Not Found
    - 405 Method Not Allowed
    - 413 Payload Too Large
    - 414 URI Too Long
    - 415 Unsupported Media Type
    - 422 Unprocessable Entity

Cliftonz:
@michaldziuba03 I also agree with @dammy001 we should provide a callback for the user to dynamically select the condition if they so want.

packages/node/src/index.ts
```diff
@@ -30,3 +30,4 @@ export * from './lib/feeds/feeds.interface';
 export * from './lib/topics/topic.interface';
 export * from './lib/integrations/integrations.interface';
 export * from './lib/messages/messages.interface';
+export { defaultRetryCondition } from './lib/retries';
```
michaldziuba03, line 33: as I decided to make retryCondition callback configurable by user, they may want to "extend" retry condition like that:

```ts
import { Novu, defaultRetryCondition } from '@novu/node';

function myCustomRetryCondition(err) {
  if (err.response.status === 404) return true;

  return defaultRetryCondition(err);
}

const novu = new Novu('my-key', {
   retryConfig: {
     retryMax: 10,
     retryCondition: myCustomRetryCondition
   }
});
```

packages/node/src/lib/retries.ts
```diff
@@ -0,0 +1,86 @@
...
+    const idempotencyKey = axiosConfig.headers['Idempotency-Key'];
+    // that means intercepted request is retried, so don't generate new idempotency key
+    if (idempotencyKey) {
+      return axiosConfig;
+    }
+
+    axiosConfig.headers['Idempotency-Key'] = uuid();
+
+    return axiosConfig;
+  });
```
michaldziuba03, line 30: I'm using axios interceptor to attach idempotency key header to each request.

Idempotency key is attached only to `POST` and `PATCH` methods because they are **not** idempotent by definition.

packages/node/src/lib/retries.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const RETRIABLE_HTTP_CODES = [408, 429];
+
+export function defaultRetryCondition(err: AxiosError): boolean {
+  // retry on TCP/IP error codes like ECONNRESET
+  if (isNetworkError(err)) {
+    return true;
+  }
```
michaldziuba03, line 67: `isNetworkError` is imported from `axios-retry` library.

As we know HTTP is based on TCP/IP protocol and sometimes we facing issues on this network level.

TCP/IP error typically contains some `code` like `"ECONNRESET"` and `isNetworkError` uses under the hood library [is-retry-allowed](https://github.com/sindresorhus/is-retry-allowed), because not every network error should be retried.

packages/node/src/lib/retries.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  if (err.response.status >= 500 && err.response.status <= 599) {
+    return true;
+  }
+
+  if (RETRIABLE_HTTP_CODES.includes(err.response.status)) {
+    return true;
+  }
```
michaldziuba03, line 83: On HTTP level, I decided to retry on every `5xx` error, `408` (Request Timeout) and `429` (Too Many Requests)

packages/node/src/lib/retries.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        waitMin: 1,
+        waitMax: 1,
+        retryMax: 6,
+      },
+    });
+
+    const result = await novuClient.topics.list({});
+    expect(result.status).toEqual(200);
+  });
+});
```
michaldziuba03, line 105: I think this test will be seperated from unit tests, because it's more like integration test + it takes some time, mainly because number of retries and backoff strategy.

To make it more reliable, I decided to use `nock` library to mock server behavior - it makes easy to simulate http and network errors.

packages/node/package.json
```diff
@@ -77,7 +81,7 @@
     "preset": "ts-jest",
     "testEnvironment": "node",
     "moduleNameMapper": {
-      "axios": "axios/dist/node/axios.cjs"
+      "^axios$": "axios/dist/node/axios.cjs"
```
michaldziuba03, line 84: without that change, tests for retries don't work.

I observed strange behavior - it sends GET request to `127.0.0.1:80` even when you send POST request to baseUrl, it messes up the request config.

change to `^axios$` fixed that issue. This is also recommendation inside [nock readme for axios](https://github.com/nock/nock#axios).

I also tested manually using symlink in testing repo's `node_modules` and retries work as expected.

packages/node/src/lib/retries.ts
Cliftonz: You should also retry on 422 but regen the idempotency key

Cliftonz:
@michaldziuba03 Great overview of the PR!

packages/node/src/lib/retries.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (
+        error.response?.status === 422 &&
+        requestConfig.headers &&
+        requestConfig.method &&
+        NON_IDEMPOTENT_METHODS.includes(requestConfig.method)
+      ) {
+        requestConfig.headers['Idempotency-Key'] = uuid();
+      }
+    },
+  });
```
michaldziuba03, line 68: regenerating idempotency key for HTTP status `422 Unprocessable Content`

michaldziuba03:
@Cliftonz @scopsy  PR is ready for judgment :)

packages/node/package.json
```diff
@@ -43,19 +43,23 @@
   },
   "dependencies": {
     "@novu/shared": "^0.20.0-alpha.0",
+    "axios-retry": "^3.8.0",
     "handlebars": "^4.7.7",
     "lodash.get": "^4.4.2",
-    "lodash.merge": "^4.6.2"
+    "lodash.merge": "^4.6.2",
+    "uuid": "^9.0.1"
```
dammy001, line 50: @michaldziuba03 `UUID` is already part of node internals. You can just import `import { randomUUID } from 'node:crypto'` instead of adding a new package

packages/node/src/index.ts
dammy001: Great!

packages/node/package.json
michaldziuba03: yes, I know but not sure about the support of `randomUUID` in older Node.js versions.

https://nodejs.org/api/crypto.html#cryptorandomuuidoptions

It was added in `v15.6.0, v14.17.0` and in `package.json` technically SDK supports Node.js 10+
```json
  "engines": {
    "node": ">=10"
  },
```

dammy001: Oh wow. I think Novu maintainers need to bump the node version to 14 at least for the package.

dammy001: ...And drop support for node 10

packages/node/src/lib/retries.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      to: { subscriberId: '123' },
+      payload: {},
+    });
+
+    expect(allEqual(idepotencyKeys)).toBeFalsy();
+    expect(result.status).toEqual(201);
+    expect(result.request.headers['idempotency-key']).toBeDefined();
+  });
+
+  it('should retry getting topics list', async () => {
```
dammy001, line 86:
```suggestion
  it('should not retry getting topics list', async () => {
```

michaldziuba03: Not really

rifont (changes requested):

packages/node/src/lib/retries.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export function defaultRetryCondition(err: AxiosError): boolean {
+  // retry on TCP/IP error codes like ECONNRESET
+  if (isNetworkError(err)) {
+    return true;
+  }
+
+  if (!err.response) {
+    return true;
+  }
```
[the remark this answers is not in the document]
michaldziuba03: I think it's mistake (should return `false`) - if statements after this check are dependent on `err.response` (possibly undefined). In new changes I made it look cleaner.

packages/node/src/lib/retries.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+      .reply(404, { message: 'Not Found' });
+
+    nock(BACKEND_URL).get(TOPICS_PATH).reply(200, [{}, {}]);
+
+    try {
+      await novu.topics.list({});
+      throw new Error();
+    } catch (err) {
+      expect((err as AxiosError).response?.status).toEqual(404);
+    }
```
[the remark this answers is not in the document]
michaldziuba03: suggestion didn't work but I got that idea and I made it bit different.

packages/node/src/lib/retry.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      const result = await novu.trigger('fake-workflow', {
+        to: { subscriberId: '123' },
+        payload: {},
+      });
+
+      idempotencyKeys.push(result.request?.headers['idempotency-key']);
+    }
+
+    expect(allEqual(idempotencyKeys)).toEqual(false);
+  });
```
michaldziuba03, line 82: new test case - I want to make sure that idempotency key is different between **requests**

packages/node/src/lib/retry.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    async (status, message) => {
+      nock(BACKEND_URL).get(TOPICS_PATH).times(3).reply(status, { message });
+      nock(BACKEND_URL).get(TOPICS_PATH).reply(200, [{}, {}]);
+
+      await expect(novu.topics.list({})).rejects.toMatchObject({
+        response: { status },
+      });
+    }
+  );
+
```
michaldziuba03, line 160: Not sure if I were supposed to use jest-loop to test *non-recoverable* or *retryable* 4xx errors. In new changes I made it for non recoverable 4xx errors.

packages/node/src/lib/retry.ts
```diff
@@ -0,0 +1,94 @@
...
+import { v4 as uuid } from 'uuid';
+import { INovuConfiguration } from './novu.interface';
+
+const NON_IDEMPOTENT_METHODS = ['post', 'patch'];
+const IDEMPOTENCY_KEY = 'Idempotency-Key'; // header key
+
+const DEFAULT_RETRY_MAX = 0;
+const DEFAULT_WAIT_MIN = 1;
+const DEFAULT_WAIT_MAX = 30;
+
```
michaldziuba03, line 12: new global variables - for default values as suggested + I decided to move string value for header key to `IDEMPOTENCY_KEY` variable (to avoid "magic strings")

michaldziuba03:
@rifont

packages/node/src/lib/retry.ts
```diff
@@ -0,0 +1,94 @@
...
+        requestConfig.method &&
+        NON_IDEMPOTENT_METHODS.includes(requestConfig.method)
+      ) {
+        requestConfig.headers[IDEMPOTENCY_KEY] = uuid();
+      }
+    },
+  });
+}
+
+const RETRYABLE_HTTP_CODES = [408, 429, 422];
```
rifont, line 73:
```suggestion
const RETRYABLE_HTTP_CODES = [408, 422, 429];
```

Prefer numerical order for readability.

packages/node/src/lib/retry.ts
```diff
@@ -0,0 +1,94 @@
...
+  ) {
+    return true;
+  }
+
+  if (err.response && RETRYABLE_HTTP_CODES.includes(err.response.status)) {
+    return true;
+  }
+
+  return false;
+}
```
rifont, line 94: We have some great integration tests covering the non-recoverable error scenarios.

Could we also add a unit test for the `defaultRetryCondition`? Would like to see the following scenarios covered:
* jest loop over the 4xx retryable status codes
* a single test for a 500 error
* a single test for an ECONNRESET error

This way we have all feature entry points covered.

rifont (changes requested):

packages/node/src/lib/retry.spec.ts
```diff
@@ -0,0 +1,205 @@
+import nock from 'nock';
+import { Novu } from '../index';
+
+const BACKEND_URL = 'http://example.com';
+const TOPICS_PATH = '/v1/topics';
+const TRIGGER_PATH = '/v1/events/trigger';
+
+jest.setTimeout(15000);
+
+const allEqual = (arr: Array<string>) => arr.every((val) => val === arr[0]);
```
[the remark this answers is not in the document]
michaldziuba03: I disagree with second sentence but I understand where problem is. Function `allEqual` (currently renamed to `hasAllEqual`) works well when our goal is to check if all items are equal (it iterates over all elements and compares them to the first element, it also compares first element with itself). This is the goal when we check in tests if idempotency-keys are equal between retries (in the scope of same request).

I agree this function is used in wrong way when we check if idempotency-keys are unique between requests - we expect `allEqual`  to return `false`. Negation of this function does not guarantee that all items are unique, so this expectation will pass in case of `[1,2,2,2,2]` scenario, but shouldn't.

I decided to let them coexist :)

packages/node/src/lib/retry.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+      const arr = ['a', 'b', 'c', 'd'];
+      expect(hasUniqueOnly(arr)).toEqual(true);
+    });
+
+    it('should return false when items are not unique', () => {
+      const arr = ['a', 'a', 'c', 'd'];
+      expect(hasUniqueOnly(arr)).toEqual(false);
+    });
+  });
+});
```
rifont, line 275: Excellent tests 👍

rifont (approved):
Great job @michaldziuba03 🙌

packages/node/src/lib/retry.spec.ts
rifont: hey @michaldziuba03 , just a few [spellcheck issues](https://github.com/novuhq/novu/actions/runs/6927540924/job/18841594745?pr=4462) to resolve then we are good to merge 🙌

michaldziuba03: @rifont done

michaldziuba03:
@rifont we are waiting for something before merge? Spellcheck issues were resolved week ago.

Let me know if there is something else to do.

rifont:
> @rifont we are waiting for something before merge? Spellcheck issues were resolved week ago.
> 
> Let me know if there is something else to do.

Hi @michaldziuba03 , we had some failing tests on `next` that we were seeking to fix. However we will now fix these separately. I will get this merged today. Thanks!

PR #4466, opened 2023-10-09
Create README-PT.md

Pedro-Ivo-Neves:

I added your readme file in portuguese.

I think soo, I know some friends that struggle reading in English and would want to use Novu.


scopsy:
Hi @Pedro-Ivo-Neves thank you so much for the contribution, but due to the maintenance, we are not adding more languages to the readme section as of now. Would love to reach out once this will become relevant 🙏

PR #4475, opened 2023-10-10
Feat : add readiness health check retry

djabarovgeorge:
Add validation in order to check isHealth is ready and not paused.
In addition, we want to make sure that the validation in the App is the same as in the hosting infra.

In order to not have race conditions between the app health check in the bootstrap and the hosting infra


packages/application-generic/src/services/readiness/readiness.service.ts
djabarovgeorge, line 1: If anyone can think of a cleaner solution i am up to it! 👍🏻

packages/application-generic/src/services/readiness/readiness.service.ts
```diff
@@ -24,16 +25,36 @@ const LOG_CONTEXT = 'ReadinessService';
...
-    private workflowQueueServiceHealthIndicator: WorkflowQueueServiceHealthIndicator
+    private workflowQueueServiceHealthIndicator: WorkflowQueueServiceHealthIndicator,
+    private subscriberProcessQueueHealthIndicator: SubscriberProcessQueueHealthIndicator
   ) {}
 
   async areQueuesEnabled(): Promise<boolean> {
     Logger.log('Enabling queues as workers are meant to be ready', LOG_CONTEXT);
 
+    const maxRetries = 10;
+    const delay = 1000; // 1 seconds
```
djabarovgeorge, line 36: @scopsy are you aware what params we should set here?

packages/application-generic/src/services/readiness/readiness.service.ts
```diff
@@ -85,6 +106,12 @@ export class ReadinessService {
           throw error;
         }
       }
+    } else {
+      const error = new Error('Queues are not enabled');
+
+      Logger.error(error, 'Queues are not enabled', LOG_CONTEXT);
+
+      throw error;
```
djabarovgeorge, line 114: In case the service is unhealthy we should throw exceptions so we won't stick with zombie instances.

packages/application-generic/src/health/workflow-queue.health-indicator.ts
```diff
@@ -19,8 +19,9 @@ export class WorkflowQueueServiceHealthIndicator extends HealthIndicator {
 
   async isHealthy(): Promise<HealthIndicatorResult> {
     const isReady = this.workflowQueueService.isReady();
+    const isPaused = this.workflowQueueService.isPaused();
```
p-fernandez, line 22: This is a promise. You need to await it.

djabarovgeorge: True, strange that I did not get any warning about it in the editor

packages/application-generic/src/services/readiness/readiness.service.ts
```diff
@@ -24,16 +25,36 @@ const LOG_CONTEXT = 'ReadinessService';
...
+
+    for (let i = 0; i < maxRetries; i++) {
+      const result = await this.checkServicesHealth();
+
+      if (result) {
+        return true;
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-loop-func
+      await new Promise((resolve) => setTimeout(resolve, delay));
```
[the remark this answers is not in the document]
djabarovgeorge: Interesting to tell the truth I have never used it :)

packages/application-generic/src/health/standard-queue.health-indicator.ts
```diff
@@ -33,4 +33,22 @@ export class StandardQueueServiceHealthIndicator extends HealthIndicator {
       this.getStatus(this.INDICATOR_KEY, false)
     );
   }
+
+  async isWorkerHealthy(): Promise<HealthIndicatorResult> {
```
p-fernandez, line 37: I have concern with the name. This checks the queue situation, not the worker. So for me the name is misleading. Not sure if provides more value than the `isHealthy` method. Besides that the duplication logs the same without making any difference so it could be misleading.

djabarovgeorge: I see what you mean, what do you think about "is active"? do you have any other suggestions?
@p-fernandez

PR #4489, opened 2023-10-11
Update tsconfig.json

DHRUVKADAM22:
1. Add Paths for Module Resolution: You can configure paths to help with module resolution. For example, if you have source code in a src directory, you can add a path to simplify imports

This way, you can import modules from your project like this: import { someModule } from '@/module';

2. Set Target and Module: Specify the ECMAScript target version and module system. Depending on your project's requirements, you can set

3. Remove Unused Types: If your project doesn't use all the specified types, consider removing the unnecessary ones from the types array. This can reduce unnecessary type checking.

4. OutDir and RootDir: If your project structure requires it, you can set the outDir and rootDir options to manage the output directory and the root directory of input files. For example

5. Optimize strict Mode: Enabling strict mode ("strict": true) enforces strict type checking. It's a good practice, but if you encounter performance issues during development, you can disable some strict checks temporarily or adjust specific options.

6. Use Declaration Map (d.ts Map): You can generate declaration maps to improve code editor and IDE performance when working with TypeScript

7. Remove Trailing Commas: Ensure there are no trailing commas in your JSON. While trailing commas are allowed in JSON, some tools and linters may flag them.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,14 @@
     "moduleResolution": "node",
     "baseUrl": ".",
     "paths": {
-    }
+      "@/*": ["src/*"]
+    },
+    "target": "ESNext",
+    "module": "CommonJS",
+    "strict": false,
+    "declarationMap": true,
+    "outDir": "./dist",
+    "rootDir": "./src"
   },
   "types": [
     "node",


scopsy:
Thank you for the contribution @DHRUVKADAM22 but please submit an issue with the reason for the change before submitting a PR.

Regarding this PR it was an intentional why of not working with modules like this. Also the extra items added are already existing in the base config it extends from .

DHRUVKADAM22:
Thanks for your guidance @scopsy and sorry for than extra items added which were already present but still looking forward for contribution in NOVU

PR #4495, opened 2023-10-11
feat(worker): handle different push flows with better exec details

p-fernandez:

Improves the handling of different push notifications messages sending:
- Doesn't throw anymore breaking the loop so cases of multiple push providers are handled properly.
- Increases the details in the activity feed for the cases of missing configuration, missing device tokens, when there is an error in one or many of multiple providers active.
- Creates a new notification error detail for the summary card of the execution details.

Code wasn't handling properly cases where multiple Push providers were active at the very same time.
Also there was no test suite for sending push notifications so it was added.

Screenshot 2023-10-11 at 16 03 10
Screenshot 2023-10-11 at 16 04 15

### DIFF
--- a/apps/api/src/app/events/e2e/send-message-push.e2e.ts
+++ b/apps/api/src/app/events/e2e/send-message-push.e2e.ts
@@ -1,0 +1,208 @@
+import axios from 'axios';
+import { expect } from 'chai';
+import {
+  ExecutionDetailsRepository,
+  IntegrationRepository,
+  MessageRepository,
+  NotificationTemplateEntity,
+} from '@novu/dal';
+import { DetailEnum } from '@novu/application-generic';
+import { ChannelTypeEnum, PushProviderIdEnum, StepTypeEnum } from '@novu/shared';
+import { UserSession } from '@novu/testing';
+
+const axiosInstance = axios.create();
+
+describe('Trigger event - Send Push Notification - /v1/events/trigger (POST)', () => {
+  let session: UserSession;
+  let template: NotificationTemplateEntity;
+
+  const executionDetailsRepository = new ExecutionDetailsRepository();
+  const integrationRepository = new IntegrationRepository();
+  const messageRepository = new MessageRepository();
+
+  before(async () => {
+    process.env.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED = 'true';
+
+    session = new UserSession();
+    await session.initialize();
+
+    template = await session.createTemplate({
+      steps: [
+        {
+          active: true,
+          type: StepTypeEnum.PUSH,
+          title: 'Title',
+          content: 'Welcome to {{organizationName}}' as string,
+        },
+      ],
+    });
+  });
+
+  after(() => {
+    process.env.IS_MULTI_PROVIDER_CONFIGURATION_ENABLED = 'false';
+  });
+
+  describe('Multiple providers active', () => {
+    before(async () => {
+      const payload = {
+        providerId: PushProviderIdEnum.EXPO,
+        channel: ChannelTypeEnum.PUSH,
+        credentials: { apiKey: '123' },
+        _environmentId: session.environment._id,
+        active: true,
+        check: false,
+      };
+
+      await session.testAgent.post('/v1/integrations').send(payload);
+
+      const integrations = await integrationRepository.find({
+        _environmentId: session.environment._id,
+        channel: ChannelTypeEnum.PUSH,
+        active: true,
+      });
+
+      expect(integrations.length).to.equal(2);
+    });
+
+    afterEach(async () => {
+      await executionDetailsRepository.delete({ _environmentId: session.environment._id });
+    });
+
+    it('should not create any message if subscriber has no configured channel', async () => {
+      await triggerEvent(session, template);
+
+      await session.awaitRunningJobs(template._id);
+
+      const messages = await messageRepository.find({
+        _environmentId: session.environment._id,
+        _templateId: template._id,
+        _subscriberId: session.subscriberId,
+      });
+
+      expect(messages.length).to.equal(0);
+
+      const executionDetails = await executionDetailsRepository.find({
+        _environmentId: session.environment._id,
+      });
+
+      expect(executionDetails.length).to.equal(5);
+      const noActiveChannel = executionDetails.find((ex) => ex.detail === DetailEnum.SUBSCRIBER_NO_ACTIVE_CHANNEL);
+      expect(noActiveChannel).to.be.ok;
+      expect(noActiveChannel?.providerId).to.equal('fcm');
+      const genericError = executionDetails.find((ex) => ex.detail === DetailEnum.NOTIFICATION_ERROR);
+      expect(genericError).to.be.ok;
+    });
+
+    it('should not create any message if subscriber has configured two providers without device tokens', async () => {
+      await updateCredentials(session, session.subscriberId, PushProviderIdEnum.FCM, []);
+      await updateCredentials(session, session.subscriberId, PushProviderIdEnum.EXPO, []);
+
+      await triggerEvent(session, template);
+
+      await session.awaitRunningJobs(template._id);
+
+      const messages = await messageRepository.find({
+        _environmentId: session.environment._id,
+        _templateId: template._id,
+        _subscriberId: session.subscriberId,
+      });
+
+      expect(messages.length).to.equal(0);
+
+      const executionDetails = await executionDetailsRepository.find({
+        _environmentId: session.environment._id,
+      });
+
+      expect(executionDetails.length).to.equal(6);
+      const fcm = executionDetails.find(
+        (ex) => ex.detail === DetailEnum.PUSH_MISSING_DEVICE_TOKENS && ex.providerId === PushProviderIdEnum.FCM
+      );
+      expect(fcm).to.be.ok;
+      const expo = executionDetails.find(
+        (ex) => ex.detail === DetailEnum.PUSH_MISSING_DEVICE_TOKENS && ex.providerId === PushProviderIdEnum.EXPO
+      );
+      expect(expo).to.be.ok;
+      const genericError = executionDetails.find((ex) => ex.detail === DetailEnum.NOTIFICATION_ERROR);
+      expect(genericError).to.be.ok;
+    });
+
+    it('should not create any message if subscriber has configured one provider without device tokens and the other has invalid device token', async () => {
+      await updateCredentials(session, session.subscriberId, PushProviderIdEnum.FCM, ['invalidDeviceToken']);
+      await updateCredentials(session, session.subscriberId, PushProviderIdEnum.EXPO, []);
+
+      await triggerEvent(session, template);
+
+      await session.awaitRunningJobs(template._id);
+
+      const messages = await messageRepository.find({
+        _environmentId: session.environment._id,
+        _templateId: template._id,
+        _subscriberId: session.subscriberId,
+      });
+
+      expect(messages.length).to.equal(0);
+
+      const executionDetails = await executionDetailsRepository.find({
+        _environmentId: session.environment._id,
+      });
+
+      expect(executionDetails.length).to.equal(8);
+      const fcmMessageCreated = executionDetails.find(
+        (ex) =>
+          ex.detail === `${DetailEnum.MESSAGE_CREATED}: ${PushProviderIdEnum.FCM}` &&
+          ex.providerId === PushProviderIdEnum.FCM
+      );
+      expect(fcmMessageCreated).to.be.ok;
+      const fcmProviderError = executionDetails.find(
+        (ex) => ex.detail === DetailEnum.PROVIDER_ERROR && ex.providerId === PushProviderIdEnum.FCM
+      );
+      expect(fcmProviderError).to.be.ok;
+
+      const expo = executionDetails.find(
+        (ex) => ex.detail === DetailEnum.PUSH_MISSING_DEVICE_TOKENS && ex.providerId === PushProviderIdEnum.EXPO
+      );
+      expect(expo).to.be.ok;
+      const genericError = executionDetails.find((ex) => ex.detail === DetailEnum.NOTIFICATION_ERROR);
+      expect(genericError).to.be.ok;
+    });
+  });
+});
+
+async function triggerEvent(session: UserSession, template: NotificationTemplateEntity) {
+  await axiosInstance.post(
+    `${session.serverUrl}/v1/events/trigger`,
+    {
+      name: template.triggers[0].identifier,
+      to: [{ subscriberId: session.subscriberId }],
+      payload: {},
+    },
+    {
+      headers: {
+        authorization: `ApiKey ${session.apiKey}`,
+      },
+    }
+  );
+}
+
+async function updateCredentials(
+  session: UserSession,
+  subscriberId: string,
+  providerId: PushProviderIdEnum,
+  deviceTokens: string[]
+) {
+  await axiosInstance.put(
+    `${session.serverUrl}/v1/subscribers/${subscriberId}/credentials`,
+    {
+      subscriberId,
+      providerId,
+      credentials: {
+        deviceTokens,
+      },
+    },
+    {
+      headers: {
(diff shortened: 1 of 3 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
         Object.values(PushProviderIdEnum).includes(chan.providerId as PushProviderIdEnum)
       ) || [];
 
-    const messagePayload = Object.assign({}, command.payload);
-    delete messagePayload.attachments;
-
     if (!pushChannels.length) {
-      await this.sendErrors(command);
```
p-fernandez, line 120: Moved above to try to fail as soon as possible and because the calculations for pushChannels are now together.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
...
+      await this.sendNoActiveChannelError(command.job);
+      await this.sendNotificationError(command.job);
 
       return;
     }
 
-    for (const channel of pushChannels) {
-      if (!channel.credentials?.deviceTokens) {
```
p-fernandez, line 126: Empty device tokens arrays were not captured here.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
...
-          CreateExecutionDetailsCommand.create({
-            ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
-            detail: DetailEnum.SUBSCRIBER_NO_ACTIVE_INTEGRATION,
-            source: ExecutionDetailsSourceEnum.INTERNAL,
-            status: ExecutionDetailsStatusEnum.FAILED,
-            isTest: false,
-            isRetry: false,
-          })
-        );
+      const [isChannelMissingDeviceTokens, integration] = await Promise.all([
```
p-fernandez, line 137: Now we can execute this paralelly as we do not break the loop anymore, nor neither want to break it.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
...
+      }
+    }
+
+    if (integrationsWithErrors > 0) {
+      Logger.error(
+        { jobId: command.jobId },
+        `There was an error sending the push notification(s) for the jobId ${command.jobId}`,
+        LOG_CONTEXT
+      );
+      await this.sendNotificationError(command.job);
```
p-fernandez, line 175: This is a new execution detail that throws so the execution details summary will show a failure.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -196,6 +225,53 @@ export class SendMessagePush extends SendMessageBase {
...
+    job: JobEntity,
+    contextData?: {
+      messageId?: string;
+      providerId?: ProvidersIdEnum;
+      raw?: string;
+    }
+  ): Promise<void> {
+    // We avoid to throw the errors to be able to execute all actions in the loop
+    try {
+      await this.createExecutionDetails.execute(
```
p-fernandez, line 259: Unfortunately downstream any `ExecutionDetailsStatusEnum.FAILED` will throw an error. I find it is not a good thing but changing that would affect in all channels so it is better to handle it in a different PR.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
...
       );
+
+      if (!result) {
+        integrationsWithErrors++;
+      }
+    }
+
+    if (integrationsWithErrors > 0) {
+      Logger.error(
+        { jobId: command.jobId },
```
djabarovgeorge, line 171:
```suggestion
        { jobId: command.jobId ,
        errorCount: integrationsWithErrors},
```
Should we add the count as well?

apps/api/src/app/events/e2e/send-message-push.e2e.ts
```diff
@@ -0,0 +1,207 @@
...
+        _environmentId: session.environment._id,
+        active: true,
+        check: false,
+      };
+
+      await session.testAgent.post('/v1/integrations').send(payload);
+
+      const integrations = await integrationRepository.find({
+        _environmentId: session.environment._id,
+        channel: ChannelTypeEnum.PUSH,
```
djabarovgeorge, line 60: Should we add active:true flag?

apps/api/src/app/events/e2e/send-message-push.e2e.ts
```diff
@@ -0,0 +1,207 @@
...
+      const expo = executionDetails.find(
+        (ex) => ex.detail === DetailEnum.PUSH_MISSING_DEVICE_TOKENS && ex.providerId === PushProviderIdEnum.EXPO
+      );
+      expect(expo).to.be.ok;
+      const genericError = executionDetails.find((ex) => ex.detail === DetailEnum.NOTIFICATION_ERROR);
+      expect(genericError).to.be.ok;
+    });
+
+    it('should not create any message if subscriber has configured one provider without device tokens and the other has wrong credentials', async () => {
+      await updateCredentials(session, session.subscriberId, PushProviderIdEnum.FCM, ['deviceToken']);
```
djabarovgeorge, line 129: how come this one is an wrong credential?

djabarovgeorge (commented):
Looks amazing to me 💪, left some comments there.

the most relevant comments imo were here https://github.com/novuhq/novu/pull/4495/files#r1355595433

apps/worker/src/app/workflow/usecases/send-message/send-message.base.ts
```diff
@@ -27,6 +27,8 @@ import {
 import { SendMessageType } from './send-message-type.usecase';
 import { CreateLog } from '../../../shared/logs';
 
+const LOG_CONTEXT = 'SendMessageBase';
+
```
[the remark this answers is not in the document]
p-fernandez: I used that for debugging logging purposes. I can delete.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
p-fernandez: Will it be helpful the count number? I added this more thinking in a potential NR dashboard monitoring the failures in the Push channel notifications. For debugging goals the execution details logs should be enough unless I missed any case!

apps/api/src/app/events/e2e/send-message-push.e2e.ts
p-fernandez: It will protect this from failures. Good idea.

apps/api/src/app/events/e2e/send-message-push.e2e.ts
p-fernandez: Because the it is not a valid device token. I will rename it to avoid confusion.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
...
-            ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
-            detail: DetailEnum.SUBSCRIBER_NO_ACTIVE_INTEGRATION,
-            source: ExecutionDetailsSourceEnum.INTERNAL,
-            status: ExecutionDetailsStatusEnum.FAILED,
-            isTest: false,
-            isRetry: false,
-          })
-        );
+      const [isChannelMissingDeviceTokens, integration] = await Promise.all([
+        this.isChannelMissingDeviceTokens(channel, command),
```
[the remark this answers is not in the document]
p-fernandez: Unfortunately I need the channel (interface IChannelSettings) to provide the right context to the execution details stored in `sendPushMissingDeviceTokensError`. I can't use the Job providerId because for more than one active push provider it is only storing one of them.
That's another bug we would need to review in a different task that might come from lack of knowledge in the team that for Push multiple providers could be active. 😕

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
p-fernandez: I considered it while thinking how we should manage all these cases but I went this direction to show everything that is failing in once.

Now that you raise the number of calls we need to make, thinking about it I can see this solution should executing less calls in an error case situation based on this generic scenario:
- All the amount of db calls we do to trigger a notification (let's call it N pre-calls).
- In SendMessagePush we fail because missing device tokens.
- User fixes that and sends again N pre-calls.
- In SendMessagePush the Redis SelectIntegration call returns an user error. (+1 call)
- User fixes that and sends again N pre-calls.
- Notification is being sent successfully with SendMessagePush X post-calls.

vs

- N pre-calls
- In SendMessagePush we fail because missing device tokens and select integration (+1 call)
- User fixes those and sends again N pre-calls.
- Notification is being sent successfully with SendMessagePush X post-calls.

For this kind of scenario makes:
3 * N pre-calls + 1 + X vs 2 * N pre-calls + 1 + X

Also the less trigger events generated the less load our systems will have and the less the customers will pay.

Maybe if we start to add more conditions on that phase of the flow we can incur in extra SelectIntegration calls while are only other preconditions failing. So I can't predict in total numbers how our users will not set properly the configuration, but I am guessing that statistically should be better to do an extra call that having to do a set of N pre-calls if the select integration brings that the subscriber got no integrations assigned.
I would be happy to brainstorm the flow with you to improve and make it more performant if you want.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
...
+        this.getSubscriberIntegration(channel, command),
+      ]);
 
+      // We avoid to send a message if subscriber has not an integration or if the subscriber has no device tokens for said integration
+      if (!deviceTokens || !integration || isChannelMissingDeviceTokens) {
```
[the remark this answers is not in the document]
p-fernandez: Adding it here was purely to solve a TS warning in line 155 when accessing `channel.credentials.deviceTokens` as credentials was potentially undefined.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
djabarovgeorge: I see what you mean here and it is a possible use case that I was not thinking of.
The reason I raised this comment was because this 'error' is on the job-subscriber-channel level and not the whole notification (all the workflow flow). 
If there is a case where the user sees that he has X subscribers that failed to send push messages from N subscribers, IMO he should resent the notification only for X subscribers and potentially only for this step, otherwise these subscribers will get duplicated messages.

Anyway, those two cases should be pretty rare so I am good with any of them :)

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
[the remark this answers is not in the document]
djabarovgeorge: missed that one 🤦‍♂️
good point!!

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
djabarovgeorge: no, i think it should be fine, I was not sure if needed or not. we could add it in the future if we find it helpful.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -113,81 +120,103 @@ export class SendMessagePush extends SendMessageBase {
         Object.values(PushProviderIdEnum).includes(chan.providerId as PushProviderIdEnum)
       ) || [];
 
-    const messagePayload = Object.assign({}, command.payload);
-    delete messagePayload.attachments;
-
     if (!pushChannels.length) {
-      await this.sendErrors(command);
+      await this.sendNoActiveChannelError(command.job);
+      await this.sendNotificationError(command.job);
```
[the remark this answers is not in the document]
djabarovgeorge: I forgot to raise this before, I am not sure if we need to use the word `notification` as it is in NOTIFICATION_ERROR because we are in the `job` level here and not in the notification level, the notification should cover all of the workflow.

regarding the comment:
If we want to add a summary we should add it to the run-job use case in its catch block so we would add a summary to all of the potential cases, but I am not sure it is necessary addition to that we do not add there any data besides that there was an undefined errors. let me know what you think about it.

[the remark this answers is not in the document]
p-fernandez: > I forgot to raise this before, I am not sure if we need to use the word `notification` as it is in NOTIFICATION_ERROR because we are in the `job` level here and not in the notification level, the notification should cover all of the workflow.
> 
I agree with you in the codebase but UI wise for the user what it fails is the notification for a certain subscriber (or notifications as the Push is being a different case where multiple providers can be in play). I can change the DetailEnum for this case for something more correct codewise though for the UI I find proper to use notification.
Any suggestion for the DetailEnum?


 
> regarding the comment: If we want to add a summary we should add it to the run-job use case in its catch block so we would add a summary to all of the potential cases, but I am not sure it is necessary addition to that we do not add there any data besides that there was an undefined errors. let me know what you think about it.

Currently Activity Feed takes the last execution details to show in the summary for the card for every subscriber and notification for all channels. The different flow of Push made me care about if something fails to have that execution details to tell "Hey look inside something was wrong" without having to think in all potential combinations.

It is true that for the case of NoActiveChannelError might not be needed, but I tried to be consistent for the behaviour of Push in the ActivityFeed. I can change it making `sendNoActiveChannelError` to throw and removing `sendNotificationError` if you think it is better for the user awareness in the UI. 

I would avoid moving that responsibility to RunJob as that triggers the different channels in different ways (email and SMS has single provider per node with multi provider store, chat and push have no multi provider store but can in the same node multiple active providers).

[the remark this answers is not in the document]
djabarovgeorge: > I agree with you in the codebase but UI wise for the user what it fails is the notification for a certain subscriber (or notifications as the Push is being a different case where multiple providers can be in play). I can change the DetailEnum for this case for something more correct codewise though for the UI I find proper to use notification.
Any suggestion for the DetailEnum?

I see the reasoning there now, thanks for the explanation. i think we can keep it like this.

> Currently Activity Feed takes the last execution details to show in the summary for the card for every subscriber and notification for all channels. The different flow of Push made me care about if something fails to have that execution details to tell "Hey look inside something was wrong" without having to think in all potential combinations.

If the concern is in the UI maybe we could add this message generated in the UI based on the data. wdyt?

> It is true that for the case of NoActiveChannelError might not be needed, but I tried to be consistent for the behaviour of Push in the ActivityFeed. I can change it making sendNoActiveChannelError to throw and removing sendNotificationError if you think it is better for the user awareness in the UI.

the NoActiveChannelError is needed and provides more information, but the addition sendNotificationError does not provide additional information to the debugging.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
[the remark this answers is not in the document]
djabarovgeorge: oh i see, i love those 🤦‍♂️

djabarovgeorge (approved):

scopsy (approved):

PR #4502, opened 2023-10-12
feat: add variant flow

BiswaViraj:

- add variants flow for the editorSidebarHeaderActions

### DIFF
--- a/apps/web/src/pages/templates/components/EditorSidebarHeaderActions.tsx
+++ b/apps/web/src/pages/templates/components/EditorSidebarHeaderActions.tsx
@@ -1,22 +1,22 @@
 import { Group } from '@mantine/core';
-import { useState } from 'react';
+import { FilterPartTypeEnum } from '@novu/shared';
+import { useEffect, useRef, useState } from 'react';
 import { useFormContext } from 'react-hook-form';
 import { useLocation, useNavigate, useParams } from 'react-router-dom';
-import { FilterPartTypeEnum } from '@novu/shared';
 
 import { Conditions, IConditions } from '../../../components/conditions';
+import { When } from '../../../components/utils/When';
+import { ActionButton } from '../../../design-system/button/ActionButton';
 import { Condition, ConditionPlus, ConditionsFile, Trash, VariantPlus } from '../../../design-system/icons';
+import { useEnvController } from '../../../hooks';
+import { useBasePath } from '../hooks/useBasePath';
 import { useFilterPartsList } from '../hooks/useFilterPartsList';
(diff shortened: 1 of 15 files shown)


apps/web/src/pages/templates/components/EditorSidebarHeaderActions.tsx
```diff
@@ -38,17 +39,26 @@ export const EditorSidebarHeaderActions = () => {
   const hasNoFilters = (filters && filters?.length === 0) || !filters;
 
   const onAddVariant = () => {
-    const newVariant = addVariant(stepUuid);
-    if (newVariant) {
-      // TODO: show the conditions sidebar first and then when conditions are applied show the variant editor
-      navigate(basePath + `/${newVariant?.template.type}/${stepUuid}/variants/${newVariant.uuid}`);
+    const variant = addVariant(stepUuid);
+    if (variant) {
+      navigate(basePath + `/${variant?.template.type}/${stepUuid}/variants/${variant?.uuid}/conditions`);
```
ainouzgali, line 44: @BiswaViraj what if after navigation to conditions, user clicks cancel. As the variant was already "created" inside the form, wouldn't it be updated as well when updating the workflow?

BiswaViraj: It redirects to the variant editor.
Yes it will create the variant. Without creating the variant it will be difficult to display the conditions screen, as we do show the variant name and also apply the conditions

LetItRock (commented):
-

apps/web/src/pages/templates/components/EditorSidebarHeaderActions.tsx
LetItRock: I do agree with Gali. Maybe we need to have a path like: `/${variant?.template.type}/${stepUuid}/variants/new/conditions` and create the variant on the conditions apply button click?

BiswaViraj:
@LetItRock , The other scenarios are handled in this PR: https://github.com/novuhq/novu/pull/4538

apps/web/src/pages/templates/components/EditorSidebarHeaderActions.tsx
BiswaViraj: We can do that, just thinking how we will handle the Conditions name on the header

LetItRock: The variant is a copy of the root step, so we can take a title from the root and add the prefix `V1` in the beginning, like the same way we create the variant name.

LetItRock (approved):
🚀

PR #4504, opened 2023-10-12
fix: change custom data type and add test in node sdk

jainpawan21:
Recently added customData field was not allowing `{ templateId : "templateId" }` due to typescript type


jainpawan21:
@scopsy this pr is blocker for sendgrid templates using email overrides

scopsy (approved):

PR #4508, opened 2023-10-12
feat(providers): improve expo response management and multi device token

p-fernandez:

Improves the response management of the Expo Provider and finally handles error cases. 
In SendMessagePush we improve the information provided for the responses showing the device token that has failed or succeeded for situations of subscribers with multiple device tokens in the channel integration.

This work is part of the bug discovered by @jainpawan21. Execution Details wasn't showing clearly that some device tokens were failing when using multiple active providers and using multiple device tokens for the same subscriber.
Also the Expo provider implementation had a lot of room to improve and was part to blame of the situation we were having.

Expo provider will still miss multi token management (inherited tech debt).

Screenshots:
Screenshot 2023-10-12 at 16 59 21
Screenshot 2023-10-12 at 16 59 28


apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -148,21 +153,26 @@ export class SendMessagePush extends SendMessageBase {
...
-        content,
-        command,
-        command.payload,
-        overrides,
-        stepData
-      );
-
-      if (!result) {
-        integrationsWithErrors++;
+      const target = (overrides as { deviceTokens?: string[] }).deviceTokens || deviceTokens;
```
p-fernandez, line 156: Extracting the `deviceTokens` out of `this.sendMessage` allowed us to easily loop them reusing the new flow implemented in https://github.com/novuhq/novu/pull/4495.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -148,21 +153,26 @@ export class SendMessagePush extends SendMessageBase {
...
       const overrides = command.overrides[integration.providerId] || {};
-
-      const result = await this.sendMessage(
-        subscriber,
-        integration,
-        deviceTokens,
-        title,
-        content,
-        command,
-        command.payload,
```
p-fernandez, line 159: Also taking advantage of cleaning the function arguments.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -273,59 +283,24 @@ export class SendMessagePush extends SendMessageBase {
   }
 
   private async sendMessage(
+    command: SendMessageCommand,
+    message: MessageEntity,
     subscriber: IPushOptions['subscriber'],
     integration: IntegrationEntity,
-    target: string[],
+    deviceToken: string,
```
p-fernandez, line 290: Now we will miss the functionalities of sending multiple tokens at the same time in exchange of better execution details without doing breaking changes in the providers.
We need to implement auxiliary methods in the providers to handle the multiple token and to handle properly their responses.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -340,7 +315,7 @@ export class SendMessagePush extends SendMessageBase {
           status: ExecutionDetailsStatusEnum.SUCCESS,
           isTest: false,
           isRetry: false,
-          raw: JSON.stringify(result),
+          raw: JSON.stringify({ result, deviceToken }),
```
p-fernandez, line 318: We add the device token in the response, though it can be arguable if we shouldn't as identifies the device of a user. We need to be careful with them.

Some providers treat the device token as not PII, like for example OneSignal:
https://documentation.onesignal.com/docs/handling-personal-data#push-tokens, but we should analyse our own risks.
CC: @scopsy @Cliftonz @djabarovgeorge @jainpawan21

packages/application-generic/src/factories/push/push.factory.ts
```diff
@@ -17,7 +17,7 @@ export class PushFactory implements IPushFactory {
     new PushWebhookHandler(),
   ];
 
-  getHandler(integration: IntegrationEntity) {
+  getHandler(integration: IntegrationEntity): IPushHandler {
```
p-fernandez, line 20: Type wasn't being infered by TypeScript. +1 for team explicit here.

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
```diff
@@ -43,6 +44,8 @@ const LOG_CONTEXT = 'SendMessagePush';
 export class SendMessagePush extends SendMessageBase {
   channelType = ChannelTypeEnum.PUSH;
 
+  private pushFactory;
```
scopsy, line 47: @p-fernandez due to how nestjs works, we cannot have this one here as a private method, as nestjs @Injectable classes are singletons, and this will be reused across multiple executions. Which we don't want We need here to not store it as a method but rather pass it.

scopsy (approved):

apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts
jainpawan21: @scopsy is it ok to proceed?

PR #4513, opened 2023-10-13
♻️ Refactor Create Queue Health Indicator

djabarovgeorge:

Create and use queue health indicator base abstract class.

In order to extract common code to base class so we could reuse it.

### DIFF
--- a/packages/application-generic/src/health/queue-health-indicator.service.ts
+++ b/packages/application-generic/src/health/queue-health-indicator.service.ts
@@ -37,7 +37,7 @@ export abstract class QueueHealthIndicator
       Logger.verbose(`${this.serviceName} is not ready`, this.logContext);
 
       throw new HealthCheckError(
-        `${this.serviceName} Health`,
+        `${this.serviceName} Health is not ready`,
         this.getStatus(this.indicatorKey, false)
       );
     }
@@ -55,7 +55,7 @@ export abstract class QueueHealthIndicator
       Logger.verbose(`${this.serviceName} is not active`, this.logContext);
 
       throw new HealthCheckError(
-        `${this.serviceName} Health`,
+        `${this.serviceName} Health is not active`,
         this.getStatus(this.indicatorKey, false)
       );
     }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17580,8 +17580,47 @@ packages:
       react-refresh: 0.11.0
       schema-utils: 3.3.0
       source-map: 0.7.4
-      webpack: 5.78.0(@swc/core@1.3.49)(esbuild@0.18.20)(webpack-cli@5.1.4)
+      webpack: 5.78.0
       webpack-dev-server: 4.11.1(webpack@5.78.0)
+    dev: true
+
+  /@pmmmwh/react-refresh-webpack-plugin@0.5.10(react-refresh@0.11.0)(webpack@5.78.0):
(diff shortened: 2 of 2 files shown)


packages/application-generic/src/health/queue-health-indicator.service.ts
```diff
@@ -0,0 +1,67 @@
...
+import { QueueBaseService } from '../services';
+import { IHealthIndicator } from './health-indicator.interface';
+
+@Injectable()
+export abstract class QueueHealthIndicator
+  extends HealthIndicator
+  implements IHealthIndicator
+{
+  constructor(
+    private queueBaseService: QueueBaseService,
```
p-fernandez, line 17:
```suggestion
    private queueService: QueueBaseService,
```
Just because we are passing other services that satisfy the QueueBaseService interface.

packages/application-generic/src/health/queue-health-indicator.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    private indicatorKey: string,
+    private serviceName: string,
```
p-fernandez, line 19: Maybe let's think if some enums limiting the values that can be passed here would be something that would benefit us.

packages/application-generic/src/health/queue-health-indicator.service.ts
```diff
@@ -0,0 +1,67 @@
...
+      throw new HealthCheckError(
+        `${this.serviceName} Health`,
+        this.getStatus(this.indicatorKey, false)
+      );
+    }
+
+    Logger.verbose(`${this.serviceName} is active`, this.logContext);
+
+    return this.getStatus(this.indicatorKey, true);
+  }
```
p-fernandez, line 66: I think we can avoid the use of argument flags and with a single function compose both methods in a more maintainable way. What do you think?

```suggestion
  async isHealthy(): Promise<HealthIndicatorResult> {
    const isReady = this.checkIsReady();
     
    Logger.verbose(`${this.serviceName} is ready`, this.logContext);

    return this.getStatus(this.indicatorKey, true);
  }

  async isActive(): Promise<HealthIndicatorResult> {
    const isReady = this.checkIsReady();
    const isPaused = await this.queueBaseService.isPaused();
    const isActive = isReady && !isPaused;

    if (!isActive) {
      Logger.verbose(`${this.serviceName} is not active`, this.logContext);

      throw new HealthCheckError(
        `${this.serviceName} Health`,
        this.getStatus(this.indicatorKey, false)
      );
    }

    Logger.verbose(`${this.serviceName} is active`, this.logContext);

    return this.getStatus(this.indicatorKey, true);
  }
  
  private checkIsReady(): boolean {
      const isReady = this.queueBaseService.isReady();

    if (!isReady) {
      Logger.verbose(`${this.serviceName} is not ready`, this.logContext);

      throw new HealthCheckError(
        `${this.serviceName} Health`,
        this.getStatus(this.indicatorKey, false)
      );
    }
    
    return isReady;
  }
```

p-fernandez (approved):
Top quality work. I am into what you went for.
🌟

packages/application-generic/src/health/queue-health-indicator.service.ts
djabarovgeorge: Now sure I follow, could you elaborate on what you mean here or provide an example? :)

packages/application-generic/src/health/queue-health-indicator.service.ts
djabarovgeorge: Do you mean that one function will check isHealth and isActive altogether?

PR #4526, opened 2023-10-15
Refactor the broadcast method to utilize the new queue

djabarovgeorge:

Refactor the new broadcast method to utilize the new trigger-handler and process-subscriber queues, instead of doing that at the moment of the trigger.


packages/application-generic/src/modules/queues.module.ts
```diff
@@ -67,3 +67,18 @@ const PROVIDERS: Provider[] = [
...
+  WebSocketsQueueService,
+  WebSocketsQueueServiceHealthIndicator,
+  WorkflowQueueService,
+  WorkflowQueueServiceHealthIndicator,
+];
+@Module({
+  providers: [...APP_PROVIDERS],
+  exports: [...APP_PROVIDERS],
+})
+export class BaseAppQueuesModule {}
```
djabarovgeorge, line 84: This is not the final solution but a temporary one, in this part, I extracted the relevant queues, so we wont initialize not relevant queues in the API APP.
Will make a discovery in order to check if we could take advantage of nest dynamic modules in order to make it more convenient.
https://docs.nestjs.com/fundamentals/dynamic-modules

apps/api/src/app/widgets/widgets.module.ts
```diff
@@ -8,7 +10,7 @@ import { SubscribersModule } from '../subscribers/subscribers.module';
 import { IntegrationModule } from '../integrations/integrations.module';
 
 @Module({
-  imports: [SharedModule, SubscribersModule, AuthModule, IntegrationModule],
+  imports: [SharedModule, SubscribersModule, AuthModule, IntegrationModule, BaseAppQueuesModule],
```
djabarovgeorge, line 13: Was needed because we removed QueueModule from the shared module.
Here we need BaseAppQueuesModule because we use WebSocketsQueueService.

apps/api/src/app/shared/shared.module.ts
```diff
@@ -100,9 +100,8 @@ const PROVIDERS = [
         version: packageJson.version,
       })
     ),
-    QueuesModule,
   ],
   providers: [...PROVIDERS],
-  exports: [...PROVIDERS, LoggerModule, QueuesModule],
```
djabarovgeorge, line 106: Not sure why we need this one here therefore I removed it.
If someone knows the reason please let me know.

apps/api/src/app/inbound-parse/inbound-parse.module.ts
```diff
@@ -1,22 +1,21 @@
...
 import { AuthModule } from '../auth/auth.module';
 
 const PROVIDERS = [InboundParseQueueService, GetMxRecord, CompileTemplate];
 
 @Module({
-  imports: [SharedModule, AuthModule, QueuesModule],
+  imports: [SharedModule, AuthModule, BaseAppQueuesModule],
   controllers: [InboundParseController],
   providers: [...PROVIDERS, ...USE_CASES],
-  exports: [...USE_CASES, QueuesModule],
```
djabarovgeorge, line 19: Do we need to export this one here?

apps/api/src/app/health/health.controller.ts
```diff
@@ -26,7 +25,6 @@ export class HealthController {
   healthCheck(): Promise<HealthCheckResult> {
     const checks: HealthIndicatorFunction[] = [
       async () => this.dalHealthIndicator.isHealthy(),
-      async () => this.standardQueueHealthIndicator.isHealthy(),
```
djabarovgeorge, line 29: We no longer need it because we will use only the workflow queue in the API.

apps/api/src/app/events/usecases/process-bulk-trigger/process-bulk-trigger.usecase.ts
```diff
@@ -10,7 +9,7 @@ import { ParseEventRequest } from '../parse-event-request/parse-event-request.us
 
 @Injectable()
 export class ProcessBulkTrigger {
-  constructor(private parseEventRequest: ParseEventRequest, private mapTriggerRecipients: MapTriggerRecipients) {}
```
djabarovgeorge, line 13: redundant 🧹

apps/api/src/app/events/events.module.ts
```diff
@@ -39,24 +26,11 @@ import { LayoutsModule } from '../layouts/layouts.module';
...
-  DigestFilterStepsTimed,
   GetNovuProviderCredentials,
   StorageHelperService,
   EventsDistributedLockService,
-  ProcessSubscriber,
-  ProcessTenant,
   SendTestEmail,
-  StoreSubscriberJobs,
-  TriggerEvent,
-  MapTriggerRecipients,
```
djabarovgeorge, line 59: 🧹

djabarovgeorge: i suspect we have more redundant providers in the DI.

apps/api/src/app/events/events.controller.ts
```diff
@@ -118,7 +111,7 @@ export class EventsController {
     @Body() body: TriggerEventToAllRequestDto
   ): Promise<TriggerEventResponseDto> {
     const transactionId = body.transactionId || uuidv4();
-    const mappedActor = body.actor ? this.mapActor(body.actor) : null;
+    const mappedActor = typeof body.actor === 'string' ? { subscriberId: body.actor } : null;
```
djabarovgeorge, line 114: make the mapping in order to remove the dep of mapTriggerRecipients from the API.

scopsy:
```suggestion
    const mappedActor = typeof body.actor === 'string' ? { subscriberId: body.actor } : body.actor ;
```

Shouldn't it be like this? in case the actor was passed as an object of `{ subscriberId: body.actor }`

apps/api/src/app/events/usecases/trigger-event-to-all/trigger-event-to-all.usecase.ts
```diff
@@ -42,18 +41,18 @@ export class TriggerEventToAll {
   }
 
   private async trigger(command: TriggerEventToAllCommand, list: SubscriberEntity[]) {
```
scopsy, line 43: @djabarovgeorge shouldn't we also move the logic in the execute method to make sure that finding the subscriber happens in the job and not here?

scopsy: So that the trigger-handler during the fanout will find all relevant subscribers, like it does with the `topic` ?

apps/api/src/app/events/events.controller.ts
djabarovgeorge: Good point thanks for noticing.

apps/api/src/app/events/usecases/trigger-event-to-all/trigger-event-to-all.usecase.ts
djabarovgeorge: I thought about it but then we will need to add additional logic to the trigger-handler, like passing to it a flag or something that will let it know that we need to broadcast, therefore I was not sure if we would want to do it in this task. let me know wdyt
@scopsy

scopsy: I think that to truly decouple this and make sure that the bulk trigger responds quick to the user we will need to do that. 👌

djabarovgeorge: Done

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -265,4 +238,122 @@ export class TriggerEvent {
...
+    template: NotificationTemplateEntity
+  ) {
+    const mappedRecipients = await this.mapTriggerRecipients.execute(
+      MapTriggerRecipientsCommand.create({
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        recipients: command.to,
+        transactionId: command.transactionId,
+        userId: command.userId,
+        actor: actorProcessed,
```
djabarovgeorge, line 293: Noticed that in the previous state is was using `mappedActor` Does anyone knows if there were particular reason? if there is i will refactor in order to pass `mappedActor`

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
```diff
@@ -14,8 +14,8 @@ export class ParseEventRequestCommand extends EnvironmentWithUserCommand {
   @IsDefined()
   overrides: Record<string, Record<string, unknown>>;
 
-  @IsDefined()
-  to: TriggerRecipients;
+  @IsOptional()
+  to?: TriggerRecipients;
```
djabarovgeorge, line 18: Was debating with myself is to update the `to`, to be able to get AddressingTypeEnum as well or not.
in the end decided to add addressingType.
Will love to hear if anyone has any ideas regarding the new Command interface.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
djabarovgeorge, line 1: Was debating with myself is to update the to, to be able to get AddressingTypeEnum as well or not.
in the end decided to add addressingType.
Will love to hear if anyone has any ideas regarding the new Command interface.

apps/api/src/app/inbound-parse/inbound-parse.module.ts
p-fernandez: It probably not anymore. I think it can be safe to be removed.

apps/api/src/app/shared/shared.module.ts
p-fernandez: It was added to provide the QueuesModule as provider in any module that is being dependant of SharedModule so in case of need QueuesModule would be available without having to inject it in said module.

packages/application-generic/src/modules/queues.module.ts
```diff
@@ -67,3 +67,19 @@ const PROVIDERS: Provider[] = [
...
+  WebSocketsQueueServiceHealthIndicator,
+  WorkflowQueueService,
+  WorkflowQueueServiceHealthIndicator,
+];
+
+@Module({
+  providers: [...APP_PROVIDERS],
+  exports: [...APP_PROVIDERS],
+})
+export class BaseApiQueuesModule {}
```
p-fernandez, line 85: Are you trying to gather all the QueuesModule dependencies used only in API app?

packages/application-generic/src/modules/index.ts
```diff
@@ -1 +1,2 @@
 export { QueuesModule } from './queues.module';
+export { BaseApiQueuesModule } from './queues.module';
```
p-fernandez, line 2: You can reuse the same export line.

packages/application-generic/src/usecases/trigger-event/trigger-event.command.ts
```diff
@@ -19,8 +24,8 @@ export class TriggerEventCommand extends EnvironmentWithUserCommand {
   @IsDefined()
   overrides: Record<string, Record<string, unknown>>;
 
-  @IsDefined()
-  to: ISubscribersDefine[];
+  @IsOptional()
+  to?: ISubscribersDefine[];
```
p-fernandez, line 28: What's the reason behind making this optional?

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
p-fernandez: I personally have no idea.
@ainouzgali tagging you because it is my memory about the implementation of mapped actors was made by you, but I might be wrong, so apologies in advance.

packages/application-generic/src/modules/queues.module.ts
djabarovgeorge: Yes exactly, for two reasons 
1. that way we won't initialize redundant clients in API.
2. in the near future (i hope) i want to try creating dynamic health validation, meaning if we inject a new HealthIndicator in the DI it will be validated out of the box, for example without adding it to the healthCheck in the api/src/app/health/health.controller.ts.

packages/application-generic/src/usecases/trigger-event/trigger-event.command.ts
djabarovgeorge: Good question, it is related to this thought i had 
https://github.com/novuhq/novu/pull/4526#discussion_r1368511918
I was not sure how to provide extra data to parse-event and trigger-event usecases that we need to execute broadcast trigger.

p-fernandez: Oh, wasn't sure what that meant.
I think we might need to explore to create specific command to each case sharing a base command.

packages/application-generic/src/modules/queues.module.ts
djabarovgeorge: From the small test I made locally before making the changes I saw that they do initialize even if not injected, i guess because they are singletons maybe?

Regarding the separation i totally agree, make it only for the API in order to keep the scope as small as possible for this PR as it is an extra for it. We definitely can update the other ones as well.

Once again I agree we can merge this one for now, the additional improvements can wait for now. I made this one as a preparation for the future.

p-fernandez (approved):
🌟

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -265,4 +238,122 @@ export class TriggerEvent {
...
+        data: {
+          environmentId: command.environmentId,
+          organizationId: command.organizationId,
+          userId: command.userId,
+          transactionId: command.transactionId,
+          identifier: command.identifier,
+          payload: command.payload,
+          overrides: command.overrides,
+          tenant: command.tenant,
+          ...(actorProcessed && actorProcessed && { actor: actorProcessed }),
```
ainouzgali, line 343:
```suggestion
          ...(actorProcessed && { actor: actorProcessed }),
```

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
ainouzgali: I believe only the subscriberId is needed, so thats why only the mapping was sent. But the command MapTriggerRecipientsCommand is expecting ISubscriberDefine, not SubscriberEntity.

ainouzgali: @djabarovgeorge , a few days ago my PR for adding actor was merged, let me know if you need help resolving conflicts or reintegrating it. As it seems we touched a lot of the same logic

djabarovgeorge: Thanks for bringing it up, i removed all of the actor/tenant i saw before the trigger event use-case (there was one redundant due to duplication in the flow), so now the trigger event use-case is responsible for the mapping.

PR #4534, opened 2023-10-15
Remove submodules from deps

djabarovgeorge:

Please don't review this PR i want to test with the CI.


apps/api/package.json
```diff
@@ -25,7 +25,8 @@
     "test:e2e:ee": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true mocha --timeout 10000 --require ts-node/register --exit --file e2e/setup.ts src/**/*.e2e-ee.ts",
     "migration": "cross-env NODE_ENV=local MIGRATION=true ts-node --transpileOnly ./migrations/expire-at/expire-at.migration.ts",
     "migration:in-app": "cross-env NODE_ENV=local MIGRATION=true ts-node --transpileOnly ./migrations/integration-scheme-update/add-primary-priority-migration.ts",
-    "migration:primary-provider": "cross-env NODE_ENV=local MIGRATION=true ts-node --transpileOnly ./migrations/integration-scheme-update/add-primary-priority-migration.ts"
+    "migration:primary-provider": "cross-env NODE_ENV=local MIGRATION=true ts-node --transpileOnly ./migrations/integration-scheme-update/add-primary-priority-migration.ts",
+    "link:submodules": "cd ../../enterprise/packages/packages/auth && yarn link && cd ../../../../apps/api && yarn link @novu/ee-auth"
```
LetItRock, line 29: similar idea about the separate script that will do the linking

.github/workflows/test.yml
```diff
@@ -7,6 +7,75 @@ on:
...
+      - uses: ./.github/actions/run-worker
+        with:
+          launch_darkly_sdk_key: ${{ secrets.LAUNCH_DARKLY_SDK_KEY }}
+
+        # Runs a single command using the runners shell
+      - name: Build API
+        run: CI='' pnpm build:api
+
+      - name: Build Submodules
+        run: CI='' pnpm build:ee-auth
```
LetItRock, line 65: maybe we should have a command that will run the script where we will list and build all ee deps?

apps/api/package.json
LetItRock:
```suggestion
    "link:submodules": "cd ../../enterprise/packages/packages/auth && yarn link && cd ../../../../apps/api && pnpm link @novu/ee-auth"`
```
please change to `pnpm link`, but check the docs: https://pnpm.io/cli/link

.github/actions/setup-project/action.yml
```diff
@@ -61,6 +65,18 @@ runs:
...
+      shell: bash
+      if: ${{ inputs.submodules == 'true' }}
+      run: |
+        cd enterprise/packages
+        pnpm install --frozen-lockfile
+
+    - name: Link submodules
+      shell: bash
+      if: ${{ inputs.submodules == 'true' }}
+      run: pnpm -C apps/api run link:submodules
```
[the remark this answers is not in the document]
djabarovgeorge: Totally agree :)
Done

.github/workflows/test.yml
djabarovgeorge: For now, we need to build the ee-auth module, am i missing something? or did you have something else in mind?
@LetItRock

LetItRock: I thought about having just one command to build all the ee packages ;) rn we are building only this one, but it will change when we have translation management ready... and then we would need to update this pipeline

djabarovgeorge: I see, we could.
Here i was afraid to build redundant libs so the runtime would increase, but we could build it all for now so it will be more simple and then optimize if needed.

PR #4539, opened 2023-10-16
Nv 2885 extract the design system into a separate package

davidsoderberg:


libs/design-system/package.json
```diff
@@ -0,0 +1,90 @@
+{
+  "name": "@novu/design-system",
```
davidsoderberg, line 2: Is this name good? any suggestions?

davidsoderberg:
How should we handle the storybook? It seems like no one updates it...

apps/web/src/hooks/useInlineComponent.tsx
```diff
@@ -1,7 +1,6 @@
 import { useMemo } from 'react';
 import type { ComponentType, ReactNode } from 'react';
-
-import { useDataRef } from './useDataRef';
+import { useDataRef } from '.';
```
p-fernandez, line 3: This import feels a bit weird.

apps/web/src/pages/templates/workflow/digest/ScheduleMonthlyFields.tsx
```diff
@@ -3,7 +3,7 @@ import { Controller, useFormContext } from 'react-hook-form';
 import { DigestUnitEnum, OrdinalEnum, OrdinalValueEnum, MonthlyTypeEnum, DigestTypeEnum } from '@novu/shared';
 
 import { When } from '../../../../components/utils/When';
-import { colors, Select } from '../../../../design-system';
+import { colors, Select } from '@novu/design-system';
```
p-fernandez, line 6: When is now in the Design System.

libs/design-system/package.json
```diff
@@ -0,0 +1,77 @@
...
+    "tslib": "^2.3.1",
+    "typescript": "4.9.5",
+    "url-loader": "^4.1.1",
+    "vite": "^4.4.5",
+    "vite-plugin-dts": "^3.6.0"
+  },
+  "peerDependencies": {
+    "react": ">=16.8.0",
+    "react-dom": ">=16.8.0",
+    "react-router-dom": "^6.10.0"
```
p-fernandez, line 60: Is this the version we finally need? I saw in an earlier file that 6.2.2 was pinned.

libs/design-system/src/navigation/NavMenu.cy.tsx
```diff
@@ -1,28 +1,27 @@
 import { NavMenu } from './NavMenu';
-import { TestWrapper } from '../../testing';
-import { ROUTES } from '../../constants/routes.enum';
```
p-fernandez, line 3: Should we export the ROUTES to `@novu/shared`?

p-fernandez (approved):
Massive brilliant work! 🌟

apps/web/src/pages/templates/workflow/digest/ScheduleMonthlyFields.tsx
davidsoderberg: Yes it is used by some components, maybe we should not export it?

libs/design-system/src/navigation/NavMenu.cy.tsx
davidsoderberg: Right now I do not see the need it is only used in tests for the design system. But in the future, it might be needed.

apps/web/src/pages/templates/workflow/digest/ScheduleMonthlyFields.tsx
p-fernandez: What I meant is that you can import it from the package so we would not need line 5.

PR #4542, opened 2023-10-16
fix(infra): some pnpm workspace improv for dev flow

p-fernandez:

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,8 +1,8 @@
 packages:
   # all packages in subdirs of packages/ and components/
-  - 'providers/*'
+  - 'providers/**'
   - 'packages/*'
-  - 'enterprise/packages/*'
+  - 'enterprise/packages/**'
   - 'apps/*'
   - 'libs/*'
   # exclude packages that are inside test directories
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,6 +1,6 @@
 packages:
   # all packages in subdirs of packages/ and components/
-  - 'providers/**'
+  - 'providers/*'
   - 'packages/*'
   - 'enterprise/packages/*'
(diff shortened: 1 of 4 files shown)


pnpm-workspace.yaml
```diff
@@ -1,9 +1,9 @@
 packages:
   # all packages in subdirs of packages/ and components/
-  - 'providers/*'
+  - 'providers/**'
   - 'packages/*'
-  - 'enterprise/packages/*'
+  - 'enterprise/packages/**'
```
rifont, line 5: In line with your previous suggestion @p-fernandez , what do you think of this? I think it's better to be fully explicit about the folder depth to avoid any doubt about which folders are targeted. `**` doesn't produce explicitness.

```suggestion
  - 'enterprise/packages/**'
  + 'enterprise/packages/*'
  + 'enterprise/libs/*'
```

p-fernandez: For this change we will need to do two PRs in a row.
I address some of the submodule changes here now.
Then we will go on here: https://github.com/novuhq/packages-enterprise/pull/16
And then finally trim the things back in the monorepo after syncing.
The goal is to move the `libs` package one level down as agreed.

CC: @scopsy

rifont (approved):
LGTM 👍

pnpm-workspace.yaml
```diff
@@ -1,8 +1,10 @@
 packages:
   # all packages in subdirs of packages/ and components/
-  - 'providers/*'
+  - 'providers/**'
```
rifont, line 3: This change needs to be rolled back too.

```suggestion
  - 'providers/*'
```

rifont (changes requested):

rifont (approved):
I've removed the nested providers workspace config, LGTM now 👍

p-fernandez:
Ok, let's try then.

PR #4543, opened 2023-10-16
fix: Remove templateId while creating variant

BiswaViraj:

- remove templateId while creating a new variant


apps/web/src/pages/templates/components/TemplateEditorFormProvider.tsx
```diff
@@ -100,7 +100,7 @@ const makeTemplateCopy = ({
 });
 
 export const makeVariantFromStep = (stepToVariant: IFormStep): IFormStep => {
-  const { id: _, variants, template, ...rest } = stepToVariant;
+  const { id: _, variants, template, _templateId, ...rest } = stepToVariant;
```
BiswaViraj, line 103: previously, it was saving the _templateId of the root step, which lead to a scenario where the content of root step and newly created variant step was using the same message template

LetItRock (approved):

PR #4544, opened 2023-10-16
feat(worker): manage default workers config through env variables

p-fernandez:

Allow control the default values for the workers through environment variables.

We need to study which are the right values for concurrency in our workers. 
Also it is a first step to be able to create a generic worker that could be instantiated with certain configuration by IaC through environment variables to have dedicated fleets.


apps/api/src/config/env-validator.ts
```diff
@@ -73,6 +72,12 @@ const validators: { [K in keyof any]: ValidatorSpec<any[K]> } = {
   LAUNCH_DARKLY_SDK_KEY: str({
     default: '',
   }),
+  WORKER_DEFAULT_CONCURRENCY: num({
+    default: undefined,
```
scopsy, line 76: Should we default to 200 here?

packages/application-generic/src/config/workers.ts
```diff
@@ -0,0 +1,59 @@
...
+  concurrency: number;
+  lockDuration: number;
+}
+
+type WorkersConfig = Record<WorkerEnum, IWorkerConfig>;
+
+const workersConfig: WorkersConfig = {
+  [WorkerEnum.INBOUND_PARSE_MAIL]: {
+    concurrency: defaultConcurrency ?? 200,
+    lockDuration: defaultLockDuration ?? 900,
```
scopsy, line 26:
```suggestion
    lockDuration: defaultLockDuration ?? 90000,
```

I think the old value is `90000` ?

p-fernandez: Good catch. I got confused with Active and Completed metrics workers.

Cliftonz (approved):
lgtm

PR #4555, opened 2023-10-17
style: update log message to better reflect context

scopsy:

NV-3010

Update the log message to better reflect the next job scheduling to not confuse with the start of the processing of a job.

This happens in the end of an existing job, while scheduling the next one. hence starting is a bit confusing


djabarovgeorge (approved):
♥️

PR #4014, opened 2023-08-22
Update Novu to Node 18 LTS

raikasdev:
### What change does this PR introduce?

This PR updates the project to use Node 18 everywhere.

### Why was this change needed?

Closes #1959 

The only currently supported version, Node 16 LTS has it's **end-of-life on September 11th 2023**, about three weeks away now. 

### Other information (Screenshots)

Probably should be discussed if the project should add support up to version 20, which is currently active and will become LTS soon. I will happily make any changes necessary.

All docker container image tag changes have been checked to exist.

Currently this PR sets the node version requirement to `>=18.17.1 <19`, meaning it accepts the current latest LTS patch or a later patch in 18.x, but not versions with a older or a newer major (19, 20).

PR #4556, opened 2023-10-17
v0.20.0 LTS node upgrade

scopsy:

Derived from https://github.com/novuhq/novu/pull/4014 by @raikasdev


rifont:
Node 20 goes LTS next week (24th October), what do we think about fast following with a PR to add Node 20 support?

That way contributor disruption when updating Node versions will be tightly clustered and ideally kept to a minimum, and we won't we'll be on LTS for ~2 years.

dammy001:
> Node 20 enters [Active LTS next week](https://nodejs.dev/en/about/releases/) (24th October), what do we think about fast following with a PR to add Node 20 support?
> 
> 
> 
> That way contributor disruption when updating Node versions will be tightly clustered and ideally kept to a minimum, and we'll be on LTS for ~2.5 years.

@rifont yes, i agree with you. If you check the PR referenced in this, you'll see that I added support for node 20.

rifont:
> @scopsy @rifont how about we drop support for node 10 in public packages? maybe bump to >=14. Currently, it's >=10. I don't think we still have projects on node 10, 12, 14

@dammy001 Great question. We don't yet have a formal stance on support for EOL versions across languages & frameworks for package consumers, so I'd be hesitant to remove support without more data on versions that consumers are using.

What I can say for now is this topic is on the Novu team's radar, and we can seek to drop support after the stance is formalised. We'd be happy to hear your thoughts on the topic to better inform our decision.

djabarovgeorge (approved):
Tested the change in DEV env, and the flow worked as expected.

Issue #4242, opened 2023-09-30
[NV-2846] Improve the ellipsis on the workflow node content


LetItRock:
## Reproduction Steps

The overflow problem arises when the content within the workflow node contains a lengthy, non-breakable word.

## Expected Behaviour

The ellipsis should be shown for the lengthy word and the content should not overflow the workflow node card.

**Current state:**
![image](https://github.com/novuhq/novu/assets/2607232/97844373-2064-4bf2-832b-6faeb55cca55)

**Expected state:**
![image](https://github.com/novuhq/novu/assets/2607232/35324310-d3a3-40ab-acbd-312170b9978b)

Right padding 24px
![image](https://github.com/novuhq/novu/assets/2607232/8ff19f4f-5f0f-47e2-8974-16e05f4641dd)

ShatilKhan:
I would like to take on this issue. Please assign me.
I have done frontend contributions before

LetItRock:
> I would like to take on this issue. Please assign me. I have done frontend contributions before

Sure assigning it to you ;)

capGoblin:
hey hi, i made the necessary changes kindly review.

LetItRock:
hey @ShatilKhan, looking forward for your PR ;)

ShatilKhan:
Thanks for considering,  I was a bit concerned about the rules
I'm working on it & will make my PR soon

Manhatthien-98:
> ## Các bước sinh sản
> Sự cố tràn xảy ra khi nội dung trong nút quy trình công việc chứa một từ dài, không thể ngắt quãng.
> 
> ## Hành vi mong đợi
> Dấu ba chấm phải được hiển thị cho từ dài và nội dung không được tràn ra khỏi thẻ nút quy trình làm việc.
> 
> **Trạng thái hiện tại:** ![image](https://github.com/novuhq/novu/assets/2607232/97844373-2064-4bf2-832b-6faeb55cca55)
> 
> **Trạng thái mong đợi:** ![image](https://github.com/novuhq/novu/assets/2607232/35324310-d3a3-40ab-acbd-312170b9978b)
> 
> Đệm phải 24px ![image](https://github.com/novuhq/novu/assets/2607232/8ff19f4f-5f0f-47e2-8974-16e05f4641dd)

https://github.com/novuhq/novu/issues/4242#issue-1920388571
https://github.com/novuhq/novu/issues/4242

PR #4557, opened 2023-10-17
Improve ellipsis on workflow node content

ShatilKhan:
Issue #4242 
Improves the ellipsis on the workflow node content

The long texts kept getting out of scope on the workflow node content

### DIFF
--- a/apps/web/src/pages/templates/workflow/workflow/node-types/WorkflowNode.tsx
+++ b/apps/web/src/pages/templates/workflow/workflow/node-types/WorkflowNode.tsx
@@ -336,6 +336,12 @@ const StyledContentWrapper = styled.div`
   justify-content: flex-start;
   width: 100%;
   flex: 1;
+    .workflow-node-subtitle {
+    overflow: 'hidden',
+    whiteSpace: 'nowrap',
+    textOverflow: 'ellipsis',
+    maxWidth: '60%',
+  }
 `;
 
 const UnstyledButtonStyled = styled.div`


apps/web/src/pages/templates/workflow/workflow/node-types/WorkflowNode.tsx
```diff
@@ -336,6 +336,12 @@ const StyledContentWrapper = styled.div`
   justify-content: flex-start;
   width: 100%;
   flex: 1;
+    .workflow-node-subtitle {
+    overflow: 'hidden',
+    whiteSpace: 'nowrap',
+    textOverflow: 'ellipsis',
+    maxWidth: '60%',
+  }
```
ainouzgali, line 344: workflow-node-subtitle is a data-test-id. Styles could not be applied on that id, these styles will have no affect on the node content.

ainouzgali (commented):
Hi @ShatilKhan , please look at my comment. 

The screenshot you added shows the situation where there is a space in the content and it looks right because of the rows={1} on the Text component (applies lineClamp). That already exists. 

The problem in the issue is when there are no spaces in the content or when its a block of html -> then the lineClamp does not work.
To reproduce and see what I mean try with content '123456789123456789123456789` for example 

Thank you!!

scopsy:
Hi @ShatilKhan did you had a chance to look at Gali's comment?

scopsy:
Closing due to inactivity

PR #4559, opened 2023-10-17
Add Trigger Identifier to the Subscriber Preferences response

djabarovgeorge:

Add the triggers field to be part of the response to the query. The mapping functionality is included in this file: https://github.com/novuhq/novu/blob/58c46d72367f0f71bb0dd6b2014c2d5c6a8adcc3/packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts#L252

Some users would love to filter the preferences based on the identifier of the trigger.

Created this initial solution and still have a couple of questions left hanging, will change this PR to ready from draft state after i figure it out.

### DIFF
--- a/packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
+++ b/packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
@@ -47,6 +47,7 @@ export class GetSubscriberTemplatePreference {
     if (!subscriber) {
       throw new ApiException(`Subscriber ${command.subscriberId} not found`);
     }
+
     const initialActiveChannels = await this.getActiveChannels(command);
     const subscriberPreference =
       await this.subscriberPreferenceRepository.findOne({
@@ -253,6 +254,7 @@ function mapTemplateConfiguration(
     name: template.name,
     tags: template?.tags || [],
     critical: template.critical != null ? template.critical : true,
+    identifier: template.triggers[0].identifier,
     ...(template.data ? { data: template.data } : {}),
   };
 }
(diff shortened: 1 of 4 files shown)


libs/dal/src/repositories/notification-template/notification-template.entity.ts
```diff
@@ -121,13 +123,3 @@ export class StepFilter {
...
-
-export interface ITriggerVariable {
-  name: string;
-  type: TemplateVariableTypeEnum;
-}
-
-export interface ITriggerReservedVariable {
-  type: TriggerContextTypeEnum;
-  variables: ITriggerVariable[];
-}
```
djabarovgeorge, line 133: Do we have a reason for having those in the DAL?
i noticed that they are duplicated and not aligned with the interface in the Shared lib.

packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -253,6 +254,7 @@ function mapTemplateConfiguration(
     name: template.name,
     tags: template?.tags || [],
     critical: template.critical != null ? template.critical : true,
+    triggers: template.triggers,
```
LetItRock, line 257: can you please update the Swagger API response in the subscriber's controller?

LetItRock (approved):

packages/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
djabarovgeorge: Definitely!  i knew I forgot something here 🤦‍♂️

scopsy (approved):

PR #4573, opened 2023-10-18
fix: promotion of step not having the prod variant template id

ainouzgali:

In promotion of a notification template that has steps with variants, the production step that is created has the template ids of the of the development variant message template. 
Resulted in changes made(update) in development shown immediately in production, without having to promote the change.


apps/api/src/app/change/usecases/promote-notification-template-change/promote-notification-template-change.usecase.ts
```diff
@@ -53,6 +57,30 @@ export class PromoteNotificationTemplateChange {
         return message._parentId === step._templateId;
       });
 
+      if (step.variants && step.variants.length > 0) {
+        step.variants = step.variants
+          ?.map(mapNewVariantItem)
+          .filter((variant): variant is StepVariantEntity => variant !== undefined);
+      }
```
LetItRock, line 64: question, can you please tell me in which case the variant could be `undefined`? I might be missing something

apps/api/src/app/change/usecases/promote-notification-template-change/promote-notification-template-change.usecase.ts
```diff
@@ -53,6 +57,30 @@ export class PromoteNotificationTemplateChange {
...
+
+        return undefined;
+      }
+
+      if (step?._templateId && oldMessage._id) {
+        step._templateId = oldMessage._id;
+      }
+
+      return step;
+    };
```
LetItRock, line 77: can we please cover all of these with tests?

LetItRock (approved):

PR #4574, opened 2023-10-18
Fix icon overlapping

### LINKED ISSUE #4533: ✨ Polishing: Integration Store Test Provider Sidebar - Overlaping Icons
### 📜 Description

In integration store when we click on any test provider a sidebar appears, On right top of sidebar icons are overlapping,

image

FULL SCREENSHOT
image

### 👟 Reproduction steps

1. Integrations store , 2. Add a provider if not added, 3. Click on any Provider, 4. Look at right top ( close icon )

### 👍 Expected behavior

Icons should be separated by some Margin

It is just css issue and this issue should be considered as Good first issue

sudhanshugautam2911:
This PR fixes the icons overlapping in Integration Store Test Providers Sidebar .

The change was in the CSS part, Initially margin property was applied on button but it was not working properly so my change fixed that problem.

This change was needed because Icons overlapping do not look good and impact user experience.

Closes #4533 

Sharing Screenshots after changes
image

Before :
image


Vishal-K-988 (commented):
great !

Cliftonz:
Thank you for doing this, however the spacing between each icon should be equal. Currently the horizontal ellipse and the X icon does not have the same spacing as the rest. 

I would recommend using a flex box instead of just setting a margin.

Cliftonz (changes requested):
We should be using something different then marginLeft to ensure that the spacing is equal between all icons.

sudhanshugautam2911:
@Cliftonz  Test Provider and its right icons except 'x' are grouped together and x is in another div so we can only apply flex to first 4 buttons but when it comes to 'x' which is separate from these 4 we can only apply margin left.

sudhanshugautam2911:
Screenshot after these changes,

image

Now it look way better please check changes @Cliftonz

Cliftonz:
That look great, thanks for the contribution!

Cliftonz (approved):

PR #4575, opened 2023-10-18
Revert "fix(infra): some pnpm workspace improv for dev flow"

p-fernandez:
Reverts novuhq/novu#4542


p-fernandez:
Reverting as it didn't work properly.

PR #4576, opened 2023-10-18
fix: updated the print-affected-array script after updating the nx

LetItRock:

Updated the `print-affected-array` script after we updated the Nx version to 16.10.

Without this fix the pipelines using that script are failing.

The results running script for this branch:
Screenshot 2023-10-18 at 17 16 23

### DIFF
--- a/apps/api/src/main.ts
+++ b/apps/api/src/main.ts
@@ -1,6 +1,3 @@
 import { bootstrap } from './bootstrap';
 
 bootstrap();
-
-// eslint-disable-next-line no-console
-console.log('Test affected apps');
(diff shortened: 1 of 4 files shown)


apps/api/src/main.ts
```diff
@@ -1,3 +1,6 @@
 import { bootstrap } from './bootstrap';
 
 bootstrap();
+
+// eslint-disable-next-line no-console
+console.log('Test affected apps');
```
scopsy, line 6: Can remove those now :)

scopsy (approved):

PR #4577, opened 2023-10-18
feat: create new dir link

djabarovgeorge:


djabarovgeorge:
@p-fernandez Do we still need this one or can we close it?

p-fernandez:
> @p-fernandez Do we still need this one or can we close it?

Wait just a bit in case we need something else but probably we will be able to use this to merge.

djabarovgeorge:
Was implemented in https://github.com/novuhq/novu/pull/4677

PR #4580, opened 2023-10-19
Update README.md

lokeshnahata:


p-fernandez:
Thank you for trying to contribute. Unfortunately this is not a sensible change to be merged.
Please look for any other issues that are open and unassigned from the issues list that can be more interesting to help us. 🙏🏻

Issue #4582, opened 2023-10-19
[NV-3037] 🐛 Bug Report: Wrong Socket Close in useUnreadCount.ts


Khongchai:
### 📜 Description

In[ this hook](https://github.com/novuhq/novu/blob/e332a6c8b46aef1e886330d5560c4ec705afdb78/packages/notification-center/src/hooks/useUnreadCount.ts#L43), you're doing `socket.on(WebSocketEventEnum.UNREAD)` but you're doing `socket.off(WebSocketEventEnum.UNSEEN)`.

### 👟 Reproduction steps

None

### 👍 Expected behavior

It should closes the correct socket channel.

### 👎 Actual Behavior with Screenshots

```ts
    socket.on(
      WebSocketEventEnum.UNREAD, // <<<<<<<<<<<<< here
      debounce((data?: { unreadCount: number }) => {
        if (Number.isInteger(data?.unreadCount)) {
          queryClient.setQueryData<{ count: number }>(queryKeysRef.current.unreadCountQueryKey, (oldData) => ({
            count: data?.unreadCount ?? oldData.count,
          }));

          // when unread count changes, we need to refetch unseen count
          queryClient.refetchQueries(queryKeysRef.current.unseenCountQueryKey, {
            exact: false,
          });
          queryClient.refetchQueries(queryKeysRef.current.fetchNotificationsQueryKey, {
            exact: false,
          });
          // refetch all feeds unseen count that is shown on the tabs
          queryClient.refetchQueries([...FEED_UNSEEN_COUNT_QUERY_KEY], {
            exact: false,
          });

          dispatchUnreadCountEvent(data.unreadCount);
        }
      }, DEBOUNCE_TIME)
    );

    return () => {
      socket.off(WebSocketEventEnum.UNSEEN); // <<<<<<<<<<<<< here
    };
  }, [socket, queryClient, setQueryKey]);
```

### Novu version

0.17.0

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes

<sub>[NV-3037](https://linear.app/novu/issue/NV-3037/🐛-bug-report-wrong-socket-close-in-useunreadcountts)</sub>

Khongchai:
[...]

PR #4590, opened 2023-10-19
[NV-3037] Fixed closing off the wrong socket in useUnreadCount.ts

Khongchai:

This fixes an issue where the useUnreadCount hook closes the wrong websocket channel.

https://github.com/novuhq/novu/issues/4582

### DIFF
--- a/packages/notification-center/src/hooks/useUnreadCount.ts
+++ b/packages/notification-center/src/hooks/useUnreadCount.ts
@@ -75,7 +75,7 @@ export const useUnreadCount = ({ onSuccess, ...restOptions }: UseQueryOptions<IC
     () => apiService.getUnreadCount({ limit: 100 }),
     {
       ...restOptions,
-      enabled: isSessionInitialized && fetchingStrategy.fetchUnseenCount,
+      enabled: isSessionInitialized && fetchingStrategy.fetchUnreadCount,
       onSuccess: (data) => {
         dispatchUnreadCountEvent(data.count);
         onSuccess?.(data);
--- a/packages/notification-center/src/hooks/useUnreadCount.ts
+++ b/packages/notification-center/src/hooks/useUnreadCount.ts
@@ -66,7 +66,7 @@ export const useUnreadCount = ({ onSuccess, ...restOptions }: UseQueryOptions<IC
     );
 
     return () => {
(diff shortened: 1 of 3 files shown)


djabarovgeorge:
Hey @Khongchai,
Awesome work discovering the issue, could you update the refetch key issue that you located as well here?

packages/notification-center/src/hooks/useUnreadCount.ts
```diff
@@ -66,7 +66,7 @@ export const useUnreadCount = ({ onSuccess, ...restOptions }: UseQueryOptions<IC
     );
 
     return () => {
-      socket.off(WebSocketEventEnum.UNSEEN);
+      socket.off(WebSocketEventEnum.UNREAD);
```
p-fernandez, line 69: I think in line 78 we would need to update to: 
```
enabled: isSessionInitialized && fetchingStrategy.fetchUnreadCount
```

Khongchai:
> Hey @Khongchai, Awesome work discovering the issue, could you update the refetch key issue that you located as well here?

Hi, sorry for my late reply. 

The refetch key issue, as I have mentioned in [this issue](https://github.com/novuhq/novu/issues/4582#issuecomment-1773707401). I'm not quite sure if this really is a problem. I thought it was at first, but after having looked into it further, it was just making sure the unseen value is updated. 

I haven't checked Novu's code on this part, but should unseen be updated if unread is updated? I thought there was a bug, but after reading this comment, I'm not so sure

image


For example, if something is marked as read should they also be marked automatically as seen? If so, the code is already working as intended (maybe some Novu backend logic to update all read fields as seen).

If my understanding is incorrect and the refetch function was really not working as intended in some way, please let me know.

packages/notification-center/src/hooks/useUnreadCount.ts
Khongchai: Ok I'll add this

Khongchai: @p-fernandez  I've updated the condition. But I needed to add a couple of things more because the fetchingStrategy interface was missing the fetchUnreadCount key.

For the default value, I set it to `true` so that existing code won't break (it was using unseen's default and it was true)

p-fernandez (approved):
🌟

PR #4596, opened 2023-10-20
refactor(providers): migrate from @sendinblue/client SDK to direct API calls with Axios 

### LINKED ISSUE #4589: [NV-3038] Migrate from @sendinblue/client SDK to direct API calls with Axios
## What?
In the `@novu/sendinblue` package we would like to migrate from using the `@sendinblue/client` SDK to having a code that directly calls to the provider APIs. Check the documentation about the API here: https://developers.brevo.com/

## Why? (Context)
Due to an un-maintained SDK, we should move from the official SDK' to working directly with the provider's API spec.

## Definition of Done
- the SDK package has been removed from the package.json file
- the current provider functionality works with the calls to their APIs using Axios
  - attachments

Dhoni77:

- Remove sendinblue sdk
- Use @getbrevo/brevo node library
- Updated unit tests

Closes #4589

Due to an un-maintained SDK, we should move from the official SDK to working directly with the provider's API spec.


p-fernandez (changes requested):

Sendinblue has been rebranded as Brevo. They also have Node SDK (https://github.com/getbrevo/brevo-node) to connect to their API, so the change I think should be only the library.
We are currently having a contributor (https://github.com/novuhq/novu/pull/3667) to help to the changes in the rebranding but I think you can tackle the replacement of the SDK in this PR.

Dhoni77:
> Thank you for your contribution.
> 
> Sendinblue has been rebranded as Brevo. They also have Node SDK (https://github.com/getbrevo/brevo-node) to connect to their API, so the change I think should be only the library. We are currently having a contributor (#3667) to help to the changes in the rebranding but I think you can tackle the replacement of the SDK in this PR.

sure will do @p-fernandez 👍

Dhoni77:
cc @p-fernandez

providers/sendinblue/src/lib/brevo.provider.ts
```diff
@@ -27,17 +29,16 @@ export class SendinblueEmailProvider implements IEmailProvider {
       senderName: string;
     }
   ) {
-    this.transactionalEmailsApi = new TransactionalEmailsApi();
-    this.transactionalEmailsApi.setApiKey(
-      TransactionalEmailsApiApiKeys.apiKey,
-      this.config.apiKey
-    );
+    this.defaultClient = brevo.ApiClient.instance;
```
scopsy, line 32: I'm a bit concerned that this will be reused across class institutions. Can we ensure that we have a brand new object here each time instead of referencing this like it? @djabarovgeorge wdyt?

djabarovgeorge: Agree, we should have same behavior between the providers, i am bit confused by the lib https://github.com/getbrevo/brevo-node/blob/584a95946ffd88a4874e2db85d35bff98bc54715/src/ApiClient.js#L598 does it mean that on every assign a new ApiClient will be initialized or it will be reused?

scopsy (approved):

PR #4599, opened 2023-10-20
docs: enhanced punctuation and omitting abstraction.

anujshandillya:

This PR enhances punctuation for the betterment of the project.

It was needed because some points were abstract and were not professional.


p-fernandez:
Unfortunately this is an opinionated change that we find is not needed to apply and also it changes some of meaning and nuances of what we want to transmit.
Feel free to go through the list of issues opened, some of them are labeled as good first issue, and tackle any of them to help us to grow. 🙏🏻

PR #4602, opened 2023-10-20
Update README.md for  written beautification

FarukhS52:

a small change in readme .md to make it more good-looking

it doesn't fix a issue but helps in beautifying README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Your contribution, no matter its size, holds immense value. We eagerly await to 
 
   <p align="center">
     <br />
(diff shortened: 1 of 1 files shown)


p-fernandez:
Unfortunately this is an opinionated change that we find is not needed to apply.
Feel free to go through the list of issues opened, some of them are labeled as good first issue, and tackle any of them to help us to grow. 🙏🏻

FarukhS52:
Thanks Novuhq/Novu for the feedback ,will definitely go through it.

PR #4616, opened 2023-10-22
fix(api): Cancel all jobs related to a transactionId from the 'cancel' method

scopsy:

Closes NV-3019

### DIFF
--- a/apps/api/src/app/events/e2e/delay-events.e2e.ts
+++ b/apps/api/src/app/events/e2e/delay-events.e2e.ts
@@ -24,13 +24,13 @@ describe('Trigger event - Delay triggered events - /v1/events/trigger (POST)', f
   let standardQueueService: StandardQueueService;
   const messageRepository = new MessageRepository();
 
-  const triggerEvent = async (payload, transactionId?: string, overrides = {}) => {
+  const triggerEvent = async (payload, transactionId?: string, overrides = {}, to = [subscriber.subscriberId]) => {
     await axiosInstance.post(
       `${session.serverUrl}/v1/events/trigger`,
       {
         transactionId,
         name: template.triggers[0].identifier,
-        to: [subscriber.subscriberId],
+        to,
         payload,
         overrides,
       },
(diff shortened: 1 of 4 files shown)


apps/api/src/app/events/e2e/delay-events.e2e.ts
```diff
@@ -24,13 +24,13 @@ describe('Trigger event - Delay triggered events - /v1/events/trigger (POST)', f
   let standardQueueService: StandardQueueService;
   const messageRepository = new MessageRepository();
 
-  const triggerEvent = async (payload, transactionId?: string, overrides = {}) => {
+  const triggerEvent = async (payload, transactionId?: string, overrides = {}, to = [subscriber.subscriberId]) => {
```
scopsy, line 27: Added the ability to test more than 1 subscriber

apps/api/src/app/events/e2e/delay-events.e2e.ts
```diff
@@ -342,7 +344,7 @@ describe('Trigger event - Delay triggered events - /v1/events/trigger (POST)', f
           content: '',
           metadata: {
             unit: DigestUnitEnum.SECONDS,
-            amount: 0.1,
+            amount: 5,
```
scopsy, line 347: To avoid flakiness in the test

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
```diff
@@ -7,20 +7,25 @@ export class CancelDelayed {
...
-      status: JobStatusEnum.DELAYED,
-    });
+    const jobs = await this.jobRepository.find(
+      {
+        _environmentId: command.environmentId,
+        transactionId: command.transactionId,
+        status: JobStatusEnum.DELAYED,
+      },
+      '_id'
+    );
```
ainouzgali, line 17: it will still find only the delay job - as its the only one that has DELAYED status. The others are pending. Or am I missing something?

ainouzgali: Never mind. Got you. its multiple subscribers

ainouzgali (approved):

PR #4625, opened 2023-10-23
Fixes #4261: Modified Scroll Bar as required

### LINKED ISSUE #4261: [NV-2837] 💔 👀 Update UX/UI of the scroll
## What?
Scroll component look and feel

## Why? (Context)
It does not look like a part of the application with borders and creates visual clutter.
It also behaves strangely for some pages. Like in the WF editor, it shouldn't appear at all, because there is not so much content. It also does not allow you to scroll down, feels like Novu is frozen :-0

## Definition of Done
- The scroll displays only when the content of a page(section) is bigger than a container of the content (so in the WE it shouldn't appear when the content is fit into the container
- The scroll uses the same scroll component everywhere
- The component with 8px height max (see the component in Figma)
- The visible scroll body is 4px and it's 40% transparent (see the component in Figma) 
- The component background is 100% transparent
- The scroll should perfectly fit into the container with the scrolling content
- The scroll sick to the right side of the container

**Design:** 
[Workflows / New workflow](https://www.figma.com/file/mxViLXxtC0PnB0ehXDcafw/Crafting-updates?type=design&node-id=730%3A5240&mode=design&t=Xs7u6G979yV7n8RA-1)

anmol-matharu-100:

My PR modifies the scroll bar as was required to resolve the issue #4261 . ScrollArea from Mantine has been used to resolve the issue.

ScrollArea is an pre-built component in Mantine library that has been used to resolve this issue.

### DIFF
--- a/apps/web/src/components/layout/components/SideNav.tsx
+++ b/apps/web/src/components/layout/components/SideNav.tsx
@@ -7,6 +7,7 @@ import {
   Navbar,
   Popover,
   useMantineColorScheme,
+  ScrollArea
 } from '@mantine/core';
 import { useEffect, useState } from 'react';
 import { Link, useNavigate } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


scopsy:
Hi @anmol-matharu-100 did you had a chance to run those changes locally? From looking at the diff it will fail to compile.

scopsy:
Closing due to inactivity, feel free to reopen 🙏

PR #4631, opened 2023-10-23
feat: add template id support in mailersend

jainpawan21:

1. Added `templateId` and `personalization` in `mailersend` provider
2. Added tests
3. Fixed _subscriberId bug in message matcher usecase


apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -321,7 +321,7 @@ export class MessageMatcher extends Filter {
     filterProcessingDetails: FilterProcessingDetails
   ): Promise<boolean> {
     const subscriber = await this.subscriberRepository.findOne({
-      _id: command.subscriberId,
+      _id: command._subscriberId,
```
jainpawan21, line 324: https://discord.com/channels/895029566685462578/1146039495398010982/1166071042213679144

djabarovgeorge: Totally right i just now noticed that we still have not merged the deprecation of this use case on behalf of using ConfitionFilter. 🤦‍♂️

The refactor is located here
https://github.com/novuhq/novu/pull/4187

libs/shared/src/types/events/index.ts
```diff
@@ -33,7 +33,7 @@ export interface IEmailOptions {
   payloadDetails?: any;
   notificationDetails?: any;
   ipPoolName?: string;
-  customData?: Record<string, Record<string, unknown>>;
+  customData?: Record<string, any>;
```
djabarovgeorge, line 36: why is this change needed?

jainpawan21: This change is from old PR. https://github.com/novuhq/novu/pull/4504
But this PR was towards main so I included that commit in this PR to make it work

djabarovgeorge: I see, let's merge it for now but we could find another way to do so without `any` there.

djabarovgeorge (approved):

PR #4632, opened 2023-10-23
refactor: replace findby id with findone

scopsy:


apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.usecase.ts
```diff
@@ -55,7 +55,10 @@ export class SetIntegrationAsPrimary {
   async execute(command: SetIntegrationAsPrimaryCommand): Promise<IntegrationEntity> {
     Logger.verbose('Executing Set Integration As Primary Usecase');
 
-    const existingIntegration = await this.integrationRepository.findById(command.integrationId);
+    const existingIntegration = await this.integrationRepository.findOne({
+      _id: command.integrationId,
+      _organizationId: command.organizationId,
```
djabarovgeorge, line 60: I think we should fetch here by environment id, and not the one we have in the command/user: IJwtPayload, shouldn't we pass the env id related to the integration in the client integration store?

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -101,7 +101,10 @@ export class UpdateIntegration {
   async execute(command: UpdateIntegrationCommand): Promise<IntegrationEntity> {
     Logger.verbose('Executing Update Integration Command');
 
-    const existingIntegration = await this.integrationRepository.findById(command.integrationId);
+    const existingIntegration = await this.integrationRepository.findOne({
+      _id: command.integrationId,
+      _organizationId: command.organizationId,
```
djabarovgeorge, line 106: same here.

djabarovgeorge (approved):
Looks good to me, left one comment regarding the topic I was not sure about, if you think there is an issue we can open a separate ticket for it.

apps/api/src/app/integrations/usecases/set-integration-as-primary/set-integration-as-primary.usecase.ts
scopsy: The id is unique enough here, so I don't really see a need in that. I believe that the current setup should be good enough, and provide the adequate specificity for that manner

djabarovgeorge: Agree that the case if unique but human error still could persist, where the user of the API will enter production id instead of the development id. 
Specifying the exact environment with additional parameter could help reduce the odds.

PR #4643, opened 2023-10-24
feat: add translation web package

davidsoderberg:

Add translations routes

Screenshot 2023-10-24 at 13 26 59


apps/web/src/pages/TranslationPages.tsx
djabarovgeorge, line 1: what do you think about importing only if NOVU_MANAGE is true, and if the import fails then throw an exception?
At the moment we could have the cloud without the translation page and we won't know about it.

libs/design-system/src/layout/PageContainer.tsx
djabarovgeorge, line 1: 🚀

apps/web/src/components/layout/components/PageContainer.tsx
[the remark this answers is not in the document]
davidsoderberg: We could I just did not like to make to huge changes in this PR keep it easy to review.

djabarovgeorge:
Looks great but won't the installation fail because of the existence of non-existent on @novu/ee-translation-web in this state?

LetItRock (approved):
🚀

PR #4648, opened 2023-10-24
feat(shared): tidy up filters operators to reuse in events broadcast

p-fernandez:

Creates an enum with existing filter operators that were hardcoded in the codebase.

In order to implement events broadcast filtering in the payload I found useful to reuse a similar implementation for filtering in a similar way we did for the Workflow steps filtering.

### DIFF
--- a/libs/shared/src/types/builder/builder.types.ts
+++ b/libs/shared/src/types/builder/builder.types.ts
@@ -1,20 +1,43 @@
-export type BuilderGroupValues = 'AND' | 'OR';
+export enum FieldOperatorEnum {
+  ALL_IN = 'ALL_IN',
+  ANY_IN = 'ANY_IN',
+  BETWEEN = 'BETWEEN',
+  EQUAL = 'EQUAL',
+  IN = 'IN',
+  IS_DEFINED = 'IS_DEFINED',
+  LARGER = 'LARGER',
+  LARGER_EQUAL = 'LARGER_EQUAL',
+  LIKE = 'LIKE',
+  NOT_BETWEEN = 'NOT_BETWEEN',
+  NOT_EQUAL = 'NOT_EQUAL',
+  NOT_IN = 'NOT_IN',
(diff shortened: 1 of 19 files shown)


libs/shared/src/types/builder/builder.types.ts
```diff
@@ -1,20 +1,43 @@
...
+  LARGER = 'LARGER',
+  LARGER_EQUAL = 'LARGER_EQUAL',
+  LIKE = 'LIKE',
+  NOT_BETWEEN = 'NOT_BETWEEN',
+  NOT_EQUAL = 'NOT_EQUAL',
+  NOT_IN = 'NOT_IN',
+  NOT_LIKE = 'NOT_LIKE',
+  SMALLER = 'SMALLER',
+  SMALLER_EQUAL = 'SMALLER_EQUAL',
+}
```
p-fernandez, line 17: We might not be able to implement them all or just a subset because the nature of the payload for the events broadcast, but big part of the subset can be reused for filtering therefore the creation of this enum (besides cleaning the code).

packages/application-generic/src/utils/filter.ts
```diff
@@ -23,31 +24,31 @@ export abstract class Filter {
...
-    if (fieldFilter.operator === 'NOT_IN') {
+    if (fieldFilter.operator === FieldOperatorEnum.NOT_IN) {
       result = !actualValue.includes(filterValue);
     }
-    if (fieldFilter.operator === 'IN') {
+    if (fieldFilter.operator === FieldOperatorEnum.IN) {
       result = actualValue.includes(filterValue);
     }
-    if (fieldFilter.operator === 'IS_DEFINED') {
+    if (fieldFilter.operator === FieldOperatorEnum.IS_DEFINED) {
```
p-fernandez, line 51: I have noticed in this method `ALL_IN` functionality is not implemented. I ignore which was the reason for that.

LetItRock (approved):
thanks! ⭐

Issue #4639, opened 2023-10-24
🐛 Bug Report: FCM data-type notification expect a message property in the payload


cherfia:
### 📜 Description

Currently, when sending a data-type notification, it lacks the push message content that was initially set during the push provider configuration step in the workflow. After inspecting the code below https://github.com/novuhq/novu/blob/bfc3862868ed5b023e18881e3ecbb16b2a638f95/providers/fcm/src/lib/fcm.provider.ts#L69

It turned out that it uses the `body` property instead of `message`.

According to Expo documentation, as stated in [Send notifications with FCM & APNs](https://docs.expo.dev/push-notifications/sending-notifications-custom/), the correct property to use should be `message` rather than `body`. I further confirmed this by introducing a custom field labeled `message` to the payload given that the `data` property essentially combines the different payload and overrides properties. It worked, however, the drawback is that the `message` property originated from the backend, which implies that we would be unable to configure it within the workflow, templates, and internationalization (i18n).

### 👟 Reproduction steps
```
curl --location --request POST 'https://eu.api.novu.co/v1/events/trigger' \
--header 'Authorization: ApiKey <REPLACE_WITH_API_KEY>' \
--header 'Content-Type: application/json' \
--data-raw '{
         "name": "space-membership-request",
         "to": {
           "subscriberId": "4e4bcfe9-1b9e-4383-9784-7d36572adfd3"
         },
         "payload": {
           "data": {
             "user": {
               "id": "345a91b3-941b-4274-8e33-49485c4d7cda",
               "full_name": "Luigi",
             },
             "space": {
               "id": "c1c1989b-1b51-489b-9219-57274a8d65cb",
               "name": "super-mario-bros-wonder",
               "title": "Super Mario Bros. Wonder",
             },
             "request_message": "Mama mia!",
             "request_datetime": "2023-10-23T14:55:58.976406"
           },
           "notification_type": "SPACE_MEMBERSHIP_REQUEST",
           "notification_version": "1.0.0"
         },
         "overrides": {
          "type": "data",
           "fcm": {
            "type": "data"
           }
         },
         "tenant": null
}'
```
### 👍 Expected behavior

![expected](https://github.com/novuhq/novu/assets/10255018/25b184e8-27f3-4a8d-b790-7a7b9688df75)

### 👎 Actual Behavior with Screenshots

![actual](https://github.com/novuhq/novu/assets/10255018/ce8ddcc8-9faa-465e-96a3-e3cca3ea4447)

### Novu version

[...]

PR #4652, opened 2023-10-25
fix(@novu/fcm): use message instead of body for data-type notifications

cherfia:

Fixed the bug reported https://github.com/novuhq/novu/issues/4639 by including the `message` property in the `data` sent by the FCM provider, aligning with [Expo documentation](https://docs.expo.dev/push-notifications/sending-notifications-custom/).

The notifications were displaying only the titles and not rendering the notification message body.

### Other information
While the `body` property is not required for data-type notifications, it has been retained for backward compatibility to support users who already rely on it.


p-fernandez:
Thanks for your contribution. 
You will need to update the tests based on the change you have done.

cherfia:
@p-fernandez I have updated [fcm.provider.spec.ts](https://github.com/novuhq/novu/pull/4652/files#diff-b3fcdc39b9827c10db8e1dbc330b523a52fc0fc7eba11bef3c9a3da29f7fd6fb) to include `message` property and `NX Successfully ran target test for project @novu/fcm`

p-fernandez (approved):
🌟

cherfia:
Thanks! I'm committed to ongoing contributions to Novu as it's integral part of our open-source solution ⭐

PR #4656, opened 2023-10-25
v0.20.1

scopsy:


libs/design-system/package.json
```diff
@@ -0,0 +1,77 @@
...
+  "name": "@novu/design-system",
+  "version": "0.20.0",
+  "repository": "https://github.com/novuhq/novu",
+  "description": "",
+  "scripts": {
+    "start": "npm run build:watch",
+    "prebuild": "rimraf dist",
+    "lint": "eslint --ext .ts,.tsx src",
+    "build": "npm run build:cjs && npm run build:esm && npm run build:types",
+    "build:cjs": "cross-env node_modules/.bin/tsc -p tsconfig.json",
```
dammy001, line 11: @scopsy why have to build ESM and CJS separately? why don't we use tsup package?

PR #4664, opened 2023-10-26
feat: updated the schema and broke down into re-usable schema

BiswaViraj:

- updated the schema to include variants validation


apps/web/src/pages/templates/components/notificationTemplateSchema.ts
```diff
@@ -79,6 +79,184 @@ const validateUnit = ({
   }
 };
 
+const digestMetadataSchema = z.object({
```
BiswaViraj, line 82: extracted these schemas to re-usable variables

ainouzgali:
@BiswaViraj when clicking update, it passes only the template object for the variants. I mean instead of having 
`variants: [ {name: .. , filters: [..], template: {...}}]`
it now sends to BE `variants: [ {template: {..} } ]` 

Can you please take a look? Tru saving a variant and you'll see the error I mean

ainouzgali (approved):
🚀

PR #4667, opened 2023-10-26
feat(application-generic, shared): Add Rate Limiting feature flag

rifont:
* Adds the `IS_REQUEST_RATE_LIMITING_ENABLED` shared type enum
* Refactors the `feature-flags` shared type to strictly type the flag naming conventions, and adds relevant type checking tests
* Adds the `GetIsRequestRateLimitingEnabled` use case

A feature flag is required for the new API Request Rate Limiting feature to safely introduce and release changes.

n/a


libs/shared/src/types/feature-flags/flags.types.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  IS_INVALID_ENABLED = 'INVALID_ENABLED',
+}
+// @ts-expect-error - Invalid value on IS_INVALID_ENABLED: 'INVALID_ENABLED'
+testFlagEnumValidity(InvalidValueFlagsEnum);
+
+/**
+ * Verifying declared FlagEnums
+ */
+testFlagEnumValidity(FeatureFlagsKeysEnum);
+testFlagEnumValidity(SystemCriticalFlagsEnum);
```
rifont, line 54: These tests will break if the in-use `enum`s do not match formatting.

packages/application-generic/src/usecases/get-feature-flag/get-use-merged-digest-id.use-case.ts
```diff
@@ -12,7 +12,7 @@ export class GetUseMergedDigestId extends GetFeatureFlag {
       value,
       fallbackValue
     );
-    const key = FeatureFlagsKeysEnum.IS_USE_MERGED_DIGEST_ID;
+    const key = FeatureFlagsKeysEnum.IS_USE_MERGED_DIGEST_ID_ENABLED;
```
rifont, line 15: This FF key on the enum was missing the `_ENABLED` suffix and broke the tests. Only the `enum` key has been updated, not the environment variable - so this will not affect customers.

libs/shared/src/types/feature-flags/flags.types.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * If the test fails, you should review your `enum` to verify that both the
+ * keys and values match the format specified by the `IFlagKey` template literal type.
+ * ref: https://stackoverflow.com/a/58181315
+ *
+ * @param testEnum - the Enum to type check
+ */
+export declare function testFlagEnumValidity<TEnum extends IFlags, IFlags = Record<IFlagKey, IFlagKey>>(
+  testEnum: TEnum & Record<Exclude<keyof TEnum, keyof IFlags>, ['Key must follow `IFlagKey` format']>
+): true;
```
LetItRock, line 20: nice! :)

LetItRock (approved):

PR #4690, opened 2023-10-29
feat: add support for tls servername for azure redis

Cliftonz:


scopsy (approved):

packages/application-generic/src/services/in-memory-provider/providers/elasticache-cluster-provider.ts
```diff
@@ -48,7 +48,9 @@ export const getElasticacheClusterProviderConfig =
       keepAlive: process.env.REDIS_CLUSTER_KEEP_ALIVE,
       family: process.env.REDIS_CLUSTER_FAMILY,
       keyPrefix: process.env.REDIS_CLUSTER_KEY_PREFIX,
-      tls: process.env.REDIS_CLUSTER_TLS as ConnectionOptions,
```
rifont, line 51: Are we deprecating the `REDIS_CLUSTER_TLS` env var, or was this a mistake earlier?

Cliftonz: @rifont what do you mean? We needed to do this so that we can connect to azure redis with ssl.

p-fernandez: @rifont that's a bit of a legacy env var we left to reuse. It comes from before when there was just an instance of Redis for all the platform.

p-fernandez:
Now this is PR is merged but I think we should have done the following:
- create a specific Azure Redis provider config to set the things properly instead of reusing other providers configs
- added an env variable to select cloud provider
- based on the cloud provider selected to apply the proper InMemoryProviderService configs through the strategy pattern for the Workflow Engine and the Cache / Distributed Lock cluster

Cliftonz:
I think this 

> Now this is PR is merged but I think we should have done the following:
> 
>     * create a specific Azure Redis provider config to set the things properly instead of reusing other providers configs
> 
>     * added an env variable to select cloud provider
> 
>     * based on the cloud provider selected to apply the proper InMemoryProviderService configs through the strategy pattern for the Workflow Engine and the Cache / Distributed Lock cluster

I think this is very vaild and I think we need to separate the configuration where I can independently configure the cache and the queue separately.

PR #4691, opened 2023-10-29
Added a scroll to top option

vaibhav-009:
Added a scroll to top option at the end of the readme file. On clicking, it will scroll to the top.

User will not have to scroll through the long file.

image


scopsy:
Hi @vaibhav-009  but we will leave the current readme as is, without adding this button to avoid clutter 🙏

PR #4697, opened 2023-10-29
fix(api): swagger schema for integrations return array

### LINKED ISSUE #4696: 📚 Docs Feedback: Swagger API Schema Incorrect
### Duplicates

- [X] I have searched the existing issues

### Related page

https://api.novu.co/api#/Integrations/IntegrationsController_getIntegrations

### Kind of issue

Other

### Issue description

Incorrect Swagger Specification for `GetIntegrations` and `GetActiveIntegrations`. It should be an array of integrations not just one.

image

### Context 🔦

I was working on `novu-rust` integrations module support while implementing the methods I saw that swagger spec shows `getIntegrations` returns a single `IntegrationResponseDto` instead of a list to verify I checked the `novu-go` module and it should be a list as per the sdk as well.

king-11:

correct swagger schema for integrations module

fixes #4696


king-11:
@Cliftonz requesting your review

apps/api/src/app/integrations/integrations.controller.ts
```diff
@@ -60,7 +60,7 @@ export class IntegrationsController {
 
   @Get('/')
   @ApiOkResponse({
-    type: IntegrationResponseDto,
+    type: IntegrationResponseDto[],
```
scopsy, line 63: I'm not sure if this will solve the issue, I believe there is an `isArray` property that needs to be passed, and this should be as it was previously.

scopsy (changes requested):
Have you had a chance to test the the generated Swagger by visiting `http://localhost:3000/api` ? I believe that the current implementation will not work

apps/api/src/app/integrations/integrations.controller.ts
king-11: I have made changes based on what was used in other places, should work now

Cliftonz:
Looks correct from the docs
https://docs.nestjs.com/openapi/types-and-parameters#arrays

scopsy (approved):

king-11:
Hi @scopsy I didn't really get a chance to run the full novu setup on my system is there some other faster option to see the changes if it's still required.

king-11:
@scopsy can you also add `hacktoberfest` tag 😬 I made the PR during that period.

PR #4702, opened 2023-10-30
feat(app-generic): create filters service

p-fernandez:

Creates Filters service.

Gathering in the same domain the current filters for conditions and message matching, will help to be able to implement payload filtering properly for the broadcast.


apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -69,12 +69,6 @@ export class SendMessage {
     const stepType = command.step?.template?.type;
 
     if (!command.payload?.$on_boarding_trigger) {
-      const usedFilters = shouldRun.conditions.reduce(MessageMatcher.sumFilters, {
```
p-fernandez, line 72: Moving this bit of code that belongs to MessageMatcher simplifies the inheritance of the `sumFilters` functionality.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -159,9 +153,7 @@ export class SendMessage {
       identifier: command.job.identifier,
     });
 
-    const data = await this.matchMessage.getFilterData(messageMatcherCommand);
```
p-fernandez, line 162: Call MessageMatcher to get the filter data (subscriber and payload) to then send it again to the MessageMatcher doesn't make much sense to me.

packages/application-generic/src/services/filters/filters.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+    if (operator === FieldOperatorEnum.IN) {
+      return actualValue.includes(filterValue);
+    }
+    if (operator === FieldOperatorEnum.IS_DEFINED) {
+      return actualValue !== undefined;
+    }
+
+    return false;
+  }
```
p-fernandez, line 82: Isolate this in a function will help a lot when implementing the payload filtering for the Broadcast.

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
```diff
@@ -67,13 +67,8 @@ export class SelectIntegration {
           continue;
         }
 
-        const { passed } = await this.conditionsFilter.filter(
-          ConditionsFilterCommand.create({
-            filters: currentIntegration.conditions,
-            environmentId: command.environmentId,
-            organizationId: command.organizationId,
-            userId: command.userId,
```
p-fernandez, line 75: This wasn't even used in the conditions filtering.

djabarovgeorge:
Hey, I noticed you refactored the message-matcher, unfortunately, message-matcher logic was merged to the condition-filter class but was yet to merge because it is in
https://github.com/novuhq/novu/pull/4187
I made this change in order to remove the duplication code we had so we will not need to maintain two classes with the same responsibility.
I wonder if we could merge those changes there as well?

p-fernandez:
> Hey, I noticed you refactored the message-matcher, unfortunately, message-matcher logic was merged to the condition-filter class but was yet to merge because it is in #4187 I made this change in order to remove the duplication code we had so we will not need to maintain two classes with the same responsibility. I wonder if we could merge those changes there as well?

Thanks for flagging this. Let me check your PR and I will try to bring all the most changes possible to this so we both have the most code aligned. Will keep you updated.

packages/application-generic/src/services/filters/filter-conditions.service.ts
p-fernandez, line 1: This is 99% the content of MessageMatcher use case with some flow mods to gather all the related logic plus a change on the filter on subscriber.

packages/application-generic/src/services/filters/filter-conditions.service.ts
```diff
@@ -0,0 +1,636 @@
...
+      );
+    }
+
+    if (filter.on === FilterPartTypeEnum.SUBSCRIBER) {
+      passed = this.processFilterEquality(
+        filter,
+        filterProcessingDetails,
+        filterData
+      );
+    }
```
p-fernandez, line 355: As I understand this filter is based on the custom properties stored in the subscriber entity instead of processing the filter based on the variables of the event trigger payload, we need to do it in the stored subscriber entity that is being retrieved by `this.getFilterData` earlier on and stored in `filterData`.

packages/application-generic/src/services/filters/filters.service.ts
```diff
@@ -0,0 +1,157 @@
...
+  IBaseFieldFilterPart,
+  ICondition,
+  ITenantDefine,
+  ITriggerPayload,
+} from '@novu/shared';
+
+export interface IFilterData {
+  subscriber?: SubscriberEntity;
+  payload?: any;
+}
```
djabarovgeorge, line 15: Isn't this one a duplication of IFilterVariables?

packages/application-generic/src/services/filters/filter-conditions.service.ts
```diff
@@ -0,0 +1,636 @@
...
+  private messageRepository: MessageRepository;
+  private subscriberRepository: SubscriberRepository;
+
+  constructor() {
+    super();
+    this.environmentRepository = new EnvironmentRepository();
+    this.executionDetailsRepository = new ExecutionDetailsRepository();
+    this.jobRepository = new JobRepository();
+    this.messageRepository = new MessageRepository();
+    this.subscriberRepository = new SubscriberRepository();
```
djabarovgeorge, line 97: Could you elaborate more on why not use injection here please?
The FilterConditionsService is part of worker shared module so the dal repos should exist in the same context.

packages/application-generic/src/services/filters/filter-conditions.service.ts
```diff
@@ -0,0 +1,636 @@
...
+  }
+
+  public async filter(
```
djabarovgeorge, line 100: do we have any conventions regarding the main service method? like we have execute for the use case.

djabarovgeorge (commented):
Looks like a lot of work, did not find anything that caught my eye, so would love it if someone else would look into it. did you have anywhere challenges you aren't sure about? so I could double-check it.

packages/application-generic/src/services/filters/filters.service.ts
p-fernandez: You are right. It is just a subset. I will try to check if I can reuse IFilterVariables there. Good spot.

p-fernandez: It works but now I should merge the `data` parameter and `variables` and do it in a way that the different use cases work properly.

packages/application-generic/src/services/filters/filter-conditions.service.ts
p-fernandez: Me just being messy.

packages/application-generic/src/services/filters/filter-conditions.service.ts
p-fernandez: As far as I know not, mainly because we indistinctly use Service for any class that does any operation all across the different layers and domains, so it is hard to come to a convention. I chose `filter` as it was the closer to the domain operation we are executing.

p-fernandez:
> Looks like a lot of work, did not find anything that caught my eye, so would love it if someone else would look into it. did you have anywhere challenges you aren't sure about? so I could double-check it.

My only kind of blocker is to understand how to set the precedence between the payload provided and the `getFilterData` operation. I considered only for the Subscriber filter we needed to take precedence of the custom properties stored in the database, but I ignore if we offered the possibility of override it with the payload of the trigger event.

ainouzgali:
Due to time constraints, this great refactor will be revisited in the near future.

PR #4711, opened 2023-10-30
Create Workflow Override Repository

djabarovgeorge:
* Channel Preferences

 
### Definition of Done
Create a new entity and schema for WorkflowOverride
* the entity will store the workflow → tenant relation
* fields: _id, _workflowId, _tenantId, active, preferences

### DIFF
--- a/libs/dal/src/repositories/workflow-override/workflow-override.entity.ts
+++ b/libs/dal/src/repositories/workflow-override/workflow-override.entity.ts
@@ -1,0 +1,43 @@
+import { IPreferenceChannels } from '@novu/shared';
+
+import type { OrganizationId } from '../organization';
+import type { EnvironmentId } from '../environment';
+import type { ChangePropsValueType } from '../../types';
+import { WorkflowOverrideId } from './types';
+import { NotificationGroupEntity } from '../notification-group';
+import { TenantEntity } from '../tenant';
+
+export class WorkflowOverrideEntity {
+  _id: WorkflowOverrideId;
+
+  _organizationId: OrganizationId;
+
+  _environmentId: EnvironmentId;
+
+  _workflowId: string;
+
+  readonly workflow?: NotificationGroupEntity;
+
+  _tenantId: string;
+
+  readonly tenant?: TenantEntity;
+
+  active: boolean;
+
+  preferenceSettings: IPreferenceChannels;
+
+  deleted: boolean;
+
+  deletedAt?: string;
+
+  deletedBy?: string;
+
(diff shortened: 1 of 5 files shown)


libs/dal/src/repositories/workflow-override/workflow-override.schema.ts
```diff
@@ -0,0 +1,81 @@
...
+      type: Schema.Types.ObjectId,
+      ref: 'Environment',
+    },
+    _organizationId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Organization',
+    },
+    _workflowId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Environment',
```
BiswaViraj, line 24:
```suggestion
      ref: 'NotificationTemplate',
```

libs/dal/src/repositories/workflow-override/workflow-override.schema.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+    _tenantId: {
+      type: Schema.Types.ObjectId,
+      ref: 'Organization',
```
BiswaViraj, line 28:
```suggestion
      ref: 'Tenant',
```

BiswaViraj (changes requested):

libs/dal/src/repositories/workflow-override/workflow-override.schema.ts
djabarovgeorge: Thanks for noticing 🤦‍♂️

libs/dal/src/repositories/workflow-override/workflow-override.schema.ts
djabarovgeorge: Thanks for noticing 🤦‍♂️

libs/dal/src/repositories/workflow-override/workflow-override.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+export class WorkflowOverrideEntity {
+  _id: WorkflowOverrideId;
+
+  _organizationId: OrganizationId;
+
+  _environmentId: EnvironmentId;
+
+  _workflowId: string;
+
+  readonly workflow?: NotificationGroupEntity;
```
djabarovgeorge, line 19:
```suggestion
  readonly workflow?: NotificationTemplateEntity;
```

LetItRock (approved):
🚀

PR #4714, opened 2023-10-30
fix(web): variants project ui issues

LetItRock:

Fixed two small UI issues found for the Variants project:
- the workflow node dropdown menu was closing when hovered on the last item (which was rendered outside the node)
- the digest node was missing the content information about the time

https://github.com/novuhq/novu/assets/2607232/8463be05-f97a-4247-8e17-c534845ed64f

### DIFF
--- a/apps/web/src/pages/templates/components/VariantItemCard.tsx
+++ b/apps/web/src/pages/templates/components/VariantItemCard.tsx
@@ -131,7 +131,11 @@ export const VariantItemCard = ({
     channel: StepTypeEnum;
     stepUuid: string;
   }>();
-  const subtitle = useStepSubtitle(variant, channel);
+  const subtitle = useStepSubtitle({
+    path: `steps.${stepIndex}.variants.${variantIndex}`,
+    step: variant,
+    channelType: channel,
+  });
   const navigate = useNavigate();
   const basePath = useBasePath();
   const [areConditionsOpened, setConditionsOpened] = useState(false);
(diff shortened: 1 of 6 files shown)


apps/web/src/pages/templates/components/VariantItemCard.tsx
```diff
@@ -131,7 +131,11 @@ export const VariantItemCard = ({
     channel: StepTypeEnum;
     stepUuid: string;
   }>();
-  const subtitle = useStepSubtitle(variant, channel);
+  const subtitle = useStepSubtitle({
+    path: `steps.${stepIndex}.variants.${variantIndex}`,
+    step: variant,
+    channelType: channel,
+  });
```
LetItRock, line 138: This hook is used in the step node and the variant node. It is responsible for creating the node content that's why I'm passing here the form path so it can calculate that text.

Screenshot 2023-10-30 at 22 55 02
Screenshot 2023-10-30 at 22 54 45

apps/web/src/pages/templates/workflow/workflow/node-types/WorkflowNodeActions.tsx
```diff
@@ -126,6 +126,7 @@ export const WorkflowNodeActions = ({
             />
           )}
           <Dropdown
+            withinPortal
```
LetItRock, line 129: fix the node dropdown issue

ainouzgali (approved):

PR #4717, opened 2023-10-31
Create workflow override use-case

djabarovgeorge:
* Channel Preferences

 
### Definition of Done
* add create workflow override usecase.
* e2e

### DIFF
--- a/apps/api/src/app/shared/dtos/preference-channels.ts
+++ b/apps/api/src/app/shared/dtos/preference-channels.ts
@@ -1,14 +1,30 @@
 import { ApiPropertyOptional } from '@nestjs/swagger';
+import { IPreferenceChannels } from '@novu/shared';
+import { IsBoolean, IsOptional } from 'class-validator';
 
-export class PreferenceChannels {
+export class PreferenceChannels implements IPreferenceChannels {
   @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
   email?: boolean;
+
   @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
   sms?: boolean;
+
   @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
   in_app?: boolean;
+
   @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
   chat?: boolean;
+
   @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
   push?: boolean;
 }
--- a/apps/api/src/app/workflows/dto/create-workflow-override-request.dto.ts
+++ b/apps/api/src/app/workflows/dto/create-workflow-override-request.dto.ts
@@ -1,0 +1,37 @@
+import { IsBoolean, IsDefined, IsOptional, IsString, ValidateNested } from 'class-validator';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { Type } from 'class-transformer';
+
+import { ICreateWorkflowOverrideRequestDto } from '@novu/shared';
+
+import { PreferenceChannels } from '../../shared/dtos/preference-channels';
+
+export class CreateWorkflowOverrideRequestDto implements ICreateWorkflowOverrideRequestDto {
+  @ApiPropertyOptional()
+  @IsString()
+  @IsOptional()
+  triggerIdentifier?: string;
+
+  @ApiPropertyOptional()
+  @IsString()
+  @IsOptional()
+  _workflowId?: string;
+
+  @ApiProperty()
+  @IsString()
+  @IsDefined()
+  tenantIdentifier: string;
+
+  @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
+  active?: boolean;
+
+  @ApiPropertyOptional({
+    type: PreferenceChannels,
+  })
+  @IsOptional()
+  @ValidateNested()
+  @Type(() => PreferenceChannels)
+  preferenceSettings?: PreferenceChannels;
+}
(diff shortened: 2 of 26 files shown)


apps/api/src/app/workflows/workflow.controller.ts
```diff
@@ -202,4 +207,32 @@ export class WorkflowController {
...
+  @ApiOperation({
+    summary: 'Create workflow override',
+    description:
+      'In order to create workflow override please make sure to identify the workflow by triggerIdentifier or _workflowId, and tenant by tenantIdentifier',
+  })
+  @ExternalApiAccessible()
+  createWorkflowOverride(
+    @UserSession() user: IJwtPayload,
+    @Body() body: CreateWorkflowOverrideRequestDto,
+    @Param('workflowId') workflowId: string
```
djabarovgeorge, line 223: redundant 🧹
```suggestion
```

apps/api/src/app/workflows/dto/create-workflow-override-request.dto.ts
```diff
@@ -0,0 +1,37 @@
...
+export class CreateWorkflowOverrideRequestDto implements ICreateWorkflowOverrideRequestDto {
+  @ApiPropertyOptional()
+  @IsString()
+  @IsOptional()
+  triggerIdentifier?: string;
+
+  @ApiPropertyOptional()
+  @IsString()
+  @IsOptional()
+  _workflowId?: string;
```
djabarovgeorge, line 18: thinking about it again i wonder if update to `workflowId` as the outside world should not be aware of the internal conversions.

```suggestion
  workflowId?: string;
```

apps/api/src/app/workflows/workflow.controller.ts
djabarovgeorge: removed in the next branch 'NV-3135-add-update-workflow-override-use-case'

apps/api/src/app/shared/dtos/preference-channels.ts
```diff
@@ -1,14 +1,30 @@
 import { ApiPropertyOptional } from '@nestjs/swagger';
+import { IPreferenceChannels } from '@novu/shared';
+import { IsBoolean, IsOptional } from 'class-validator';
 
-export class PreferenceChannels {
+export class PreferenceChannels implements IPreferenceChannels {
   @ApiPropertyOptional()
+  @IsBoolean()
+  @IsOptional()
```
LetItRock, line 8: ❤️

apps/api/src/app/workflows/workflow.controller.ts
LetItRock: I actually feel like we should have this endpoint be like `POST /workflows/:workflowId/overrides` and then we won't need to have the `_workflowId` nor `triggerIdentifier` in the request body. We only use the `triggerIdentifier` to trigger the workflow, in all other cases like get/update/delete/status we use `workflowId`.

apps/api/src/app/workflows/usecases/create-workflow-override/create-workflow-override.usecase.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (!tenant) {
+      throw new NotFoundException(`Tenant with identifier ${command.tenantIdentifier} is not found`);
+    }
+
+    let workflow: NotificationTemplateEntity | null = null;
+
+    if (!command.triggerIdentifier && !command._workflowId) {
+      throw new BadRequestException(`Either triggerIdentifier or _workflowId must be provided`);
+    }
```
LetItRock, line 50: maybe let's put it to the top, there is no sense in fetching the tenant and then rejecting the request

apps/api/src/app/workflows/usecases/create-workflow-override/create-workflow-override.usecase.ts
```diff
@@ -0,0 +1,85 @@
...
+      workflow = await this.notificationTemplateRepository.findByTriggerIdentifier(
+        command.environmentId,
+        command.triggerIdentifier
+      );
+
+      if (!workflow) {
+        throw new NotFoundException(`Workflow with identifier ${command.triggerIdentifier} is not found`);
+      }
+
+      return { tenant, workflow };
```
LetItRock, line 62: if we remove this

apps/api/src/app/workflows/usecases/create-workflow-override/create-workflow-override.usecase.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (command._workflowId) {
+      workflow = await this.notificationTemplateRepository.findOne({
+        _environmentId: command.environmentId,
+        _id: command._workflowId,
+      });
+
+      if (!workflow) {
+        throw new NotFoundException(`Workflow with _workflowId ${command._workflowId} is not found`);
+      }
```
LetItRock, line 73: and this, the code will be more clean :)

apps/api/src/app/workflows/usecases/create-workflow-override/create-workflow-override.usecase.ts
```diff
@@ -0,0 +1,85 @@
...
+      );
+
+      if (!workflow) {
+        throw new NotFoundException(`Workflow with identifier ${command.triggerIdentifier} is not found`);
+      }
+
+      return { tenant, workflow };
+    }
+
+    if (command._workflowId) {
```
LetItRock, line 65: make this `else if`

apps/api/src/app/workflows/usecases/create-workflow-override/create-workflow-override.usecase.ts
djabarovgeorge: Yes, agree this one is independent and go up :)

apps/api/src/app/workflows/usecases/create-workflow-override/create-workflow-override.usecase.ts
djabarovgeorge: yes, but the error message won't be informative as before, but i think i am ok with it :)

apps/api/src/app/workflows/workflow.controller.ts
djabarovgeorge: Actually, it makes a lot of sense to me, i make it like this because of our sync before the cycle. i thought we wanted to couple it to the workflow entity this way.

Do you mind if we merge all of the existing branches to `workflow-override` so i could then create a new PR in order to refactor the location of all of the routes altogether?

PR #4718, opened 2023-10-31
feat(infra): safe value in-memory provider configs

p-fernandez:

Converts null and undefined strings to proper values.

Azure Redis configurations demand non empty values even if not set but we don't want to pass them down to our system.

### DIFF
--- a/packages/application-generic/src/services/in-memory-provider/providers/variable-mappers.ts
+++ b/packages/application-generic/src/services/in-memory-provider/providers/variable-mappers.ts
@@ -1,0 +1,19 @@
+/**
+ * Due some problems with Azure Redis DB that doesn't allow for certain
+ * configuration values to be empty or have an empty string and as we don't
+ * want to process them in our provider configuration files, we implement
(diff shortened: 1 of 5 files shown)


packages/application-generic/src/services/in-memory-provider/providers/variable-mappers.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Due some problems with Azure Redis DB that doesn't allow for certain
+ * configuration values to be empty or have an empty string and as we don't
+ * want to process them in our provider configuration files, we implement
+ * this mapper function to be able to overcome that limitation in Azure
+ * temporarily while we find a better solution
+ */
+export const convertStringValues = (
```
Cliftonz, line 8: Why do you have this function written twice?

p-fernandez: I forgot to remove it from the initial location in the index file. Thanks for spotting it.

Cliftonz (approved):

PR #4724, opened 2023-10-31
feat: add support for variables in filter values

ainouzgali:

When filtering a step, allow using payload/subscriber/actor.tenant values 
for example:

Payload :
  Key: data.customProp
  Value: {{subscriber.data.someOtherValue}}

Subscriber:
   Key: data.subscriberValue
   Value: {{someProp.from.payload}}

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
@@ -21,6 +21,7 @@ import {
   Instrument,
   ConditionsFilterCommand,
   ConditionsFilter,
+  IFilterVariables,
 } from '@novu/application-generic';
 import {
   JobEntity,
@@ -28,6 +29,8 @@ import {
   NotificationTemplateRepository,
   JobRepository,
   JobStatusEnum,
+  TenantEntity,
+  TenantRepository,
 } from '@novu/dal';
 
 import { SendMessageCommand } from './send-message.command';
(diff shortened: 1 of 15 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -164,8 +177,7 @@ export class SendMessage {
           isTest: false,
           isRetry: false,
           raw: JSON.stringify({
-            payload: shouldRun.variables,
-            filters: command.step.filters,
+            conditions: shouldRun.conditions,
```
ainouzgali, line 180: So it presented with the filters ui

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -126,31 +134,36 @@ export class SendMessage {
...
-  private async filter(command: SendMessageCommand) {
+  private async filter(command: SendMessageCommand, payload: IFilterVariables) {
     const shouldRun = await this.conditionsFilter.filter(
       ConditionsFilterCommand.create({
         filters: command.job.step.filters || [],
         environmentId: command.environmentId,
         organizationId: command.organizationId,
         userId: command.userId,
+        step: command.step,
+        job: command.job,
```
ainouzgali, line 165: Unrelated to this PR. It was not filtering correctly a group filter. This fixes it.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -223,6 +235,36 @@ export class SendMessage {
...
+      subscriber,
+      payload: command.payload,
+      step: {
+        digest: !!command.events?.length,
+        events: command.events,
+        total_count: command.events?.length,
+      },
+      ...(tenant && { tenant }),
+      ...(actor && { actor }),
+    };
```
ainouzgali, line 265: This is used in all send message usecases and now in the filters as well. Extracted this object so we fetch and build it once.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -411,7 +422,11 @@ export class ConditionsFilter extends Filter {
 
     if (child.on === FilterPartTypeEnum.WEBHOOK) {
       if (process.env.NODE_ENV === 'test') return true;
-
+      child.value = await this.compileFilter(
+        child.value,
+        variables,
+        command.job
+      );
```
ainouzgali, line 429: Should this be in the value of webhook? I am not 100% certain

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
```diff
@@ -555,6 +576,9 @@ export class ConditionsFilter extends Filter {
       ? command.variables?.payload
       : command.job?.payload ?? undefined;
 
+    filterVariables.step = command.variables?.step ?? undefined;
+    filterVariables.actor = command.variables?.actor ?? undefined;
```
LetItRock, line 580: no need in `?? undefined` part ;)

PR #4725, opened 2023-10-31
fix(dal): Mongodb data layer loose typings

rifont:
* DAL
  * Upgrade generic typings for Entity -> DbModel transformation to transparently handle PK & FK identifiers and date attributes
    * Takes a type-widening approach to handle the PK and FK identifiers, as there are currently many type errors with only `Types.ObjectId` from `mongoose`
    * Tests for new generic typings
    * Add generic type transformations for all date attributes
  * Make the `create` and `update` BaseRepository methods fully typed
  * Fix all repository typings to match required Entity and DbModel typings
* Worker
  * TBA, there are 16 type errors to fix in Worker
* API
  * TBA, there are 28 type errors to fix in API

MongoDB currently has loose any typings for all of the `BaseRepository` typings, most importantly the mutative operations such as `create`, `insertMany`, and `update`. This can cause the database layer to become corrupted if incorrect data is sent and processed, as demonstrated in the intellisense snippet for `apiServiceLevel` below allowing `any` types.

Pasted Graphic 7

We want to enforce strict typings for mutative operations to reduce the likelihood of database layer corruption.

`apiServiceLevel` field after the type fixes, fully typed and producing type error:

Pasted Graphic 6


libs/dal/src/types/helpers.ts
```diff
@@ -1,5 +1,42 @@
...
+export type TransformValues<T, U, V> = Identity<
+  Omit<T, Extract<keyof T, U>> & {
+    [P in keyof Pick<T, Extract<keyof T, U>>]: V;
+  }
+>;
+
+/**
+ * Transform the values of T with keys in union of type `ObjectIdKey` to type `Types.ObjectId`
+ */
+export type TransformEntityToDbModel<T> = TransformValues<TransformValues<T, ObjectIdKey, ObjectIdType>, DateKey, Date>;
```
rifont, line 44: This was the old `ChangePropsValueTypes` generic, updated to use the new base generic, with a new Date property transform, and renamed to better fit the purpose.

libs/dal/src/repositories/base-repository.ts
```diff
@@ -7,9 +7,23 @@ import {
...
+ */
+type QueryData<T_DBModel, T_Enforcement> = FilterQuery<T_DBModel> & T_Enforcement;
+
+/**
+ * The create type used for enforcing the type of the create data payload
+ */
+type CreateData<T_DBModel, T_Enforcement> = Omit<T_DBModel, '_id'> & T_Enforcement;
+
+export class BaseRepository<
+  T_DBModel extends { _id: ObjectIdType },
```
rifont, line 23: All database models must include an `_id` attribute.

libs/dal/src/repositories/base-repository.ts
```diff
@@ -7,9 +7,23 @@ import {
...
+  T_MappedEntity,
+  T_Enforcement extends Partial<T_DBModel>
```
rifont, line 25: Adding some type safety to the enforced attributes.

rifont:
The Worker and API collectively have 44 type errors we need to resolve after we finalise the fixes for the DAL.

Let's first agree on the direction taken in the DAL fixes, then we can fix the typing issues on the Worker and API.

libs/dal/src/types/mongo.ts
```diff
@@ -0,0 +1,69 @@
+import { Condition } from 'mongoose';
```
rifont, line 1: Stricter overrides for all the `mongoose` generics, linked in the `@see` directive.

libs/dal/src/types/helpers.ts
```diff
@@ -1,5 +1,63 @@
...
+ */
+export type DotPrefix<T extends string> = T extends '' ? '' : `.${T}`;
+
+/**
+ * Construct a union type from the leaf key paths in T
+ *
+ * The following reference specifically mentions Mongo as a use-case for the generic.
+ * @see https://stackoverflow.com/a/68404823
+ */
+export type LeafKeys<T> = (
```
rifont, line 57: Used in the `RootQuerySelector` generic to extract the leaf keys used in some queries.

libs/dal/src/types/mongo.ts
```diff
@@ -0,0 +1,69 @@
...
+    $language?: string;
+    $caseSensitive?: boolean;
+    $diacriticSensitive?: boolean;
+  };
+  /** @see https://www.mongodb.com/docs/manual/reference/operator/query/where/#op._S_where */
+  $where?: string;
+  /** @see https://www.mongodb.com/docs/manual/reference/operator/query/comment/#op._S_comment */
+  $comment?: string;
+} & Partial<{
+  [Key in LeafKeys<T>]: T[Key] | T[Key][];
```
rifont, line 30: `Key` can be a dot notation of a leaf field - this still needs some work to make sure the `T[Key]` resolves to the correct type.

libs/dal/src/types/mongoose.ts
```diff
@@ -0,0 +1,87 @@
...
+    $language?: string;
+    $caseSensitive?: boolean;
+    $diacriticSensitive?: boolean;
+  };
+  /** @see https://www.mongodb.com/docs/manual/reference/operator/query/where/#op._S_where */
+  $where?: string;
+  /** @see https://www.mongodb.com/docs/manual/reference/operator/query/comment/#op._S_comment */
+  $comment?: string;
+} & Partial<{
+  [Key in DeepKeys<T>]: ExtractDot<T, Key> | ExtractDot<T, Key>[];
```
rifont, line 48: We are using the helper `DeepKeys` and `ExtractDot` generics here to introduce stronger typings on the `RootQuerySelect`. This allows us to strongly type queries targeting nested attributes through Mongo's Dot Notation, as seen in `"partnerConfigurations.configurationId"?: string | string[] | undefined` below:
image

There is significant work required to update the codebase to align on consistent use of these typings, because `mongoose` is very flexible by design. We will most likely want to decide on a pattern to keep the typings as simple as possible.

Other notation such as `'partnerConfigurations.$': 1` is not currently covered, and we should discuss if we'd like to include this pattern as a valid typing.

rifont: These generic typings for `mongoose` are not presently in use, and are introduced in this PR only for discussion.

libs/dal/src/types/helpers.ts
djabarovgeorge: Love this one, now we won't need to do manual mapping for every entity.

libs/dal/src/types/helpers.ts
```diff
@@ -1,5 +1,119 @@
...
+ * TODO: Investigate and decide on approach to choose either `Types.ObjectId` or `string`
+ *
+ * @see https://mongoosejs.com/docs/schematypes.html#objectids
+ */
+export type ObjectIdType = Types.ObjectId | string;
+
+/**
+ * The type union of the PK identifier and the template literal type of an FK identifier
+ */
+export type ObjectIdKey = '_id' | `_${string}Id`;
```
djabarovgeorge, line 18: Cute! I was not aware of the benefit of using Template Literal in Types like this.

libs/dal/src/repositories/base-repository.ts
```diff
@@ -1,15 +1,28 @@
...
+
+/**
+ * The query type used for enforcing the type of the query data payload
+ */
+type QueryData<T_DBModel, T_Enforcement> = FilterQuery<T_DBModel> & T_Enforcement;
+
+/**
+ * The create type used for enforcing the type of the create data payload
+ */
+type CreateData<T_DBModel, T_Enforcement> = Omit<T_DBModel, '_id'> & T_Enforcement;
```
djabarovgeorge, line 19: I think we should add here some more params to Omit like all the date ones, `createdAt, updatedAt` maybe the `deleted` as well. There could be a couple of more as well.

djabarovgeorge (commented):
So my couple of cents related to the issue I initially opened is that in QueryData we do not validate type the same way we do in CreateData, meaning while executing `findOne` we could pass to it `{ nonExistingParam: 'id' }` which could fail in runtime.
Let me know what you think about it, maybe i missed something on the way.

libs/dal/src/repositories/base-repository.ts
```diff
@@ -1,15 +1,28 @@
...
+
+export class BaseRepository<
+  T_DBModel extends { _id: Types.ObjectId | string },
```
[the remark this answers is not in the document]
rifont: `BaseRepository` is dependent on the `_id` attribute in the `insertMany` method - https://github.com/novuhq/novu/pull/4725/files/98a5324a471ffc6ff5aa03e4d169ebca1d33b928#diff-f1be5cdc382c28f12393da10b6aab3123a77a6e7b30ae019a654758c8e790074R149.

I agree, creating a BaseEntity class will make the code more DRY and maintainable.

libs/dal/src/repositories/base-repository.ts
rifont: Good catch. I agree that some date attributes should be removed, however there are cases where we might like to customise the property outside the DAL. A good example here is the `expireAt` property,  which we often specify as part of an `insert` operation.

We could find a middle ground by selecting all date attributes via the Date template literal type, then make them optional via Partial.

@LetItRock what do you think?

libs/dal/src/repositories/base-repository.ts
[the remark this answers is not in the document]
rifont: I have added `IEntity` and `IDBModel` as opposed to a base class, to give us more flexibility when decorating the `_id` attribute. This solution also enables us to use custom `_id` string types when implementing the Entity class.

Let me know what you think @djabarovgeorge

libs/dal/src/repositories/base-repository.ts
rifont: I have added `IEntity` and `IDBModel` as opposed to a base class, to give us more flexibility when decorating the `_id` attribute. This solution also enables us to use custom `_id` string types when implementing the Entity class.

Let me know what you think @djabarovgeorge

libs/dal/src/repositories/base-repository.ts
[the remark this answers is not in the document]
djabarovgeorge: Looks good to me :)

libs/dal/src/repositories/base-repository.ts
```diff
@@ -77,7 +87,7 @@ export class BaseRepository<T_DBModel, T_MappedEntity, T_Enforcement> {
   }
 
   async *findBatch(
-    query: FilterQuery<T_DBModel> & T_Enforcement,
+    query: QueryData<T_DBModel, T_Enforcement>,
```
djabarovgeorge, line 90: So my couple of cents related to the issue I initially opened is that in QueryData we do not validate type the same way we do in CreateData, meaning while executing findOne we could pass to it { nonExistingParam: 'id' } which could fail in runtime.
Let me know what you think about it, maybe i missed something on the way.
@rifont

scopsy:
@rifont should we close this one due to inactivity?

PR #4727, opened 2023-10-31
Fixes: #4256

### LINKED ISSUE #4256: [NV-2732] Workflow adjust the initial workflow steps location
## What?
Centralize the initial location of the workflow tree.
Current location:

image

## Why?
After the recent iteration, the initialized location of the workflow tree needs to be updated.

## Definition of done
- the workflow tree should be centered
- when there is no enough space on the screen it should be centered to the top (like it is right now)
- it looks good on smaller screens 1280x720 and also on the bigger ex. 1920x1080

anmol-matharu-100:

Element has been centered as required

Closes #4256

### DIFF
--- a/apps/web/src/components/workflow/FlowEditor.tsx
+++ b/apps/web/src/components/workflow/FlowEditor.tsx
@@ -91,10 +91,12 @@ export function FlowEditor({
 
   useEffect(() => {
     const clientWidth = reactFlowWrapper.current?.clientWidth;
-    const middle = clientWidth ? clientWidth / 2 - 100 : 0;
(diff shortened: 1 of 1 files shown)


scopsy (changes requested):
@anmol-matharu-100 Unfortunately when there are multiple nodes in the screen it still starts from the same position.

We need to make it a bit more dynamic, so that it centers the whole block  of items like in the second image
CleanShot 2023-11-06 at 09 29 47@2x

Expected behaviour:
CleanShot 2023-11-06 at 09 30 29@2x

When having more nodes than bounding space in view, we should just attach to the top like previously.

LetItRock:
closing as not active

PR #4733, opened 2023-10-31
feat(infra): create dedicated services for in-memory for wf and cache

p-fernandez:

Creates a parent class for both Workflow and Cache in-memory provider instances, to be used between the features and the InMemoryProviderService.

This change will help us to implement the selection of the in-memory provider for each service in a more dynamic way and also to be able to independently set cluster mode.
Is a previous step to be able to implement properly Azure Redis DB provider and from there any in-memory provider.

### DIFF
--- a/apps/api/src/app/widgets/e2e/get-unseen-count.e2e.ts
+++ b/apps/api/src/app/widgets/e2e/get-unseen-count.e2e.ts
@@ -6,9 +6,8 @@ import { ChannelTypeEnum } from '@novu/shared';
 import {
   buildFeedKey,
   buildMessageCountKey,
+  CacheInMemoryProviderService,
   CacheService,
-  InMemoryProviderEnum,
-  InMemoryProviderService,
   InvalidateCacheService,
 } from '@novu/application-generic';
 
@@ -22,12 +21,12 @@ describe('Unseen Count - GET /widget/notifications/unseen', function () {
     _id: string;
   } | null = null;
 
-  let inMemoryProviderService: InMemoryProviderService;
+  let cacheInMemoryProviderService: CacheInMemoryProviderService;
   let invalidateCache: InvalidateCacheService;
 
   before(async () => {
-    inMemoryProviderService = new InMemoryProviderService(InMemoryProviderEnum.REDIS);
-    const cacheService = new CacheService(inMemoryProviderService);
+    cacheInMemoryProviderService = new CacheInMemoryProviderService();
+    const cacheService = new CacheService(cacheInMemoryProviderService);
     await cacheService.initialize();
     invalidateCache = new InvalidateCacheService(cacheService);
   });
@@ -68,7 +67,7 @@ describe('Unseen Count - GET /widget/notifications/unseen', function () {
 
     const messages = await messageRepository.findBySubscriberChannel(
       session.environment._id,
-      subscriberProfile!._id,
+      String(subscriberProfile?._id),
       ChannelTypeEnum.IN_APP
     );
     const messageId = messages[0]._id;
@@ -97,7 +96,7 @@ describe('Unseen Count - GET /widget/notifications/unseen', function () {
 
     const messages = await messageRepository.findBySubscriberChannel(
       session.environment._id,
-      subscriberProfile!._id,
+      String(subscriberProfile?._id),
       ChannelTypeEnum.IN_APP
     );
     const messageId = messages[0]._id;
@@ -126,7 +125,7 @@ describe('Unseen Count - GET /widget/notifications/unseen', function () {
 
     const messages = await messageRepository.findBySubscriberChannel(
       session.environment._id,
-      subscriberProfile!._id,
+      String(subscriberProfile?._id),
       ChannelTypeEnum.IN_APP
     );
     const messageId = messages[0]._id;
@@ -155,7 +154,7 @@ describe('Unseen Count - GET /widget/notifications/unseen', function () {
 
     const messages = await messageRepository.findBySubscriberChannel(
       session.environment._id,
-      subscriberProfile!._id,
+      String(subscriberProfile?._id),
       ChannelTypeEnum.IN_APP
     );
     const messageId = messages[0]._id;
--- a/packages/application-generic/src/health/queue-health-indicator.service.ts
+++ b/packages/application-generic/src/health/queue-health-indicator.service.ts
@@ -5,7 +5,7 @@ import {
 } from '@nestjs/terminus';
 import { Injectable, Logger } from '@nestjs/common';
 
-import { QueueBaseService } from '../services';
+import { QueueBaseService } from '../services/queues/queue-base.service';
 import { IHealthIndicator } from './health-indicator.interface';
 
 @Injectable()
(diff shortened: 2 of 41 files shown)


packages/application-generic/src/health/queue-health-indicator.service.ts
```diff
@@ -5,7 +5,7 @@ import {
 } from '@nestjs/terminus';
 import { Injectable, Logger } from '@nestjs/common';
 
-import { QueueBaseService } from '../services';
+import { QueueBaseService } from '../services/queues/queue-base.service';
```
p-fernandez, line 8: There is a circular dependency when executing the App Generic tests for the Queues and Workers. Still haven't been able to find the guilty one. These import path changes are part of that investigation.

packages/application-generic/src/services/bull-mq/bull-mq.service.ts
```diff
@@ -46,35 +43,18 @@ export {
...
-   * if MemoryDB not configured properly. That's happening in the provider
-   * mapping in the /in-memory-provider/providers/index.ts
-   */
-  private selectProvider(): InMemoryProviderEnum {
-    if (process.env.IS_DOCKER_HOSTED) {
-      return InMemoryProviderEnum.REDIS;
-    }
-
-    return InMemoryProviderEnum.MEMORY_DB;
+    this.workflowInMemoryProviderService =
```
p-fernandez, line 52: Now the BullMQ service will have no knowledge of the in-memory provider selected as it is not its responsibility.

packages/application-generic/src/services/bull-mq/old-instance-bull-mq.service.ts
```diff
@@ -49,7 +49,8 @@ export class OldInstanceBullMqService {
   constructor() {
     if (this.shouldInstantiate()) {
       this.inMemoryProviderService = new InMemoryProviderService(
-        InMemoryProviderEnum.OLD_INSTANCE_REDIS
+        InMemoryProviderEnum.OLD_INSTANCE_REDIS,
+        true
```
p-fernandez, line 53: We now have to pass the cluster mode as variable, therefore this setting here because the old Redis instance of the Workflow Engine is in cluster.

packages/application-generic/src/services/in-memory-provider/cache-in-memory-provider.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.isCluster = this.isClusterMode();
+
+    const enableAutoPipelining =
+      process.env.REDIS_CACHE_ENABLE_AUTOPIPELINING === 'true';
+
+    this.inMemoryProviderService = new InMemoryProviderService(
+      provider,
+      this.isCluster,
+      enableAutoPipelining
+    );
```
p-fernandez, line 33: With these changes this parent service and WorkflowInMemoryProviderService, will drive down which provider will be set, instead of delegating that responsibility to the InMemoryProviderService.

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.spec.ts
```diff
@@ -147,8 +146,6 @@ describe('In-memory Provider Service', () => {
       });
 
       it('should instantiate the provider properly', async () => {
-        expect(inMemoryProviderService.isClusterMode()).toEqual(true);
```
p-fernandez, line 150: Superfluous check as we inject the cluster mode in the InMemoryProviderService now when setting the tests.

packages/application-generic/src/usecases/get-feature-flag/get-is-in-memory-cluster-mode-enabled.use-case.ts
```diff
@@ -7,9 +7,9 @@ import { GetSystemCriticalFlag } from './get-system-critical-flag.use-case';
 @Injectable()
 export class GetIsInMemoryClusterModeEnabled extends GetSystemCriticalFlag {
   execute(): boolean {
-    const value =
-      process.env.IS_IN_MEMORY_CLUSTER_MODE_ENABLED ??
-      process.env.IN_MEMORY_CLUSTER_MODE_ENABLED;
+    const value = process.env.IS_IN_MEMORY_CLUSTER_MODE_ENABLED
```
p-fernandez, line 10: Change needed as there was a failing test when `IS_IN_MEMORY_CLUSTER_MODE_ENABLED` was empty string. I added an extra test to confirm right behaviour.

packages/application-generic/src/modules/queues.module.ts
```diff
@@ -23,16 +25,13 @@ import {
...
   WebSocketsWorkerService,
   WorkflowWorkerService,
   OldInstanceStandardWorkerService,
   OldInstanceWorkflowWorkerService,
 } from '../services/workers';
 
-import { SubscriberProcessQueueService } from '../services/queues/subscriber-process-queue.service';
-import { SubscriberProcessWorkerService } from '../services/workers/subscriber-process-worker.service';
-import { SubscriberProcessQueueHealthIndicator } from '../health/subscriber-process-queue.health-indicator';
-
```
rifont, line 35: Nice cleanup 👍

packages/application-generic/src/services/in-memory-provider/cache-in-memory-provider.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    return `[Provider: ${this.selectProvider()}] ${message}`;
+  }
+
+  private isClusterMode(): boolean {
+    const isClusterModeEnabled = this.getIsInMemoryClusterModeEnabled.execute();
+
+    Logger.log(
+      this.descriptiveLogMessage(
+        `Cluster mode ${
+          isClusterModeEnabled ? 'is' : 'is not'
```
rifont, line 62: Nit for readability on the differentiator.

```suggestion
          isClusterModeEnabled ? 'IS' : 'IS NOT'
```

apps/api/src/app/widgets/e2e/get-unseen-count.e2e.ts
```diff
@@ -68,7 +67,7 @@ describe('Unseen Count - GET /widget/notifications/unseen', function () {
 
     const messages = await messageRepository.findBySubscriberChannel(
       session.environment._id,
-      subscriberProfile!._id,
+      String(subscriberProfile?._id),
```
rifont, line 70: Each of these statements have the same result on typings and functionality, but the underlying issue still exists - what if `subscriberProfile` is nullish? `String(undefined)` just results in `'undefined'` as a string.

Is this fixing something in this PR?

rifont (approved):

apps/api/src/app/widgets/e2e/get-unseen-count.e2e.ts
p-fernandez: The goal of these changes is to reduce the TS warnings shown that just provide noise to the linting.
The changes casting to String is to get rid of few TypeScript warnings due the nature of the implementation of the values retrieved for the subscriber. It happens in more other test files. Technical debt we have, indeed.

The rule warning is that we are not allowing non-null assertions in the ruleset. Considering these changes were done inside of tests that if we don't have a subscriber it must fail or even if the id used to retrieve results from the database is `'null'` or `'undefined'` the tests must fail. I found that was the less intrusive change for this situation without having to touch tenths of files more.

PR #4737, opened 2023-11-01
Nv 2888 select the default language

davidsoderberg:
Move use auth controller to design system and add defaultLocale to Organization entity.


apps/web/src/api/api.client.ts
```diff
@@ -1,71 +1,3 @@
-import axios from 'axios';
-import { IParamObject } from '@novu/shared';
-import { API_ROOT } from '../config';
+import { api } from '@novu/design-system';
```
p-fernandez, line 1: I find this move complicated to understand thinking in domain terms. I will explain myself.
I always associate a design system to the smallest pieces that help to build bigger more complicated components. And this pieces are expected to be stateless. So I am struggling to think the scenario that an HTTP client implementation would live here.
Would you mind to share the intentions with this?

libs/design-system/src/config/vars.ts
```diff
@@ -0,0 +1,74 @@
+import { getContextPath, NovuComponentEnum } from '@novu/shared';
```
p-fernandez, line 1: Same as with the HTTP client I struggle to see environment variables management in a Design System. Reasoning would be similar. But I am open to hear the goal of this.

apps/web/src/api/api.client.ts
davidsoderberg: Yeah I also think it is a bit misplaced and maybe we should create client-generic package as we do have one for backends in application-generic. But I did not like to create too big of a change right now. @scopsy and @LetItRock what do you think? I could do a new refactoring on this in the cooldown.

p-fernandez:
As discussed privately we are aware this is not the ideal situation but there are some tech debt and environment variables technicalities that difficult the implementation.
Will be raised with the core team and tackled in forthcoming cycles.

libs/design-system/src/config/vars.ts
davidsoderberg: My comment above touches this as well.

p-fernandez (approved):
🌟

Issue #3934, opened 2023-08-07
🚀 Feature: Custom backendUrl in invite-member.usecase and resend-invite.usecase


mohamed-cherifi:
### 🔖 Feature description

Hi Folks, 

It would be great if when instantiating the Novu instance [here](https://github.com/novuhq/novu/blob/v0.13.x/apps/api/src/app/invites/usecases/invite-member/invite-member.usecase.ts#L38) and [here](https://github.com/novuhq/novu/blob/v0.13.x/apps/api/src/app/invites/usecases/resend-invite/resend-invite.usecase.ts#L37) we could also pass a custom `config` object in case we want to inject our own `backendUrl` for example.

Thank you.

### 🎤 Why is this feature needed ?

More flexibility for injecting a custom `backendUrl` for `invite-member.usecase` and `resend-invite.usecase`

### ✌️ How do you aim to achieve this?

Add a custom config param to the Novu instance in [invite-member.usecase](https://github.com/novuhq/novu/blob/v0.13.x/apps/api/src/app/invites/usecases/invite-member/invite-member.usecase.ts#L38).

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

pritipsingh:
Hey folks, can I take this up??

davidsoderberg:
@pritipsingh will assign you on this now :)

pritipsingh:
thanks @davidsoderberg

jainpawan21:
@pritipsingh 

Just checking on how is it going with this issue.
Do you need any help?

pritipsingh:
Hey @jainpawan21 , I am actually done with it. Just E2E test case was failing but let me just try to get done with this in 2 days and get back to you

pritipsingh:
Hey @jainpawan21 just wanted to take your review on this one. Can you please check

jsharma44:
Please check https://github.com/novuhq/novu/pull/5742m . 
Thanks

mohamed-cherifi:
Any news on this ticket?

PR #4738, opened 2023-11-01
feat: Custom backendUrl in invite-member.usecase and resend-invite.usecase #3934

pritipsingh:

More flexibility for injecting a custom backendUrl for invite-member.usecase and resend-invite.usecase
I have added config in all the interfaces for invites, resend & bulk and passed it in. Password one is left but just wanted to get it reviewed first.

Closes [3934](https://github.com/novuhq/novu/issues/3934)

### DIFF
--- a/apps/api/src/app/auth/usecases/password-reset-request/password-reset-request.usecase.ts
+++ b/apps/api/src/app/auth/usecases/password-reset-request/password-reset-request.usecase.ts
@@ -2,7 +2,7 @@ import { Injectable, UnauthorizedException } from '@nestjs/common';
 import { v4 as uuidv4 } from 'uuid';
 import { differenceInHours, differenceInSeconds, parseISO } from 'date-fns';
 import { Novu } from '@novu/node';
-import { UserRepository, UserEntity, IUserResetTokenCount } from '@novu/dal';
+import { UserRepository, UserEntity, IUserResetTokenCount, MemberRepository } from '@novu/dal';
 import { buildUserKey, InvalidateCacheService } from '@novu/application-generic';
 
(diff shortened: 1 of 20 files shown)


apps/api/src/app/auth/usecases/password-reset-request/password-reset-request.usecase.ts
```diff
@@ -14,16 +14,22 @@ export class PasswordResetRequest {
...
+    private userRepository: UserRepository,
+    private memberRepository: MemberRepository
+  ) {}
 
   async execute(command: PasswordResetRequestCommand): Promise<{ success: boolean }> {
     const email = normalizeEmail(command.email);
     const foundUser = await this.userRepository.findByEmail(email);
     if (foundUser && foundUser.email) {
       const { error, isBlocked } = this.isRequestBlocked(foundUser);
+      //  const member = await this.memberRepository.findById(foundUser._id);
```
markmssd, line 28: Curious if keeping this commented is intended 😸

pritipsingh: yup haven't implemented it for passwords. cannot find the member just by id. and there is no fn for `findById` in `memberRepository`  we could use others but will need more for it

scopsy: Could you please remove this code if it's not in use in this file? Including the imports part

scopsy:
@pritipsingh i'm not sure regarding the approach of passing this information from the client. Why not just having an env variable the backend can have to connect with the proper backend URL by it?

mirzap:
Any news on this?

scopsy:
Closing due to inactivity

PR #4715, opened 2023-10-30
Feature: azure SMS provider added

parveshsaini:
### What change does this PR introduce?

- Introduces Azure SMS provider
- Doc reference: https://learn.microsoft.com/en-us/azure/communication-services/quickstarts/sms/receive-sms
- 
### Why was this change needed?

- Closes: https://github.com/novuhq/novu/issues/4681

PR #4740, opened 2023-11-01
fix(providers): azure sms logo and identifier

michaldziuba03:

This PR fixes Azure SMS id and logo square variant (makes it 32x32 as it should be)

Because provider's id in one place is `azure-sms` but in other places this is `azureSms`. Square logo variant is called `azure-sms.svg` but provider's id in enum is `azureSms`.

But it's not even about the logo itself, but I just wonder if something like this will work at all?

@parveshsaini Have you tested the provider manually with credentials?

Related merged PR: https://github.com/novuhq/novu/pull/4715

#### Before: 

before

#### After: 

after


libs/shared/src/consts/providers/provider.enum.ts
```diff
@@ -89,7 +89,7 @@ export enum SmsProviderIdEnum {
   Bandwidth = 'bandwidth',
   MessageBird = 'messagebird',
   Simpletexting = 'simpletexting',
-  AzureSms = 'azureSms',
+  AzureSms = 'azure-sms',
```
michaldziuba03, line 92: In `@novu/azure-sms` identifier is: `azure-sms`:

https://github.com/parveshsaini/novu/blob/8bffdb1b6dfd1a7ac3d73d88dcffa954be73f6bd/providers/azure-sms/src/lib/azure-sms.provider.ts#L9C1-L12

michaldziuba03:
@BiswaViraj @Cliftonz

Cliftonz (approved):

PR #4741, opened 2023-11-01
Get workflow override use-case

djabarovgeorge:
* Channel Preferences

 
### Definition of Done
* add get workflow override usecase.
* e2e

### DIFF
--- a/apps/api/src/app/workflows/usecases/get-workflow-override/get-workflow-override.usecase.ts
+++ b/apps/api/src/app/workflows/usecases/get-workflow-override/get-workflow-override.usecase.ts
@@ -1,8 +1,8 @@
 import { Injectable, NotFoundException } from '@nestjs/common';
 
 import { TenantRepository, WorkflowOverrideRepository } from '@novu/dal';
-import { GetWorkflowOverrideCommand } from './update-workflow-override.command';
 import { GetWorkflowOverrideResponseDto } from '../../dto/get-workflow-override-response.dto';
+import { GetWorkflowOverrideCommand } from './get-workflow-override.command';
 
 @Injectable()
 export class GetWorkflowOverride {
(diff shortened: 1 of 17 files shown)


apps/api/src/app/workflows/usecases/get-workflow-override/get-workflow-override.usecase.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  async execute(command: GetWorkflowOverrideCommand): Promise<GetWorkflowOverrideResponseDto> {
+    const tenant = await this.tenantRepository.findOne({
+      _environmentId: command.environmentId,
+      identifier: command.tenantIdentifier,
+    });
+
+    if (!tenant) {
+      throw new NotFoundException(`Tenant with identifier ${command.tenantIdentifier} is not found`);
+    }
```
LetItRock, line 22: if we will have `tenantId` then we don't need this

apps/api/src/app/workflows/workflow.controller.ts
```diff
@@ -265,4 +269,27 @@ export class WorkflowController {
       })
     );
   }
+
+  @Get('/:workflowId/tenants/:tenantIdentifier/overrides')
```
LetItRock, line 273: `tenantId` instead of `identifier`

apps/api/src/app/workflows/usecases/get-workflow-override/get-workflow-override.usecase.ts
djabarovgeorge: Do you think we should switch the Param from `tenantIdentifier` to `tenantId` in the route? 
Please see https://github.com/novuhq/novu/pull/4729#discussion_r1382541366
I wonder what should be the interface of the ids in the route, should we use the `identifiers` or `_ids`

PR #4758, opened 2023-11-03
feat(dal, shared, api): Add rate limit DAL attributes

rifont:
* `@novu/shared`:
  * Adds `ApiServiceLevelTypeEnum`
  * Adds `RateLimitCategoryTypeEnum`
* `@novu/dal`:
  * Adds optional `apiServiceLevel` field to the Organization DAL
  * Adds optional `rateLimits` map field to the Environment DAL
* `@novu/api`
  * Modifies create organization use-case to specify a `apiServiceLevel: free` for all new organizations

The Organization and Environment entities require new fields to support API Request Rate Limiting.

N/A


libs/dal/src/repositories/organization/organization.entity.ts
```diff
@@ -1,10 +1,15 @@
...
+
 export class OrganizationEntity {
   _id: string;
 
   name: string;
 
   logo?: string;
 
+  // TODO: NV-3067 - Remove optional once all organizations have a service level
+  apiServiceLevel?: ApiServiceLevelTypeEnum;
```
scopsy, line 11: @rifont should we exclude this entity from returning using the Exclude decorator here? So it won't be returned by default when querying for his org? To me it's more of an internal property. Wdyt?

rifont: I agree that `apiServiceLevel` should be excluded from the API response for `/organizations` due to it being an internal property. Though, I'm not convinced it should be excluded at the database layer - this creates extra work in the use-case layer, complicates typings, and makes debugging for missing properties less transparent and difficult.

I experimented with the [`ApiHideProperty`](https://github.com/nestjs/swagger/blob/master/lib/decorators/api-hide-property.decorator.ts) decorator from @nestjs/swagger, however, this only removes the attribute from the swagger doc.

Longer term, we probably want a way to exclude certain properties from an API response, I see 2 solutions here:
* Implicitly exclude non-decorated properties by design. Properties intended to be exposed must be decorated on a response DTO. This is a more failsafe option.
* Explicitly exclude decorated properties. This is more intentional but requires discipline to ensure properties are decorated correctly.

I'll go with the DTO layer exclude approach in absence of the API layer exclusion for now.

rifont: @scopsy just thinking on this some more. On the other hand, it could be useful to customers building platforms on top of Novu to have access to the `apiServiceLevel`, and the `apiRateLimits` property, to display these in Admin dashboards to aid in debugging/supporting the service they are providing to their own customers via their platform.

This B2B2B2C mindset for Novu as a notification infrastructure will provide the best flexibility and transparency for customers to pick and choose how they provide access to the underlying Novu platform resources with features such as rate limiting.

rifont: Given that the Swagger doc should be seen as the source of truth for API integrations, we can technically merge and proceed with this implementation and modify the behavior of this property later without being a breaking change.

I'll merge this to keep proceedings with the rate-limiting implementation, but let's discuss this further.

PR #4761, opened 2023-11-05
feat(web): Enable to copy on activity item feed - [NV-3056]

### LINKED ISSUE #4757: [NV-3056] Enable to copy transactionId from the activity feed
### What?

The user is unable to copy transactionId from the activity feed. In the activity feed page, we show notification items with `transactionId` and `suscriberId`.

### Why? (Context)

When a user shares any issue with us, we need to check transactionId. But It is not an easy task from user's point of view to copy transactionId. When the user clicks on the item card. A modal is opened with job logs. So the only way is to copy from the developer console.
In my opinion, we can add a transaction ID copy option in the modal like this 

### Definition of Done

* Normal state: display Subscriber ID and Transaction ID without control "Copy"
* When a user hovers over a card in the activity feed the control "Copy" appears on the right side of the transaction ID and a Subscriber ID
* By clicking on the control "CopY" a user copies the Subscriber ID and Transaction ID to the clipboard

[Hover state - Crafting updates (Figma)](https://www.figma.com/file/mxViLXxtC0PnB0ehXDcafw/Crafting-updates?type=design&node-id=1561%3A7947&mode=dev&t=QfQgdk3zFB0BnfjC-1)

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-3056](https://linear.app/novu/issue/NV-3056/enable-to-copy-transactionid-from-the-activity-feed)</sub>

tquocanvn:

- Add copy functionality to transaction and subscriber ID of activity item in activity feed.

Closes #4757

https://github.com/novuhq/novu/assets/89153099/91739695-e1e3-4d40-a105-5498e70d0f82


jainpawan21:
@scopsy can we please review this PR?
This feature will be helpful in new release 0.22.0

scopsy (approved):

PR #4773, opened 2023-11-06
Stacked PR's of - Workflow override

djabarovgeorge:


libs/dal/src/repositories/workflow-override/workflow-override.schema.ts
```diff
@@ -0,0 +1,84 @@
...
+
+workflowOverrideSchema.plugin(mongooseDelete, { deletedAt: true, deletedBy: true, overrideMethods: 'all' });
+
+workflowOverrideSchema.index(
+  {
+    _tenantId: 1,
+    _workflowId: 1,
+    _environmentId: 1,
+  },
+  { unique: true }
```
djabarovgeorge, line 78: Create it as unique so we could not create 2 overrides with the same tenant workflow.

apps/api/src/app/workflow-overrides/workflow-overrides.controller.ts
```diff
@@ -0,0 +1,230 @@
...
+        environmentId: user.environmentId,
+        userId: user._id,
+        active: body.active,
+        preferenceSettings: body.preferenceSettings,
+        overrideId: overrideId,
+      })
+    );
+  }
+
+  @Put('/workflows/:workflowId/tenants/:tenantIdentifier')
```
djabarovgeorge, line 113: I am still not sure if we should use here tenantIdentifier or _tenantId.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -46,7 +54,38 @@ export class ParseEventRequest {
...
+        _tenantId: tenant._id,
+      });
+    }
+
+    const inactiveWorkflow = !workflowOverride && !template.active;
+    const inactiveWorkflowOverride = workflowOverride && !workflowOverride.active;
+
+    if (inactiveWorkflowOverride || inactiveWorkflow) {
+      const message = workflowOverride ? 'Workflow is not active by workflow override' : 'Workflow is not active';
+      Logger.log(message, LOG_CONTEXT);
```
djabarovgeorge, line 87: I wonder if we should log more here because at the moment we do not and if we would want to debug something we would be missing the information as to why the trigger failed.

LetItRock (approved):
looks good! 🚀

PR #4785, opened 2023-11-08
refactor(providers): migrate from sparkpost sdk to axios

### LINKED ISSUE #4588: [NV-3039] Migrate from sparkpost SDK to direct API calls with Axios
## What?
In the `@novu/sparkpost` package we would like to migrate from using the `sparkpost` SDK to have a code that directly calls to the provider APIs. Check the documentation about the API here: https://developers.sparkpost.com/api/

## Why? (Context)
Due to an un-maintained SDK, we should move from the official SDK' to working directly with the provider's API spec.

## Definition of Done
- the SDK package has been removed from the package.json file
- the current provider functionality works with the calls to their APIs using Axios
  - attachments

michaldziuba03:

Migrate from non mantained SparkPost SDK to direct API calls with axios.
This PR is tested manually with real credentials against real API.

screenshot

> Email sent by provider with attachment

Closes #4588

- SparkPost API reference: https://developers.sparkpost.com/api/transmissions/

### DIFF
--- a/providers/sparkpost/src/lib/sparkpost.error.ts
+++ b/providers/sparkpost/src/lib/sparkpost.error.ts
@@ -1,0 +1,16 @@
+export interface ISparkPostErrorResponse {
+  errors: Array<{
+    description: string;
+    code: string;
+    message: string;
+  }>;
+}
+
+export class SparkPostError extends Error implements ISparkPostErrorResponse {
+  readonly errors: ISparkPostErrorResponse['errors'];
+
+  constructor(response: ISparkPostErrorResponse, readonly statusCode: number) {
+    super();
+    this.errors = response.errors;
+  }
+}
--- a/providers/sparkpost/src/lib/sparkpost.provider.ts
+++ b/providers/sparkpost/src/lib/sparkpost.provider.ts
@@ -6,13 +6,22 @@ import {
   ICheckIntegrationResponse,
   CheckIntegrationResponseEnum,
 } from '@novu/stateless';
+import axios, { AxiosError } from 'axios';
 import { randomUUID } from 'crypto';
-import SparkPost from 'sparkpost';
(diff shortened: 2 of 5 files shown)


providers/sparkpost/src/lib/sparkpost.error.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export class SparkPostError extends Error implements ISparkPostErrorResponse {
+  readonly errors: ISparkPostErrorResponse['errors'];
+
+  constructor(response: ISparkPostErrorResponse, readonly statusCode: number) {
+    super();
+    this.errors = response.errors;
+  }
+}
```
michaldziuba03, line 16: Why I try to create SparkPostError instead just throwing axios exception?

Because AxiosError is circular object (refers to itself in some field) and worker throws error when trying to stringify object like that.

I want to make sure that user will get helpful execution details.

providers/sparkpost/src/lib/sparkpost.provider.ts
```diff
@@ -59,12 +65,29 @@ export class SparkPostEmailProvider implements IEmailProvider {
...
+          baseURL: this.endpoint,
+        }
+      );
+
+      return {
+        id: sent.data.results.id,
+        date: new Date().toISOString(),
+      };
+    } catch (err) {
+      this.createSparkPostError(err);
```
michaldziuba03, line 88: "factory" for SparkPostError

michaldziuba03:
@scopsy

providers/sparkpost/src/lib/sparkpost.error.ts
scopsy: Oh really nice catch! I was wondering what was causing those errors with some providers

scopsy (approved):

PR #4786, opened 2023-11-08
refactor: use execution queue instead of writing directly to execution detail collection

mahendraHegde:

- [x] Refractor the exetution detail flow to use queue in between for better performance
- [x] Adds execution log queue and worker, where usecases push the execution logs to the queue instead of directly writing to DB, worker will pickup them and writes to DB.
- [x] add tests 
- [x] add manual test results

closes ...

#### workflow

Screenshot 2023-11-10 at 7 49 32 PM

#### Execution details
Screenshot 2023-11-10 at 7 51 12 PM

Screenshot 2023-11-10 at 7 51 25 PM

Screenshot 2023-11-10 at 7 51 33 PM

Screenshot 2023-11-10 at 7 51 42 PM

#### Notifications
Screenshot 2023-11-10 at 7 53 52 PM

Screenshot 2023-11-10 at 7 50 37 PM

### DIFF
--- a/apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
@@ -30,7 +30,6 @@ import {
 } from '@novu/dal';
 import {
   DetailEnum,
-  CreateExecutionDetails,
   CreateExecutionDetailsCommand,
   buildSubscriberKey,
   CachedEntity,
@@ -38,6 +37,7 @@ import {
   Filter,
   FilterProcessingDetails,
   IFilterVariables,
+  ExecutionLogQueueService,
 } from '@novu/application-generic';
 import { EmailEventStatusEnum } from '@novu/stateless';
 
@@ -60,7 +60,7 @@ const differenceIn = (currentDate: Date, lastDate: Date, timeOperator: TimeOpera
 export class MessageMatcher extends Filter {
   constructor(
     private subscriberRepository: SubscriberRepository,
-    private createExecutionDetails: CreateExecutionDetails,
+    private executionLogQueueService: ExecutionLogQueueService,
     private environmentRepository: EnvironmentRepository,
     private executionDetailsRepository: ExecutionDetailsRepository,
     private messageRepository: MessageRepository,
@@ -101,8 +101,11 @@ export class MessageMatcher extends Filter {
         if (singleRule) {
           const result = await this.processFilter(variables, children[0], command, filterProcessingDetails);
           if (!prefiltering) {
-            await this.createExecutionDetails.execute(
+            const metadata = CreateExecutionDetailsCommand.getExecutionLogMetadata();
+            await this.executionLogQueueService.add(
+              metadata._id,
               CreateExecutionDetailsCommand.create({
+                ...metadata,
                 ...CreateExecutionDetailsCommand.getDetailsFromJob(command.job),
                 detail: DetailEnum.PROCESSING_STEP_FILTER,
                 source: ExecutionDetailsSourceEnum.INTERNAL,
@@ -110,7 +113,8 @@ export class MessageMatcher extends Filter {
                 isTest: false,
                 isRetry: false,
                 raw: filterProcessingDetails.toString(),
-              })
+              }),
+              command.organizationId
             );
           }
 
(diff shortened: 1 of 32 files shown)


mahendraHegde:
@Cliftonz can you please take a look.

Cliftonz:
@mahendraHegde  This looks great.

@p-fernandez  Can I get your glance at this? I want to make sure we are not missing anything.

Cliftonz:
@mahendraHegde  After dicussing this with the team, one of the things we need to look out for is unencrypted PII  data in the execution logs.
The simplest way to get around this is to purely encrypt the raw notification while it is in the queue and decrypt when writing to the database.
Here is the encryption usecase.
https://github.com/novuhq/novu/blob/2ff7977b363a9dfb7eeb38900f70bbeb76958c19/packages/application-generic/src/encryption/cipher.ts#L7

mahendraHegde:
> @mahendraHegde After dicussing this with the team, one of the things we need to look out for is unencrypted PII data in the execution logs. The simplest way to get around this is to purely encrypt the raw notification while it is in the queue and decrypt when writing to the database. Here is the encryption usecase.
> 
> https://github.com/novuhq/novu/blob/2ff7977b363a9dfb7eeb38900f70bbeb76958c19/packages/application-generic/src/encryption/cipher.ts#L7

okay, but today we dont encrypt the job data that we push to the queue and even data in mongo is not encrypted, so it is a separate issue that we need to tackle in general or anything specific you see for the execution log?

Cliftonz:
> > @mahendraHegde After dicussing this with the team, one of the things we need to look out for is unencrypted PII data in the execution logs. The simplest way to get around this is to purely encrypt the raw notification while it is in the queue and decrypt when writing to the database. Here is the encryption usecase.
> > https://github.com/novuhq/novu/blob/2ff7977b363a9dfb7eeb38900f70bbeb76958c19/packages/application-generic/src/encryption/cipher.ts#L7
> 
> okay, but today we dont encrypt the job data that we push to the queue and even data in mongo is not encrypted, so it is a separate issue that we need to tackle in general or anything specific you see for the execution log?

That is a very good point, I think I need to take this back and look further into it.
I think what you have done is great, so I will have the team review and go from there.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -322,7 +337,8 @@ export class SendMessageChat extends SendMessageBase {
         content,
       });
 
-      await this.createExecutionDetails.execute(
+      await this.executionLogQueueService.add(
+        command.jobId,
```
scopsy, line 341: This would need to be a unique value for later easier debugging, could you just pass here a random value?

apps/worker/src/app/workflow/usecases/message-matcher/message-matcher.usecase.ts
```diff
@@ -101,7 +101,8 @@ export class MessageMatcher extends Filter {
         if (singleRule) {
           const result = await this.processFilter(variables, children[0], command, filterProcessingDetails);
           if (!prefiltering) {
-            await this.createExecutionDetails.execute(
+            await this.executionLogQueueService.add(
```
scopsy, line 104: I believe we should consider having the`createdAt` timestamp generated here, instead of being generated by mongoose at write time. So the user can better see the timed history of an execution detail. 

Wdyt?

mahendraHegde: i think thats a good suggestion, eventhough it take milliseconds to write in most of the scenarios, in some cases queue might be busy and the write might be delayed. I will make that change

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
mahendraHegde: okay. may be i can generate and pass the mongo _id here, which could help in relating the data with event, but do you think would be benefitial? i'm not sure what kind of debugging you are talking here, because if its just for failures then anyway we wont have the data stored it will be useless.

mahendraHegde:
@Cliftonz  @scopsy review comments addressed.

rifont (approved):
LGTM. Great work @mahendraHegde 🙌

PR #4820, opened 2023-11-10
chore: update npm badges in readme

michaldziuba03:

Changes package in npm badges from `novu` to `@novu/node`.
It also fixes some weird line with mixed markdown syntax and HTML.

In my opinion we should show `@novu/node` package in readme badges. Why? Because this is package that developers are going to use and it also has 50k+ downloads/month, so it will show better numbers.

HTML and markdown shouldn't be mixed in single line:

> This is how README looks like in official GitHub Mobile app


rifont:
Great improvement, thanks @michaldziuba03 ! LGTM

rifont (approved):

PR #4824, opened 2023-11-10
feat(infra): decommission old instance of redis cluster

p-fernandez:

Removes the code related to the old instance of Redis Cluster as it is not needed anymore.

All data has been successfully consumed. So we are just spending money on a useless instance.


Cliftonz (approved):

Cliftonz:
@p-fernandez When ever the test are ready, let me know and I will start the stress tests and verification

PR #4826, opened 2023-11-11
feat(node): add methods for organization apis

### LINKED ISSUE #4775: 🚀 Feature: add nodejs sdk methods for organization apis
### 🔖 Feature description

We recently made organizations apis public and accessible using API keys and nodejs methods are missing for these apis

https://api.novu.co/api#/Organizations/OrganizationController_createOrganization

So that these apis can be accessed using nodejs SDK methods

By adding organization methods in sdk

### 🔄️ Additional Information

1. Crete required DTOs and interfaces
2. Add required tests
3. Document in readme.md

Take subscriber methods as reference
https://github.com/novuhq/novu/blob/next/packages/node/src/lib/subscribers/subscribers.ts

Take documentation for subscriber as reference
https://github.com/novuhq/novu/blob/next/packages/node/README.md#subscribers

None

michaldziuba03:

This PR introduces new SDK methods for organization API.

Work in progress

Closes #4775

### DIFF
--- a/packages/node/src/lib/organizations/organizations.interface.ts
+++ b/packages/node/src/lib/organizations/organizations.interface.ts
@@ -1,0 +1,34 @@
+export interface IOrganizations {
+  list();
+  create(payload: IOrganizationCreatePayload);
+  rename(payload: IOrganizationRenamePayload);
+  getCurrent();
+  removeMember(memberId: string);
+  updateMemberRole(
+    memberId: string,
+    payload: IOrganizationUpdateMemberRolePayload
+  );
+  getMembers();
+  updateBranding(payload: IOrganizationBrandingPayload);
+}
+
+export interface IOrganizationCreatePayload {
+  name: string;
+  logo?: string;
+}
+
+export interface IOrganizationRenamePayload {
+  name: string;
+}
+
+export interface IOrganizationUpdateMemberRolePayload {
+  role: string;
+}
+
+export interface IOrganizationBrandingPayload {
+  logo: string;
+  color: string;
+  fontColor?: string;
+  contentBackground?: string;
+  fontFamily: string;
+}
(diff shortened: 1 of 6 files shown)


packages/node/src/lib/organizations/organizations.interface.ts
```diff
@@ -0,0 +1,34 @@
+export interface IOrganizations {
+  list();
+  create(payload: IOrganizationCreatePayload);
```
michaldziuba03, line 3: Why these two methods are public? For me it doesn't make sense - API KEY is attached to single organization, so why we want to allow to cross that boundary?

Whose organizations will be returned? Organizations of the owner?

I'm just wondering - but I implemented that anyway.

packages/node/src/lib/organizations/organizations.interface.ts
```diff
@@ -0,0 +1,34 @@
...
+  rename(payload: IOrganizationRenamePayload);
+  getCurrent();
+  removeMember(memberId: string);
+  updateMemberRole(
+    memberId: string,
+    payload: IOrganizationUpdateMemberRolePayload
+  );
```
michaldziuba03, line 10: These two methods also may cause problems when full Role-based access control will be implemented?

I can imagine member that may bypass role-based access control by using API KEY to remove other members or update their role - they will do it as "organization owner" (assuming that member would be able to see api key).

michaldziuba03:
@jainpawan21 @scopsy

michaldziuba03:
@jainpawan21 @scopsy

jainpawan21:
Thanks for your contribution @michaldziuba03 
Your concern is valid with create organization and update member role apis.
We can revisit them with RBAC feature

jainpawan21 (approved):

PR #4835, opened 2023-11-13
NV-2610 - [WF editor] Provide ability to rearrange steps

gitstart:
This PR enhances the user experience in the workflow area, allowing users to easily update the position of individual steps. Previously, users had to remove all steps below a wrongly placed step to reposition it. With this update, users can now directly drag and move a step to a different position, making the process more intuitive and efficient.

The following improvements were implemented to achieve a seamless repositioning experience:

- When a user starts dragging a step, it is detached from its parent and child nodes, ensuring that only the selected step is moved without affecting its descendants.
- To maintain clarity and prevent confusion, the descendants of the repositioned step remain at their original positions during the drag operation.
- The step position is now correctly updated in the form state, reflecting the changes made by the user during the repositioning.
- A special styling is applied to the step being repositioned, providing a visual cue that it can be placed in a new position. 

To allow users to change the position of steps without having to remove all steps down to the wrongly placed step.

### Other information

Closes https://github.com/novuhq/novu/issues/3854

Loom Video https://www.loom.com/share/ba3167669ba545bd996181466f22cd67

---


gitstart:
@LetItRock this PR is ready for review

PR #4836, opened 2023-11-13
NV-2057 - 🚀 Feature: Get topics of subscriber

### LINKED ISSUE #3142: [NV-2057] 🚀 Feature: Get topics of subscriber

gitstart:
This feature enables user to filter topics by subscriber ID.

closes https://github.com/novuhq/novu/issues/3142

### Other information

---

### DIFF
--- a/apps/api/src/app/topics/e2e/filter-topics.e2e.ts
+++ b/apps/api/src/app/topics/e2e/filter-topics.e2e.ts
@@ -114,6 +114,43 @@ describe('Filter topics - /topics (GET)', async () => {
     expect(topic.subscribers).to.have.members([firstSubscriber.subscriberId, secondSubscriber.subscriberId]);
   });
 
+  it('should retrieve the topic filtered by the query param subscriber ID for the user', async () => {
+    const url = `${BASE_PATH}?subscriberId=${firstSubscriber.subscriberId}`;
+    const response = await session.testAgent.get(url);
+
+    expect(response.statusCode).to.eql(200);
+
+    const { data, totalCount, page, pageSize } = response.body;
+    const [topic] = data;
+
+    expect(data.length).to.eql(1);
+    expect(totalCount).to.eql(1);
+    expect(page).to.eql(0);
+    expect(pageSize).to.eql(10);
+    expect(topic._environmentId).to.eql(session.environment._id);
+    expect(topic._organizationId).to.eql(session.organization._id);
+    expect(topic.subscribers).to.have.members([firstSubscriber.subscriberId, secondSubscriber.subscriberId]);
+  });
+
+  it('should retrieve the topic filtered by multiple query param', async () => {
+    const topicKey = 'topic-key-2';
+    const url = `${BASE_PATH}?subscriberId=${firstSubscriber.subscriberId}&key=${topicKey}&pageSize=2`;
+    const response = await session.testAgent.get(url);
+
+    expect(response.statusCode).to.eql(200);
+
+    const { data, totalCount, page, pageSize } = response.body;
+    const [topic] = data;
+
+    expect(data.length).to.eql(1);
+    expect(totalCount).to.eql(1);
+    expect(page).to.eql(0);
+    expect(pageSize).to.eql(2);
+    expect(topic._environmentId).to.eql(session.environment._id);
+    expect(topic._organizationId).to.eql(session.organization._id);
+    expect(topic.subscribers).to.have.members([firstSubscriber.subscriberId, secondSubscriber.subscriberId]);
+  });
+
   it('should retrieve an empty response if filtering by a key that is not in the database for the user', async () => {
     const topicKey = 'topic-key-not-existing';
     const url = `${BASE_PATH}?key=${topicKey}`;
--- a/apps/api/src/app/topics/topics.controller.ts
+++ b/apps/api/src/app/topics/topics.controller.ts
@@ -51,11 +51,14 @@ import {
   RemoveSubscribersUseCase,
   RenameTopicCommand,
   RenameTopicUseCase,
+  GetSubscriberTopicsUseCase,
+  GetSubscriberTopicsCommand,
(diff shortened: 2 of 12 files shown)


gitstart:
@scopsy @p-fernandez this PR is ready for review

apps/api/src/app/topics/topics.controller.ts
```diff
@@ -204,13 +214,52 @@ export class TopicsController {
...
         key: query?.key,
+        subscriberId: query?.subscriberId,
         organizationId: user.organizationId,
         page: query?.page,
         pageSize: query?.pageSize,
       })
     );
   }
 
+  @Get('/subscribers/:subscriberId')
```
djabarovgeorge, line 225: Is this new endpoint should be part of this PR? i did not find any mentions of this in the tagged issue.

apps/api/src/app/topics/e2e/filter-topics.e2e.ts
```diff
@@ -114,6 +114,43 @@ describe('Filter topics - /topics (GET)', async () => {
...
+    const { data, totalCount, page, pageSize } = response.body;
+    const [topic] = data;
+
+    expect(data.length).to.eql(1);
+    expect(totalCount).to.eql(1);
+    expect(page).to.eql(0);
+    expect(pageSize).to.eql(10);
+    expect(topic._environmentId).to.eql(session.environment._id);
+    expect(topic._organizationId).to.eql(session.organization._id);
+    expect(topic.subscribers).to.have.members([firstSubscriber.subscriberId, secondSubscriber.subscriberId]);
```
djabarovgeorge, line 132: Should we check if the subscriber belongs to the predefined in the `describe` topic?

apps/api/src/app/topics/e2e/filter-topics.e2e.ts
```diff
@@ -114,6 +114,43 @@ describe('Filter topics - /topics (GET)', async () => {
...
+    const { data, totalCount, page, pageSize } = response.body;
+    const [topic] = data;
+
+    expect(data.length).to.eql(1);
+    expect(totalCount).to.eql(1);
+    expect(page).to.eql(0);
+    expect(pageSize).to.eql(2);
+    expect(topic._environmentId).to.eql(session.environment._id);
+    expect(topic._organizationId).to.eql(session.organization._id);
+    expect(topic.subscribers).to.have.members([firstSubscriber.subscriberId, secondSubscriber.subscriberId]);
```
djabarovgeorge, line 151: same here

libs/dal/src/repositories/topic/topic.repository.ts
```diff
@@ -55,30 +55,38 @@ export class TopicRepository extends BaseRepository<TopicDBModel, TopicEntity, E
...
+      $match: {
+        $expr: {
+          $in: [subscriberId, '$topicSubscribers.externalSubscriberId'],
+        },
       },
+    };
+    const commonAggQuery = [{ $match: parsedQuery }, lookup, ...(subscriberFilter ? [subscriberFilter] : [])];
+
+    const [data, totalCount] = await Promise.all([
+      this.aggregate([
```
djabarovgeorge, line 76: Although i like the optimization here with one mongo query, maybe we can keep the previous implementation but add `Promise.alll` to the two queries? my main concerns are that not filterTopics is responsible for two actions (get data and get totalCount) and the additional complexity.
We most definitely can back to it once we notice that we have a high load in here and we need to optimize more. wdyt?

djabarovgeorge (commented):
Amazing job, i left a couple of comments for your review.

apps/api/src/app/topics/topics.controller.ts
gitstart: @djabarovgeorge  The new endpoint was proposed by @p-fernandez on the previously raised PR, where we even had to revert another implementation for this 
see comment for reference https://github.com/novuhq/novu/pull/3551#discussion_r1227969854

The idea was proposed so we could have a simplified implementation rather than compounding two distinct implementations into one
cc: @jainpawan21

apps/api/src/app/topics/e2e/filter-topics.e2e.ts
gitstart: We needed to check for that, to confirm that all listed subscribers are returned for a topic so that the user can have a complete list of subscribers associated with a topic

apps/api/src/app/topics/e2e/filter-topics.e2e.ts
gitstart: Same response as [here](https://github.com/novuhq/novu/pull/4836#discussion_r1407706533)

brucek:
I'd like to weigh in here that this functionality would be very helpful for my project too in case @gitstart you need any moral support 🤞🙂

estubmo:
Any update on this?

scopsy:
@brucek @estubmo the team will be picking this issue soon to address 🙏

PR #4839, opened 2023-11-13
fix(ci): docker build test.yml for forks

scopsy:

Loads the built docker to the local machine.

Community tests been failing due to latest tag not found. See docker docs: https://docs.docker.com/engine/reference/commandline/buildx_build/#load


p-fernandez (approved):
🌟

PR #4840, opened 2023-11-13
refactor(providers): Migrate from onesignal-node SDK to the official @onesignal/node-onesignal

### LINKED ISSUE #4635: [NV-3053] Migrate from onesignal-node SDK to the official @onesignal/node-onesignal
## What?
In the @novu/one-signal package we would like to migrate from using the onesignal-node SDK to the official @onesignal/node-onesignal SDK. Check the documentation about the it here: https://github.com/OneSignal/onesignal-node-api#readme

## Why? (Context)
Due to vulnerabilities and un-maintained SDK, we should move from using unofficial onesignal-node SDK.

## Definition of Done
- the onesignal-node SDK package has been replaced with @onesignal/node-onesignal
- the current provider functionality works with the calls to the new @onesignal/node-onesignal SDK:
  - attachments

Dhoni77:

- Remove onesignal-node sdk
- Use `@onesignal/node-onesignal` sdk
- Updated unit tests

Closes #4635

Due to vulnerabilities and un-maintained SDK, we should move from using unofficial onesignal-node SDK.


packages/application-generic/src/factories/push/handlers/one-signal.handler.ts
```diff
@@ -8,13 +8,18 @@ export class OneSignalHandler extends BasePushHandler {
...
+      !credentials.appKey
+    ) {
       throw Error('Config is not valid for OneSignal');
     }
 
     this.provider = new OneSignalPushProvider({
       appId: credentials.applicationId,
-      apiKey: credentials.apiKey,
+      appKey: credentials.appKey,
+      userKey: credentials.userKey,
```
scopsy, line 22: Unfortunately this change is causing a breakin change for all of one signal users, is there some mapping between the old apiKey to the new appKey? is that the same value?

scopsy (changes requested):
We need to make sure that we are not providing a breaking change for old configuration users who used the apiKey and applicationId approach.

scopsy:
@Dhoni77 another approach we can take is using the rest API directly which works with API key and AppId like previously: https://documentation.onesignal.com/reference/create-notification we can use axios for the API call, wdyt?

providers/one-signal/package.json
```diff
@@ -31,7 +31,7 @@
   },
   "dependencies": {
     "@novu/stateless": "^0.21.0",
-    "onesignal-node": "^3.4.0"
+    "axios": "1.3.4"
```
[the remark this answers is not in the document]
Dhoni77: Ok👍

providers/one-signal/src/lib/one-signal.provider.spec.ts
```diff
@@ -1,56 +1,50 @@
 import { OneSignalPushProvider } from './one-signal.provider';
+import axios from 'axios';
+import { IPushOptions } from 'packages/stateless/build/main';
```
scopsy, line 3:
```suggestion
import { IPushOptions } from '@novu/stateless';
```

scopsy (approved):

PR #4843, opened 2023-11-14
Ensure that our queue services get validated object data

djabarovgeorge:

At the moment we do not validate the data we pass queue services for example SubscriberProcessQueueService.

Suggestions:
What we could do is create a middleware function 'add' in the child class, for example, SubscriberProcessQueueService with command and after then pass it to the parent 'add'.

We could provide an invalid object to the queue services by mistake.

### Other information and DOD's (Screenshots)

The remaining things that still need to be done:
* Refactor the job data object to be defined, at the moment we allow to pass `undefined` data. IMO it was made in order to create metric cron jobs that do not pass any data, however, this option adds complexity to the code and could lead to a failure point where someone won't pass the data object. Instead, we could add minimal data on the metric identifier.
* Make a discovery in order to make the base queue service move generic by getting the type from its child classes, this way we won't need to duplicate the methods in the services. in order to do so we will need to remove the mapping of the bull mq in the queue service and move it to the bull mq service.
* Refactor so that all of the workers will execute the creation of the Command. was not sure if it should be done in this PR or not, because it could be a breaking change.
* Align all of DTO's params to use align internal conventions, meaning if it is novu internal id it should have a prefix of `_` at the moment we sometimes use it and sometimes not. for example userId _userId. This will be a breaking change that will require backward combability support so I'd prefer to do it in separate dedicated PR.


apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -104,18 +104,12 @@ export class ParseEventRequest {
...
-    let jobData = {
+    const jobData = {
       ...command,
-      actor: command.actor,
       transactionId,
     } as ParseEventRequestCommand;
 
-    if ((command as ParseEventRequestMulticastCommand).to?.length > 0) {
-      jobData = jobData as ParseEventRequestMulticastCommand;
-      jobData.to = (command as ParseEventRequestMulticastCommand).to;
```
djabarovgeorge, line 115: Done by me this month, it is not needed as it will be destructured above in `...command,`.
More than that i regret that i created two types of one usecase command, Apologies in advance, i will revisit it in order to think of a better way.

apps/api/src/app/inbound-parse/services/inbound-parse.queue.service.ts
```diff
@@ -34,7 +33,7 @@ export class InboundParseQueueService {
   }
 
   public getWorkerProcessor() {
-    return async ({ data }: { data: InboundEmailParseCommand }) => {
+    return async ({ data }: { data: IInboundParseDataDto }) => {
       Logger.verbose({ data }, 'Processing the inbound parsed email', LOG_CONTEXT);
       await this.emailParseUsecase.execute(InboundEmailParseCommand.create({ ...data }));
```
djabarovgeorge, line 38: I wonder if we should do the execution of create in this PR or not. it could be a breaking change.

apps/api/src/app/inbound-parse/usecases/inbound-email-parse/inbound-email-parse.command.ts
```diff
@@ -66,70 +75,3 @@ export class InboundEmailParseCommand extends BaseCommand {
...
-
-export interface IEnvelopeFrom {
-  address: string;
-  args: boolean;
-}
-
-export interface IEnvelopeTo {
-  address: string;
-  args: boolean;
-}
```
djabarovgeorge, line 135: Extracted to application-generic so we could reuse it.

apps/api/src/app/widgets/usecases/mark-all-messages-as/mark-all-messages-as.usecase.ts
```diff
@@ -56,24 +56,15 @@ export class MarkAllMessagesAs {
...
       command.markAs === MarkMessagesAsEnum.READ || command.markAs === MarkMessagesAsEnum.UNREAD;
 
-    const countQuery = isUnreadCountChanged ? { read: false } : { seen: false };
-
-    const count = await this.messageRepository.getCount(
-      command.environmentId,
-      subscriber._id,
-      ChannelTypeEnum.IN_APP,
-      countQuery
-    );
```
djabarovgeorge, line 66: Was not used.

apps/inbound-mail/src/server/inbound-mail.service.spec.ts
```diff
@@ -65,7 +66,11 @@ describe('Inbound Mail Service', () => {
         _organizationId,
         _userId,
       };
-      await inboundMailService.inboundParseQueueService.add(jobId, jobData, _organizationId);
+      await inboundMailService.inboundParseQueueService.add({
+        name: jobId,
+        data: jobData as unknown as IInboundParseDataDto,
```
djabarovgeorge, line 71: Make for the test, IInboundParseDataDto contained a lot of attributes that needed to be added. for now, added casting as those attributes were not tested.

apps/worker/src/app/workflow/services/standard.worker.ts
```diff
@@ -45,7 +41,7 @@ export class StandardWorker extends StandardWorkerService implements INovuWorker
 
     this.initWorker(this.getWorkerProcessor(), this.getWorkerOptions());
 
-    this.worker.on('failed', async (job: Job<IJobData, void, string>, error: Error): Promise<void> => {
+    this.worker.on('failed', async (job: Job<IStandardDataDto, void, string>, error: Error): Promise<void> => {
```
djabarovgeorge, line 44: we will need either merge IJobData and IStandardDataDto, or connect them, as they share the same interface.

apps/worker/src/app/workflow/services/standard.worker.ts
```diff
@@ -59,22 +55,26 @@ export class StandardWorker extends StandardWorkerService implements INovuWorker
...
-    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = job;
+    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = data;
 
     if (!environmentId || !jobId || !organizationId || !userId) {
-      const message = job.payload.message;
+      const message = data.payload?.message;
+
+      if (!message) {
+        throw new Error('Job data is missing required fields' + JSON.stringify(data));
+      }
```
djabarovgeorge, line 71: Should not happen. if it does we need to throw it as no message exists.

apps/worker/src/app/workflow/services/standard.worker.ts
```diff
@@ -59,22 +55,26 @@ export class StandardWorker extends StandardWorkerService implements INovuWorker
...
     organizationId: string;
     jobId: string;
     userId: string;
   } {
-    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = job;
+    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = data;
 
     if (!environmentId || !jobId || !organizationId || !userId) {
-      const message = job.payload.message;
+      const message = data.payload?.message;
```
djabarovgeorge, line 67: This object is structure `data.payload?.message` deprecated, i could not locate any place in the code that passes this data.

apps/worker/src/app/workflow/services/workflow.worker.ts
```diff
@@ -44,7 +45,7 @@ export class WorkflowWorker extends WorkflowWorkerService implements INovuWorker
 
             storage.run(new Store(PinoLogger.root), () => {
               _this.triggerEventUsecase
-                .execute(data)
+                .execute(data as TriggerEventCommand)
```
djabarovgeorge, line 48: We should create a command here, but before and solve this issue of complex command interface https://github.com/novuhq/novu/pull/4843/files#r1392301028

apps/ws/src/socket/services/web-socket.worker.ts
```diff
@@ -37,14 +39,16 @@ export class WebSocketWorker extends WebSocketsWorkerService implements INovuWor
           'WS Service',
           function () {
             const transaction = nr.getTransaction();
+            const { data: jobData } = job;
+            const data = jobData as IWebSocketDataDto;
```
djabarovgeorge, line 43: I am not sure what is the type of `job` above therefore i create this mapping so we could at least get the `data` object right here.

packages/application-generic/src/dtos/index.ts
djabarovgeorge, line 1: Created those DTO's of our queue-worker communication for us to use.

packages/application-generic/src/dtos/process-subscriber-job.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+  templateId: string;
+}
+
+export interface IProcessSubscriberJobDto extends IJobParams {
+  data?: IProcessSubscriberDataDto;
+}
+
+export interface IProcessSubscriberBulkJobDto extends IBulkJobParams {
+  data: IProcessSubscriberDataDto;
+}
```
djabarovgeorge, line 29: This is part of the complexity i was talking about in the issue description, due to the metrics add execution allowing undefined data objects we need to create two separate interfaces because we do not want to allow undefined in the data as much as possible e.x the bulk DTO.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -71,52 +71,60 @@ export class QueueBaseService {
...
-    });
+    await this.instance.add(
+      name,
+      jobData,
+      {
+        removeOnComplete: true,
+        removeOnFail: true,
+        ...options,
+      },
+      groupId
```
djabarovgeorge, line 94: I removed the dependency of the `add` method.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -71,52 +71,60 @@ export class QueueBaseService {
     await this.gracefulShutdown();
   }
 
-  public async addMinimalJob(
-    id: string,
-    data?: any,
-    groupId?: string,
-    options: JobsOptions = {}
-  ) {
+  public async addMinimalJob(params: IJobParams) {
```
djabarovgeorge, line 74: Refactored to get object instead of params list so we could be explicit about the params.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -71,52 +71,60 @@ export class QueueBaseService {
...
+    );
   }
 
-  public async add(
-    name: string,
-    data?: any,
-    groupId?: string,
-    options: JobsOptions = {}
-  ) {
+  public async add(params: IJobParams) {
```
djabarovgeorge, line 98: Refactored to get object instead of params list so we could be explicit about the params.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -71,52 +71,60 @@ export class QueueBaseService {
...
+  groupId?: string;
+  options?: JobsOptions;
+}
+
+export interface IBulkJobParams {
+  name: string;
+  data: any;
+  groupId?: string;
+  options?: BulkJobOptions;
+}
```
djabarovgeorge, line 130: This is part of the complexity i was talking about in the issue description, due to the metrics add execution allowing undefined data objects we need to create two separate interfaces because we do not want to allow undefined in the data as much as possible e.x the bulk DTO.

* We need to find a better place for those interfaces.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -71,52 +71,60 @@ export class QueueBaseService {
...
-      {
-        name: string;
-        data: any;
-        options: BulkJobOptions;
-        groupId?: string;
-      }
-    ]
-  ) {
+
+  public async addBulk(data: IBulkJobParams[]) {
```
djabarovgeorge, line 113: Refactored to get object instead of params list so we could be explicit about the params.

packages/application-generic/src/services/queues/standard/standard-queue.service.spec.ts
```diff
@@ -58,14 +59,17 @@ describe('Standard Queue service', () => {
       const _environmentId = 'standard-environment-id';
       const _organizationId = 'standard-organization-id';
       const _userId = 'standard-user-id';
-      const jobData = {
-        _id: jobId,
-        test: 'standard-job-data',
```
djabarovgeorge, line 63: invalid param

packages/application-generic/src/services/queues/standard/standard-queue.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public async add(data: unknown) {
+    // in order to implement we need to know what should be the data dto first
+    throw new Error('Not implemented');
+  }
+
+  public async addBulk(data: unknown[]) {
+    // in order to implement we need to know what should be the data dto first
+    throw new Error('Not implemented');
+  }
```
djabarovgeorge, line 30: We do not use those methods at the moment, therefore i can not tell what should be the DTO here, i wonder if we should keep the exception or allow any here.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -71,52 +71,60 @@ export class QueueBaseService {
...
     await this.instance.addBulk(data);
   }
 }
+
+export interface IJobParams {
+  name: string;
```
djabarovgeorge, line 119: Changed the `id` to `name` as it is the perm that is used in pull mq so i wanted to align with it. Same for IBulkJobParams.

PR #4844, opened 2023-11-14
feat(api): Add evaluate api rate limit use case

rifont:
* Adds `evaluate-api-rate-limit` use-case to determine if a request should be rate-limited for a given `environmentId`, service category, and cost category.
* Adds `evaluate-token-bucket-rate-limit` use-case to encapsulate the token bucket algorithm. The use-case uses the `@upstash/ratelimit` library to implement the token bucket rate limit algorithm.

* The `evaluate-api-rate-limit` use-case will be used in the Rate Limiting guard to determine if a request will be accepted or rejected

N/A


apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-configuration/get-api-rate-limit-configuration.usecase.ts
```diff
@@ -0,0 +1,44 @@
+import { Injectable } from '@nestjs/common';
```
rifont, line 1: We have hardcoded defaults for the rate limit configuration which can be overidden with environment variables.

packages/application-generic/src/services/cache/cache.service.ts
```diff
@@ -23,6 +23,15 @@ export interface ICacheService {
...
   cacheEnabled();
+  sadd<TData extends string | number | Buffer>(
+    key: string,
+    ...members: TData[]
+  ): Promise<number>;
+  eval<TArgs extends (string | Buffer | number)[], TData = unknown>(
+    script: string,
+    keys: string[],
+    ...args: TArgs
+  ): Promise<TData>;
```
rifont, line 34: Additional Redis operations required by the `@upstash/ratelimiting` library.

packages/application-generic/src/services/cache/cache.service.ts
```diff
@@ -219,6 +228,25 @@ export class CacheService implements ICacheService {
...
+    TArgs extends (string | Buffer | number)[],
+    TData = unknown
+  >(script: string, keys: string[], ...args: TArgs): Promise<TData> {
+    return this.client?.eval(
+      script,
+      keys.length,
+      ...keys,
+      ...args
+    ) as Promise<TData>;
+  }
```
rifont, line 249: Typesafe adaptator for a "generic" Redis `eval` method to the `ioredis` method. Notably, the `ioredis` library requires the length of keys (`keys.length`) to be passed as a parameter whereas many other libraries do not.

libs/shared/src/types/rate-limiting/index.ts
```diff
@@ -7,3 +7,26 @@ type ApiRateLimitNamespace = 'API_RATE_LIMIT';
 
 export type ApiRateLimitEnvVarFormat =
   Uppercase<`${ApiRateLimitNamespace}_${ApiServiceLevelTypeEnum}_${ApiRateLimitCategoryTypeEnum}`>;
+
+export enum ApiRateLimitConfigurationEnum {
+  BURST_ALLOWANCE = 'burstAllowance',
+  REFILL_INTERVAL = 'refillInterval',
+}
+
+export class IApiRateLimitConfiguration implements Record<ApiRateLimitConfigurationEnum, unknown> {
```
rifont, line 16: This `implements` pattern allows us to create a typing that strictly required all class attributes have keys in the `ApiRateLimitConfigurationEnum`.

apps/api/src/app/rate-limiting/usecases/evaluate-api-rate-limit/evaluate-api-rate-limit.usecase.ts
```diff
@@ -0,0 +1,98 @@
...
+import { EvaluateApiRateLimitResponse } from './evaluate-api-rate-limit.types';
+
+const LOG_CONTEXT = 'EvaluateApiRateLimit';
+
+type UpstashRedisClient = ConstructorParameters<typeof Ratelimit>[0]['redis'];
+
+@Injectable()
+export class EvaluateApiRateLimit {
+  private ephemeralCache = new Map<string, number>();
+  public readonly DEFAULT_WINDOW_DURATION = 60;
```
LetItRock, line 16: not used?

LetItRock (approved):
looks awesome! 🚀

apps/api/src/app/rate-limiting/usecases/evaluate-api-rate-limit/evaluate-api-rate-limit.usecase.ts
rifont: You're right, it wasn't. Now deleted.

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-configuration/get-api-rate-limit-configuration.usecase.ts
```diff
@@ -0,0 +1,44 @@
...
+      }
+    });
+
+    return mergedApiRateLimits;
+  }
+
+  private getEnvVarName(
+    apiRateLevelConfiguration: ApiRateLimitConfigurationEnum
+  ): ApiRateLimitConfigurationEnvVarFormat {
+    return `API_RATE_LIMIT_${apiRateLevelConfiguration.toUpperCase() as Uppercase<ApiRateLimitConfigurationEnum>}`;
```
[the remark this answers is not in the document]
rifont: It's a good point. I would prefer the `CAPITAL_CASE` approach, but that would require either:
1. a custom `camelCase` to `CAPITAL_CASE` transformer
2. the enum values to be non-`camelCased`

I think I'll go with option 2, as we can just use the enum keys as the reference.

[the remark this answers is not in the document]
rifont: I've resolved it like this, changing the `camelCase` enum values to `snake_case`:
```typescript
export enum ApiRateLimitAlgorithmEnum {
  BURST_ALLOWANCE = 'burst_allowance',
  WINDOW_DURATION = 'window_duration',
}
```

This way, the value is transformed `toUpperCase` without further changes, and gives us flexibility to introduce a custom transformer in the future - in the case we need to store these config values on a per-environment/organization basis.

LetItRock (approved):
looks perfect to me! 💎

apps/api/src/app/rate-limiting/usecases/evaluate-api-rate-limit/evaluate-api-rate-limit.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+    private evaluateTokenBucketRateLimit: EvaluateTokenBucketRateLimit
+  ) {}
+
+  @InstrumentUsecase()
+  async execute(command: EvaluateApiRateLimitCommand): Promise<EvaluateApiRateLimitResponseDto> {
+    const maxLimitPerSecond = await this.getApiRateLimitMaximum.execute({
+      apiRateLimitCategory: command.apiRateLimitCategory,
+      environmentId: command.environmentId,
+      organizationId: command.organizationId,
+    });
```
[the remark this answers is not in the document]
rifont: thanks! fixed.

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-maximum/get-api-rate-limit-maximum.usecase.ts
```diff
@@ -49,7 +49,7 @@ export class GetApiRateLimit {
 
     const { apiRateLimits } = environment;
 
-    let environmentApiRateLimits: IApiRateLimits;
+    let environmentApiRateLimits: IApiRateLimitMaximum;
```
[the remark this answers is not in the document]
rifont: Good point, I've updated to make it more clear.

PR #4849, opened 2023-11-15
Create change on group and feed creation

djabarovgeorge:

Steps to reproduce: 
Go to the workflow page
click on the Create Workflow button
choose password reset workflow (or any other) from pre-made workflows
update something if you want to update some content in that
Go to the changes page, find out changes related to this newly added workflow
Try to promote this change
You will get this error similar to the attached image

 

Its a bug.

The fix for this is to go back to workflow settings change the notification group to other and update and promote

### DIFF
--- a/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
+++ b/apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
@@ -35,9 +35,7 @@ export class CreateNotificationTemplate {
   ) {}
 
   async execute(usecaseCommand: CreateNotificationTemplateCommand) {
-    const parentChangeId: string = NotificationTemplateRepository.createObjectId();
-
-    const blueprintCommand = await this.processBlueprint(usecaseCommand, parentChangeId);
+    const blueprintCommand = await this.processBlueprint(usecaseCommand);
     const command = blueprintCommand ?? usecaseCommand;
 
     const contentService = new ContentService();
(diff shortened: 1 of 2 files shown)


LetItRock (approved):
can we please cover this with tests? thank you 🙏

apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
```diff
@@ -263,6 +278,18 @@ export class CreateNotificationTemplate {
...
       });
+
+      await this.createChange.execute(
+        CreateChangeCommand.create({
+          item: group,
+          environmentId: command.environmentId,
+          organizationId: command.organizationId,
+          userId: command.userId,
+          type: ChangeEntityTypeEnum.NOTIFICATION_GROUP,
+          parentChangeId: parentChangeId,
```
ainouzgali, line 289:
```suggestion
          parentChangeId,
```
@djabarovgeorge doesn't setting the parent here will make it into the same row? Like, it won't show as an independent change?

djabarovgeorge:
> can we please cover this with tests? thank you 🙏

Unfortunately no.
In order to create a test that will simulate the issue we need to make a request to the `dev` environment, which should be fine because we took it into account in blueprint feature discovery.
However, it is not possible to create such a test before we fix [issue](https://linear.app/novu/issue/NV-3188/incorrect-creation-of-the-blueprint-on-cross-origin) i found while fixing this issue

apps/api/src/app/workflows/usecases/create-notification-template/create-notification-template.usecase.ts
djabarovgeorge: Yes, and it is the desired behavior i think we should merge all the possible changes into one workflow change in order to reduce potential issues like we had in the past
Do you have another opinion on the matter?

ainouzgali: Its just not the current behavior we have for notification group. They are showing separately ->notification group change can be promoted as itself. But if you promote the notification template change  it promotes both changes

PR #4857, opened 2023-11-16
chore(api, shared, app-generic): Convert API rate limit FF to be LaunchDarkly compatible

rifont:
* Convert API rate limit flag from system-critical to feature-flag
* Rename `IS_REQUEST_RATE_LIMITING_ENABLED` -> `IS_API_RATE_LIMITING_ENABLED` for naming consistency with features under implementation

* The feature flag was not compatible with LaunchDarkly and could not be resolved at runtime. This change enables rate limiting to be rolled out gradually through strategies such as A/B.

N/A


apps/api/src/.env.development
```diff
@@ -25,7 +25,7 @@ REDIS_CACHE_FAMILY=
 REDIS_CACHE_KEY_PREFIX=
 REDIS_CACHE_ENABLE_AUTOPIPELINING=true
 
-IS_REQUEST_RATE_LIMITING_ENABLED=false
+IS_API_RATE_LIMITING_ENABLED=false
```
rifont, line 28: "API Rate Limiting" aligns naming with the use-cases under development. If we decide to add Web Socket rate limiting or other types, we can then toggle these features independently.

packages/application-generic/src/usecases/get-feature-flag/get-is-request-rate-limiting-enabled.use-case.ts
```diff
@@ -1,21 +1,22 @@
...
 
-import { GetSystemCriticalFlag } from './get-system-critical-flag.use-case';
+import { FeatureFlagCommand } from './get-feature-flag.command';
+import { GetFeatureFlag } from './get-feature-flag.use-case';
 
 @Injectable()
-export class GetIsRequestRateLimitingEnabled extends GetSystemCriticalFlag {
-  execute(): boolean {
-    const value = process.env.IS_REQUEST_RATE_LIMITING_ENABLED;
+export class GetIsApiRateLimitingEnabled extends GetFeatureFlag {
```
rifont, line 8: Changing `GetFeatureFlag` -> `GetSystemCriticalFlag` and modifying the `execute` method accordingly.

packages/application-generic/src/usecases/get-feature-flag/get-is-api-rate-limiting-enabled.use-case.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@nestjs/common';
```
rifont, line 1: This file was also renamed from `get-is-request-rate-limiting-enabled` -> `get-is-api-rate-limiting-enabled`

LetItRock (approved):
looks good to me 🎉

PR #4868, opened 2023-11-17
feat(infra): move provider selection to be owned by services

p-fernandez:

Moves the responsibility to choose the in-memory provider from the providers section of InMemoryProviderService to the Cache and the Workflow in-memory services.

This allows finer control and to allow each of the services to own their fallbacks and configurations. This is a previous step to allow free selection of provider by allowing multiple instances of the same provider and creating dedicated service environment variables rather than per-service dedicated environment variables like it is happening now.


packages/application-generic/src/services/cache/cache-service.spec.ts
```diff
@@ -8,11 +8,7 @@ import {
 import { CacheInMemoryProviderService } from '../in-memory-provider';
 import { MockCacheService } from './cache-service.mock';
 
-/**
- * TODO: Maybe create a Test single Redis instance to be able to run it in the
- * pipeline. Local wise they work
- */
-describe.skip('Cache Service - Redis Instance - Non Cluster Mode', () => {
+describe('Cache Service - Redis Instance - Non Cluster Mode', () => {
```
p-fernandez, line 11: The changes also allowed to this to be run properly.

packages/application-generic/src/services/in-memory-provider/cache-in-memory-provider.service.ts
```diff
@@ -36,32 +43,38 @@ export class CacheInMemoryProviderService {
...
-    if (process.env.IS_DOCKER_HOSTED) {
-      return InMemoryProviderEnum.REDIS;
+  private selectProvider(): IProviderCluster | IProviderRedis {
+    if (this.isClusterMode()) {
+      const providers = [
+        InMemoryProviderEnum.ELASTICACHE,
+        InMemoryProviderEnum.REDIS_CLUSTER,
+      ];
+
+      const selectedProvider = providers.find((provider) =>
```
p-fernandez, line 59: Checks providers in order and loads the first one that is validated.

packages/application-generic/src/services/in-memory-provider/cache-in-memory-provider.service.ts
```diff
@@ -93,10 +106,7 @@ export class CacheInMemoryProviderService {
   }
 
   public providerInUseIsInClusterMode(): boolean {
-    const providerConfigured =
-      this.inMemoryProviderService.getProvider.configured;
-
-    return this.isCluster || providerConfigured !== InMemoryProviderEnum.REDIS;
+    return this.loadedProvider.provider !== InMemoryProviderEnum.REDIS;
```
p-fernandez, line 109: For the cases that providers are misconfigured and it fell backs to other providers we can't rely on if the cluster mode has been set or not.

packages/application-generic/src/services/in-memory-provider/providers/providers.spec.ts
```diff
@@ -37,84 +34,28 @@ describe('Client and config for cluster', () => {
...
-      InMemoryProviderEnum.MEMORY_DB
-    );
+    const { getConfig } = getClusterProvider(InMemoryProviderEnum.MEMORY_DB);
     const config: IMemoryDbClusterProviderConfig = getConfig();
     expect(config.host).toEqual(memoryDbUrl);
     expect(config.port).toEqual(Number(memoryDbPort));
     expect(config.ttl).toEqual(7200);
   });
 
-  it('should return Redis Cluster config after validating Elasticache faulty URL config', () => {
```
p-fernandez, line 49: These fallback cases doesn't happen in the providers area anymore.

packages/application-generic/src/services/in-memory-provider/cache-in-memory-provider.service.ts
```diff
@@ -15,53 +22,80 @@ export class CacheInMemoryProviderService {
...
-   * if Elasticache not configured properly. That's happening in the provider
-   * mapping in the /in-memory-provider/providers/index.ts
+   * if Elasticache not configured properly. If Redis Cluster is wrong too, we will
+   * fall back to Redis single instance.
    */
-  private selectProvider(): InMemoryProviderEnum {
-    if (process.env.IS_DOCKER_HOSTED) {
-      return InMemoryProviderEnum.REDIS;
+  private selectProvider(): IProviderCluster | IProviderRedis {
+    if (this.isClusterMode()) {
```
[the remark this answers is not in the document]
p-fernandez: There is a small risk that if someone swaps the order of execution in the constructor this change could led to a buggy behaviour.

packages/application-generic/src/services/in-memory-provider/in-memory-provider.service.ts
```diff
@@ -20,44 +20,29 @@ const LOG_CONTEXT = 'InMemoryProviderService';
...
   public inMemoryProviderClient: InMemoryProviderClient;
-  public inMemoryProviderConfig: InMemoryProviderConfig;
-
   public isProviderClientReady: (string) => boolean;
 
   constructor(
-    private provider: InMemoryProviderEnum,
-    private isCluster: boolean,
-    private enableAutoPipelining?: boolean
+    private loadedProvider: IProviderCluster | IProviderRedis,
```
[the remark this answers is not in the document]
p-fernandez: This is a good point and that possibility would lead to buggy behaviour. Something to give it a thought the best way of solving it.

PR #4871, opened 2023-11-19
refactor(ws): Add newrelic metrics and refactor handle disconnect

scopsy:


apps/ws/src/socket/ws.gateway.ts
```diff
@@ -1,41 +1,113 @@
...
       return this.disconnect(connection);
     }
 
     await connection.join(subscriber._id);
     await this.subscriberOnlineService.handleConnection(subscriber);
-
-    connection.on('disconnect', async () => {
-      const activeConnections = await this.getActiveConnections(connection, subscriber._id);
-      await this.subscriberOnlineService.handleDisconnection(subscriber, activeConnections);
-    });
```
scopsy, line 38: This could potentially lead to a memory hog with all existing running connections

apps/ws/src/socket/ws.gateway.ts
```diff
@@ -1,41 +1,113 @@
...
+    const _this = this;
 
-    if (!token || token === 'null') {
-      return this.disconnect(connection);
-    }
+    return new Promise((resolve, reject) => {
+      nr.startBackgroundTransaction(
+        ObservabilityBackgroundTransactionEnum.WS_HANDLE_SOCKET_CONNECTION,
+        'WS Service',
+        function () {
```
scopsy, line 47: Wrapped the handler functions in observanilty log.

apps/ws/src/socket/ws.gateway.ts
```diff
@@ -1,41 +1,113 @@
...
 
 @WebSocketGateway()
-export class WSGateway implements OnGatewayConnection {
+export class WSGateway implements OnGatewayConnection, OnGatewayDisconnect {
   constructor(private jwtService: JwtService, private subscriberOnlineService: SubscriberOnlineService) {}
 
   @WebSocketServer()
   server: Server;
 
+  async handleDisconnect(connection: Socket) {
```
scopsy, line 16: Refactored this to utilize the nestjs hook for disconnect instead of managing it by ourself.

apps/ws/src/socket/ws.gateway.ts
```diff
@@ -1,41 +1,113 @@
...
     } catch (e) {
+      return;
+    }
+  }
+
+  private async processDisconnectionRequest(connection: Socket) {
+    const token = this.extractToken(connection);
+
+    if (!token || token === 'null') {
+      return;
```
djabarovgeorge, line 85: should we try to disconnect here and on the 'return' below??
```suggestion
      return this.disconnect(connection);
 ```

Comment (scopsy) [reply apps/ws/src/socket/ws.gateway.ts:85]: This hook is fired on the disconnect action, so it will disconnect regardless here

File: apps/ws/src/socket/ws.gateway.ts
Comment (djabarovgeorge) [inline apps/ws/src/socket/ws.gateway.ts:1]: There are a lot of blind spots, should we add logs there in case we will need to debug why there were missing subscriber/token etc?

File: libs/shared/src/config/job-queue.ts [hunk 5]
```diff
@@ -13,4 +13,6 @@ export enum ObservabilityBackgroundTransactionEnum {
   SUBSCRIBER_PROCESSING_QUEUE = 'subscriber-processing-queue',
   TRIGGER_HANDLER_QUEUE = 'trigger-handler-queue',
   WS_SOCKET_QUEUE = 'ws_socket_queue',
+  WS_HANDLE_SOCKET_CONNECTION = 'ws_handle_socket_connection',
```
djabarovgeorge, line 16:
```suggestion
  WS_HANDLE_SOCKET_CONNECTION = 'ws_socket_handle_connection',
```

libs/shared/src/config/job-queue.ts
```diff
@@ -13,4 +13,6 @@ export enum ObservabilityBackgroundTransactionEnum {
...
+  WS_HANDLE_SOCKET_DISCONNECT = 'ws_handle_socket_disconnect',
```
djabarovgeorge, line 17:
```suggestion
  WS_HANDLE_SOCKET_DISCONNECT = 'ws_socket_handle_disconnect',
```

libs/shared/src/config/job-queue.ts
djabarovgeorge: feels more readable to me, <service-name><action>
what do you think?

djabarovgeorge (approved):

apps/ws/src/socket/ws.gateway.ts
scopsy: That's defiently true, I just won't be able to do this as part of the current scope of work. Will create a smaller task to handle this down the road

PR #4873, opened 2023-11-19
Refactor SubscribersService for Dependency Injection

Atharva1723:
1) Refactored **'SubscribersService'** to use dependency injection for **'SubscriberRepository'** and **'IntegrationRepository'**.
2) Updated the constructor to accept these repositories as parameters.

1) Using dependency injection makes the class more testable and flexible.
2) Allows for easy substitution of repository implementations, improving maintainability.


scopsy:
Thank you for the contribution @Atharva1723 but please create an issue with proposed changes and their reasoning before submitting a PR.

In this case all the test suite will be broken, since you haven'y updated the users of this class.

Also, this is the intentional design as the subscriber repositories are an implementation detail we don't want to expose or to be able to dynamically swap. 

Closing this for now 🙏

Atharva1723:
Okay understood 👍

PR #4884, opened 2023-11-22
feat: add launch darkly context in frontend

davidsoderberg:
See example I did follow on this page: https://docs.launchdarkly.com/sdk/client-side/react/react-web#initializing-the-react-sdk


Cliftonz:
Just to follow up, this covers also when they switch between organizations correct?

davidsoderberg:
> Just to follow up, this covers also when they switch between organizations correct?

Yes so when either the organization id or name change it will update.

Cliftonz (approved):

apps/web/src/components/providers/AuthProvider.tsx
```diff
@@ -26,6 +27,19 @@ export const useAuthContext = (): UserContext => useContext(AuthContext);
...
+    if (!organization?._id) {
+      return;
+    }
+
+    ldClient?.identify({
+      kind: 'Organization',
+      key: organization._id,
+      name: organization.name,
+    });
+  }, [organization?._id, ldClient, organization?.name]);
```
rifont, line 42: What do you think about extracting this into a separate hook? That way, if we change FF providers we can keep the code change isolated.

```typescript
import { useLDClient } from 'launchdarkly-react-client-sdk';

const useFeatureFlagClient = (organization: IOrganizationEntity) => {

  useEffect(() => {
    if (!organization?._id) {
      return;
    }

    ldClient?.identify({
      kind: 'Organization',
      key: organization._id,
      name: organization.name,
    });
  }, [organization?._id, ldClient, organization?.name]);
}
```

rifont (approved):

apps/web/src/components/providers/AuthProvider.tsx
Cliftonz: We many need to change in the future due to the change with the platform so I think this would be a good idea.

PR #4888, opened 2023-11-23
Feature : Ability to cancel any arbitrary events from digest

djabarovgeorge:

Subscriber receives messages in a discord like platform. Each time a new message is sent to user, workflow is triggered and added into digest. This digest is send to user at some defined time of the day. before that defined time, if user sees any of the message, it should be removed from digest and rest of unseen messages should stay in digest

#### Current Working
Novu allows canceling only first event from digest. and if first event is cancelled then digest is deleted and all events in digest got deleted.

Additional cancellation support.

#### Defintion of Done.
if there are 4 events in digest, user should be able to cancel 3rd or 2nd or 4th event before digest time completion and that cancelled event should not be included in final digest events array

### The current solution I came up with help from Gali, David, and Dima.
The resolution involves:

* When canceling the main digest through API, responsibility for the digest is shifted to the subsequently merged digest transaction.
* Once acknowledging the cancellation of the main digest in the trigger phase, if a digest follower has carried responsibility for its execution, their transaction becomes the primary one.
That means executing its digest, updating all associated child job steps, and queuing their jobs for execution under 'QueueNextJob'.

Another note that i want to raise is that i was trying to avoid creating a new job in the API level as it would mean that we would need to add `AddJob` use case and its queue service dep that connects to `standard` queue. Creating those changes in the API would corrupt its responsibility and couple us to the `standard` queue that we just decoupled.

Miro REF - https://miro.com/app/board/uXjVNL6VbtQ=/

The main concern is about other than regular digest flow, as at his task i was mainly working with regular digest. would love to hear your opinion on the matter @davidsoderberg @ainouzgali


apps/api/src/app/events/e2e/cancel-event.e2e.ts
djabarovgeorge, line 1: There is some duplication in the new tests and comments as I thought it would be easier to read in the future.

apps/api/src/app/events/e2e/delay-events.e2e.ts
djabarovgeorge, line 1: Moved to dedicated file cancel-event.e2e.ts

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
```diff
@@ -1,30 +1,33 @@
...
   constructor(private jobRepository: JobRepository) {}
 
   public async execute(command: CancelDelayedCommand): Promise<boolean> {
-    const jobs = await this.jobRepository.find(
+    const transactionJobs: PartialJob[] = await this.jobRepository.find(
       {
         _environmentId: command.environmentId,
         transactionId: command.transactionId,
-        status: JobStatusEnum.DELAYED,
+        status: [JobStatusEnum.DELAYED, JobStatusEnum.MERGED, JobStatusEnum.PENDING],
```
djabarovgeorge, line 17: JobStatusEnum.MERGED was added by David in order to solve the 2nd, 3rd, etc.. cancellations.

In addition, I added JobStatusEnum.PENDING as it looks like we were keeping transaction child pending jobs in the database that would never resolve. IMO they should be canceled as well, let me know if someone knows if it was intentional.

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
```diff
@@ -34,6 +37,71 @@ export class CancelDelayed {
       }
     );
 
+    const mainDigestJob = transactionJobs.find((job) => isMainDigest(job.type, job.status));
```
djabarovgeorge, line 40: Here i want to point to the naming issue I had, i am not entirely sure what to call the digest that is responsible for executing the digest in the end of the digest time.

At the current state, I referred to this digest job as `main`, and if we would find a new transaction that will place it a `follower`.

If someone has any other ideas I'd love to hear them.

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
```diff
@@ -34,6 +37,71 @@ export class CancelDelayed {
...
+        _environmentId: job._environmentId,
+        _subscriberId: job._subscriberId,
+      },
+      '',
+      {
+        query: { sort: { createdAt: 1 } },
+      }
+    );
+
+    // meaning that only one trigger was send, and it was cancelled in the CancelDelayed.execute
```
djabarovgeorge, line 64: Not sure if keep those comments in the code or not, because IMO this flow is very complicated to grasp.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -114,19 +119,54 @@ export class RunJob {
...
+    job: JobEntity
+  ): Promise<{ canceled: boolean; activeDigestFollower: JobEntity | null }> {
+    let activeDigestFollower: JobEntity | null = null;
+
     if (job.type !== StepTypeEnum.DIGEST && job.type !== StepTypeEnum.DELAY) {
-      return false;
+      return { canceled: false, activeDigestFollower };
     }
 
-    const count = await this.jobRepository.count({
```
djabarovgeorge, line 123: Removed this query because we already have the newly fetched job and its status.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -114,19 +119,54 @@ export class RunJob {
...
+    const canceled = job.status === JobStatusEnum.CANCELED;
+
+    if (job.status === JobStatusEnum.CANCELED) {
+      activeDigestFollower = await this.activeDigestMainFollowerExist(job);
```
djabarovgeorge, line 150: If the current digest is cancelled we are looking for follower.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -114,19 +119,54 @@ export class RunJob {
...
+    return { canceled, activeDigestFollower };
+  }
+
+  @Instrument()
+  private async activeDigestMainFollowerExist(job: JobEntity): Promise<JobEntity | null> {
+    if (job.type !== StepTypeEnum.DIGEST) {
+      return null;
+    }
+
+    return await this.jobRepository.findOne({
```
djabarovgeorge, line 162: Does anyone with digest expertise could let me know if we need to fetch by digest key here?

packages/application-generic/src/usecases/get-feature-flag/get-use-merged-digest-id.use-case.ts
```diff
@@ -6,7 +6,8 @@ import { GetFeatureFlag } from './get-feature-flag.use-case';
 @Injectable()
 export class GetUseMergedDigestId extends GetFeatureFlag {
   async execute(featureFlagCommand: FeatureFlagCommand): Promise<boolean> {
-    const value = process.env.IS_USE_MERGED_DIGEST_ID_ENABLED;
+    const value = 'true';
+    // const value = process.env.IS_USE_MERGED_DIGEST_ID_ENABLED;
```
djabarovgeorge, line 10: 🗑️

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
davidsoderberg: If I remember correctly we skipped pending because there could be a matter of who getting first in workflows without digest. Like what happens if they cancel it and we find it here at the same time it is picked up and put on the queue. How will it be handled?

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
davidsoderberg: I do not think so

davidsoderberg (approved):
Really amazing work with this 👑

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
djabarovgeorge: I see, so i think about two different points we need to think about here:
1. this should be okay because, at the current project state, we do not support cancelation of a job of CHANNEL type, meaning only ACTION which is delay and digest supports cancelation.
2. we do not support Channel cancelation, meaning if we cancel here workflow without digest we will get the following flow with corrupted status:
    1. run a job (in-app for example), 
    2. which will be executed because we won't check its status (cancel)
    3. once we finish the execution we mark it as Completed.
    

Let me know if you think I missed something in the flow.

One solution would be to add another query after
```
.find({
      {
        status: [JobStatusEnum.DELAYED, JobStatusEnum.MERGED,
        })
```
if transactionJobs is truthy it means that this is a digest workflow transaction fetch for its PENDING jobs and then cancel them all.

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
djabarovgeorge, line 1: I remembered some doubts i had here, i wonder if we should add lock in this use case
@davidsoderberg what do you think?

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
davidsoderberg: I think we can go with your solution because cancel and then complete it could create confusion for the user.

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
davidsoderberg: And use the main digest as lock object?

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
djabarovgeorge: Awesome thanks for the feedback we could have missed it otherwise!! i will create an update :)

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
djabarovgeorge: Well to tell the truth I am not sure

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
LetItRock: @davidsoderberg shouldn't we? 
we do merge digests by the "digest key" so the logic that calculates the new "follower" in the cancel use-case is fine... however, I think that here we need to fetch by "digest key" because there might be many delayed digests with different "digest keys"...

LetItRock (commented):
thank you George, amazing work! 🔥 👏 
left a few small suggestions and questions ;)

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
davidsoderberg: Good catch, my thought was that there should only be one delayed digest job for the specific subscriber and template so digest key would be not needed. But maybe it is good to add it as well to be 100% sure.

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
```diff
@@ -1,30 +1,49 @@
...
+      const possiblePendingJobs: PartialJob[] = await this.jobRepository.find(
+        {
+          _environmentId: command.environmentId,
+          transactionId: command.transactionId,
+          status: [JobStatusEnum.PENDING],
+        },
+        '_id type status _environmentId _subscriberId'
+      );
+
+      transactionJobs = [...transactionJobs, ...possiblePendingJobs];
```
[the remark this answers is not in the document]
djabarovgeorge: At the current state while we process a notification we create all its step's jobs as pending.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -50,13 +50,18 @@ export class RunJob {
...
+
+    if (canceled && !activeDigestFollower) {
       Logger.verbose({ canceled }, `Job ${job._id} that had been delayed has been cancelled`, LOG_CONTEXT);
 
       return;
     }
 
+    if (activeDigestFollower) {
+      job = this.assignNewDigestExecutor(activeDigestFollower);
+    }
```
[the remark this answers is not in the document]
djabarovgeorge: Good catch! We definitely should because the transaction/job ids are updated.
I will re-assign the logger if we re-assign the job. and move the new relic lower because once the custom attribute has been created and reported to new relic with a value, the value cannot be changed or updated.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
djabarovgeorge: Added 🙏

djabarovgeorge: Thanks for the feedback, I added digest key🙏

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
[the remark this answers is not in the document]
djabarovgeorge: @LetItRock If you think we should rethink the flow please create an initial ticket for us so we can track this thought.

packages/application-generic/src/usecases/digest-filter-steps/digest-filter-steps.usecase.ts
djabarovgeorge, line 1: this function is not digest filter related and we can move to utils/object in order to reuse it.

apps/worker/src/app/workflow/usecases/send-message/digest/get-digest-events.usecase.ts
```diff
@@ -24,30 +22,12 @@ const LOG_CONTEXT = 'GetDigestEvents';
 export abstract class GetDigestEvents {
   constructor(protected jobRepository: JobRepository, private createExecutionDetails: CreateExecutionDetails) {}
 
-  protected getJobDigest(job: JobEntity): {
```
djabarovgeorge, line 27: this function is not get-digest-events related and we can move to utils/digest in order to reuse it.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -114,19 +121,77 @@ export class RunJob {
...
+    return { canceled, activeDigestFollower };
+  }
+
+  @Instrument()
+  private async activeDigestMainFollowerExist(job: JobEntity): Promise<JobEntity | null> {
+    if (job.type !== StepTypeEnum.DIGEST) {
+      return null;
+    }
+
+    const { digestKey, digestValue } = getJobDigest(job);
```
djabarovgeorge, line 178: after we moved the getNestedValue getJobDigest function into separated stateless functions now we can reuse getJobDigest here.

libs/dal/src/repositories/job/job.entity.ts
```diff
@@ -21,7 +21,7 @@ export class JobEntity {
   _notificationId: string;
   subscriberId: string;
   _subscriberId: string;
-  _mergedDigestId?: string;
+  _mergedDigestId?: string | null;
```
djabarovgeorge, line 24: Now, we can remove assignments from _mergedDigestId. When a merged digest becomes a `main follower`, it's no longer merged, so it can be empty/null.

apps/api/src/app/events/usecases/cancel-delayed/cancel-delayed.usecase.ts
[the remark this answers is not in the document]
LetItRock: this is the ticket actually: https://linear.app/novu/issue/NV-2671/optimize-regular-digest ;)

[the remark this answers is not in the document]
djabarovgeorge: Oh sorry, now i understand the comment you raised in discord better. Thanks for raising it up.

LetItRock (approved):

PR #4895, opened 2023-11-26
docs(providers): Fix syntax mistake

Atharva1723:

The import statement in the code was not right. The correct way to import is using ' ' quotes

The current code will not handle the import as there is an syntax mistake . It should be corrected for the correct working of the code

image


scopsy (approved):
Thank you!

PR #4898, opened 2023-11-26
refactor(dal): improve subscriber preference schema

scopsy:


libs/dal/src/repositories/subscriber-preference/subscriber-preference.schema.ts
djabarovgeorge, line 1: do you think we need to keep _environment index? 
and if we need to keep it maybe we could create a new compound index so all the indexes in this collection would be under the same type.

libs/dal/src/repositories/subscriber-preference/subscriber-preference.schema.ts
```diff
@@ -55,6 +52,16 @@ const subscriberPreferenceSchema = new Schema<SubscriberPreferenceDBModel>(
...
 );
 
+subscriberPreferenceSchema.index({
+  _subscriberId: 1,
+  _templateId: 1,
+});
+
+subscriberPreferenceSchema.index({
+  _subscriberId: 1,
+  level: 1,
```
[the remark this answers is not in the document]
scopsy: I added this here since we are doing a global check for each trigger (for both global and template specific). Index size will be quite small, due to low cardinality. But I would keep this one here just for the small boost in query time.

libs/dal/src/repositories/subscriber-preference/subscriber-preference.schema.ts
scopsy: I think we can keep the existing _environmentId, and later we can do some consolidation and cleanup. (Won't have time to go over it right now, there are other things there that I noticed we can potentially remove)

djabarovgeorge (approved):
🚀

PR #4903, opened 2023-11-28
fix: senderName and subject override for email providers

jainpawan21:

- Fix sendername and subject override bug
- Add support of overridden sendername in those email providers which support separate field for sendername


LetItRock (approved):
thank you! 🔥

PR #4906, opened 2023-11-28
Race condition when reading the service instance - SW Health Indicator

djabarovgeorge:

Reproduction Steps

There is a race condition happening in the WS service when reading the server instance sockets. The error message that was noticed has been like Cannot read proprties of sockets of null. 

The code part where it is happening:

```
private async connectionExist(command: ExternalServicesRouteCommand) {
  return !!(await this.wsGateway.server.sockets.in(command.userId).fetchSockets()).length;
}
```

The issue should be reproducible when spinning the service (or a few instances).

Will validate the WS server health.

### DIFF
--- a/apps/ws/src/health/health.controller.ts
+++ b/apps/ws/src/health/health.controller.ts
@@ -3,13 +3,15 @@ import { HealthCheck, HealthCheckResult, HealthCheckService, HttpHealthIndicator
 import { DalServiceHealthIndicator, WebSocketsQueueServiceHealthIndicator } from '@novu/application-generic';
 
 import { version } from '../../package.json';
+import { WSHealthIndicator } from '../socket/services';
 
 @Controller('v1/health-check')
 export class HealthController {
   constructor(
     private healthCheckService: HealthCheckService,
     private dalHealthIndicator: DalServiceHealthIndicator,
-    private webSocketsQueueHealthIndicator: WebSocketsQueueServiceHealthIndicator
+    private webSocketsQueueHealthIndicator: WebSocketsQueueServiceHealthIndicator,
+    private wsHealthIndicator: WSHealthIndicator
   ) {}
 
   @Get()
@@ -18,6 +20,7 @@ export class HealthController {
     const result = await this.healthCheckService.check([
       async () => this.dalHealthIndicator.isHealthy(),
       async () => this.webSocketsQueueHealthIndicator.isHealthy(),
+      async () => this.wsHealthIndicator.isHealthy(),
       async () => {
         return {
           apiVersion: {
--- a/apps/ws/src/socket/services/ws-health-indicator.service.ts
+++ b/apps/ws/src/socket/services/ws-health-indicator.service.ts
@@ -1,0 +1,25 @@
+import { HealthIndicator, HealthIndicatorResult } from '@nestjs/terminus';
+import { Injectable } from '@nestjs/common';
+
+import { IHealthIndicator } from '@novu/application-generic';
+
+import { WSGateway } from '../ws.gateway';
+
+@Injectable()
+export class WSHealthIndicator extends HealthIndicator implements IHealthIndicator {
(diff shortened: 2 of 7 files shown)


apps/ws/src/health/health.controller.ts
```diff
@@ -18,6 +20,7 @@ export class HealthController {
     const result = await this.healthCheckService.check([
       async () => this.dalHealthIndicator.isHealthy(),
       async () => this.webSocketsQueueHealthIndicator.isHealthy(),
+      async () => this.wsHealthIndicator.isHealthy(),
```
djabarovgeorge, line 23: Now we will check if the WS server is up.

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
```diff
@@ -15,18 +15,21 @@ export class ExternalServicesRoute {
...
+    if (command.event === WebSocketEventEnum.RECEIVED) {
+      await this.processReceivedEvent(command);
+    }
+
+    if (command.event === WebSocketEventEnum.UNSEEN) {
+      await this.sendUnseenCountChange(command);
+    }
+
+    if (command.event === WebSocketEventEnum.UNREAD) {
+      await this.sendUnreadCountChange(command);
```
djabarovgeorge, line 32: Small refactor, not related to the PR

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
```diff
@@ -127,7 +130,13 @@ export class ExternalServicesRoute {
     }
   }
 
-  private async connectionExist(command: ExternalServicesRouteCommand) {
+  private async connectionExist(command: ExternalServicesRouteCommand): Promise<boolean | undefined> {
+    if (!this.wsGateway.server) {
```
djabarovgeorge, line 134: If the server is not initialized we want to log it.

apps/ws/src/socket/ws.gateway.ts
```diff
@@ -1,17 +1,21 @@
...
 
+const LOG_CONTEXT = 'WSGateway';
+
 @WebSocketGateway()
 export class WSGateway implements OnGatewayConnection, OnGatewayDisconnect {
   constructor(private jwtService: JwtService, private subscriberOnlineService: SubscriberOnlineService) {}
 
   @WebSocketServer()
-  server: Server;
+  server: Server | null;
```
djabarovgeorge, line 18: Now we know that there is an edge case where the Server can be nullish.

apps/ws/src/socket/ws.gateway.ts
```diff
@@ -112,6 +116,12 @@ export class WSGateway implements OnGatewayConnection, OnGatewayDisconnect {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   async sendMessage(userId: string, event: string, data: any) {
+    if (!this.server) {
+      Logger.error('No sw server available to send message', LOG_CONTEXT);
+
```
djabarovgeorge, line 121: If the server is not initialized we want to log it.

apps/ws/src/socket/services/ws-health-indicator.service.ts
```diff
@@ -0,0 +1,25 @@
+import { HealthIndicator, HealthIndicatorResult } from '@nestjs/terminus';
+import { Injectable } from '@nestjs/common';
+
+import { IHealthIndicator } from '@novu/application-generic';
+
+import { WSGateway } from '../ws.gateway';
+
+@Injectable()
+export class WSHealthIndicator extends HealthIndicator implements IHealthIndicator {
```
djabarovgeorge, line 9: Will rename in the upcoming PR's in order to make clear what the health indicator is responsible for.

```suggestion
export class WSServerHealthIndicator extends HealthIndicator implements IHealthIndicator {
```

PR #4908, opened 2023-11-28
Refactor readiness service to receive indicator list

djabarovgeorge:

In this PR we readiness service to receive an indicator list, it is a preparation so we could add readiness to the WS app.

In order to make readiness service more modular.


packages/application-generic/src/services/readiness/readiness.service.ts
```diff
@@ -58,13 +50,15 @@ export class ReadinessService {
...
+      const healths = await Promise.all(
+        this.indicators.map((indicator) => indicator.isHealthy())
+      );
+
+      const statuses = healths.map(
+        (health: HealthIndicatorResult) => Object.values(health)[0].status
+      );
 
-      return healths.every((health) => !!health === true);
+      return statuses.every((status: HealthIndicatorStatus) => status === 'up');
```
djabarovgeorge, line 61: Created this explicit check of the status.
The old way works now because we throw exceptions if the status `false`, however, i would like to be explicit if possible.

PR #4909, opened 2023-11-28
Set up a readiness check on the Websocket system

djabarovgeorge:

We have observed that the websocket container is having issues starting due to some issue with the queue and websocket service starting at the same time.

This is causing the websockets to take 3+ minutes to become ready to service requests.

Websocket has a readiness check that prevents the health check from failing until the system is ready


apps/worker/src/bootstrap.ts
```diff
@@ -66,15 +66,16 @@ export async function bootstrap(): Promise<INestApplication> {
...
 
   Logger.log('BOOTSTRAPPED SUCCESSFULLY');
 
-  await app.listen(process.env.PORT);
+  await app.init();
 
   await startAppInfra(app);
 
+  await app.listen(process.env.PORT);
+
```
djabarovgeorge, line 78: Does anyone know if we need to start listening before we startAppInfra? IMO we first need to set up the app and then start listening to new requests.

I made the same for WS, there i noticed that nest is not initialized properly, and `app.init()` helps to initialize the app without starting to listen for new requests.

apps/ws/src/readiness/readiness.module.ts
djabarovgeorge, line 1: Created a module in order to collect the indicators in one place.

apps/ws/src/readiness/readiness.module.ts
```diff
@@ -0,0 +1,47 @@
...
+  DalServiceHealthIndicator,
+
+  WebSocketsQueueService,
+
+  indicatorList,
+  ReadinessService,
+];
+
+@Module({
+  imports: [SharedModule, SocketModule],
```
djabarovgeorge, line 43: If we import the needed indicator services here separately I noticed that Nest will create separate instances, which is not the desired behavior because we want to validate the health of the connection that the APP is using.

In order to get the same instance we need to use the module that initializes the services.
* for DalServiceHealthIndicator we import SharedModule.
* for the wsServerHealthIndicator we import SocketModule.

apps/ws/src/readiness/readiness.module.ts
```diff
@@ -0,0 +1,47 @@
...
+import { WSServerHealthIndicator } from '../socket/services';
+import { SharedModule } from '../shared/shared.module';
+import { SocketModule } from '../socket/socket.module';
+
+export const indicatorList = {
+  provide: 'INDICATOR_LIST',
+  useFactory: (
+    wsServerHealthIndicator: WSServerHealthIndicator,
+    webSocketsQueueServiceHealthIndicator: WebSocketsQueueServiceHealthIndicator,
+    dalServiceHealthIndicator: DalServiceHealthIndicator
```
djabarovgeorge, line 18: Unlike the worker implementation, here i added all the service indicators and not only the queue. would love to hear what you think about it.

apps/ws/src/socket/services/cold-start.service.ts
djabarovgeorge, line 1: For now i duplicated the cold start code, we could make it more dynamic in the future.

packages/application-generic/src/modules/queues.module.ts
```diff
@@ -38,11 +39,13 @@ export const workerIndicatorList = {
     workflowQueueServiceHealthIndicator: WorkflowQueueServiceHealthIndicator,
     subscriberProcessQueueHealthIndicator: SubscriberProcessQueueHealthIndicator
   ) => {
-    return [
+    const indicatorList: IHealthIndicator[] = [
```
djabarovgeorge, line 42: make the type explicit just in case

packages/application-generic/src/services/workers/active-jobs-metric-worker.service.ts
```diff
@@ -9,6 +9,5 @@ const LOG_CONTEXT = 'ActiveJobsMetricWorkerService';
 export class ActiveJobsMetricWorkerService extends WorkerBaseService {
   constructor() {
     super(JobTopicNameEnum.ACTIVE_JOBS_METRIC);
-    Logger.log(`Worker ${this.topic} instantiated`, LOG_CONTEXT);
```
djabarovgeorge, line 12: removed this duplicated log, we already log that the worker was initiated in the initWorker() in the base class, we could add something else here if needed.

LetItRock (approved):
💎

PR #4910, opened 2023-11-28
feat(api): Add API rate limiting NestJS guard

rifont:
* Adds an API Rate Limit guard to throttle inbound API requests
  * Extended from `@nestjs/throttler`
  * Implements the guard as an interceptor, to access Auth guard context in execution flow
  * Add Cost decorator to use on controllers and methods
  * Add Category decorator to use on controllers and methods
  * Unit test for single-cost endpoint use and multi-category
* Specifies the guard as an App Interceptor (Rate Limiting is currently toggled off by default)
* Add strong environment variable typings

API Rate Limiting increases API resiliency by limiting the number of requests that a single API client can make.
The guards will be used to implement variable-cost endpoints and variable-category rate limits.

_Note_
E2E tests for bulk cost endpoints are included in a subsequent PR: https://github.com/novuhq/novu/pull/4911

_Tests_
image


libs/dal/src/repositories/environment/environment.repository.ts
```diff
@@ -87,4 +88,23 @@ export class EnvironmentRepository extends BaseRepository<EnvironmentDBModel, En
...
+            apiRateLimits: {
+              $mergeObjects: ['$apiRateLimits', apiRateLimits],
+            },
+          },
+        },
+      ]
+    );
+
+    return await this.getApiKeys(environmentId);
+  }
```
rifont, line 109: This will be used to add custom rate limits per environment when enterprise customers need custom limits. It's also used for testing.

libs/dal/src/repositories/organization/organization.repository.ts
```diff
@@ -58,6 +59,19 @@ export class OrganizationRepository extends BaseRepository<OrganizationDBModel,
...
+        _id: organizationId,
+      },
+      {
+        $set: {
+          apiServiceLevel,
+        },
+      }
+    );
+  }
+
```
rifont, line 74: This will be used to set a custom API service level when customers upgrade to higher product tiers. It's only used for testing now.

libs/testing/src/user.session.ts
```diff
@@ -332,4 +334,14 @@ export class UserSession {
...
+
+  public async updateOrganizationServiceLevel(serviceLevel: ApiServiceLevelEnum) {
+    const organizationService = new OrganizationService();
+
+    await organizationService.updateServiceLevel(this.organization._id, serviceLevel);
+  }
+
+  public async updateEnvironmentApiRateLimits(apiRateLimits: Partial<IApiRateLimitMaximum>) {
+    await this.environmentService.updateApiRateLimits(this.environment._id, apiRateLimits);
+  }
```
rifont, line 346: Adding methods to customise the organization service level and environment rate limit in tests.

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -26,6 +26,11 @@ export class JwtAuthGuard extends AuthGuard('jwt') {
     const request = context.switchToHttp().getRequest();
     const authorizationHeader = request.headers.authorization;
 
+    if (authorizationHeader) {
+      const authScheme = authorizationHeader.split(' ')[0];
+      request.authScheme = authScheme;
+    }
+
```
rifont, line 33: This is a little ugly right now. It's required because we currently override the existing auth header when an API key is used to authenticate. Once we remove this override, we can let the downstream execution flow read the auth header again as normal.

It will be fixed in https://linear.app/novu/issue/NV-3055/🔐-fix-apikey-auth-guard-performance

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-maximum/get-api-rate-limit-maximum.usecase.ts
```diff
@@ -24,9 +24,9 @@ export class GetApiRateLimit {
   }
 
   @CachedEntity({
-    builder: (command: GetApiRateLimitCommand) =>
+    builder: (command: { apiRateLimitCategory: ApiRateLimitCategoryEnum; _environmentId: string }) =>
       buildMaximumApiRateLimitKey({
-        _environmentId: command.environmentId,
+        _environmentId: command._environmentId,
```
djabarovgeorge, line 29: sorry about that, i still have not found a way to do the cache decorator type safe. the input signature in the builder should be the same as the wrapped function input.

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts
```diff
@@ -0,0 +1,86 @@
...
+    const previousHash = await this.cacheService.get(cacheKey);
+    const newHash = this.getConfigHash(newDefault);
+
+    if (previousHash !== newHash) {
+      await this.cacheService.set(cacheKey, newHash);
+
+      this.invalidateCache.invalidateQuery({
+        key: buildMaximumApiRateLimitKey({
+          _environmentId: '*',
+          apiRateLimitCategory: '*',
```
djabarovgeorge, line 45: what is the motivation behind the `*` character?

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const cacheKey = buildServiceConfigApiRateLimitMaximumKey();
+    const previousHash = await this.cacheService.get(cacheKey);
+    const newHash = this.getConfigHash(newDefault);
+
+    if (previousHash !== newHash) {
+      await this.cacheService.set(cacheKey, newHash);
+
+      this.invalidateCache.invalidateQuery({
+        key: buildMaximumApiRateLimitKey({
```
djabarovgeorge, line 43: I would not suggest using invalidateQuery with `type: CacheKeyTypeEnum.ENTITY` that is used in buildMaximumApiRateLimitKey, unless i am missing some new implementation that i am not aware of.

invalidateCache.invalidateByKey - responsibility to delete by key
invalidateCache.invalidateQuery - while responsible for deleting all nested SET members that are stored in the redis.

libs/dal/src/repositories/environment/environment.repository.ts
```diff
@@ -87,4 +88,23 @@ export class EnvironmentRepository extends BaseRepository<EnvironmentDBModel, En
...
+          $set: {
+            apiRateLimits: {
+              $mergeObjects: ['$apiRateLimits', apiRateLimits],
+            },
+          },
+        },
+      ]
+    );
+
+    return await this.getApiKeys(environmentId);
```
djabarovgeorge, line 108: Do we need to return the api keys here?

djabarovgeorge (commented):
Looks like an amazing job done, left a couple of comments regarding stuff i was not sure about.

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -0,0 +1,182 @@
...
+  RATE_LIMIT_LIMIT = 'RateLimit-Limit',
+  RATE_LIMIT_RESET = 'RateLimit-Reset',
+  RATE_LIMIT_POLICY = 'RateLimit-Policy',
+  RETRY_AFTER = 'Retry-After',
+  USER_AGENT = 'User-Agent',
+  AUTHORIZATION = 'Authorization',
+}
+
+export const THROTTLED_EXCEPTION_MESSAGE = 'API rate limit exceeded';
+export const ALLOWED_AUTH_SCHEMES = ['ApiKey'];
```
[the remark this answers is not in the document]
rifont: Right now we are not planning to rate limit requests with a Bearer token, because those requests will come from the browser (unless someone spoofs an auth flow outside the browser) and browser-based activity is unlikely to cause extreme server load.

Keen to hear your thoughts on whether you think we should indeed be rate limiting Bearer Auth requests.

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -0,0 +1,182 @@
...
+    if (!authorization) {
+      return false;
+    }
+
+    return ALLOWED_AUTH_SCHEMES.some((scheme) => req.authScheme === scheme);
+  }
+
+  private getReqUser(context: ExecutionContext): IJwtPayload {
+    const req = context.switchToHttp().getRequest();
+    const token = req.headers[RateLimitHeaderKeysEnum.AUTHORIZATION.toLowerCase()].split(' ')[1];
```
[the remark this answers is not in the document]
rifont: See https://github.com/novuhq/novu/pull/4910#discussion_r1409001401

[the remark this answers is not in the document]
rifont: Right now, we are overwriting the `Authorization: ApiKey: XXXXX` header with an `Authorization: Bearer XXXXX` when a request uses API key auth. So this will always be a JWT token right now.

https://linear.app/novu/issue/NV-3055/🔐-fix-apikey-auth-guard-performance will clean this logic up, and we can adjust the flow here - notably, the `Remove all use of JWT (e.g. jwt.decode)` in the task description will deal with this particular instance.

libs/dal/src/repositories/environment/environment.repository.ts
rifont: Good catch! This was accidentally left behind from a copy-paste. I've updated it to return the `this.update` result instead.

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts
```diff
@@ -39,7 +39,7 @@ export class GetApiRateLimitServiceMaximumConfig implements OnModuleInit {
     if (previousHash !== newHash) {
       await this.cacheService.set(cacheKey, newHash);
 
-      this.invalidateCache.invalidateQuery({
+      this.invalidateCache.invalidateByKey({
```
rifont, line 42: @djabarovgeorge made this change per the PR feedback: https://github.com/novuhq/novu/pull/4910#discussion_r1408493196

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts
rifont: The intention here is to cater to the following scenario:
1. A Novu platform administrator makes changes to the default API rate limit service config environment variables. All [cached maximum environment limits](https://github.com/novuhq/novu/blob/3c1fd8a30889097da4840dd19cd7975dffc67453/apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-maximum/get-api-rate-limit-maximum.usecase.ts#L26) for all Organizations, for all Environments need to be invalidated. This cached entity supports fast lookups for an Environment's max limit.
2. When a change is detected (via the modified hash), invalidate all cached entities.
3. Update the hash of the service config

This pattern could be used to support other platform service configuration variables that require cache invalidation. This will typically be the case when entities are cached for performance reasons, but still depend on defaults.

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts
rifont: Resolved to use the `invalidateByKey` method. Thanks for highlighting.

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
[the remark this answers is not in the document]
rifont: @djabarovgeorge and I discussed this point, we agreed that unless there is a good reason not to limit Bearer token requests, we should limit their consumption.

However, we will need to add support for a custom cache prefix to be passed to the evaluate rate limit use-case to support multiple token buckets to consume from - it's not desirable for both authentication schemes to share the same bucket as this may be confusing to the user.

Given the additional work required and deviation from the initial scope, we should address this in a separate PR if we think we should limit Bearer token requests.

apps/api/src/app/rate-limiting/guards/throttler.guard.e2e.ts
```diff
@@ -0,0 +1,309 @@
...
+describe('API Rate Limiting', () => {
+  let session: UserSession;
+  const pathPrefix = '/v1/rate-limiting';
+  let request: (
+    path: string,
+    authHeader?: string
+  ) => Promise<Awaited<ReturnType<typeof UserSession.prototype.testAgent.get>>>;
+
+  beforeEach(async () => {
+    process.env.IS_API_RATE_LIMITING_ENABLED = 'true';
```
rifont, line 33: The [E2E tests are all failing in CI](https://github.com/novuhq/novu/actions/runs/7033263232/job/19138794738?pr=4910), it looks as though this runtime flag is not propagating.

Is it necessary to set `IS_API_RATE_LIMITING_ENABLED` to `true` in the `.env/test` for this to succeed? If yes, how do we test for the disabled flag case?

packages/application-generic/src/services/auth/auth.guard.ts
Cliftonz: I will make sure to prioritize NV-3055 for the next cycle

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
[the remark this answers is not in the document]
Cliftonz: @djabarovgeorge I have prioritized 3055 in the next cycle as this debt.

apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts
Cliftonz: Amazing!

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -0,0 +1,182 @@
...
+export const THROTTLED_EXCEPTION_MESSAGE = 'API rate limit exceeded';
+export const ALLOWED_AUTH_SCHEMES = ['ApiKey'];
+
+const defaultApiRateLimitCategory = ApiRateLimitCategoryEnum.GLOBAL;
+const defaultApiRateLimitCost = ApiRateLimitCostEnum.SINGLE;
+
+/**
+ * An interceptor is used instead of a guard to ensure that Auth context is available.
+ * This is currently necessary because we do not currently have a global guard configured for Auth,
+ * therefore the Auth context is not guaranteed to be available in the guard.
```
Cliftonz, line 37: What would it take to switch this to a guard? and should we?

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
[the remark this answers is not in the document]
djabarovgeorge: I think we should add a client rate limit with a separate bucket as Richard said in order to guard our system from things like:
* malicious actors that could retrieve the jwt token and automate requests.
* misconfigured systems - could be someone that implements the notification center (jwt token user), which could cause high-traffic requests by bug or misconfigured.

there could be potential other cases.

But this should not be part of this PR/Cycle, we could benefit a lot from the first stage of the rate limit and prioritize the next stage accordingly.

Cliftonz (approved):

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
rifont: After making these changes, we can remove a considerable number of individual controller-level and method-level use of the AuthGuard, providing better maintainability.

Given the sizeable scope and many controllers/methods touched, I think we should make these change in a separate PR.

Cliftonz: I agree, Can you please create a set of tasks to do this in linear so that way I can prioritize this follow-up tasks?

PR #4918, opened 2023-11-30
refactor: do not perform fetchsockets when service is shutting down

djabarovgeorge:

When service is shutting down don't perform a look for the active sockets connections. 

When the server shuts down we have a spike of disconnects.

### DIFF
--- a/apps/ws/src/socket/ws.gateway.ts
+++ b/apps/ws/src/socket/ws.gateway.ts
@@ -94,8 +94,7 @@ export class WSGateway implements OnGatewayConnection, OnGatewayDisconnect {
       return;
     }
 
-    const activeConnections = await this.getActiveConnections(connection, subscriber._id);
-    await this.subscriberOnlineService.handleDisconnection(subscriber, activeConnections);
+    await this.subscriberOnlineService.handleDisconnection(subscriber);
   }
 
   private async processConnectionRequest(connection: Socket) {
@@ -127,11 +126,5 @@ export class WSGateway implements OnGatewayConnection, OnGatewayDisconnect {
 
   private disconnect(socket: Socket) {
     socket.disconnect();
-  }
-
-  private async getActiveConnections(socket: Socket, subscriberId: string) {
-    const activeSockets = await socket.in(subscriberId).fetchSockets();
-
-    return activeSockets.length;
   }
 }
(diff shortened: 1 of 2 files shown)


packages/application-generic/src/services/workers/worker-base.service.ts
```diff
@@ -18,7 +19,7 @@ export type WorkerProcessor =
 
 export { WorkerOptions };
 
-export class WorkerBaseService {
+export class WorkerBaseService implements INovuWorker {
```
djabarovgeorge, line 22: This class is the base for all of the workers meaning it should implement the worker interface more than that he is responsible for implementing it.
so I see no reason for the duplication in all of the worker child classes.

apps/worker/src/app/workflow/services/execution-log.worker.ts
```diff
@@ -16,7 +15,7 @@ import { ObservabilityBackgroundTransactionEnum } from '@novu/shared';
 const LOG_CONTEXT = 'ExecutionLogWorker';
 
 @Injectable()
-export class ExecutionLogWorker extends ExecutionLogWorkerService implements INovuWorker {
+export class ExecutionLogWorker extends ExecutionLogWorkerService {
```
djabarovgeorge, line 18: moved the INovuWorker to the base class.

apps/ws/src/socket/usecases/external-services-route/external-services-route.usecase.ts
djabarovgeorge, line 1: Redundant call to the redis we already check the connection exists in line 17 and skip the flow if it is missing in line 20.

apps/ws/src/socket/ws.gateway.ts
```diff
@@ -129,9 +129,28 @@ export class WSGateway implements OnGatewayConnection, OnGatewayDisconnect {
...
+        Logger.error('WS server was not initialized while executing shutdown', LOG_CONTEXT);
+
+        return;
+      }
+
+      Logger.log('Closing WS server for incoming new connections', LOG_CONTEXT);
+      this.server.close();
+
+      Logger.log('Disconnecting active sockets connections', LOG_CONTEXT);
+      this.server.sockets.disconnectSockets();
```
djabarovgeorge, line 144: At the moment I have not implemented disconnection with jitter due to a lack of data, i think we should make a small discovery on how many connections we have and how we want to split them unless someone is familiar with general best practices that will fit our needs that we could implement here.

packages/application-generic/src/modules/interfaces.ts
djabarovgeorge, line 1: I already raised with Richard that I think that in my opinion, we should implement this interface in all of the modules in case we need to dispose of available resources.

LetItRock (approved):
🥇

PR #4936, opened 2023-12-04
feat: move in app preview render to backend

davidsoderberg:
Screenshot 2023-12-04 at 09 12 16


ainouzgali (commented):
@davidsoderberg can we re-use this new usecase from the send-message-in-app that is in the worker?

davidsoderberg:
> @davidsoderberg can we re-use this new usecase from the send-message-in-app that is in the worker?

Yes we can, great catch will refactor so it is used in that one as well.

ainouzgali (approved):

PR #4945, opened 2023-12-05
perf(worker): dont refetch subscriber for global preferences

scopsy:

Reduce a subscriber query global pereferences


ainouzgali (approved):

scopsy:
@ainouzgali added another change 😅 We can reduce an update call round trip to the db here

djabarovgeorge (approved):
Looks good, 
small note what do you think about combining lines 40 41 and the `if (canceled)` logic into a function?

PR #4946, opened 2023-12-05
Add execution log on invalid delay configuration

### LINKED ISSUE #4748: [NV-3126] Novu should fail delay error gracefully
### What?

The delay step accepts the scheduled time variable

### Why? (Context)

If the user sets this variable a past date time, then novu fails the step and stop the execution. This error is not logged in activity feed logs.

### Definition of Done

Delay errors should fail gracefully with logs in activity feed

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-3126](https://linear.app/novu/issue/NV-3126/novu-should-fail-delay-error-gracefully)</sub>

djabarovgeorge:

The delay step accepts the scheduled time variable

If the user sets this variable a past date time, then novu fails the step and stop the execution. This error is not logged in activity feed logs.

#### Definition of Done

Delay errors should fail gracefully with logs in activity feed
Closes #4748 

image

image


packages/application-generic/src/usecases/add-job/add-delay-job.usecase.ts
```diff
@@ -27,16 +40,47 @@ export class AddDelayJob {
...
+    let delay;
+
+    try {
+      delay = this.calculateDelayService.calculateDelay({
+        stepMetadata: data.step.metadata,
+        payload: data.payload,
+        overrides: data.overrides,
+      });
+
+      await this.jobRepository.updateStatus(
```
djabarovgeorge, line 52: update as delayed only if calculation was successful

packages/application-generic/src/usecases/add-job/add-delay-job.usecase.ts
```diff
@@ -27,16 +40,47 @@ export class AddDelayJob {
...
+          source: ExecutionDetailsSourceEnum.INTERNAL,
+          status: ExecutionDetailsStatusEnum.FAILED,
+          isTest: false,
+          isRetry: false,
+          raw: JSON.stringify({ error: error.message }),
+        }),
+        command.organizationId
+      );
+
+      await this.jobRepository.updateStatus(
```
djabarovgeorge, line 75: If the calculation was unsuccessful update the job as cancelled

packages/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
djabarovgeorge, line 1: This one is a real pain, we were unable to create a job because we got an exception while creating the expire at.
Let me know what you think about the workaround, i don't like it but i am not sure what would be a better way to execute the execution log creation.

djabarovgeorge:
@jainpawan21 Would love to hear your opinion on the matter, do you think this execution log will help understand what happened?

packages/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -162,22 +162,30 @@ export class CreateNotificationJobs {
...
+        )
+        .reduce((sum, delayAmount) => sum + delayAmount, 0);
+
+      return addMilliseconds(Date.now(), delay);
+    } catch (e) {
+      /*
+       * If the user has entered an incorrect negative delay,
+       * we'll accept it as a temporary solution to enable printing error execution details later in the process when a job is available.
+       */
+      return Date.now();
```
LetItRock, line 188: won't this be a problem if we expire the entity before it gets executed? like I'm not sure how the MongoDB Atlas will behave in this kind of situation, would it archive it right away?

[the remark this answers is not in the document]
djabarovgeorge: As Gali said, we have a fixed time we add to expire, here we make sure to add the delay/digest about of time.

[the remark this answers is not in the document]
LetItRock: Maybe we can just return 1 month from now for safety because I'm not sure if that won't override the default on schema `expireAt`.

jainpawan21:
> @jainpawan21 Would love to hear your opinion on the matter, do you think this execution log will help understand what happened?

Thanks George. Its helpful for user to debug now

ainouzgali (approved):
💃🏼

PR #4952, opened 2023-12-06
Add support for cid based attachment in email

djabarovgeorge:

CID attachment is used for inline image embedding

Example:- 

To show novu logo as inline embed in html

```
<html>
  <body>
    <p>Before image</p>
    <img src="cid:novu-logo" alt="Novu logo">
    <p>After image</p>
  </body>
</html>
```
 
Sendgrid attachment example code:-

```
{
  "content": "imageFileBase64",
  "filename": "index.jpeg",
  "type": "image/jpeg",
  "disposition": "inline",
  "content_id": "novu-logo"
}
```

### Why? (Context)

Novu has 3 fields for attachments currently:-

```
{
   file: "base64 of file",
   name: 'file_name.jpeg',
   mime: 'image/jpeg',
}
```

Definition of Done

Add new fields to support CID based attachments

### DIFF
--- a/providers/sendgrid/src/lib/sendgrid.provider.ts
+++ b/providers/sendgrid/src/lib/sendgrid.provider.ts
@@ -7,9 +7,12 @@ import {
   ICheckIntegrationResponse,
   CheckIntegrationResponseEnum,
   IEmailEventBody,
+  IAttachmentOptions,
 } from '@novu/stateless';
 
 import { MailDataRequired, MailService } from '@sendgrid/mail';
+
+type AttachmentJSON = MailDataRequired['attachments'][0];
 
 export class SendgridEmailProvider implements IEmailProvider {
   id = 'sendgrid';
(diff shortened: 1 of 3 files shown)


providers/sendgrid/src/lib/sendgrid.provider.ts
```diff
@@ -7,10 +7,13 @@ import {
   ICheckIntegrationResponse,
   CheckIntegrationResponseEnum,
   IEmailEventBody,
+  IAttachmentOptions,
 } from '@novu/stateless';
 
 import { MailDataRequired, MailService } from '@sendgrid/mail';
 
+type AttachmentJSON = MailDataRequired['attachments'][0];
```
djabarovgeorge, line 15: No idea why but the AttachmentJSON is not exported from '@sendgrid/mail', created this workaround so we won't use any in the code.

providers/sendgrid/src/lib/sendgrid.provider.ts
```diff
@@ -74,6 +77,26 @@ export class SendgridEmailProvider implements IEmailProvider {
     delete options.customData?.dynamicTemplateData;
     delete options.customData?.templateId;
 
+    const attachments = options.attachments?.map(
+      (attachment: IAttachmentOptions) => {
+        const attachmentJson: Partial<AttachmentJSON> = {
```
LetItRock, line 82: here you have `Partial<AttachmentJSON>` but at the end you do `as AttachmentJSON`, so maybe define it from the beginning like that, or you have some issues?

LetItRock (approved):
⭐

providers/sendgrid/src/lib/sendgrid.provider.ts
djabarovgeorge: Good point i think I had some issues before we created the local `type AttachmentJSON`

PR #4953, opened 2023-12-06
Add subscriber source to metric

djabarovgeorge:

Add subscriber source to metric.

So we would have the ability to separate the sending sources.

The primary aim was to grant us the capability to distinguish trigger sources without increasing the number of metrics/logs or generating additional computational burden through iteration over subscriber arrays.


libs/shared/src/types/subscriber/index.ts
```diff
@@ -19,6 +19,16 @@ export interface ISubscribersDefine extends ISubscriberPayload {
   subscriberId: string;
 }
 
+export interface ISubscribersSource extends ISubscribersDefine {
+  _subscriberSource: SubscriberSourceEnum;
+}
+
+export enum SubscriberSourceEnum {
+  BROADCAST = 'broadcast',
+  SINGLE = 'single',
```
djabarovgeorge, line 28: Was not sure about the naming here, not on the enum and not on the `single`.

Cliftonz: I think this is good

packages/application-generic/src/usecases/map-trigger-recipients/map-trigger-recipients.use-case.ts
```diff
@@ -159,8 +165,33 @@ export class MapTriggerRecipients {
...
+  public mapSubscriber(
+    subscriber: TriggerRecipientSubscriber,
+    source: SubscriberSourceEnum
+  ): ISubscribersSource | null {
+    if (!subscriber) return null;
+
+    let mappedSubscriber: Partial<ISubscribersSource>;
+
+    if (typeof subscriber === 'string') {
+      mappedSubscriber = { subscriberId: subscriber };
```
rifont, line 177: This runtime type check here is a little uneasy. We should really only have a single object type.

Is this due to legacy tech debt? Is it easy to fix?

packages/application-generic/src/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
```diff
@@ -92,6 +93,7 @@ export class SubscriberJobBound {
         _organization: command.organizationId,
         channels: template?.steps.map((step) => step.template?.type),
         source: command.payload.__source || 'api',
+        subscriberSource: _subscriberSource || 'undefined',
```
rifont, line 96: What is the case for the `subscriberSource` being undefined?

The value of `undefined` doesn't currently align with the enum value options. If we really need a falsy value, I think it's better to be more declarative with `null`.

libs/shared/src/types/subscriber/index.ts
rifont: I agree the naming makes sense, it aligns with Rate Limting nomenclature.

packages/application-generic/src/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
djabarovgeorge: This should never happen at the current logic, but in the future, you never know so I added it in the backoff, null makes sense to me :)

packages/application-generic/src/usecases/map-trigger-recipients/map-trigger-recipients.use-case.ts
djabarovgeorge: Correct this is a legacy logic in order to align the subscriber interface.
Please note that at this addition code, we did not add any additional iterations/checks.
We slitted `mapSubscriber` to two functions of `mapActor` and `mapSubscriber` so that for `mapActor` we do not add `_subscriberSource` because it does not make sense, and for `mapSubscriber` we add `_subscriberSource`.

rifont (approved):
LGTM! Great work 👍

apps/api/src/app/events/e2e/map-trigger-recipients.e2e.ts
```diff
@@ -569,19 +582,21 @@ describe('MapTriggerRecipientsUseCase', () => {
...
+        secondSubscriber.subscriberId,
         firstTopicRecipient,
         secondRecipient,
-        thirdSubscriber._id,
+        thirdSubscriber.subscriberId,
       ]);
       const result = await useCase.execute(command);
 
-      // We process first recipients that are not topics so they will take precedence when deduplicating
+      expect(result.length).to.equal(3);
```
djabarovgeorge, line 593: Looks like this test was not testing the reduplicating at all because the `to.include.deep.members` validated that the provided set exist in the expected object `result`, which was true but the result list was with other subscribers as well.
I am not sure if this was the expected behavior or not, but IMO the test should have tested the duplication `filtering` .

PR #4963, opened 2023-12-10
feat: add queue and workers dto typed

djabarovgeorge:

At the moment we do not validate the data we pass queue services for example SubscriberProcessQueueService.

Suggestions:
What we could do is create a middleware function 'add' in the child class, for example, SubscriberProcessQueueService with command and after then pass it to the parent 'add'.

We could provide an invalid object to the queue services by mistake.

### Other information and DOD's (Screenshots)

The remaining things that still need to be done:
* Refactor the job data object to be defined, at the moment we allow to pass `undefined` data. IMO it was made in order to create metric cron jobs that do not pass any data, however, this option adds complexity to the code and could lead to a failure point where someone won't pass the data object. Instead, we could add minimal data on the metric identifier.
* Make a discovery in order to make the base queue service move generic by getting the type from its child classes, this way we won't need to duplicate the methods in the services. in order to do so we will need to remove the mapping of the bull mq in the queue service and move it to the bull mq service.
* Refactor so that all of the workers will execute the creation of the Command. was not sure if it should be done in this PR or not, because it could be a breaking change.
* Align all of DTO's params to use align internal conventions, meaning if it is novu internal id it should have a prefix of `_` at the moment we sometimes use it and sometimes not. for example userId _userId. This will be a breaking change that will require backward combability support so I'd prefer to do it in separate dedicated PR.

### DIFF
--- a/apps/api/src/app/inbound-parse/usecases/inbound-email-parse/inbound-email-parse.command.ts
+++ b/apps/api/src/app/inbound-parse/usecases/inbound-email-parse/inbound-email-parse.command.ts
@@ -1,7 +1,16 @@
 import { IsDefined, IsNumber, IsOptional, IsString } from 'class-validator';
-import { BaseCommand } from '@novu/application-generic';
+import {
+  BaseCommand,
+  IConnection,
+  IEnvelopeFrom,
+  IEnvelopeTo,
+  IFrom,
+  IHeaders,
+  IInboundParseDataDto,
+  ITo,
+} from '@novu/application-generic';
 
-export class InboundEmailParseCommand extends BaseCommand {
+export class InboundEmailParseCommand extends BaseCommand implements IInboundParseDataDto {
   @IsDefined()
   @IsString()
   html: string;
@@ -65,71 +74,4 @@ export class InboundEmailParseCommand extends BaseCommand {
 
   @IsDefined()
   envelopeTo: IEnvelopeTo[];
-}
-
-export interface IHeaders {
-  'content-type': string;
-  from: string;
-  to: string;
-  subject: string;
-  'message-id': string;
-  date: string;
-  'mime-version': string;
-}
-
-export interface IFrom {
-  address: string;
-  name: string;
-}
-
-export interface ITo {
-  address: string;
-  name: string;
-}
-
-export interface ITlsOptions {
-  name: string;
-  standardName: string;
-  version: string;
-}
-
-export interface IMailFrom {
-  address: string;
-  args: boolean;
-}
-
-export interface IRcptTo {
-  address: string;
-  args: boolean;
-}
-
-export interface IEnvelope {
-  mailFrom: IMailFrom;
-  rcptTo: IRcptTo[];
-}
-
-export interface IConnection {
-  id: string;
-  remoteAddress: string;
-  remotePort: number;
-  clientHostname: string;
-  openingCommand: string;
-  hostNameAppearsAs: string;
-  xClient: any;
-  xForward: any;
-  transmissionType: string;
-  tlsOptions: ITlsOptions;
-  envelope: IEnvelope;
-  transaction: number;
-  mailPath: string;
-}
-
-export interface IEnvelopeFrom {
-  address: string;
-  args: boolean;
-}
-
-export interface IEnvelopeTo {
-  address: string;
-  args: boolean;
 }
--- a/apps/worker/src/app/workflow/services/standard.worker.ts
+++ b/apps/worker/src/app/workflow/services/standard.worker.ts
@@ -1,6 +1,7 @@
 const nr = require('newrelic');
 import { forwardRef, Inject, Injectable, Logger } from '@nestjs/common';
-import { IJobData, ObservabilityBackgroundTransactionEnum } from '@novu/shared';
+
+import { ObservabilityBackgroundTransactionEnum } from '@novu/shared';
 import {
   BullMqService,
   getStandardWorkerOptions,
@@ -24,6 +25,7 @@ import {
   HandleLastFailedJobCommand,
   HandleLastFailedJob,
 } from '../usecases';
+import { IStandardDataDto } from '@novu/application-generic/build/main/dtos/standard-job.dto';
 
 const LOG_CONTEXT = 'StandardWorker';
 
@@ -43,7 +45,7 @@ export class StandardWorker extends StandardWorkerService {
 
     this.initWorker(this.getWorkerProcessor(), this.getWorkerOptions());
 
-    this.worker.on('failed', async (job: Job<IJobData, void, string>, error: Error): Promise<void> => {
+    this.worker.on('failed', async (job: Job<IStandardDataDto, void, string>, error: Error): Promise<void> => {
       await this.jobHasFailed(job, error);
     });
   }
@@ -57,22 +59,26 @@ export class StandardWorker extends StandardWorkerService {
     };
   }
 
-  private extractMinimalJobData(job: any): {
+  private extractMinimalJobData(data: IStandardDataDto): {
     environmentId: string;
-    organizationId: string;
     jobId: string;
+    organizationId: string;
     userId: string;
   } {
-    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = job;
+    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = data;
 
     if (!environmentId || !jobId || !organizationId || !userId) {
-      const message = job.payload.message;
+      const message = data.payload?.message;
+
+      if (!message) {
+        throw new Error('Job data is missing required fields' + JSON.stringify(data));
+      }
 
       return {
         environmentId: message._environmentId,
         jobId: message._jobId,
         organizationId: message._organizationId,
-        userId: job.userId,
+        userId: userId,
       };
     }
 
@@ -85,7 +91,7 @@ export class StandardWorker extends StandardWorkerService {
   }
 
   private getWorkerProcessor() {
-    return async ({ data }: { data: IJobData | any }) => {
+    return async ({ data }: { data: IStandardDataDto }) => {
       const minimalJobData = this.extractMinimalJobData(data);
 
       Logger.verbose(`Job ${minimalJobData.jobId} is being processed in the new instance standard worker`, LOG_CONTEXT);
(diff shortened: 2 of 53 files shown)


apps/api/src/app/inbound-parse/usecases/inbound-email-parse/inbound-email-parse.command.ts
```diff
@@ -66,70 +75,3 @@ export class InboundEmailParseCommand extends BaseCommand {
   @IsDefined()
   envelopeTo: IEnvelopeTo[];
 }
-
-export interface IHeaders {
```
djabarovgeorge, line 70: Extracted to application-generic so we could reuse it.

apps/inbound-mail/src/server/inbound-mail.service.spec.ts
```diff
@@ -63,8 +63,9 @@ describe('Inbound Mail Service', () => {
         _environmentId,
         _organizationId,
         _userId,
-      };
-      await inboundMailService.inboundParseQueueService.add(jobId, jobData, _organizationId);
+      } as any;
```
djabarovgeorge, line 66: For the test, IInboundParseDataDto contained a lot of attributes that needed to be added. for now, added casting as those attributes were not tested.

apps/worker/src/app/workflow/services/standard.worker.ts
```diff
@@ -57,22 +59,26 @@ export class StandardWorker extends StandardWorkerService {
...
     jobId: string;
+    organizationId: string;
     userId: string;
   } {
-    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = job;
+    const { _environmentId: environmentId, _id: jobId, _organizationId: organizationId, _userId: userId } = data;
 
     if (!environmentId || !jobId || !organizationId || !userId) {
-      const message = job.payload.message;
+      const message = data.payload?.message;
```
djabarovgeorge, line 71: This object is structure data.payload?.message deprecated, i could not locate any place in the code that passes this data.

packages/application-generic/src/dtos/process-subscriber-job.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+  subscriber: ISubscribersDefine;
+  templateId: string;
+}
+
+export interface IProcessSubscriberJobDto extends IJobParams {
+  data?: IProcessSubscriberDataDto;
+}
+
+export interface IProcessSubscriberBulkJobDto extends IBulkJobParams {
+  data: IProcessSubscriberDataDto;
```
djabarovgeorge, line 28: This is part of the complexity i was talking about in the issue description, due to the metrics add execution allowing undefined data objects we need to create two separate interfaces because we do not want to allow undefined in the data as much as possible e.x the bulk DTO.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -66,56 +66,64 @@ export class QueueBaseService {
...
     await this.instance.addBulk(data);
   }
 
   async onModuleDestroy(): Promise<void> {
     await this.gracefulShutdown();
   }
 }
+
+export interface IJobParams {
+  name: string;
```
djabarovgeorge, line 118: Changed the `id` to `name` as it is the perm that is used in pull mq so i wanted to align with it. Same for IBulkJobParams.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -66,56 +66,64 @@ export class QueueBaseService {
...
+  groupId?: string;
+  options?: JobsOptions;
+}
+
+export interface IBulkJobParams {
+  name: string;
+  data: any;
+  groupId?: string;
+  options?: BulkJobOptions;
+}
```
djabarovgeorge, line 129: This is part of the complexity i was talking about in the issue description, due to the metrics add execution allowing undefined data objects we need to create two separate interfaces because we do not want to allow undefined in the data as much as possible e.x the bulk DTO.

* We need to find a better place for those interfaces.

packages/application-generic/src/services/queues/standard-queue.service.ts
```diff
@@ -21,4 +22,18 @@ export class StandardQueueService extends QueueBaseService {
...
+
+  public async add(data: unknown) {
+    // in order to implement we need to know what should be the data dto first
+    throw new Error('Not implemented');
+  }
+
+  public async addBulk(data: unknown[]) {
+    // in order to implement we need to know what should be the data dto first
+    throw new Error('Not implemented');
+  }
```
djabarovgeorge, line 38: We do not use those methods at the moment, therefore i can not tell what should be the DTO here, i wonder if we should keep the exception or allow any here.

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -12,14 +12,14 @@ import {
...
-  public instance: BullMqService;
+  private instance: BullMqService;
 
   public readonly DEFAULT_ATTEMPTS = 3;
   public queue: Queue;
 
   constructor(
     public readonly topic: JobTopicNameEnum,
-    public bullMqService: BullMqService
+    private bullMqService: BullMqService
```
djabarovgeorge, line 22: This poses a major problem. Exposing the provider used for delegation means our provider becomes vulnerable to errors as it's visible externally. Additionally, tying our queue service to a specific provider's behavior will complicate switching, for instance, from Kafka to Pulsar.

Also, it's important to keep the 'queue' attribute private for now to maintain simplicity in this PR.

apps/worker/src/app/workflow/services/standard.worker.ts
```diff
@@ -24,6 +25,7 @@ import {
   HandleLastFailedJobCommand,
   HandleLastFailedJob,
 } from '../usecases';
+import { IStandardDataDto } from '@novu/application-generic/build/main/dtos/standard-job.dto';
```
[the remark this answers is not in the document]
djabarovgeorge: Good catch!  I was missing an export.

BiswaViraj (approved):
🚀

packages/application-generic/src/services/queues/queue-base.service.ts
```diff
@@ -66,56 +86,64 @@ export class QueueBaseService {
...
+    await this.instance.add(
+      name,
+      jobData,
+      {
+        removeOnComplete: true,
+        removeOnFail: true,
+        ...options,
+      },
+      groupId
+    );
```
LetItRock, line 110: why we are not using the `add` function from this class?

packages/application-generic/src/services/queues/standard-queue.service.ts
```diff
@@ -21,4 +22,18 @@ export class StandardQueueService extends QueueBaseService {
 
     this.createQueue();
   }
+
+  public async addMinimalJob(data: IStandardJobDto) {
+    return await super.addMinimalJob(data);
+  }
```
LetItRock, line 28: Do we need this overridden function if it's just calling the super? If you want to force the `data` type here I think it's better to do with generics on the `QueueBaseService`. Then we won't need to do everywhere: 
```
public async add(data: T) {
    return await super.add(data);
  }

  public async addBulk(data: T[]) {
    return await super.addBulk(data);
  }
```

LetItRock, line 28: the `addMinimalJob` is specific just for the `standard queue` do we need to have it in the base?

packages/application-generic/src/services/queues/standard-queue.service.ts
LetItRock: The `add` function should be deprecated for the `standard queue` right?

apps/worker/src/app/workflow/services/standard.worker.ts
djabarovgeorge: Hard to say, I am okay with keeping it like that for now, but create a ticket to remove this redundant code so it won't stick with us and confuse the developers in the future.

* i will create a ticket rn

packages/application-generic/src/services/queues/standard-queue.service.ts
djabarovgeorge: Are you saying we should delete it? In my view, we should get rid of addMinimalJob and all its related stuff like Worker.extractMinimalJobData. Instead,  just use the add function in the standard queue with the minimal data.

I didn't change it because I made this PR mainly to add DTO types in the code. So, the PR focuses on adding the necessary logic for these types without messing with the main logic.

packages/application-generic/src/services/queues/standard-queue.service.ts
djabarovgeorge: Good point i will remove it from the base.

apps/ws/src/socket/services/web-socket.worker.ts
```diff
@@ -41,14 +42,16 @@ export class WebSocketWorker extends WebSocketsWorkerService {
           'WS Service',
           function () {
             const transaction = nr.getTransaction();
+            const { data: jobData } = job;
```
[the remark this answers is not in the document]
djabarovgeorge: Good point, however, these two id's aren't connected. Adding specific types helps open our eyes to better understand the types that are in the field. However, our current solution isn't completely typed.

To clarify, the 'name' serves as the Job's name, but it's relatively redundant now. It was initially used in the 'bull' library to distribute jobs among workers. However, 'bullmq' has deprecated this to avoid confusion. [more info here](https://github.com/taskforcesh/bullmq/issues/297#issuecomment-715406744
).

The job ID you mentioned is an internal ID used by 'bullmq'. To assign it, in order to override it we need to pass it as 'option.jobId' in the third 'add' parameter [more about it here](https://docs.bullmq.io/guide/jobs/job-ids).

packages/application-generic/src/services/queues/queue-base.service.ts
djabarovgeorge: At this state, we could not call this.add because for StandardQueueService it is not implemented.

However, due to other comments you made I will remove the addMinimalJob from base.

packages/application-generic/src/services/queues/standard-queue.service.ts
djabarovgeorge: It will take more time in order to add those methods typed due to the BullMqService.add method restrictions, so for this PR scope I would suggest keeping it like this. wdyt we can pospond it?

djabarovgeorge: In the end, i removed it altogether. Now we will use add with addMinimalJob data dto.

LetItRock (approved):
🚀

PR #4972, opened 2023-12-11
fix(api): ApiKey auth guard performance

rifont:
* Update auth service to return expected auth validation User entity
* Authentication strategy
  * Add passport apikey strategy
  * Remove redundant logger assign in jwt strategy
* Guards/decorator tidy-up - removing redundant JWT check
  * `Roles` guard
  * `RootEnvironment` guard
  * `UserSession` decorator
* Use req.user payload in:
  * throttler guard
  * idempotency interceptors
* Remove redundant logger.assign in trigger use-case

API requests using ApiKey authentication currently sign and decode a JWT token every request. The signing introduces considerable latency whilst the decoding adds unnecessary memory consumption.

We also needed an interface to retrieve User context from the request, regardless of auth scheme.

_Auth Guard Tests_
image

_ApiKey Auth Log_
image

_Bearer Auth Log_
image

_No Auth Log_
image


apps/api/src/app/auth/auth.module.ts
```diff
@@ -18,8 +18,9 @@ import { EnvironmentsModule } from '../environments/environments.module';
 import { JwtSubscriberStrategy } from './services/passport/subscriber-jwt.strategy';
 import { JwtAuthGuard } from './framework/auth.guard';
 import { RootEnvironmentGuard } from './framework/root-environment-guard.service';
+import { ApiKeyStrategy } from './services/passport/apikey.strategy';
 
-const AUTH_STRATEGIES: Provider[] = [];
+const AUTH_STRATEGIES: Provider[] = [JwtStrategy, ApiKeyStrategy, JwtSubscriberStrategy];
```
rifont, line 23: Moving the passport strategies out to where they belong.

apps/api/src/app/auth/framework/roles.guard.ts
```diff
@@ -13,18 +13,6 @@ export class RolesGuard implements CanActivate {
...
-
-    const token = request.headers.authorization.split(' ')[1];
-    if (!token) return false;
-
-    const authorizationHeader = request.headers.authorization;
-    if (!authorizationHeader?.includes('ApiKey')) {
-      const user = jwt.decode(token) as IJwtPayload;
-      if (!user) return false;
-    }
-
```
rifont, line 27: This is handled in the Authentication flow, and shouldn't be part of Authorization.

apps/api/src/app/auth/framework/root-environment-guard.service.ts
```diff
@@ -13,14 +13,7 @@ export class RootEnvironmentGuard implements CanActivate {
...
   async canActivate(context: ExecutionContext) {
     const request = context.switchToHttp().getRequest();
-    if (!request.headers.authorization) return false;
-
-    const token = request.headers.authorization.split(' ')[1];
-    if (!token) return false;
-
-    const user = jwt.decode(token) as IJwtPayload;
-    if (!user) return false;
-    if (!user.environmentId) return false;
```
rifont, line 23: As above, this is handled in the Authentication flow, and shouldn't be part of Authorization.

apps/api/src/app/auth/services/passport/jwt.strategy.ts
```diff
@@ -20,12 +20,6 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
       throw new UnauthorizedException();
     }
 
-    this.logger.assign({
-      userId: user._id,
-      environmentId: payload.environmentId,
-      organizationId: payload.organizationId,
-    });
-
```
rifont, line 28: This is now handled in a single place in the Auth Guard, for all Auth strategies.

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -165,18 +164,17 @@ export class ApiRateLimitInterceptor extends ThrottlerGuard implements NestInter
...
-    return ALLOWED_AUTH_SCHEMES.some((scheme) => req.authScheme === scheme);
+    return ALLOWED_AUTH_SCHEMES.some((scheme) => authScheme === scheme);
   }
 
   private getReqUser(context: ExecutionContext): IJwtPayload {
     const req = context.switchToHttp().getRequest();
-    const token = req.headers[RateLimitHeaderKeysEnum.AUTHORIZATION.toLowerCase()].split(' ')[1];
 
-    return jwt.decode(token);
+    return req.user;
```
rifont, line 178: `req.user` is always present now.

apps/api/src/app/shared/framework/idempotency.interceptor.ts
```diff
@@ -93,26 +92,22 @@ export class IdempotencyInterceptor implements NestInterceptor {
...
-    }
-    if (req.headers?.authorization?.length) {
-      const token = req.headers.authorization.split(' ')[1];
-      if (token) {
-        return jwt.decode(token);
-      }
-    }
 
-    return null;
+    return req.user;
```
rifont, line 98: As above, `req.user` is always present now.

packages/application-generic/src/decorators/user-session.decorator.ts
```diff
@@ -11,28 +14,12 @@ export const UserSession = createParamDecorator((data, ctx) => {
...
-        throw new UnauthorizedException('bad_token');
-      if (!tokenParts[1]) throw new UnauthorizedException('bad_token');
-
-      const user = jwt.decode(tokenParts[1]);
-
-      return user;
-    }
-  }
-
-  return null;
```
rifont, line 37: As above, this is handled in the Authentication flow, and shouldn't be part of Authorization.

If we don't have a `req.user`, this is indicative of a developer error, and results in a `500 Server error`

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -1,54 +1,75 @@
...
-      if (!apiEnabled)
-        throw new UnauthorizedException('API endpoint not available');
+        return {
+          session: false,
+          defaultStrategy: 'headerapikey',
+        };
+      default:
+        throw new UnauthorizedException('Invalid auth scheme');
+    }
+  }
```
rifont, line 48: Handling invocation different authentication schemes passport strategies with an authScheme switch statement

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -196,32 +193,30 @@ export class AuthService {
   }
 
   @Instrument()
-  async apiKeyAuthenticate(apiKey: string) {
-    const { environment, user, key, error } = await this.getUserData({
-      apiKey,
-    });
+  public async validateApiKey(apiKey: string): Promise<IJwtPayload> {
```
rifont, line 196: Making sure the ApiKey validation returns the same interface as Bearer authentication.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -242,33 +237,7 @@ export class AuthService {
...
-        environmentId,
-      },
-      {
-        expiresIn: '1 day',
-        issuer: 'novu_api',
-        audience: 'api_token',
-      }
-    );
-  }
-
```
rifont, line 270: Not used.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -397,17 +358,6 @@ export class AuthService {
...
-  @Instrument()
-  @CachedEntity({
-    builder: ({ apiKey }: { apiKey: string }) =>
-      buildEnvironmentByApiKey({
-        apiKey: apiKey,
-      }),
-  })
-  private async getEnvironment({ apiKey }: { apiKey: string }) {
-    return await this.environmentRepository.findByApiKey(apiKey);
-  }
```
rifont, line 409: Not used.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -369,15 +338,7 @@ export class AuthService {
     });
   }
 
-  async decodeJwt<T>(token: string) {
-    return this.jwtService.decode(token) as T;
-  }
-
-  async verifyJwt(jwt: string) {
-    return this.jwtService.verify(jwt);
-  }
```
rifont, line 378: Not used.

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -196,32 +193,30 @@ export class AuthService {
...
+    const {
+      environment,
+      user,
+      error,
+      key, // In the future, roles/scopes will be assigned to the API Key.
+    } = await this.getApiKeyUser(apiKey);
 
     if (error) throw new UnauthorizedException(error);
-    if (!environment) throw new UnauthorizedException('API Key not found');
-    if (!user) throw new UnauthorizedException('User not found');
```
rifont, line 206: The `if(error)` check above already handles these exceptions.

apps/api/src/app/auth/services/passport/apikey.strategy.ts
```diff
@@ -0,0 +1,24 @@
...
+import { IJwtPayload } from '@novu/shared';
+
+@Injectable()
+export class ApiKeyStrategy extends PassportStrategy(HeaderAPIKeyStrategy) {
+  constructor(private readonly authService: AuthService) {
+    super(
+      { header: 'Authorization', prefix: 'ApiKey ' },
+      true,
+      (apikey: string, verified: (err: Error | null, user?: IJwtPayload | false) => void) => {
+        this.authService.validateApiKey(apikey).then((user) => {
```
scopsy, line 14: Should we add a catch here?

LetItRock (approved):
looks good to me 💎

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -1,54 +1,76 @@
...
+      userId: user._id,
+      environmentId: user.environmentId,
+      organizationId: user.organizationId,
+    });
 
-        return true;
-      });
-    }
+    /**
+     * This helps with sentry and other tools that need to know who the user is based on `id` property.
```
djabarovgeorge, line 68: 🚀

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -165,18 +164,14 @@ export class ApiRateLimitInterceptor extends ThrottlerGuard implements NestInter
...
   private isAllowedAuthScheme(context: ExecutionContext): boolean {
     const req = context.switchToHttp().getRequest();
-    const authorization = req.headers[RateLimitHeaderKeysEnum.AUTHORIZATION.toLowerCase()];
-    if (!authorization) {
-      return false;
-    }
+    const authScheme = req.authScheme;
 
-    return ALLOWED_AUTH_SCHEMES.some((scheme) => req.authScheme === scheme);
+    return ALLOWED_AUTH_SCHEMES.some((scheme) => authScheme === scheme);
```
djabarovgeorge, line 169: What do you think about creating an auth scheme enum so it would be more transparent what scheme we have, this way by looking at the allowed scheme list we could figure out that Bearer is not allowed just by looking at the code.

```
export enum AuthSchemeEnum {
  API_KEY = 'ApiKey',
  BEARER = 'Bearer',
  NONE = 'None',
}

export const ALLOWED_AUTH_SCHEMES = [AuthSchemeEnum.API_KEY];
```

djabarovgeorge (commented):
Love how much the code became cleaner and optimized 🤩

The main concern i had was regarding auth.guard.ts file and the defaultStrategy

apps/api/src/app/shared/framework/idempotency.interceptor.ts
```diff
@@ -93,26 +92,22 @@ export class IdempotencyInterceptor implements NestInterceptor {
...
   }
 
   private getCacheKey(context: ExecutionContext): string {
-    const { organizationId } = this.getReqUser(context) || {};
+    const user = this.getReqUser(context);
+    if (user === undefined) {
+      const message = 'Cannot build cache key without user';
```
rifont, line 104:
```suggestion
      const message = 'Cannot build idempotency cache key without user';
```

apps/api/src/app/auth/services/passport/apikey.strategy.ts
rifont: Good question! I'll add investigate whether an unhandled error here results in an `Unauthenticated` error and address accordingly.

packages/application-generic/src/usecases/trigger-event/trigger-event.usecase.ts
```diff
@@ -84,8 +84,6 @@ export class TriggerEvent {
 
       this.logger.assign({
         transactionId: mappedCommand.transactionId,
-        environmentId: mappedCommand.environmentId,
-        organizationId: mappedCommand.organizationId,
```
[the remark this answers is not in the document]
rifont: Good catch! I'll revert this change.

packages/application-generic/src/services/auth/auth.guard.ts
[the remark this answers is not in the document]
rifont: Great point. The name of this auth guard is not that representative of what we're trying to achieve here.

Instead of `JwtAuthGuard`, we should really rename this to `UserAuthGuard`, as we're seeking to authenticate only Novu Users with this guard. The controller/endpoint the AuthGuard is placed, combined with the Passport Strategies accepted by the Auth guard determines which clients have access to that API controller/endpoint. The contract of the Auth guard with a given number of strategies is to return the Entity (e.g. User/Subscriber) relating to the authenticated client.

We have a [`SubscriberRouteGuard`](https://github.com/novuhq/novu/blob/next/apps/api/src/app/auth/framework/subscriber-route.guard.ts) which does not use Passport, is [instantiated as a global guard](https://github.com/novuhq/novu/blob/d78af0351244ded26a903a1d2e9c1ef3659a0eb0/apps/api/src/bootstrap.ts#L85) and has a corresponding `SubscriberRoute` decorator that is currently unused across the whole codebase - dead code.

Instead, we have the `JwtSubscriberStrategy` you mentioned, used with the hardcoded `subscriberJwt` in several locations in the [Widgets controller](https://github.com/novuhq/novu/blob/d78af0351244ded26a903a1d2e9c1ef3659a0eb0/apps/api/src/app/widgets/widgets.controller.ts#L102). We should extract `@UseGuards(AuthGuard('subscriberJwt'))` into a reusable guard as in this guard, and call it `SubscriberAuthGuard`.

I'll add this comment and the relevant action to the Subscriber Auth guard clean up ([NV-3281](https://linear.app/novu/issue/NV-3281/remove-subscriber-jwtdecode)).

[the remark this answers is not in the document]
rifont: I will also rename this guard to `UserAuthGuard`.

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -1,54 +1,76 @@
...
-        'external_api_accessible',
-        context.getHandler()
-      );
-      if (!apiEnabled)
-        throw new UnauthorizedException('API endpoint not available');
+    switch (authScheme) {
+      case 'Bearer':
+        return {
+          session: false,
+          defaultStrategy: 'jwt',
```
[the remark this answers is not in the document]
rifont: Also noteworthy - if you specify a list of strategies with length greater than 1, Passport will attempt to validate all strategies by default. This is useful to easily configure passport, but actually results in a performance hit whereby a DB-read is required for each strategy, so O(n) per strategy.

This implementation using `getAuthenticateOptions` ensures that only a single strategy is validated per request.

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
rifont: I considered this, however we have the `UserAuthGuard` (formerly `JwtAuthGuard`) living in `application-generic` right now, with strategies living in `api`. These should all really be in the same place.

My suggestion would be to bring the `UserAuthGuard` back to `api`, and have both the guard and strategies co-located. We can move them all at once in the future if we need common API packages as part of API microservice direction or otherwise.

Wdyt @djabarovgeorge ?

rifont: And yes, we 100% should create this enum!

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -1,54 +1,76 @@
...
+        throw new UnauthorizedException(
+          `Invalid authentication scheme: "${authScheme}"`
+        );
+    }
+  }
 
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      return this.authService.apiKeyAuthenticate(key).then<any>((result) => {
-        request.headers.authorization = `Bearer ${result}`;
+  handleRequest<TUser = IJwtPayload>(
```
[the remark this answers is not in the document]
rifont: As above, I'll investigate this.

rifont:
These changes broke a bunch of [`api-ee` tests](https://github.com/novuhq/novu/actions/runs/7167933910/job/19515110155?pr=4972#step:11:940), they are all returning a `422` error.

It may be a result of the Idempotency changes, but could be something else. Further investigation needed.

scopsy:
@rifont seems like also happening in non e2e tests aswell

apps/api/src/app/auth/services/passport/apikey.strategy.ts
rifont: I simulated an error on the `validateApiKey` method, unhandled errors are not managed internally by passport. I've added a `.catch`. As a side, my preference would have been to use async/await but the `passport-headerapikey` lib does not support this.

_Before with unhandled error_
image

_After with `.catch` handler, for simulated 500 error_
image

_After with `.catch` handle, for 400 auth error_

packages/application-generic/src/services/auth/auth.guard.ts
```diff
@@ -1,54 +1,76 @@
...
   Injectable,
+  Logger,
   UnauthorizedException,
 } from '@nestjs/common';
-import { AuthGuard } from '@nestjs/passport';
+import { AuthGuard, IAuthModuleOptions } from '@nestjs/passport';
 import { Reflector } from '@nestjs/core';
-import { AuthService } from './auth.service';
-import { Instrument } from '../../instrumentation';
+import { PinoLogger } from 'nestjs-pino';
```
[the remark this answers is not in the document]
rifont: I have removed the logger altogether, user context logging config is now handled separately.

packages/application-generic/src/services/auth/auth.guard.ts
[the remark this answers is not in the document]
rifont: I have updated to call the `super`.

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
rifont: I haven't consolidated auth strategies and guards in this PR, however I have created an `auth.types` file in the `shared` lib to contain the auth strategies and auth scheme enums. Everywhere is now updated to use those 🎉

rifont:
> These changes broke a bunch of [`api-ee` tests](https://github.com/novuhq/novu/actions/runs/7167933910/job/19515110155?pr=4972#step:11:940), they are all returning a `422` error.
> 
> It may be a result of the Idempotency changes, but could be something else. Further investigation needed.

I have ruled out Idempotency issues.

It appears that the throttler guard is processing a different `environmentId` than the one present in the User Test session context. I suspect this may be related to changes in the ApiKey part of the Auth service. Continuing investigation there
image

rifont:
> > These changes broke a bunch of [`api-ee` tests](https://github.com/novuhq/novu/actions/runs/7167933910/job/19515110155?pr=4972#step:11:940), they are all returning a `422` error.
> > It may be a result of the Idempotency changes, but could be something else. Further investigation needed.
> 
> I have ruled out Idempotency issues.
> 
> It appears that the throttler guard is processing a different `environmentId` than the one present in the User Test session context. I suspect this may be related to changes in the ApiKey part of the Auth service. Continuing investigation there image

Turns out this was a simple fix. I'd changed the signature of the `getApiKeyUser` method and hadn't updated the `CachedEntity` signature to match (see the [commit diff](https://github.com/novuhq/novu/pull/4972/commits/552c383642911858e202b62bdd95b0e0c9141687)). Making them match line fixed all the test issues. cc @djabarovgeorge , related to what we'd discussed but unfortunately Typescript doesn't support decorator typings right now in the way we need to, for us to make this typing strict.

rifont:
Added tests for both Bearer and ApiKey auth to make sure the auth logic and strategies are working as expected - https://github.com/novuhq/novu/pull/4972/commits/6ba5265f60040befd1e477a621ad30f1bc376b56

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
djabarovgeorge: > I considered this, however we have the `UserAuthGuard` (formerly `JwtAuthGuard`) living in `application-generic` right now, with strategies living in `api`. These should all really be in the same place.

Thanks for sharing your thoughts on the separation of the code, I agree that this guard's responsibility is to be used in the API domain but I believe that the JwtAuthGuard was moved to application-generic because we reuse it in the enterprise package controllers.

apps/api/src/app/auth/services/passport/apikey.strategy.ts
djabarovgeorge: i wanted to suggest an async-await pattern as well because i pretty much hate nested promises, but then i saw that they do not expect an async function :(

packages/application-generic/src/services/auth/auth.guard.ts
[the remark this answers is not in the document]
djabarovgeorge: > We should extract `@UseGuards(AuthGuard('subscriberJwt'))` into a reusable guard as in this guard, and call it `SubscriberAuthGuard`.

Sounds good to me! :)

PR #4983, opened 2023-12-13
New enterprise flow

djabarovgeorge:

Change the internal process for utilizing enterprise modules in our codebase.

So we could use the enterprise module.


Cliftonz:
@djabarovgeorge Can you add more context to this?

.source
BiswaViraj, line 1: why do we need this `.source` ?

djabarovgeorge: This is the new submodule link, instead of the old one 'enterprise'

.github/workflows/reusable-api-e2e.yml
```diff
@@ -27,6 +27,7 @@ on:
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   check_submodule_token:
+    if: ${{ inputs.ee }}
```
djabarovgeorge, line 30: Run this check only if the matrix name is `'novu/api-ee'` otherwise, it is not needed.

LetItRock (approved):
looks good to me ⭐

Issue #2813, opened 2023-02-16
[NV-1721] 🐛 Bug Report: Issues at scale. Subscriber tab lagging


alaboudi:
### 📜 Description

We have a lot of users in our business, in high 10s of millions, that we have hydrated into our self hosted Novu service / mongoDB. I have realized that the subscribers tab, in the lefthand side navigation, lags when clicked on at this scale

### 👟 Reproduction steps

1. Hydrate the Novu database with 50M+ users
2. Go to the subscribers tab 
3. Experience the 1 minute+ lag/crash 😆 

### 👍 Expected behavior

There should be no lag when visiting the subscribers tab

### 👎 Actual Behavior with Screenshots

<img width="1745" alt="image" src="https://user-images.githubusercontent.com/7456621/219496728-cd9a218a-9416-43cd-a2fe-ab0414c796ed.png">
Hangs like this for 5+ minutes

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

<sub>[NV-1721](https://linear.app/novu/issue/NV-1721/🐛-bug-report-issues-at-scale-subscriber-tab-lagging)</sub>

davidsoderberg:
test

davidsoderberg:
test

scopsy:
Related to #2099

SebastianStehle:
It is the count query. You need full index coverage here and I would add a dedicated collection for the count

scopsy:
@SebastianStehle what do you mean by a dedicated collection?

SebastianStehle:
Yes.

the problem is the size of the index. If it exceeds the memory limit you have to do calls to the disk again and I have never managed to get these queries really fast.

scopsy:
We can also limit the count for now, in case there are more than 100k (or any other) entries, just write: 100k+.

SebastianStehle:
How do you know that there are more than 100k without counting? ;)

scopsy:
@SebastianStehle from what I recall mongo have a `limit` property for `count` commands it will basically stop counting at the limit, and we can have the assumption if the count equals to the limit we can show `limit+` on the UI.

oba2311:
limit+ works IMO.

alaboudi:
[...]

PR #4985, opened 2023-12-13
fix(api): subscribers page performance

PaulFasola:

Our hosted Novu, housing tens of millions subscriber documents, underwent an upgrade from v13 to v21 recently. However, we encountered a persistent issue with our MongoDB instance being impacted by excessive hits, even after the deployment of v21. Although the problem was reported as fixed in version 17.1 (see #2813), it persisted in the subsequent release.

Upon investigation, it was discovered that an unused query in the `get-subscribers` command was the root cause. It was calling a - unused- count query that appears to be a remnant from a previous optimization effort on the pagination, that was modified to be non-totalCount based. As a nit but for coherency purposes, I also renamed the file names in that folder to match class names.

Our subscribers page is now loading, and very fast ✅ 

It ain't much, but it's honest work 🙂 

This change was needed to fix the subscribers page given a large-scale dataset.


scopsy:
@PaulFasola great catch! Thank you so much for the contribution

scopsy (approved):

scopsy:
@PaulFasola I think we missed an import with the refactoring of the file name. Could you take a look? 

```
> @novu/api@0.22.0 build /home/runner/work/novu/novu/apps/api
> nest build
ERROR in /home/runner/work/novu/novu/apps/api/src/app/subscribers/usecases/get-subscribers/get-subscribers.usecase.ts
3:38-64
[tsl] ERROR in /home/runner/work/novu/novu/apps/api/src/app/subscribers/usecases/get-subscribers/get-subscribers.usecase.ts(3,39)
      TS2[30](https://github.com/novuhq/novu/actions/runs/7207319249/job/19634025137?pr=4985#step:5:32)7: Cannot find module './get-subscriber.command' or its corresponding type declarations.
ts-loader-default_43d7ed5e51b1fe49
webpack 5.88.2 compiled with 1 error in 30238 ms
 ELIFECYCLE  Command failed with exit code 1.
```

PR #4986, opened 2023-12-15
Refactor move inbound mail parser to worker

djabarovgeorge:

Move the inbound mail parser worker to the worker app.

There is no reason for us to spin the inbound mail parser worker in the API app.


scopsy:
Just a note, it will require a terraform change aswell*

LetItRock (approved):
from the code perspective it looks good to me! 🔥 
@Cliftonz we would need to implement appropriate cloud infra changes after this PR is merged

Cliftonz:
@djabarovgeorge I see the spell check failing for the moment. 

however I want to confirm that now the workers will be processing the inbound mail and it will go like:

email -> dns -> inbound-email service-> queue -> email parse (on worker) -> standard queue

If this is true then we can merge this in and the general worker service should be able to handle this.

djabarovgeorge:
> email -> dns -> inbound-email service-> queue -> email parse (on worker) -> standard queue

@Cliftonz It will be even shorter
`email -> dns -> inbound-email service-> queue -> email parse (on worker)`
After we parse the email we respond user's webhook

Cliftonz:
@djabarovgeorge There are a few conflicts but I would like to get this merged in soon

djabarovgeorge:
> @djabarovgeorge There are a few conflicts but I would like to get this merged in soon

Tend to agree here 💪

PR #4991, opened 2023-12-17
refactor: extracted worker queue chunk size to env var

Khongchai:

Extract hard-coded queue size into env var.

To make it easier for people who self-host to optimize queue size.

### Other information

https://discord.com/channels/895029566685462578/1185606056005611642/1185861985808695336


Khongchai:
@scopsy

packages/application-generic/src/usecases/trigger-broadcast/trigger-broadcast.usecase.ts
```diff
@@ -30,7 +30,7 @@ import { SubscriberProcessQueueService } from '../../services/queues/subscriber-
 import { TriggerBroadcastCommand } from './trigger-broadcast.command';
 
 const LOG_CONTEXT = 'TriggerBroadcastUseCase';
-const QUEUE_CHUNK_SIZE = 100;
+const QUEUE_CHUNK_SIZE = process.env.QUEUE_CHUNK_SIZE ?? 100;
```
scopsy, line 33: I wonder should we maybe have 2 separate more specific names here?

```suggestion
const QUEUE_CHUNK_SIZE = process.env.BROADCAST_QUEUE_CHUNK_SIZE ?? 100;
```

and for the multicast `MULTICAST_QUEUE_CHUNK_SIZE` wdyt?

Khongchai: I agree! Will apply your suggestion

Khongchai: I also forgot to parse int from process env string. Will add parseInt as well

punchvv:
Any update on this PR?

scopsy (approved):

PR #4992, opened 2023-12-17
fix: digest with passing filters are not aggregated

### LINKED ISSUE #4957: 🐛 Bug Report: Digests with filter behave improperly
### 📜 Description

When a digest has a filter attached and that filter succeeds, messages are sent immediately, resulting in one message per trigger with each message having a digest of the triggers leading up to it.

### 👟 Reproduction steps

This branch https://github.com/MasonD/novu/tree/digest-filter-failing-test has a failing test "should digest events within time interval with filter".

### 👍 Expected behavior

Digests behave identically to when they have no filter attached when they have a filter that passes. The test I wrote passes.

The test I wrote fails. Digests send one message per trigger when they have a filter even when the filter passes.

### Novu version

next (commit cf7ec7d1d327d82ac24d81444015d3c94dcd8a8b)

### npm version

8.11.0

### node version

v16.15.1

_No response_

ainouzgali:

For a digest with filters that passes - it sent separate message for each trigger, instead of aggregating the events into one message. 
Another problem fixed in this PR, is that a digest that is filtered (message should be sent immediately) was merged into a current delayed digest.

Closes #4957

### DIFF
--- a/apps/worker/src/app/workflow/usecases/queue-next-job/queue-next-job.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/queue-next-job/queue-next-job.usecase.ts
@@ -33,6 +33,7 @@ export class QueueNextJob {
           environmentId: command.environmentId,
           organizationId: command.organizationId,
           userId: command.userId,
+          job,
         })
       );
 
--- a/libs/dal/src/repositories/job/job.repository.ts
+++ b/libs/dal/src/repositories/job/job.repository.ts
@@ -182,8 +182,7 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
     job: JobEntity,
     digestKey?: string,
     digestValue?: string | number,
-    digestMeta?: IDigestRegularMetadata,
-    filtered = false
+    digestMeta?: IDigestRegularMetadata
   ): Promise<IDelayOrDigestJobResult> {
     const isBackoff = job.digest?.type === DigestTypeEnum.BACKOFF || (job.digest as IDigestRegularMetadata)?.backoff;
     if (isBackoff) {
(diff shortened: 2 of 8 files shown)


apps/worker/src/app/workflow/usecases/queue-next-job/queue-next-job.usecase.ts
```diff
@@ -33,6 +33,7 @@ export class QueueNextJob {
           environmentId: command.environmentId,
           organizationId: command.organizationId,
           userId: command.userId,
+          job,
```
ainouzgali, line 36: The conditions filter did not have the payload so all digests were created

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -207,6 +208,27 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
       }
     }
 
+    if (filtered) {
```
ainouzgali, line 211: If filtered, we want a new digest to be created, and not merged to an existing delayed digest

davidsoderberg (approved):

libs/dal/src/repositories/job/job.repository.ts
```diff
@@ -207,6 +208,27 @@ export class JobRepository extends BaseRepository<JobDBModel, JobEntity, Enforce
...
+      await this._model.updateOne(
+        {
+          _environmentId: job._environmentId,
+          _templateId: job._templateId,
+          _subscriberId: job._subscriberId,
+          _id: job._id,
+        },
+        {
+          $set: {
+            status: JobStatusEnum.DELAYED,
```
scopsy, line 221: @ainouzgali if this one was filtered out, why not marking it as skipped instead?

LetItRock: also then `digestResult: DigestCreationResultEnum.SKIPPED` should be updated ;)

ainouzgali: We can. But then we won't have the execution details. We will be having a different flow for digest than for the other steps.

packages/application-generic/src/usecases/add-job/merge-or-create-digest.usecase.ts
```diff
@@ -213,4 +219,23 @@ export class MergeOrCreateDigest {
...
+        ...metadata,
+        ...CreateExecutionDetailsCommand.getDetailsFromJob(job),
+        detail: filtered ? DetailEnum.FILTER_STEPS : DetailEnum.DIGEST_SKIPPED,
+        source: ExecutionDetailsSourceEnum.INTERNAL,
+        status: ExecutionDetailsStatusEnum.SUCCESS,
+        isTest: false,
+        isRetry: false,
+      }),
+      job._organizationId
+    );
```
ainouzgali, line 239: Show execution details for a skipped digest - for backoff as well. We only showed 'step created'

scopsy: Great addition!

scopsy (approved):

PR #4996, opened 2023-12-17
POC -  Refactor provider initialize

djabarovgeorge:

Provider initialization changes:
####  Novu
For all of the providers, the provider id (env variable) is be mandatory.
* COMMUNICATION_PROVIDER_ID
* CACHE_PROVIDER_ID
* DISTRIBUTED_LOCK_PROVIDER_ID

#### Community
There is a fallback to a single Redis instance, if the desirable provider is different the user will need to provide the proper provider id with its credentials.


packages/application-generic/src/services/in-memory-provider/services/cache-in-memory-provider.service.ts
```diff
@@ -0,0 +1,145 @@
...
+    return {
+      enableAutoPipelining,
+      showFriendlyErrorStack,
+    };
+  }
+
+  /**
+   * New way of selecting provider with priority to select through environment variables
+   * and fallback to the previous way retro compatible
+   */
```
djabarovgeorge, line 49: 🧹

Issue #5006, opened 2023-12-19
[NV-3300] most recent change should be on the top in changes page


jainpawan21:
steps to reproduce:-

1. Create a new workflow or update an existing workflow
2. Check if there is new pending change in changes page
3. update few more workflows
4. Now on changes page, changes are list in order of last updated date having last updated change at the top

Most recently updated change should be on the top. As a user I want to promote the most recent change only. To get to that change, I have to use pagination and go till the last item.

<sub>From [SyncLinear.com](https://synclinear.com) | [NV-3300](https://linear.app/novu/issue/NV-3300/most-recent-change-should-be-on-the-top-in-changes-page)</sub>

Ujjwal55:
@jainpawan21 can I take this issue? Please assign it to me

sudhanshugautam2911:
@jainpawan21 I would like to work on this issue. Please assign me

jainpawan21:
> @jainpawan21 can I take this issue? Please assign it to me

@Ujjwal55 Assigning this to you 🙂

m-Chetan:
Hello @jainpawan21 I have already worked on this issue. Could you please merge it if everything is fine?

Ujjwal55:
> > @jainpawan21 can I take this issue? Please assign it to me
> 
> @Ujjwal55 Assigning this to you 🙂

Please review this pr https://github.com/novuhq/novu/pull/5010 @jainpawan21

mahendraHegde:
@Ujjwal55  @m-Chetan  you both followed the same approach of sorting in mem in the UI, but i believe it should be done on `/changes` api side while querying DB.
@jainpawan21  correct me if i'm wrong.

Ujjwal55:
> @Ujjwal55 @m-Chetan you both followed the same approach of sorting in mem in the UI, but i believe it should be done on `/changes` api side while querying DB. @jainpawan21 correct me if i'm wrong.

Okk, I'll do it

m-Chetan:
@mahendraHegde sure we can fetch it from DB in descending order from api side

m-Chetan:
@jainpawan21 I have made changes in the PR to fetch the changes from api side while querying DB.
If all fine could you please merge it

scopsy:
@Ujjwal55 are you still on it? Or shouldIassign @m-Chetan for this issue?

Ujjwal55:
> @mahendraHegde sure we can fetch it from DB in descending order from api side

[...]

PR #5010, opened 2023-12-20
fix: Done [NV-3300] most recent change should be on the top in change…

Ujjwal55:

Fixed [NV-3300] most recent change should be on the top in changes page #5006

The changed was needed because the recent workflows were coming in bottom below the list

Before:
Screenshot from 2023-12-21 01-18-49

After:
Screenshot from 2023-12-21 01-27-32

### DIFF
--- a/apps/web/src/pages/changes/components/ChangesTableLayout.tsx
+++ b/apps/web/src/pages/changes/components/ChangesTableLayout.tsx
@@ -138,7 +138,7 @@ export const ChangesTable = ({
     <Table
       data-test-id={dataTestId}
       loading={loading}
-      data={changes || []}
+      data={changes?.slice()?.reverse() || []}
       columns={columns}
       pagination={{
         pageSize: pageSize,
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/changes/components/ChangesTableLayout.tsx
```diff
@@ -138,7 +138,7 @@ export const ChangesTable = ({
     <Table
       data-test-id={dataTestId}
       loading={loading}
-      data={changes || []}
+      data={changes?.slice()?.reverse() || []}
```
mahendraHegde, line 141: its supposed to be an api leve change,
you need to sort in `/changes` api(while querying the DB)
https://github.com/novuhq/novu/blob/next/apps/api/src/app/change/changes.controller.ts#L53

Ujjwal55: Okk, I will do it, Thanks

mahendraHegde (commented):
lgtm

jainpawan21 (approved):

PR #5014, opened 2023-12-21
feat(pkg): Add Open-Telemetry and Prometheus Montioring to Novu

AliaksandrRyzhou:

This PR adds the initial commit of nestjs-otel to Novu's API, WS, and Webhook containers.
This system set up a Prometheus metric server on 9464 on the path '/metrics' and the ability to add spans in the future.

The otel output is supported by NewRelic and will allow logs to be connected with spans and traced for performance on the open source (with jaeger and prometheus) and saas deployments as discussed in [#2878](https://github.com/novuhq/novu/pull/2878) 

image


packages/application-generic/src/tracing/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  OtelHistogram,
+  OtelObservableGauge,
+  OtelObservableCounter,
+  OtelObservableUpDownCounter,
+  OtelCounter,
+  Span,
+} from 'nestjs-otel';
+export { Counter } from '@opentelemetry/api';
+
+export const otelModule = OpenTelemetryModule.forRoot({
```
scopsy, line 18:
```suggestion
export const OtelModule = OpenTelemetryModule.forRoot({
```

Usually we try to do PascalCased module names :)

apps/api/src/bootstrap.ts
```diff
@@ -137,3 +139,13 @@ function isWidgetRoute(url: string) {
...
+
+process.on('SIGTERM', () => {
+  otelSDK
+    .shutdown()
+    .then(
+      () => console.log('SDK shut down successfully'),
+      (err) => console.log('Error shutting down SDK', err)
+    )
+    .finally(() => process.exit(0));
+});
```
scopsy, line 151: Nestjs provides shutdown hooks we can utilize: https://docs.nestjs.com/fundamentals/lifecycle-events

apps/api/src/bootstrap.ts
```diff
@@ -39,9 +39,11 @@ if (process.env.SENTRY_DSN) {
 
 // Validate the ENV variables after launching SENTRY, so missing variables will report to sentry
 validateEnv();
+const otelSDK = getOTELSDK(packageJson.name);
 
 export async function bootstrap(expressApp?): Promise<INestApplication> {
   BullMqService.haveProInstalled();
+  await otelSDK.start();
```
scopsy, line 46: Is it possible to only enable OTEL if an env variable indicating it was provided?

packages/application-generic/src/tracing/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  MetricService,
+  OtelInstanceCounter,
+  OtelUpDownCounter,
+  OtelHistogram,
+  OtelObservableGauge,
+  OtelObservableCounter,
+  OtelObservableUpDownCounter,
+  OtelCounter,
+  Span,
+} from 'nestjs-otel';
```
scopsy, line 15: Is it possible to create our own wrappers around those functions? so that we can also send metrics to new relic or other custom monitoring tools like Clouwatch in case we would want to?

.husky/commit-msg
```diff
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx --no-install commitlint --edit "$1"
+npx --no -- commitlint --edit
```
scopsy, line 5: Could you elaborate on the need for this change?

scopsy:
@AliaksandrRyzhou overall fantastic work! Added a few notes, and will let someone from the team to do a final review 🙏

packages/application-generic/src/tracing/index.ts
Cliftonz: Just as a note, I am not sure if that is needed as both of these monitoring systems support OTEL.
https://aws.amazon.com/about-aws/whats-new/2023/08/amazon-cloudwatch-agent-opentelemetry-traces-x-ray/
https://docs.newrelic.com/docs/more-integrations/open-source-telemetry-integrations/opentelemetry/opentelemetry-introduction/

However, at what capacity I am not sure and I think this requires a bit more understanding of the trade-offs and usages.

packages/application-generic/src/tracing/index.ts
AliaksandrRyzhou: Initially it was exactly what I had done but I faced a typescript-eslint/naming-convention issue when I wanted to commit my changes :)
OtelModule

.husky/commit-msg
AliaksandrRyzhou: It allows to easily set up Git hooks in your project as a pre-commit hook to enforce commit message linting using commitlint before allowing a commit to be made. I played around with it and it can be deleted. My bad :)

apps/api/src/bootstrap.ts
djabarovgeorge: I am not sure it is possible, because here we clean the otelSDK resources that are initialized outside of the nest app (and before nest app was initiated) unless I am missing something.
what we could do is maybe move this code into an application-generic/tracing file so it could be reused and we won't need to call for it in every bootstrap.

djabarovgeorge (commented):
Should we init the sdk in the @novu/worker as well?

packages/application-generic/src/tracing/index.ts
scopsy: Yes, but from previous test we made with NewRelic is very poor compared to their native APM functionality, I would love to to keep the current functionality and not deprecate it until we are happy with the OTEL metrics we neeed.

apps/api/src/bootstrap.ts
```diff
@@ -43,7 +43,9 @@ const otelSDK = initializeOtelSdk(packageJson.name);
 
 export async function bootstrap(expressApp?): Promise<INestApplication> {
   BullMqService.haveProInstalled();
-  await otelSDK.start();
+  if (process.env.ENABLE_OTEL_SDK === 'true') {
```
Cliftonz, line 46: I think ``` ENABLE_OTEL ``` would be a better name here

apps/api/src/bootstrap.ts
djabarovgeorge: Definitely, I was wondering if we could move this code to the `initializeOtelSdk` scope. meaning we initialize the SDK and then:

```typescript
process.on('SIGTERM', () => {
  otelSDK
    .shutdown()
    .then(
      () => console.log('SDK shut down successfully'),
      (err) => console.log('Error shutting down SDK', err)
    )
    .finally(() => process.exit(0));
});
```
then return the SDK instance, this way we could have the `SIGTERM` listener inside of the `application-generic/tracing` file so we wouldn't need to add this listener separately in each bootstrap.

packages/application-generic/src/tracing/index.ts
AliaksandrRyzhou: @Cliftonz @scopsy. Could you take a look at the last changes within packages/application-generic/src/tracing/index.ts and packages/application-generic/src/tracing/otel-wrapper.ts if I'm on right way? I wrapped only `Span`, `TraceService`, `MetricService` and if it's ok I'll add other.

Cliftonz:
@AliaksandrRyzhou this looks great, other then the conflicting pnpm-lock file which you can resolve by merging next and running ```pnpm i ```.

One thing we did miss is that we did not add the promethious metric port to the following docker files:
- https://github.com/novuhq/novu/blob/next/docker/local/development/docker-compose.yml
- https://github.com/novuhq/novu/blob/next/docker/local/deployment/docker-compose.yml

Cliftonz (approved):
LGTM

packages/application-generic/src/tracing/otel-wrapper.ts
```diff
@@ -0,0 +1,134 @@
...
+export function Span(name?: string, options?: SpanOptions) {
+  return OtelSpan(name, options);
+}
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function OtelInstanceCounter(options?: MetricOptions) {
+  return setOtelInstanceCounter(options);
+}
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export function OtelUpDownCounter(
+  ...dataOrPipes: (
```
rifont, line 24: We can extract this to an exported typing in this file, then reuse it throughout, e.g.

import { PipeTransform, Type } from '@nestjs/common';
export type OtelDataOrPipe = string | PipeTransform<any, any> | Type<PipeTransform<any, any>>;

...
dataOrPipes: OtelDataOrPipe[],

rifont: Unless I am missing something that necessitates a dynamic import?

rifont (commented):
Looks great, very clean implementation! Nice job.

Added a few comments for consideration.

packages/application-generic/src/tracing/otel-wrapper.ts
AliaksandrRyzhou: I think you are right, I've pushed the changes

rifont (commented):
I missed one @AliaksandrRyzhou !

Changes look great.

packages/application-generic/src/tracing/tracing.ts
```diff
@@ -0,0 +1,48 @@
...
+  });
+  process.on('SIGTERM', () => {
+    instance
+      .shutdown()
+      .then(
+        () => console.log('SDK shut down successfully'),
+        (err) => console.log('Error shutting down SDK', err)
+      )
+      .finally(() => process.exit(0));
+  });
```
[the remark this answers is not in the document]
rifont: For an example, see [here](https://github.com/novuhq/novu/blob/4eb51163468547d41bd165b93ee4340d0a55811b/apps/api/src/app/rate-limiting/usecases/get-api-rate-limit-service-maximum-config/get-api-rate-limit-service-maximum-config.usecase.ts#L24). We would need the class to implement both init and shutdown, like:
```typescript
class TracingService implements OnModuleInit, OnApplicationShutdown { 
...
  onModuleInit {
  ...
  }

  onApplicationShutdown {
  ...
  }
}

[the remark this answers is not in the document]
Comment (AliaksandrRyzhou) [reply packages/application-generic/src/tracing/tracing.ts:45]: Thanks @rifont I'll take care about it!

File: packages/application-generic/src/tracing/tracing.service.ts [hunk 9]
```diff
@@ -0,0 +1,34 @@
...
+    if (this.otelSDKInstance) {
+      await this.otelSDKInstance.shutdown();
+    }
+  }
+  onModuleInit() {
+    if (!this.hasValidParameters()) return;
+
+    this.otelSDKInstance = initializeOtelSdk(this.serviceName);
+    this.otelSDKInstance.start();
+  }
```
rifont, line 29: Excellent!

rifont (approved):
Great work @AliaksandrRyzhou . Let's ship it!

PR #5022, opened 2023-12-24
Nv 2991 endpoint to return all the template variables

ainouzgali:
Implement the endpoint that will return for a given template all the system and translation variables.

We need to provide the information about what variables are accessible to the user when editing the template content in the editor. We have introduced the translation variables which are dynamic and based on the uploaded translation files.. System variables are hardcoded on the FE, which is bad. All of the above is a reason for creating the new endpoint.


libs/shared/src/consts/handlebar-helpers/getTemplateVariables.ts
```diff
@@ -14,6 +14,10 @@ export function getTemplateVariables(bod: any[]): IMustacheVariable[] {
     .filter((body) => body.type === 'MustacheStatement')
     .flatMap((body) => {
       const varName = body.params[0]?.original || (body.path.original as string);
+
+      if (body.path.original === HandlebarHelpersEnum.I18N) {
```
ainouzgali, line 18: This is so the translation vars will not show in the trigger snippet

BiswaViraj (approved):

Issue #4505, opened 2023-10-12
[NV-2859] Add the Application Identifier field to the Integration Sidebar


LetItRock:
## Why?

In a recent customer feedback, we have received feedback stating that the difference between provider identifier and application identifier is not very clear. And they got confused between the two during implementation.

## What?

We should add a new read-only field to the Integration Sidebar showing the `Application Identifier` field under the `Provider identifier field.`

The Application identifier should have the following description:
```
The application identifier is a unique public key, used by the notification center to identify you Novu account.
```

## Definition of Done
- The new `Application Identifier` field should be added under the `Provider identifier` field in the In-App Integration Sidebar in the Integration Store 
 - The field is read-only
 - User can copy the value of it using the "copy" button
- An appropriate identifier will be displayed on the prod and dev environment

![image](https://github.com/novuhq/novu/assets/2607232/d3a8cb6a-af05-4c59-8c31-a3b9279ed022)

LetItRock:
> I would really love to work on this, can I get assigned to it?

hey 👋 sure, just assigned it to you 🤝

Osilaja78:
Nice. thanks

MrLEGENDx10:
me too ?

djabarovgeorge:
> me too ?

Hello @MrLEGENDx10 , the issue you're interested in is currently assigned. We'll inform you if it becomes available.

Ninad3721:
Hello if @Osilaja78  inactive on this issue @LetItRock  can you please assign it to me ?

Osilaja78:
Hi @LetItRock I'm able to solve the issue, but there's no way to verify what I've done because I'm having issues running the novu `web` platform. I sought help in the support channel on discord here (https://discord.com/channels/895029566685462578/1163793177883451453), but no solution. I am able to run `api` and `worker` sucessfullly but `web` just refused to start up for me 😥 so i can confirm the changes i made accordingly.

Please, you can re-assign it to whoever you want to pending the time I'm able to fix the local problem. I hope I find another issue to work on by then.
Thanks for your pateince 👍🏼.

LetItRock:
[...]

PR #5025, opened 2023-12-25
feat: added application identifier read only field in integration sidebar

learner-mohit:

Application Identifier is added in integration sidebar for Novu In App integrations. The field is read only, has a copy button and has a description as well.

This PR closes https://github.com/novuhq/novu/issues/4505


learner-mohit:
I found the application identifier fields on the settings page as well and it was mapped to environment id.
ref: https://github.com/novuhq/novu/blob/626b61b0b6c404187f10124656ab982868d32859/apps/web/src/pages/settings/tabs/ApiKeysCard.tsx#L90-L98

PR #5026, opened 2023-12-25
Store the questionnaire data

djabarovgeorge:

We would like to store the questionnaire form data sent from the Web application. 
Implement the new endpoint or extend the existing "create organization" endpoint. 
The form data will contain information like:
the user role in the organization: engineer, architect, manager, etc.
organization domain: github.com
the list with the use cases they would like to use Novu for, for example: In-App, Multi-Channel, Digest, etc.

 

We would like to understand our customer's needs better and to provide a better onboarding experience.

### Definition of Done
* implement or extend the existing endpoint
* store the questionnaire data in the database
* send the questionnaire data to the analytics service
* e2e tests


libs/dal/src/repositories/organization/organization.entity.ts
```diff
@@ -20,7 +20,14 @@ export class OrganizationEntity {
   };
 
   partnerConfigurations?: IPartnerConfiguration[];
+
   defaultLocale?: string;
+
+  jobTitle: string;
```
ainouzgali, line 26:
```suggestion
  jobTitle?: string;
```

ainouzgali (approved):

libs/shared/src/entities/organization/organization.interface.ts
```diff
@@ -2,26 +2,24 @@ import { ApiServiceLevelEnum } from '../../types';
...
-    fontFamily: string;
-    contentBackground: string;
-    direction: 'ltr' | 'rtl';
+    fontColor?: string;
+    fontFamily?: string;
+    contentBackground?: string;
+    direction?: 'ltr' | 'rtl';
   };
   defaultLocale?: string;
+  jobTitle?: string;
```
scopsy, line 20: I think this should be on the user entity?  Or there is some reason for it on the organization?

djabarovgeorge: No, you're right it was my mistake putting it there, will refactor it shortly, thanks for highlighting it!

apps/api/src/app/auth/usecases/register/user-register.command.ts
```diff
@@ -9,18 +11,34 @@ export class UserRegisterCommand extends BaseCommand {
...
   @IsOptional()
+  @IsEnum(SignUpOriginEnum)
   origin?: SignUpOriginEnum;
+
+  @IsOptional()
+  @IsEnum(JobTitleEnum)
+  jobTitle?: JobTitleEnum;
+
+  @IsString()
+  @IsOptional()
```
dammy001, line 39: I think `IsOptional` should come before `IsString` here

apps/api/src/app/organization/dtos/create-organization.dto.ts
```diff
@@ -1,11 +1,23 @@
...
   @IsString()
   @IsOptional()
   logo?: string;
+
+  @IsOptional()
+  @IsEnum(JobTitleEnum)
+  jobTitle?: JobTitleEnum;
+
+  @IsString()
+  @IsOptional()
```
dammy001, line 18: same as above

apps/api/src/app/auth/usecases/register/user-register.command.ts
dammy001: Okay I see precedence doesn't matter

djabarovgeorge: Good point, i think in both cases, IsOptional will tell class-validator to skip validation, however, it will be more optimized to check IsOptional first and then skip the next decorator.

Made a small Google search and saw this comment from one of the members
https://github.com/typestack/class-validator/issues/1688#issuecomment-1353943121, So i wonder if they are supporting order at all.

BiswaViraj (approved):

PR #5027, opened 2023-12-25
Questionnaire in the sign-up flow

djabarovgeorge:

Create a questionnaire with 3 questions to a new customer.

We want to understand our customers better to boost sales and onboarding.
Also, we want to display appropriate use-case on the Get Started page to onboard more smoothly.

image

https://www.loom.com/share/3baad9ecebe04a3e80eaddad547acebc


apps/web/src/pages/auth/components/QuestionnaireForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+import { api } from '../../../api/api.client';
+import { useAuthContext } from '../../../components/providers/AuthProvider';
+import { useVercelIntegration, useVercelParams } from '../../../hooks';
+import { ROUTES } from '../../../constants/routes.enum';
+import { DynamicCheckBox } from './dynamic-checkbox/DynamicCheckBox';
+
+type Props = {};
+
+export function QuestionnaireForm({}: Props) {
```
ainouzgali, line 36:
```suggestion
export function QuestionnaireForm() {
```

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+        render={({ field }) => {
+          return (
+            <Select
+              label="Job title"
+              data-test-id="questionnaire-job-title"
+              required
+              error={errors.jobTitle?.message}
+              {...register('jobTitle', {
+                required: 'Please input your job title',
+              })}
```
ainouzgali, line 120: If you are using the Controller you don't need the register as well. For the error message you can use 
```
<Controller ...
rules={{
          required: 'Required - Job title',
        }}
render = ...

```

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
```diff
@@ -0,0 +1,214 @@
...
+              getCreateLabel={(newOrganization) => <div>+ Add "{newOrganization}"</div>}
```
ainouzgali, line 121: Why is this needed? Isn't it just a dropdown from a pre defined set of jobs?

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
ainouzgali, line 1: @djabarovgeorge I feel a better practice would be to either go with Controller or register - and not a mix of them :)

apps/web/src/pages/auth/components/dynamic-checkbox/DynamicCheckBox.tsx
ainouzgali, line 1: Maybe not now - but we could consider adding this to our design system

libs/design-system/src/icons/general/DelayAction.tsx
```diff
@@ -25,7 +24,7 @@ export function DelayAction(props: React.ComponentPropsWithoutRef<'svg'>) {
       </mask>
       <g mask="url(#mask0_433_2481)">
         <path
-          fill="url(#paint0_linear_433_2481)"
```
ainouzgali, line 28: Can you please make sure this doesn't remove the gradient on the step in the workflow editor?
Screen Shot 2023-12-27 at 11 30 39

ainouzgali (commented):
Looks really good @djabarovgeorge

libs/design-system/src/icons/general/DelayAction.tsx
djabarovgeorge: You are I used the incorrect icon by mistake! :(

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
djabarovgeorge: Good catch! we don't need it, like it is leftover 🧹

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
djabarovgeorge: Good point! looks like i made some mess while playing with the form 🤦‍♂️

ainouzgali (approved):

PR #5028, opened 2023-12-27
fix(web): specify novu version in client

### LINKED ISSUE #3073: [NV-1903] 🚀 Feature: specify Novu version in client
### 🔖 Feature description

Current Novu version should be mentioned somewhere in client.

In case of self hosting, user will be aware that this specific version is currently used

I will suggest to add somewhere in left bottom below side menu links.

### 🔄️ Additional Information

_No response_

None

<sub>[NV-1903](https://linear.app/novu/issue/NV-1903/🚀-feature-specify-novu-version-in-client)</sub>

michaldziuba03:

It changes a way how Novu version is shown on client-side. It uses Webpack configuration to take `version` from `package.json` (because we don't want to bundle package.json) and put it into `process.env.NOVU_VERSION` as string.

I was assigned to the issue that was solved in other PR, but previous solution doesn't work on my machine so I fixed it anyway.

I ran basic Docker deployment to test this feature with default setup. It doesn't show Novu version properly.

Closes #3073

#### Before

before

#### After

after


michaldziuba03:
Related PR: #4241

apps/web/config-overrides.js
```diff
@@ -1,8 +1,16 @@
...
+const { version } = require('./package.json');
 // const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
 function overrideConfig(config, env) {
-  const plugins = [...config.plugins /* new BundleAnalyzerPlugin() */];
+  const plugins = [
+    ...config.plugins,
+    new DefinePlugin({
+      'process.env.NOVU_VERSION': JSON.stringify(version),
+    }),
```
michaldziuba03, line 11: it will replace `process.env.NOVU_VERSION` with version string.

We need JSON.stringify because Webpack will put this value into bundle as is without string quotes.

michaldziuba03:
@scopsy @jainpawan21

In my opinion this is a way how feature flags should be implemented on the frontend.

https://webpack.js.org/plugins/define-plugin/

What do you think about it?

scopsy:
@michaldziuba03  I think it's an option, CRA already has this functionality built in aswell for the other things. (By working with the env variables)

jainpawan21 (approved):

PR #5038, opened 2023-12-31
fix: self host env file and docker compose file

jainpawan21:
Update .example.env file and docker compose file as per feedback from community members

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -546,7 +546,8 @@
     "openapi",
     "headerapikey",
     "INITDB",
-    "isend"
+    "isend",
+    "Idand"
   ],
   "flagWords": [],
   "patterns": [
--- a/apps/api/src/app/integrations/e2e/get-webhook-support-status.e2e.ts
+++ b/apps/api/src/app/integrations/e2e/get-webhook-support-status.e2e.ts
@@ -1,0 +1,154 @@
+import { expect } from 'chai';
+import { UserSession } from '@novu/testing';
+import {
+  ChannelTypeEnum,
+  ChatProviderIdEnum,
+  EmailProviderIdEnum,
+  InAppProviderIdEnum,
+  PushProviderIdEnum,
+  SmsProviderIdEnum,
+} from '@novu/shared';
+import { IntegrationEntity, IntegrationRepository } from '@novu/dal';
+
+describe('Get Webhook Support Status - /webhook/provider/:providerOrIntegrationId/status (GET)', function () {
+  let session: UserSession;
+  const integrationRepository = new IntegrationRepository();
+
+  const checkBadRequestIntegration = async (integration: IntegrationEntity, message: string) => {
+    const { body } = await session.testAgent.get(`/v1/integrations/webhook/provider/${integration._id}/status`);
+    expect(body.statusCode).to.equal(400);
+    expect(body.error).to.equal('Bad Request');
+    expect(body.message).to.equal(message);
+  };
+
+  beforeEach(async () => {
+    session = new UserSession();
+    await session.initialize();
+    await integrationRepository.deleteMany({
+      _organizationId: session.organization._id,
+      _environmentId: session.environment._id,
+    });
+  });
+
+  it("should throw not found error when integration doesn't exist", async () => {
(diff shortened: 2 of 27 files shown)


docker/.env.example
```diff
@@ -46,5 +56,5 @@ WIDGET_CONTEXT_PATH=
 
 # Analytics
 SENTRY_DSN=
-NEW_RELIC_APP_NAME=
-NEW_RELIC_LICENSE_KEY=
+NEW_RELIC_APP_NAME=new-relic-novu-app
+NEW_RELIC_LICENSE_KEY=new-relic-novu-license-key
```
djabarovgeorge, line 60: do we need the mock values here, I am afraid that if the user will not provide/provide invalid credentials it will be less discoverable this way.

docker/.env.example
```diff
@@ -1,18 +1,28 @@
 # Secrets
 # YOU MUST CHANGE THESE BEFORE GOING INTO PRODUCTION
+# used as a secret to verify the JWT token signature
 JWT_SECRET=your-secret
-STORE_ENCRYPTION_KEY="<ENCRYPTION_KEY_MUST_BE_32_LONG>"
+# used to encrypt/decrypt the provider credentials
```
djabarovgeorge, line 5: 🌟

docker/.env.example
```diff
@@ -1,18 +1,28 @@
...
+# MONGO PASSWORD
+MONGO_INITDB_ROOT_PASSWORD=secret
+MONGO_URL=mongodb://root:secret@mongodb:27017/novu-db?authSource=admin
 REDIS_HOST=redis
 REDIS_PASSWORD=
 REDIS_CACHE_SERVICE_HOST=
 
 # AWS
-S3_LOCAL_STACK=http://localhost:4566
+S3_LOCAL_STACK=$HOST_NAME:4566
```
djabarovgeorge, line 25: i was not aware of this syntax in the.env file, looks good, love the reusability 😉

docker/.env.example
```diff
@@ -1,18 +1,28 @@
...
-NODE_ENV=local
-MONGO_URL=mongodb://mongodb:27017/novu-db
+NODE_ENV=production
 MONGO_MAX_POOL_SIZE=500
+MONGO_MIN_POOL_SIZE=100
+# MONGO USER
+MONGO_INITDB_ROOT_USERNAME=root
+# MONGO PASSWORD
+MONGO_INITDB_ROOT_PASSWORD=secret
+MONGO_URL=mongodb://root:secret@mongodb:27017/novu-db?authSource=admin
```
djabarovgeorge, line 19: will the mongodb manage to connect with this url locally for development purposes?

docker/.env.example
```diff
@@ -1,18 +1,28 @@
...
+# used to encrypt/decrypt the provider credentials
+STORE_ENCRYPTION_KEY=<ENCRYPTION_KEY_MUST_BE_32_LONG>
+
+# Host
+HOST_NAME=http://localhost
 
 # General
-NODE_ENV=local
-MONGO_URL=mongodb://mongodb:27017/novu-db
+NODE_ENV=production
```
djabarovgeorge, line 12: i think the default should be `local`, and will be updated to production only if it was intentional in order to reduce human mistakes. wdyt?

djabarovgeorge (commented):
Look good to me :)
I left some comments about the things I was not sure about.

docker/.env.example
djabarovgeorge: Do you have any ref of this error? or why it gives 500, it should not happen with local NODE_ENV

docker/.env.example
djabarovgeorge: Do we have ref of the errors? because if it is a warning of not being able to connect to the service it should be fine and mock key won't solve it.

Additionally, I am not sure what behavior we will have with the mock of NEW_RELIC_LICENSE_KEY. I think the best one for us would be if `key` is provided but invalid do not initialize the app, if `key` is not provided skip metric initialization.

djabarovgeorge (approved):
🚀

PR #5046, opened 2024-01-03
Add field-level encryption to API Keys

djabarovgeorge:

* Using the encryptProviderSecret function in application-generic, update the following use-cases to encrypt prior to write:
  * Update the create-environment use-case to encrypt the generated key prior to insertion in the environment DAL
* Using the decryptProviderSecret function in application-generic, update the following use-cases to decrypt after read and before sending back in response:
  * get-api-keys
  * get-environment
  * get-my-environments
* Update the authService.apiKeyAuthenticate method to:
  * Remove the apiKey from the cached result in authService.getUserData
  * Generate a hash from the unencrypted apiKey to act as the cache key in the @CachedEntity builder for authService.getUserData
  * Update the authService.getUserData method to encrypt the provided apiKey prior to environmentRepository.findByApiKey lookup
  Remove the apiKey from return of authService.getUserData
  Remove the apiKey from JWT payload in authService.getApiSignedToken
* Create a migration in the API to idempotently (i.e. only once through subsequent migrations) encrypt the API keys on all environments
* Update the [data migration docs](https://docs.novu.co/additional-resources/data-migrations) to include the new migration for the relevant version

The API key is not currently encrypted at rest, this poses a vulnerability if the raw contents of the database or the cache are available to a bad actor who may use the api key with bad intentions. We need to obfuscate the value of the api key at rest to prevent direct use of the key in the event of a database breach

We want to be able to show the API key to users from the dashboard to simplify and ease product adoption, therefore hashing of the API key in the database is not an option. We must therefore implement encryption prior to writes and decryption during reads to support the necessary obfuscation.

### Definition of Done

* API keys are encrypted at rest in the database
* API keys are hashed at rest in the cache
* API keys in dashboard are visible in decrypted form
* API keys remain usable as authentication credentials for the API
* All existing API keys become encrypted through a data migration script
* Running the data migration script more than once does not re-encrypt the api keys
https://docs.novu.co/additional-resources/data-migrations has the new migration listed


djabarovgeorge:
Things I was not sure about, these and could not validate with Richard because he is on vacation 🎉

Are the notes regarding authService.apiKeyAuthenticate and authService.getUserData redundant? because i could not find those functions.

Why generate a hash from the unencrypted apiKey to act as the cache key in the @CachedEntity builder for authService.getUserData
• we can use an encrypted key, so we won't expose the key, and the key should be relatively small. so i wonder if we need to hash it? 


Remove the apiKey from the JWT payload in authService.getApiSignedToken
• is it already removed? i did not find apiKey in the JWT payload.

apps/api/src/app/environments/dtos/environment-response.dto.ts
```diff
@@ -15,11 +13,14 @@ export class EnvironmentResponseDto {
   @ApiProperty()
   identifier: string;
 
-  @ApiProperty({
-    type: [ApiKey],
-  })
-  apiKeys: IApiKey[];
+  @ApiPropertyOptional()
+  apiKeys?: IApiKeyDto[];
```
djabarovgeorge, line 17: Optional because get-environment.usecase does not return the api keys, i wonder if we need to return them at all under environment-response.dto. I would suggest removing it at altogether and providing the keys only under GET `/environments/api-keys`

libs/shared/src/types/shared/index.ts
```diff
@@ -1 +1,5 @@
 export type CustomDataType = { [key: string]: string | string[] | boolean | number | undefined };
+
+export const NOVU_SUB_MASK = 'nvsk.';
+
+export type EncryptedSecret = `${typeof NOVU_SUB_MASK}${string}`;
```
djabarovgeorge, line 5: kudos to Richard for helping me to be more familiar with Template Literal Types.
This small type highlighted some potential issues we could have in compilation time while developing this feature, and will probably prevent future issues.

packages/application-generic/src/encryption/encrypt-provider.ts
```diff
@@ -41,19 +44,34 @@ export function decryptCredentials(
...
+  return encryptSecret(apiKey);
+}
+
+export function decryptApiKey(apiKey: string): string {
+  if (novuEncrypted(apiKey)) {
+    return decryptSecret(apiKey);
+  }
+
+  return apiKey;
+}
```
djabarovgeorge, line 69: those are middleware for the encryption-decryption that will help us make sure that we won't encrypt/decrypt twice for no reason.

apps/api/src/app/environments/dtos/environment-response.dto.ts
LetItRock: The `GET /environments` does use that interface and returns the `apiKeys`.

apps/api/src/app/environments/usecases/generate-unique-api-key/generate-unique-api-key.usecase.ts
```diff
@@ -27,6 +27,18 @@ export class GenerateUniqueApiKey {
...
   }
 
+  private async validateIsApiKeyUsed(apiKey: string) {
+    const encryptedApiKey = encryptApiKey(apiKey);
+
+    // backward compatibility - update to `findByApiKey` after encrypt-api-keys-migration run
+    const environment = await this.environmentRepository.findByApiKeyBackwardCompatibility({
+      decryptedKey: apiKey,
+      encryptedKey: encryptedApiKey,
+    });
```
LetItRock, line 37: let's have a ticket for this so we won't forget

packages/application-generic/src/encryption/encrypt-provider.ts
LetItRock: shouldn't we have these checks actually in the `encryptSecret, decryptSecret` functions? because I see that they are exposed and used in other places.

packages/application-generic/src/services/cache/key-builders/entities.ts
```diff
@@ -80,7 +81,7 @@ const buildGroupedBlueprintsKey = (): string =>
     identifier: BLUEPRINT_IDENTIFIER,
   });
 
-const buildAuthServiceKey = ({ apiKey }: { apiKey: string }): string =>
+const buildAuthServiceKey = ({ apiKey }: { apiKey: EncryptedSecret }): string =>
```
LetItRock, line 84: the ticket says to store the hash in the cache, I guess just to save memory?

apps/api/src/app/environments/usecases/create-environment/create-environment.usecase.ts
```diff
@@ -1,9 +1,10 @@
+import { nanoid } from 'nanoid';
```
LetItRock, line 1: could you please also cover this functionality with e2e tests for all use-cases touched?

apps/api/src/app/environments/dtos/environment-response.dto.ts
djabarovgeorge: True, that is why I left it as optional because `GET /environments/me` do not return it.

packages/application-generic/src/encryption/encrypt-provider.ts
djabarovgeorge: That is a good point but i did not want to merge the base functions `encryptSecret, decryptSecret` that at the moment responsible for adding and removing the prefix.  as it could affect the integration credentials as well.

apps/api/src/app/environments/usecases/generate-unique-api-key/generate-unique-api-key.usecase.ts
djabarovgeorge: Good point! i will create one.

packages/application-generic/src/services/cache/key-builders/entities.ts
djabarovgeorge: True please note the comment i left https://github.com/novuhq/novu/pull/5046#issuecomment-1875520097 regarding the ticket. i am not sure if we will benefit too much from a hash because the key is relatively small, but I am not open to suggestions.

apps/api/src/app/environments/usecases/create-environment/create-environment.usecase.ts
djabarovgeorge: I was not sure what would be the best way to test it as the API key is i the code of our testing environment. i wanted to revisit the tests tomorrow once again.

small note, our test env organically simulates backward compatibility because we do not store an encrypted api key by the @novu/test service on sesstion init. i could not add the encryption because we do not propose application genetic in test lib.
meaning that we created not encrypted api keys in the DB on session init and the application (e2e tests) works as expected (backward compatibility).

LetItRock: I see, we should have in the `EnvironmentService` the same encryption ideally by reusing the utils `encryptApiKey, decryptApiKey`.
Theoretically, both packages are built to CJS, so we might include `@novu/application-generic` in `@novu/testing`, but not suggesting that 😅 Another way would be to have another lib that is pure for reusable utils and without app-related logic, which is headache 🤕 so the only easy thing would be to duplicate the code?

packages/application-generic/src/services/cache/key-builders/entities.ts
LetItRock:
```
apiKey nvsk.41e558c55f4c1d18f45354bbcfc31c87:7db2803321f4d5b6db48625505621d922dda7a09623361ca2947f13d057ce586d2666bbfb5afabb30613be5743681239
apiKey - string size 134

apiKeyHash 9c2e166d6bc2f95fb8f92442c22e54edb3cd0529c4260af3c9f4c5f3fe98c6cb
apiKeyHash - string size 64
```

it wouldn't be a huge memory win for us (~4mb if all envs cached) but still, it's 2 times smaller in size...

[codesandbox link](https://codesandbox.io/p/devbox/encryption-9xvr3w?layout=%257B%2522sidebarPanel%2522%253A%2522EXPLORER%2522%252C%2522rootPanelGroup%2522%253A%257B%2522direction%2522%253A%2522horizontal%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522id%2522%253A%2522ROOT_LAYOUT%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522UNKNOWN%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522clqyzv07l00063b6jfgu0uvl1%2522%252C%2522sizes%2522%253A%255B70%252C30%255D%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522EDITOR%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522EDITOR%2522%252C%2522id%2522%253A%2522clqyzv07k00023b6jxr53oi7g%2522%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522direction%2522%253A%2522horizontal%2522%252C%2522id%2522%253A%2522SHELLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522SHELLS%2522%252C%2522id%2522%253A%2522clqyzv07l00043b6jfgu7omra%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%257D%252C%257B%2522type%2522%253A%2522PANEL_GROUP%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522direction%2522%253A%2522vertical%2522%252C%2522id%2522%253A%2522DEVTOOLS%2522%252C%2522panels%2522%253A%255B%257B%2522type%2522%253A%2522PANEL%2522%252C%2522contentType%2522%253A%2522DEVTOOLS%2522%252C%2522id%2522%253A%2522clqyzv07l00053b6j5prsb177%2522%257D%255D%252C%2522sizes%2522%253A%255B100%255D%257D%255D%252C%2522sizes%2522%253A%255B50%252C50%255D%257D%252C%2522tabbedPanels%2522%253A%257B%2522clqyzv07k00023b6jxr53oi7g%2522%253A%257B%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clqyzv07k00013b6jsd57vg3s%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522FILE%2522%252C%2522filepath%2522%253A%2522%252FREADME.md%2522%257D%255D%252C%2522id%2522%253A%2522clqyzv07k00023b6jxr53oi7g%2522%252C%2522activeTabId%2522%253A%2522clqyzv07k00013b6jsd57vg3s%2522%257D%252C%2522clqyzv07l00053b6j5prsb177%2522%253A%257B%2522tabs%2522%253A%255B%255D%252C%2522id%2522%253A%2522clqyzv07l00053b6j5prsb177%2522%257D%252C%2522clqyzv07l00043b6jfgu7omra%2522%253A%257B%2522id%2522%253A%2522clqyzv07l00043b6jfgu7omra%2522%252C%2522tabs%2522%253A%255B%257B%2522id%2522%253A%2522clqyzv07k00033b6jzgsuhabw%2522%252C%2522mode%2522%253A%2522permanent%2522%252C%2522type%2522%253A%2522TASK_LOG%2522%252C%2522taskId%2522%253A%2522start%2522%257D%255D%252C%2522activeTabId%2522%253A%2522clqyzv07k00033b6jzgsuhabw%2522%257D%257D%252C%2522showDevtools%2522%253Atrue%252C%2522showShells%2522%253Atrue%252C%2522showSidebar%2522%253Atrue%252C%2522sidebarPanelSize%2522%253A15%257D)

djabarovgeorge: omg i hope it did not take too much time to create the sandbox, well why not 😊,  i will look into it. 
Thanks for the comparison help!

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -1,26 +1,46 @@
...
       }
 
       environment.apiKeys = [];
 
       return environment;
     });
   }
+
+  private decryptApiKeys(environment: EnvironmentEntity): EnvironmentResponseDto {
+    const res = { ...environment };
```
rifont, line 35: I think it's always better to be descriptive with naming, because there's always a chance the function could be extended in the future.
```suggestion
    const decryptedApiKeysEnvironment = { ...environment };
```

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -1,26 +1,46 @@
...
+
+    res.apiKeys = environment.apiKeys.map((apiKey) => {
+      return {
+        _userId: apiKey._userId,
```
rifont, line 39: In case the apiKey entity grows, we should spread here to handle that future case.
```suggestion
        ...apiKey,
```

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -1,26 +1,46 @@
...
+        key: decryptApiKey(apiKey.key),
+      };
+    });
+
+    return res;
```
rifont, line 44: Fix per above
```suggestion
    return decryptedApiKeysEnvironment;
```

libs/dal/src/repositories/environment/environment.repository.ts
```diff
@@ -51,12 +51,21 @@ export class EnvironmentRepository extends BaseRepository<EnvironmentDBModel, En
...
   }
 
+  /*
+   * backward compatibility -
+   * * delete findByApiKeyBackwardCompatibility
+   * * use findByApiKey instead once encrypt-api-keys-migration executed
+   */
+  async findByApiKeyBackwardCompatibility({ key, hash }: { key: string; hash: string }) {
+    return await this.findOne({ $or: [{ 'apiKeys.key': key }, { 'apiKeys.hash': hash }] });
+  }
```
rifont, line 67: Some self-hosted Customers may never execute this migration under the current status-quo. So we'll need a way to support both flows until we have a forced deprecation of unencrypted API keys.

I think we should rename this method to `findByApiKey` and remove the other one, and create a Linear ticket to address removal of this sometime in the distant future after we have notified self-hosted Customers to ensure they run this migration.

libs/shared/src/types/shared/index.ts
```diff
@@ -1 +1,5 @@
 export type CustomDataType = { [key: string]: string | string[] | boolean | number | undefined };
+
+export const NOVU_SUB_MASK = 'nvsk.';
```
rifont, line 3: It's possible we'll add other forms of sub masks in the future, this provides avoidance of doubt and preparation for that future:
```suggestion
export const NOVU_ENCRYPTION_SUB_MASK = 'nvsk.';
```

packages/application-generic/src/encryption/encrypt-provider.ts
```diff
@@ -41,19 +44,34 @@ export function decryptCredentials(
...
+
+function novuEncrypted(text: string): text is EncryptedSecret {
```
rifont, line 71: Prefixing with `is` to provide better description that return value is boolean.

```suggestion
function isNovuEncrypted(text: string): text is EncryptedSecret {
```

packages/application-generic/src/encryption/encrypt-provider.ts
```diff
@@ -41,19 +44,34 @@ export function decryptCredentials(
...
   return text.startsWith(NOVU_SUB_MASK);
 }
 
-function needEncryption(key: string): boolean {
+function needCredentialEncryption(key: string): boolean {
```
rifont, line 75: Whilst we're changing the name here, I think we should also follow the pattern above and go with the following:
```suggestion
function isCredentialEncryptionRequired(key: string): boolean {
```

packages/application-generic/src/services/auth/auth.service.ts
```diff
@@ -387,15 +384,35 @@ export class AuthService {
...
+     */
+    const environment =
+      await this.environmentRepository.findByApiKeyBackwardCompatibility({
+        key: apiKey,
+        hash: hashedApiKey,
+      });
+
     if (!environment) {
-      return { error: 'API Key not found' };
+      return { error: 'Environment not found' };
```
rifont, line 402: This error text is returned to the API Client when their API key does not have a matching environment. Whilst this change is logical based on the code flow, it's not actually helpful to the API Client. Their API key may have had a simple spelling error, the proposed change might leave them scratching their head.
I'd suggest adding a comment here for future reference.
```suggestion
      // Failed to find the environment for the provided API key.
      return { error: 'API Key not found' };
```

rifont:
> Things I was not sure about, these and could not validate with Richard because he is on vacation 🎉
> 
> Are the notes regarding authService.apiKeyAuthenticate and authService.getUserData redundant? because i could not find those functions.
Yes, they are redundant - the API Key auth flow was recently simplified and these functions were removed.
> 
> Why generate a hash from the unencrypted apiKey to act as the cache key in the @CachedEntity builder for authService.getUserData • we can use an encrypted key, so we won't expose the key, and the key should be relatively small. so i wonder if we need to hash it?
The approach you have taken is much better. Nice work.
> 
> Remove the apiKey from the JWT payload in authService.getApiSignedToken • is it already removed? i did not find apiKey in the JWT payload.
Yes, they were removed in teh API Key auth flow simplification mentioned above.

apps/api/migrations/encrypt-api-keys/encrypt-api-keys-migration.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    expect(firstMigrationExecution[1].identifier).to.contains(secondMigrationExecution[1].identifier);
+    expect(firstMigrationExecution[1].name).to.exist;
+    expect(firstMigrationExecution[1]._organizationId).to.equal(secondMigrationExecution[1]._organizationId);
+    expect(firstMigrationExecution[1].apiKeys[0].key).to.contains(secondMigrationExecution[1].apiKeys[0].key);
+    expect(firstMigrationExecution[1].apiKeys[0].hash).to.contains(secondMigrationExecution[1].apiKeys[0].hash);
+    expect(firstMigrationExecution[1].apiKeys[0]._userId).to.equal(secondMigrationExecution[1].apiKeys[0]._userId);
+  });
+});
+
+async function pruneIntegration({ environmentRepository }: { environmentRepository: EnvironmentRepository }) {
```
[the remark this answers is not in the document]
djabarovgeorge: Agree and disagree on that one, I love the most relevant function to be at the top, for example, the `execute` method for our use-cases classes, and then all of the other functions. This way you can focus right away on the main logic and browse the other functions if needed more context.

I think the best way of doing so where you get the best of two cases is the C++ way of doing it, when you first declare and then provide the implementation.

However, I am good with both ways.

apps/api/src/app/environments/usecases/create-environment/create-environment.e2e.ts
```diff
@@ -24,8 +25,13 @@ describe('Create Environment - /environments (POST)', async () => {
     expect(body.data.identifier).to.be.ok;
     const dbApp = await environmentRepository.findOne({ _id: body.data._id });
 
+    if (!dbApp) {
+      throw new Error('App not found');
+    }
```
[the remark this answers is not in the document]
djabarovgeorge: Actually, this is needed only in order to remove the ts errors and should never occur 😅

however, i do agree that his code could be reused/copied in other places due to missing knowledge of its purpose, in order to prevent it i can do the following.

```
 if (!dbApp) {
      expect(dbApp).to.be.ok;
      throw new Error('App not found');
    }
```

wdyt?

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
djabarovgeorge: Agree, those types of variable names created are usually temporary while developing, and then I forget to update them in the end.
Thanks for highlighting i am always for naming improvement.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
djabarovgeorge: Good point!

libs/dal/src/repositories/environment/environment.repository.ts
djabarovgeorge: This is another way of doing it, feels more clean to me, ill update the code :) 
We already have a dedicated ticket  💪

packages/application-generic/src/encryption/encrypt-provider.ts
djabarovgeorge: Love it, I totally agree, before novu I always prefixed with 'is' but in novu it felt to me that people liked it less.

packages/application-generic/src/services/auth/auth.service.ts
djabarovgeorge: Good point, thanks for bringing it up.

apps/api/src/app/environments/usecases/create-environment/create-environment.e2e.ts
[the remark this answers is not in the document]
rifont: It still feels a little unnatural due to throwing an error after an assertion, but I think it's a suitable compromise.

rifont (approved):
Awesome work @djabarovgeorge . Great security posture improvement 🦾

PR #5058, opened 2024-01-08
[Onboarding] Get the blueprint by the trigger identifier

djabarovgeorge:

Implement or extend the existing endpoint GET /blueprints/:templateId that will return the workflow aka blueprint by the provided trigger identifier.

The steps in the onboarding use cases will use the constant workflow identifiers to create a copy of the workflow in the user environment.

### DOD
* implement/extends the endpoint to return the blueprint by the trigger identifier
* e2e tests


apps/api/src/app/blueprint/usecases/get-blueprint/get-blueprint.command.ts
```diff
@@ -1,8 +1,8 @@
-import { IsDefined, IsMongoId } from 'class-validator';
+import { IsDefined, IsString } from 'class-validator';
 import { BaseCommand } from '@novu/application-generic';
 
 export class GetBlueprintCommand extends BaseCommand {
   @IsDefined()
-  @IsMongoId()
+  @IsString()
```
djabarovgeorge, line 6: The thing that I was not sure about was if we needed to change this param and where to provide the data that the API user can pass either id or trigger identifier.

djabarovgeorge: i am ok with templateIdOrIdentifier :)

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -50,11 +50,34 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+          triggers: { $elemMatch: { identifier: id } },
+          isBlueprint: true,
+          _organizationId: this.blueprintOrganizationId,
+        };
+
+        break;
+      }
+      default:
+        throw new DalException(`Invalid id ${id}`);
+    }
```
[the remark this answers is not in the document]
djabarovgeorge: For me, switch statements always felt less complex, but looks like most people prefer regular if statements.

LetItRock (approved):
🚀

PR #5062, opened 2024-01-09
feat(api): Add LanchDarkly feature flag for API Idempotency

rifont:
* Add a LaunchDarkly feature flag for API idempotency
* Fix API Idempotency tests
  * Move test spec into folder matching test paths
  * Add tests for allowed authentication schemes

* We would like to safely toggle on API Idempotency to a select number of users

image


apps/api/src/app/shared/framework/idempotency.e2e.ts
rifont, line 1: This file was in a folder that didn't match the test glob pattern - so it wasn't running in CI. The file is moved and the test is fixed with these changes.

apps/api/src/app/shared/framework/idempotency.e2e.ts
```diff
@@ -222,6 +225,41 @@ describe('Idempotency Test', async () => {
...
+        const key = '10';
+        const { headers } = await session.testAgent
+          .post(path)
+          .set(HttpResponseHeaderKeysEnum.IDEMPOTENCY_KEY, key)
+          .set('authorization', '')
+          .send({ data: 201 })
+          .expect(401);
+        expect(headers[HttpResponseHeaderKeysEnum.IDEMPOTENCY_KEY.toLowerCase()]).not.to.exist;
+      });
+    });
```
rifont, line 262: Adding tests to verify that API Idempotency is only enabled for desired authentication schemes.

apps/api/src/app/shared/framework/idempotency.interceptor.ts
```diff
@@ -183,7 +218,7 @@ export class IdempotencyInterceptor implements NestInterceptor {
 
     //already seen the request return cached response
     if (parsed.status === ReqStatusEnum.ERROR) {
-      Logger.error(`returning cached error response. key:${idempotencyKey}`, LOG_CONTEXT);
+      Logger.verbose(`returning cached error response. key: "${idempotencyKey}"`, LOG_CONTEXT);
```
rifont, line 220: No need to log these as errors, as they are technically not a server error and are client errors. We should only log them in verbose mode.

rifont:
I have finished up the PR from before Christmas so the work was not lost - it is ready for review.

Cliftonz (approved):

PR #5065, opened 2024-01-10
fix(provider): add better proxy support for brevo mail provider

nohira:

Add better proxy for brevo mail provider as axios may leads to 502 error when having proxy over HTTPS

Using actual provider leads to 502 error (Squid proxy error with zero size reply) as axios have some issues with HTTPS over proxy.

Using cross-fetch with proxy-agent to have better proxy management

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -5996,9 +5996,12 @@ importers:
       '@novu/stateless':
         specifier: ^0.22.0
         version: link:../../packages/stateless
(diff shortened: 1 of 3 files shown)


nohira:
Is it possible to have this fix in the next release? I know it involve a provider that may be already productive, but there is an issue with axios and some proxies over https calls:

https://github.com/axios/axios/issues/5078

I'm on an on premise that involve a proxy for external calls

PR #5066, opened 2024-01-10
feat(provider): add smsmode provider

nohira:

Add [smsmode](https://www.smsmode.com/) provider implementation

This provider is used in my product


nohira:
> @nohira Could you look into merge conflicts please

I've rebase onto next and resolved conflict due to a new sms provider previously merged

nohira:
I've updated the project after brevo-sms PR merge

providers/smsmode/src/lib/smsmode.provider.ts
```diff
@@ -0,0 +1,61 @@
...
+} from '@novu/stateless';
+import { ProxyAgent } from 'proxy-agent';
+import 'cross-fetch';
+
+declare global {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  interface RequestInit {
+    agent: ProxyAgent;
+  }
+}
```
jainpawan21, line 15: @scopsy shall we allow `proxy-agent` and `cross-fetch` in place of `axios`?

nohira: I used this two for avoiding 502 proxy issue that axios have. You can see my other PR about proxy on sendinblue. Also, I've used it in brevo sms provider.

PR #5068, opened 2024-01-10
feat(ci): Docker images tagging

AliaksandrRyzhou:

Set up correct Semver tagging on docker deployments. It will allow labels of Pull Requests to manage the next Version of docker images.
image

The current versioning strategy for Docker images, limited to "dev" and "prod" tags, lacks the necessary granularity and precision. 

Challenges:
Limited Granularity: The simplicity of the "dev" and "prod" tags restricts the ability to express nuanced version information.
Lack of Precision: The absence of a structured versioning system hinders precise tracking and understanding of the development lifecycle.
Blue-Green and Canary Deployments: Future deployment strategies such as blue-green or canary deployments require more sophisticated versioning to ensure seamless transitions between different environments.

### Other information 

More information [here](https://www.notion.so/novuhq/SemVer-tagging-on-docker-deployments-f507a7f2621149edb0ce16bab787233f)


Cliftonz:
@AliaksandrRyzhou There is some missing words in the .cspell file that you will need to add.

.github/workflows/get-version.yml
```diff
@@ -0,0 +1,87 @@
...
+        id: get-labels
+        env:
+          PR_NUMBER: ${{ steps.show-pr-data.outputs.pr_number }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "The Pull request Number for labels getting is ${{ env.PR_NUMBER }}"
+          gh api -H "Accept: application/vnd.github+json" /repos/ras-devops/novu/issues/${{ env.PR_NUMBER }}/labels | jq '[.[].name]' > /tmp/label_list
+
+      - name: Semantic Custom Version Bumping
+        uses: ras-devops/determine-version-bump@v1.0.10
```
Cliftonz, line 83: This will be replace with a novu published one soon

.github/workflows/prod-deploy-api.yml
```diff
@@ -63,23 +73,25 @@ jobs:
           REGISTRY_OWNER: novuhq
           DOCKER_NAME: ${{matrix.name}}
           IMAGE_TAG: ${{ github.sha }}
+          SEMVER_IMAGE_TAG: ${{ needs.get_version.outputs.current_tag }}-prod
```
Cliftonz, line 76: Why is this ```-prod```

Cliftonz (commented):
@AliaksandrRyzhou would like one clarification

.github/workflows/prod-deploy-api.yml
Cliftonz: If there is an infra change we need to make to fully support semver let me know.

AliaksandrRyzhou: It's a temporary solution until we can exclude a docker build for a production environment. The idea is to build the image in the beginning and after use the already built and tested image to avoid possible issues. At the current moment, we have no chance to use equal tags for both environments because the docker build step will replace the image that was created earlier.

.github/workflows/get-version.yml
AliaksandrRyzhou: Done, replaced to novuhq/semver-tag-action@v0.0.1

PR #5069, opened 2024-01-11
fix: add options in resend senderName

jainpawan21:
Added `options.senderName` so that sendername from payload/overrides/template is used and in absence of all of three config one is used


Rieranthony:
Thank you so much for that, this will unblock me 🙌🏻

Rieranthony:
Any news on that one guys? It's blocking me quite a bit :/ Thanks!

BiswaViraj (approved):

PR #5071, opened 2024-01-11
[Onboarding] Allow hiding get started page

djabarovgeorge:

Implement the option to "hide" the Get started page from the side-bar nav item.

Users who are onboarded should have the ability to hide the Get started page.

### Definition of Done
* Add icon + tooltip to the side menu -> get started 
* Add logic to not show Get started in side menu. (we already have showOnBoarding field for user in the scheme)
* The eye icon is hidden by default
* The eye icon is displayed on hovering over menu item

https://www.loom.com/share/33c9b72798fa40528bf5a22429255f04


libs/design-system/src/navigation/NavMenu.tsx
```diff
@@ -9,32 +9,59 @@ interface INavMenuProps {
...
+              </Group>
+              {rightSideSection({ rightSide: rightSide, isHover: isHovered === link })}
+            </NavLink>
+          );
+        })}
     </div>
   );
 }
+
+const rightSideSection = ({
```
djabarovgeorge, line 47: I'm not completely satisfied with this solution, and I'm uncertain if it's the most effective approach. 
The reason it's placed here is because we want to show the VisibilityOff Icon only when the navigation link is hovered. Additionally, the component's appearance was quite intense, so I looked for a way to create a smoother transition. This is the best method I could come up with.

BiswaViraj: It will be easy to just use the native CSS display: block display none style
but it wont be easy to animate those transitions

Also i will suggest you to change the hover calculation

Create a NavLink component
and have the hover state inside that  Navlink

```jsx
const NavLinkComponent = (props) => {
const [isHover, setIsHover] = useState(false)

return <div onMouseEnter onMouseLeave>
<MantineNavlink  {...props}/>
{rightSideSection}
</div>

}


menuItems.map(item => <NavlinkComponent  ...item/>
```

libs/design-system/src/navigation/NavMenu.tsx
```diff
@@ -1,40 +1,106 @@
...
+                  setPopoverOpened: setPopoverOpened,
+                })}
+            </NavLink>
+          );
+        })}
     </div>
   );
 }
+
+const rightSideWithHover = ({
```
LetItRock, line 64: This can be also potentially extracted to a separate component. I also feel like this is a unique use-case and we should keep this component code close to the icon, and we can pass the `hover` state from the parent.

libs/design-system/src/navigation/NavMenu.tsx
```diff
@@ -1,40 +1,106 @@
...
-    rightSide?: ReactNode;
+    rightSide?: ReactNode | { component: ReactNode; displayOnHover: boolean };
     condition?: boolean;
+    tooltipLabel?: string;
   }[];
 }
 
 export function NavMenu({ menuItems }: INavMenuProps) {
   const { classes } = useStyles();
+  const { classes: popoverClasses } = usePopoverStyles();
```
LetItRock, line 21: this should be close to the extracted component, no need to propagate through the props

apps/web/src/api/hooks/useUserOnBoarding.ts
```diff
@@ -0,0 +1,56 @@
...
+  error: string;
+  message: string;
+  statusCode: number;
+}
+
+interface IRequestPayload {
+  showOnBoarding: boolean;
+}
+
+export const useUserOnBoarding = () => {
```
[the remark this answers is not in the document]
djabarovgeorge: initially, i wanted to create a general user hook, but then noticed that it would be too much 🙃

apps/web/src/components/layout/components/SideNav.tsx
```diff
@@ -64,6 +70,12 @@ export function SideNav({}: Props) {
   const { classes } = usePopoverStyles();
   const isMultiTenancyEnabled = useIsMultiTenancyEnabled();
   const isTranslationManagerEnabled = useIsTranslationManagerEnabled();
+  const {
+    showOnBoarding: showOnBoardingState,
+    isLoading: isLoadingShowOnBoarding,
```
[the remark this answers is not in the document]
djabarovgeorge: agree, sorry about that...

apps/web/src/api/hooks/useUserOnBoarding.ts
```diff
@@ -0,0 +1,56 @@
...
+  const { currentUser } = useAuthContext();
+  const queryClient = useQueryClient();
+
+  const [isUserLoading, setIsUserLoading] = useState<boolean>(true);
```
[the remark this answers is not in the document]
djabarovgeorge: Love the idea, thanks.

apps/web/src/api/hooks/useUserOnBoarding.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const mutationFunction = ({ showOnBoarding }) => updateUserOnBoarding(showOnBoarding);
+  const {
+    mutateAsync: updateOnBoardingStatus,
+    isLoading: isLoadingUpdate,
+    ...rest
+  } = useMutation<IUserEntity, IErrorResponseType, IRequestPayload>(mutationFunction, {
+    onSuccess: async (data, variables, context) => {
+      if (data) {
+        await queryClient.refetchQueries({ queryKey: ['/v1/users/me'], exact: true });
```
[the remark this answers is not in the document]
djabarovgeorge: Good point

libs/design-system/src/navigation/NavMenu.tsx
djabarovgeorge: Agree i have another issue that was solved but your other comment is by separating NavLink so we won't have infinite re-rendering.

libs/design-system/src/navigation/NavMenu.tsx
```diff
@@ -1,40 +1,117 @@
...
+          isHover: isHovered === link,
+          tooltipLabel: tooltipLabel,
+          popoverOpened: popoverOpened,
+          setPopoverOpened: setPopoverOpened,
+        })}
+    </ReactNavLink>
+  );
+};
+
+const rightSideWithHover = ({
```
LetItRock, line 75: Please move this to the separate component or include it in the `NavLink` body, because it's anti-pattern to do it that way and using additionally the hooks inside `usePopoverStyles`.

libs/design-system/src/navigation/NavMenu.tsx
```diff
@@ -1,40 +1,117 @@
...
+            />
+          );
+        })}
     </div>
   );
 }
+
+const NavLink = ({ rightSide, link, testId, icon, label, tooltipLabel }: IMenuItem) => {
+  const [popoverOpened, setPopoverOpened] = useState(false);
+  const [isHovered, setIsHovered] = useState<string>(null);
```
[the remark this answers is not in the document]
djabarovgeorge: I started with boolean, i promise i had good reason to change to string, i just do not remember why 😅
looks like it works as expected with boolean, thanks for raising it up!

PR #5073, opened 2024-01-11
fix(web): prevent updating variants to have empty conditions (front-end only)

antonjoel82:

Front-end fixes for preventing empty variant conditions
- Introduces test to catch issue with saving variants with empty conditions
- Disable submit button and add tooltip when removing all conditions from a variant

Fixes NV-3147

### DIFF
--- a/apps/web/src/components/conditions/Conditions.tsx
+++ b/apps/web/src/components/conditions/Conditions.tsx
@@ -26,6 +26,19 @@ import { OnlineConditionRow } from './OnlineConditionRow';
 import { DefaultGroupOperatorData, DefaultOperatorData } from './constants';
 import { PreviousStepsConditionRow } from './PreviousStepsConditionRow';
 
+export interface IConditionsComponentProps {
+  isOpened: boolean;
+  isReadonly?: boolean;
+  onClose: () => void;
+  updateConditions: (data: IConditions[]) => void;
+  conditions?: IConditions[];
+  name: string;
+  label?: string;
+  filterPartsList: IFilterTypeList[];
+  defaultFilter?: FilterPartTypeEnum;
+  shouldDisallowEmptyConditions?: boolean;
+}
+
 export function Conditions({
   isOpened,
   isReadonly = false,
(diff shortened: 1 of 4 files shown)


apps/web/src/components/conditions/Conditions.tsx
```diff
@@ -26,6 +26,19 @@ import { OnlineConditionRow } from './OnlineConditionRow';
...
+  isOpened: boolean;
+  isReadonly?: boolean;
+  onClose: () => void;
+  updateConditions: (data: IConditions[]) => void;
+  conditions?: IConditions[];
+  name: string;
+  label?: string;
+  filterPartsList: IFilterTypeList[];
+  defaultFilter?: FilterPartTypeEnum;
+  shouldDisallowEmptyConditions?: boolean;
```
antonjoel82, line 39: Flag here is optional / opt-in to preserve the original functionality for other callers of this component.

BiswaViraj (changes requested):
@antonjoel82, Sorry for not adding this in the task description
But we will need to update the API too to disallow update of empty conditions variants.
[more details](https://linear.app/novu/issue/NV-3147/🔀-able-to-save-variants-without-any-conditions#comment-26036b30)

antonjoel82:
> @antonjoel82, Sorry for not adding this in the task description But we will need to update the API too to disallow update of empty conditions variants. [more details](https://linear.app/novu/issue/NV-3147/%F0%9F%94%80-able-to-save-variants-without-any-conditions#comment-26036b30)

Sounds good, thanks for the call-out! I would typically do those in two different PRs, but am happy to do that in this PR as well if that's the preferred way.

BiswaViraj:
@antonjoel82, it will be better to have two different PRs. I thought API changes were might be overlooked due to not being in the task description 🙏 🙈.
Also just how linear works, I'll suggest you to create a sub-issue for API over linear and tackle it on that task. This way we will have our linear and prs synced.
Let me know if you need anything.

antonjoel82:
> @antonjoel82, it will be better to have two different PRs. I thought API changes were might be overlooked due to not being in the task description 🙏 🙈. Also just how linear works, I'll suggest you to create a sub-issue for API over linear and tackle it on that task. This way we will have our linear and prs synced. Let me know if you need anything.

Great, thank you! I'll update this PR description to make it more explicit, and have created a [sub-issue](https://linear.app/novu/issue/NV-3363/prevent-saving-variants-without-any-conditions-on-the-server) as requested.

BiswaViraj (approved):
🚀

LetItRock (approved):
awesome work!🔥 verified it with the dev testing and works like a charm :) let's now also implement the changes on API side :)

PR #5079, opened 2024-01-14
[Onboarding] Create the layout component for the onboarding page

djabarovgeorge:

Create the layout component for the redesigned Onboarding page based on the provided design specifications.

This layout component would be used as the parent for all the onboarding use cases.

image


apps/web/src/pages/get-started/GetStartedUseCase.tsx
```diff
@@ -0,0 +1,170 @@
...
+import PageHeader from '../../components/layout/components/PageHeader';
+import { ROUTES } from '../../constants/routes.enum';
+import { When } from '../../components/utils/When';
+
+import { createStyles, MantineTheme } from '@mantine/core';
+
+export const useTabs = createStyles((theme: MantineTheme, _params, getRef) => {
+  const dark = theme.colorScheme === 'dark';
+
+  const tab = getRef('tab');
```
LetItRock, line 19: not used

apps/web/src/pages/get-started/GetStartedUseCase.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+export function GetStartedTabs() {
+  const { currentOrganization } = useAuthContext();
+  const selfHosted = process.env.REACT_APP_DOCKER_HOSTED_ENV === 'true';
+  const { classes } = useTabs();
+  const navigate = useNavigate();
+  const { pathname } = useLocation();
+  const value = useMemo(() => {
+    return pathname === ROUTES.GET_STARTED ? '/' : pathname.replace(ROUTES.GET_STARTED, '');
+  }, [pathname]);
```
LetItRock, line 117: All this memoization is not needed. 
Also, you can take the `onboardingUseCase` value from the `useParams`.

apps/web/src/pages/get-started/GetStartedUseCase.tsx
```diff
@@ -0,0 +1,170 @@
...
+    );
+  }
+
+  return (
+    <Container fluid mt={15} ml={5}>
+      <Tabs
+        orientation="horizontal"
+        keepMounted={true}
+        onTabChange={(tabValue) => {
+          navigate(ROUTES.GET_STARTED + tabValue);
```
LetItRock, line 133:
```suggestion
          navigate(`${ROUTES.GET_STARTED}/${tabValue}`);
```

apps/web/src/pages/get-started/GetStartedUseCase.tsx
```diff
@@ -0,0 +1,170 @@
...
+        }}
+        variant="default"
+        value={value}
+        classNames={classes}
+        mb={15}
+      >
+        <Tabs.List>
+          <Tabs.Tab value="/inapp" icon={<RingingBell style={{ height: 20, width: 20, marginBottom: '12px' }} />}>
```
LetItRock, line 141: I would suggest not connecting the tab values with the URL path, but just having `inapp`, of course, please move it to constants and reuse it between the code.

LetItRock, line 141: we have same icon style for all the icons here, please move to the constant

apps/web/src/pages/get-started/consts.tsx
```diff
@@ -0,0 +1,53 @@
+import { ProductUseCasesEnum } from '@novu/shared';
+import { RingingBell } from '@novu/design-system/dist/cjs';
+
+export const onboardingUseCases: OnboardingUseCases = {
+  [ProductUseCasesEnum.IN_APP]: {
+    steps: ['Configure In-App provider', 'Build a workflow', 'Connect trigger and run test', 'Track activity feed'],
```
LetItRock, line 6: Please keep in mind that these steps might become more complex, as they are tied with the descriptions, and also descriptions have links, etc.

apps/web/src/pages/get-started/GetStartedUseCase.tsx
```diff
@@ -0,0 +1,170 @@
...
+    },
+
+    tabIcon: {
+      ref: tabIcon,
+      color: colors.B60,
+    },
+  };
+});
+
+export function GetStartedUseCase() {
```
LetItRock, line 96: GetStartedPage

apps/web/src/pages/get-started/GetStartedUseCase.tsx
```diff
@@ -0,0 +1,170 @@
...
+        <HeaderLayout>
+          <PageHeader title="Get Started" />
+        </HeaderLayout>
+        <GetStartedTabs />
+      </PageContainer>
+    </>
+  );
+}
+
+export function GetStartedTabs() {
```
LetItRock, line 109: please move it to the separate file

apps/web/src/App.tsx
```diff
@@ -168,7 +171,19 @@ function App() {
...
+                    <Route
+                      path="multichannel"
+                      element={<GetStartedTab onboardingUseCase={onboardingUseCases.multi_channel} />}
+                    />
+                    <Route path="digest" element={<GetStartedTab onboardingUseCase={onboardingUseCases.digest} />} />
+                    <Route path="delay" element={<GetStartedTab onboardingUseCase={onboardingUseCases.delay} />} />
+                    <Route
+                      path="translation"
+                      element={<GetStartedTab onboardingUseCase={onboardingUseCases.translation} />}
+                    />
```
LetItRock, line 185: I'm wondering whether we need this duplication, the `onboardingUseCase` can be derived from `useParams`, and then we won't need to define any new route if we will have more use cases. But then we will need to handle each use case individually in `GetStartedTab`. They are similar but just the Digest is different a little. WDYT?

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
```diff
@@ -97,7 +96,9 @@ export function QuestionnaireForm() {
 
       return;
     }
-    navigate(ROUTES.GET_STARTED);
+
+    const firstUsecase = findFirstUsecase(data.productUseCases) ?? '';
```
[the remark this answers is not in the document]
djabarovgeorge: The input for the use case is required, meaning that the user must provide it. However, if for some reason it is not provided or we decide to eliminate this requirement, I suggest the first option, which is currently the `in-app` option, as you suggested.

apps/web/src/pages/get-started/GetStartedUseCase.tsx
djabarovgeorge: thanks, i will remove it.

apps/web/src/pages/get-started/GetStartedUseCase.tsx
djabarovgeorge: Good idea. let me look into it.

apps/web/src/pages/get-started/consts.tsx
djabarovgeorge: Yes i am aware, other then link we should not have other complexity at the moment.
small note for digest i will probably wont reuse this const as we would need to make some custom code there, but for the simple steps with just text and link we could use reuse same component.

apps/web/src/App.tsx
djabarovgeorge: My plan was to create a separate component for the digest because it's more complex. Having a distinct component would be simpler than adding complexity to the GetStartedTab and onboardingUseCases const.

We could possibly incorporate routing in the GetStartedTab for the digest. If we compare the two approaches: 5 routes in App vs 1 in GetStartedTab for digest.

I'll modify the route with parameters and assess the outcome. In the next cycle, we'll observe how the digest behaves.

apps/web/src/pages/get-started/GetStartedUseCase.tsx
djabarovgeorge: Done :)

djabarovgeorge: Forgot to go over it. will create a separate const for it for now.

apps/web/src/pages/get-started/layout/GetStartedTab.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+interface IGetStartedTabProps {
+  usecase?: ProductUseCasesEnum;
+}
+
+export function GetStartedTab(props: IGetStartedTabProps) {
+  const navigate = useNavigate();
+  const { usecase: usecaseParam } = useParams<Record<OnboardingParams, OnboardingUseCasesTabsEnum | undefined>>();
+
+  const usecase = (props.usecase || usecaseParam?.replace('-', '_')) as ProductUseCasesEnum | undefined;
```
LetItRock, line 20: not sure if we need the `OnboardingUseCasesTabsEnum` and `ProductUseCasesEnum` and do this conversion? maybe better just to use `ProductUseCasesEnum`?

LetItRock (approved):

apps/web/src/pages/get-started/layout/GetStartedTab.tsx
djabarovgeorge: Agree here I didn't want this duplication. It happened because of the following reasons:

- We use underscores (_) in ProductUseCasesEnum to utilize our storage conversion. This enum is then reused in our DAL layer to represent different use cases on the organization entity.
- On the other hand, OnboardingUseCasesTabsEnum uses hyphens (-) to interact with the web URLs use case.

LetItRock: thanks, I've checked the code before :) that's why I was suggesting to use `ProductUseCasesEnum` also in the URLs ;)

djabarovgeorge: Agree on the maintenance part etc. Do you mean to use underscores in the URL or convert without the enum?

PR #5083, opened 2024-01-15
Execution Detail - Route by FF to store by queue or directly to DAL layer

djabarovgeorge:

Implement the middleware service that will use the feature flag to decide whether execution details should be written in sync manner or put on the queue.

We would like to do the A/B testing to understand how the Execution Details impact the services performance.

the middleware service that utilises the feature flag
when true it should put the execution details job on the queue
when false it should write the execution details to the repository in sync manner


apps/api/src/app/shared/shared.module.ts
```diff
@@ -95,6 +98,9 @@ const PROVIDERS = [
   InvalidateCacheService,
   storageService,
   ...DAL_MODELS,
+  ExecutionLogRoute,
+  CreateExecutionDetails,
+  getIsExecutionLogQueueEnabled,
```
djabarovgeorge, line 103: Those are here because of ConditionsFilter deps we have in IntegrationModule. i believe if we were able to merge Pablo's PR regarding the filtering service this part would not be needed.

packages/application-generic/src/usecases/get-feature-flag/get-is-execution-log-queue-enabled.use-case.ts
```diff
@@ -0,0 +1,22 @@
...
+@Injectable()
+export class GetIsExecutionLogQueueEnabled extends GetFeatureFlag {
+  async execute(featureFlagCommand: FeatureFlagCommand): Promise<boolean> {
+    const value = process.env.IS_API_EXECUTION_LOG_QUEUE_ENABLED;
+    const fallbackValue = true;
+    const defaultValue = this.prepareBooleanStringFeatureFlag(
+      value,
+      fallbackValue
+    );
+    const key = FeatureFlagsKeysEnum.IS_API_IDEMPOTENCY_ENABLED;
```
LetItRock, line 16: it uses wrong feature flag key should be `IS_API_EXECUTION_LOG_QUEUE_ENABLED`

packages/application-generic/src/usecases/get-feature-flag/get-is-execution-log-queue-enabled.use-case.ts
```diff
@@ -0,0 +1,22 @@
...
+import { FeatureFlagsKeysEnum } from '@novu/shared';
+
+import { FeatureFlagCommand } from './get-feature-flag.command';
+import { GetFeatureFlag } from './get-feature-flag.use-case';
+
+@Injectable()
+export class GetIsExecutionLogQueueEnabled extends GetFeatureFlag {
+  async execute(featureFlagCommand: FeatureFlagCommand): Promise<boolean> {
+    const value = process.env.IS_API_EXECUTION_LOG_QUEUE_ENABLED;
+    const fallbackValue = true;
```
LetItRock, line 11: shouldn't be fallback `false`?

djabarovgeorge: Well, I thought about it, isn't it dependent on what we want to have as default? I thought it should be to continue to use the queue. and then we use the FF from there. lmkwyt.

LetItRock (approved):

packages/application-generic/src/usecases/execution-log-route/execution-log-route.usecase.ts
```diff
@@ -0,0 +1,60 @@
...
+      case false: {
+        await this.createExecutionDetails.execute(
+          CreateExecutionDetailsCommand.create(command)
+        );
+        break;
+      }
+      default: {
+        Logger.warn('Execution log queue feature flag is not set', LOG_CONTEXT);
+      }
+    }
```
scopsy, line 58: A nitpick, but I think an `if` statement would be easier to read :P

packages/application-generic/src/usecases/get-feature-flag/get-is-execution-log-queue-enabled.use-case.ts
scopsy: I think it should use the queue by default, and we will turn it off explicitly. Later we can remove the FF once we finalize on the wanted solution.

PR #5084, opened 2024-01-16
fix(pkg): user 1000 instead of root

AliaksandrRyzhou:

These change will allow to use user 1000 instead of ROOT to support running containers on k8.

It's needed to support running containers on k8 and for security reasons.

The main process is run by user 1000 after these changes

Screenshot from 2024-01-16 11-26-13


scopsy:
Perfect @AliaksandrRyzhou! have we verified that the service is properly running with those changes?

AliaksandrRyzhou:
> Perfect @AliaksandrRyzhou! have we verified that the service is properly running with those changes?

@scopsy, I ran every service to make sure that the main processes feel good itself and they didn't fail. The screen above from a new build of images. To be fair it's a main checks for such changes, but there can be special cases

Cliftonz (approved):

PR #5091, opened 2024-01-17
fix: filters dont apply to a delay that is the first step in workflow

ainouzgali:

Bug fix - a delay as a first step didn't apply filters, the delay was still executed. The problem was that the check of the filters for action steps was done in the queue-next-step usecase but when delay is the first step that usecase is not executed and add-job is first (there was no problem with digest because we add a trigger step for digest)


LetItRock (approved):
💎

PR #5092, opened 2024-01-17
feat: add raw body buffer to request object for stripe webhooks

davidsoderberg:


apps/api/src/.example.env
```diff
@@ -78,3 +78,6 @@ API_RATE_LIMIT_MAXIMUM_FREE_GLOBAL=
 API_RATE_LIMIT_MAXIMUM_UNLIMITED_TRIGGER=
 API_RATE_LIMIT_MAXIMUM_UNLIMITED_CONFIGURATION=
 API_RATE_LIMIT_MAXIMUM_UNLIMITED_GLOBAL=
+
+STRIPE_API_KEY=
+STRIPE_CONNECT_SECRET=
```
rifont, line 83: Could we also list these under https://github.com/novuhq/novu/blob/next/apps/api/src/types/env.d.ts#L25 for type safety?

rifont: We should also add these env-vars to the env-validator with undefined default values:
https://github.com/novuhq/novu/blob/next/apps/api/src/config/env-validator.ts#L84

apps/api/src/bootstrap.ts
```diff
@@ -88,8 +91,18 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
...
+      return;
+    }
+
+    if (buffer && buffer.length) {
+      req.rawBody = Buffer.from(buffer);
+    }
+  };
+
+  app.use(bodyParser.json({ verify: rawBodyBuffer }));
+  app.use(bodyParser.urlencoded({ extended: true, verify: rawBodyBuffer }));
```
rifont, line 105: How difficult would it be (is it possible) to export a verify function from the BillingModule and import it here for use?
It's a little uncomfortable for separation of concerns to have `stripe-signature` header here in the OSS edition.

apps/api/src/bootstrap.ts
```diff
@@ -45,9 +45,12 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
 
   let app: INestApplication;
   if (expressApp) {
-    app = await NestFactory.create(AppModule, new ExpressAdapter(expressApp));
+    app = await NestFactory.create(AppModule, new ExpressAdapter(expressApp), {
+      bodyParser: false,
+      rawBody: true,
+    });
```
LetItRock, line 51: I'm wondering shouldn't we only do this when there is a billing module? Like:
```
if (hasBillingModule) {
   app = await NestFactory.create(AppModule, new ExpressAdapter(expressApp), {
      bodyParser: false,
      rawBody: true,
    });
    const rawBodyBuffer = require('@novu/ee-billing')?.rawBodyBuffer;
    app.use(bodyParser.json({ verify: rawBodyBuffer }));
    app.use(bodyParser.urlencoded({ extended: true, verify: rawBodyBuffer }));
} else {
   app = await NestFactory.create(AppModule, new ExpressAdapter(expressApp));
}

```
of course in a nicer way ;) because I'm worried whether this will still work for OS

rifont: This condition is now handled via the `NOVU_ENTERPRISE=true` flag below, however the presence of `bodyParser:false , rawBody: true` is tightly coupled to the code block below.

I prefer the approach Paweł has suggested, to keep the coupled functionality under the same conditional block. If we use the `NOVU_ENTERPRISE=true` flag for the condition instead of `hasBillingModule`, we get the best of both.

apps/api/src/bootstrap.ts
```diff
@@ -88,8 +91,20 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
   app.use(extendedBodySizeRoutes, bodyParser.json({ limit: '20mb' }));
   app.use(extendedBodySizeRoutes, bodyParser.urlencoded({ limit: '20mb', extended: true }));
 
-  app.use(bodyParser.json());
-  app.use(bodyParser.urlencoded({ extended: true }));
+  let rawBodyBuffer: undefined | (() => void) = undefined;
+
+  try {
+    if (process.env.NOVU_ENTERPRISE === 'true' || process.env.CI_EE_TEST === 'true') {
```
rifont, line 97: I see that we are using this OR condition in several places, but the `CI_EE_TEST` flag only ever occurs next to the `NOVU_ENTERPRISE` flag. This extra flag seems redundant and adds [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity).

Can we simplify to the following?
```suggestion
    if (process.env.NOVU_ENTERPRISE === 'true') {
```

apps/api/src/bootstrap.ts
davidsoderberg: I have now updated my code please check if that was something similar to what you have in mind :)

rifont (approved):

PR #5111, opened 2024-01-24
feat(ci): Get a build process back to a docker build instead of a dep…

AliaksandrRyzhou:

We want to shut down our account of depot.dev as we are not getting very much utility out of there platform. We want to move our build process back to docker buildx in the CI pipelines.

Cost Optimization

Just in case, I decided not to change the "package" files.json" until we make sure everything is working fine


scopsy (approved):
Great work!

Cliftonz (approved):

PR #5118, opened 2024-01-25
feat(worker): Add customData overrides for sms and fix gupshup provider

BiswaViraj:
- added customData overrides option for sms
- added DLT support for Gupshup provider
- Fixed gupshup providerID


packages/application-generic/src/factories/sms/handlers/gupshup.handler.ts
```diff
@@ -4,7 +4,7 @@ import { GupshupSmsProvider } from '@novu/gupshup';
 
 export class GupshupSmsHandler extends BaseSmsHandler {
   constructor() {
-    super('Gupshup', ChannelTypeEnum.SMS);
+    super('gupshup', ChannelTypeEnum.SMS);
```
BiswaViraj, line 7: We are using lowercase 'gupshup' provider id everywhere else, hence it was not picking up the correct provider.

scopsy (approved):

scopsy (approved):

PR #5119, opened 2024-01-25
fix(docker): use correct quotes in shell comparison in dockerfiles

### LINKED ISSUE #5052: 🐛 Bug Report: Docker containers always try to build enterprise version
### 📜 Description

The dockerfiles use the syntax:

```
RUN if [ -n '${BULL_MQ_PRO_NPM_TOKEN}' ] ; then echo 'Building with Enterprise Edition of Novu'; rm -f .npmrc ; cp .npmrc-cloud .npmrc ; fi
```

But since single quotes are used in the if, string interpolation isn't performed, and the literal string '${BULL_MQ_PRO_NPM_TOKEN}' is always non-null. Thus the npmrc swap is always performed.

### 👟 Reproduction steps

`pnpm docker:build` without BULL_MQ_PRO_NPM_TOKEN set. Observe the build output echo "Building with Enterprise Edition of Novu"

### 👍 Expected behavior

~It should use double quotes instead of single so it actually works as intended.~

It should securely leverage the Bullmq pro token to correctly identify if we are building the community or enterprise edition.

### Novu version

0\.23.0

### npm version

n/a

### node version

n/a

Yes I am willing to submit a PR! Ping me if it's more convenient for me to make those couple quote changes and submit a PR :)

MasonD:

Change from single quotes to double quotes in Dockerfiles

Closes #5052

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -19,9 +19,8 @@ COPY --chown=1000:1000 ./meta .
 COPY --chown=1000:1000 ./deps .
 COPY --chown=1000:1000 ./pkg .
 
-RUN if [ -n '${BULL_MQ_PRO_NPM_TOKEN}' ] ; then echo 'Building with Enterprise Edition of Novu' ; fi
-RUN if [ -n '${BULL_MQ_PRO_NPM_TOKEN}' ]; then rm -f .npmrc ; fi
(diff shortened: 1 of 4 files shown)


scopsy:
@AliaksandrRyzhou / @Cliftonz would appreciate your assistance in verifying this change

MasonD:
This sounds like a problem that exists with the current implementation regardless of whether the quotes are fixed, so seems pretty important to solve quickly. Are you going to handle that internally?

PR #5124, opened 2024-01-28
Translation management beta

ainouzgali:


apps/web/src/pages/templates/components/ChannelPreview.tsx
```diff
@@ -0,0 +1,59 @@
...
+      return <PushPreview />;
+
+    case StepTypeEnum.DELAY:
+      return <>DELAY</>;
+
+    case StepTypeEnum.DIGEST:
+      return <>DIGEST</>;
+
+    default:
+      return <>dummy</>;
```
BiswaViraj, line 36: @ainouzgali  now that we have implemented all the channels preview, we should have some proper UI here, 
"Unexpected error occurred!" something like that

ainouzgali: Great points, I will do that after we deploy to dev :) I want Nikolay to be able to start testing

LetItRock (approved):

PR #5125, opened 2024-01-28
feat(headless): add remove notifications method

djabarovgeorge:

### DIFF
--- a/packages/headless/src/lib/types.ts
+++ b/packages/headless/src/lib/types.ts
@@ -77,4 +77,3 @@ export type UpdateResult<
 
 export type IMessageId = string | string[];
 export type IFeedId = string | string[];
-export type EmptyObject = Record<string, never>;
--- a/packages/headless/src/lib/types.ts
+++ b/packages/headless/src/lib/types.ts
@@ -77,3 +77,4 @@ export type UpdateResult<
 
 export type IMessageId = string | string[];
 export type IFeedId = string | string[];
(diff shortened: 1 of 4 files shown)


packages/headless/src/lib/types.ts
```diff
@@ -77,3 +77,4 @@ export type UpdateResult<
 
 export type IMessageId = string | string[];
 export type IFeedId = string | string[];
+export type EmptyObject = Record<string, never>;
```
djabarovgeorge, line 80: I'd like to hear your thoughts on this approach. I created it because when the delete method returns an empty object, it seems more polished in terms of typing compared to the current approach, where the return type is 'void'. You can observe this in the method `HeadlessService.removeAllNotifications`.

LetItRock: from where that empty object is coming from? the API responds with 200 OK and doesn't return anything 🤔

djabarovgeorge: That is correct however you will still get an empty object under the response data it is the same for axios, postman, and curl I validated them separately just in case (see the screenshot attached).

Let me know what you think, i am ok with void, however, it feels like an empty object more represents the response.

image

LetItRock: I would say let's be consistent, if in the other cases, we return `void` let's do the same. Also to me, an empty object is just useless.

PR #5131, opened 2024-01-30
fix(ci): Changed the process of token getting. From docker arguments …

AliaksandrRyzhou:

Docker containers always try to build enterprise version. [Here](https://github.com/novuhq/novu/pull/5119) it was suggested to use double quotes instead of single so it actually works as intended. But if we change single quotes to double quotes the variable BULL_MQ_PRO_NPM_TOKEN will be visible every time during a docker build process. Even if we try to hide it the Docker build best practice doesn't recommend to use arguments and environment variables for secret values. Anyone can take a look at this values using  "docker image history" ot "docker inspect"

* [ Closes 5052](https://github.com/novuhq/novu/issues/5052)
* It should securely leverage the Bullmq pro token to correctly identify if we are building the community or enterprise edition.
* Build arguments and environment variables are inappropriate for passing secrets to your build, because they persist in the final image. Instead, should use secret mounts or SSH mounts, which expose secrets to your builds securely.

### DIFF
--- a/.github/actions/docker/build-api/action.yml
+++ b/.github/actions/docker/build-api/action.yml
@@ -64,7 +64,7 @@ runs:
         GH_PASSWORD: ${{ inputs.github_token }}
       run: |
         set -x
-        cd apps/api && pnpm run docker:build --build-arg BULL_MQ_PRO_TOKEN=${BULL_MQ_PRO_NPM_TOKEN}
+        cd apps/api && pnpm run docker:build
 
     - name: Tag and test
       id: build-image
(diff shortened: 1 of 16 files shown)


LetItRock (approved):
looks good to me! 🔥
maybe before merging these changes to `next` we can try the full deployment to the dev env? and we can also verify that the images that are built include the `bullmq-pro` package in the `node_modules`.

AliaksandrRyzhou:
> looks good to me! 🔥
> 
> maybe before merging these changes to `next` we can try the full deployment to the dev env? and we can also verify that the images that are built include the `bullmq-pro` package in the `node_modules`.

This is a very good idea Pawel. I'll do as you say

scopsy:
Just as a small note, it's not really a huge issues since the ee dockers are in the private repository. But I totally agree that it's a better practice

PR #5139, opened 2024-01-31
ARM Architecture Support

AliaksandrRyzhou:

These changes will allow us to build docker images for ARM platforms too. 

* This will also lower our cloud cost as ARM is more efficient and typically less expensive.
* It will allow to use Novu on ARM Platform so is really useful for many users

* We will have to remove the base image: nikolaik/python-nodejs:python3.10-nodejs16-alpine as this couldn't be compiled for arm. Because of it it's really important to make sure that it works fine after the base image changes.
* Build times might increase significantly...

Example of what we expect to see:
Screenshot from 2024-01-31 17-19-10

Screenshot from 2024-01-31 17-19-15


apps/web/Dockerfile
```diff
@@ -1,5 +1,6 @@
 # start build stage
-FROM nikolaik/python-nodejs:python3.10-nodejs20-alpine as builder
+FROM node:20-alpine3.19 as builder
+RUN apk add g++ make py3-pip
```
Cliftonz, line 3: Did you test this without python?

I am unsure if we need python in this container

apps/webhook/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM nikolaik/python-nodejs:python3.10-nodejs20-alpine as dev_base
+FROM node:20-alpine3.19 as dev_base
+RUN apk add g++ make py3-pip
```
Cliftonz, line 2: Same here

apps/ws/Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM node:20-alpine3.16
+FROM node:20-alpine3.19
 
 ENV NX_DAEMON=false
 
 RUN npm install -g pnpm@8.9.0 --loglevel notice
 RUN npm i pm2 -g
+RUN apk --no-cache add curl g++ make py3-pip
```
Cliftonz, line 7: Why was this added?

apps/web/Dockerfile
AliaksandrRyzhou: I assumed that python was needed where basic images include python. Also Pawel said that python is used on the build and in some images the ARM architecture build hung without it. I will recheck and let you know 😊

apps/webhook/Dockerfile
AliaksandrRyzhou: Same here

apps/ws/Dockerfile
AliaksandrRyzhou: I think it's my bad, forgot to delete after some tests, I'll remove it

apps/web/Dockerfile
Cliftonz: @AliaksandrRyzhou Lets double check this, I know its needed for inbound-mail for sure but I have not seen anything else in the project that needs it.

@LetItRock Are you aware of any build tools that leverage python for our project?

I want to make sure we are only doing what we need to cut build times down as much as possible, but @AliaksandrRyzhou if this takes more then 2 hours to debug lets create a ticket and look at it in the future so we can get this moved through.

AliaksandrRyzhou: This I get when I try to build the web service without Python
image

Cliftonz (approved):

AliaksandrRyzhou:
@Cliftonz I've made a quick investigation and found out that we can't build the docker images for the ARM platform without Python. It fails even in the images where Python wasn't installed initially. It looks like some npm packages include native dependencies that need to be compiled during installation. These dependencies might be platform-specific and iinstalling Python might provide a necessary build environment for these native dependencies.

estubmo:
I wonder if something has broken this PR, because the OS/Arch tab is no longer present in the built images on ghrc

AliaksandrRyzhou:
> I wonder if something has broken this PR, because the OS/Arch tab is no longer present in the built images on ghrc

@estubmo Please accept our apologies. We decided to change an approach of a building process in a bit, you will see multi-architecture images soon.

estubmo:
@AliaksandrRyzhou No problem, thanks for your reply. Do you have access to some arm64 tags I could use in the meantime, preferably for the prod images?

udleinati:
Same issue here. @AliaksandrRyzhou if you have any tag that works on arm, it would be great! thanks

estubmo:
Any update on this @AliaksandrRyzhou?

paulwer:
@udleinati @estubmo we have a working production on arm with an old image, i dont know it this helps someone in the meantime:
image

paulwer:
@AliaksandrRyzhou still no updates for us?

Cliftonz:
@paulwer @AliaksandrRyzhou and I are not working on Novu anymore.

This is something that will hopefully be fixed in the next version. @SokratisVidros would you be able to speak to this?

SokratisVidros:
Currently, we are reshuffling the roadmap and aiming at a stable, multi-architecture, Novu V2 docker release in early September.

paulwer:
@SokratisVidros is this considdered in the upcomming v2 release?

SokratisVidros:
@merrcury ☝️

paulwer:
@merrcury @SokratisVidros does it help, when I open an other issue for better clarity on this topic or are there further informations for us now?

merrcury:
Hey, @paulwer We are working on V2 release. We will see if we will be able to solve this with that. For community Announcement, Please check discord. [Last Announcement ](https://discord.com/channels/895029566685462578/1040040454906986578/1292917556503384195). 

Otherwise, I'll separately add this to our roadmap. Stay Tuned

PR #5148, opened 2024-02-02
feat: New Icon library -- Material Symbols

antonjoel82:

- New Icon set from Material Symbols
- Adds deprecation warning to existing Icons

- Fixes [NV-3365](https://linear.app/novu/issue/NV-3365/%F0%9F%8D%AD-implement-new-icon-set-google-material-icons)
- Design has begun using the

- Can be viewed via Storybook under `Icons/MaterialIcon`

### TODO

- [ ] Fix Cypress error with style importing
- [ ] Component test -- my proposal is to do these in a follow-up PR with `@testing-library/react` (see below)

### Future Considerations

- We will be changing how Colors and Sizes work in the future
- We should remove Cypress until absolutely necessary and instead use `@testing-library/react` for Component tests because it's faster, lighter weight, is the industry standard for Component testing,  and requires less configuration


LetItRock (approved):
amazing work! 🔥

antonjoel82:
Confirmed that using the CSS for the icons from `design-library` -> `web` seems to be what's broken because using the icons and CSS directly from a 3rd party library works: https://github.com/novuhq/novu/pull/5171

antonjoel82:
Closing in favor of #5196 as we will not be using Material Symbols

PR #5156, opened 2024-02-04
chore: upgrade bullmq packages

scopsy:

Fixes a security vulnerability and introduces a few performance optimizations


scopsy:
Inspected the change log of the version bumps, couldn't find any related breaking changes for us.

djabarovgeorge:
> Inspected the change log of the version bumps, couldn't find any related breaking changes for us.

Are the type errors in the tests are redundant?

scopsy:
@djabarovgeorge I believe they are, from my memory we use the connection in our actual code. 

But haven't verified, if you will have some time to help checking it out would appreciate!

PR #5159, opened 2024-02-06
Fix client payload on remove messages

djabarovgeorge:

Fixes the payload on bulk remove messages in the API client.


jainpawan21:
@djabarovgeorge is this test run failure relevant for this PR?
https://github.com/novuhq/novu/actions/runs/7796836089/job/21262313358?pr=5159#step:17:202

LetItRock (approved):

PR #5160, opened 2024-02-06
fix(api): bulk create subscribers after they were deleted

LetItRock:

When deleted subscribers are "recreated" with bulk create endpoint we will "mark" them as `deleted: false` (not deleted).

https://github.com/novuhq/novu/assets/2607232/247f749a-498f-4092-bee4-4bb2b161c10d


djabarovgeorge (approved):
🚀

PR #5162, opened 2024-02-06
refactor(infra): EE symlinks

rifont:
* [feat: Add ee symlinking script](https://github.com/novuhq/novu/commit/145dfdf84bfa77b16ec35e14d2be51cb246f5df5)
* [fix: Add nx cache to gitignore](https://github.com/novuhq/novu/commit/17258086251b2f78b851355ab04768f29b6e4a85)
* [fix: Git ignore ee symlink binaries](https://github.com/novuhq/novu/commit/6e9d20adf23dfb94473554463b26a0969780ff23)
* [refactor: Remove redundant symlink script from ee package.json](https://github.com/novuhq/novu/commit/5a22ac94d009928500176f76e962338c41d17d32)

* Each EE package needed a symlink script with custom file depth, we can abstract this into a script following a common pattern.
  * Symlinks were being checked into source, this will cause broken symlinks for community contributors
  * Symlinks were recursively nesting in EE src folders
  * Without the `-f` flag on the `ln` command, the symlink creation fails on repeated script executions. Adding `-f` makes the operation idempotent without modifying the symlinking behaviour
* NX was producing cache files that could be checked in accidentally

_Symlink script run (existing `pnpm symlink:submodules` script with some better logging and using new script)_
image

_Symlink recursive src folder nesting issue (fixed)_
image

_Symlink pnpm script error (fixed)_
image

_NX Cache files showing in git pre-stage (fixed)_
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ docker/.env
 
 # NX Build System
 .nx/cache
+
+# EE Symlinked folders
+enterprise/packages/**/src
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,6 @@ docker/.env
 .vercel
 
 .pnpm-store/
+
+# NX Build System
+.nx/cache
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "jarvis": "node scripts/jarvis.js",
     "dev-environment-setup": "sh ./scripts/dev-environment-setup.sh",
     "get-affected": "node scripts/print-affected-array.mjs",
-    "symlink:submodules": "nx run-many --target=symlink --all",
+    "symlink:submodules": "pnpm --filter \"@novu/ee-*\" exec node \"$(pwd)/scripts/symlink-ee.mjs\"",
     "install:with-ee": "pnpm install && pnpm symlink:submodules"
   },
   "devDependencies": {
--- a/scripts/symlink-ee.mjs
+++ b/scripts/symlink-ee.mjs
@@ -1,0 +1,23 @@
+import fs from 'fs';
+import path from 'path';
+import { execSync } from 'child_process';
+
+// Get the current directory
+const currentDir = process.cwd();
+
+// Remove the prefix up to and including "enterprise/packages/"
+const relativePath = currentDir.split('enterprise/packages/')[1];
+
+// Count the number of slashes in the relative path to determine the depth
+const depth = (relativePath.match(/\//g) || []).length;
+
+// Generate the correct number of "../" for the symlink command
+const prefix = '../'.repeat(depth + 3);
+
+// Generate the symlink command
+const symlinkCommand = `ln -sf ${prefix}.source/${relativePath}/src`;
+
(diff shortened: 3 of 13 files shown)


scripts/symlink-ee.mjs
```diff
@@ -0,0 +1,23 @@
...
+const depth = (relativePath.match(/\//g) || []).length;
+
+// Generate the correct number of "../" for the symlink command
+const prefix = '../'.repeat(depth + 3);
+
+// Generate the symlink command
+const symlinkCommand = `ln -sf ${prefix}.source/${relativePath}/src`;
+
+// Execute the symlink command
+execSync(symlinkCommand);
```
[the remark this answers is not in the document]
rifont: Bad bot. There is no use of environment variables, only use of Node's `process.cwd`

package.json
```diff
@@ -68,7 +68,7 @@
     "jarvis": "node scripts/jarvis.js",
     "dev-environment-setup": "sh ./scripts/dev-environment-setup.sh",
     "get-affected": "node scripts/print-affected-array.mjs",
-    "symlink:submodules": "nx run-many --target=symlink --all",
+    "symlink:submodules": "pnpm --filter \"@novu/ee-*\" exec node \"$(pwd)/scripts/symlink-ee.mjs\"",
```
rifont, line 71: Calling a single script for these types of operations ensures we only have a single script to modify, reducing maintenance cost.

.gitignore
```diff
@@ -102,3 +102,9 @@ docker/.env
 .vercel
 
 .pnpm-store/
+
+# NX Build System
+.nx/cache
+
+# EE Symlinked folders
+enterprise/packages/**/src
```
djabarovgeorge, line 110: 🤩

scripts/symlink-ee.mjs
```diff
@@ -0,0 +1,23 @@
...
+const relativePath = currentDir.split('enterprise/packages/')[1];
+
+// Count the number of slashes in the relative path to determine the depth
+const depth = (relativePath.match(/\//g) || []).length;
+
+// Generate the correct number of "../" for the symlink command
+const prefix = '../'.repeat(depth + 3);
+
+// Generate the symlink command
+const symlinkCommand = `ln -sf ${prefix}.source/${relativePath}/src`;
```
djabarovgeorge, line 18: we don't miss ` ./src` in the end of the string?
```suggestion
const symlinkCommand = `ln -sf ${prefix}.source/${relativePath}/src ./src`;
```

package.json
djabarovgeorge: Love the one command. it is true that it will be easier to maintain it unless we will need to maintain symlink-ee script 😅

scripts/symlink-ee.mjs
rifont: This resolves the nesting issue.

djabarovgeorge (approved):

PR #5172, opened 2024-02-08
Nv 3475 - typeerror cannot create property buttons on string

djabarovgeorge:

Resolve the issue where the user is storing an invalid value under message-template.cta.action

bug

### DIFF
--- a/apps/api/src/app/message-template/usecases/create-message-template/create-message-template.usecase.ts
+++ b/apps/api/src/app/message-template/usecases/create-message-template/create-message-template.usecase.ts
@@ -1,13 +1,13 @@
 import { Injectable } from '@nestjs/common';
 import { MessageTemplateEntity, MessageTemplateRepository } from '@novu/dal';
-import { ChangeEntityTypeEnum } from '@novu/shared';
+import { ChangeEntityTypeEnum, IMessageAction } from '@novu/shared';
 
 import { CreateMessageTemplateCommand } from './create-message-template.command';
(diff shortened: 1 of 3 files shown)


apps/api/src/app/message-template/usecases/create-message-template/create-message-template.usecase.ts
```diff
@@ -18,6 +18,10 @@ export class CreateMessageTemplate {
   ) {}
 
   async execute(command: CreateMessageTemplateCommand): Promise<MessageTemplateEntity> {
+    if ((command?.cta?.action as IMessageAction | undefined | '') === '') {
+      throw new ApiException('Please provide a valid CTA action');
+    }
```
djabarovgeorge, line 23: I do not like this solution but this would be the best way we could validate and alert the API user with a proper message, and this is because we do not support nested command validation [REF](https://github.com/novuhq/novu/pull/4306),  I will raise the topic with the team.

apps/web/src/pages/templates/components/in-app-editor/preview/InAppWidgetPreview.tsx
```diff
@@ -41,7 +41,7 @@ export function InAppWidgetPreview({
 
   function onRemoveTemplate() {
     setIsButtonsTemplateSelected(false);
-    onChange('');
+    onChange({});
```
djabarovgeorge, line 44: This solved the web side issue.

LetItRock (approved):
🔥

PR #5182, opened 2024-02-12
Nv 3230 auto remove deleted subscribers from topic

djabarovgeorge:

Subscribers can be added in topic using subscriberId. 
Subscriber can be deleted later using API or SDK, however a user manually also has to delete them from every topic they where apart of.

If subscriber is deleted, it is still part of topic
Currently user had to call API to first check if subscriber is part of any topic then remove that subscriber from topic.

### Definition of Done
* Novu should automatically handle this usecase
* The migration script to remove not existing subscribers from the topics


djabarovgeorge:
Closing this PR because the branch was branched out not from next.
Please see the new PR https://github.com/novuhq/novu/pull/5187

PR #5184, opened 2024-02-12
[Onboarding] - Refactor tabs and routing

antonjoel82:

- Keep Get Started Tabs as a single route with search params for changing tabs
- Make `GetStartedTab` a dumb component that just accepts a rendering logic from props
- Separate concerns from tabs

- Per UX conversation with Nikolay, this is preferred for routing
- Sets foundation for us to use a React.Context provider within the tabs page if we need to

https://www.loom.com/share/67bc4170b0fa439087521d1730a95222


apps/web/src/AppRoutes.tsx
```diff
@@ -99,10 +97,7 @@ export const AppRoutes = () => {
...
         </Route>
         {isImprovedOnboardingEnabled ? (
-          <Route path={ROUTES.GET_STARTED} element={<GetStartedPage />}>
-            <Route path="" element={<GetStartedTab usecase={ProductUseCasesEnum.IN_APP} />} />
-            <Route path=":usecase" element={<GetStartedTab />} />
-          </Route>
+          <Route path={ROUTES.GET_STARTED} element={<GetStartedPage />} />
         ) : (
           <Route path={ROUTES.GET_STARTED} element={<GetStarted />} />
         )}
```
antonjoel82, line 103: This conditional (that requires a re-render) is why search params don't persist across refresh. Once we remove this, they will work as expected

antonjoel82:
@djabarovgeorge This is the branch / PR I would branch off of!

LetItRock (approved):
nice and clean! 🙌

apps/web/src/pages/get-started/GetStartedPage.tsx
```diff
@@ -1,19 +1,30 @@
...
+import { useAuthContext } from '../../components/providers/AuthProvider';
+import { Center, Loader } from '@mantine/core';
+import { colors } from '@novu/design-system';
+import { useTabSearchParams } from './components/get-started-tabs/useGetStartedTabs';
 
 const PAGE_TITLE = 'Get started';
 
 export function GetStartedPage() {
+  const { currentOrganization } = useAuthContext();
+  const { currentTab, setTab } = useTabSearchParams();
```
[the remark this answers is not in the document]
antonjoel82: Thanks for the feedback! I will merge it as is for now, but we can memoize the tab behaviors so that they don't re-render the whole component tree if it becomes necessary.

apps/web/src/AppRoutes.tsx
djabarovgeorge: Not related to this scope
I wonder if we could/want to create a "loading" indication on our FF's so in such cases we would create a loading state, so we would avoid unnecessary rerenders and screen changes.

@antonjoel82 would love to hear your thoughts on it :)

antonjoel82: @djabarovgeorge What is `FF` in this context?

Sounds like a good UX improvement! We could do a temporary one, but overall my preference would be to upgrade our React version to use `Suspense` which will be a huge improvement overall.

Alternatively, one way I've done this in the past to avoid the re-routing issue is to route to a container page (just a page component that uses the hook to determine which content to show). Then, we'd delete that page later and point directly to the new functionality

djabarovgeorge: Sorry by i meant feature flag by FF.
Still had not the chance to use Suspense wonder how good is it.
Agree that a middleware component could help.

thanks for the feedback 🙃

PR #5189, opened 2024-02-13
feat: add workflow creation from blueprint

djabarovgeorge:

Create a component that will be responsible for the creation of a use case workflow from a blueprint.

So we could reuse it across the get started onboarding.


apps/web/src/api/hooks/notification-templates/useFetchBlueprints.ts
```diff
@@ -1,12 +1,12 @@
 import { useQuery, UseQueryOptions } from '@tanstack/react-query';
-import { IconName } from '@fortawesome/fontawesome-svg-core';
 import localforage from 'localforage';
 import { addWeeks } from 'date-fns';
 import { IGroupedBlueprint } from '@novu/shared';
 
 import { getBlueprintsGroupedByCategory } from '../../notification-templates';
 import { QueryKeys } from '../../query.keys';
 import { IBlueprintTemplate } from '../../types';
+import { getWorkflowBlueprintDetails } from '../../../utils';
```
antonjoel82, line 9: **👏 praise:** ‏Thank you for extracting this!

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+}) {
+  const openNewTab = (route: ROUTES, param: string) =>
+    window.open(`${window.location.origin}${route.replace(':templateId', param)}`, '_blank', 'noreferrer noopener');
+
+  const { createTemplateFromBlueprint } = useCreateTemplateFromBlueprint({
+    onSuccess: (template) => {
+      openNewTab(ROUTES.WORKFLOWS_EDIT_TEMPLATEID, template._id ?? '');
+    },
+    onError: () => {
+      errorMessage('Something went wrong while creating template from blueprint, please try again later.');
```
antonjoel82, line 29:
```suggestion
      errorMessage('Something went wrong while creating a template from the blueprint. Please try again later.');
```

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+    },
+  });
+
+  const { refetch } = useQuery(
+    ['blueprint', blueprintIdentifier],
```
antonjoel82, line 34: **❔ question:** Isn't `blueprintIdentifier` already a `string`?‏

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+      refetchOnWindowFocus: false,
+
+      onSuccess: (data) => {
+        if (data) {
+          const { name: blueprintName } = getWorkflowBlueprintDetails(data.name);
+          createTemplateFromBlueprint({
+            blueprint: { ...data, name: blueprintName },
+            params: { __source: TemplateCreationSourceEnum.ONBOARDING_GET_STARTED },
+          });
+        }
```
antonjoel82, line 47: **❔ question:** Is this function async? If so, we need to await or return `createTemplateFromBlueprint`, right?‏

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+      },
+    }
+  );
+
+  const segment = useSegment();
+  const handleCreateTemplateClick = () => {
+    segment.track('[Template Store] Click Create Notification Template', {
```
antonjoel82, line 54: **❔ question:** Is this an accurate tracking event (in reference to `[Template Store]`)?‏

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export const UnstyledButton = styled.button`
+  background: none;
+  border: none;
+  padding: 0;
+  font: inherit;
+  cursor: pointer;
+  outline: inherit;
+  color: ${colors.gradientEnd};
+`;
```
antonjoel82, line 73: **🤓 nitpick:** Please use our `Button` from the Design System here if possible. If we need to reset styles, I think we can use the `unstyled` prop: https://v5.mantine.dev/styles/styles-api/#unstyled-components‏

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { getWorkflowBlueprintDetails } from '../../../utils';
+import { TemplateCreationSourceEnum } from '../../templates/shared';
+import { useSegment } from '../../../components/providers/SegmentProvider';
+
+export function CreateWorkflowButton({
+  blueprintIdentifier,
+  children,
+}: {
+  blueprintIdentifier: string;
+  children: string;
```
antonjoel82, line 19: **💣 chore:** Please rename this Prop or use `React` types for children‏

apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
```diff
@@ -25,7 +28,7 @@ export const DelayUseCaseConst: OnboardingUseCase = {
         return (
           <StepDescription>
             <StepText>Novu pre-built workflow with a digest node.</StepText>
-            <Link children={' Customize '} href={'https://mantine.dev/core/timeline/'} />
+            <CreateWorkflowButton children={' Customize '} blueprintIdentifier={USECASE_BLUEPRINT_IDENTIFIER} />
```
antonjoel82, line 31: **👏 praise:** NOICE‏

apps/web/src/utils/templates.ts
```diff
@@ -0,0 +1,13 @@
...
+export const getWorkflowBlueprintDetails = (templateName: string): { name: string; iconName: IconName } => {
+  const regexResult = /^:.{1,}:/.exec(templateName);
+  let name = '';
+  let iconName = 'fa-solid fa-question';
+  if (regexResult !== null) {
+    name = templateName.replace(regexResult[0], '').trim();
+    iconName = regexResult[0].replace(/:/g, '').trim();
+  }
+
+  return { name, iconName: iconName as IconName };
```
antonjoel82, line 12: **💭 thought:** I realize this is just refactoring / moving this, but this IconName coupling is going to be a problem with de-coupling from our icon set(s)‏

antonjoel82 (approved):
Great work! Added some comments and questions, and only one small "chore" regarding re-naming / typing

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
djabarovgeorge: I wonder if we could create something reusable so in most cases the text would be the same but a param in this case could have been `creating template from blueprint`

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
djabarovgeorge: Good question thanks for bringing it up, i was totally not sure about the events here, i left them so i could comment on them but in the end forgot to 🤦‍♂️

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
djabarovgeorge: Another place i wanted to raise, a button is a proper element here, right?

apps/web/src/utils/templates.ts
djabarovgeorge: what do you mean by `our icon set(s)`?

small note: this logic was created so we could store icons by string conversion, this way we can store them easily in the template store under workflow.name

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
djabarovgeorge: I wonder what prefix should we keep here because we create the workflow by `Template Store` and we have the location `location: TemplateCreationSourceEnum.ONBOARDING_GET_STARTED`.
@antonjoel82  wdyt?

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
antonjoel82: Semantically, a link (`a`) makes more sense here indeed. If we stick with `Button`, we should also set the `role: link` in the JSX

apps/web/src/utils/templates.ts
antonjoel82: Yeah, this is a tough one, but we're tightly coupled with FontAwesome as an icon set due to this implementation

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
antonjoel82: I would say that we're never in the template store though, so this would mis-represent the source of the action in segment IMO. I would say `[Get Started]` may be a better fit

djabarovgeorge: Done

apps/web/src/pages/get-started/components/CreateWorkflowButton.tsx
antonjoel82: This is where we'd want to lean on `i18n` more, but we could start by creating a "default" error message

PR #5190, opened 2024-02-13
[Onboarding] - Groundwork for all use cases

antonjoel82:

- Add all static links for each use case
- Connect In-App tab to specific Integration store side panel
- Create abstraction to make adding animations easier for the different color schemes
- Fix Timeline styling in light mode

- Reduces future work while we await next steps (for [NV-3243](https://linear.app/novu/issue/NV-3243/🙋-in-app-use-case-tab) - 3247)

https://www.loom.com/share/1d42b72c833e4f128353721d0cc420b3?sid=6fde9997-ecb6-4bd0-806a-a2bacf1d76b0

### DIFF
--- a/apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
+++ b/apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
@@ -18,7 +18,7 @@ export const DelayUseCaseConst: OnboardingUseCase = {
             <StepText>
               Novu has set up trial email and SMS providers for you. To expand your options, add more providers in the
             </StepText>
-            <Link children={' Integration store'} href={'https://mantine.dev/core/timeline/'} />
+            <Link href={ROUTES.INTEGRATIONS}> Integration store</Link>
             <StepText>.</StepText>
           </StepDescription>
         );
--- a/apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
+++ b/apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
@@ -1,4 +1,5 @@
 import { CreateWorkflowButton } from '../components/CreateWorkflowButton';
+import { ROUTES } from '@novu/shared-web';
 import { Link, StepDescription, StepText } from './shared';
 import { OnboardingUseCase } from './types';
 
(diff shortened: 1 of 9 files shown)


apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
```diff
@@ -25,7 +27,7 @@ export const DelayUseCaseConst: OnboardingUseCase = {
         return (
           <StepDescription>
             <StepText>Novu pre-built workflow with a digest node.</StepText>
-            <Link children={' Customize '} href={'https://mantine.dev/core/timeline/'} />
+            <Link href={ROUTES.INTEGRATIONS}> Integration store</Link>
```
LetItRock, line 30: The value here should be actually `Customize`, [ref](https://www.figma.com/file/hlsUxjq2kxUU7aVBf2QyGu/Onboarding--experience?node-id=3608%3A25001&mode=dev)
But the link (button) would call the API endpoint to create the workflow from the blueprint constant trigger id.

apps/web/src/pages/get-started/consts/InAppUseCase.const.tsx
```diff
@@ -10,9 +15,20 @@ export const InAppUseCaseConst: OnboardingUseCase = {
     {
       title: 'Configure In-App provider',
       Description: function () {
+        const { integrations } = useGetIntegrationsByChannel({ channelType: ChannelTypeEnum.IN_APP });
+
+        const getInAppIntegrationUrl = () => {
+          const inAppIntegration = integrations?.[0];
+          if (!inAppIntegration) {
+            return ROUTES.INTEGRATIONS;
```
LetItRock, line 23: in this case, I think we can open the Select provider sidebar with a preselected In-App tab... this URL should do that `/integrations/create?scrollTo=in_app`

antonjoel82: Sounds good, thanks!

apps/web/src/pages/get-started/consts/DelayUseCase.const.tsx
antonjoel82: Shoot thank you, good catch

LetItRock (approved):
🔥

PR #5193, opened 2024-02-14
fix: sender name and preheader not promoted

ainouzgali:

Sender name and preheader not promoted to production


LetItRock (approved):
thank you! ❤️

PR #5196, opened 2024-02-15
Nv-3365: Material Design Icons

antonjoel82:

- New Icon set from Material Design
- Adds deprecation warning to existing Icons

- Fixes [NV-3365](https://linear.app/novu/issue/NV-3365/%F0%9F%8D%AD-implement-new-icon-set-google-material-icons)
- Design has begun using the new icons

- Can be viewed via Storybook under `Icons/IconsV2`

### TODO

- [ ] Types are not being enforced / mapped properly when importing to `web`. Icons can be imported with the proper names, but for some reason the Icon props are not properly restricting `size`

### Future Considerations

- We will be changing how Colors and Sizes work in the future
- We should remove Cypress until absolutely necessary and instead use `@testing-library/react` for Component tests because it's faster, lighter weight, is the industry standard for Component testing,  and requires less configuration


apps/web/src/components/layout/components/HeaderNav.tsx
```diff
@@ -41,13 +44,14 @@ const Icon = () => {
   const { themeStatus } = useLocalThemePreference();
 
   if (themeStatus === 'dark') {
-    return <Moon {...headerIconsSettings} />;
+    // FIXME: Revert these changes before merging!
+    return <IconDarkMode size="18" color="blue" />;
```
antonjoel82, line 48: These will be removed before merge! Testing that things work in `web` -- I give it a solid 7/10.

- Icons imported with proper names
- They have the proper defaults (color and size)
- Providing properties changes appearance... but `size` isn't being typed properly (it's using `react-icons`' underlying types

LetItRock (approved):
nice! 🙌

antonjoel82:
Created a follow-up ticket to enforce types: https://linear.app/novu/issue/NV-3508/refine-icon-types-to-enforce-design-system-values

PR #5199, opened 2024-02-15
[Onboarding] Digest Tab

antonjoel82:

- A couple stylistic fixes
- Sets up the [temporarily unused] work for supporting the Digest Playground (via hash routing)

- https://linear.app/novu/issue/NV-3245/%F0%9F%99%8B-digest-use-case-tab

### Notes

This would benefit from a refactor when we go to implement it. Primary action items include:

- Convert each `UseCase.const` into an actual component so that we are able to customize them better, share state between the parts (i.e. `BottomSection` and `steps`)
- Use a `React.ContextProvider` within `GetStartedTabs` with a `Reducer` to wrap all the tabs to share state and the ability to dispatch action


apps/web/src/components/TimerControl.tsx
djabarovgeorge, line 1: I wonder if we should reuse the `DigestDemoFlowProvider` we have from the last playground.  it already contains all of the main logic we need to maintain the playground + the event trigger

antonjoel82: We can -- I didn't find it until I'd already implemented this one with the proper icons

apps/web/src/pages/get-started/layout/HeaderLayout.tsx
antonjoel82, line 1: Unused file

apps/web/src/pages/get-started/layout/GetStartedTab.tsx
```diff
@@ -1,23 +1,65 @@
...
+      return tabProps;
+    }
+
+    return views[currentView] ?? tabProps;
+  }, [views, currentView, tabProps]);
 
-export function GetStartedTab({ steps, Demo, title, description, useCaseLink, type }: IGetStartedTabProps) {
   return (
     <Grid align="stretch" justify={'space-between'}>
       <Grid.Col span={3} mt={12}>
```
LetItRock, line 51: I would also suggest setting this column a min-width because on the smallest screen that we support it has a huge gap IMO, 18.75rem looks fine. But when shrinking the viewport width the animation will jump below, having nowrap on parent helps but creates the scroll :/

Screenshot 2024-02-21 at 19 14 02
Screenshot 2024-02-21 at 19 20 49

LetItRock (approved):
🔥

apps/web/src/pages/get-started/consts/DigestUsecasePlaygroundView.const.tsx
```diff
@@ -0,0 +1,69 @@
...
+    {
+      title: 'Run trigger multiple times',
+      Description: function () {
+        return (
+          <StepDescription>
+            <StepText>Click the button multiple times to generate a few notifications.</StepText>
+            <Button variant="outline" mt="0.5rem">
+              Run workflow trigger
+            </Button>
+          </StepDescription>
```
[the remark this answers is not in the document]
antonjoel82: Good catch, thanks!

PR #5203, opened 2024-02-17
chore(web): marketing utm campaign on doc links

LetItRock:

To all `https://novu.co` and `https://docs.novu.co` references in the app, we would like to add the marketing `utm_campaign=in-app` query param.


ainouzgali (approved):

ainouzgali:
@LetItRock does this need to be added for [those](https://github.com/novuhq/novu/blob/e14a6870026b92cc34a861688f2be4fbedfd1c73/libs/shared/src/consts/providers/channels/email.ts) as well?

LetItRock:
> @LetItRock does this need to be added for [those](https://github.com/novuhq/novu/blob/e14a6870026b92cc34a861688f2be4fbedfd1c73/libs/shared/src/consts/providers/channels/email.ts) as well?

ahh good catch, I'm not sure how the search didn't show me these :/ thank you!

PR #5206, opened 2024-02-17
[Use Case Onboarding Project] - Open onboarding workflow node.

djabarovgeorge:

This ticket is responsible for:
* navigate to a specific workflow node.
* reuse existing workflow.
* allow to fetch workflow by identifier.


djabarovgeorge:
recreated in https://github.com/novuhq/novu/pull/5207

PR #5207, opened 2024-02-17
Nv 3510 - Open onboarding workflow node

djabarovgeorge:

This ticket is responsible for:
* navigate to a specific workflow node.
* reuse existing workflow.
* allow to fetch workflow by identifier.


apps/web/src/pages/get-started/components/OpenWorkflowButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+    },
+  });
+
+  const openNewTab = (url) => window.open(url, '_blank', 'noreferrer noopener');
+
+  const handleOpenWorkflowClick = async () => {
+    segment.track('[Get Started] Click Create Notification Template', {
+      templateIdentifier: blueprintIdentifier,
+      location: TemplateCreationSourceEnum.ONBOARDING_GET_STARTED,
+    });
```
djabarovgeorge, line 42: This way we will know if the user wanted to `Create` new use case workflow, or `Configure` use case node.

```suggestion
if(node){
    segment.track('Click Configure Use-Case Node [Get Started] ', {
      templateIdentifier: blueprintIdentifier,
      node: node
    });
    } else {
        segment.track('Click Create Workflow [Get Started] ', {
      templateIdentifier: blueprintIdentifier,
      node: node
    });
```

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -47,27 +47,44 @@ export class NotificationTemplateRepository extends BaseRepository<
     return this.mapEntity(item);
   }
 
-  async findBlueprint(idOrIdentifier: string) {
-    if (!this.blueprintOrganizationId) throw new DalException('Blueprint environment id was not found');
-
+  buildIdOrIdentifierQuery(idOrIdentifier: string) {
```
antonjoel82, line 50: **💭 thought:** This seems like something that may be handled better at the service layer rather than making than opening up this query to multiple input types. While this may be a bit of an exaggeration, if we continue with this pattern, we could end up with a single query for 7 different identifier types.

I'd love to hear other opinions on it, but I think this query should be "closed" for modification and instead create a separate query function for the other type of identifier. This allows another function to delegate which query should be used. I also imagine a worst-case scenario that could involve `|` across an unindexed field resulting in terrible performance for every execution.

Definitely not a blocker, but perhaps a discussion item!‏

apps/web/src/pages/get-started/components/OpenWorkflowButton.tsx
antonjoel82: Not sure I understand why we'd switch the `[ ]` to the end.

Also, this seems like something that could be handled with a ternary:

```suggestion
    const action = node ? 'Click Create Notification Template' : 'Click Create Workflow'; 
    segment.track(`[Get Started] ${action}`, {
      templateIdentifier: blueprintIdentifier,
      location: TemplateCreationSourceEnum.ONBOARDING_GET_STARTED,
    });
```

antonjoel82 (commented):
Thanks for the updates!

apps/web/src/pages/get-started/components/OpenWorkflowButton.tsx
djabarovgeorge: Let's centralize the topic regarding the `[area]` in one place, i already addressed this question here :) 
https://github.com/novuhq/novu/pull/5212#discussion_r1496120200

Regarding the ternary, it has two cons in my opinion
* one event we have a `node` and another does not.
* as I love less code with reusable variables, in those scenarios where we log data it will be harder to debug code and search when the log is dynamic.

let me know what you think on the matter

apps/web/src/pages/get-started/consts/types.ts
```diff
@@ -13,3 +13,10 @@ export interface OnboardingUseCase {
   steps: IOnboardingStep[];
   Demo: React.ComponentType<any>;
 }
+
+export enum OnboardingNodeEnum {
```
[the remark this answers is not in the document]
djabarovgeorge: Could you point me to the right enum, I did not find a proper one.

Now when I am looking at the name I find it misleading, as we do not pass a node but a route. I will update it. i think OnboardingWorkflowRouteEnum would be more appropriate but naming in not the easy part 😅

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: Firstly, let me say that I completely agree. I really dislike having to add it here instead of at the application layer. Now, let me explain why it's included here, and if you have any other suggestions, I'd be more than happy to hear them, given the reasons mentioned above.

Here's why it's included here:

- We want to provide users of those endpoints, starting with the blueprint and now the template itself, the ability to fetch by the identifier in cases where our user might not be aware of the `_id`.
- The main reason is the validation we perform on the query payload build. To construct a proper payload, we need to determine whether it is a MongoDB ID or not.

djabarovgeorge: But you got me thinking 👀, let me try something out.

apps/web/src/pages/get-started/consts/DigestUseCase.const.tsx
```diff
@@ -19,7 +20,7 @@ export const DigestUseCaseConst: OnboardingUseCase = {
             <StepText>
               Novu has set up trial email and SMS providers for you. To expand your options, add more providers in the
             </StepText>
-            <Link href={ROUTES.INTEGRATIONS}> Integration store</Link>
+            <Link children={' Integration store'} href={ROUTES.INTEGRATIONS_CREATE} />
```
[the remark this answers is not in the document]
djabarovgeorge: Agree, sorry added those by mistake

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: Done! :)
let me know what you think, and btw thanks for questioning the implementation ♥️

libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -47,26 +47,28 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+      _organizationId: this.blueprintOrganizationId,
+      _id: id,
+    };
+
+    const item = await this.MongooseModel.findOne(requestQuery).populate('steps.template').lean();
+
+    return this.mapEntity(item);
+  }
+
+  async findBlueprintByTriggerIdentifier(identifier: string) {
```
antonjoel82, line 64: **👏 praise:** Nice, well done!‏

apps/web/src/pages/get-started/components/OpenWorkflowButton.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { ROUTES } from '../../../constants/routes.enum';
+import { useCreateTemplateFromBlueprint } from '../../../api/hooks';
+import { errorMessage } from '../../../utils/notifications';
+import { getBlueprintTemplateById, getTemplateById } from '../../../api/notification-templates';
+import { getWorkflowBlueprintDetails } from '../../../utils';
+import { TemplateCreationSourceEnum } from '../../templates/shared';
+import { useSegment } from '../../../components/providers/SegmentProvider';
+import { OnboardingWorkflowRouteEnum } from '../consts/types';
+
+export function OpenWorkflowButton({
```
antonjoel82, line 15: **🎯 suggestion:** This component and file are quite overloaded with functionality. I'd suggest we refactor it to improve readability and testability‏

antonjoel82 (approved):

antonjoel82:
Per discussion with @LetItRock, we will merge this for now, and handle the discussion items above in a separate PR

PR #5223, opened 2024-02-20
NV-3238 | Onboarding Animations

antonjoel82:

- Animations for Rive

- https://linear.app/novu/issue/NV-3238/%F0%9F%99%8B-use-cases-animations

### Notes

- We are currently putting all assets directly into the application. This is less than ideal, and would be better if we do it via CDN. _We may be able ask Pixel Point to host it on their Rive CDN_

https://www.loom.com/share/b217b9f652284d83a096ea07aa5677b2?sid=71c4b9a2-569d-486f-8692-e195a8aa00ed


apps/web/src/pages/get-started/components/GetStartedAnimation.tsx
```diff
@@ -0,0 +1,91 @@
...
+// uses `public` as the default base directory
+const ROOT_ANIMATION_PATH = `animations/get-started`;
+const STATE_MACHINE_NAME = 'SM';
+const STATE_MACHINE_INPUT_NAME = 'white theme';
+
+const getAnimationPath = (useCase: OnboardingUseCasesTabsEnum) => `${ROOT_ANIMATION_PATH}/${useCase}.riv`;
+
+const AnimationContainer = styled.div`
+  /* taken from Figma to try to get a good estimate on aspect ratio */
+  aspect-ratio: 540 / 472;
```
LetItRock, line 20: I would suggest additionally setting the max-width (you can decide about the value, IMO 62.5rem looks fine) because on a bigger screen, it looks huge. And maybe margin auto on a sides to center it.

apps/web/src/pages/get-started/components/GetStartedAnimation.tsx
```diff
@@ -0,0 +1,91 @@
...
+       * Here, we load a font asset with a random font on load of the Rive file
+       * and return true, because this callback handler is responsible for loading
+       * the asset, as opposed to the runtime
+       */
+      if (asset.isFont) {
+        /*
+         * decodeFont creates a Rive-specific Font object that `setFont()` takes
+         * on the asset from assetLoader
+         */
+        fetch('/fonts/Lato.ttf').then(async (res) => {
```
LetItRock, line 50: With the new design system, we will be using the system-defined fonts, so this might not be needed I guess?

LetItRock, line 50: btw not sure why but I see its fetched 16 times for me locally

LetItRock (approved):

apps/web/src/pages/get-started/components/GetStartedAnimation.tsx
antonjoel82: As in browser-native ones? Unfortunately we will still need to [decode a ttf](https://help.rive.app/runtimes/loading-assets#handling-assets) either way for it to work with Rive

apps/web/src/pages/get-started/components/GetStartedAnimation.tsx
antonjoel82: Good suggestions, thanks! In the future, I think we should ask Nikolay to specify during the design phase

apps/web/src/pages/get-started/components/GetStartedAnimation.tsx
antonjoel82: Took a brief crack at this but to no avail. I created a follow-up ticket [here](https://linear.app/novu/issue/NV-3527/fix-font-loading-in-rive-animations). While it is definitely not ideal, I would prefer not to block on this since it behaves properly for the user (albeit non-performantly)

PR #4128, opened 2023-09-09
Add whatsapp-business provider

antoniodinovi:
### What change does this PR introduce?

This PR introduce the whatsapp-business chat provider, is a simple implementation based on rest call to the graph api

### Why was this change needed?

This PR try to Close #1019
### Other information (Screenshots)

--

Issue #2689, opened 2023-02-06
[NV-1638] ✨ Polishing: WhatsApp Chat Provider Workflow as SMS


gepd:
### 📜 Description

I tried to add whatsapp as provider but whatsapp workflow is more similar to a SMS than a chat provider.

Related PR https://github.com/novuhq/novu/pull/1297

To send a message you do this:

```cpp
curl -i -X POST `
  https://graph.facebook.com/v15.0/NumberId/messages `
  -H 'Authorization: Bearer authorizationtoken' `
  -H 'Content-Type: application/json' `
  -d '{ \"messaging_product\": \"whatsapp\", \"to\": \"receiverNumber\", \"recipient_type\":\"individual\", \"type\": \"text\", "text": { \"preview_url\": \"false\", \"body\":\"Hello World\" } }'
```

We have 3 important parameters

1. NumberId: This is the sender number ID
2. Authorization Token: No much explanation here, we need this token to send the message
3. We add the receiver phone number in the payload (`to` property)

The rest of the parameters are static

### 👟 Reproduction steps

Current whatsapp implementation as SMS provider https://github.com/novuhq/novu/pull/1297

### 👍 Expected behavior

The chat provider needs a webhook url which whatsapp do not use, so it's confused to use whatsapp as SMS provider, we will need to update the chat workflow to work similtar to the SMS provider for cases like this

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

<sub>[NV-1638](https://linear.app/novu/issue/NV-1638/✨-polishing-chat-provider-workflow-as-sms)</sub>

oba2311:
Thanks @gepd !
We'll definitely consider it when designing this flow which is planed later on this year.

PR #5232, opened 2024-02-23
feat(providers): Add Whatsapp business as provider

### LINKED ISSUE #1019: [NV-782] - WhatsApp Provider
## What?
It would be nice to have a provider to send WhatsApp notifications

## Why?
WhatsApp is being currently used in many use cases to notify users about deliveries, app updates, etc.

## How?
Integrating WhatsApp for Business API. The difficult part, I think, may be the WhatsApp enterprise onboarding, after that, integrating the API is simpler. Another option would be Twillio Messaging for WhatsApp.

vitoorgomes:

This PR adds the option to use WhatsApp Business as a provider. The requested issue is #1019.

Although there is already a PR opened, I decide to create this fresh one for two main reasons: firstly because I don't think the OP it's working anymore on that based on replies not being answered and secondly because it only implements sending message as a template, where the Meta API allows to send it as a simple text as well, I wanted to give a little more customization to the user.

There some improvements that could be done here, but they aren't breaking changes, so I think it's worth to open this now, like allowing some other configurations like `preview_url` and other fields available in the Meta's API.

There is also the #2689 issue, while it's valid, it would be nice to have this provider as a SMS now, and port this later when Chat Providers are able to receive more data following this issue thread updates.

This Closes #1019 and Closes #4128 

SCR-20240223-pxj
SCR-20240223-pxe


Cliftonz:
@vitoorgomes Thanks for bringing this to our attention, we will take a look at it as soon as we can!

Cliftonz:
@vitoorgomes  What are the other improvements you think could be done?

Cliftonz (changes requested):
@vitoorgomes There is two things you need to fix here.
1. The hard coded authentication header
2. adding the whatsapp names to the cspell list

libs/shared/src/consts/providers/credentials/provider-credentials.ts
```diff
@@ -1120,3 +1120,21 @@ export const eazySmsConfig: IConfigCredentials[] = [
     description: 'Your SMS Channel Id',
   },
 ];
+
+export const whatsaAppBusinessConfig: IConfigCredentials[] = [
```
antonjoel82, line 1124: ☑ **chore:** Please correct this typo‏
```suggestion
export const whatsAppBusinessConfig: IConfigCredentials[] = [
```

libs/shared/src/consts/providers/provider.enum.ts
```diff
@@ -45,6 +45,7 @@ export enum CredentialsKeyEnum {
   state = 'state',
   externalLink = 'externalLink',
   channelId = 'channelId',
+  phoneNumberIdentification = 'phoneNumberIdentification',
```
antonjoel82, line 48:
```suggestion
  PhoneNumberIdentification = 'phoneNumberIdentification',
```

antonjoel82 (commented):
Thanks for the hard work here, Vitor! Added a couple comments regarding CI and to ensure we have good handling around phone numbers.

providers/whatsapp-business/src/lib/whatsapp-business.provider.ts
```diff
@@ -0,0 +1,76 @@
...
+import Axios, { AxiosInstance } from 'axios';
+import { ISendMessageRes } from '../types/whatsapp-business.types';
+import { WhatsAppMessageTypeEnum } from '../consts/whatsapp-business.enum';
+
+export class WhatsappBusinessSmsProvider implements ISmsProvider {
+  id = 'whatsapp-business';
+  channelType = ChannelTypeEnum.SMS as ChannelTypeEnum.SMS;
+
+  private readonly axiosClient: AxiosInstance;
+  private readonly baseUrl = 'https://graph.facebook.com/v18.0/';
```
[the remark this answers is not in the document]
scopsy: Yep. tested it out.

libs/shared/src/consts/providers/credentials/provider-credentials.ts
scopsy: Fixed

libs/shared/src/consts/providers/credentials/provider-credentials.ts
```diff
@@ -1120,3 +1120,21 @@ export const eazySmsConfig: IConfigCredentials[] = [
...
+    key: CredentialsKeyEnum.ApiToken,
+    displayName: 'Access API token',
+    description: 'Your WhatsApp Business access API token',
+    type: 'string',
+    required: true,
+  },
+  {
+    key: CredentialsKeyEnum.phoneNumberIdentification,
+    displayName: 'Phone Number Identification',
+    description: 'Your WhatsApp Business phone number identification',
```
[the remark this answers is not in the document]
scopsy: This is provided in the WhatsApp interface, you can't really control it. The needs to copy paste this. I will add this to the doccs section of the whole process

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -108,6 +110,14 @@ export class SendMessageChat extends SendMessageBase {
         Object.values(ChatProviderIdEnum).includes(chan.providerId as ChatProviderIdEnum)
       ) || [];
 
+    const phone = subscriber.phone;
+    chatChannels.push({
+      providerId: ChatProviderIdEnum.WhatsAppBusiness,
+      credentials: {
+        phoneNumber: phone,
+      },
+    });
```
scopsy, line 119: I don't love this implementation, but until we refactor the whole "Chat" channel flow, this workaround is needed.

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -167,33 +177,50 @@ export class SendMessageChat extends SendMessageBase {
...
 
     const chimeraOverride = this.getChimeraOverride(command.chimeraData?.providers, integration);
 
     const chatWebhookUrl =
       chimeraOverride?.webhookUrl || command.payload.webhookUrl || subscriberChannel.credentials?.webhookUrl;
+    const phoneNumber = subscriberChannel.credentials?.phoneNumber;
     const channelSpecification = subscriberChannel.credentials?.channel;
 
-    if (!chatWebhookUrl) {
+    if (!chatWebhookUrl && integration.providerId !== ChatProviderIdEnum.WhatsAppBusiness) {
```
scopsy, line 223: Those workaround will be handeled during a more thorough refactoring of the "Chat" Channel

apps/worker/src/app/workflow/usecases/send-message/send-message-type.usecase.ts
```diff
@@ -26,6 +26,7 @@ export abstract class SendMessageType {
   ) {
     const errorString =
       stringifyObject(error?.response?.body) ||
+      stringifyObject(error?.response?.data) ||
```
scopsy, line 29: This needed to gracefully map axios errors

packages/node/src/lib/events/events.spec.ts
```diff
@@ -85,6 +85,9 @@ describe('test use of novus node package - Events', () => {
               name: 'Fourth Child',
             },
           },
+          headers: {
+            'X-Novu-Custom-Header': 'test-data',
+          },
```
scopsy, line 90: This fixes a failing test on next branch

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -157,7 +167,7 @@ export class SendMessageChat extends SendMessageBase {
     chatChannel: NotificationStepEntity,
     content: string
   ) {
-    const integration = await this.getIntegration({
+    const searchCriteria: SelectIntegrationCommand = {
```
SokratisVidros, line 170:
```suggestion
    const integrationCommand: SelectIntegrationCommand = {
```

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -167,33 +177,50 @@ export class SendMessageChat extends SendMessageBase {
...
-          status: ExecutionDetailsStatusEnum.FAILED,
-          isTest: false,
-          isRetry: false,
-          raw: JSON.stringify({
-            reason: `Integration with integrationId: ${subscriberChannel?._integrationId} is either deleted or not active`,
-          }),
-        })
-      );
+    /**
+     * Current a workaround as chat providers for whatsapp is more similar to sms than to our chat implementation
```
SokratisVidros, line 183: Would you consider inverting the logic for this workaround to make it easier to reason about? That is, build the command payload without an ID and add the provider is not WhatsApp.

SokratisVidros: Alternatively you can redefine a `getIntegration` method on this class and encapsulate the workaround only in that method.

providers/whatsapp-business/src/types/whatsapp-business.types.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+interface IContact {
+  input: string;
+  wa_id: string;
+}
+
+interface IMessage {
+  id: string;
+}
```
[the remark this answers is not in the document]
SokratisVidros: +1. Also, does the IMessage interface only have an ID?

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -239,8 +267,8 @@ export class SendMessageChat extends SendMessageBase {
       })
     );
 
-    if (chatWebhookUrl && integration) {
-      await this.sendMessage(chatWebhookUrl, integration, content, message, command, channelSpecification);
+    if ((chatWebhookUrl && integration) || (phoneNumber && integration)) {
```
SokratisVidros, line 270: In an attempt to start defining a naive strategy pattern until we do the proper refactoring, I want to suggest introducing two private methods:

- sendChannelMessageViaWebhookUrl
- sendChannelMessageViaAPIKey

That way, we will ensure that messages created in Mongo have only defined fields per use case and have fewer and shorter if conditions across the file. 

The current `sendError` can be implemented inside the above two private methods separately for now.

SokratisVidros (approved):
Nice!

I've added some comments regarding code structure aiming to isolate the necessary patch for now in dedicated functions that will be easier to refactor in the future.

zegomess:
thanks @scopsy for helping out, I didn't see the notifications from months ago, totally my bad

really happy that this was merged 🤝

scopsy:
Thank you @vitoorgomes so much for your contribution ❤️ and sorry it took a while to merge 🙏

PR #5234, opened 2024-02-24
fix: subscriber details not available for digest filters

### LINKED ISSUE #5163: 🐛 Bug Report & fix: subscriber details not available for digest filters
### 📜 Description

If you use `on: 'subscriber'` for a filter on a digest step, that filter always fails because a bug causes all subscriber details to be null when evaluating filters on digests. I've figured out the root cause (a weirdness in ConditionsFilter's `normalizeVariablesData` that requires an optional argument on ConditionsFilterCommand to work correctly. We might want to fix that separately) and am ready to submit a PR once this issue is assigned to me.

### 👟 Reproduction steps

Add a workflow with a digest with the following filter:
```
{
  "value": "AND",
  "children": [
    {
      "field": "data.digest",
      "value": "digest",
      "operator": "EQUAL",
      "on": "subscriber"
    }
  ]
}
```

Trigger the workflow with a user that has the correct data:
```js
await novu.subscribers.identify("subscriberId", {
  firstName: "Pawan",
  lastName: "Jain",
  email: "pawan.jain@domain.com",
  data: { digest: "digest" }
});
await novu.trigger('workflow-id',
  {
    to: ['subscriberId'],
    payload: { example: 'example' },
  }
);
await novu.trigger('workflow-id',
  {
    to: ['subscriberId'],
    payload: { example: 'example' },
  }
);
```

### 👍 Expected behavior

The digesting should happen, as the susbcriber has the correct data for the filter to pass.

The subscriber receives two notifications, each digest saying in the web interface "Step was filtered based on steps filters"
image

### Novu version

next

### npm version

10.1.0

### node version

v20.8.1

_No response_

MasonD:
Passes step to conditionsFilter.filter so it correctly recognises when a digest filter depends on subscriber.

Closes #5163

### DIFF
--- a/packages/application-generic/src/usecases/add-job/add-job.usecase.ts
+++ b/packages/application-generic/src/usecases/add-job/add-job.usecase.ts
@@ -85,6 +85,7 @@ export class AddJob {
           environmentId: command.environmentId,
           organizationId: command.organizationId,
           userId: command.userId,
+          step: job.step,
           job,
         })
       );


ainouzgali (approved):
Thank you @MasonD !!

PR #5243, opened 2024-02-28
Pyroscope Integration

Cliftonz:

This PR introduces the integration of Pyroscope, an open-source continuous profiling platform, into the Novu codebase. By incorporating Pyroscope, we enable real-time performance monitoring and profiling of our Node.js application. This integration involves adding the Pyroscope Node.js agent to our dependencies and initializing it within our application's entry point to begin automatic profiling.

Why was this change needed?

As Novu continues to grow and scale, ensuring optimal performance and resource utilization becomes increasingly critical. The integration of Pyroscope provides us with a powerful tool for continuous profiling, allowing us to identify performance bottlenecks, optimize code, and enhance the overall efficiency and reliability of Novu. This proactive approach to performance monitoring ensures that we can maintain a high-quality service for our users, even as the complexity and scale of our operations increase.

By implementing continuous profiling with Pyroscope, we gain deeper insights into the runtime behavior of our application, enabling us to make data-driven decisions for performance improvements. This change is a step forward in our commitment to delivering a robust and high-performing notification system.


packages/application-generic/src/profiling/profiling.controller.ts
```diff
@@ -0,0 +1,33 @@
+import { Controller, Get, HttpException } from '@nestjs/common';
+import { ApiExcludeController } from '@nestjs/swagger';
+import { PinoLogger } from '../logging';
+import { ProfilingService } from './profiling.service';
+
+@Controller('/debug/pprof')
+@ApiExcludeController()
```
LetItRock, line 7: I guess these won't be accessible from the outside right?

LetItRock (approved):

packages/application-generic/src/profiling/profiling.controller.ts
Cliftonz: Currently, I have not hooked them to the api yet as with AWS ALB it is not able to block these request but istio can. 
I can also have this set up where it is conditionally enabled.

PR #5244, opened 2024-02-29
fix: Update Content-Type header to application/x-www-form-urlencoded

BiswaViraj:
- fixed Gupshup provider request format to use form-encoded
- [Gupshup docs](https://docs.gupshup.io/docs/sample-code#sample-nodejs-code-for-sending-a-single-message)

### DIFF
--- a/providers/gupshup/src/lib/gupshup.provider.ts
+++ b/providers/gupshup/src/lib/gupshup.provider.ts
@@ -12,7 +12,7 @@ if (!globalThis.fetch) {
 
 export class GupshupSmsProvider implements ISmsProvider {
   channelType = ChannelTypeEnum.SMS as ChannelTypeEnum.SMS;
-  public static BASE_URL = 'http://enterprise.smsgupshup.com/GatewayAPI/rest';
+  public static BASE_URL = 'https://enterprise.smsgupshup.com/GatewayAPI/rest';
 
   constructor(
     private config: {
--- a/providers/gupshup/src/lib/gupshup.provider.ts
+++ b/providers/gupshup/src/lib/gupshup.provider.ts
@@ -46,9 +46,9 @@ export class GupshupSmsProvider implements ISmsProvider {
     const opts = {
       method: 'POST',
       headers: {
-        'Content-Type': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded',
       },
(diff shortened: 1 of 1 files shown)


providers/gupshup/src/lib/gupshup.provider.ts
```diff
@@ -46,9 +46,9 @@ export class GupshupSmsProvider implements ISmsProvider {
     const opts = {
       method: 'POST',
       headers: {
-        'Content-Type': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded',
       },
-      body: JSON.stringify(params),
+      body: new URLSearchParams(params),
```
LetItRock, line 51: shouldn't we send the data in the URL according to their [documentation](https://docs.gupshup.io/docs/send-message-to-single-number#send-api-request)? and I'm not sure about the POST method.
Btw the URL should start with `https`.

BiswaViraj:
```js
var request = require("request");
var options = { method: 'POST',
url: 'https://enterprise.smsgupshup.com/GatewayAPI/rest',
form:
{ method: 'sendMessage',
send_to: '919820XXXXXX',
msg: 'This is sample test message from GupShup',
msg_type: 'TEXT',
userid: 'XXXXXX',auth_scheme: 'PLAIN',
password: 'XXXXX',
format: 'JSON' } };
request(options, function (error, response, body) {
if (error) throw new Error(error);
console.log(body);
});
```

```python

import requests
url = "https://enterprise.smsgupshup.com/GatewayAPI/rest"
payload =
"method=sendMessage&send_to=919820XXXXXX&msg=This%20is%20sample%20test%20message%20from%20
GupShup&msg_type=TEXT&userid=XXXXXX&auth_scheme=PLAIN&password=XXXXX&format=JSON"
response = requests.request("POST", url, data=payload)
print(response.text)

```

@LetItRock They support both `GET` and `POST` method. I took the example from their `Node` and `Python` codes.
Should we update it to use `GET` instead?

LetItRock: @BiswaViraj I was assuming they support both :) all right let's leave it then :)

LetItRock (approved):

PR #5255, opened 2024-03-04
fix(provider): Change payload + allow http legacy SSL renegotiation

RahulDutta007:

1. Changed Gupshup (Provider) BASE URL
2. Changed provider request "Content-Type" to accept "application/json"
3. Added httpsAgent to allow legacy SSL renegoation for Gupshup API endpoint

1. To allow delivery of Gupshup (provider) SMS

### DIFF
--- a/providers/gupshup/src/lib/gupshup.provider.ts
+++ b/providers/gupshup/src/lib/gupshup.provider.ts
@@ -46,12 +46,7 @@ export class GupshupSmsProvider implements ISmsProvider {
       }),
     };
 
-    const response = await axios.post(GupshupSmsProvider.BASE_URL, params, {
-      httpsAgent: new Agent({
-        rejectUnauthorized: false,
-        secureOptions: crypto.constants.SSL_OP_LEGACY_SERVER_CONNECT,
-      }),
-    });
+    const response = await axios.post(GupshupSmsProvider.BASE_URL, params);
 
     const body = response.data;
     const result = body.split(' | ');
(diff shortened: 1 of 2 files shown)


providers/gupshup/src/lib/gupshup.provider.ts
```diff
@@ -43,16 +46,14 @@ export class GupshupSmsProvider implements ISmsProvider {
...
-      method: 'POST',
-      headers: {
-        'Content-Type': 'application/json',
-      },
-      body: JSON.stringify(params),
-    };
+    const response = await axios.post(GupshupSmsProvider.BASE_URL, params, {
+      httpsAgent: new Agent({
+        rejectUnauthorized: false,
+        secureOptions: crypto.constants.SSL_OP_LEGACY_SERVER_CONNECT,
```
BiswaViraj, line 52: @RahulDutta007,
Why is this change required?

cc: @LetItRock

RahulDutta007: I was getting the fetch error "ERR_SSL_UNSAFE_LEGACY_RENEGOTIATION_DISABLED". Possibly due to running it as a localhost as from Node v18, it does not allow legacy TLS renegotiation and hence, unable to test.

BiswaViraj: @RahulDutta007, can you try using`127.0.0.1` instead of `localhost`?
I think we wont need this change.

RahulDutta007:
Let me keep this as a simple axis request without using httsAgent. Works
for you?

BiswaViraj:
> Let me keep this as a simple axis request without using httsAgent. Works for you?
Yes, that will work.

PR #5257, opened 2024-03-04
[Design System] NV-3517 | Use Tokens and Panda in `web`

antonjoel82:

- Fixes issues with Panda using dark / light modes by exposing a `useColorScheme` hook
- Connects / demos Panda functionality when used in `web`

- Closes [NV-3517](https://linear.app/novu/issue/NV-3517/💈-testing-and-validation-for-v0)

_Note: these are not meant to look pretty (or correct for the page), but rather demo Panda being used in two different ways downstream! The demo file will be reverted before merge._

The change is in the header section: you can see the title (font, size, color) and background colors use Design System values and change with dark / light mode.

Screenshot 2024-03-06 at 2 35 27 PM

Screenshot 2024-03-06 at 2 35 52 PM

### DIFF
--- a/apps/web/panda.config.ts
+++ b/apps/web/panda.config.ts
@@ -17,6 +17,12 @@ export default defineConfig({
 
   presets: [NovuPandaPreset],
 
+  /**
+   * Prefixes generated classes with the specified string (e.g. `nv-text_blue`)
+   * https://panda-css.com/docs/references/config#prefix
+   */
+  prefix: 'nv',
+
   /*
    * Any additional configuration that is specific only to `web`.
    * Most things should likely go in the `NovuPandaPreset`, but if there is something
--- a/apps/web/panda.config.ts
+++ b/apps/web/panda.config.ts
@@ -2,9 +2,6 @@ import { defineConfig } from '@pandacss/dev';
 import { NovuPandaPreset } from '@novu/design-system';
 
 export default defineConfig({
-  // Whether to use css reset
-  // preflight: true,
-
   /**
    * https://panda-css.com/docs/guides/component-library#include-the-src-files
    *
--- a/apps/web/panda.config.ts
+++ b/apps/web/panda.config.ts
@@ -2,6 +2,9 @@ import { defineConfig } from '@pandacss/dev';
 import { NovuPandaPreset } from '@novu/design-system';
 
 export default defineConfig({
+  // use CSS reset
+  preflight: true,
+
   /**
    * https://panda-css.com/docs/guides/component-library#include-the-src-files
    *
--- a/apps/web/panda.config.ts
+++ b/apps/web/panda.config.ts
@@ -3,14 +3,17 @@ import { NovuPandaPreset } from '@novu/design-system';
 
 export default defineConfig({
   // Whether to use css reset
-  preflight: true,
+  // preflight: true,
 
   /**
    * https://panda-css.com/docs/guides/component-library#include-the-src-files
    *
    * Per Panda docs, use the above approach if app code lives in an internal monorepo
    */
-  include: ['./src/**/*.{js,jsx,ts,tsx}', '../../libs/design-system/src/**/*.{js,jsx,ts,tsx}'],
+  include: [
+    './src/**/*.{js,jsx,ts,tsx}',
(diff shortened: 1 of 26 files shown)


apps/web/panda.config.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export default defineConfig({
+  /**
+   * https://panda-css.com/docs/guides/component-library#include-the-src-files
+   *
+   * Per Panda docs, use the above approach if app code lives in an internal monorepo
+   */
+  include: [
+    './src/**/*.{js,jsx,ts,tsx}',
+    // '../../libs/design-system/src/**/*.{js,jsx,ts,tsx}'
```
antonjoel82, line 12: Leaving this as it will relevant for the future iteration

libs/design-system/src/color-scheme/useColorScheme.ts
```diff
@@ -0,0 +1,45 @@
+import { useState, useEffect, useCallback } from 'react';
+import { useLocalThemePreference } from '@novu/shared-web';
+import { useColorScheme as useMantineColorScheme } from '@mantine/hooks';
+import { ColorScheme } from './ColorScheme';
+import { mapThemeStatusToColorScheme } from './mapThemeStatusToColorScheme';
+
+/**
+ * Handle behavior for changing ColorSchemes or ThemeStatuses
+ */
+export const useColorScheme = () => {
```
antonjoel82, line 10: @BiswaViraj I ended up needing to modify the behavior of setting the color scheme, so I extracted it to this hook. It should still work with your "track" changes, but I will integrate it with your hook when we start bring our branches together

libs/design-system/src/panda/NovuPandaPreset.ts
```diff
@@ -27,16 +29,26 @@ export const NovuPandaPreset = definePreset({
     tokens: {
       sizes: SIZES_TOKENS,
       spacing: SPACING_TOKENS,
-      colors: COLOR_PALETTE_TOKENS,
+      colors: {
+        ...COLOR_PALETTE_TOKENS,
+        ...LEGACY_COLOR_TOKENS,
```
LetItRock, line 34: what are your thoughts on the legacy tokens? maybe there is no sense in exposing them and just refactoring the old code to the new colors?

apps/web/src/components/layout/components/PageHeader.tsx
```diff
@@ -1,11 +1,18 @@
 import { Group } from '@mantine/core';
-import { Title, Container } from '@novu/design-system';
+import { Container } from '@novu/design-system';
+import { css } from '../../../styled-system/css';
+import { styled } from '../../../styled-system/jsx';
+import { title as titleRecipe } from '../../../styled-system/recipes';
+
+const Title = styled('h2', titleRecipe);
```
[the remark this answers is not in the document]
antonjoel82: Correct, I could not get the CSS to generate properly in my first pass with writing a standalone component in Design System-- it was only generating styles for the default variant, even when all of the variants were used in storybook. However, this was out of scope anyway for the first phase, so it was a bit of a side quest to _try_ to make things easier.

libs/design-system/src/panda/NovuPandaPreset.ts
antonjoel82: I just added this last night because it will make our gradual adoption MUCH easier. Nikolay did not use the new values in the Information Architecture designs, which made things very difficult to implement. The way I did this, I'm excited that it should be a good user experience while keeping styles consistent:

- Devs just use the new semantic tokens (i.e. `surface.page`)
- The Preset has legacy overrides for the semantic tokens that point them to the legacy colors.
- Once we're ready to switch colors, we just get rid of the semantic-to-legacy mapping layer, and voila!

apps/web/src/components/layout/components/PageHeader.tsx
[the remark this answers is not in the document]
LetItRock: ahh yeah, now I remember you was mentioning this before :)

libs/design-system/src/panda/NovuPandaPreset.ts
LetItRock: awesome!

LetItRock (approved):

PR #5262, opened 2024-03-05
Create a script to increase the amount of storage space on default runner

Cliftonz:
This adds a script to the parts of the pipeline that do the docker build. 

This script is a temporary fix by removing unused and preinstalled tooling and removes apt caches.

### DIFF
--- a/.github/actions/docker/build-api/action.yml
+++ b/.github/actions/docker/build-api/action.yml
@@ -37,6 +37,9 @@ runs:
       shell: bash
       run: pnpm build:api
 
+    # TODO Removed when migrated to action matrix for each build type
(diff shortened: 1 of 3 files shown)


AliaksandrRyzhou (approved):
LGTM. I've added cspell new words. Do we have to add the same step to the production build? As I remember currently Gali faces this issue only with API service or we have to take care about other services too. How do you think? Anyway we can merge it and see as it goes on

scopsy:
@Cliftonz @AliaksandrRyzhou I see that we haven't tested this on all test.yml pipelines. https://github.com/novuhq/novu/actions/runs/8172265709/job/22342316168?pr=5253
Falling for me, can someone assist, please?

PR #5268, opened 2024-03-06
feat(api): search workflows by name or trigger identifier

LetItRock:

Allow searching workflows by name or trigger identifier.

Screenshot 2024-03-07 at 00 52 29
Screenshot 2024-03-07 at 00 52 59

### DIFF
--- a/apps/api/src/app/shared/dtos/pagination-request.ts
+++ b/apps/api/src/app/shared/dtos/pagination-request.ts
@@ -5,8 +5,8 @@ import { IsInt, Max, Min } from 'class-validator';
 import { Constructor } from '../types';
 
 export interface IPagination {
-  page?: number;
-  limit?: number;
+  page: number;
+  limit: number;
 }
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
@@ -18,7 +18,7 @@ export function PaginationRequestDto(defaultLimit = 10, maxLimit = 100): Constru
     })
     @Type(() => Number)
     @IsInt()
-    page?: number = 0;
+    page = 0;
 
     @ApiPropertyOptional({
       type: Number,
@@ -30,7 +30,7 @@ export function PaginationRequestDto(defaultLimit = 10, maxLimit = 100): Constru
     @IsInt()
     @Min(1)
     @Max(maxLimit)
-    limit?: number = defaultLimit;
+    limit = defaultLimit;
   }
 
   return PaginationRequest;
--- a/apps/api/src/app/shared/dtos/pagination-request.ts
+++ b/apps/api/src/app/shared/dtos/pagination-request.ts
@@ -2,10 +2,15 @@ import { ApiPropertyOptional } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
 import { IsInt, Max, Min } from 'class-validator';
 
-export type Constructor<I> = new (...args: any[]) => I;
+import { Constructor } from '../types';
+
+export interface IPagination {
+  page?: number;
+  limit?: number;
+}
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
-export function PaginationRequestDto(defaultLimit = 10, maxLimit = 100): Constructor<any> {
+export function PaginationRequestDto(defaultLimit = 10, maxLimit = 100): Constructor<IPagination> {
   class PaginationRequest {
     @ApiPropertyOptional({
       type: Number,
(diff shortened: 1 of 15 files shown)


libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -165,15 +165,31 @@ export class NotificationTemplateRepository extends BaseRepository<
...
-  async getList(organizationId: string, environmentId: string, skip = 0, limit = 10) {
-    const totalItemsCount = await this.count({ _environmentId: environmentId });
+  async getList(organizationId: string, environmentId: string, skip = 0, limit = 10, query?: string) {
+    let searchQuery: FilterQuery<NotificationTemplateDBModel> = {};
+    if (query) {
+      searchQuery = {
+        $or: [
+          { name: { $regex: regExpEscape(query), $options: 'i' } },
+          { 'triggers.identifier': { $regex: regExpEscape(query), $options: 'i' } },
+        ],
```
LetItRock, line 175: allow searching by the name or trigger identifier case-insensitive

libs/dal/src/repositories/notification-template/notification-template.schema.ts
```diff
@@ -240,6 +240,11 @@ notificationTemplateSchema.index({
   'triggers.identifier': 1,
 });
 
+notificationTemplateSchema.index({
+  _environmentId: 1,
+  name: 'text',
+});
```
LetItRock, line 246: the index for a more performant MongoDB full-text search

apps/api/src/app/subscribers/subscribers.controller.ts
```diff
@@ -485,10 +485,10 @@ export class SubscribersController {
       organizationId: user.organizationId,
       environmentId: user.environmentId,
       subscriberId: subscriberId,
-      page: query.page != null ? parseInt(query.page) : 0,
+      page: query.page != null ? parseInt(query.page as any) : 0,
```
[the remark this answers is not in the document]
djabarovgeorge: are we certain that PaginationRequestDto paras are validated in the transport dto phase, if we do then I think it should be saved to remove the parsing because if it is not number @IsInt() will throw an exception.

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: Sorry, i missed this part of the search implementation.
if we want to search by both name and identifier in the same query (by OR operator), as far as I know, we need to make sure we have an index for both (all cases) meaning:
* user query by name, index `_environmentId.name`
* user query by name and identifier, index `_environmentId.identifier.name` (covers `_environmentId.identifier` and  `_environmentId.identifier.name` )

This is because if we have only `_environmentId.identifier.name` we could not use it for queries by `name` because we would be missing the `identifier` in the index.

djabarovgeorge (commented):
Amazing work, left couple of comment for your review :)

libs/dal/src/repositories/notification-template/notification-template.repository.ts
LetItRock: Ok, I do understand what you mean. I will add it, thanks! 🙌 

I see that we also have `_organizationId.triggers.identifier`, but I couldn't find any references in the code where we use `_organizationId and identifier`. Do we need it?

Also, in Atlas `_environmentId_1_triggers.identifier_1` means that we just have to extend it right?

apps/api/src/app/shared/dtos/pagination-request.ts
```diff
@@ -2,10 +2,15 @@ import { ApiPropertyOptional } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
 import { IsInt, Max, Min } from 'class-validator';
 
-export type Constructor<I> = new (...args: any[]) => I;
+import { Constructor } from '../types';
+
+export interface IPagination {
+  page?: number;
+  limit?: number;
```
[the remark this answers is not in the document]
LetItRock: they are optional but have their defaults, meaning that during the dto transformation if the query param value doesn't exist it will be defined with defaults on dto... I'll update it

libs/dal/src/repositories/notification-template/notification-template.repository.ts
djabarovgeorge: The only places i found that depend on `_organizationId` in their query are blueprint-related which are located in:
* NotificationTemplateRepository.findBlueprintByTriggerIdentifier
* NotificationTemplateRepository.getBlueprintList
we could easily refactor the code to use environmentId instead, however for now I would keep the index so we won't scan the collections until we refactor the code.

Thats right we could extend `_environmentId_1_triggers.identifier_1` to `_environmentId_1_triggers.identifier_1_name`

djabarovgeorge (approved):
💪

PR #5276, opened 2024-03-07
feat(web): workflows allow searching by name or trigger identifier

LetItRock:

Workflows page: 
- added the search input component at the top of the workflows list
- allow searching by the name or trigger identifier
- clear functionality
- `query` parameter in the URL along with pagination params

https://github.com/novuhq/novu/assets/2607232/ca80309a-c4e7-4bae-a913-858d0aed1bd2


apps/web/src/components/layout/components/OrganizationSelect.tsx
```diff
@@ -42,7 +42,7 @@ export default function OrganizationSelect() {
       setLoadingSwitch(true);
       const token = await changeOrganization(organizationId);
       setToken(token);
-      await queryClient.refetchQueries();
+      await queryClient.clear();
```
LetItRock, line 45: when switching the organization in the dropdown we don't need to refetch the queries but rather clear them all, otherwise some of the queries might be in the fetching state resulting in no loading indicators for example on the workflows page

apps/web/src/hooks/usePaginatedQuery.ts
```diff
@@ -34,13 +35,16 @@ export const usePaginatedQuery = <TResponse extends object>({
   paginationOptions = {},
 }: UsePaginatedQueryProps<TResponse>) => {
   const { currentPageNumber, pageSize, setCurrentPageNumber, setPageSize } = usePaginationState(paginationOptions);
+  const { search, setSearch } = useSearchState(paginationOptions);
```
LetItRock, line 38: created the `useSearchState` hook that reads the `query` value from the URL query params and allows to updating it

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -156,15 +157,18 @@ function WorkflowListPage() {
...
     currentPageNumber,
+    pageSize,
+    search,
+    setSearch,
     setCurrentPageNumber,
     setPageSize,
-    pageSize,
   } = useTemplates({ areSearchParamsEnabled: true });
-  const isLoading = areNotificationGroupLoading || loading;
+  const [searchValue, setSearchValue] = useState(search ?? '');
```
LetItRock, line 171: the search value state for the input to reflect quickly what user types; the request, and URL updates are debounced

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -199,15 +205,28 @@ function WorkflowListPage() {
...
-  const handleCreateDigestDemoWorkflow = () => {
-    segment.track(TemplateAnalyticsEnum.TRY_DIGEST_CLICK);
-    createDigestDemoWorkflow();
-  };
-
   function onRowClick(row) {
     navigate(parseUrl(ROUTES.WORKFLOWS_EDIT_TEMPLATEID, { templateId: row.values._id }));
   }
 
+  const debouncedSearchChange = useDebounce((value: string) => setSearch(value), 2000);
```
LetItRock, line 212: the time amount is debatable

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -199,15 +205,28 @@ function WorkflowListPage() {
...
+
+  const onSearchClearClick = () => {
+    debouncedSearchChange.cancel();
+    setSearchValue('');
+    setSearch('');
+  };
```
LetItRock, line 218: on input clear button click - cancel the last debounced request and update value and make request right away

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -199,15 +205,28 @@ function WorkflowListPage() {
...
+
+  const onSearchChange: ChangeEventHandler<HTMLInputElement> = ({ target: { value } }) => {
+    if (value === '') {
+      onSearchClearClick();
+
+      return;
+    }
+    setSearchValue(value);
+    debouncedSearchChange(value);
+  };
```
LetItRock, line 228: on search change, update the state for input and debounce the request

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -221,72 +240,50 @@ function WorkflowListPage() {
       }}
     >
       <Container fluid sx={{ padding: '0 24px 8px 24px' }}>
-        {isTemplateStoreEnabled ? (
```
LetItRock, line 224: removed the template store FF, as we use it with `true` option for a while now

apps/web/src/pages/templates/WorkflowListPage.tsx
```diff
@@ -221,72 +240,50 @@ function WorkflowListPage() {
...
+            onAllTemplatesClick={openModal}
+          />
+          <SearchInput
+            value={searchValue}
+            placeholder="Type name or identifier..."
+            onChange={onSearchChange}
+            onClearClick={onSearchClearClick}
+            disabled={isSearchInputDisabled}
+            data-test-id="workflows-search-input"
+          />
```
LetItRock, line 261: the search input added

libs/design-system/src/pagination/usePaginationState.ts
```diff
@@ -35,13 +34,11 @@ export const usePaginationState = ({
...
-        [pageNumberParamName]: `${currentPageNumber}`,
-        [pageSizeParamName]: `${pageSize}`,
-      },
-      { replace: true }
-    );
+    const newSearchParams = new URLSearchParams(document.location.search);
+    newSearchParams.set(pageNumberParamName, `${currentPageNumber}`);
+    newSearchParams.set(pageSizeParamName, `${pageSize}`);
+
+    setSearchParams(newSearchParams, { replace: true });
```
LetItRock, line 41: the `usePaginationState` and `useSearchState` are both separate hooks that both update the query parameters. This code ensures that they are not overridden but all combined together. Otherwise, if the `useEffect` in the `useSearchState` hook is called after, it will override all the pagination query params.

antonjoel82 (changes requested):
Looks good overall! Nice job on separating the rendering from the logic, and making the logic reusable.

Code thoughts:
- Should we add a minimum string length before we fire the query? Generally I see 3 char min

Testing feedback:
- What is the expected behavior if the user types an illegal character? Currently it removes the entire search string. I feel like it would be better to just omit the illegal chars.
- We should change the `search` icon color when the input is disabled
- Query param stays even when empty (i.e. `...?query=`) --> could we clear it for better UX?
- Should we convert all query strings to lower case (in the URL)?
- Changing between environments removes the query string, but it doesn't between orgs

Design feedback (I think we can make a decision here, and consult Nik when he returns):
- I think the input will feel / look nicer if we have an active state that adds either a border or background color.
- We could add a hover state (to make it more discoverable), and use the same style as above

As for the flickering behavior you mentioned on Slack, I unfortunately think this is because our `WorkflowListPage` is monolithic and over-complex, so everything re-renders together on Organization change.

Screenshot 2024-03-08 at 12 03 11 PM

I suggest we acknowledge that minor visual glitch as acceptable for now (since users aren't going to rapidly click between orgs), and aim to tackle it in a future refactor that aims to break apart our larger pages into more performant, reusable patterns.

libs/design-system/src/search-input/useSearchState.ts
```diff
@@ -0,0 +1,43 @@
+import { useCallback, useEffect, useState } from 'react';
+import { useSearchParams } from 'react-router-dom';
+
+const URL_PARAM_NAME_SEARCH = 'query';
```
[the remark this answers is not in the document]
LetItRock: I thought about that as well, but then wanted to align it more with API params, but there is no need actually.

LetItRock:
Thanks for the review! 🙌 

> * Should we add a minimum string length before we fire the query? Generally I see 3 char min

There was no specific requirement on this from the product. I see that it might be even useful with 2 chars, but for now, I would say let's leave it like this.

> * What is the expected behavior if the user types an illegal character? Currently it removes the entire search string. I feel like it would be better to just omit the illegal chars.

All the characters are decoded in the URL, and we also don't have any restrictions on the API side for the workflow name, so the users can even call it `%test`. 
Can you provide an example? I've fixed one decoding issue that I introduced, but I haven't seen that the entire search was removed.

> * We should change the `search` icon color when the input is disabled

Mantine is not propagating the opacity on the icons :/ will add css manually

> * Query param stays even when empty (i.e. `...?query=`) --> could we clear it for better UX?

Yes, will update it.

> * Should we convert all query strings to lower case (in the URL)?

There is no need, as the API searches the DB case-insensitively. 

> * Changing between environments removes the query string, but it doesn't between orgs

The environment switch change redirects the user to the `/` => `/workflows` page, which basically resets all query parameters. The organization switch doesn't do that, but I think it should behave the same. Now it leads to the issue that the params are preserved and the user might not see the workflows if the new org doesn't have that many pages.

> Design feedback (I think we can make a decision here, and consult Nik when he returns):
> 
> * I think the input will feel / look nicer if we have an active state that adds either a border or background color.
> * We could add a hover state (to make it more discoverable), and use the same style as above
> 
> As for the flickering behavior you mentioned on Slack, I unfortunately think this is because our `WorkflowListPage` is monolithic and over-complex, so everything re-renders together on Organization change.

Yes, I agree it would be nice to have these micro-interactions on the search input so the user will have a better experience. Let's talk to Nikolay about that when he will be back.

About the `WorkflowListPage` yes it's gigantic and screams for the refactoring.

apps/web/src/api/notification-templates.ts
```diff
@@ -1,11 +1,19 @@
...
+  IPaginationWithQueryParams,
+} from '@novu/shared';
 
 import { api } from './api.client';
 import { BLUEPRINTS_API_URL } from '../config';
 
-export function getNotificationsList(page = 0, limit = 10) {
-  return api.getFullResponse(`/v1/notification-templates`, { page, limit });
+export function getNotificationsList({ page = 0, limit = 10, query }: IPaginationWithQueryParams) {
+  const params = { page, limit, ...(query && { query }) };
```
antonjoel82, line 12: **❔ question:** Is there another way we could write the query portion?‏

libs/design-system/src/search-input/SearchInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Input, InputProps } from '@mantine/core';
+import styled from '@emotion/styled';
+
+import { IconClose, IconSearch } from '../iconsV2';
+import { colors } from '../config';
+
+export interface ISearchInputProps {
+  placeholder?: string;
+  value?: string;
+  ref?: React.Ref<HTMLInputElement>;
```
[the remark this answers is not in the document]
antonjoel82: Ah, so is it meant to represent the type of the ref-forwarded component then?

antonjoel82 (approved):
Awesome, love the changes! Thank you for reusing interfaces from `shared` as well -- I think that's a great addition.

Everything looked good in my smoke testing 🎉

apps/web/src/api/notification-templates.ts
LetItRock: if statement or ternary

libs/design-system/src/search-input/SearchInput.tsx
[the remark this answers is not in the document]
LetItRock: exactly :)

PR #5278, opened 2024-03-08
Add Mongodb Atlas index recommendations

Cliftonz:

This PR adds a new set of indexs based on mongodb atlas recommendation.

### DIFF
--- a/libs/dal/src/repositories/message-template/message-template.schema.ts
+++ b/libs/dal/src/repositories/message-template/message-template.schema.ts
@@ -89,6 +89,13 @@ messageTemplateSchema.index({
   'triggers.identifier': 1,
 });
 
+messageTemplateSchema.index({
+  _environmentId: 1,
+  _layoutId: 1,
+  _feedId: 1,
+  deleted: 1,
+});
+
 messageTemplateSchema.plugin(mongooseDelete, { deletedAt: true, deletedBy: true, overrideMethods: 'all' });
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
--- a/libs/dal/src/repositories/message/message.schema.ts
+++ b/libs/dal/src/repositories/message/message.schema.ts
@@ -286,6 +286,28 @@ messageSchema.index({
   deleted: 1,
 });
 
+/*
+ * This index was initially created to optimize:
+ *
+ * Path: apps/api/src/app/messages/usecases/get-messages.usecase.ts
+ * Context: execute()
+ * Query:  const query: Partial<Omit<MessageEntity, 'transactionId'>> & { _environmentId: string; transactionId?: string[] } =
+ *    {
+ *      _environmentId: command.environmentId,
+ *    };
+ *
+ *  if (command.subscriberId) {
+ *    const subscriber = await this.getSubscriberUseCase.execute(
+ *      GetSubscriberCommand.create({
+ *        subscriberId: command.subscriberId,
+ *        environmentId: command.environmentId,
+ *        organizationId: command.organizationId,
+ *      })
+ *    );
+ *
+ *    query._subscriberId = subscriber._id;
+ *  }
+ */
 messageSchema.index({
   _environmentId: 1,
   _subscriberId: 1,
--- a/libs/dal/src/repositories/message/message.schema.ts
+++ b/libs/dal/src/repositories/message/message.schema.ts
@@ -286,6 +286,15 @@ messageSchema.index({
   deleted: 1,
 });
 
+messageSchema.index({
+  _environmentId: 1,
+  _subscriberId: 1,
+  channel: 1,
+  seen: 1,
+  read: 1,
+  deleted: 1,
+});
+
 // eslint-disable-next-line @typescript-eslint/naming-convention
 export const Message =
   (mongoose.models.Message as mongoose.Model<MessageDBModel>) ||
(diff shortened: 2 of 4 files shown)


libs/dal/src/repositories/message-template/message-template.schema.ts
djabarovgeorge, line 1: I know it is not related to your change but do we need the index above? 
```
index({
  _organizationId: 1,
  'triggers.identifier': 1,
});
```

it feels redundant as we don't have `triggers` under `messageTemplate`.

libs/dal/src/repositories/message-template/message-template.schema.ts
```diff
@@ -89,6 +89,13 @@ messageTemplateSchema.index({
   'triggers.identifier': 1,
 });
 
+messageTemplateSchema.index({
+  _environmentId: 1,
+  _layoutId: 1,
+  _feedId: 1,
+  deleted: 1,
```
djabarovgeorge, line 96: Is `deleted` auto added by `mongoose-delete` on every query?

libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -283,6 +283,43 @@ messageSchema.index({
...
+ *  }
+ */
+messageSchema.index({
+  _environmentId: 1,
+  _subscriberId: 1,
+  channel: 1,
+  seen: 1,
+  read: 1,
+  deleted: 1,
+});
```
djabarovgeorge, line 318: small suggestion so we will have it a bit more compact.

i made it as a workaround I wonder if we could have a better way to visualize the source of the index creation.

```suggestion
 * Query:  find({
 *    {
 *        environmentId: command.environmentId,
 *        _subscriberId: command.subscriberId,
 *        channel = command.channel;
 *        transactionId = command.transactionIds;
 *    });
 *
 *    query._subscriberId = subscriber._id;
 *  }
 */
messageSchema.index({
  _environmentId: 1,
  _subscriberId: 1,
  channel: 1,
  seen: 1,
  read: 1,
  deleted: 1,
});
```

libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -283,6 +283,43 @@ messageSchema.index({
...
+ *
+ *    query._subscriberId = subscriber._id;
+ *  }
+ */
+messageSchema.index({
+  _environmentId: 1,
+  _subscriberId: 1,
+  channel: 1,
+  seen: 1,
+  read: 1,
```
djabarovgeorge, line 316: I do not see the `seen` and `read` I the query of getMessges metod, do we add it in some phase?

libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -283,6 +283,43 @@ messageSchema.index({
...
+ *        organizationId: command.organizationId,
+ *      })
+ *    );
+ *
+ *    query._subscriberId = subscriber._id;
+ *  }
+ */
+messageSchema.index({
+  _environmentId: 1,
+  _subscriberId: 1,
```
djabarovgeorge, line 313: IMO the subscriber should be before the environment because it has a bigger cardinality than the environment. were we recommended to add it in this order?

libs/dal/src/repositories/message/message.schema.ts
djabarovgeorge, line 1: I think that the new index should be considered alongside the existing index:
```
messageSchema.index({
  _subscriberId: 1,
  _environmentId: 1,
  channel: 1,
  seen: 1,
  read: 1,
  createdAt: -1,
});
```

because it was created to solve the `get-messages.usecase` use case.

libs/dal/src/repositories/message/message.schema.ts
```diff
@@ -283,6 +283,43 @@ messageSchema.index({
...
+  channel: 1,
+  seen: 1,
+  read: 1,
+  deleted: 1,
+});
+
+messageSchema.index({
+  _environmentId: 1,
+  _organizationId: 1,
 });
```
djabarovgeorge, line 323: Why is this index needed? i think we could manage to query by environment unless there is a use-case for this query.

libs/dal/src/repositories/notification-template/notification-template.schema.ts
```diff
@@ -248,6 +248,14 @@ notificationTemplateSchema.virtual('notificationGroup', {
...
   _organizationId: 1,
   'triggers.identifier': 1,
+  deleted: 1,
+});
+
+notificationTemplateSchema.index({
+  _environmentId: 1,
+  _organizationId: 1,
+  createdAt: -1,
+  deleted: 1,
```
djabarovgeorge, line 258: It seems that Pawel plans to create a new index in this collection. I think it would be better to combine the work if possible to minimize workload for both us and the database.

It's worth noting that including organization in this index may not necessarily optimize the query performance. Since the query primarily filters by environment first, sorting by the organization might not add significant value, in addition, we need to consider that if the organization is missing in the query, other parameters like createdAt and deleted won't be utilized in the query by MongoDB.

libs/dal/src/repositories/subscriber/subscriber.schema.ts
djabarovgeorge, line 1: What query requires this index?

Similarly, when considering the `_organizationId` parameter in such a large collection, we must weigh the trade-off between having a fully covered index and the associated RAM costs. Do you have any best practices or rules of thumb regarding this matter? because for me it is super hard to decide sometimes.

djabarovgeorge (commented):
Amazing work 💪

libs/dal/src/repositories/message-template/message-template.schema.ts
Cliftonz: Yes it does

PR #5283, opened 2024-03-08
[Information Architecture] - New Sidebar Nav

antonjoel82:

- New components for the Sidebar Nav
- Refactors some existing components to decouple them
- Adds a couple more frequently used tokens to the Design Tokens

- Closes [NV-3438](https://linear.app/novu/issue/NV-3438/%F0%9F%8F%9B%EF%B8%8F-main-navigation-left-menu)

## Notes

A few places do not match designs where they were inconsistent with other patterns or unclear
- Background Color of the icon in the `Environment` selector
- A few icons in designs do not match what is available in our current icon set, so I used the closest approximation
- Added parentheses around the Org Name in the Settings menu section headers
  - _The decision to add org name here was very last minute, and so it didn't go through the normal review_
- Using the `<--` to go back to the previous _non-settings_ page would require some special handling, and wasn't explicitly covered in the requirements. My recommendation is to just return the user to the home (`/`) page if they use the nested menu back button.

### Future Work

- Next PR: tests, visibility toggle
- Handled in other feature PRs: first-class routes for the new standalone pages (i.e. `settings/organization`)

Screenshot 2024-03-11 at 10 10 16 PM
Screenshot 2024-03-11 at 10 10 00 PM
Screenshot 2024-03-11 at 10 11 17 PM
Screenshot 2024-03-11 at 10 10 39 PM


apps/web/.eslintrc.js
```diff
@@ -36,7 +36,8 @@ module.exports = {
       },
     ],
     "react-hooks/rules-of-hooks": 'error',
-    "react-hooks/exhaustive-deps": 'warn'
+    "react-hooks/exhaustive-deps": 'warn',
+    "import/extensions": 'off',
```
antonjoel82, line 40: Would love an alternative suggestion if anyone has one! This rule failed repeatedly in CI for the Panda `styled-system` imports, but none of the tweaks I made seemed to deactivate the rule. ([example commit](https://github.com/novuhq/novu/pull/5283/commits/a8dcdb1aab805c99ea3be804bdc7d6a1d02a6d6f))

apps/web/src/SettingsRoutes.tsx
```diff
@@ -0,0 +1,92 @@
...
+import { ROUTES } from './constants/routes.enum';
+import { useFeatureFlag } from './hooks';
+import { BillingRoutes } from './pages/BillingPages';
+import { BrandingForm } from './pages/brand/tabs';
+import { MembersInvitePage as MembersInvitePageNew } from './pages/invites/v2/MembersInvitePage';
+import { SettingsPage } from './pages/settings/SettingsPage';
+import { ApiKeysCard } from './pages/settings/tabs';
+import { EmailSettings } from './pages/settings/tabs/EmailSettings';
+
+/** Note: using a hook is the only way to separate routes */
```
BiswaViraj, line 14: Other way will be use `Routes` to wrap the `Route` and return it as `<SettingsRoutes/>`.
We do something similar in our enterprise packages.

BiswaViraj: also another way will be to  migrate to [useRoutes](https://reactrouter.com/en/main/hooks/use-routes) hook of `react-router-dom`. It will enable us to use objects to define the routes.

apps/web/src/components/nav/EnvironmentSelect/EnvironmentPopover.tsx
```diff
@@ -0,0 +1,86 @@
...
+const closeButtonStyles = css({
+  position: 'absolute',
+  cursor: 'pointer',
+  top: '7px',
+  zIndex: '2',
+  right: '10px',
+  '& svg': {
+    fill: 'typography.text.main',
+  },
+});
```
BiswaViraj, line 42: we can move the styles to a `.styles` file

BiswaViraj (approved):
Amazing work 🚀 
Left a small comment.

apps/web/src/components/nav/NavMenuButton/NavMenuLinkButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const NavMenuLinkButton: FC<PropsWithChildren<INavMenuLinkButtonProps>> = ({
+  rightSide,
+  testId,
+  icon,
+  link,
+  label,
+  isVisible = true,
+}) => {
+  const [popoverOpened, setPopoverOpened] = useState(false);
+  const [isHovered, setIsHovered] = useState<boolean>(false);
```
LetItRock, line 36: not used?

apps/web/src/components/nav/NavMenuButton/NavMenuToggleButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+}) => {
+  const [isOpen, setIsOpen] = useState<boolean>(false);
+
+  const handleClick = () => {
+    setIsOpen((prevIsOpen) => !prevIsOpen);
+  };
+
+  return (
+    <>
+      <button className={css(rawButtonBaseStyles)} data-test-id={testId} onClick={handleClick}>
```
LetItRock, line 23: missing cursor pointer

apps/web/src/components/nav/OrganizationSelect/OrganizationSelect.tsx
```diff
@@ -0,0 +1,50 @@
...
+      data={data}
+    />
+  );
+};
+
+/**
+ * @deprecated
+ */
+export default function OrganizationSelect() {
+  return <OrganizationSelectRenderer {...useOrganizationSelect()} />;
```
LetItRock, line 49: there is a room for improvements in this hook, but let's leave this for later

apps/web/src/components/nav/RootNavMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { ROUTES } from '@novu/shared-web';
+import { ChangesCountBadge } from '../layout/components/ChangesCountBadge';
+import { EnvironmentSelect } from './EnvironmentSelect';
+import { NavMenu } from './NavMenu';
+import { NavMenuLinkButton } from './NavMenuButton/NavMenuLinkButton';
+import { NavMenuSection } from './NavMenuSection';
+import { OrganizationSelect } from './OrganizationSelect/v2/OrganizationSelect';
+import { RootNavMenuFooter } from './RootNavMenuFooter';
+
+export const RootNavMenu: React.FC = () => {
```
LetItRock, line 23: love the way it's clear, simple and extendable :)

LetItRock (changes requested):
Really incredible amount of work you put into this!  🤝 
I think it easily could eat even more time, but I think that we can iterate over some parts (of old code) later, it's in much much better shape than it was. Hats down 🎩 

Let's just tackle a few issues/suggestions that I've found:

1. When opening the org dropdown something is appearing undeneath.

https://github.com/novuhq/novu/assets/2607232/516ea5b7-f664-42de-aa91-5f75a8184bd7

2. The org/env icons are not aligned with the menu items and seems that are overflowing the select.
Screenshot 2024-03-14 at 17 47 35

3. We need to have a gap between the menu items, but on smaller screens it won't work, like there is not enough space.
Screenshot 2024-03-14 at 17 49 41

4. The settings menu when the env accordion is opened on smaller screens the menu items gets cut off and not reachable, no scroll available.
Screenshot 2024-03-14 at 17 57 35

5. The env dropdown should behave similarly to organization dropdown and keep the user on the same page, or navigate to the root.

https://github.com/novuhq/novu/assets/2607232/f6b9981d-10f5-4acc-88c5-e5a46836641c

6. Seems like after switching the org/env we are making same requests twice (just noticed it's how it is currently on prod too, let's ignore for now).

Screenshot 2024-03-14 at 18 14 52

7.  Missing the popover that should be shown when the org select is opened.
Screenshot 2024-03-14 at 22 22 57

apps/web/src/components/nav/EnvironmentSelect/useEnvironmentSelect.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { setEnvironment, isLoading, environment, readonly } = useEnvController({
+    onSuccess: (newEnvironment) => {
+      setIsPopoverOpened(!!newEnvironment?._parentId);
+    },
+  });
+
+  async function handlePopoverLinkClick(e) {
+    e.preventDefault();
+
+    await setEnvironment('Development');
```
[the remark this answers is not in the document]
antonjoel82: Agreed -- however, I couldn't find a good "source" for these.

apps/web/src/components/nav/NavMenuButton/NavMenuLinkButton.tsx
antonjoel82: They're used in the next phase since we need the one-off visibility icon

apps/web/src/components/nav/OrganizationSelect/OrganizationSelect.tsx
antonjoel82: Agreed, there's too much legacy code that needs refactor for the scope of this feature. Extracting it to a standalone hook was the start :)

antonjoel82:
@LetItRock I believe I addressed all the changes requested -- the one exception being:

> We need to have a gap between the menu items, but on smaller screens it won't work, like there is not enough space.

Nikolay had explicitly mentioned not having a gap between items. I'm okay if we want to iterate on that later, but I think we can discuss it during the final project review

LetItRock (commented):
Looks much better, but found a few more things.
1. There is now a gap between the main navigation and the sidebar, we should change this variable `NAVIGATION_WIDTH` to equal `272` or better to connect it with `sidebarStyle`.
Screenshot 2024-03-19 at 18 27 37

2. Switching environment / organization doesn't work for the Translation Groups and Tenant Edit Sidebar, like it supposed to redirect to the root page, like Translation page or Tenants Page.

https://github.com/novuhq/novu/assets/2607232/fe1cc884-aabf-42bd-96da-e0cb42c7bc27

3. Not sure if it's in this PR but the hover background color has changed for the `HeaderNav` 
Screenshot 2024-03-19 at 18 34 56

4. The shadow on the nav items is cut, check the comment about the overflow.
Screenshot 2024-03-19 at 19 03 15

apps/web/src/components/nav/NavMenu.tsx
```diff
@@ -0,0 +1,86 @@
...
+  '&:disabled svg': {
+    opacity: '40%',
+  },
+});
+
+const navStyles = css({
+  h: '100%',
+  display: 'flex',
+  flexDirection: 'column',
+  overflowY: 'auto',
```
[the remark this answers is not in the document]
antonjoel82: Great catch, thank you!

apps/web/src/components/nav/OrganizationSelect/v2/OrganizationSelect.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <Tooltip
+      label="Type a name to add organization"
+      position="left"
+      classNames={{
+        tooltip: tooltipStyles,
+        arrow: arrowStyles,
+      }}
+      withinPortal
+    >
```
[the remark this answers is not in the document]
antonjoel82: I added this for now, but I believe this tooltip is an indicator of a poor UX pattern to begin with, and we should ultimately change this Organization component

antonjoel82:
@LetItRock Thanks again for the thorough review! I believe I have addressed all this feedback except the green coloring that should be handled separately.

Apologies, but I was unable to verify the gap in Translations locally because Translations throws an error every time I try to create a group. It seems to be due to locales, but I didn't see relevant info for fixing it in Notion nor Slack

LetItRock:
> Apologies, but I was unable to verify the gap in Translations locally because Translations throws an error every time I try to create a group. It seems to be due to locales, but I didn't see relevant info for fixing it in Notion nor Slack

No need to apologize, all good! you did awesome work on this, many things changed for the better! 🙌
About the Translations Management, I suggest you to run these steps:
```
1. From the root folder: `npm run clean` and then after remove the `node_modules`.
2. `pnpm i`
3. `pnpm run symlink:submodules`
4. `nx run-many --target=build --all --skip-nx-cache` 
```
Usually, these steps help to mitigate the issues with enterprise packages.

apps/web/src/components/nav/OrganizationSelect/v2/OrganizationSelect.tsx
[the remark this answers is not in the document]
LetItRock: agree 💯

LetItRock (approved):
amazing work! 🙌

PR #5290, opened 2024-03-10
Editor autocomplete - does not open in a second use

ainouzgali:
In the editor, after adding a {{..}} once. Writing only one { opens the autocomplete dropdown, when adding the second { (so its {{ ) the dropdown does not show again.

The slice was wrong, at the second input of {{ it took the whole content and so used the indexes of the first {{


BiswaViraj:
It seems to not work after adding an empty line in between

https://github.com/novuhq/novu/assets/29251776/4841f8e3-0576-444b-9732-3744b9e69dd0

BiswaViraj (approved):
Amazing 🚀

PR #5295, opened 2024-03-12
Fix: Unhandled Error: Objects are not valid as a React child

ainouzgali:

The error: Objects are not valid as a React child (found: object with keys {type, content, styles}). If you meant to render a collection of children, use an array instead.

It happens when a user has email content for editor type and in an update he switches to custom html but leaves the content empty, the problem originates from a bug where we didn't allow to update content to be empty => so contentType is updated to customHtml but the content value is left as an object (in the editor content is an array) 

It also resulted that a content that had value but then updated to be empty will still have the original value.


apps/api/src/app/message-template/usecases/update-message-template/update-message-template.usecase.ts
```diff
@@ -33,7 +33,7 @@ export class UpdateMessageTemplate {
       updatePayload.name = command.name;
     }
 
-    if (command.content) {
+    if (command.content !== null) {
```
ainouzgali, line 36: Allow the update of an empty string

LetItRock (approved):
thank you for the fix @ainouzgali! 🙌

djabarovgeorge (commented):
sorry for the late response, just noticed that I did not submit the review I had 🤦‍♂️

apps/api/src/app/message-template/usecases/update-message-template/update-message-template.usecase.ts
ainouzgali: Thanks @djabarovgeorge, added the undefined in this [PR](https://github.com/novuhq/novu/pull/5302)

PR #5296, opened 2024-03-12
fix: provide notification group on creation by blueprint

djabarovgeorge:

When creating a blueprint template, internal IDs like _notificationGroupId and _feedId are utilized. However, using these IDs may result in errors when generating templates for environments other than DEV and US, as the blueprint is extracted from these specific environments.

Scenario of the issue:
An EU user receives a notification template (workflow) blueprint containing internal IDs. This blueprint is extracted from the US-Centralized blueprint storage. Upon attempting to create this workflow in the EU-API, the system encounters an error while trying to extract entities by their IDs, as these entities don’t exist in the EU environment.

### Definition of Done
* In order to make this flow simpler we could pass the existing default notification group from the existing environment.
* The US-Centralized storage should provide a workflow that propagates with nested entities without any internal IDs.
* Create tests that will use centralized storage.


LetItRock (approved):
hey @djabarovgeorge, approving it! 🙌 

The only thing that is bothering me is our decision to have a "central" place for managing the blueprints. Don't you think that it might be much simpler to maintain if we decide to split it to EU and US, like each region will be independent. Like to me I'm not sure if it's worth to have such logic and maintain it, if I can just spend few mins and create the same set of templates in any EU/US region. WDYT?

djabarovgeorge:
> hey @djabarovgeorge, approving it! 🙌
> 
> The only thing that is bothering me is our decision to have a "central" place for managing the blueprints. Don't you think that it might be much simpler to maintain if we decide to split it to EU and US, like each region will be independent. Like to me I'm not sure if it's worth to have such logic and maintain it, if I can just spend few mins and create the same set of templates in any EU/US region. WDYT?

I believe it's better to keep the blueprints separate from the user environment. This means we shouldn't include any internal user identifiers in the blueprints. By keeping the creation of a workflow stateless and isolated within the user's environment, we prevent any dependencies on user-specific information. Providing isolated blueprints ensures those bugs don't occur.

IMO managing two sources of truth could lead to additional maintenance and potential issues with alignment and future bugs. please note that in the future we could have more regions which could lead to more maintenance for us.

PR #5298, opened 2024-03-12
[Information Architecture] - Nav Part 2

antonjoel82:

- Visibility toggle for Get Started
- Cypress test for Main Nav

- [NV-3438](https://linear.app/novu/issue/NV-3438/%F0%9F%8F%9B%EF%B8%8F-main-navigation-left-menu)


apps/web/src/components/nav/VisibilityButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const buttonStyles = css({
+  bg: 'transparent',
+  transition: 'background 0.2s ease',
+  _hover: {
+    '& svg': {
+      fill: 'typography.text.secondary',
+    },
+  },
+  borderRadius: '50',
```
BiswaViraj, line 12: is it meant to be `50%` ?

BiswaViraj (approved):

apps/web/src/components/nav/VisibilityButton.tsx
antonjoel82: It is not -- "50" is one of the `size` values we have in the design-system, but in this case, it doesn't seem to be getting picked up by borderRadius

LetItRock (approved):
🙌

PR #5300, opened 2024-03-13
fix(web): workflow settings - providers warnings were cut off

LetItRock:

Fixed the issue in the Workflow Settings - Providers tab, that the channel "warning banners" were cut off.

https://github.com/novuhq/novu/assets/2607232/02d7cecf-2a9d-46b8-94c3-fc0a4e14c711

### DIFF
--- a/apps/web/src/pages/templates/components/ListProviders.tsx
+++ b/apps/web/src/pages/templates/components/ListProviders.tsx
@@ -1,8 +1,10 @@
+import { MouseEventHandler, useMemo } from 'react';
+import styled from '@emotion/styled';
 import { Group, Stack, Text, UnstyledButton, useMantineColorScheme } from '@mantine/core';
 import { ChannelTypeEnum, NOVU_SMS_EMAIL_PROVIDERS } from '@novu/shared';
+import { Button, colors, Tooltip } from '@novu/design-system';
 
 import { When } from '../../../components/utils/When';
-import { Button, colors, Tooltip } from '@novu/design-system';
 import { useEnvController } from '../../../hooks';
 import { IntegrationEnvironmentPill } from '../../integrations/components/IntegrationEnvironmentPill';
 import { IntegrationStatus } from '../../integrations/components/IntegrationStatus';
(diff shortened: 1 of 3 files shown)


apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -11,165 +13,142 @@ import { stepNames } from '../constants';
...
+      ),
+    [channelProviders]
+  );
+  const providersForTheCurrentEnvironment = useMemo(
+    () =>
+      channelProviders.filter((provider) => provider.connected && provider.environmentId === currentEnvironment?._id),
+    [channelProviders, currentEnvironment?._id]
+  );
+  const hasProviders = providersForTheCurrentEnvironment.length > 0;
+  const hasNovuProvider = providersForTheCurrentEnvironment.length === 1 && containsNovuProvider;
```
LetItRock, line 87: Under the Providers tab, we are showing all the integrations from the current environment, so the one change that I introduced is the logic to show the warnings is based on the assumption of the current environment.

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -11,165 +13,142 @@ import { stepNames } from '../constants';
...
-      >
-        <Group position="apart">
-          <ChannelTitle spacing={8} channel={channel} />
-          <Button
-            sx={{
-              height: '32px',
-              padding: '7.5px 15px',
-            }}
-            variant={providers.filter((provider) => provider.connected).length > 0 ? 'outline' : 'gradient'}
+    <ListProvidersContainer>
```
LetItRock, line 91: I could spend probably more time refactoring it, but I don't think it's worth it and to me, the current state is better than what we had before.

apps/web/src/pages/templates/components/ListProviders.tsx
antonjoel82, line 1: **💭 thought:** ‏This could be a great opportunity to use `panda` instead! It should simplify some of this CSS

apps/web/src/pages/templates/components/ListProviders.tsx
```diff
@@ -11,165 +13,142 @@ import { stepNames } from '../constants';
...
+      NOVU_SMS_EMAIL_PROVIDERS.some(
+        (providerId) => providerId === channelProviders.find((provider) => provider.connected)?.providerId
+      ),
+    [channelProviders]
+  );
+  const providersForTheCurrentEnvironment = useMemo(
+    () =>
+      channelProviders.filter((provider) => provider.connected && provider.environmentId === currentEnvironment?._id),
+    [channelProviders, currentEnvironment?._id]
+  );
```
antonjoel82, line 85: **🎯 suggestion:** These are more or less "selectors," and could be written and reused as standalone functions outside of this component and file. This will help us from re-inventing the wheel because I imagine we have an instance or two of this implemented elsewhere in the app
‏

antonjoel82 (approved):
LGTM, thanks!

apps/web/src/pages/templates/components/ListProviders.tsx
LetItRock: I thought about it, but wasn't sure how should I map the current colors to the new color palette.

apps/web/src/pages/templates/components/ListProviders.tsx
LetItRock: We don't because they are specific to this case, but we can extract it into more reusable code.

PR #5302, opened 2024-03-14
Translation with params not included in payload generation

ainouzgali:
Params used within a translation are not being recognized as workflow variables and are not included in the created payload trigger. 
example: `{{i18n "test.key" var=payload-var}}` => `payload-var` should be included in the trigger snippet :
`novu.trigger(......, payload: { payload-var: ...})`

Screen Shot 2024-03-14 at 12 11 25
Screen Shot 2024-03-14 at 12 11 48


apps/api/src/app/message-template/usecases/update-message-template/update-message-template.usecase.ts
```diff
@@ -33,7 +33,7 @@ export class UpdateMessageTemplate {
       updatePayload.name = command.name;
     }
 
-    if (command.content !== null) {
+    if (command.content !== null || command.content !== undefined) {
```
ainouzgali, line 36: Related to a different [PR](https://github.com/novuhq/novu/pull/5295#discussion_r1523214826), added it here after a comment there

libs/shared/src/consts/handlebar-helpers/getTemplateVariables.ts
```diff
@@ -10,27 +10,48 @@ export interface IMustacheVariable {
...
+      }
+
+      return {
+        type: TemplateVariableTypeEnum.STRING,
+        name: body.value?.original as string,
+        defaultValue: '',
+        required: false,
+      };
+    });
+
```
ainouzgali, line 29: This could be used in the future for other handlebars helpers that we might want to allow variables assignments in.

libs/shared/src/consts/handlebar-helpers/getTemplateVariables.ts
```diff
@@ -10,27 +10,48 @@ export interface IMustacheVariable {
...
+    .flatMap((body) => {
+      const varName = body.value?.original as string;
+
+      if (!shouldAddVariable(varName)) {
+        return [];
+      }
+
+      return {
+        type: TemplateVariableTypeEnum.STRING,
+        name: body.value?.original as string,
```
LetItRock, line 24:
```suggestion
        name: varName,
```

libs/shared/src/consts/handlebar-helpers/getTemplateVariables.ts
LetItRock: maybe we should extract it to a reusable function then?

libs/shared/src/consts/handlebar-helpers/getTemplateVariables.ts
```diff
@@ -10,27 +10,48 @@ export interface IMustacheVariable {
...
+
   const stringVariables: IMustacheVariable[] = bod
     .filter((body) => body.type === 'MustacheStatement')
     .flatMap((body) => {
       const varName = body.params[0]?.original || (body.path.original as string);
 
-      if (body.path.original === HandlebarHelpersEnum.I18N) {
+      if (body.path?.original === HandlebarHelpersEnum.I18N) {
+        if (body.hash?.pairs) {
+          return getTemplateVariables(body.hash.pairs);
```
LetItRock, line 37: I'm not sure if we need a recurrent call here? 
Can't we just use the extracted above function (lines 13-29) to get the variables?

ainouzgali: Currently, yes, you are right. But when we have time, in the cooldown maybe I want to go back to it and add support for sub expressions. In i18n, but also for general use. And that will need the recurrent calls.
For example:
`{{titlecase (uppercase msg) }}` (not a useful example, but just to show my point)
or for i18n 
`{{i18n "group.key" var=(titlecase msg) }}`

`msg` in both examples would not be recognized as a variable, and in the first example it would set `uppercase` as a variable instead. 

I do have something working for those states, but it requires some more work. I didn't want to over do it with the time as I feel its not the biggest priority

LetItRock: ok, makes sense now :)

LetItRock (approved):

PR #5305, opened 2024-03-15
[Information Architecture] API Keys Page

antonjoel82:

- Create a page with refactored components for the API Keys
- Per [discussion with Dima](https://novu.slack.com/archives/C06E70CF0E5/p1710465576052029), we removed the ability to interact with API Keys for environments other than the current one
- While we had scoped it out initially, we were able to add "Regenerate" API Key functionality

- Closes https://linear.app/novu/issue/NV-3456/%F0%9F%8F%9B%EF%B8%8F-settings-api-keys

### Notes

- Added the environment in parentheses for page titles of `API Keys` and `Webhooks` since there's not a good way to tell otherwise. This was not in designs, but is highly necessary IMO. Happy to take other suggestions for how to show it!

https://www.loom.com/share/985d9efd140a482caadd310e0a63a3ae?sid=b20627dd-7566-40a5-9cc7-a9f8619bfa4a

_Yes, I did regenerate my API Key after this 😅_


apps/web/src/SettingsRoutes.tsx
```diff
@@ -15,8 +16,13 @@ import { EmailSettings } from './pages/settings/tabs/EmailSettings';
 export const useSettingsRoutes = () => {
   const isInformationArchitectureEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_INFORMATION_ARCHITECTURE_ENABLED);
 
-  return isInformationArchitectureEnabled ? (
+  /*
+   * FIXME: reload doesn't work properly with conditional routes,
+   * so replace with isInformationArchitectureEnabled before merging
+   */
+  return true ? (
```
antonjoel82, line 23: I would like to merge this into the main `information-architecture` branch as it makes dev much easier for the time being

apps/web/src/pages/settings/ApiKeysPage/useApiKeysPage.ts
```diff
@@ -0,0 +1,54 @@
...
+   * TODO: we will eventually want to use the URL params instead of the current environment, but
+   * this will come at a later time once we have the APIs to support it.
+   *
+   * const { env } = useParams<ApiKeysPageUrlParams>();
+   */
+
+  const clipboardApiKey = useClipboard({ timeout: CLIPBOARD_TIMEOUT_MS });
+  const clipboardEnvironmentIdentifier = useClipboard({ timeout: CLIPBOARD_TIMEOUT_MS });
+  const clipboardEnvironmentId = useClipboard({ timeout: CLIPBOARD_TIMEOUT_MS });
+  const { data: apiKeys, refetch: refetchApiKeys } = useQuery<{ key: string }[]>(['getApiKeys'], getApiKeys);
```
LetItRock, line 28: not used by the `ApiKeysRenderer`

apps/web/src/pages/settings/ApiKeysPage/ApiKeysPage.tsx
```diff
@@ -0,0 +1,109 @@
...
+  );
+};
+
+export const ApiKeysPage = () => {
+  const props = useApiKeysPage();
+
+  return (
+    <SettingsPageContainer title={`API keys (${props.pageEnv})`}>
+      <ApiKeysRenderer {...props} />
+    </SettingsPageContainer>
```
LetItRock, line 107: I'm not sure about this pattern, like I see the value when we have something that could be reusable, but IMO we create more hassle with props drilling, `ApiKeysPage -> ApiKeysRenderer -> ConfirmRegenerationModal`. Just the my opinion, don't take it as a must.

LetItRock (approved):
looks good! 🙌

apps/web/src/pages/settings/ApiKeysPage/ApiKeysPage.tsx
antonjoel82: I can certainly understand that! There are two things that I'd add to considerations here:

1. **General**: I agree we are not likely to reuse the whole page content. However, this is not only for reusability, but also makes things much easier to test when doing integration tests (since we can pass whatever we want as props and validate against that).
2. **Specific**: While this `ConfirmRegenerationModal` may not be particularly applicable to other situations, I think our implementation strategy for modals is currently broken, and this case shows a major part of it. Our modals should not have their "roots" placed on each page IMO. Instead, we could have a `ModalProvider` that is included in the base layout and register modals by a key. Then we can call: `showModal(ModalKey.REGENERATE_API_KEYS)`. More realistically, we'd see that most modals fall into a few main varieties -- like confirmation (yes / no), acknowledgement (ok), etc. Then we can more simply do `showModal(ModalKey.CONFIRMATION, { /* props */ })` rather than writing one-off components.

PR #5308, opened 2024-03-17
Nv 3567 subscribers are duplicated during the burst of events

djabarovgeorge:

During the burst of trigger events the processed subscribers might get duplicated in the database. This could happen due to the race conditions in the worker's concurrent processing.

Duplicated subscribers mean that the wrong document might be picked from the database and for example some fields might be missing (email), resulting in messages not being delivered.

We have to introduce the unique index on the subscriberId and environmentId combo to prevent and fix this problem.

### DIFF
--- a/apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
+++ b/apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
@@ -1,0 +1,121 @@
+import { TopicSubscribersRepository, SubscriberRepository } from '@novu/dal';
+import { SubscribersService, UserSession } from '@novu/testing';
+
+import { removeDuplicatedSubscribers } from './remove-duplicated-subscribers.migration';
+import { expect } from 'chai';
+
+describe('Migration: Remove Duplicated Subscribers', () => {
+  let session: UserSession;
+  let subscriberService: SubscribersService;
+  const subscriberRepository = new SubscriberRepository();
+  const topicSubscribersRepository = new TopicSubscribersRepository();
+
+  beforeEach(async () => {
+    session = new UserSession();
+    await session.initialize();
+    subscriberService = new SubscribersService(session.organization._id, session.environment._id);
+  });
+
+  it('should remove duplicated subscribers', async () => {
+    const duplicatedSubscriberId = '123';
+
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'first_subscriber',
+      _environmentId: session.environment._id,
+      _organizationId: session.organization._id,
+    });
+
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'mid_subscriber',
+      _environmentId: session.environment._id,
+      _organizationId: session.organization._id,
+    });
+
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'last_subscriber',
+      _environmentId: session.environment._id,
+      _organizationId: session.organization._id,
+    });
+
+    const duplicates = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: session.environment._id,
+    });
+
+    expect(duplicates.length).to.equal(3);
+
+    await removeDuplicatedSubscribers();
+
+    const remainingDuplicates = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: session.environment._id,
+    });
+
+    expect(remainingDuplicates.length).to.equal(1);
+    expect(remainingDuplicates[0].firstName).to.equal('last_subscriber');
+  });
+
+  it('should always keep one subscriber per environment', async () => {
+    const duplicatedSubscriberId = '123';
+    const firstEnvironmentId = session.environment._id;
+
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'env_1',
+      _environmentId: firstEnvironmentId,
+      _organizationId: session.organization._id,
+    });
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'env_1',
+      _environmentId: firstEnvironmentId,
+      _organizationId: session.organization._id,
+    });
+
+    const secondEnvironmentId = session.organization._id;
+
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'env_2',
+      _environmentId: secondEnvironmentId,
+      _organizationId: session.organization._id,
+    });
+    await subscriberRepository.create({
+      subscriberId: duplicatedSubscriberId,
+      firstName: 'env_2',
+      _environmentId: secondEnvironmentId,
+      _organizationId: session.organization._id,
+    });
+
+    const duplicates = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: session.environment._id,
+    });
+    expect(duplicates.length).to.equal(2);
+
+    const duplicates2 = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: session.environment._id,
+    });
+    expect(duplicates2.length).to.equal(2);
+
+    await removeDuplicatedSubscribers();
+
+    const remainingDuplicates = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: session.environment._id,
+    });
+    expect(remainingDuplicates.length).to.equal(1);
+    expect(remainingDuplicates[0].firstName).to.equal('env_1');
+
+    const remainingDuplicates2 = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: secondEnvironmentId,
+    });
+    expect(remainingDuplicates2.length).to.equal(1);
+    expect(remainingDuplicates2[0].firstName).to.equal('env_2');
+  });
+});
--- a/apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
+++ b/apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
@@ -3,6 +3,7 @@ import { SubscribersService, UserSession } from '@novu/testing';
 
 import { removeDuplicatedSubscribers } from './remove-duplicated-subscribers.migration';
 import { expect } from 'chai';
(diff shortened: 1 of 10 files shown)


apps/ws/src/shared/subscriber-online/subscriber-online.service.ts
```diff
@@ -30,7 +30,7 @@ export class SubscriberOnlineService {
 
   private async updateOnlineStatus(subscriber: ISubscriberJwt, updatePayload: IUpdateSubscriberPayload) {
     await this.subscriberRepository.update(
-      { _id: subscriber._id, _organizationId: subscriber.organizationId },
+      { _id: subscriber._id, _environmentId: subscriber.environmentId },
```
djabarovgeorge, line 33: Does anyone know if there is a reason to query by organization id here?
Updating it to environment id means that we can remove the _organizaitionId index from the DB.

libs/dal/src/repositories/subscriber/subscriber.schema.ts
```diff
@@ -10,12 +10,10 @@ const subscriberSchema = new Schema<SubscriberDBModel>(
     _organizationId: {
       type: Schema.Types.ObjectId,
       ref: 'Organization',
-      index: true,
     },
     _environmentId: {
       type: Schema.Types.ObjectId,
       ref: 'Environment',
-      index: true,
```
djabarovgeorge, line 18: redundant index 🧹

apps/ws/src/shared/subscriber-online/subscriber-online.service.ts
scopsy: Let's remove it.

libs/dal/src/repositories/subscriber/subscriber.schema.ts
```diff
@@ -165,11 +163,14 @@ subscriberSchema.index({
...
-subscriberSchema.index({
-  subscriberId: 1,
-  _environmentId: 1,
-  _id: 1,
-});
+subscriberSchema.index(
+  {
+    subscriberId: 1,
+    _environmentId: 1,
+    _id: 1,
```
scopsy, line 170: I think the `_id` part here can create duplicates still, as the id will always be random and unique. 

Mongodb already have a unique id index by default, So I think it should be 👍 
```
 subscriberId: 1,
    _environmentId: 1,
```

Wdyt?

djabarovgeorge: Good point, i think the main trade off here is if we want full coverage on the query or reduce memory, this is a big collection so I believe it will be significant.
looking at it again I wonder if mongo db would perform Index Prefix Intersection with both 
{subscriberId: 1, _environmentId: 1} and {_id} indexes.

LetItRock:
@djabarovgeorge sorry that I didn't noticed that the PR #5309  was pointing to this one

djabarovgeorge:
> @djabarovgeorge sorry that I didn't noticed that the PR #5309 was pointing to this one

No worries both of them are my PR's are they are small, so there won't be any issue with it ;)

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
djabarovgeorge, line 1: These tests won't work with the new index because it won't allow to create of new documents with duplicated indexes.

i ran them locally without the new index because I was not able to find a better way to manipulate the indexes from the API without creating an additional connection to the DB.

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.ts
```diff
@@ -0,0 +1,71 @@
...
+  ];
+
+  const cursor = await subscriberRepository._model.aggregate(pipeline, {
+    batchSize: batchSize,
+    readPreference: 'secondaryPreferred',
+    allowDiskUse: true,
+  });
+
+  for (const group of cursor) {
+    const docsToRemove = group.docs.slice(0, -1); // Keep the last created document, remove others
```
djabarovgeorge, line 43: At the moment I decided to keep the last created subscriber, does someone know other practices for removing invalid entries?

libs/dal/src/repositories/subscriber/subscriber.repository.ts
```diff
@@ -10,6 +10,8 @@ import type { EnforceEnvOrOrgIds } from '../../types';
 import { EnvironmentId, ISubscribersDefine, OrganizationId } from '@novu/shared';
 
 type SubscriberQuery = FilterQuery<SubscriberDBModel> & EnforceEnvOrOrgIds;
+type SubscriberDeleteQuery = Pick<SubscriberQuery, 'subscriberId' | '_environmentId'> & EnforceEnvOrOrgIds;
+type SubscriberDeleteManyQuery = Pick<SubscriberQuery, 'subscriberId' | '_id' | '_environmentId'> & EnforceEnvOrOrgIds;
```
djabarovgeorge, line 14: i decided to create those types because, without those types, the client of the methods could think that other query params are in use in the query while they are not at the moment.

libs/dal/src/repositories/subscriber/subscriber.repository.ts
```diff
@@ -151,21 +153,31 @@ export class SubscriberRepository extends BaseRepository<SubscriberDBModel, Subs
...
       throw new DalException(`Could not find subscriber ${query.subscriberId} to delete`);
     }
 
-    const requestQuery: SubscriberQuery = {
-      _environmentId: foundSubscriber._environmentId,
-      subscriberId: foundSubscriber.subscriberId,
+    return await this.subscriber.delete(requestQuery);
+  }
+
+  async deleteMany(query: SubscriberDeleteManyQuery) {
```
djabarovgeorge, line 171: we need `deleteMany` override method in order to perform `soft delete`

libs/dal/src/types/error.enum.ts
```diff
@@ -0,0 +1,3 @@
+export enum ErrorCodesEnum {
+  DUPLICATE_KEY = '11000',
+}
```
djabarovgeorge, line 3: I made this enum as a temporary solution. It helps the application layer recognize errors from the data access layer without being dependent on MongoDB. But I'm considering whether it's better to generate our own error codes and link them to the application layer for better control and clarity.

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.ts
```diff
@@ -0,0 +1,150 @@
...
+          'lastOnlineAt',
+        ].includes(key)
+      ) {
+        continue;
+      }
+
+      // Update with non-null/undefined values from subsequent subscribers
+      if (currentSubscriber[key] !== null && currentSubscriber[key] !== undefined) {
+        mergedSubscriber[key] = currentSubscriber[key];
+      }
```
LetItRock, line 145: ⚠ **issue:** I think that this part might be problematic for the `channels` field, which is an array, so the latter will override the first. Should we have here more custom logic and compare channels by `integrationId` and merge all the unique `deviceTokens` together? wdyt?

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    try {
+      await subscriberRepository.update(
+        {
+          _id: mergedSubscriber._id,
+          subscriberId: subscriberId,
+          _environmentId: environmentId,
+        },
+        {
+          $set: mergedSubscriber,
```
LetItRock, line 71: ❓ **question:** shouldn't we just update some fields like `firstName, lastName, email, channels...` etc., without `_id, _environmentId, _organizationId` etc?

libs/dal/src/repositories/subscriber/subscriber.schema.ts
```diff
@@ -165,15 +164,31 @@ subscriberSchema.index({
...
+ * We expect an exception to be thrown when attempting to create two subscribers with the same subscriberId (e.g., 2022) within the same environment.
+ */
+index(
+  {
+    subscriberId: 1,
+    _environmentId: 1,
+    deleted: 1,
+  },
+  { unique: true }
+);
```
LetItRock, line 183: ❓ **question:** Wouldn't this index be an issue in this case:
The client already has deleted subscriber `{ subscriberId: '123', _environmentId: '123', deleted: true }`. Then he will create a new one with the same values `{ subscriberId: '123', _environmentId: '123', deleted: false }`. And I think that the index will fail when he tries to delete that new subscriber because he already has one deleted.

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    const remainingDuplicates = await subscriberRepository.find({
+      subscriberId: duplicatedSubscriberId,
+      _environmentId: session.environment._id,
+    });
+
+    expect(remainingDuplicates.length).to.equal(1);
+    expect(remainingDuplicates[0]._id).to.equal(firstCreatedSubscriber._id);
+  });
+});
```
LetItRock, line 264: ✏ **suggestion:** it would be nice to see some tests with the properly merged `channels`

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.ts
djabarovgeorge: Here, I wasn't sure how specific we wanted to be because we might say the same with 'data.' but, since we're uncertain about the type of data being stored, we could check by object size. Yet again, I wasn't certain if this level of detail was what we were aiming for.

libs/dal/src/repositories/subscriber/subscriber.schema.ts
djabarovgeorge: That's a valid observation! We should address this scenario through multiple approaches:

1. If one subscriber was already deleted, perform a hard delete on the old one, followed by a soft delete on the newly deleted subscriber.
2. revisit the necessity of soft deleting subscribers altogether.
3. Remove the 'deleted' status from the index.

In my opinion, it's reasonable to proceed with option 3. The number of deleted subscribers under the same subscriberId and environmentId should be relatively small, so the optimization gained from adding 'deleted' in the index might not be significant.

djabarovgeorge: Done

apps/api/migrations/subscribers/remove-duplicated-subscribers/remove-duplicated-subscribers.migration.spec.ts
djabarovgeorge: @LetItRock Done :)

LetItRock (approved):

PR #5309, opened 2024-03-17
feat(dal): add deleted index and type safety index

djabarovgeorge:

We want to optimize the call checking for non deleted subscribers.
This change is recommended by Mongodb Atlas.
As an alternative, the query may need to be changed.

This will increase the query speed of calls looking for non deleted subscribers and will increase the trigger speed.

### DIFF
--- a/libs/dal/src/repositories/subscriber/subscriber.schema.ts
+++ b/libs/dal/src/repositories/subscriber/subscriber.schema.ts
@@ -1,9 +1,10 @@
 import * as mongoose from 'mongoose';
-import { Schema } from 'mongoose';
+import { IndexOptions, Schema } from 'mongoose';
 import * as mongooseDelete from 'mongoose-delete';
 
 import { schemaOptions } from '../schema-default.options';
-import { SubscriberDBModel } from './subscriber.entity';
+import { SubscriberDBModel, SubscriberEntity } from './subscriber.entity';
+import { IndexDefinition } from '../../shared/types';
 
 const subscriberSchema = new Schema<SubscriberDBModel>(
   {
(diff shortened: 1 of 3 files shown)


libs/dal/src/repositories/subscriber/subscriber.schema.ts
```diff
@@ -178,3 +181,7 @@ subscriberSchema.plugin(mongooseDelete, { deletedAt: true, deletedBy: true, over
 export const Subscriber =
   (mongoose.models.Subscriber as mongoose.Model<SubscriberDBModel>) ||
   mongoose.model<SubscriberDBModel>('Subscriber', subscriberSchema);
+
+function index(fields: IndexDefinition<SubscriberEntity>, options?: IndexOptions) {
+  subscriberSchema.index(fields, options);
```
djabarovgeorge, line 186: @rifont How about such an approach? while developing I noticed we don't have our indexes typed, meaning we can do indexed typos.
Doing this as a POC, I tried to make it a bit more generic but TS gave me problems :/

LetItRock: `mongoose` is so poor when it comes to the types... yes the problems might be because we force some fields...
I like this approach 🙌

LetItRock (approved):

libs/dal/src/repositories/subscriber/subscriber.schema.ts
rifont: I like it too, let's revisit generic typings as part of a holistic approach to strongly typing Mongoose. Ship it 🚀

rifont (approved):

PR #5314, opened 2024-03-18
feat: optimize organization queries

djabarovgeorge:

* Stop using single indexes on the mongoose key implementation
* Add compound indexes for common queries and patterns using MongoDB index best practices
* Remove unused indexes from MongoDB atlas after implementing the new compound indexes


libs/dal/src/repositories/organization/organization.repository.ts
```diff
@@ -12,25 +12,24 @@ export class OrganizationRepository extends BaseRepository<OrganizationDBModel,
...
     return this.mapEntity(data.toObject());
   }
 
-  async findOrganizationById(organizationId: string): Promise<OrganizationEntity | null> {
-    const data = await this.MongooseModel.findById(organizationId).read('secondaryPreferred');
-    if (!data) return null;
+  async findUserActiveOrganizations(userId: string): Promise<OrganizationEntity[]> {
+    const organizationIds = await this.getUsersMembersOrganizationIds(userId);
 
-    return this.mapEntity(data.toObject());
```
djabarovgeorge, line 25: @scopsy do you remember why we created new `findOrganizationById` that reads from secondaryPreferred for the widget API route? instead of reusing the existing `findById` and updating its query to use secondaryPreferred?

At the moment I do not see reasons for duplication, and all `findById` can use secondaryPreferred.

libs/dal/src/repositories/notification/notification.schema.ts
```diff
@@ -140,12 +140,39 @@ notificationSchema.index({
...
+ *          'notifications.createdAt': {
+ *            $gte: command.startDate,
+ *            $lt: command.endDate,
+ *          },
+ *        }
+ */
+notificationSchema.index({
+  createdAt: -1,
+});
+
```
djabarovgeorge, line 175: I'm unsure about adding this. I'd like to hear what others think.

Here's why we might consider it:
- We require this index for our hourly corn job. Without it, we'll need to do a full collection scan.

However, there are drawbacks to consider:
- The collection is quite large. Adding this index will impact all queries, not just the hourly one, by consuming a large amount of ram in the database.

libs/dal/src/repositories/organization/organization.repository.ts
scopsy: Sounds good

scopsy (approved):
Looking good! Would just remove `createdAt` index

libs/dal/src/repositories/notification/notification.schema.ts
djabarovgeorge: totally agree :)

PR #5319, opened 2024-03-19
fix: Surface all auth server errors during login in email input.

SokratisVidros:

This PR patches the following issue:

1. Sign up with Github with user A
2. Sign out
3. Try to sign in with the email of user A
4. The API returns a 400 error not displayed in the Login form right away in smaller screens.

After this PR, all server auth errors will be displayed in the UI.

It also contains some minor copywriting fixes and clean-ups.

#### Before

https://github.com/novuhq/novu/assets/1352422/674ac4ef-7914-419d-bf48-99d3e163a752

#### After

Screenshot 2024-03-20 at 00 39 54

### DIFF
--- a/apps/api/src/app/auth/e2e/login.e2e.ts
+++ b/apps/api/src/app/auth/e2e/login.e2e.ts
@@ -13,154 +13,179 @@ describe('User login - /auth/login (POST)', async () => {
     password: '123Qwerty@',
   };
 
-  before(async () => {
-    session = new UserSession();
-    await session.initialize();
+  context('with email/password', async () => {
+    before(async () => {
+      session = new UserSession();
+      await session.initialize();
 
-    const { body } = await session.testAgent
-      .post('/v1/auth/register')
-      .send({
+      const { body } = await session.testAgent
+        .post('/v1/auth/register')
+        .send({
+          email: userCredentials.email,
+          password: userCredentials.password,
+          firstName: 'Test',
+          lastName: 'User',
+        })
+        .expect(201);
+    });
+
+    it('should login the user correctly', async () => {
+      const { body } = await session.testAgent.post('/v1/auth/login').send({
         email: userCredentials.email,
         password: userCredentials.password,
-        firstName: 'Test',
-        lastName: 'User',
-      })
-      .expect(201);
-  });
-
-  it('should login the user correctly', async () => {
-    const { body } = await session.testAgent.post('/v1/auth/login').send({
-      email: userCredentials.email,
-      password: userCredentials.password,
-    });
-
-    const jwtContent = (await jwt.decode(body.data.token)) as IJwtPayload;
-
-    expect(jwtContent.firstName).to.equal('test');
-    expect(jwtContent.lastName).to.equal('user');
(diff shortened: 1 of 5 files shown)


apps/api/src/app/auth/e2e/login.e2e.ts
```diff
@@ -13,154 +13,179 @@ describe('User login - /auth/login (POST)', async () => {
...
       expect(body.statusCode).to.equal(401);
-    }
+      expect(body.message).to.contain('Account blocked');
+    });
+  });
 
-    const { body } = await session.testAgent.post('/v1/auth/login').send({
-      email: userCredentials.email,
-      password: userCredentials.password,
+  context('with OAuth', async () => {
```
SokratisVidros, line 169: All the above ☝️ are just indentation changes after adding a `context`.

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -78,29 +79,28 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
     }
   };
 
-  const serverErrorString = useMemo<string>(() => {
-    return Array.isArray(error?.message) ? error?.message[0] : error?.message;
-  }, [error]);
+  const emailClientError = errors.email?.message;
+  let emailServerError = parseServerErrorMessage(error);
```
SokratisVidros, line 83: Similar changes can be applied to the register flow.

apps/api/src/app/auth/usecases/login/login.usecase.ts
```diff
@@ -41,7 +41,8 @@ export class Login {
       throw new UnauthorizedException(`Account blocked, Please try again after ${blockedMinutesLeft} minutes`);
     }
 
-    if (!user.password) throw new ApiException('OAuth user login attempt');
+    // TODO: Trigger a password reset flow automatically for existing OAuth users instead of throwing an error
+    if (!user.password) throw new ApiException('Please sign in using Github.');
```
SokratisVidros, line 45: Note that returning different errors per auth flow results in User enumeration issues. User enumeration is mostly a privacy concern and should be tackled in a separate PR.

SokratisVidros: Moreover, the error message mentions Github specifically. Currently, it's the only supported OAuth method. When more methods are added, the copywriting should be tweaked.

apps/api/src/app/auth/e2e/login.e2e.ts
scopsy: A new one for me! What is the difference between context and describe?

SokratisVidros: `context` is a synonym for `describe` that is usually available in BDD style APIs. For example https://www.coreycleary.me/better-test-structuring-using-mochas-context

apps/api/src/app/auth/usecases/login/login.usecase.ts
LetItRock: Btw we have the Google OAuth logic but currently, it's enterprise-only and "disabled".

LetItRock (approved):
🙌

PR #5325, opened 2024-03-20
[Information Architecture] User Profile | Set Password Flow

antonjoel82:

- Adds context to the User Profile page to make sharing state easier
- Adds sidebar management through URL params
- Provides a foundation for the "Update Password Flow'
 
- Closes [NV-3445](https://linear.app/novu/issue/NV-3445/%F0%9F%8F%9B%EF%B8%8F-settings-user-profile-set-a-password)

https://www.loom.com/share/b75e812bd60c4fb09dd8a93d64c7b381?sid=4bb24aff-8c88-46f4-8e57-c70a55e10ace

### DIFF
--- a/apps/api/src/app/auth/usecases/password-reset-request/password-reset-request.usecase.ts
+++ b/apps/api/src/app/auth/usecases/password-reset-request/password-reset-request.usecase.ts
@@ -38,7 +38,7 @@ export class PasswordResetRequest {
 
       if ((process.env.NODE_ENV === 'dev' || process.env.NODE_ENV === 'production') && process.env.NOVU_API_KEY) {
         const novu = new Novu(process.env.NOVU_API_KEY);
-        const resetPasswordLink = PasswordResetRequest.getResetRedirectLink(token, command.src);
+        const resetPasswordLink = PasswordResetRequest.getResetRedirectLink(token, foundUser, command.src);
 
         novu.trigger(process.env.NOVU_TEMPLATEID_PASSWORD_RESET || 'password-reset-llS-wzWMq', {
           to: {
@@ -57,14 +57,19 @@ export class PasswordResetRequest {
     };
   }
 
-  private static getResetRedirectLink(token: string, src?: PasswordResetFlowEnum): string {
-    switch (src) {
-      case PasswordResetFlowEnum.USER_PROFILE:
-        return `${process.env.FRONT_BASE_URL}/settings/profile?token=${token}`;
-      case PasswordResetFlowEnum.FORGOT_PASSWORD:
-      default:
-        return `${process.env.FRONT_BASE_URL}/auth/reset/${token}`;
+  private static getResetRedirectLink(token: string, user: UserEntity, src?: PasswordResetFlowEnum): string {
+    // ensure that only users without passwords are allowed to reset
+    if (src === PasswordResetFlowEnum.USER_PROFILE && !user.password) {
+      return `${process.env.FRONT_BASE_URL}/settings/profile?token=${token}`;
     }
+
+    /**
+     * Default to the existing "forgot password flow". Works for:
+     * 1. No src
+     * 2. When src is explicitly FORGOT_PASSWORD
+     * 3. User already has a password
+     */
+    return `${process.env.FRONT_BASE_URL}/auth/reset/${token}`;
   }
 
   private isRequestBlocked(user: UserEntity) {
--- a/apps/web/src/pages/settings/user-profile-page/UserProfilePasswordEmailVerificationSection.tsx
+++ b/apps/web/src/pages/settings/user-profile-page/UserProfilePasswordEmailVerificationSection.tsx
@@ -3,11 +3,10 @@ import { css, cx } from '../../../styled-system/css';
 import { HStack, Stack } from '../../../styled-system/jsx';
 import { text } from '../../../styled-system/recipes';
 import { IUserProfilePasswordEmailVerificationProps } from './UserProfilePasswordSidebar.shared';
-import { Text } from './UserProfilePasswordSidebar';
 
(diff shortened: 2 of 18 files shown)


apps/web/src/pages/settings/user-profile-page/UserProfilePasswordSidebarEnum.ts
```diff
@@ -0,0 +1,7 @@
+export enum UserProfileSidebarTypeEnum {
```
antonjoel82, line 1: Chose an enum so that we could easily support other use cases

libs/shared/src/utils/checkIsResponseError.ts
```diff
@@ -0,0 +1,5 @@
+import { IResponseError } from '../types';
```
antonjoel82, line 1: Currently have a PR out for this to merge it separately: https://github.com/novuhq/novu/pull/5338

This file will disappear once that gets merged and we rebase.

apps/web/src/pages/settings/user-profile-page/UserProfilePasswordEmailVerificationSection.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <IconOutlineMarkunreadMailbox size="40" className={css({ flexShrink: 0 })} />
+        <p className={cx(text(), css({ color: 'typography.text.secondary' }))}>
+          We have just sent a verification link to your email address{' '}
+          <span className={text({ variant: 'main' })}>{email}</span>.<br />
+          Please verify your email address to proceed with setting a password.
+        </p>
+      </HStack>
+      <HStack gap={'200'} justifyContent="flex-end">
+        <span className={text()}>
+          Didn't get the link? Resend in{' '}
```
BiswaViraj, line 27: **suggestion:** When the timer stops, `Resend in 0 seconds` looks odd. Maybe we can just write `resend` or `resend now` when the timer is 0‏

antonjoel82: Yeah, we didn't have designs that handled this, and agreed. However, I tried it, and it causes the text to jump which I don't think looks great either. We can get more input in the UX hand-off

apps/api/src/app/auth/usecases/password-reset-request/password-reset-request.usecase.ts
```diff
@@ -55,6 +57,16 @@ export class PasswordResetRequest {
     };
   }
 
+  private static getResetRedirectLink(token: string, src?: PasswordResetFlowEnum): string {
+    switch (src) {
+      case PasswordResetFlowEnum.USER_PROFILE:
+        return `${process.env.FRONT_BASE_URL}/settings/profile?token=${token}`;
```
BiswaViraj, line 63: **suggestion:** I think we should add a step of validation here to check only the users with no password available can access this API.
Currently, anyone can access it.‏

apps/web/src/pages/settings/user-profile-page/useUserProfileSetPassword.ts
```diff
@@ -0,0 +1,58 @@
...
+      email: string;
+    }
+  >((data) => api.post(`/v1/auth/reset/request`, data, { src: PasswordResetFlowEnum.USER_PROFILE }), {
+    onError: (err) => {
+      showNotification({
+        color: 'red',
+        message: `There was an error sending your email. Please refresh and try again. Error: ${err.message}`,
+      });
+    },
+  });
```
BiswaViraj, line 33: **todo:** You can use `errorMessage` and `successMessage` notification helpers from our design-system‏

apps/web/src/pages/settings/user-profile-page/useUserProfileSetPassword.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const useUserProfileSetPassword = () => {
+  const { currentUser } = useAuthContext();
+  const {
+    timeRemaining: countdownTimerSeconds,
+    resetTimer,
+    stopTimer,
+  } = useCountdownTimer({
+    // start at 0 in the event the user opens the page directly to the set password flow via URL
+    startTimeSeconds: 0,
```
BiswaViraj, line 17: **thought (non-blocking):** This can go both ways. I see the reason why it has to be 0 by default. But this also allows the user to refresh the page to reset the timer every time.‏

apps/web/src/pages/settings/user-profile-page/useUserProfileSetPassword.ts
BiswaViraj: image
The long error message also looks bad on the UI. Not sure if it using the one from the design-system will make it better.

BiswaViraj (changes requested):
Awesome work 🙌🏼 

Left a few comments. let me know wdyt

Also, I think we should close the sidebar after successful password set, wdyt?

apps/web/src/pages/settings/user-profile-page/useUserProfileSetPassword.ts
antonjoel82: Ultimately we are handling a back-end concern solely through the front-end here, but we do have a safeguard for total number of password requests I believe. I agree it's not ideal, but we can see what comes up in UX feedback

apps/web/src/pages/settings/user-profile-page/useUserProfileSetPassword.ts
antonjoel82: Thanks for letting me know about these utils! I hadn't seen them, and the code I was looking at used `showNotification` directly. Will update

BiswaViraj (approved):

PR #5327, opened 2024-03-21
feat: add product feature interceptor and hook

davidsoderberg:


apps/api/src/app/shared/interceptors/product-feature.interceptor.ts
```diff
@@ -0,0 +1,51 @@
...
+        handler,
+        classRef,
+      ]);
+
+      const enabled = productFeatureEnabledForServiceLevel[requestedFeature].includes(
+        organization?.apiServiceLevel as ApiServiceLevelEnum
+      );
+
+      if (!enabled) {
+        throw new HttpException('Payment Required', 402);
```
rifont, line 37: Is there a `PaymentRequiredError` class we can use out of the box here?

apps/api/src/app/shared/interceptors/product-feature.interceptor.ts
```diff
@@ -0,0 +1,51 @@
...
+export class ProductFeatureInterceptor implements NestInterceptor {
+  constructor(private reflector: Reflector, private organizationRepository: OrganizationRepository) {}
+
+  async intercept(context: ExecutionContext, next: CallHandler): Promise<Observable<any>> {
+    try {
+      const user = this.getReqUser(context);
+
+      const { organizationId } = user;
+
+      const organization = await this.organizationRepository.findById(organizationId);
```
rifont, line 23: We should probably throw an error when organization is not found, and remove the optional chaining below.

apps/api/src/app/shared/interceptors/product-feature.interceptor.ts
davidsoderberg: No I tried to find one but could not find one and it is not listed here: https://docs.nestjs.com/exception-filters#built-in-http-exceptions

apps/api/src/app/shared/interceptors/product-feature.interceptor.ts
```diff
@@ -0,0 +1,71 @@
...
+      }
+
+      const { organizationId } = user;
+
+      const organization = await this.organizationRepository.findById(organizationId);
+
+      if (!organization) {
+        throw new NotFoundException();
+      }
+
```
rifont, line 51: Thinking back on this one, I don't think it's required because the Auth Guard has already ensured the orgnaization must be present. We can remove.

```suggestion
```

rifont (approved):
LGTM!

PR #5328, opened 2024-03-21
.github: change github runners for e2e web/widget to blacksmith

adityamaru:
This change is a proof of concept to show if blacksmith runners can be faster for some of the longest running pipelines in novu's CI.

This is as per a conversation with Dima over email!


adityamaru:
There is a setup-project/action.yml file that we could point to our much faster Blacksmith cache (https://docs.blacksmith.sh/getting-started/blacksmith-cache) but that can only be done once all workflows relying on this action file is on blacksmith runners.

I would need someone to approve the workflows and also touch the relevant files to get `Test Web Cypress` and `Test Widget Cypress` running!

scopsy:
Closing for now 🙏

PR #5329, opened 2024-03-21
feat(web): setting user profile page - security section

LetItRock:

Information Architecture:
The User Profile Page - security section. For the rest waiting on Biswa to finish with the reusable component.

Screenshot 2024-03-21 at 22 26 24
Screenshot 2024-03-21 at 22 26 16
Screenshot 2024-03-21 at 22 25 51


apps/api/src/app/user/usecases/base-user-profile.usecase.ts
```diff
@@ -0,0 +1,36 @@
...
+      lastName,
+      email,
+      profilePicture,
+      createdAt,
+      showOnBoarding,
+      servicesHashes,
+      jobTitle,
+      hasPassword: !!password,
+    };
+  }
```
antonjoel82, line 35: **💭 thought:** Alternative, though not particularly important!‏
```suggestion
  protected mapToDto({ password, /* other exclusions */ ...userWithoutPw }: UserEntity): UserResponseDto {

    return {
      ...userWithoutPw,
      hasPassword: !!password,
    };
  }
```

apps/web/src/pages/settings/components/InputPlain.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  '& label': {
+    margin: '0 !important',
+  },
+  '& input': {
+    border: 'none !important',
+    paddingLeft: '0 !important',
+    height: '125 !important',
+    minHeight: '125 !important',
+    color: 'typography.text.secondary !important',
```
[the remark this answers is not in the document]
antonjoel82: Perhaps we can look into something with the layers or global resets like [here](https://github.com/chakra-ui/panda/discussions/1728)?

apps/web/src/pages/settings/user-profile-page/UserProfilePage.tsx
```diff
@@ -1,11 +1,53 @@
...
 export const UserProfilePage: FC = () => {
+  const { currentUser } = useAuthContext();
+
   return (
     <SettingsPageContainer title="User profile">
-      <div>TODO: replace with page content</div>
+      <Title mb="100" variant="section">
+        Profile security
+      </Title>
+      <div className={css({ maxWidth: '37.5rem' })}>
```
antonjoel82, line 23:
```suggestion
      <section className={css({ maxWidth: '37.5rem' })}>
```

libs/design-system/src/panda/semanticColors.tokens.ts
```diff
@@ -42,6 +42,14 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
        */
     },
   },
+  button: {
+    text: {
+      outline: {
+        value: { base: '{colors.legacy.gradientMiddle}', _dark: '{colors.legacy.white}' },
```
antonjoel82, line 48: 👏

antonjoel82 (approved):
Looks good, thanks! Are there any tests we could add to validate the back-end `hasPassword` addition?

antonjoel82 (commented):
@LetItRock FYI, I'm getting a build error in DAL

apps/api/src/app/user/usecases/base-user-profile.usecase.ts
LetItRock: The thing is that with this approach whenever you add something to the `UserEntity` it will be automatically available in the `UserResponseDto` even if the fields are not defined in the DTO, with this entity it's better to control which fields are exposed.

antonjoel82: Very good call given the use case! Thanks for the explanation!

PR #5331, opened 2024-03-22
feat: image component

BiswaViraj:
- added re-usable image preview and upload component


apps/web/src/components/shared/ProfileImage.tsx
```diff
@@ -0,0 +1,72 @@
...
+  control: Control<any>;
+  /**
+   * The name of the image input field that is used in react-hook-form
+   */
+  name: string;
+};
+
+type UploadInputContainerProps = Pick<ProfileImageProps, 'control' | 'onImageSubmit' | 'name'>;
+
+export function ProfileImage({ onImageSubmit, imageUrl, control, name }: ProfileImageProps) {
```
LetItRock, line 32: This component should accept: `ref, name, value, onChange` props which will be passed by the parent component, that will wrap it with the `Controller` or `useController` hook.

apps/web/src/components/shared/useProfileImageForm.ts
```diff
@@ -0,0 +1,38 @@
...
+  const { field, fieldState } = useController({
+    name,
+    control,
+  });
+
+  const handleOnSubmit = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const file = e.target.files && e.target.files.length > 0 ? e.target.files[0] : null;
+
+    // check if file is not an image
+    if (file && !file.type.includes('image')) {
```
LetItRock, line 23: We accept mime types `image/png, image/jpeg` but here are not validating that.

apps/web/src/components/shared/useProfileImageForm.ts
```diff
@@ -0,0 +1,38 @@
...
+      errorMessage('Invalid file type');
```
LetItRock, line 24: I think that we should call here `onError` prop, which will make give more flexibility

LetItRock (commented):
left you a few comments, props for the tests! 🙌

LetItRock (approved):
looks good! 🙌

PR #5332, opened 2024-03-22
feat: Tweak semantic color tokens

antonjoel82:

- Adds a couple more semantic tokens 

- To make working with design-system easier (and more backwards compatible)

### DIFF
--- a/libs/design-system/src/panda/semanticColors.tokens.ts
+++ b/libs/design-system/src/panda/semanticColors.tokens.ts
@@ -12,7 +12,7 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
       type: 'color',
     },
     popover: {
-      value: { base: '{colors.mauve.120.light}', _dark: '{colors.mauve.120.dark}' },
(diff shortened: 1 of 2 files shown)


libs/design-system/src/panda/semanticColors.tokens.ts
```diff
@@ -30,16 +30,25 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
...
-       * disabled: {
-       *   value: { base: '{colors.legacy.mauve.30.light}', _dark: '{colors.legacy.mauve.30.dark}' },
-       *   type: 'color',
-       * },
-       * accent: {
-       *   value: { base: '{colors.legacy.blue.20.light}', _dark: '{colors.legacy.blue.20.dark}' },
-       *   type: 'color',
-       * },
-       */
+      // not actually legacy, but makes the merging of the two easier for now.
```
antonjoel82, line 33: These are from the new design system, but since our feedback colors are kind of all over the place and used much more situationally, I decided exposing these here and now was a bit easier.

LetItRock (approved):

PR #5338, opened 2024-03-25
feat: Add type guard to improve error handling

antonjoel82:

- Add type guard to improve error handling

- Clean-up error handling (especially in front-end code)

Notes:

- Parameters for `catch` blocks should be typed as unknown ([ref](https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables))

Usage example:

```
catch (err: unknown) {
  if (!checkIsResponseError(err) {
     // ... handle unexpected error type
     return;
  }
  // type-checker knows that err is IResponseError, so can access the fields as normal
  showNotification('red', err.message);
}
```


libs/shared/src/utils/checkIsResponseError.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { checkIsResponseError } from './checkIsResponseError';
+import { IResponseError } from '../types';
+
+describe('checkIsResponseError', () => {
```
scopsy, line 4: **praise:** Love it!‏

scopsy (approved):

PR #5346, opened 2024-04-01
Small dx touch leveraging env var in node sdk

SokratisVidros:

This PR introduces a slight DX improvement to use `NOVU_API_KEY` env variable if available in `@novu/node` SDK. It's also a test PR to play with how libraries and packages interact in the monorepo with conditional exports and modern Node resolution.

It's just a tiny, quality of life DX improvement.


libs/shared/src/index.ts
```diff
@@ -1,26 +1,27 @@
-export * from './entities/user';
```
SokratisVidros, line 1: Ignore these changes. I just ordered them alphabetically.

libs/shared/src/utils/env.ts
```diff
@@ -0,0 +1,51 @@
+type CloudflareEnv = { env: Record<string, string> };
```
SokratisVidros, line 1: This is more of a preview for now in terms of how things should be approached in the new JS isomorphic runtime world. The package works in Node for now so process.env will work.

packages/node/src/lib/novu.ts
```diff
@@ -40,8 +41,25 @@ export class Novu extends EventEmitter {
   readonly organizations: Organizations;
   readonly workflowOverrides: WorkflowOverrides;
 
-  constructor(apiKey: string, config?: INovuConfiguration) {
+  constructor(config?: INovuConfiguration);
+  constructor(apiKey: string, config?: INovuConfiguration);
```
SokratisVidros, line 45: Overloading ensures backwards compatibility.

.eslintrc.js
```diff
@@ -66,6 +66,7 @@ module.exports = {
       {
         patterns: [
           '@novu/shared/*',
+          '!@novu/shared/utils',
```
scopsy, line 69: **question:** this excludes the utils path from bad imports rule right? ‏

libs/shared/package.json
```diff
@@ -25,6 +25,18 @@
   "files": [
     "dist/"
   ],
+  "exports": {
+    ".": {
+      "require": "./dist/cjs/index.js",
+      "import": "./dist/esm/index.js",
+      "types": "./dist/esm/index.d.js"
+    },
+    "./utils": {
```
scopsy, line 34: **question:** What is the benefit for introducing another barrel entry point for shared vs just importing from root?‏

libs/shared/src/utils/env.ts
```diff
@@ -0,0 +1,51 @@
...
+};
+
+/**
+ *
+ * Utility function to get env variables across Node and Edge runtimes.
+ *
+ * @param name Pass the name of the environment variable. The param is case-sensitive.
+ * @returns string Returns the value of the environment variable if exists.
+ */
+export const getEnvVariable = (name: string, context?: any): string => {
```
scopsy, line 19: **praise:** Really nice addition!‏

packages/node/src/lib/novu.ts
```diff
@@ -40,8 +41,25 @@ export class Novu extends EventEmitter {
...
+    if (arguments.length === 2) {
+      apiKey = args[0];
+      config = args[1];
+    } else if (arguments.length === 1) {
+      const { apiKey: key, ...rest } = args[0];
+      apiKey = key;
+      config = rest;
+    } else {
+      apiKey = getEnvVariable('NOVU_API_KEY');
+    }
```
scopsy, line 61: **note (non-blocking):** Should we add a unit test for this behaviour? Just so that we are confident in this tricky part here‏

packages/node/src/lib/novu.ts
```diff
@@ -40,8 +41,25 @@ export class Novu extends EventEmitter {
...
-  constructor(apiKey: string, config?: INovuConfiguration) {
+  constructor(config?: INovuConfiguration);
+  constructor(apiKey: string, config?: INovuConfiguration);
+  constructor(...args: any) {
     super();
+
+    let apiKey: string | undefined;
+    let config: INovuConfiguration | undefined;
+
+    if (arguments.length === 2) {
```
scopsy, line 52: **nitpick (non-blocking):** 
‏
```suggestion
    const haveApiKeyParam = arguments.length === 2;
    if (haveApiKeyParam) {
```

Just for easier readability.

packages/node/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "strictNullChecks": true,
     "allowSyntheticDefaultImports": true,
     "outDir": "build/main",
-    "module": "commonjs",
+    "module": "CommonJS",
+    "moduleResolution": "NodeNext",
```
scopsy, line 8: **question:** What does moduleResolution add?‏

scopsy (approved):

.eslintrc.js
SokratisVidros: Yes.

libs/shared/package.json
SokratisVidros: It's a new pattern to avoid one level of barrel files since node now support subpath imports. When used extensively, Barrel files affect [performance](https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/).

Of course, just one change won't have any impact, I am mostly testing whether the pattern can be applied moving forward.

packages/node/tsconfig.json
SokratisVidros: `nodenext` modes integrate with Node’s [native ECMAScript Module support](https://nodejs.org/api/esm.html) and if not mistaken we also get subpath exports.

packages/node/src/lib/novu.ts
SokratisVidros: Done.

.cspell.json
```diff
@@ -702,6 +703,7 @@
     "angular.json",
     "ng-package.json",
     "libs/shared/src/types/timezones/timezones.types.ts",
-    "*.riv"
+    "*.riv",
+    "cafebabe"
```
scopsy, line 707: @SokratisVidros i think it should be added at the top section of words, vs the file paths key here

SokratisVidros: Fixed.

PR #5350, opened 2024-04-01
[Information Architecture] - API Changes for Reset Password

antonjoel82:

- Implement a URL Search Param while resetting password
- Implement function to get the Email verification redirect link based on the search param

- This will enable us to reuse logic across password flows and is the basis for front-end changes in #5325


BiswaViraj (approved):
🚀

SokratisVidros (approved):

PR #5352, opened 2024-04-02
fix: cache validation during notification template promotion

BiswaViraj:

- Fixed the cache invalidation to use correct `environmentId` and `notificationTemplateId` during promotion of notification templates.


apps/api/src/app/change/usecases/promote-notification-template-change/promote-notification-template-change.usecase.ts
```diff
@@ -181,19 +181,7 @@ export class PromoteNotificationTemplateChange {
...
-        _environmentId: command.environmentId,
-      }),
-    });
-
-    await this.invalidateCache.invalidateByKey({
-      key: buildNotificationTemplateIdentifierKey({
-        templateIdentifier: newItem.triggers[0].identifier,
-        _environmentId: command.environmentId,
-      }),
-    });
```
BiswaViraj, line 196: Previously, It was using the development environment Id and development environment's notification templates' id.
and was not invalidating for the Prod env

djabarovgeorge (approved):
LGTM :) 
have you had the chance to test it as well?

BiswaViraj:
Yes, I was able to reproduce the issue locally and tested the fix. It works
now

PR #5362, opened 2024-04-04
[Information Architecture] User Profile | Update Password Flow

antonjoel82:

- Allow user to update their password when they already have one set

- Closes https://linear.app/novu/issue/NV-3446/🏛️-settings-user-profile-update-a-password

_Note_: outlines are showing re-renders but will not be included in actual code

https://www.loom.com/share/7e2a8be4a9824789a6ae49b588d902fd?sid=cf7d340c-8e56-4faf-9192-3cb895591b31

### DIFF
--- a/apps/web/src/pages/settings/user-profile-page/UserProfileUpdatePasswordForm.tsx
+++ b/apps/web/src/pages/settings/user-profile-page/UserProfileUpdatePasswordForm.tsx
@@ -1,6 +1,6 @@
 import { Button, errorMessage, IconOutlineLockPerson, PasswordInput, successMessage } from '@novu/design-system';
 import { checkIsResponseError, IResponseError } from '@novu/shared';
-import { api, useAuthContext } from '@novu/shared-web';
+import { api } from '@novu/shared-web';
 import * as Sentry from '@sentry/react';
 import { useMutation } from '@tanstack/react-query';
 import { useForm } from 'react-hook-form';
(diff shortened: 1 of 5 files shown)


apps/web/src/pages/settings/user-profile-page/UserProfileUpdatePasswordForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+  onSuccess?: () => void;
+};
+
+interface IPasswordUpdateData {
+  currentPassword: string;
+  newPassword: string;
+  confirmPassword: string;
+}
+
+export const UserProfileUpdatePasswordForm: React.FC<UserProfileUpdatePasswordFormProps> = ({ onSuccess }) => {
```
antonjoel82, line 23: 🗒 **note (non-blocking):** Given that this form and the other password form have different field names, different numbers of inputs, and hit different APIs, it didn't seem worthwhile to try to make an abstraction to reuse code -- especially since they might diverge more in the future.‏

SokratisVidros: Since we are using TypeScript we can start replacing React.FC with typed functions ([more info](https://github.com/typescript-cheatsheets/react#reacttypescript-cheatsheets:~:text=Why%20is%20React.FC%20not%20needed%3F))

SokratisVidros: As you said with TS 5.1 and React 18 it's mostly a matter of preference. My argument aligns with [this post](https://www.totaltypescript.com/you-can-stop-hating-react-fc#should-you-use-it) and is basically; it just reads simpler to do:

`export const UserProfileUpdatePasswordForm = ({ onSuccess }: UserProfileUpdatePasswordFormProps) => { ... }`

antonjoel82: I generally agree with Matt Pocock and enjoy his content in general, but in this case I feel that the typing is less clear and resembles a normal function (which is not necessarily what I want since I think Components should be considered "first-class" constructs in React).

Either way, we do not have a code styling guide yet, and this is something that we can decide on as a group and put into the guide & accompanying linting

apps/web/src/pages/settings/user-profile-page/UserProfileUpdatePasswordForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+      return;
+    }
+
+    try {
+      await mutateAsync(data);
+
+      showNotification({
+        message: 'Password was set successfully',
+        color: 'green',
+      });
```
BiswaViraj, line 58: **todo:** Please use the `errorMessage` and `successMessage` notification helpers from our design-system‏

BiswaViraj (changes requested):

BiswaViraj (approved):

PR #5364, opened 2024-04-05
fix(worker): Skip template rendering for Echo Workflow steps

rifont:
* Skip template rendering for Push, SMS, In-App, and Chat channels when chimera content is present

Push, SMS, and Chat channels re-map some payload attributes (e.g. for SMS, `content` becomes `body`). This causes the template rendering to be passed a falsy object, and subsequently, template rendering fails with the following error seen in the Activity Log (see other info).

The Email channel already has this conditional skip (https://github.com/novuhq/novu/blob/next/apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts#L205) and therefore is working correctly. The remaining channels are updated to skip rendering.

Due to the In-App channel not having any payload remapping, this channel was not impacted, but has been updated also for consistency.

_Push_
image

_SMS_
image

_Chat_
image
N/A

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
@@ -89,12 +89,14 @@ export class SendMessageChat extends SendMessageBase {
     let content = '';
 
     try {
-      content = await this.compileTemplate.execute(
-        CompileTemplateCommand.create({
-          template: step.template.content as string,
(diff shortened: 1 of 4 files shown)


scopsy (approved):
Can we cherry pick to deploy?

rifont:
Yes we should cherry-pick. Let's verify the changes will succeed in CI testing, then we can merge to `next` and cherry pick into [0.25.x](https://github.com/novuhq/novu/tree/v0.25.x)

PR #5019, opened 2023-12-21
feat(notification-center): improve notification list styling

katyaterletskaya:
### What change does this PR introduce?

Add the ability to override the height value in styles by using [scrollable target approach](https://github.com/ankeetmaini/react-infinite-scroll-component/blob/master/src/stories/ScrollableTargetInfScroll.tsx)

### Why was this change needed?

Closes #5013

PR #5371, opened 2024-04-05
fix(notification-center,widget): infinite scroll issue

LetItRock:

Fixes the issue introduced in: #5019, with not scrollable parent component.
Fixes the Widget test for the infinite scroll which was always passing, and makes sure that the scroll event is triggered on the scrollable element.

https://github.com/novuhq/novu/assets/2607232/9bd2fd91-e95a-4a8b-8da0-6e7fac39462d

https://github.com/novuhq/novu/assets/2607232/be3ec601-f2d7-4baf-b900-ec4e074e19af

### DIFF
--- a/apps/widget/cypress/e2e/notifications-list.spec.ts
+++ b/apps/widget/cypress/e2e/notifications-list.spec.ts
@@ -171,11 +171,12 @@ describe('Notifications List', function () {
     cy.getByTestId('unseen-count-label').should('contain', '99+');
   });
 
-  it('pagination check', async function () {
+  it('pagination check', function () {
     cy.wait('@getNotificationsFirstPage');
     cy.task('createNotifications', {
       organizationId: this.session.organization._id,
       enumerate: true,
+      ordered: true,
       identifier: this.session.templates[0].triggers[0].identifier,
       token: this.session.token,
       subscriberId: this.session.subscriber.subscriberId,
@@ -206,5 +207,5 @@ describe('Notifications List', function () {
 });
 
 function scrollToBottom() {
-  cy.getByTestId('notifications-scroll-area').get('.infinite-scroll-component').scrollTo('bottom');
+  cy.getByTestId('notifications-scroll-area').scrollTo('bottom', { ensureScrollable: true });
 }
--- a/packages/notification-center/src/components/notification-center/components/NotificationsList.tsx
+++ b/packages/notification-center/src/components/notification-center/components/NotificationsList.tsx
@@ -47,4 +47,5 @@ export function NotificationsList({
 
 const notificationsListCss = css`
   height: 400px;
+  overflow-y: auto;
 `;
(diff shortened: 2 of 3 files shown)


apps/widget/cypress/e2e/notifications-list.spec.ts
```diff
@@ -171,11 +171,12 @@ describe('Notifications List', function () {
     cy.getByTestId('unseen-count-label').should('contain', '99+');
   });
 
-  it('pagination check', async function () {
```
LetItRock, line 174: because of `async` this test was always passing, no matter of the body results

apps/widget/cypress/e2e/notifications-list.spec.ts
```diff
@@ -171,11 +171,12 @@ describe('Notifications List', function () {
...
+  it('pagination check', function () {
     cy.wait('@getNotificationsFirstPage');
     cy.task('createNotifications', {
       organizationId: this.session.organization._id,
       enumerate: true,
+      ordered: true,
```
LetItRock, line 179: this test verifies notifications in the order, so I've added the option to create notifications ordered (trigger with a small delay after)

apps/widget/cypress/e2e/notifications-list.spec.ts
```diff
@@ -206,5 +207,5 @@ describe('Notifications List', function () {
 });
 
 function scrollToBottom() {
-  cy.getByTestId('notifications-scroll-area').get('.infinite-scroll-component').scrollTo('bottom');
+  cy.getByTestId('notifications-scroll-area').scrollTo('bottom', { ensureScrollable: true });
```
LetItRock, line 210: `ensureScrollable` ensures that the target element is scrollable element, otherwise the test will fail

packages/notification-center/src/components/notification-center/components/NotificationsList.tsx
```diff
@@ -47,4 +47,5 @@ export function NotificationsList({
 
 const notificationsListCss = css`
   height: 400px;
+  overflow-y: auto;
```
LetItRock, line 50: the fix... the `InfiniteScroll` component has a prop `scrollableTarget="notifications-list"` which points to the parent element and it should be a scrollable element otherwise the infinite scroll won't work

apps/widget/cypress/e2e/notifications-list.spec.ts
antonjoel82: ✏ **suggestion:** This is helpful context in the code‏

davidsoderberg (approved):

antonjoel82 (approved):
This looks good to me!

SokratisVidros:
Nicely done @LetItRock for the quick fix. We can take the fix a step further and hide the scrollbar all together while scrolling giving a more polished feeling to the UI especially in non OS devices

```
/* Hide scrollbar for Chrome, Safari and Opera */
.example::-webkit-scrollbar {
  display: none;
}

/* Hide scrollbar for IE, Edge and Firefox */
.example {
  -ms-overflow-style: none;  /* IE and Edge */
  scrollbar-width: none;  /* Firefox */
}```

Issue #5349, opened 2024-04-01
🐛 Bug Report: in-app editor content writing space is less than its height


jainpawan21:
### 📜 Description

in-app editor content writing space is less than its height

### 👟 Reproduction steps

1. create a workflow 
2. add an in-app step
3. click on the step and write some content in in-app editor
4. You will notice that actual space to edit is less than editor height

### 👍 Expected behavior

Height should be same as editor space

### 👎 Actual Behavior with Screenshots

<img width="1609" alt="in-app-test-notification | Novu Manage Platform 2024-04-02 00-06-18" src="https://github.com/novuhq/novu/assets/39362422/e0a13541-f622-4711-9269-b801b711dcd4">

### Novu version

Novu SaaS

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

PR #5372, opened 2024-04-06
fix(web): make InApp code editor fit the height of its parent container

FlacorLopes:

The code editor was using a fixed height of 100px. Making it scroll even when there is available space.
This changes makes it take full height of its parent

Before:
Screenshot from 2024-04-06 02-58-01

After:
Screenshot from 2024-04-06 02-59-13


FlacorLopes:
#5349

scopsy (approved):

scopsy:
Thank you @FlacorLopes 🙏

PR #5375, opened 2024-04-07
feat(worker): add swc compiler for the worker service

abrl91:
Adding [swc](https://docs.nestjs.com/recipes/swc) to the `Worker`

Integrating SWC with the Nest CLI has greatly enhanced the efficiency of the development workflow.

#### Before:
- on start
image
- on change
image

#### After:
- on start
image
- on change
image


apps/worker/src/generate-metadata.ts
```diff
@@ -0,0 +1,9 @@
+import { PluginMetadataGenerator } from '@nestjs/cli/lib/compiler/plugins';
+
+const generator = new PluginMetadataGenerator();
+generator.generate({
```
scopsy, line 4: I think that since we don't need to have swagger here, is this file and the metadata file can be removed? Or it's still required?

scopsy:
An interesting find I've noticed, when removing `webpack: true` from the nest-cli config, and relying only on SWC to compile, the compilation takes `70ms`, with the current setup in this branch it takes for me `3000ms` which is exactly similar to swc turned off. Which makes me think it's not really taking effect here.

Should we consider just go with plain SWC without webpack? (Some path adjustments need to be done when importing the .env files for it to work, But should be manageable)

CleanShot 2024-04-08 at 11 20 37@2x

apps/worker/src/generate-metadata.ts
abrl91: Yeah...Without adding any visitors, we won't see a metadata.ts file pop up (I found that out when Swagger was in the mix). Also, since the 'Worker' isn't using any CLI plugins, we can totally ditch that file.

scopsy:
This is slightly confusing, although we are a monorepo, we don't really have nest setup in a monorepo mode. Each app is treated independently without a centralized nest cli file.

I think worth giving it a go without webpack. With webpack it seems doesn't really affect build time before and after the change. Or maybe i'm missing something. Have you had a chance to compare reaults of 'start:dev' compared to the next branch?

abrl91:
> This is slightly confusing, although we are a monorepo, we don't really have nest setup in a monorepo mode. Each app is treated independently without a centralized nest cli file.
> 
> I think worth giving it a go without webpack. With webpack it seems doesn't really affect build time before and after the change. Or maybe i'm missing something. Have you had a chance to compare reaults of 'start:dev' compared to the next branch?

You mean `start:dev` from the `worker` dir, right? I'll check. 
Just yesterday I realized that I can run the `worker` from its path :) 
I noticed that after removing the changes I have made to the webpack file, and changed the nest-cli configuration 
image

it was faster, but I got the error about the `.env` path.

scopsy:
@abrl91 yep! I think that's the way. Fixing the path errors should be relatively simple. It's usually just adding or removing the "../../" in the imports of those files to be relative to their new location in the `dist` folder.

First one is from the `config/index.ts` folder I believe

abrl91:
@scopsy I think you can re-review it :)

scopsy:
@coderabbitai review

Cliftonz:
@abrl91  Thank you for submitting this, unfortunately this does not cover internal changes we have recently made nor does if handle all of the services.

We appreciate this PR however we are going to handle this upgrade internally.

Thanks you!

PR #5378, opened 2024-04-08
fix: free trial banner

davidsoderberg:


rifont (approved):

libs/shared-web/src/config.ts
```diff
@@ -60,7 +60,7 @@ export const LAUNCH_DARKLY_CLIENT_SIDE_ID =
   window._env_.REACT_APP_LAUNCH_DARKLY_CLIENT_SIDE_ID || process.env.REACT_APP_LAUNCH_DARKLY_CLIENT_SIDE_ID;
 
 export const FEATURE_FLAGS = Object.values(FeatureFlagsKeysEnum).reduce((acc, key) => {
-  const defaultValue = isCypress ? true : false;
+  const defaultValue = isCypress ? 'true' : false;
```
rifont, line 63: @davidsoderberg this is quite uncomfortable, we now have both string boolean and primitive boolean. Why is this necessary?

davidsoderberg: because the converter for the strings in the useFeatureFlags expects a string. 
Used here:
https://github.com/novuhq/novu/blob/e259c7f5d206f5a8288c54e6fa4d1aadfe5d7028/libs/shared-web/src/hooks/useFeatureFlags.ts#L30

Defined here:
https://github.com/novuhq/novu/blob/e259c7f5d206f5a8288c54e6fa4d1aadfe5d7028/libs/shared/src/services/feature-flags/feature-flags.util.ts#L4

rifont: That makes sense. Should we also update the 3rd party of the ternary expression to be `'false'` in that case?

rifont (approved):
Awesome!

PR #5379, opened 2024-04-08
refactor(web): import echo from echo ee echo web

djabarovgeorge:

move the echo from billing to its own package.

it is not connected to the billing.

[EE REF](https://github.com/novuhq/packages-enterprise/pull/116)


apps/web/src/pages/templates/components/InputVariables.tsx
```diff
@@ -18,11 +18,11 @@ export const InputVariables = ({
...
+    const module = require('@novu/ee-echo-web');
     const InputVariablesComponent = module.InputVariables;
 
     return <InputVariablesComponent schema={inputs} onSubmit={onSubmit} onChange={onChange} />;
-  } catch (e) {}
-
-  return null;
+  } catch (e) {
+    throw e;
+  }
```
djabarovgeorge, line 27: I'm curious if there's a rationale behind returning null when it's not self-hosted. In my opinion, we should consider the web failed if the import was intended but unsuccessful.

pnpm-workspace.yaml
```diff
@@ -11,3 +11,4 @@ packages:
   # all packages in enterprise modules
   - 'enterprise/packages/*'
   - 'enterprise/packages/libs/*'
+  - 'enterprise/packages/web/*'
```
djabarovgeorge, line 14: created so we could organize a bit the packages by folders

LetItRock:
@coderabbitai review

apps/web/src/pages/templates/components/InputVariables.tsx
BiswaViraj: Valid point. I also tend to agree with this

BiswaViraj (approved):

apps/web/src/pages/templates/components/InputVariables.tsx
djabarovgeorge: Thanks for the feedback, awesome so i will keep the throw here

enterprise/packages/web/echo/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "@mantine/core": "^5.7.1",
+    "@mantine/hooks": "^5.7.1",
+    "@novu/design-system": "^0.24.1",
+    "@novu/shared-web": "^0.24.1",
+    "@rjsf/core": "^5.17.1",
+    "@rjsf/validator-ajv8": "^5.17.1",
+    "@tanstack/react-query": "^4.20.4",
+    "eslint": "^8.33.0",
+    "eslint-plugin-react-hooks": "^4.4.0",
+    "react-router-dom": "6.2.2",
```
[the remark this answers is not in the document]
djabarovgeorge: Done :)

PR #5381, opened 2024-04-08
feat(web,api): user profile page form

LetItRock:

Information Architecture:
- User Profile Page form
- Changed the API endpoint to update user profile data `PUT /users/profile`
- The image uploading will be implemented in separate PR

https://github.com/novuhq/novu/assets/2607232/754fc609-8a02-4458-b197-8b6689bddfd0

https://github.com/novuhq/novu/assets/2607232/dc5c4918-64b5-4a82-a2dc-9c4fefb1b47b

### DIFF
--- a/apps/api/src/app/user/dtos/update-profile-request.dto.ts
+++ b/apps/api/src/app/user/dtos/update-profile-request.dto.ts
@@ -8,5 +8,5 @@ export class UpdateProfileRequestDto {
   lastName: string;
 
   @ApiProperty()
-  imageUrl: string;
+  profilePicture?: string;
 }
--- a/apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.command.ts
+++ b/apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.command.ts
@@ -1,10 +1,17 @@
 import { IsDefined, IsOptional, IsString, IsUrl } from 'class-validator';
 import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
+import { IsImageUrl } from '../../../shared/validators/image.validator';
 
 export class UpdateNameAndProfilePictureCommand extends EnvironmentWithUserCommand {
-  @IsUrl({ require_tld: false })
+  @IsUrl({
+    require_protocol: true,
+    protocols: ['https'],
+  })
+  @IsImageUrl({
+    message: 'Profile picture must be a valid image URL with one of the following extensions: jpg, jpeg, png, gif, svg',
+  })
   @IsOptional()
-  imageUrl: string;
+  profilePicture?: string;
 
   @IsDefined()
   @IsString()
--- a/apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.usecase.ts
+++ b/apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.usecase.ts
@@ -1,16 +1,20 @@
 import { Injectable, NotFoundException } from '@nestjs/common';
 import { ApiException, buildUserKey, InvalidateCacheService } from '@novu/application-generic';
 import { UserEntity, UserRepository } from '@novu/dal';
+
+import { BaseUserProfileUsecase } from '../base-user-profile.usecase';
 import { UpdateNameAndProfilePictureCommand } from './update-name-and-profile-picture.command';
 
 @Injectable()
-export class UpdateNameAndProfilePicture {
-  constructor(private invalidateCache: InvalidateCacheService, private readonly userRepository: UserRepository) {}
+export class UpdateNameAndProfilePicture extends BaseUserProfileUsecase {
+  constructor(private invalidateCache: InvalidateCacheService, private readonly userRepository: UserRepository) {
+    super();
+  }
 
   async execute(command: UpdateNameAndProfilePictureCommand) {
     if (!command.firstName || !command.lastName) throw new ApiException('First name and last name are required');
 
-    const user = await this.userRepository.findById(command.userId);
+    let user = await this.userRepository.findById(command.userId);
     if (!user) throw new NotFoundException('User not found');
 
     const updatePayload: Partial<UserEntity> = {
@@ -20,11 +24,8 @@ export class UpdateNameAndProfilePicture {
 
     const unsetPayload: Partial<Record<keyof UserEntity, string>> = {};
 
-    // if imageUrl exists, update the profile picture to the imageUrl, otherwise, unset/remove it
-    if (command.imageUrl) {
-      updatePayload.profilePicture = command.imageUrl;
-    } else {
-      unsetPayload.profilePicture = '';
+    if (command.profilePicture) {
+      updatePayload.profilePicture = command.profilePicture;
     }
 
     await this.userRepository.update(
@@ -42,5 +43,10 @@ export class UpdateNameAndProfilePicture {
         _id: command.userId,
       }),
     });
+
+    user = await this.userRepository.findById(command.userId);
+    if (!user) throw new NotFoundException('User not found');
+
+    return this.mapToDto(user);
   }
 }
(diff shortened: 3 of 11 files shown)


apps/api/src/app/user/dtos/update-profile-request.dto.ts
```diff
@@ -8,5 +8,5 @@ export class UpdateProfileRequestDto {
   lastName: string;
 
   @ApiProperty()
-  imageUrl: string;
+  profilePicture?: string;
```
LetItRock, line 11: changed this to be consistent with the entity name

apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.usecase.ts
```diff
@@ -20,11 +24,8 @@ export class UpdateNameAndProfilePicture {
 
     const unsetPayload: Partial<Record<keyof UserEntity, string>> = {};
 
-    // if imageUrl exists, update the profile picture to the imageUrl, otherwise, unset/remove it
-    if (command.imageUrl) {
-      updatePayload.profilePicture = command.imageUrl;
-    } else {
-      unsetPayload.profilePicture = '';
```
LetItRock, line 27: no need to remove the profile picture if not present in the dto

apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.usecase.ts
```diff
@@ -42,5 +43,10 @@ export class UpdateNameAndProfilePicture {
         _id: command.userId,
       }),
     });
+
+    user = await this.userRepository.findById(command.userId);
+    if (!user) throw new NotFoundException('User not found');
+
+    return this.mapToDto(user);
```
LetItRock, line 50: return the user data, so on the FE we don't need to refetch it

apps/web/src/api/hooks/useUpdateUserProfile.ts
```diff
@@ -0,0 +1,35 @@
...
+    IUserEntity,
+    IResponseError,
+    IUpdateUserProfileVariables
+  >(updateUserProfile, {
+    ...options,
+    onSuccess: async (data, variables, context) => {
+      options.onSuccess?.(data, variables, context);
+      if (data) {
+        queryClient.setQueryData(['/v1/users/me'], data);
+      }
```
LetItRock, line 27: update user data in query cache on success

apps/web/src/pages/settings/user-profile-page/UserProfileForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+interface IUserProfileFormProps {
+  currentUser?: IUserEntity | null;
+}
+
+const formInputStyles = css({
+  minWidth: '18.75rem',
+  '& label': { marginTop: '0 !important' },
+  '& .mantine-TextInput-error': { position: 'absolute', marginTop: '0.5rem' },
```
LetItRock, line 23: render errors on position absolute, otherwise, the form is jumping like:

Screenshot 2024-04-08 at 18 39 20

apps/web/src/pages/settings/user-profile-page/UserProfileForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+});
+
+const FIRST_NAME_LABEL = 'First name';
+const LAST_NAME_LABEL = 'Last name';
+const MIN_LENGTH_RULE = { value: 2, message: 'Should be longer than 2 characters' };
+
+const makeFormData = (data?: IUserEntity | null): IUserProfileForm => ({
+  firstName: capitalize(data?.firstName ?? ''),
+  lastName: capitalize(data?.lastName ?? ''),
+});
```
LetItRock, line 34: the first and last names are lowercased in db, wow :)

apps/web/src/pages/settings/user-profile-page/UserProfileForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+    formState: { isDirty, errors },
+  } = useForm<IUserProfileForm>({
+    mode: 'onSubmit',
+    defaultValues: makeFormData(currentUser),
+  });
+
+  const { updateUserProfile, isLoading: isUpdatingUserProfile } = useUpdateUserProfile({
+    onSuccess: (newUserData) => {
+      successMessage('Profile updated successfully');
+      reset(makeFormData(newUserData));
```
LetItRock, line 50: reset the form on success

apps/web/src/pages/settings/user-profile-page/UserProfileForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+    },
+  });
+
+  const onUpdateUserProfileHandler: SubmitHandler<IUserProfileForm> = ({ firstName, lastName }) => {
+    updateUserProfile({ firstName, lastName });
+  };
+
+  const onUpdateUserProfileImageHandler = async (_: File[]) => {
+    // TODO: Implement image upload
+  };
```
LetItRock, line 63: will be implemented in the next PR

LetItRock:
> Looks good overall, thanks!
> 
> Couple small UX things with smaller screens, but I believe these could be cleaned-up in a later iteration:
> 
> 1. Photo gets clipped on small screens. We should talk w/ Nik about how the form should look
> 2. When there's not enough room for the Loading icon in the button, it causes the whole form to shift

I think that for now, I'll give the min-width to the image and that will cause the horizontal scroll which is not ideal but at least would be accessible and will cover two of the cases. We can review that later.

apps/api/src/app/user/dtos/update-profile-request.dto.ts
LetItRock: @antonjoel82 yes, agree but on the other hand, it won't be aligned with the `UserResponseDto` which will return the `profilePicture`

apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.command.ts
```diff
@@ -1,10 +1,17 @@
...
 
 export class UpdateNameAndProfilePictureCommand extends EnvironmentWithUserCommand {
-  @IsUrl({ require_tld: false })
+  @IsUrl({
+    require_protocol: true,
+    protocols: ['https'],
+  })
+  @IsImageUrl({
+    message: 'Profile picture must be a valid image URL with one of the following extensions: jpg, jpeg, png, gif, svg',
+  })
```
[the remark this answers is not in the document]
LetItRock: This annotation verifies whether the image URL has a correct extension at the end of the string. The `/user` endpoints are internal APIs used by us and are not exposed, so it's just the information to us at the end ;)

apps/api/src/app/user/usecases/update-name-and-profile-picture/update-name-and-profile-picture.usecase.ts
LetItRock: This is a similar approach we use for the topics; but it's a general pattern when you need to transfer data between different layers of the application, in this case, the controller and use-case layers.

apps/web/src/components/shared/ProfileImage.tsx
```diff
@@ -9,7 +9,7 @@ interface ProfileImageProps {
   /**
    * **NOTE**: Value should be URL string
    */
-  value: string;
+  value?: string | null;
```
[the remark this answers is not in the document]
LetItRock: Extended it because we support in code both. But mainly needed because the `profilePicture` prop from the dto is defined like that. Otherwise, I would have to do `value={currentUser?.profilePicture ?? ''}` which is a hacky and also will fall under the `false` case of the `ProfileImage` ternary statement.

apps/web/src/pages/settings/user-profile-page/UserProfileForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <form
+      noValidate
+      onSubmit={handleSubmit(onUpdateUserProfileHandler)}
+      className={css({ display: 'flex', alignItems: 'flex-end', gap: '150', marginBottom: '2.5rem' })}
```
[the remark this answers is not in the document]
LetItRock: it's coming from the designs where we have a spacing between the profile form and the "Profile security" sections.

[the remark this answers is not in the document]
LetItRock: sure will do that 🙌

PR #5385, opened 2024-04-09
Change pull request template  

ainouzgali:


ainouzgali:
@coderabbitai review

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,12 +1,24 @@
-### What change does this PR introduce?
+### Summary
 
-<!-- Explain here the changes your PR introduces and text to help us understand the context of this change. -->
+<!-- Provide a concise summary "Why are the changes needed"? 
+Include any relevant links, such as Linear tickets, Slack discussions, 
+or design documents. -->
 
-### Why was this change needed?
```
SokratisVidros, line 5: The purpose of this section should be on the "Why". The changes made can be seen in the PR but why did we make them is very important context for the reviewer and for posterity.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,12 +1,24 @@
...
+### Changes Made
 
-<!-- If your PR fixes an open issue, use `Closes #999` to link your PR with the issue. #999 stands for the issue number you are fixing, Example: Closes #31 -->
+<!-- Provide details on the approach taken to address the problem 
+and any notable implementation details. -->
 
-### Other information (Screenshots)
+### Screenshots
+<!-- If the changes are visual, include screenshots or GIFs. -->
```
SokratisVidros, line 13:
```suggestion
<!-- If the changes are visual, include screenshots or screencasts. -->
```

.github/PULL_REQUEST_TEMPLATE.md
LetItRock: Most of our PRs don't have anything in this section, usually because the "why" is in the Linear ticket or GH issue.

.github/PULL_REQUEST_TEMPLATE.md
LetItRock: I would say that for the OS community it should be required to attach the screenshots. We've been so many times in the past when someone created the PR and said that the visual issue was fixed, but then after checking the PR locally, it wasn't.

.github/PULL_REQUEST_TEMPLATE.md
ainouzgali: Also, I felt people would write the why in the Summary part. Right now, the questions we have are usually ignored, the what vs why is not so clear and are usually combined into a few sentences into the "What change does this PR introduce?". 
So, my thought was that it would feel more natural to write why in the Summary (its also in the description of the question) because they see they have the next question that has the space to talk about what they actually did

.github/PULL_REQUEST_TEMPLATE.md
ainouzgali: But some PRs don't have any visual - like BE work. Thats why its phrased as optional. 
I suggest that when there are visual changes, but no screenshot is included, then we ask to please add screenshots before we can start reviewing their PR.
Also, now that there is a dedicated section for that, they will be more likely to include them IMO

.github/PULL_REQUEST_TEMPLATE.md
SokratisVidros: In this case, we can combine them: `What changed? Why did it change?`

ainouzgali: OK. But I removed the Changes Made part. I don't want it to go back to being confusing with 2 too similar questions. What do you think?

SokratisVidros: Looks good.

SokratisVidros: Can we also use the same in the ee repo?

ainouzgali: of course. There is already a PR for the ee, just waited until we agreed here to change it there. Thanks. Can you please approve so I can merge this one? :)

davidsoderberg (approved):

PR #5389, opened 2024-04-10
feat: add reply to field in resend

jainpawan21:
Added missing replyTo field in resend provider


LetItRock (approved):

pontusab:
awesome! thank you!

PR #5391, opened 2024-04-10
fix: Ensure health check return the correct status

SokratisVidros:

This PR throws the Terminus error when the indicator is unhealthy and applies minor code refactorings to most health indicators. Throwing the right error results in the correct status page result (see screenshots). 

Currently, the DAL and Cache health checks won't cause the overall health status of the app to return an error, as Terminus requires us to throw a HealthCheck error from the indicator when its status is unhealthy.

For more information, refer to https://docs.nestjs.com/recipes/terminus#custom-health-indicator

### Screenshots

Screenshot 2024-04-10 at 12 36 42
Screenshot 2024-04-10 at 12 01 56

### DIFF
--- a/apps/ws/src/socket/services/ws-server-health-indicator.service.ts
+++ b/apps/ws/src/socket/services/ws-server-health-indicator.service.ts
@@ -1,4 +1,4 @@
-import { HealthIndicator, HealthIndicatorResult } from '@nestjs/terminus';
+import { HealthCheckError, HealthIndicator, HealthIndicatorResult } from '@nestjs/terminus';
 import { Injectable } from '@nestjs/common';
 
 import { IHealthIndicator } from '@novu/application-generic';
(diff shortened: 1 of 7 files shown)


scopsy:
@coderabbitai review

scopsy (approved):

packages/application-generic/src/health/cache.health-indicator.ts
```diff
@@ -3,34 +3,29 @@ import {
...
+  private static KEY = 'cacheService';
 
   constructor(private cacheService: CacheService) {
     super();
   }
 
   async isHealthy(): Promise<HealthIndicatorResult> {
-    const isReady = this.cacheService.cacheEnabled();
+    const isHealthy = this.cacheService.cacheEnabled();
+    const result = this.getStatus(CacheServiceHealthIndicator.KEY, true);
```
djabarovgeorge, line 19: Should not we pass here isHealthy?

```suggestion
    const result = this.getStatus(CacheServiceHealthIndicator.KEY, isHealthy);
```

SokratisVidros: Great catch!

apps/ws/src/socket/services/ws-server-health-indicator.service.ts
```diff
@@ -7,16 +7,21 @@ import { WSGateway } from '../ws.gateway';
...
+  private static KEY = 'ws-server';
 
   constructor(private wsGateway: WSGateway) {
     super();
   }
 
   async isHealthy(): Promise<HealthIndicatorResult> {
-    const status = !!this.wsGateway.server;
+    const isHealthy = !!this.wsGateway.server;
+    const result = this.getStatus(WSServerHealthIndicator.KEY, isHealthy);
```
LetItRock, line 18: ✏ **suggestion:** this function seems to need to be called only when `isHealthy` is `true`, so maybe consider moving it under the if statement. Same in the other places.

SokratisVidros: The result is also used in the thrown error.

LetItRock: ahh sorry, didn't noticed that :)

LetItRock (approved):

PR #5392, opened 2024-04-10
chore(repo): Disable body-max-line-length

SokratisVidros:

Disable body-max-line-length commitlint rule. It was too strict!

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,10 +1,23 @@
 {
-  "extends": ["@commitlint/config-conventional"],
+  "extends": [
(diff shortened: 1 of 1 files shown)


rifont (approved):
The few formatting changes of the JSON object are unrelated to the PR intent, but aligned with other parts of the JSON. I'm happy for us to include these formatting changes for consistency. Long term however, we probably want to lint these types of changes to enforce consistency.

Let's ship it!

SokratisVidros:
You are right. VC Code plugins did the trick. We will address it in a separate PR.

PR #5395, opened 2024-04-10
[Information Architecture] Webhook Page

antonjoel82:

- Implement the Inbound Webhook page

- Closes [NV-3458](https://linear.app/novu/issue/NV-3458/%F0%9F%8F%9B%EF%B8%8F-settings-inbound-webhook-page)

### Questions for Reviewers

- [ ] Is there anything that needs to be handled specially for self-hosted?

### Notes

- [x] e2e tests in progress, but trying to get earlier code feedback to keep moving forward

### Other information

https://www.loom.com/share/3794aa9f67bd41c1ace6807d0d37560e?sid=a7723bf7-442a-4d44-af0b-51336ee652d3

### DIFF
--- a/apps/web/src/SettingsRoutes.tsx
+++ b/apps/web/src/SettingsRoutes.tsx
@@ -1,5 +1,5 @@
 import { Route, Navigate } from 'react-router-dom';
-import { Cloud, IconLockPerson, SSO, UserAccess } from '@novu/design-system';
+import { Cloud, SSO, UserAccess } from '@novu/design-system';
 import { FeatureFlagsKeysEnum } from '@novu/shared';
 
 import { ProductLead } from './components/utils/ProductLead';
@@ -14,6 +14,7 @@ import { SettingsPageNew as SettingsPage } from './pages/settings/SettingsPageNe
 import { ApiKeysCard } from './pages/settings/tabs';
 import { EmailSettings } from './pages/settings/tabs/EmailSettings';
 import { OrganizationPage } from './pages/settings/organization';
+import { WebhookPage } from './pages/settings/WebhookPage';
 
 /** Note: using a hook is the only way to separate routes */
 export const useSettingsRoutes = () => {
@@ -33,7 +34,7 @@ export const useSettingsRoutes = () => {
         <Route path={ROUTES.ORGANIZATION} element={<OrganizationPage />} />
         <Route path={ROUTES.TEAM_SETTINGS} element={<TeamPage />} />
         <Route path={ROUTES.BILLING} element={<BillingRoutes />} />
-        <Route path={ROUTES.WEBHOOK} element={<EmailSettings />} />
+        <Route path={ROUTES.WEBHOOK} element={<WebhookPage />} />
         <Route path={ROUTES.SECURITY} element={<AccessSecurityPage />} />
         {/* Default settings route in case we didn't match with the existing */}
         <Route path={`${ROUTES.SETTINGS}/*`} element={<Navigate to={ROUTES.PROFILE} replace />} />
--- a/apps/web/src/pages/settings/SettingsPageContainer.tsx
+++ b/apps/web/src/pages/settings/SettingsPageContainer.tsx
@@ -1,11 +1,12 @@
-import { PageContainer } from '@novu/design-system';
-import { FC, PropsWithChildren } from 'react';
+import { PageContainer, When } from '@novu/design-system';
+import { FC, PropsWithChildren, ReactNode } from 'react';
 import PageHeader from '../../components/layout/components/PageHeader';
 import { css } from '../../styled-system/css';
 
 export interface ISettingsPageContainerProps {
   // TODO: this should be LocalizedMessage, but PageContainer and PageHeader don't accept it
   title: string;
+  header?: ReactNode;
 }
 
 /**
@@ -24,10 +25,23 @@ const SettingsPageContentContainer: FC<PropsWithChildren<{}>> = ({ children }) =
   );
 };
 
-export const SettingsPageContainer: FC<PropsWithChildren<ISettingsPageContainerProps>> = ({ title, children }) => {
+export const SettingsPageContainer: FC<PropsWithChildren<ISettingsPageContainerProps>> = ({
+  title,
+  children,
+  header,
+}) => {
   return (
     <PageContainer title={title}>
       <PageHeader title={title} />
+      <When truthy={!!header}>
+        <section
+          className={css({
+            mx: '150',
+          })}
+        >
+          {header}
+        </section>
+      </When>
       <SettingsPageContentContainer>{children}</SettingsPageContentContainer>
     </PageContainer>
   );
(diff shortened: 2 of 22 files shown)


apps/web/src/pages/settings/SettingsPageContainer.tsx
```diff
@@ -24,10 +25,23 @@ const SettingsPageContentContainer: FC<PropsWithChildren<{}>> = ({ children }) =
...
-export const SettingsPageContainer: FC<PropsWithChildren<ISettingsPageContainerProps>> = ({ title, children }) => {
+export const SettingsPageContainer: FC<PropsWithChildren<ISettingsPageContainerProps>> = ({
+  title,
+  children,
+  header,
+}) => {
   return (
     <PageContainer title={title}>
       <PageHeader title={title} />
+      <When truthy={!!header}>
```
antonjoel82, line 36: 🗒 **note (non-blocking):** I think Nikolay may have established a new layout pattern on this page (and only this one), so this supports it without changing other settings pages‏

apps/web/src/pages/settings/WebhookPage/WebhookClaimStatusDisplay.tsx
```diff
@@ -0,0 +1,68 @@
...
+  claimed: <IconOutlineDomainVerification />,
+};
+
+interface IWebhookClaimStatusProps {
+  status: WebhookClaimStatus;
+  handleRefresh: () => Promise<any>;
+  isLoading: boolean;
+}
+
+const iconTextRecipe = cva({
```
antonjoel82, line 34: 🗒 **note (non-blocking):** Here is my first time using `cva` (an atomic recipe)! It is a great way to have "conditional" styles while still allowing Panda to statically analyze our code and generate the necessary styles.‏

apps/web/src/pages/settings/WebhookPage/WebhookPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+  '& input': {
+    border: 'none !important',
+    background: 'surface.popover !important',
+    color: 'typography.text.secondary !important',
+    fontFamily: 'mono !important',
+  },
+});
+
+// source: https://www.geeksforgeeks.org/how-to-validate-a-domain-name-using-regular-expression/
+const DOMAIN_REGEX = /^(?!-)[A-Za-z0-9-]+([\-\.]{1}[a-z0-9]+)*\.[A-Za-z]{2,6}$/;
```
antonjoel82, line 29: ❓ **question:** Should we have this restriction? Currently, we have 0 server-side restriction on domains 😱‏

apps/web/src/pages/settings/WebhookPage/WebhookPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+        </Stack>
+      }
+    >
+      <Stack gap="200" className={css({ maxW: '960px' })}>
+        <Timeline>
+          <MantineTimeline.Item
+            bullet={1}
+            lineVariant="dashed"
+            title={<label htmlFor="inbound-parse-domain-input">Specify a domain name for {envName} environment</label>}
+          >
```
antonjoel82, line 70: 🗒 **note (non-blocking):** As mentioned in the comment at the top, I had some issues with doing my own wrapper around the Mantine `Timeline.Item` -- no styles would apply the timeline any more. However, I didn't want to get stuck on this since we'll want to reconsider the Timeline more generically in the future when we extract it to the design system‏

apps/web/src/pages/settings/WebhookPage/useWebhookPage.ts
```diff
@@ -0,0 +1,100 @@
...
+import { useMutation } from '@tanstack/react-query';
+import { useState, useEffect } from 'react';
+import { useForm } from 'react-hook-form';
+import { validateMxRecord } from '../../../api/inbound-parse';
+import { useEffectOnce } from '../../../hooks';
+import { getWebhookClaimStatusFromEnvironment } from './getWebhookClaimStatusFromEnvironment';
+import { WebhookClaimStatus } from './WebhookPage.types';
+
+export const useWebhookPage = () => {
+  const mailServerDomain = `10 ${MAIL_SERVER_DOMAIN}`;
```
antonjoel82, line 14: ❓ **question:** ‏This was copied from the previous implementation... should we do anything better here?‏

apps/web/src/pages/settings/WebhookPage/useWebhookPage.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  async function handleUpdateDnsSettings({ inboundParseDomain }) {
+    const payload = {
+      payload: { inboundParseDomain },
+      environmentId: environment?._id ?? '',
+    };
+
+    await updateDnsSettingsMutation(payload);
+    successMessage('Domain info updated successfully');
+  }
```
antonjoel82, line 66: 🗒 **note (non-blocking):** Most code here was copied from the `EmailSettings` too‏

libs/design-system/src/panda/semanticSizes.tokens.ts
```diff
@@ -0,0 +1,18 @@
+import { defineSemanticTokens } from '@pandacss/dev';
+import { INPUT_HEIGHT_PX } from '../config/inputs.styles';
+
+/**
+ * Represents a dimension of an element.
+ *
+ * Used for properties like width and height.
+ */
+export const SEMANTIC_SIZES_TOKENS = defineSemanticTokens.sizes({
```
antonjoel82, line 9: 🗒 **note (non-blocking):** Excited for this! This made it easier to work with some hardcoded values that otherwise didn't work within Panda. We can change these as the design system matures, but I suggest we continue adding here as we encounter useful, hard-coded `px / rem` values‏

apps/web/src/pages/settings/WebhookPage/WebhookPage.tsx
BiswaViraj: I also noticed the regex doesn't support URL schemes (http, https etc)

apps/web/src/SettingsRoutes.tsx
```diff
@@ -32,7 +33,7 @@ export const useSettingsRoutes = () => {
         <Route path={ROUTES.ORGANIZATION} element={<ApiKeysCard />} />
         <Route path={ROUTES.TEAM_SETTINGS} element={<MembersInvitePageNew />} />
         <Route path={ROUTES.BILLING} element={<BillingRoutes />} />
-        <Route path={ROUTES.WEBHOOK} element={<EmailSettings />} />
+        <Route path={ROUTES.WEBHOOK} element={<WebhookPage />} />
```
BiswaViraj, line 36: It is using `/:env` in the URL to differentiate between envs.
But this param is not used in the hook to fetch the relevant details.
Also, currently it allows to visit any random URL param for example: `http://localhost:4200/settings/webhook/hello-joel` still opens the webhook page

BiswaViraj: There is no way to access the webhook page for the `PROD` env

apps/web/src/pages/settings/WebhookPage/WebhookPage.tsx
antonjoel82: Correct -- from my understanding (and how its labeled on the page), this should be a domain name, not an URL

apps/web/src/SettingsRoutes.tsx
antonjoel82: 👏 **praise:** Great catch -- thank you!

I updated it with a hook to force redirects (and applied it to the API Keys as well)‏

BiswaViraj (approved):

PR #5400, opened 2024-04-11
Remove redundant step for uploading a new translation file

ainouzgali:
A user needs to open a language page and click the upload button instead of directly upload a JSON file. Originally, we had that to allow either a direct file upload or an upload from the CLI but we are not planning to implement the cli option in the near future.

[The enterprise PR](https://github.com/novuhq/packages-enterprise/pull/120) removes the empty state screen with the upload button and directly opens the file manager. Also, removed other places that mentioned the CLI option.


BiswaViraj (approved):
Easiest PR to review 🙌🏼

Issue #5402, opened 2024-04-12
🚀 Feature: Support SendGrid ASM parameters


dmgarland:
### 🔖 Feature description

[ASM groups ](https://docs.sendgrid.com/ui/sending-email/unsubscribe-groups)are a SendGrid feature designed to allow e-mail suppressions short of stopping all mail. 

The [SendGrid API ](https://docs.sendgrid.com/api-reference/mail-send/mail-send) supports `asm` as an option, where the group id is matched against the recipient to see whether they should be sent mail or not.

Currently the Novu SendGrid provider doesn't support `asm` parameters passed via `customData`. This patches that to allow an asm group id to be passed directly through to SendGrid

### 🎤 Why is this feature needed ?

We have a use case where an ASM group is shared with other parts of our platform that require to send mail, so we need to suppress mail being sent to members of that group.

### ✌️ How do you aim to achieve this?

I will change the SendGrid provider to allow the passing of this option

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

Abobos:
Can I work on this issue?

PR #5403, opened 2024-04-12
feat: Pass customData asm parameters to SendGrid

dmgarland:

Looks for an optional `asm` parameter in `customData` that is passed through to the SendGrid API. 

Addresses #5402

### DIFF
--- a/providers/sendgrid/src/lib/sendgrid.provider.ts
+++ b/providers/sendgrid/src/lib/sendgrid.provider.ts
@@ -13,6 +13,10 @@ import {
 import { MailDataRequired, MailService } from '@sendgrid/mail';
 
 type AttachmentJSON = MailDataRequired['attachments'][0];
+type ASMConfig = {
+  groupId: number;
+  groupsToDisplay: number[];
+};
 
 export class SendgridEmailProvider implements IEmailProvider {
   id = 'sendgrid';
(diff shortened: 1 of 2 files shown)


providers/sendgrid/src/lib/sendgrid.provider.ts
```diff
@@ -13,6 +13,10 @@ import {
 import { MailDataRequired, MailService } from '@sendgrid/mail';
 
 type AttachmentJSON = MailDataRequired['attachments'][0];
+type ASMConfig = {
+  groupId: number;
+  groupsToDisplay: number[];
+};
```
LetItRock, line 19:
```suggestion
type ASMConfig = {
  groupId: number | string;
  groupsToDisplay: Array<number | string>;
};
```

providers/sendgrid/src/lib/sendgrid.provider.ts
```diff
@@ -70,12 +74,22 @@ export class SendgridEmailProvider implements IEmailProvider {
...
     const templateId = options.customData?.templateId as unknown as string;
+    const asm = options.customData?.asm as ASMConfig;
+    if (asm?.groupId) {
+      asm.groupId = parseInt(asm.groupId as any); // Sendgrid expects groupId to be a number
+    }
+    if (asm?.groupsToDisplay) {
+      asm.groupsToDisplay = asm.groupsToDisplay.map((group) =>
+        parseInt(group as any)
+      ); // Sendgrid expects groupsToDisplay to be an array of numbers
+    }
```
LetItRock, line 85: ✏ **suggestion:** if we would like to allow passing the string or number for both fields, please create a mapper function that will create a new object with appropriate value types like: 

```
  private mapAsmOptions(asm: ASMConfig): MailDataRequired['asm'] {
    return {
      groupId: parseInt(`${asm.groupId}`),
      groupsToDisplay: asm.groupsToDisplay.map((group) => parseInt(`${group}`)),
    };
  }
...

const asm = this.mapAsmOptions(options.customData?.asm);
```

PR #5404, opened 2024-04-12
chore(repo): Simplify health indicators interface

SokratisVidros:

Remove the `isActive` method from the `IHealthIndicator` interface, as it's currently not in use.

Just a clean up to simplify things.


djabarovgeorge:
@SokratisVidros 
I am afraid that we removed the isActive flag use during refactoring.  If we would revert this change, it would allow us to verify if a service is healthy by checking if it's not paused as well. 
Is this validation not needed anymore?

SokratisVidros:
@djabarovgeorge, I couldn't find any usages for the current status of the code. Would taking into account the paused state of each queue in the `isActive` method resolve the concerns?

SokratisVidros:
@djabarovgeorge health check run every [10 seconds](https://github.com/novuhq/cloud-infra/blob/main/terraform/novu/aws/service/templates/ecs/cb_app.json.tpl). Any other objections before merging this for now?

djabarovgeorge:
@SokratisVidros Let's merge it! :)
 
Sorry, I meant the worst-case scenario where the service is down, looking at it again looks like the health check could take up to 150 seconds, which seems more reasonable the 10 minutes.

PR #5405, opened 2024-04-12
chore(repo): Stream logs in local dev

SokratisVidros:

Teach lerna to stream logs when local development starts via `pnpm start:dev`.

This eases the local development lifecycle.


rifont (approved):
LGTM!

PR #5407, opened 2024-04-12
Nv 3455 settings branding page

BiswaViraj:
NOTE: I decided to re-brand the current branding form and leave out the preview section as it will be too much work for a page that we are going to deprecate soon as we discussed earlier.

- migrated the current branding form to use panda CSS
- added brand color input
- updated the design 

image


apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+                    data-test-id="font-color-picker"
+                    disallowInput={false}
+                    {...field}
+                  />
+                )}
+                control={control}
+                name="fontColorValue"
+              />
+            </div>
+          </Stack>
```
BiswaViraj, line 216: Interestingly, the font color value was send as color (brand color) in the previous iteration.

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
BiswaViraj, line 1: Re-used the existing branding form and updated the mantine imports to panda CSS.

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
antonjoel82: 😱

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+          <div
+            className={css({
+              textAlign: 'right',
+              marginTop: '60px',
```
antonjoel82, line 221: ✏ **suggestion:** ‏ We should reconsider this value and confirm with Nikolay‏

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+            })}
+          >
+            <Button submit loading={isLoading} data-test-id="submit-branding-settings" px="4rem">
```
antonjoel82, line 224:
```suggestion
            <Button submit loading={isLoading} data-test-id="submit-branding-settings" **px="4rem"**>
```
✏ suggestion: ‏ We should reconsider this value and confirm with Nikolay‏

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+                            className={css({
+                              minHeight: '6.25rem',
+                              minWidth: '6.25rem',
+                              objectFit: 'contain',
+                            })}
+                            alt="avatar"
+                          />
+                        )}
+                      </div>
+                    </Dropzone>
```
antonjoel82, line 158: ☑ **chore:** This would be a good boundary for a standalone component -- please separate out‏

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+                        </div>
+                      </div>
+                    )}
+                    <Dropzone
+                      styles={{
+                        root: {
+                          background: 'none',
+                          border: 'none',
+                        },
+                      }}
```
antonjoel82, line 126: 🤓 **nitpick (non-blocking):** Please avoid `styles` where possible -- instead, let's try to leverage Panda to override. However, if it is not straight-forward and ends up being time-consuming, it's okay to skip.‏

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+                      openRef={dropzoneRef}
+                      accept={Object.keys(MIME_TYPE_TO_FILE_EXTENSION)}
```
antonjoel82, line 128: ⚠ **issue:** While this works for now, if we add another file type to that look-up that isn't valid in this situation, this will be a problem. Let's safeguard ourselves by picking the specific subset we want to accept‏

antonjoel82 (changes requested):
Nice work on the Panda changes, and thank you for updating various aspects!

**suggestion**: we should ask Nikolay for a small re-design of the layout if we're going to remove the preview; the page feels imbalanced otherwise.

**issue**: When testing, the page / tab seems to free every time after I interact with the color picker inputs (I think primarily when clicking on the hex value text).

apps/web/src/SettingsRoutes.tsx
```diff
@@ -7,6 +7,7 @@ import { ROUTES } from './constants/routes.enum';
 import { useFeatureFlag } from './hooks';
 import { BillingRoutes } from './pages/BillingPages';
 import { BrandingForm } from './pages/brand/tabs';
+import { BrandingForm as BrandingFormNew } from './pages/brand/tabs/v2';
```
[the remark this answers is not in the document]
SokratisVidros: ☝️

apps/web/src/api/hooks/useUpdateOrganizationBranding.tsx
```diff
@@ -1,12 +1,14 @@
 import type { IResponseError } from '@novu/shared';
 import { useMutation, UseMutationOptions, useQueryClient } from '@tanstack/react-query';
-
 import { errorMessage } from '../../utils/notifications';
 import { updateBrandingSettings } from '../organization';
 
 type PayloadType = {
```
SokratisVidros, line 6: It's a good idea to make this look like CSSOM properties for maintainers and users.

```suggestion
type Branding = Partial<{
  logoUrl: string,
  backgroundColor,
  color,
  fontFamily
  ...
```

SokratisVidros: For example `color` instead of `fontColor` and `backgroundColor` instead of `color` unless `color` means something else for now.

apps/web/src/api/hooks/useUpdateOrganizationBranding.tsx
```diff
@@ -19,8 +21,11 @@ export const useUpdateOrganizationBranding = (
   const { mutateAsync: updateOrganizationBranding, isLoading } = useMutation<ResultType, IResponseError, PayloadType>(
     (payload) =>
       updateBrandingSettings({
-        logo: payload.logoUrl,
-        color: payload.colorValue,
+        logo: payload.brandLogoUrl || undefined,
```
SokratisVidros, line 24: **question:** Is it possible to align this object with the API payload to avoid this transformation? If we own both ends of the APIs and there are no backward compatibility concerns I suggest we do it. ‏

apps/web/src/api/hooks/useUploadToStorage.ts
```diff
@@ -3,13 +3,20 @@ import { useMutation, UseMutationOptions } from '@tanstack/react-query';
...
 
 export const useUploadToStorage = (
   options: UseMutationOptions<string, IResponseError, { file: File; type: UploadTypesEnum }> = {}
 ) => {
   const { getSignedUrl } = useGetSignedUrl();
   const { mutateAsync: uploadToStorage, ...mutationData } = useMutation(
     async ({ file, type }) => {
+      const isValidFileExtension = MIME_TYPE_TO_FILE_EXTENSION[file.type] !== undefined;
+
+      if (!isValidFileExtension) {
```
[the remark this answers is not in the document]
SokratisVidros: ☝️

SokratisVidros:
Based on the screenshot, can we split the page into two sections to add a bit more structure as follows:

# Branding
## Global settings
...

## In-App Notification Center
...

SokratisVidros (changes requested):

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+export function BrandingForm() {
+  const { currentOrganization: organization } = useAuthContext();
+  const dropzoneRef = useRef<() => void>(null);
+
+  const { uploadToStorage } = useUploadToStorage({
+    onSuccess: (path) => {
+      setValue('brandLogoUrl', path);
+    },
+    onError: (e) => {
```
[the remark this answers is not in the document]
BiswaViraj: It is inferred as `IResponseError`

apps/web/src/pages/brand/tabs/v2/BrandingForm.tsx
BiswaViraj: This is copied from the old file

apps/web/src/api/hooks/useUpdateOrganizationBranding.tsx
```diff
@@ -5,8 +5,11 @@ import { errorMessage } from '../../utils/notifications';
 import { updateBrandingSettings } from '../organization';
 
 type PayloadType = {
-  logoUrl?: string;
-  colorValue?: string;
+  brandLogoUrl?: string;
+  brandColorValue?: string;
+  fontColorValue?: string;
+  fontFamilyValue?: string;
```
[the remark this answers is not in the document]
BiswaViraj: The issue is the API supports `string` and it creates a mismatch between the types on the UI and API

antonjoel82 (approved):
Thanks for the updates! A couple more suggestions and one **chore**: please remember to add e2e tests! 

I think once you get Sokratis' approval, we can merge this, and reach out to Nikolay for a small redesign on the page. I don't want to go live with the current layout, but I don't think it should block us from beginning integration testing.

apps/web/src/pages/brand/tabs/v2/BrandingPage.const.ts
```diff
@@ -0,0 +1,23 @@
...
+  DEFAULT_FONT_FAMILY,
+  'Fira Code',
+  'Roboto',
+  'Montserrat',
+  'Open Sans',
+  'Lato',
+  'Nunito',
+  'Oswald',
+  'Raleway',
+];
```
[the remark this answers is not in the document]
BiswaViraj: I wanted to do the same, but it differs on the API side.
To use these types 
we will have to update the UI code to `fontFamily: organization.banding.fontFamily as UserFontFamily`, 
Also, `export const FONT_FAMILIES = [...] as const`. will throw error `readonly not assignable to selectItem[]` when trying to pass it as `data` to the `Select` component

apps/web/src/api/hooks/useUpdateOrganizationBranding.tsx
SokratisVidros: Can we change the API at the same time or do we have backwards compatibility concerns? Is the payload in question used by the web app and <Notification-Center/>?

SokratisVidros (approved):

apps/web/src/api/hooks/useUpdateOrganizationBranding.tsx
SokratisVidros: I imagined so.

SokratisVidros: Noop for now :)

PR #5413, opened 2024-04-16
fix: add domain parser hook and domain condition

jainpawan21:

Add `useDomainParser` Hook
handle invalid domain field value

### Screenshots

<details>

</details>

### DIFF
--- a/apps/web/src/pages/auth/components/QuestionnaireForm.tsx
+++ b/apps/web/src/pages/auth/components/QuestionnaireForm.tsx
@@ -9,7 +9,6 @@ import { JobTitleEnum, jobTitleToLabelMapper, ProductUseCasesEnum } from '@novu/
 import type { ProductUseCases, IResponseError, ICreateOrganizationDto, IJwtPayload } from '@novu/shared';
 import {
   Button,
-  colors,
   Digest,
   HalfClock,
   Input,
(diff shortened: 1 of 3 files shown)


apps/web/src/pages/auth/components/QuestionnaireForm.tsx
```diff
@@ -161,12 +174,6 @@ export function QuestionnaireForm() {
       <Controller
         name="domain"
         control={control}
-        rules={{
```
SokratisVidros, line 164: Alternatively, we can add it as a custom rule and let it trigger as part of react-hook-form lifecycle. For example:

```
rules = {{
  validate: {
    isValidDomain: v => parse(v)...
  }
}}```

For more information PTAL https://www.react-hook-form.com/api/useform/register/#options:~:text=%7D%0A/%3E-,validate,-Function%20%7C%20Object

I am not sure what other benefits this gives us but If I had to guess it would make it simpler to trigger all validations at once on form click or maybe on blur if we want to.

File: apps/web/src/pages/auth/components/QuestionnaireForm.tsx [hunk 2]
```diff
@@ -81,6 +83,17 @@ export function QuestionnaireForm() {
   const onCreateOrganization = async (data: IOrganizationCreateForm) => {
     if (!data?.organizationName) return;
 
+    const val = parse(data?.domain as string);
+
+    if (data.domain && !val.isIcann) {
+      setError('domain', {
+        type: 'custom',
+        message: 'Please make sure you specified a valid domain',
```
SokratisVidros, line 91: To match the tone of other error messages we have in the app, let's revise this as follows:

```suggestion
        message: 'Please provide a valid domain',
```

SokratisVidros (approved):
@jainpawan21 I approve this so that we can ship but PTAL at my comments. I think it's worth using the validate function.

PR #5417, opened 2024-04-16
next > 0.25.x

scopsy:


apps/api/src/app/testing/product-feature.e2e.ts
```diff
@@ -0,0 +1,32 @@
+import { OrganizationRepository } from '@novu/dal';
+import { ApiServiceLevelEnum } from '@novu/shared';
+import { UserSession } from '@novu/testing';
+import { expect } from 'chai';
+
+describe.only('Product feature Test', async () => {
```
scopsy, line 6: @rifont should we remove the only flag here?

rifont: This has been removed.

PR #5424, opened 2024-04-17
fix: navigation width when information architecture is disabled

BiswaViraj:

- Added legacy navigation width to be used when IA feature flag is false.

### Screenshots

---
Before:
image

---

After:
image
image

---

<details>

</details>


antonjoel82 (changes requested):
Great catch! While this is definitely necessary, I think we need to offload this logic to `web`, and instead expose a `width` override.

BiswaViraj:
@antonjoel82 , then we will have to manually add this code to all the places sidebar is being used 
```
  const isInformationArchitectureEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_INFORMATION_ARCHITECTURE_ENABLED);
  const navigationWidth = isInformationArchitectureEnabled ? NAVIGATION_WIDTH : LEGACY_NAVIGATION_WIDTH;
  ```

antonjoel82:
> @antonjoel82 , then we will have to manually add this code to all the places sidebar is being used
> 
> ```
>   const isInformationArchitectureEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_INFORMATION_ARCHITECTURE_ENABLED);
>   const navigationWidth = isInformationArchitectureEnabled ? NAVIGATION_WIDTH : LEGACY_NAVIGATION_WIDTH;
>   //....
>   
>   <Sidebar ... width={navidationWidth}/>
> ```

There should only be one place where we currently use it, right? Either way, putting feature flags for app-specific work into the design-system is an anti-pattern that I'd prefer to avoid. I'm happy to make the changes since you're at your end of day!

BiswaViraj:
> There should only be one place where we currently use it, right? Either way, putting feature flags for app-specific work into the design-system is an anti-pattern that I'd prefer to avoid. I'm happy to make the changes since you're at your end of day!

@antonjoel82, I see will keep in mind about the design-system.
We don't have a single place in web app. It is directly imported from the design-system package in the code
A quick search shows it is used in ~18 files
image

> I'm happy to make the changes since you're at your end of day!

Thank you, if you don't get time to return to this today. I can fix it tomorrow morning

antonjoel82 (approved):
@BiswaViraj on second thought, I think this will be fine for the short-term fix we need it for. In the future, I think we should do our best to keep app-targeted flags out of the design-system, but since we'll be re-doing the DS anyway and given the timeline of the project, I think it's acceptable

PR #5427, opened 2024-04-17
[Design System] - Finish initial base token specification

antonjoel82:

- Finish listing out the primary tokens we may use 
  - These may not be the values we ultimately go with, should help us to stop using magic values in code and instead rely on standards that we set

[Full token output JSON](https://drive.google.com/file/d/1iXyp2dKtVTWIz4lON7-UoHC2MIp0So7U/view?usp=sharing)


libs/design-system/src/panda/radius.tokens.ts
```diff
@@ -0,0 +1,45 @@
...
+    type: 'radius',
+  },
+  pill: {
+    value: '9999px',
+    type: 'radius',
+  },
+});
+
+/** @deprecated */
+export const LEGACY_RADIUS_TOKENS = defineTokens.radii({
```
antonjoel82, line 36: 🗒 **note (non-blocking):** These line-up with values that are used frequently throughout web‏

libs/design-system/src/panda/opacity.tokens.ts
```diff
@@ -0,0 +1,41 @@
...
+    type: 'opacity',
+  },
+  '50': {
+    value: '50%',
+    type: 'opacity',
+  },
+  '60': {
+    value: '60%',
+    type: 'opacity',
+  },
```
antonjoel82, line 40: 🗒 **note (non-blocking):** These are values that appear periodically throughout web, but I think we should move to only using the values above which offer similar appearance without unnecessary variation‏

libs/design-system/src/panda/NovuPandaPreset.ts
```diff
@@ -1,8 +1,12 @@
 import { definePreset } from '@pandacss/dev';
+import { BORDER_TOKENS, BORDER_WIDTH_TOKENS } from './borders.tokens';
 import { COLOR_PALETTE_TOKENS, LEGACY_COLOR_TOKENS } from './colors.tokens';
-import { COLOR_SEMANTIC_TOKENS, LEGACY_COLOR_SEMANTIC_TOKENS } from './semanticColors.tokens';
+import { GRADIENT_TOKENS, LEGACY_GRADIENT_TOKENS } from './gradients.tokens';
+import { LEGACY_RADIUS_TOKENS } from './radius.tokens';
 import { TEXT_RECIPE } from './recipes/text.recipe';
 import { TITLE_RECIPE } from './recipes/title.recipe';
+import { COLOR_SEMANTIC_TOKENS, LEGACY_COLOR_SEMANTIC_TOKENS } from './semanticColors.tokens';
+import { LEGACY_SHADOW_TOKENS } from './shadow.tokens';
```
rifont, line 9: 👏 **praise:** Great use "meta-extensions" (e.g. `.tokens`) to descriptively name files. This pattern makes it very easy to identify relevant files between the different layers of the design system.‏

libs/design-system/src/panda/opacity.tokens.ts
```diff
@@ -0,0 +1,41 @@
+import { defineTokens } from '@pandacss/dev';
+
+/**
+ * Represents the size of an element.
+ *
+ * Used for properties like width and height.
+ */
```
rifont, line 7: ☑ **chore:** It looks like this comment may have been carried over from a different space. Can we remove it? It's already clear what these tokens are used for, I don't think a comment adds value.‏
```suggestion
```

rifont (approved):
Awesome work @antonjoel82 🎉 Just a minor chore to address.

PR #5431, opened 2024-04-18
fix(node): update online status payload error.

stanma9107:

change `online` to `isOnline` to avoid bad request error.


LetItRock:
@stanma9107 hey, thanks a lot for the fix! 🙌 
could you please take a look at failing spec: `src/lib/subscribers/subscribers.spec.ts`?
don't bother about failing API tests ;)

PR #5433, opened 2024-04-18
feat(ci): Docker build cache

AliaksandrRyzhou:
We want to update the runners to have cache for docker, pnpm, etc to decrease the build time.
To bypass the Github cache restriction in 10Gb we decide to use the GitHub package registry as a cache. After the first run the new package with the name "cache" will be created for storing the cache images separated by tags. More information about this type of cache you can see [here](https://docs.docker.com/build/cache/backends/registry/)

WEBHOOK service
**Without the cache:**
https://github.com/novuhq/novu/actions/runs/8738434410
**With the cache:**
https://github.com/novuhq/novu/actions/runs/8739026750

### Screenshots

**Without the cache:**
Screenshot from 2024-04-18 17-41-06

**With the cache:**
Screenshot from 2024-04-18 17-40-47

**Cache package:**
image

### DIFF
--- a/.github/workflows/reusable-docker.yml
+++ b/.github/workflows/reusable-docker.yml
@@ -110,8 +110,8 @@ jobs:
           GH_ACTOR: ${{ github.actor }}
           GH_PASSWORD: ${{ secrets.GH_PACKAGES }}
           DOCKER_BUILD_ARGUMENTS: >
-            --cache-from type=registry,ref=ghcr.io/novuhq/novu/cache:build-cache-${{ env.SERVICE_NAME }}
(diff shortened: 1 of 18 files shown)


.github/workflows/reusable-docker.yml
```diff
@@ -123,6 +134,11 @@ jobs:
           PROJECT_PATH: ${{ inputs.project_path }}
           GH_ACTOR: ${{ github.actor }}
           GH_PASSWORD: ${{ secrets.GH_PACKAGES }}
+          DOCKER_BUILD_ARGUMENTS: >
```
Cliftonz, line 137: **praise:** ‏Love this!

Cliftonz (approved):

PR #5443, opened 2024-04-23
Create a CI pipeline to run cypress and playwright tests 4 times a day

AliaksandrRyzhou:
Create a CI pipeline to run cypress and playwright tests 4 times a day to test a code in the main branch regularly.
GitHub actions schedule is in UTC time.
It will run tests 4 times per working day and the tests start every 2 hours from 12 AM until 19 PM (Github actions schedule is in UTC time.)
**The schedule should be clarified.**
```
  schedule:
   - cron:  '0 12-19/2 * * 1-5'
```
This workflow will checkout the last commit on default branch when it triggers. Currently we've been using `next` branch, when we switch to `main` branch this workflow works on `main` branch.

If tests of the `WEB` service fail then workflows `Deploy DEV WEB` and `Deploy PROD WEB` will be disabled.
If tests of the `Widget` service fail then workflows `Deploy DEV Widget` and `Deploy PROD Widget` will be disabled.
The workflows will be enabled back when the tests finish successfully or you can enable the workflows manually.


.github/workflows/scheduler.yml
```diff
@@ -0,0 +1,67 @@
+name: Test Scheduler
```
SokratisVidros, line 1:
```suggestion
name: Scheduled E2E tests
```

.github/workflows/scheduler.yml
```diff
@@ -0,0 +1,67 @@
...
+  id-token: write
+  actions: write
+
+env:
+  OWNER: ${{ github.repository_owner }}
+  REPO: ${{ github.event.repository.name }}
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  test_web:
```
SokratisVidros, line 20: @AliaksandrRyzhou Can we make the tests run in parallel?

.github/workflows/scheduler.yml
```diff
@@ -0,0 +1,67 @@
...
+          prod-deploy-web.yml: $web_status
+          EOF
+          )
+          while IFS= read -r line; do
+            filename=$(echo "$line" | awk -F ':' '{print $1}')
+            status=$(echo "$line" | awk -F ':' '{print $2}' | awk '{$1=$1};1') # Trim whitespace
+            echo "Workflow $filename status: $status"
+
+            if [[ "$status" == "failure" ]]; then
+              echo "Disabling workflow $filename because some of the tests failed."
```
SokratisVidros, line 59:
```suggestion
              echo "Disabling workflow $filename. The E2E test has failed."
```

.github/workflows/scheduler.yml
```diff
@@ -0,0 +1,67 @@
...
+              gh workflow disable "$filename"
+            elif [[ "$status" == "success" ]]; then
+              echo "Enabling workflow $filename because the tests look good."
```
SokratisVidros, line 62:
```suggestion
              echo "Enabling workflow $filenams. The E2E tests have passed."
```

.github/workflows/scheduler.yml
```diff
@@ -0,0 +1,67 @@
...
+
+env:
+  OWNER: ${{ github.repository_owner }}
+  REPO: ${{ github.event.repository.name }}
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  test_web:
+    name: Test Web Cypress
+    uses: ./.github/workflows/reusable-web-e2e.yml
```
SokratisVidros, line 22: @AliaksandrRyzhou, Have these workflows been built in retries for the cypress part? The assumption is that Cypress will remain unstable so we would like the workflow to retry 1-2 times without manual action.

.github/workflows/scheduler.yml
SokratisVidros: Let's also rename the file to `scheduled_e2e.yml`

SokratisVidros (changes requested):

.github/workflows/scheduler.yml
AliaksandrRyzhou: It's a good idea, I'll take care about it

SokratisVidros (approved):

SokratisVidros:
@AliaksandrRyzhou, when we test this, we need to remove e2e from PR branches for now.

SokratisVidros (approved):

PR #5446, opened 2024-04-24
fix: billing cypress tests

davidsoderberg:


apps/web/cypress/tests/billing/billing.spec-ee.ts
```diff
@@ -61,6 +91,21 @@ describe('Billing', function () {
...
+
+        response.body['data'] = [
+          {
+            ...response.body.data[0],
+            apiServiceLevel: 'business',
+          },
+        ];
+        return response;
+      });
+    }).as('organizations');
```
antonjoel82, line 108: ✏ **suggestion:** Since this is repeated a couple times, perhaps we could wrap all the tests that need it in a shared `describe` block with its own `beforeEach`?‏

antonjoel82 (approved):
Thank you for fixing this, and sorry for missing it on Information Architecture!

rifont (approved):

PR #5451, opened 2024-04-25
feat: add echo usecase and more tracking for getting started

davidsoderberg:
Screenshot 2024-04-25 at 14 54 44


apps/web/src/pages/get-started/consts/shared.tsx
```diff
@@ -16,8 +16,26 @@ export const StepButton = styled(Button)`
...
+  event?: string;
+}) {
+  const segment = useSegment();
+
+  return (
+    <StyledLink
+      {...linkProps}
+      onClick={() => {
+        if (linkProps.event) {
+          segment.track(`${linkProps.event} - [Get Started]`, { href: linkProps.href });
```
djabarovgeorge, line 32: Love it! :)

I wonder if maybe we should send the channel/action by property so that the consumers of the event could group by the channels easily, what do you think?

djabarovgeorge (commented):
Looks good to me left a small question regarding the event structure.

apps/web/src/pages/get-started/consts/shared.tsx
davidsoderberg: oh a great suggestion 🥳

apps/web/src/pages/get-started/consts/EchoUseCase.const.tsx
```diff
@@ -0,0 +1,110 @@
...
+            segment.track(`Copy echo command - [Get Started]`);
+            copy(COMMAND);
+          }}
+        >
+          <Copy width="22" height="22" />
+        </UnstyledButton>
+      </Group>
+    </div>
+  );
+};
```
antonjoel82, line 42: ☑ **chore:** Please use [panda](https://www.notion.so/novuhq/Using-PandaCSS-at-Novu-7b155517c2ff448ca65527fa7b2c9afe) styling for all new work here. Let me know if you have any questions!‏

apps/web/src/pages/get-started/consts/EchoUseCase.const.tsx
```diff
@@ -0,0 +1,110 @@
+import { StepDescription, StepText } from './shared';
+import { OnboardingUseCase } from './types';
+import { OnboardingUseCasesTabsEnum } from './OnboardingUseCasesTabsEnum';
+import { colors, Copy } from '@novu/design-system';
+import { Group, UnstyledButton, useMantineColorScheme } from '@mantine/core';
```
antonjoel82, line 5: ☑ **chore:** Please avoid direct imports to `@mantine/core` as discussed at the R&D offsite‏

apps/web/src/pages/get-started/consts/EchoUseCase.const.tsx
```diff
@@ -0,0 +1,110 @@
+import { StepDescription, StepText } from './shared';
+import { OnboardingUseCase } from './types';
+import { OnboardingUseCasesTabsEnum } from './OnboardingUseCasesTabsEnum';
+import { colors, Copy } from '@novu/design-system';
```
antonjoel82, line 4: ☑ **chore:** Could you please find the equivalent icon to the deprecated one `Copy`?‏

apps/web/src/pages/get-started/consts/EchoUseCase.const.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { Group, UnstyledButton, useMantineColorScheme } from '@mantine/core';
+import { useClipboard } from '@mantine/hooks';
+import { useSegment } from '@novu/shared-web';
+
+const COMMAND = 'npx novu-labs@latest echo';
+
+const CodeSnippet = () => {
```
antonjoel82, line 11: 🤓 **nitpick (non-blocking):** ‏Please move to a separate file -- this seems something that could be reusable with a couple props extracted‏

antonjoel82 (changes requested):
Nice work on the new addition, David! The overall outcome of the PR looks great. Asking for changes to avoid accumulating more tech / design debt

apps/web/src/pages/get-started/components/CodeSnippet.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useClipboard } from '@mantine/hooks';
+import { css, cx } from '../../../styled-system/css';
+import { ClipboardIconButton } from '../../../components';
+
+const codeValueInputClassName = css({
+  '& input': {
+    border: 'none !important',
+    background: 'surface.popover !important',
+    color: 'typography.text.secondary !important',
+    fontFamily: 'mono !important',
```
antonjoel82, line 11: 🗒 **note (non-blocking):** @davidsoderberg Updated this component to use the tokens we already have!

We had a similar functionality in the Information Architecture changes, so I updated the file to leverage that‏

antonjoel82:
One other note -- I think we will want to default to `Echo` tab instead of in-app which may require a couple tweaks

apps/web/src/pages/get-started/consts/EchoUseCase.const.tsx
```diff
@@ -0,0 +1,95 @@
...
+      className={css({ maxW: '400px' })}
+      onClick={() => {
+        segment.track(`Copy echo command - [Get Started]`);
+      }}
+    />
+  );
+};
+
+const textClassName = css({
+  color: 'legacy.B60',
```
antonjoel82, line 26:
```suggestion
  color: 'typography.text.secondary',
```
We should try to only use semantic tokens in app code as much as possible

apps/web/src/pages/get-started/consts/EchoUseCase.const.tsx
```diff
@@ -0,0 +1,95 @@
...
+          </StepDescription>
+        );
+      },
+    },
+    {
+      title: 'Create a workflow',
+      Description: function () {
+        return (
+          <StepDescription>
+            <p className={textClassName}>
```
antonjoel82, line 55:
```
const StepText = styled('p', text);
```
(`text` should be imported from `styled-system/recipes` -- you can search for examples)

```suggestion
            <StepText variant="secondary">
```
We can also use this to get built-in styles

antonjoel82 (approved):
Thank you for your hard work!

PR #5454, opened 2024-04-25
chore(api): Optimize dockerfile - WIP

SokratisVidros:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -1,11 +1,10 @@
 FROM node:20-alpine3.19 as dev_base
 RUN apk add g++ make py3-pip
 
-ENV NX_DAEMON=false
+ENV NX_DAEMON=false \
+    NODE_ENV=production
 
-RUN npm i pm2 -g
 RUN npm --no-update-notifier --no-fund --global install pnpm@8.9.0
-RUN pnpm --version
 
 USER 1000
 WORKDIR /usr/src/app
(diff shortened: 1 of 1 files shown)


apps/api/Dockerfile
```diff
@@ -66,7 +64,5 @@ RUN --mount=type=cache,id=pnpm-store-api,target=/root/.pnpm-store\
  --frozen-lockfile \
  --unsafe-perm
 
-ENV NEW_RELIC_NO_CONFIG_FILE=true
-
 WORKDIR /usr/src/app/apps/api
-CMD [ "pm2-runtime","start", "dist/main.js" ]
+CMD [ "dump-init", "node", "dist/main.js" ]
```
rifont, line 68: Should this one be `dumb-init`?

apps/api/Dockerfile
```diff
@@ -14,9 +13,7 @@ WORKDIR /usr/src/app
 FROM dev_base AS dev
 ARG PACKAGE_PATH
 
-COPY --chown=1000:1000 ./meta .
-COPY --chown=1000:1000 ./deps .
-COPY --chown=1000:1000 ./pkg .
+COPY --chown=1000:1000 ./meta ./deps ./pkg ./
```
AliaksandrRyzhou, line 16: It's interim stage, here is no sense to decrease layers because we won't use dev, it's important to decrease dev_base and prod. Docker caches each line in the Dockerfile, and that the output of one line is the input of the next. So if a line generates new output all subsequent caches are invalidated. In this case we lost an opportunity to use cache even if only last folder will be changed we copy all of them.

apps/api/Dockerfile
```diff
@@ -1,11 +1,10 @@
 FROM node:20-alpine3.19 as dev_base
 RUN apk add g++ make py3-pip
```
AliaksandrRyzhou, line 2: we can exclude cache because we use dev_base in the final step `RUN apk add --no-cache g++ make py3-pip`

apps/api/Dockerfile
SokratisVidros: I followed some best practises and found out about https://github.com/Yelp/dumb-init. Maybe we need it, maybe not. We will review it as soon as this is more testable.

PR #5458, opened 2024-04-26
refactor(app-gen,worker): move add job to worker

djabarovgeorge:
The addJob use case has been migrated from the application-generic component to the worker application. This improves the separation of concerns by keeping worker-related functionality within the worker itself. Previously, application-generic included addJob to support adding jobs via the API. However, this restriction is no longer applicable.

Some additional validation needed to be done.  This is because the application-generic app allows null values with strictNullChecks set as false, while the worker enforces stricter type safety with strictNullChecks set to true.

### Screenshots

<details>

</details>


scopsy (approved):
Amazing work!

PR #5460, opened 2024-04-26
fix(ci): Instant rollback process

AliaksandrRyzhou:
# Rollback the ECS and Netlify.

So that we don't have to use the Web interface to roll back our ECS and Netlify releases, we decided to create a workflow that should simplify this procedure.

## Workflow options

You can find this workflow [here](https://github.com/novuhq/novu/actions/workflows/rollback.yml).

When you click the button `Run workflow` you will see the menu with the following options:

**Use workflow from**

Specify the branch name to be used when running the Rollback workflow. Keep in mind that to use these options, the latest `rollback.yml` should exist in your branch.

**Service**:

- inbound_mail
- api
- web
- webhook
- widget
- worker
- ws

The micro-service that we want to rollback to the previous version.

**Environment**:

- Development
- Production

The environment where the service is located. Mostly, I've added this option for testing, but let's leave it here until we're assured that everything works fine.

**Region**:

- [EU]
- [US]
- [EU,US]

The region of the environment is necessary to know the TF workspace. When you choose the DEV environment, this option is not needed. However, if you attempt to run the Rollback for the Development environment in the US Region, it will return an error because we don't have a DEV environment in this region.

**Rollback mode**:

- Previous Version
- Commit Hash

You can roll back to the previously deployed version or to the version that has the current commit hash of this branch in an image tag name or a deployment info. The `Commit Hash` option should be tested as it will be useful for tags. When a tag is created and this workflow is run for that tag, the workflow retrieves the HEAD Commit for the current git state. It then attempts to find the `image tag` with this value or the Deploy in Netlify with the same `commit_ref` roll back the release from the current version to the version with this head commit value.

More you can find [here](https://www.notion.so/novuhq/Production-Release-Rollback-Guide-0dd4556094884a32b5825148239dbe37#c7e7d8ea63d248cda2e9ca72c3b33021)

Remember that this workflow works only in the branch where rollback.yml exists and is updated.


.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+        options:
+          - Development
+          - Production
+      region:
+        type: choice
+        description: The region of the environment. It's needed to know TF workspace. When you choose DEV environment this option is not needed.
+        options:
+          - [EU]
+          - [US]
+          - [EU,US]
```
SokratisVidros, line 30: Let's put this first so that the default mode is both regions.

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+          - web
+          - webhook
+          - widget
+          - worker
+          - ws
+      environment:
+        type: choice
+        description: The environment where the service is located.
+        options:
+          - Development
```
SokratisVidros, line 22: Let's add production first, so it's select by default. My guess is that we should almost never rollback development.

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+      - run: echo "Rolling back ${{ github.event.inputs.service }} in ${{ github.event.inputs.environment }}"
+
+      - id: commit
+        if: contains(fromJson('["Commit Hash"]'), github.event.inputs.mode)
+        uses: prompt/actions-commit-hash@v3
+
+      - name: Prepare variables
+        id: variables
+        run: |
+          if [[ "${{ matrix.region }}" == "EU" && "${{ github.event.inputs.environment }}" == "Production" ]]; then
```
SokratisVidros, line 61: **question:** ‏ I am curious, is there a way to declaring a static mapping object and avoid the if/elif logic in a Github action?

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ env.aws_region }}
+
+      - name: ECS get output
+        if: contains(fromJson('["Previous Version"]'), github.event.inputs.mode)
+        id: ecs-output
+        run: |
+          echo "Retrieving current_task_definition_arn..."
```
SokratisVidros, line 121: @AliaksandrRyzhou How much time does it take to rollback? There are several API calls and we need to make sure this action is fast.

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+        run: |
+          aws ecs update-service --cluster ${{ env.ecs_cluster }} --service ${{ env.ecs_service }} --task-definition ${{ env.PREVIOUS_TASK }}
+          aws ecs wait services-stable --cluster ${{ env.ecs_cluster }} --service ${{ env.ecs_service }}
+          echo "After Rollback:"
+          echo "The previous task definition: $(echo $CURRENT_TASK | awk -F'task-definition/' '{print $2}')"
+          echo "The current task definition: $(echo $PREVIOUS_TASK | awk -F'task-definition/' '{print $2}')"
+
+
+  netlify:
+    if: contains(fromJson('["web", "widget"]'), github.event.inputs.service)
```
SokratisVidros, line 196: 😍

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
 name: Rollback
+run-name: Rollback the ${{ inputs.service }} service in the ${{ inputs.environment }} environment
 
 on:
   workflow_dispatch:
+    inputs:
+      service:
+        type: choice
+        description: The microservice which we want to rollback to the previous version.
```
SokratisVidros, line 9:
```suggestion
        description: Select service to rollback
```

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+          - inbound_mail
+          - api
+          - web
+          - webhook
+          - widget
+          - worker
+          - ws
+      environment:
+        type: choice
+        description: The environment where the service is located.
```
SokratisVidros, line 20:
```suggestion
        description: Select the environment
```

.github/workflows/rollback.yml
```diff
@@ -1,11 +1,287 @@
...
+        options:
+          - Development
+          - Production
+      region:
+        type: choice
+        description: The region of the environment. It's needed to know TF workspace. When you choose DEV environment this option is not needed.
```
SokratisVidros, line 26:
```suggestion
        description: Select the environment region. Required only in production.
```

SokratisVidros: TIL that Github actions don't support [dynamic, dependent inputs yet.](https://github.com/orgs/community/discussions/11795)

SokratisVidros (changes requested):
@AliaksandrRyzhou this is great work. Kudos for combining all services, environments and regions in a single workflow! We should apply the same mindset in the deployment workflows very soon.

I added some comments to discuss but overall this is already in great shape.

.github/workflows/rollback.yml
AliaksandrRyzhou: > **question:** ‏ I am curious, is there a way to declaring a static mapping object and avoid the if/elif logic in a Github action?

It allows us to use matrix and not to write a lot of steps for each static value, I agree, it looks not super great but in this case it exclude extra part of code. Anyway, I got you @SokratisVidros  and I'll think on it.

.github/workflows/rollback.yml
AliaksandrRyzhou: @SokratisVidros Regarding ECS, the rollback initialization starts already in 30 seconds but other part of rollback depends on ECS update mechanism (Task definition update, health checks, Load Balancer checks etc). I've added option aws ecs wait services-stable so that we be assure that our service was deployed and full rollback takes about 4 minutes if the service is working. (https://github.com/novuhq/novu/actions/runs/8877861439/job/24372292994)
Netlify works much more faster, only a few second to make a previous deploy published. https://github.com/novuhq/novu/actions/runs/8923431167/job/24507673237

.github/workflows/rollback.yml
AliaksandrRyzhou: > TIL that Github actions don't support [dynamic, dependent inputs yet.](https://github.com/orgs/community/discussions/11795)

It's a very useful feature

Cliftonz:
@AliaksandrRyzhou  Lets make sure sok does not have any other discussion point tomorrow otherwise I am good to merge this, dont forget we need to cherry pick this to main also.

@SokratisVidros

SokratisVidros (approved):

AliaksandrRyzhou:
@Cliftonz I'm waiting for your approval :) Anyway, we have to merge it to have a chance of how it works, If something happens we will fix it

Cliftonz (approved):

PR #5461, opened 2024-04-26
fix(ci): Cypress E2E has been disabled for PR temporary

AliaksandrRyzhou:
We decided to disable E2E tests on PR request, when E2E is fast and stable, we will restore it

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,7 +93,8 @@ jobs:
   test_web:
     name: Test Web Cypress
     needs: [get-affected]
-    if: ${{ contains(fromJson(needs.get-affected.outputs.test-cypress), '@novu/web') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -93,7 +93,8 @@ jobs:
   test_web:
     name: Test Web Cypress
     needs: [get-affected]
-    if: ${{ contains(fromJson(needs.get-affected.outputs.test-cypress), '@novu/web') }}
+#    if: ${{ contains(fromJson(needs.get-affected.outputs.test-cypress), '@novu/web') }}
```
SokratisVidros, line 96: Please add a comment why we are doing this.

SokratisVidros (approved):
🚀

PR #5466, opened 2024-04-27
[Information Architecture] - Routing tweaks

antonjoel82:

- Using feature flags with routing and having default values set has caused some issues based on when feature flags load
- This PR ensures that opening the app directly to `/settings/profile` with a search param doesn't reroute and lose the param

- Also fixes the coloring of cards in dark mode since that was bugging me 😅

Fixes [DES-258](https://linear.app/novu/issue/DES-258/🏛️-[infarch]-reset-password-flow-doesnt-work)

Repro:

1. Org w/ Information Architecture disabled, route to `/settings/profile?view=password`
    1. Re-routes to settings root page
1. Org w/ Information Architecture _enabled_, route to `/settings/profile?view=password`
    1. Routes successfully to profile with password sidebar open

https://www.loom.com/share/4801c76d06ee4f8a95c96b9e7d7c566d?sid=20b39b23-c787-4009-aafb-bcf25ceff45f


libs/design-system/src/cards/CardTile.tsx
```diff
@@ -1,7 +1,11 @@
 import styled from '@emotion/styled';
+import { useMantineTheme } from '@mantine/core';
 import { colors, shadows } from '../config';
 
-export const CardTile = styled.button`
+export const CardTile = styled.button(() => {
+  const theme = useMantineTheme();
+
+  return `
```
antonjoel82, line 8: 🗒 **note (non-blocking):** This fixes the ugly dark mode issue due to a bug I've brought up with Mantine theming‏
Screenshot 2024-04-27 at 1 13 25 PM

BiswaViraj (approved):

PR #5467, opened 2024-04-29
[Design System] - New Package

antonjoel82:

- Create a new package for the design system!
- Bring-in Panda tokens + setup
- Update build systems such that we can use a single import for our components & Panda-generated code
- _The name `novui` (Novu + UI) is not final and can be replaced. It was used to be clearly different than the existing package so as not to confuse the two_

Solves [NV-3691](https://linear.app/novu/issue/NV-3691/%E2%9A%A1-package-setup-to-improve-dx) and [NV-3693](https://linear.app/novu/issue/NV-3693/%E2%9A%A1-investigate-and-set-up-the-pandacss-eslint-plugin)

### Notes

- This PR will fail the build because all Panda imports need to be updated -- this is done in a different PR (#5546) to avoid 40+ files of only import changes
- I had to abandon `tsup` :( I repeatedly got issues with `React is not defined` when importing the `Test` component in `web`, so went back to plain ol' `tsconfig`
- Storybook and vitest are configured in #5569 

### Screenshots

This incredible UI showcases a component built in the new  DS package next to a component with styles overwritten in `web` using the generated panda code imported from the new package -- all using tokens for colors

Screenshot 2024-05-13 at 12 03 13 PM

### Future iterations

- Turn-on "strict" tokens for `web` -- currently this causes a lot of errors for tokens that shouldn't seemingly be excluded from strict


apps/web/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "forceConsistentCasingInFileNames": true,
     "noFallthroughCasesInSwitch": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "Node16",
```
antonjoel82, line 13: 🗒 **note (non-blocking):** This was necessary to enable using the "new" `exports` field in `package.json` (in `novui/package.json`). This permits imports like `@novu/novui/css` which actually points to `./styled-system/css` instead of `dist`‏

apps/web/src/pages/invites/components/MembersTable.cy.tsx
```diff
@@ -1,7 +1,7 @@
 import { MembersTable } from './MembersTable';
 import { TestWrapper } from '../../../testing';
 import { MemberRoleEnum, MemberStatusEnum } from '@novu/shared';
-import { SinonSpy } from 'cypress/types/sinon';
+import { SinonSpy } from 'sinon';
```
antonjoel82, line 4: 🗒 **note (non-blocking):** This broke when changing module resolution in tsconfig -- also just not ideal anyway‏

apps/web/src/components/nav/NavMenuButton/NavMenuLinkButton.tsx
```diff
@@ -1,7 +1,7 @@
 import { FC, PropsWithChildren, useState } from 'react';
 import { NavLink } from 'react-router-dom';
-import { css } from '../../../styled-system/css';
-import { HStack } from '../../../styled-system/jsx';
+import { css } from '@novu/novui/css';
```
antonjoel82, line 3: 🗒 **note (non-blocking):** ‏Example imports using the new DS

libs/novui/package.json
```diff
@@ -0,0 +1,76 @@
...
+      "types": "./styled-system/recipes/index.d.ts",
+      "require": "./styled-system/recipes/index.js",
+      "import": "./styled-system/recipes/index.js"
+    },
+    "./jsx": {
+      "types": "./styled-system/jsx/index.d.ts",
+      "require": "./styled-system/jsx/index.js",
+      "import": "./styled-system/jsx/index.js"
+    },
+    "./styles.css": "./styled-system/styles.css"
```
antonjoel82, line 48: 🗒 **note (non-blocking):** Enables better imports for package consumers. Requires `moduleResolution: >= Node16` in the consumer‏

apps/web/src/pages/templates/TemplatesListNoData.tsx
```diff
@@ -10,6 +10,8 @@ import { IBlueprintTemplate } from '../../api/types';
 import { TemplateCreationSourceEnum } from './shared';
 import { useHoverOverItem } from '../../hooks';
 import { EchoProjectCardTile } from './components/EchoProjectWaitList';
+import { Test } from '@novu/novui';
+import { css } from '@novu/novui/css';
```
antonjoel82: 🗒 **note (non-blocking):** ‏Example of new imports!

apps/web/src/pages/templates/TemplatesListNoData.tsx
```diff
@@ -73,7 +75,10 @@ export const TemplatesListNoData = ({
 
   return (
     <NoDataHolder data-test-id="no-workflow-templates-placeholder">
-      <NoDataSubHeading>Start from a blank workflow or use a template</NoDataSubHeading>
+      <NoDataSubHeading className={css({ color: 'legacy.success !important' })}>
+        Start from a blank workflow or use a template
+      </NoDataSubHeading>
+      <Test />
```
antonjoel82: 🗒 **note (non-blocking):** We will remove this before merging! This is just to validate that everything is working across the two‏

SokratisVidros (approved):

PR #5471, opened 2024-04-29
fix: sending email provider error to sentry

davidsoderberg:


apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -455,8 +455,7 @@ export class SendMessageEmail extends SendMessageBase {
         'mail_unexpected_error',
         error.message || error.name || 'Error while sending email with provider',
         command,
-        LogCodeEnum.MAIL_PROVIDER_DELIVERY_ERROR,
-        error
```
davidsoderberg, line 459: Not providing the error to this method makes sure it does not pass it to sentry

SokratisVidros (approved):

PR #5473, opened 2024-04-29
feat(ci): E2E API tests have been disabled

AliaksandrRyzhou:
We decided to disable E2E tests on PR requests, when E2E is fast and stable, we will restore it

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -178,7 +178,7 @@ jobs:
           targets: lint,build,test
           projects: ${{join(fromJson(needs.get-affected.outputs.test-libs), ',')}}
 
-  test_api:
+  test_unit_api:
     name: Test API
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -181,6 +181,7 @@ jobs:
   test_api:
```
SokratisVidros, line 181: Shall we rename these actions to `test_e2e_api` and `test_unit_api` to make them more specific for the feature reader?

SokratisVidros (approved):

.github/workflows/test.yml
AliaksandrRyzhou: I guess we can rename `test_e2e_api` but `test_unit` runs not only for API as it seems to me

PR #5474, opened 2024-04-29
feat: add backend proxy for segment

davidsoderberg:


apps/api/src/app/analytics/analytics.controller.ts
```diff
@@ -0,0 +1,18 @@
+import { Body, Controller, Post } from '@nestjs/common';
+import { AnalyticsService } from '@novu/application-generic';
+
+@Controller({
+  path: 'analytics',
+})
+export class AnalyticsController {
+  constructor(private analyticsService: AnalyticsService) {}
+
+  @Post('/telemetry')
```
davidsoderberg, line 10: Should we add auth guard here and use that userId instead?

rifont: Yes we definitely should. Otherwise, anyone could register fake events.

rifont, line 10: ✏ **suggestion:** What do we think about aligning with the base analytics syntax, so here we could have a `/track` endpoint?
‏
```suggestion
  @Post('/track')
```

rifont (commented):
It's not currently covered in this PR - do we need to address(/remove) the Segment Identify/Group call that happens too?

apps/api/src/app/analytics/analytics.controller.ts
```diff
@@ -0,0 +1,18 @@
+import { Body, Controller, Post } from '@nestjs/common';
+import { AnalyticsService } from '@novu/application-generic';
+
+@Controller({
+  path: 'analytics',
+})
+export class AnalyticsController {
```
rifont, line 7: Once we add the Auth guard, we also need to use the `@SkipThrottle` decorator to ignore these endpoints from consuming the rate limiting quota for the User's Organization.

davidsoderberg:
> It's not currently covered in this PR - do we need to address(/remove) the Segment Identify/Group call that happens too?

Group is not used by frontend and identify only sends user id so we should already know that as we can get the user id from the user session.

apps/api/src/app/analytics/analytics.controller.ts
```diff
@@ -0,0 +1,22 @@
+import { Body, Controller, Logger, Post, UseGuards } from '@nestjs/common';
+import { SkipThrottle } from '@nestjs/throttler';
+import { AnalyticsService, UserAuthGuard, UserSession } from '@novu/application-generic';
+import { IJwtPayload } from '@novu/shared';
+
+@Controller({
+  path: 'analytics',
```
scopsy, line 7: Would suggest calling this "telemetry" and the endpoint "measure" as some blockers will look for "track" keywords

apps/api/src/app/analytics/analytics.controller.ts
scopsy: Just seen your comment :P I think actually it's better to use `/telemtry/measure` or something like this, as from experience a lot of blockers look for the analytics keyword or "track" keywords for blocking requests.

apps/api/src/app/analytics/analytics.controller.ts
rifont: Good to know, didn't realise those blockers were so aggressive. I agree, we should definitely make this change.

scopsy (approved):

rifont (approved):
LGTM! Let's 🚢 it

apps/api/src/app/analytics/analytics.controller.ts
```diff
@@ -0,0 +1,22 @@
...
+import { IJwtPayload } from '@novu/shared';
+
+@Controller({
+  path: 'telemetry',
+})
+@SkipThrottle()
+export class AnalyticsController {
+  constructor(private analyticsService: AnalyticsService) {}
+
+  @Post('/measure')
```
SokratisVidros, line 13:
```suggestion
  @Post('/track')
```

SokratisVidros:
Should we feature flag this endpoint based on the presence of `MIXPANEL_TOKEN` and `SEGMENT_TOKEN`? We would like this endpoint to be opt-in in standalone deployments.

libs/shared-web/src/utils/segment.ts
```diff
@@ -21,12 +22,11 @@ export class SegmentService {
...
   }
 
-  track(event: string, data?: Record<string, unknown>) {
-    if (!this.isSegmentEnabled()) {
-      return;
-    }
-
-    this._segment?.track(event + ' - [WEB]', data);
+  async track(event: string, data?: Record<string, unknown>) {
+    await api.post('/v1/telemetry/measure', {
```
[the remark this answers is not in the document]
scopsy: @SokratisVidros just and FYI: https://github.com/novuhq/novu/pull/5474#discussion_r1584280307

I have no strong opinions on this, just encountered some cases where adblockers will look for keywords such as "ads" "track" and so on in the past. Might not be relevant anymore, but that's why David used this path on my initial comment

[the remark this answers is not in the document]
SokratisVidros: I don't think it's relevant since we are proxying, but I also don't feel strongly about naming.

davidsoderberg:
> Should we feature flag this endpoint based on the presence of `MIXPANEL_TOKEN` and `SEGMENT_TOKEN`? We would like this endpoint to be opt-in in standalone deployments.

Not sure I understand your comment 💯 but there is a check in the analytics service for the `SEGMENT_TOKEN` so the standalone deployments will not get an error if the endpoint will be called without it being present.

scopsy:
@davidsoderberg I think he meant don't make the http request on the first place if ff is not present to avoid the network activity.

davidsoderberg:
> @davidsoderberg I think he meant don't make the http request on the first place if ff is not present to avoid the network activity.

So this should be it then :)

https://github.com/novuhq/novu/pull/5474/commits/5e6e0cfbf86dbd00636c657573796e11c9cfe91f

libs/shared-web/src/utils/segment.ts
```diff
@@ -21,12 +22,15 @@ export class SegmentService {
     this._segment?.identify(user?._id);
   }
 
-  track(event: string, data?: Record<string, unknown>) {
+  async track(event: string, data?: Record<string, unknown>) {
     if (!this.isSegmentEnabled()) {
       return;
```
SokratisVidros, line 27: Does this return 404?

SokratisVidros:
Thanks, @davidsoderberg, for the clarification. My initial comment was about not registering the Analytics module at all unless both env vars are set. We might be able to abstract the env var check via a static `isEnabled` check on the module class.

apps/api/src/app.module.ts
```diff
@@ -134,6 +133,10 @@ if (process.env.SENTRY_DSN) {
   });
 }
 
+if (process.env.SEGMENT_TOKEN) {
+  modules.push(AnalyticsModule);
+}
```
scopsy, line 138: @davidsoderberg won't it break all self-hosted applications? since they are trying to import not available module? (analytics service)

scopsy: cc. @SokratisVidros

davidsoderberg: No we use analytics service everywhere or have we moved it to enterprise?

scopsy: We import it everywhere. It's just not reporting anything (returning the track) method immediately. I believe this change will break the API in the current state.

My suggestion is leave it as it was before the commit. And let's merge this PR already :)

SokratisVidros: Are you talking about the AnalyticsService or the AnalyticsModule introduced in this PR? The service is still imported as ESM so I am not sure I understand how the conditional import breaks it.

scopsy: The Analytics Module exports a service called AnalyticsService, we inject this service in many usecases in the code. If the module is not present, all those injections will fail in boot time because the service is not available. (or at least that's what I think :P )

scopsy: nvm, i thought the the analyticsmodule exposes the analytics service. in practices the analytics service is living in SharedModule. Analytics module is just the new endpoint for proxying 😅

Issue #5445, opened 2024-04-23
🐛 Bug Report: Error when copy/paste template code to email editor - Failed to extract variables


lucafaggianelli:
### 📜 Description

When I copy/paste some template code to the email editor (typically because I created that code in other notifications channels) I receive an error when trying to save: Failed to extract variables

Indeed the custom variables are not present in variables list when clicking on the pencil icon

### 👟 Reproduction steps

1. Create a in-app channel in a workflow
2. Write some template code with some custom variables, i.e.:
```
{{#if step.digest}}
{{step.total_count}} events matched with bank transactions.
{{else}}
The expense {{expense.name}} matched with a bank transaction.
{{/if}}
```
3. copy the code
4. create an email channel in the workflow
5. paste the code you copied before
6. Click Update/Save

### 👍 Expected behavior

It should correctly save the email content recognizing the variables

### 👎 Actual Behavior with Screenshots

![image](https://github.com/novuhq/novu/assets/4471043/d5ae4f81-36a6-4590-909a-6526a5bf2843)

### Novu version

Novu SaaS

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

YugfeldID:
@lucafaggianelli you can paste template code into **Custom code** tab in the Email editor, it works for me

YugfeldID:
Also, it seems like the problem occurs because when we type in the in-app editor text like this:
`The expense {{expense.name}} `

it is represented in DOM like this:
```
<span>The&nbsp;expense&nbsp;</span>   
<span >{{</span>
<span> expense.name</span>
<span >}}</span>
```
and after copying and pasting into email editor, 'data/html' will be pasted like in the example above. But email editor is not rich text editor, so when the app tries to find variables it parses HTML tags and can't find variables.

I assume it will be proper to prohibit pasting any 'data/html' into email editor not only because variables are not parsed correctly, but also because all styles and fonts are pasted with text now and there are no controls to clear them.
If it's okay, I can do it.

lucafaggianelli:
[...]

PR #5478, opened 2024-04-30
fix: prohibit paste text/html content into email editor, allow paste …

YugfeldID:
fix for [#5445](https://github.com/novuhq/novu/issues/5445)

Blocks paste **text/html** content into the email editor (allow paste only **text/plain**). Problem: when the user copies text from the rich text editor (like the editor from the In-App channel) all styles will be copied, it leads to an inability to process variables and clear formatting.

### Screenshots
Without fix pasted content can look like this (with the inability to change the font, clear background, remove list items, and bold formatting):
Снимок экрана 2024-04-30 в 16 59 29

If this approach is okay, I will add a test


LetItRock (commented):
hello @YugfeldID! 👋 thanks for the contribution! 🤝 

I think that this is the right approach for now, but I noticed some problems with the current implementation:
- variables are not recognized
- the undo (cmd + Z) works weird

My feelings is that the inserted text node is not recognized somehow as a text change that's why the weird behaviour with undo. The another solution is to use `document.execCommand('insertText', false, pastedData);` but its deprecated and not recommended.

YugfeldID:
> hello @YugfeldID! 👋 thanks for the contribution! 🤝
> 
> I think that this is the right approach for now, but I noticed some problems with the current implementation:
> 
> * variables are not recognized
> * the undo (cmd + Z) works weird
> 
> My feelings is that the inserted text node is not recognized somehow as a text change that's why the weird behaviour with undo. The another solution is to use `document.execCommand('insertText', false, pastedData);` but its deprecated and not recommended.

Hello @LetItRock, thanks for feedback.

I pushed some changes:
- `methods.setValue(``${stepFormPath}.template.content.${blockIndex}.content``, ref.current?.innerHTML ?? '') ` should trigger changes -> variables will be parsed and placeholder will be gone
-  you are right, it seems like the only way to make undo works properly is using `document.execCommand('insertText', false, pastedData);`, because only it interacts with browser's undo manager

PR #5481, opened 2024-05-01
Remove logging for health check

Cliftonz:
This pull request is aimed at removing logging from our health-check endpoint. 

This decision comes in light of our continuous efforts towards cost optimization and efficient resource utilization.
The health-check endpoint is frequently hit to ensure the system's healthiness, whether it's by internal services, load-balancers, or monitoring systems. Each of these hits generates a log entry, which, over time, amounts to a significant volume of logs. This volume leads to increased storage needs and consequently higher costs associated with log storage and management.

While these logs do provide information, they offer little to no actionable value since health-checks usually don't carry business-critical data. The routine nature of these requests and the rare probability of error occurrence make the logs from health-check endpoints low in value yet high in cost. Hence, by removing logging from the health-check endpoint, we are aiming to:
- Reduce the costs associated with log storage and management
- Make our logging more efficient by concentrating only on value-adding logs
- Lessen the noise in our logging, making it easier to focus on more critical and actionable logs


scopsy (approved):

LetItRock (approved):
looks good! 🙌

SokratisVidros (approved):

PR #5493, opened 2024-05-03
Nv 2906 remove the is topic notification enabled feature flag

denis-kralj-novu:
A feature flag that is always set to true has been removed, along with all associated 'false' code paths. Additionally, dead code was also removed and used helpers moved to more appropriate places

I've placed 2 options in this MR and we should agree on what option we want to use. I'm leaning more towards Option 1 as it complains right away, not only when tests are run. Option 2 also seems a bit ugly, but that is more personal preference.


libs/shared/src/types/feature-flags/flags.types.spec.ts
```diff
@@ -62,3 +62,15 @@ testFlagEnumValidity(InvalidValueFlagsEnum);
  */
 testFlagEnumValidity(FeatureFlagsKeysEnum);
 testFlagEnumValidity(SystemCriticalFlagsEnum);
+
```
denis-kralj-novu, line 65: Option 1:
Defining a type limiter here that requires us to update this file every time we introduce a new enum (same as with the `testFlagEnumValidity()` function). Reports error before build.

libs/shared/src/types/feature-flags/flags.types.ts
```diff
@@ -17,4 +17,16 @@ export type IFlagKey = `IS_${Uppercase<string>}_ENABLED`;
  */
 export function testFlagEnumValidity<TEnum extends IFlags, IFlags = Record<IFlagKey, IFlagKey>>(
   testEnum: TEnum & Record<Exclude<keyof TEnum, keyof IFlags>, ['Key must follow `IFlagKey` format']>
-) {}
+) {
```
denis-kralj-novu, line 20: Option 2:
Explicit testing of match between the key and value for the enum, works for all enums provided, only shows failure when tests get run.

libs/shared/src/types/feature-flags/flags.types.spec.ts
denis-kralj-novu: Agreed with @ainouzgali to go with this option

apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
```diff
@@ -49,7 +49,6 @@ describe('Topic Trigger Event', () => {
 
     beforeEach(async () => {
       process.env.LAUNCH_DARKLY_SDK_KEY = '';
```
ainouzgali, line 51: You can also remove all the process.env.LAUNCH_DARKLY_SDK_KEY references here :) and the afterEach actions

packages/application-generic/src/usecases/map-trigger-recipients/map-trigger-recipients.use-case.ts
ainouzgali, line 1: So odd that this usecase isn't being used. But it seems so 🤷‍♀️  Good job finding it

ainouzgali (commented):
@denis-kralj-novu looks great💃🏼
Just note my comment about removing the launch darkly env variable in the test files

packages/application-generic/src/usecases/map-trigger-recipients/map-trigger-recipients.use-case.ts
denis-kralj-novu: Finding it was a side-effect of working on this task. I saw it used the flag and then just checked the uses of `execute()` to make sure I didn't miss anything. That lead to finding that it is dead code save for the one helper that I moved from it closer to use. Talked with @rifont about it, seems that there was some refactor and this was left dead in the transition.

apps/api/src/app/events/e2e/trigger-event-topic.e2e.ts
denis-kralj-novu: Sure, I will remove those as well, good comment

ainouzgali (approved):
💃🏼

PR #5500, opened 2024-05-06
E2E fail slack notification + scheduled API tests

AliaksandrRyzhou:
We want to be informed about how e2e tests are going. 
image

These changes will add the Slack notification when e2e tests fail to tell us that a workflow has been disabled. 

Also, as it was discussed [here](https://novu.slack.com/archives/C06DU7A7BPV/p1714988094136339?thread_ts=1714984786.812179&cid=C06DU7A7BPV) I've added API tests to the schedule


.github/workflows/reusable-web-e2e.yml
```diff
@@ -186,12 +200,25 @@ jobs:
         path: playwright-report
         retention-days: 14
 
+    - name: Notify Slack Action
+      uses: ravsamhq/notify-slack-action@v2
+      if: ${{ always() && inputs.slack_notification }} && ${{ needs.e2e_web.result == 'failure' }}
```
rifont, line 205: ❓ **question:**  Why is the `always()` expression ([ref to docs](https://docs.github.com/en/actions/learn-github-actions/expressions#always)) needed here? It seems redundant when the `inputs.slack_notification` is taking effect already‏

rifont (approved):
Awesome work @AliaksandrRyzhou , very clean! Added one question for review, it's non-blocking. ⚡

.github/workflows/reusable-web-e2e.yml
AliaksandrRyzhou: @rifont if some step of the merge-reports job fails then a job stops on a failed step and we won't receive a notification. Because the merge-reports job is a part of WEB tests we have to get a notification even when this job fails.

PR #5513, opened 2024-05-07
feat(web,design-system,novui): Add embeddable docs

davidsoderberg:
https://www.loom.com/share/ed59a7cc9b834426811994be310e81c6?sid=03dfc7a1-13cf-42c1-8c03-7a31efcf7bae

Screenshot 2024-05-14 at 09 44 52
Screenshot 2024-05-14 at 09 45 01
Screenshot 2024-05-14 at 09 59 44
Screenshot 2024-05-15 at 08 55 19
Screenshot 2024-05-15 at 08 56 32
Screenshot 2024-05-15 at 08 56 38


apps/web/src/AppRoutes.tsx
```diff
@@ -124,6 +125,7 @@ export const AppRoutes = () => {
           <Route path="" element={<LayoutsListPage />} />
         </Route>
         <Route path="/translations/*" element={<TranslationRoutes />} />
+        <Route path="/docs/*" element={<DocsPage />} />
```
scopsy, line 128: @davidsoderberg how is this going to work? Why do we need a new page for docs if it's accessed via a side menu?

davidsoderberg: Right now I am only doing this to be able to implement the docs component until I know how we will display it inside of the web. Radek and Nik will sync about this :)

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+const getMDXComponent = mdxBundler.getMDXComponent;
+
+export const Docs = ({ path = '', children }: { path?: string; children: ReactNode }) => {
+  const segment = useSegment();
+
+  const { isLoading, data: { code, title, description } = { code: '', title: '', description: '' } } = useQuery(
+    ['docs', path],
+    async () => {
+      const response = await fetch('http://localhost:5173/' + path);
```
davidsoderberg: This needs to change before merge

davidsoderberg:
Here is the backend service for the docs as well https://github.com/novuhq/cloud-doc

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+  };
+
+  useEffect(() => {
+    if (!enabled) {
+      return;
+    }
+    if (localStorage.getItem('inline-docs-intro') === 'false') {
+      return;
+    }
+    setOpened(false);
```
davidsoderberg, line 93: This will be changed to true when we are ready to promote the inline docs.

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
+import { ActionIcon, Popover, useMantineColorScheme } from '@mantine/core';
+import { Button, colors, IconOutlineMenuBook, QuickGuide, shadows, Tooltip } from '@novu/design-system';
```
antonjoel82: ☑ **chore:** Please avoid things that should be done via design tokens!‏
```suggestion
import { Button, IconOutlineMenuBook, QuickGuide, Tooltip } from '@novu/design-system';
```

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
+import { ActionIcon, Popover, useMantineColorScheme } from '@mantine/core';
```
antonjoel82: Please avoid direct imports from `@mantine/core` as discussed [here](https://www.notion.so/novuhq/Using-PandaCSS-at-Novu-7b155517c2ff448ca65527fa7b2c9afe?pvs=4#fe38738ef93b42b5af98f66d0142b052).
```suggestion
import { useColorScheme } from '@novu/design-system';
```

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useDocsContext } from '../providers/DocsProvider';
+
+const Title = styled('h3', title);
+const Text = styled('p', text);
+
+export const DocsButton = () => {
+  const { colorScheme } = useMantineColorScheme();
+  const isDark = colorScheme === 'dark';
+  const { toggle, enabled, path } = useDocsContext();
+  const [opened, setOpened] = useState(false);
```
antonjoel82: 🤓 **nitpick (non-blocking):** ‏
```suggestion
  const [opened, setOpened] = useState<boolean>(false);
```

apps/web/src/components/docs/DocsButton.tsx
antonjoel82: 💭 **thought (non-blocking):** Seems like a great opportunity for a feature flag then!‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
+import { ActionIcon, Popover, useMantineColorScheme } from '@mantine/core';
+import { Button, colors, IconOutlineMenuBook, QuickGuide, shadows, Tooltip } from '@novu/design-system';
+import { useSegment } from '@novu/shared-web';
+import { useEffect, useState } from 'react';
+import { css } from '../../styled-system/css';
+import { Flex, styled } from '../../styled-system/jsx';
```
antonjoel82: 👏 **praise:** Thank you for using the Panda utils!‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+            <ActionIcon variant="transparent" onClick={() => toggle()}>
+              <IconOutlineMenuBook />
+            </ActionIcon>
+          </Popover.Target>
+          <Popover.Dropdown
+            style={{
+              borderRadius: '12px',
+              boxShadow: isDark ? undefined : shadows.dark,
+              background: isDark ? colors.B30 : undefined,
+            }}
```
antonjoel82: ☑ **chore:** Please only use Panda here -- this will not use design tokens otherwise‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+          >
+            <Flex gap="125" justify="space-between">
+              <QuickGuide />
+              <div>
+                <Title
+                  className={css({
+                    marginBottom: '50',
+                    fontSize: '18px',
```
antonjoel82: ☑ **chore:** Please a proper font size token‏
```suggestion
                    fontSize: '18px',
```

apps/web/src/components/docs/Highlight.tsx
```diff
@@ -0,0 +1,26 @@
+import hljs from 'highlight.js';
+import { useEffect, useRef } from 'react';
+import 'highlight.js/styles/default.css';
+import 'highlight.js/styles/atom-one-dark.css';
+
+export const Highlight = (props: any) => {
+  const ref: any = useRef();
```
antonjoel82: ☑ **chore:** Please remove use of `any‏`

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+              })}
+              justify="end"
+            >
+              <Button
+                onClick={onClose}
+                style={{
+                  height: '32px',
+                  padding: '0 12px',
+                }}
+              >
```
antonjoel82: ☑ **chore:** Panda please‏

apps/web/src/components/docs/DocsCard.tsx
```diff
@@ -0,0 +1,43 @@
+import { useMantineColorScheme } from '@mantine/core';
```
antonjoel82:
```suggestion
```

apps/web/src/components/docs/DocsCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const DocsCard = ({
+  title,
+  text,
+  Icon,
+  onClick,
+}: {
+  title: string;
+  text: string;
+  Icon: any;
```
antonjoel82: ☑ **chore:** ‏Please use a type other than `any`

apps/web/src/components/docs/DocsHomeModal.tsx
```diff
@@ -0,0 +1,244 @@
...
+            zIndex: 10003,
+          },
+          modal: {
+            width: 800,
+            padding: '0px !important',
+            background: 'transparent',
+            boxShadow: 'none',
+            borderRadius: 12,
+          },
+        }}
```
antonjoel82:
```suggestion

```

apps/web/src/components/docs/DocsHomeModal.tsx
```diff
@@ -0,0 +1,244 @@
+import { ActionIcon, Modal, useMantineColorScheme } from '@mantine/core';
```
antonjoel82:
```suggestion
```

apps/web/src/components/docs/DocsHomeModal.tsx
```diff
@@ -0,0 +1,244 @@
...
+import {
+  colors,
```
antonjoel82:
```suggestion
```

apps/web/src/components/docs/DocsHomeModal.tsx
```diff
@@ -0,0 +1,244 @@
...
+import { Flex, Grid, GridItem, styled } from '../../styled-system/jsx';
+import { title } from '../../styled-system/recipes';
+import { useDocsContext } from '../providers/DocsProvider';
+import { When } from '../utils/When';
+import { DocsCard } from './DocsCard';
+import { ExploreLink } from './ExploreLink';
+
+const ModalTitle = styled('h2', title);
+
+export const DocsHomeModal = ({ path, open }: { path: string; open: boolean }) => {
```
antonjoel82: ❓ **question:** Where does this Modal appear?‏

apps/web/src/components/docs/ExploreLink.tsx
```diff
@@ -0,0 +1,51 @@
...
+            lineHeight: '125',
+          })}
+        >
+          {title}
+        </Text>
+        <Text
+          className={css({
+            fontSize: '75',
+            lineHeight: '125',
+            color: 'legacy.B60',
```
antonjoel82: ☑ **chore:** ‏Please avoid using base tokens wherever possible
```suggestion
            color: 'typography.text.secondary',
```

apps/web/src/components/docs/index.tsx
antonjoel82: 🤓 **nitpick (non-blocking):** Please add a `// FIXME: ...` in the code‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+const TitleH2 = styled('h2', RTitle);
+const TitleH3 = styled('h3', RTitle);
+const TitleH1 = styled('h1', RTitle);
+
+const getMDXComponent = mdxBundler.getMDXComponent;
+
+export const Docs = ({ path = '', children }: { path?: string; children: ReactNode }) => {
+  const segment = useSegment();
+
+  const { isLoading, data: { code, title, description } = { code: '', title: '', description: '' } } = useQuery(
```
antonjoel82: ☑ **chore:** Please use explicit types‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+    const docs = document.getElementById('docs');
+
+    if (!docs) {
+      return;
+    }
+
+    const images = Array.from(docs.getElementsByTagName('img'));
+
+    for (const img of images) {
+      if (img.src.startsWith('https://mintlify.s3-us-west-1.amazonaws.com')) {
```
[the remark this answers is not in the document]
antonjoel82: ☑ **chore:** Please use an environment variable‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+      const url = new URL(img.src);
+      img.src = `https://mintlify.s3-us-west-1.amazonaws.com/novu${url.pathname}`;
+    }
+  }, [Component, isLoading]);
+
+  if (isLoading) {
+    return (
+      <Center
+        className={css({
+          marginTop: 64,
```
antonjoel82: ☑ **chore:** If none of the existing design tokens fit, and you truly need a different value, please:
- use explicit units (`rem` preferred)
- use the "escape syntax"
```suggestion
          marginTop: [4rem],
```
‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+        <Loader color={colors.error} size={32} />
+      </Center>
+    );
+  }
+
+  if (Component === null) {
+    return (
+      <Flex
+        className={css({
+          marginTop: 64,
```
antonjoel82:
```suggestion
          marginTop: [4rem],
```

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+        })}
+        justify="center"
+        align="center"
+      >
+        <Text>We could not load the documentation for you, please try again.</Text>
```
antonjoel82:
```suggestion
        <Text>We could not load the documentation. Please try again.</Text>
```

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+        components={{
+          tr: ({ className, ...props }: any) => {
+            return (
+              <tr
+                {...props}
+                className={css({
+                  height: '250',
+                  lineHeight: '125',
+                  borderBottom: `1px solid`,
+                  borderBottomColor: 'legacy.B20',
```
antonjoel82: ☑ **chore:** Design tokens please ‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+          h2: ({ className, ...props }: any) => {
+            return <TitleH2 className={css({ marginTop: '75' })} {...props} />;
+          },
+          h3: ({ className, ...props }: any) => {
+            return <TitleH3 {...props} />;
+          },
+          a: ({ className, ...props }: any) => {
+            return (
+              <a
+                href={`https://docs.novu.co${props.href}`}
```
antonjoel82: ☑ **chore:** Please use an env variable (or a string constant‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+                    value: tab.props.title,
+                  };
+                })}
+                defaultValue={tabs[0].props.title}
+              />
+            );
+          },
+          Tip: ({ className, ...props }: any) => {
+            return (
+              <Alert style={{ borderRadius: 12 }} color="gray" {...props} icon={<IconInfoOutline color="white" />} />
```
antonjoel82:
```suggestion
              <Alert className={css({ borderRadius: '75' })} color="gray" {...props} icon={<IconInfoOutline color="white" />} />
```

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+const Text = styled('p', text);
+const LiText = styled('span', text);
+const TitleH2 = styled('h2', RTitle);
+const TitleH3 = styled('h3', RTitle);
+const TitleH1 = styled('h1', RTitle);
+
+const getMDXComponent = mdxBundler.getMDXComponent;
+
+export const Docs = ({ path = '', children }: { path?: string; children: ReactNode }) => {
```
antonjoel82: ✏ **suggestion:** Please consider adding some comments and explanations to this file‏

apps/web/src/components/docs/index.tsx
```diff
@@ -0,0 +1,319 @@
+import { colors, IconInfoOutline, IconOutlineWarning, Tabs, Tooltip } from '@novu/design-system';
```
antonjoel82:
```suggestion
import { IconInfoOutline, IconOutlineWarning, Tabs, Tooltip } from '@novu/design-system';
```

apps/web/src/components/layout/components/v2/HeaderNav.tsx
```diff
@@ -40,6 +40,7 @@ export function HeaderNav() {
               <div>{themeIcon}</div>
             </Tooltip>
           </ActionIcon>
+          <DocsButton />
```
antonjoel82: 👏 **praise:** Nice job and thank you for remembering it in both places!‏

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+                minWidth: '20px',
+                border: 'none',
+              }}
+              variant="transparent"
+              onClick={onVoteClick('down')}
+            >
+              <IconThumbDownAlt color={voted == 'down' ? 'white' : undefined} size={20} />
+            </ActionIcon>
+          </Flex>
+        </Flex>
```
antonjoel82: ✏ **suggestion:** Consider separating into separate component‏

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <ActionIcon
+              style={{
+                width: '20px  !important',
+                minWidth: '20px',
+                border: 'none',
+              }}
+              variant="transparent"
+              onClick={onVoteClick('down')}
+            >
+              <IconThumbDownAlt color={voted == 'down' ? 'white' : undefined} size={20} />
```
antonjoel82:
```suggestion
              <IconThumbDownAlt color={voted === 'down' ? 'white' : undefined} size={20} />
```

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <ActionIcon
+              style={{
+                width: '20px  !important',
+                minWidth: '20px',
+                border: 'none',
+              }}
+              variant="transparent"
+              onClick={onVoteClick('up')}
+            >
+              <IconThumbUpAlt color={voted == 'up' ? 'white' : undefined} size={20} />
```
antonjoel82:
```suggestion
              <IconThumbUpAlt color={voted === 'up' ? 'white' : undefined} size={20} />
```

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+      setPath('');
+      setVoted('');
+    };
+  }, [pathname]);
+
+  const toggle = () => {
+    setOpen(!open);
+  };
+
+  const onVoteClick = (vote: string) => () => {
```
antonjoel82:
```suggestion
  const onVoteClick = (vote: 'up' | 'down') => () => {
```

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+            </ActionIcon>
+          </Flex>
+        </Docs>
+        <Flex
+          className={css({
+            marginTop: '250',
+          })}
+          align="center"
+          gap="125"
+        >
```
antonjoel82: ☑ **chore:** Please hide this if the documentation fails to load -- otherwise, we'll collect some false signals‏

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <Tooltip label="Open docs website">
+              <ActionIcon
+                style={{
+                  width: '20px  !important',
+                  minWidth: '20px !important',
+                  border: 'none',
+                }}
+                variant="transparent"
+                onClick={() => {
+                  window.open(`https://docs.novu.co/${path}`);
```
antonjoel82:
```suggestion
                  window.open(`https://docs.novu.co/${path}`, '_blank', 'noopener,noreferrer,resizable');
```
Please use env var.

antonjoel82: ❓ **question:** Should we open to a new tab or window?‏

apps/web/src/components/providers/DocsProvider.tsx
```diff
@@ -0,0 +1,192 @@
...
+            maxHeight: '100%',
+          },
+          modal: {
+            width: 800,
+            padding: '24px !important',
+            borderRadius: 12,
+            position: 'relative',
+            height: '100%',
+          },
+        }}
```
antonjoel82:
```suggestion

```

apps/web/src/components/layout/AppLayout.tsx
```diff
@@ -63,16 +64,18 @@ export function AppLayout() {
...
-                  <ContentShell>
-                    <FreeTrialBanner />
-                    {isInformationArchitectureEnabled ? (
-                      <HeaderNavNew />
-                    ) : (
-                      <HeaderNav isIntercomOpened={isIntercomOpened} />
-                    )}
-                    <Outlet />
-                  </ContentShell>
+                  <DocsProvider>
```
antonjoel82: ⚠ **issue:** I'm concerned that a context wrapper around the whole application here could be problematic as `Context` doesn't provide any memorization out of the box.‏

antonjoel82 (changes requested):
Thank you for your hard work on this, David! The loom videos were super helpful for seeing the whole masterpiece come together and I think this feature will be very helpful for users.

I do have some large-scale feedback that I'm happy to discuss further; part of it may require input from Nik and Dima.

1. As mentioned in `AppLayout`, I'm concerned that the context wrapping the whole app could have significant performance implications. I'd like to propose an alternative structure that avoids touching the `*Page.tsx` files. Given that all of our documentation is tied to a specific page, we should be able to leverage the `ROUTES` as a key for a look-up in a configuration file. This would allow us to correlate a route with a documentation path and any other static fields. Furthermore, it provides type-safety and centralizes the related content to minimize touchpoints. We could use a hook to match the current route to a config, and then render the button accordingly.
2. I strongly propose that we do not put the documentation in the header nav. I don't believe it's a typical pattern, and that bar is already fairly crowded (and may get more so if we choose to add Org and Env switchers there). Instead, I'd propose a FAB (Floating Action Button). They offer more flexibility and extensibility, feel more directly related to the content on the existing page, and seem to match common practices more. I did a quick look, and saw that Linear is using one for the same purpose.
Screenshot 2024-05-15 at 4 19 08 PM

Please let me know if you have any questions about these things or the other feedback in the PR, and again, nice work!

apps/web/src/components/providers/DocsProvider.tsx
davidsoderberg: Window.open already opens this in a new tab as seen in the video :)

apps/web/src/components/docs/DocsHomeModal.tsx
davidsoderberg: This is not used anywhere right now but will be in the next version of the docs. I implemented this because it was not communicated to me that this was v2 so to say :)

davidsoderberg:
> Thank you for your hard work on this, David! The loom videos were super helpful for seeing the whole masterpiece come together and I think this feature will be very helpful for users.
> 
> I do have some large-scale feedback that I'm happy to discuss further; part of it may require input from Nik and Dima.
> 
> 1. As mentioned in `AppLayout`, I'm concerned that the context wrapping the whole app could have significant performance implications. I'd like to propose an alternative structure that avoids touching the `*Page.tsx` files. Given that all of our documentation is tied to a specific page, we should be able to leverage the `ROUTES` as a key for a look-up in a configuration file. This would allow us to correlate a route with a documentation path and any other static fields. Furthermore, it provides type-safety and centralizes the related content to minimize touchpoints. We could use a hook to match the current route to a config, and then render the button accordingly.
> 2. I strongly propose that we do not put the documentation in the header nav. I don't believe it's a typical pattern, and that bar is already fairly crowded (and may get more so if we choose to add Org and Env switchers there). Instead, I'd propose a FAB (Floating Action Button). They offer more flexibility and extensibility, feel more directly related to the content on the existing page, and seem to match common practices more. I did a quick look, and saw that Linear is using one for the same purpose.
>    Screenshot 2024-05-15 at 4 19 08 PM
> 
> Please let me know if you have any questions about these things or the other feedback in the PR, and again, nice work!

I have now fixed according to your feedback on 1. Number 2 would I love to get @scopsy thoughts on :)

scopsy:
100% agree with @antonjoel82 on the top navigation bar point. I have already mentioned this earlier this week to nikolay here: https://www.figma.com/design/ybKRCB2ZO5DqICxjxzZGnW?node-id=701-74044#804551101

I would actually go further and say, that the inline documentation should be, well inline 😄 When it's located at the top navbar I would expect it will just open the global docs page for me, as the top navbar is the "global" scope for me personally. 

Here are a couple of potentially solutions we can iterate over:
- FAB Pattern (Joels suggestion) - similar to how Linear/Figma is handling it, and we can connect all relevant things there: Support, Docs, Changelog, Feedback, etc...
- Have inline Icons close to their origin, for example: Near the "Workflows" title, this pattern can allow us having "multiple" inline docs withing specific screens and more complex areas like the workflow editor. (Resend did it nicely in their dashboard)

Putting this note assigned, I really like how the modal itself and the docs look like! 😍

rgozdzialski:
@scopsy @antonjoel82 @davidsoderberg Since we're only testing the functionality --- although the FAB idea is cool and I agree in the long run it would give us more flexibility as a small context menu --- for version one, I'd go with the "icon next to the page title" approach that @scopsy suggested. Thanks!

davidsoderberg:
It is now moved to the page title instead 🥳 

Screenshot 2024-05-16 at 09 05 38

.source
antonjoel82, line 1: ❓ **question:** Is this intentional? I don't see any files that warrant updating this‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,333 @@
...
+const TitleH3 = styled('h3', RTitle);
+const TitleH1 = styled('h1', RTitle);
+
+const getMDXComponent = mdxBundler.getMDXComponent;
+
+/*
+ *Render the mdx for our mintlify docs inside of the web.
+ *Fetching the compiled mdx from another service and then try to map the markdown to react components.
+ */
+export const Docs = ({ path = '', children, actions }: { path?: string; children: ReactNode; actions: ReactNode }) => {
```
antonjoel82:
```suggestion
export const Docs = ({ path = '', children, actions }: PropsWithChildren<{ path?: string; actions: ReactNode }>) => {
```

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,333 @@
...
+      </Flex>
+    );
+  }
+
+  return (
+    <>
+      <VStack
+        alignItems="unset"
+        id="docs"
+        gap={12}
```
antonjoel82:
```suggestion
        gap={'75'}
```

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,333 @@
...
+            },
+            thead: ({ className, ...props }: any) => {
+              return (
+                <thead
+                  {...props}
+                  className={css({
+                    height: '150',
+                    lineHeight: '125',
+                    color: 'typography.text.main',
+                    borderBottomColor: 'legacy.B30',
```
antonjoel82: ❓ **question:** Can you please talk to Nikolay about these two colors and seeing if we can use existing ones? It is possible to add them to the `semanticColors.tokens.ts` if they are consistently used elsewhere!‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,333 @@
...
+              return (
+                <div {...props}>
+                  <img alt="" src={`${MINTLIFY_IMAGE_URL}${props.children.props.src}`} />
+                  <Text className={css({ textAlign: 'center', fontStyle: 'italic' })}>{props.caption}</Text>
+                </div>
+              );
+            },
+            Info: ({ className, ...props }: any) => {
+              return (
+                <Alert style={{ borderRadius: 12 }} color="gray" {...props} icon={<IconInfoOutline color="white" />} />
```
antonjoel82:
```suggestion
                <Alert className={css({ borderRadius: '150' })}  {...props} icon={<IconInfoOutline />} />
```
Please add "color" a different way

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,333 @@
...
+                    {props.title}
+                  </LiText>
+                  <LiText className={css({ lineHeight: '125' })}>{props.children}</LiText>
+                </li>
+              );
+            },
+            Warning: ({ className, ...props }: any) => {
+              return (
+                <Alert
+                  style={{ borderRadius: 12 }}
```
antonjoel82:
```suggestion
                 className={css({ borderRadius: '150' })}
```

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,333 @@
...
+                  color="yellow"
+                  {...props}
+                  icon={<IconOutlineWarning color="white" />}
+                />
+              );
+            },
+            Note: ({ className, ...props }: any) => {
+              return (
+                <Alert style={{ borderRadius: 12 }} color="blue" {...props} icon={<IconInfoOutline color="white" />} />
```
antonjoel82:
```suggestion
                <Alert className={css({ borderRadius: '150' })} color="blue" {...props} icon={<IconInfoOutline color="white" />} />
```

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const [docsOpen, setDocsOpen] = useState<boolean>(false);
+  const { pathname } = useLocation();
+
+  useEffect(() => {
+    for (const route in PATHS) {
+      if (matchPath(route, pathname) !== null) {
+        setPath(PATHS[route]);
+        break;
+      }
+    }
```
antonjoel82, line 65: 🤓 **nitpick (non-blocking):** Prefer a more functional approach (array functions) to avoid using `break`‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    return () => {
+      setPath('');
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [pathname]);
+
+  const toggle = () => {
+    setDocsOpen(!docsOpen);
```
antonjoel82:
```suggestion
    setDocsOpen((prevOpen) => !prevOpen);
```

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+            arrowOffset={16}
+            position="bottom-end"
+            withArrow
+          >
+            <Popover.Target>
+              <ActionButton
+                sx={{
+                  height: 24,
+                  minHeight: 24,
+                }}
```
antonjoel82: ☑ **chore:** Please ensure we remove all traces of `sx` and `style` props from the PR‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+                Icon={() => <IconOutlineMenuBook />}
+                onClick={() => toggle()}
+              />
+            </Popover.Target>
+            <Popover.Dropdown
+              className={css({
+                borderRadius: '75',
+                boxShadow: isDark ? undefined : 'dark',
+                background: isDark ? 'legacy.B30' : undefined,
```
antonjoel82: ☑ **chore:** Please replace conditionals within `css` like this with I use of `cva` ([example](https://github.com/novuhq/novu/blob/1965cbe1a8c1b3ee8fc1db58c831f34ae56fe791/apps/web/src/pages/settings/WebhookPage/WebhookClaimStatusDisplay.tsx#L35)‏) as we need to help panda generate static css

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+                      marginBottom: '50',
+                      fontSize: '88',
+                      lineHeight: '150',
+                    })}
+                  >
+                    Discover inline documentation
+                  </Title>
+                  <Text
+                    className={css({
+                      color: isDark ? 'legacy.B80' : 'typography.text.secondary',
```
antonjoel82: 💭 **thought (non-blocking):** ‏This is the sign of something deviating from our styles since `secondary` should handle dark and light mode‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+                  })}
+                >
+                  Got it
+                </Button>
+              </Flex>
+            </Popover.Dropdown>
+          </Popover>
+        </div>
+      </Tooltip>
+      <DocsModal open={docsOpen} toggle={toggle} path={path} />
```
antonjoel82, line 161:
```suggestion
      {/* TODO: extract the Modal root out when modal management is improved */}
      <DocsModal open={docsOpen} toggle={toggle} path={path} />
```

apps/web/src/components/docs/DocsModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { useSegment } from '@novu/shared-web';
+import { useEffect, useState } from 'react';
+import { css } from '../../styled-system/css';
+import { Flex } from '../../styled-system/jsx';
+import { Docs } from './Docs';
+import { DOCS_URL } from './docs.const';
+import { VotingWidget } from './VotingWidget';
+
+export const DocsModal = ({ open, toggle, path }) => {
+  const [voted, setVoted] = useState<'up' | 'down' | ''>('');
```
antonjoel82:
```suggestion
  const [voted, setVoted] = useState<'up' | 'down' | undefined>();
```
(or null)

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { css } from '../../styled-system/css';
+import { Flex, styled } from '../../styled-system/jsx';
+import { text, title } from '../../styled-system/recipes';
+import { PATHS } from './docs.const';
+import { DocsModal } from './DocsModal';
+
+const Title = styled('h3', title);
+const Text = styled('p', text);
+
+export const DocsButton = () => {
```
antonjoel82, line 50: 💭 **thought (non-blocking):** I do think this button would be better as a "renderer" that receives almost all of its behavior from props. This would make it more testable (which I think is important for something that will be as widely used as this) and will clean up the rendering vs logic to make it easier to read.‏

apps/web/src/components/docs/DocsModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <Docs
+        path={path}
+        actions={
+          <Flex
+            className={css({
+              position: 'fixed',
+              top: '150',
+              right: '150',
+              background: isDark ? 'legacy.B15' : 'white',
+              zIndex: 1,
```
antonjoel82: ‏
```suggestion
              // TODO: update with a proper zIndex once we have refactored
              zIndex: [1],
```

apps/web/src/components/docs/DocsModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+              padding: '25',
+              borderBottomLeftRadius: '50',
+            })}
+            gap="75"
+          >
+            <ActionButton
+              tooltip="Open docs website"
+              onClick={() => {
+                window.open(`${DOCS_URL}/${path}`);
```
antonjoel82: ☑ **chore:** Please add the same options here as the other `window.open` calls. It may be worthwhile to extract them to a shared function‏

apps/web/src/components/docs/DocsModal.tsx
```diff
@@ -0,0 +1,99 @@
...
+        },
+        modal: {
+          width: 800,
+          padding: '1.5rem !important',
+          borderRadius: '0.75rem',
+          position: 'relative',
+          height: '100%',
+          boxShadow: 'none',
+        },
+      }}
```
antonjoel82, line 59: 🗒 **note (non-blocking):** ‏‏We discussed that this was acceptable for now, but not ideal. In the meantime, please add a TODO

apps/web/src/components/docs/VotingWidget.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Flex, styled } from '../../styled-system/jsx';
+import { text } from '../../styled-system/recipes';
+
+const Text = styled('p', text);
+
+export const VotingWidget = ({
+  voted,
+  onVoteClick,
+}: {
+  voted: 'up' | 'down' | '';
```
antonjoel82: ☑ **chore:** Please extract ``up' | 'down'` to a shared type for this file an others in this PR‏

apps/web/src/components/docs/docs.const.ts
```diff
@@ -0,0 +1,19 @@
...
+  [ROUTES.WORKFLOWS]: 'workflows/notification-workflows',
+  [ROUTES.TENANTS]: 'tenants/introduction',
+  [ROUTES.TRANSLATIONS]: 'content-creation-design/translations',
+};
+
+export const DOCS_URL = 'https://docs.novu.co';
+
+export const MINTLIFY_IMAGE_URL = 'https://mintlify.s3-us-west-1.amazonaws.com/novu';
+
+export const MDX_URL = 'https://cloud-doc.vercel.app/';
```
antonjoel82, line 19: ✏ **suggestion:** [Non-blocking] - I do believe these are best as environment variables‏

apps/web/src/components/layout/components/v2/HeaderNav.tsx
```diff
@@ -1,6 +1,5 @@
 import { ActionIcon, Group, Header } from '@mantine/core';
-
-import { colors, IconHelpOutline, Tooltip } from '@novu/design-system';
+import { IconHelpOutline, Tooltip } from '@novu/design-system';
```
antonjoel82: 👏 **praise:** Thank you for the clean-up!‏

antonjoel82 (commented):
Awesome work on all the changes, and thanks for walking me through them! There's still a bit of clean-up to be done to help with consistency and to avoid accruing more tech / design debt.

.source
davidsoderberg: This happens because next was not aligned with next enterprise as it should and when I merge next into my branch I get this file change

apps/web/src/components/docs/Docs.tsx
davidsoderberg: Yes as I said in our walkthrough I will ask Nikolay about this and fix it in a follow up pr :)

apps/web/src/components/docs/Docs.tsx
```diff
@@ -157,7 +157,17 @@ export const Docs = ({ path = '', children, actions }: { path?: string; children
             },
             Info: ({ className, ...props }: any) => {
               return (
-                <Alert style={{ borderRadius: 12 }} color="gray" {...props} icon={<IconInfoOutline color="white" />} />
+                <Alert
+                  className={css({
+                    borderRadius: '75',
+                    backgroundColor: 'mauve.60.dark !important',
```
davidsoderberg, line 165: All `!important` I have added in latest commit was needed to affect style of their components

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,363 @@
...
+    return getMDXComponent(code);
+  }, [code]);
+
+  // Workaround for img tags that is not parsed correctly by mdx-bundler
+  useEffect(() => {
+    if (Component === null || isLoading) {
+      return;
+    }
+
+    const docs = document.getElementById('docs');
```
rifont: ⚠ **issue:** I wonder if we should have a more specific `id` here, `docs` could be a little loose and result in multiple elements being targeted. I'd suggest something like `embedded-docs`

Furthermore, let's extract this out to a constant.‏

apps/web/src/components/docs/docs.const.ts
rifont, line 1: 👏 **praise:** Great extraction of constants! Make's future additions very easy.‏

rifont, line 1: ☑ **chore:** Rename to `docs.constants.ts` to maintain consistency with other constant file names.‏

pnpm-lock.yaml
rifont, line 1: ☑ **chore:** There's a huge number of changes on this file given only 2 dependencies have been added. Are all these changes relevant?‏

apps/web/src/components/docs/docs.const.ts
```diff
@@ -0,0 +1,19 @@
+import { ROUTES } from '@novu/shared-web';
+
+export const PATHS: { [key in ROUTES]?: string } = {
+  [ROUTES.ACTIVITIES]: 'activity-feed/introduction',
+  [`${ROUTES.BRAND}/layouts`]: 'content-creation-design/layouts',
```
rifont, line 5: ❓ **question:** This manual template literal here is a bit of a code smell. Can we instead add the `ROUTES.LAYOUT` route into the `ROUTES` constant? After we make this change, we should be able to simplify the typing of `PATHS` to remove the `?` symbol.‏

apps/web/src/components/docs/docs.const.ts
davidsoderberg: I agree with you that we should keep consistency and in the web project .const have been used so that is the reason for this.

apps/web/src/components/docs/docs.const.ts
davidsoderberg: this will only live until the information architecture project is released and all the if cases are gone and this one can be removed.

pnpm-lock.yaml
davidsoderberg: Good catch, how can I resolve this problem, remove lock file and re run pnpm install?

apps/web/src/components/docs/docs.const.ts
rifont: Ok got it, thanks for clarifying!

pnpm-lock.yaml
rifont: I'm encountering a similar issue and will post into the team channel shortly.

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+const Text = styled('p', text);
+const LiText = styled('span', text);
+const TitleH2 = styled('h2', RTitle);
+const TitleH3 = styled('h3', RTitle);
+const TitleH1 = styled('h1', RTitle);
+
+const getMDXComponent = mdxBundler.getMDXComponent;
+
+const wrapperId = 'embedded-docs';
```
antonjoel82:
```suggestion
const DOCS_WRAPPER_ELEMENT_ID = 'embedded-docs';
```

apps/web/src/components/docs/Docs.tsx
antonjoel82: (Sorry, just adding these in writing so that they are visible to others who may come along)

apps/web/src/components/docs/Docs.tsx
antonjoel82: 🤓 **nitpick (non-blocking):** While using the "new" base color tokens is better than using the `legacy.*` tokens, we should still be aiming to use semantic tokens as much as possible in application code so that we can better handle dark/light and in a semantic way.‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -0,0 +1,365 @@
...
+              );
+            },
+            Info: ({ className, ...props }: any) => {
+              return (
+                <Alert
+                  className={css({
+                    borderRadius: '75',
+                    backgroundColor: 'mauve.60.dark !important',
+                    '& p': {
+                      color: 'white !important',
```
antonjoel82: ☑ **chore:** Replace with a valid token‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { Flex, styled } from '../../styled-system/jsx';
+import { text, title } from '../../styled-system/recipes';
+import { SystemStyleObject } from '../../styled-system/types';
+import { PATHS } from './docs.const';
+import { DocsModal } from './DocsModal';
+
+const Title = styled('h3', title);
+const Text = styled('p', text);
+
+const popoverDropdownRecipe = cva<{ colorScheme: Record<ColorScheme, SystemStyleObject> }>({
```
antonjoel82, line 16: 👏 **praise:** Nice job!‏

apps/web/src/components/docs/DocsButton.tsx
```diff
@@ -0,0 +1,164 @@
...
+    },
+  },
+});
+
+export const DocsButton = () => {
+  const { colorScheme } = useColorScheme();
+  const [opened, setOpened] = useState<boolean>(false);
+  const segment = useSegment();
+  const [path, setPath] = useState<string>('');
+  const enabled = useMemo(() => path.length > 0, [path]);
```
antonjoel82:
```suggestion
  const shouldShowButton = useMemo(() => path.length > 0, [path]);
```

apps/web/src/components/docs/VotingWidget.tsx
```diff
@@ -0,0 +1,38 @@
...
+    >
+      <Text>Did you find it useful?</Text>
+      <Flex gap="50" align="center">
+        <ActionButton
+          onClick={onVoteClick('up')}
+          Icon={() => <IconThumbUpAlt color={voted === 'up' ? 'white' : undefined} size={20} />}
+        />
+        <ActionButton
+          onClick={onVoteClick('down')}
+          Icon={() => <IconThumbDownAlt color={voted === 'down' ? 'white' : undefined} size={20} />}
```
antonjoel82: ☑ **chore:** Use `cva` here please‏

antonjoel82 (approved):
Thanks for great improvements on this! My final recommendation would be that this should be behind a feature flag to make it easier to control. Fortunately, it should be fairly straight-forward since we only really have one "entry point."

antonjoel82:
@davidsoderberg I rebased your PR after merging the new Design System package and updated the imports for Panda so that everything should just work the same!

apps/web/src/components/docs/DocsModal.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <ActionButton
+              onClick={() => {
+                toggle();
+              }}
+              Icon={() => <IconOutlineClose />}
+            />
+          </Flex>
+        }
+      >
+        <VotingWidget onVoteClick={onVoteClick} voted={voted} />
```
antonjoel82, line 96: ✏ **suggestion:** I do still strongly believe we should hide this when docs fail to load, otherwise we will collect unhelpful data‏

davidsoderberg: It will be hidden because if Component is null it will not render the voting widget inside of the docs component.

PR #5517, opened 2024-05-07
[NV-3735]: Improve launch darkly

antonjoel82:

- Use async LaunchDarkly initialization, but ensure it works with self-hosted
- **Why?** Conditional routing (enabling routes based on feature flags) for Information Architecture is inconsistent, and causes issues with certain routes
- Resolves [NV-3735](https://linear.app/novu/issue/NV-3735/refactor-launchdarkly-implementation-to-better-support-conditional)

- [ ] Did we account properly for self-hosted?

### Repro

**Basic**
1. Check-out the branch `improve-launch-darkly`
1. Start everything normally -- upon loading, you should briefly see a reddish loader in the center of the page
1. Open a new tab and navigate to http://localhost:4200/settings/profile?view=password
1. You should see the loader again, but will go directly the the User Profile with the sidebar open

**Old route hidden behind feature flag**
1. Navigate to http://localhost:4200/brand
2. The loader should appear, and you should be directed to `Workflows`

**Check non-auth route**
1. Log-out
2. Everything should load correctly.
4. Log-in
5. The loader should appear and the app should then load

**Self-hosted**
1. In your local `web/.env.local` file (or whichever one you're using for local dev), comment out `REACT_APP_LAUNCH_DARKLY_CLIENT_SIDE_ID`
2. Start the app
3. Open the browser tools to the `Network` tab
4. The app should load correctly without the loader, but everything will be based on env variables
6. Filter network requests with `launchd` and nothing should appear


apps/web/src/AppRoutes.tsx
```diff
@@ -116,10 +117,12 @@ export const AppRoutes = () => {
         <Route path={ROUTES.TEAM} element={<MembersInvitePage />} />
         <Route path={ROUTES.CHANGES} element={<PromoteChangesPage />} />
         <Route path={ROUTES.SUBSCRIBERS} element={<SubscribersList />} />
-        <Route path={ROUTES.BRAND} element={<BrandPage />}>
-          <Route path="" element={<BrandingForm />} />
-          <Route path="layouts" element={<LayoutsListPage />} />
-        </Route>
+        {!isInformationArchitectureEnabled && (
```
antonjoel82, line 120: 🗒 **note (non-blocking):** Fix for information architecture -- previously didn't work due to LD‏

apps/web/src/Providers.tsx
```diff
@@ -22,28 +24,41 @@ const queryClient = new QueryClient({
...
+            </HelmetProvider>
+          </LaunchDarklyProvider>
+        </AuthProvider>
+      </QueryClientProvider>
     </SegmentProvider>
   );
 };
 
-export default Sentry.withProfiler(
-  withLDProvider({
```
antonjoel82, line 43: 🗒 **note (non-blocking):** Moved LD provider to its own file in `shared-web` with logic‏

libs/shared-web/src/constants/unprotected-routes.const.ts
```diff
@@ -0,0 +1,12 @@
+import { ROUTES } from './routes.enum';
+
+export const UNPROTECTED_ROUTES: ROUTES[] = [
+  ROUTES.AUTH_SIGNUP,
+  ROUTES.AUTH_LOGIN,
+  ROUTES.AUTH_RESET_REQUEST,
+  ROUTES.AUTH_RESET_TOKEN,
+  ROUTES.AUTH_INVITATION_TOKEN,
+  ROUTES.AUTH_APPLICATION,
+];
```
antonjoel82, line 10: ❓ question: Are there any other routes that belong here?‏

libs/shared-web/src/hooks/useFeatureFlags.ts
```diff
@@ -24,10 +24,12 @@ export const useFeatureFlags = (organization: IOrganizationEntity) => {
 };
 
 export const useFeatureFlag = (key: FeatureFlagsKeysEnum): boolean => {
-  const { [key]: featureFlag } = useFlags();
+  /** We knowingly break the rule of hooks here to avoid making any LaunchDarkly calls when it is disabled */
+  // eslint-disable-next-line
+  const flagValue = checkShouldUseLaunchDarkly() ? useFlags()[key] : undefined;
```
antonjoel82, line 29: 🗒 **note (non-blocking):** Breaks the rule of hooks, but should be okay since it's done via a static check.‏

libs/shared-web/src/hooks/useFeatureFlags.ts
```diff
@@ -1,15 +1,15 @@
...
 
 import { FEATURE_FLAGS } from '../config';
 
-export const useFeatureFlags = (organization: IOrganizationEntity) => {
+export const useFeatureFlags = (organization?: IOrganizationEntity) => {
   const ldClient = useLDClient();
 
   useEffect(() => {
-    if (!organization?._id) {
+    if (!checkShouldUseLaunchDarkly() || !organization?._id) {
```
antonjoel82, line 12: 🗒 **note (non-blocking):** The client should never be defined anyway, but just as an extra fail-safe‏

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const authContext = useAuthContext();
+  if (!authContext) {
+    throw new Error('LaunchDarklyProvider must be used within <AuthProvider>!');
+  }
+  const { currentOrganization } = authContext;
+
+  useEffect(() => {
+    // no need to fetch if LD is disabled or there isn't an org to query against
+    if (!checkShouldUseLaunchDarkly() || !currentOrganization) {
```
antonjoel82, line 40: 🗒 **note (non-blocking):** Avoids any LD calls for initializing the provider‏

libs/shared-web/src/constants/unprotected-routes.const.ts
scopsy: Is there a chance if somehow doing this without the need to define unprotected routes here? What is the current problem we are solving with this list? (wondering if there is another way around it)

antonjoel82: @scopsy Refactored! This was due to the initial constraints with the ordering of the providers -- LD was wrapped around the auth provider, so we didn't necessarily have the context available.

Since I switched the order of the tree in this PR, I updated it to use the logged-in status!

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+      return;
+    }
+
+    const fetchLDProvider = async () => {
+      LDProvider.current = await asyncWithLDProvider({
+        clientSideID: LAUNCH_DARKLY_CLIENT_SIDE_ID,
+        context: {
+          kind: 'organization',
+          key: currentOrganization._id,
+          name: currentOrganization.name,
```
scopsy, line 50: **question:** If I understand this correctly, this will prevent us from having feature flags on the Onboarding and Signup pages right?

What if we still load launchdarkly in those pages but without the "context" ?

‏

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+        },
+        reactOptions: {
+          useCamelCaseFlagKeys: false,
+        },
+      });
+      setIsLDReady(true);
+    };
+
+    fetchLDProvider();
```
LetItRock, line 59: ⚠ **issue:** if this promise fails, then it will load forever

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+ * Async provider for feature flags.
+ *
+ * @requires AuthProvider must be wrapped in the AuthProvider.
+ */
+export const LaunchDarklyProvider: React.FC<PropsWithChildren<ILaunchDarklyProviderProps>> = ({
+  children,
+  fallbackDisplay,
+}) => {
+  const LDProvider = useRef<GenericLDProvider>(DEFAULT_GENERIC_PROVIDER);
+  const [isLDReady, setIsLDReady] = useState<boolean>(false);
```
LetItRock, line 30: 🤓 **nitpick (non-blocking):** the same could be done with just `useState` ;)

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
```diff
@@ -0,0 +1,84 @@
...
+};
+
+/**
+ * Refreshes feature flags on org change using the LaunchDarkly client from the provider.
+ */
+function LaunchDarklyClientWrapper({ children, org }: PropsWithChildren<{ org?: IOrganizationEntity }>) {
+  useFeatureFlags(org);
+
+  return <>{children}</>;
+}
```
LetItRock, line 84: 👏 **praise:** nice catch, thank you! 🙏

apps/web/src/AppRoutes.tsx
```diff
@@ -49,6 +49,7 @@ import { useSettingsRoutes } from './SettingsRoutes';
 
 export const AppRoutes = () => {
```
LetItRock, line 50: 🗒 **note (non-blocking):** the `/auth/application` is not a protected route 😨

LetItRock (commented):
⚠ **issue:** blockers: sign up process and reset password are not working for me

https://github.com/novuhq/novu/assets/2607232/679445de-64c3-4fbc-b396-4e6551eaae98

Please make sure to run all e2e tests with this PR.

antonjoel82:
> ⚠ **issue:** blockers: sign up process and reset password are not working for me
> 
>  Screen.Recording.2024-05-08.at.00.29.23.mov 
> Please make sure to run all e2e tests with this PR.

Great catch, thanks! This whole non-automated automated tests is beating me 🙃

libs/shared-web/src/hooks/useAuthController.ts
```diff
@@ -122,7 +120,8 @@ export function useAuthController() {
   const logout = () => {
     setTokenCallback(null);
     queryClient.clear();
-    navigate('/auth/login');
+    // avoid usage of react-router here to prevent needing AuthProvider to be wrapped in the BrowserRouter
+    window.location.assign('/auth/login');
```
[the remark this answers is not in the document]
antonjoel82: It worked through my testing, but did you encounter anything specific?

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
antonjoel82: I'll investigate further. It definitely introduces some complexity, since they require fundamentally different things. Additionally, we're not handling this [properly](https://docs.launchdarkly.com/sdk/features/anonymous#javascript) at the moment

libs/shared-web/src/hooks/useAuthController.ts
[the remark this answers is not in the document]
LetItRock: nope, I was just curious why we needed that change apart from that we won't need to wrap `AuthProvider` with `BrowserRouter`

apps/web/src/Providers.tsx
antonjoel82: Most of our other similar providers are there, and many of the related behaviors / imports (i.e. `useFeatureFlags`) are there, so it seemed best IMO

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
antonjoel82: My thinking was to avoid passing around the provider when a boolean would suffice -- especially now that we're using some "selectors"

libs/shared-web/src/utils/auth-selectors/selectHasUserCompletedSignUp.ts
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Determine if a user is fully-registered; if not, they're still in onboarding.
+ */
+export const selectHasUserCompletedSignUp = (userCtx: UserContext): boolean => {
+  if (!userCtx) {
+    return false;
+  }
+
+  // User has completed registration if they have an associated orgId.
+  return !!userCtx.jwtPayload?.organizationId;
```
antonjoel82, line 12: 🗒 **note (non-blocking):** This does not seem like the best way, but it is the most robust and semantic way I found thus far‏

libs/shared-web/src/utils/auth-selectors/selectShouldInitializeLaunchDarkly.tsx
antonjoel82, line 1: I wanted to add tests to these files, but we do not have testing set up yet in `shared-web` 😞 

https://linear.app/novu/issue/NV-3773/setup-tests-in-shared-web-and-implement-tests-for-launchdarkly

apps/web/src/Providers.tsx
LetItRock: 🤓 **nitpick (non-blocking):** I suggest keeping it closer to the source of the web app unless we are sure it would be reused. Otherwise, we might end up with a landfill.

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
```diff
@@ -0,0 +1,100 @@
...
+                /**
+                 * When user is not authenticated, assigns an id to them to ensure consistent results.
+                 * https://docs.launchdarkly.com/sdk/features/anonymous#javascript
+                 */
+                kind: 'user',
+                anonymous: true,
+              },
+        });
+      } catch (err: unknown) {
+        // FIXME: what should we do here since we don't have logging?
```
LetItRock: send the error to Sentry like this?
```
Sentry.captureException(err);
```
but I guess this will require moving the Sentry provider up in the tree

LetItRock (commented):
please take a look at the failing tests: https://github.com/novuhq/novu/actions/runs/9014235363/job/24766606546

apps/web/src/Providers.tsx
SokratisVidros: I agree with @LetItRock. I feel that we might be overusing shared packages on some occasions. Other than that, great choice to split it into a separate component.

libs/shared-web/src/hooks/useAuthController.ts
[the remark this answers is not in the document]
SokratisVidros: Should we use `window.location.replace` to highjack the back button? That shouldn't work anyway, as the user just signed out.

libs/shared-web/src/utils/auth-selectors/selectShouldInitializeLaunchDarkly.tsx
```diff
@@ -0,0 +1,25 @@
...
+/** Determine if LaunchDarkly should be initialized based on the current auth context */
+export function selectShouldInitializeLaunchDarkly(userCtx: UserContext): boolean {
+  const { isLoggedIn, currentOrganization } = userCtx;
+  // don't show fallback if LaunchDarkly isn't enabled
+  if (!checkShouldUseLaunchDarkly()) {
+    return false;
+  }
+
+  // enable feature flags for unauthenticated areas of the app
+  if (!isLoggedIn) {
```
SokratisVidros: Do we have to stick to this restriction for now? There might be a case in the future where we want to feature flag public pages such as sign-in or sign-up.

libs/shared-web/src/utils/auth-selectors/selectShouldInitializeLaunchDarkly.tsx
```diff
@@ -0,0 +1,25 @@
...
+    return true;
+  }
+
+  // allow LD to load when the user is created but still in onboarding
+  if (!selectHasUserCompletedSignUp(userCtx)) {
```
SokratisVidros: Can you please elaborate on why this conditional is required?

libs/shared-web/src/utils/auth-selectors/selectShouldInitializeLaunchDarkly.tsx
antonjoel82: This check is what prevents it from blocking on public pages!

libs/shared-web/src/utils/auth-selectors/selectShouldInitializeLaunchDarkly.tsx
antonjoel82: I'll add more to the code comment as well, but after users provide their name, email, and password, we take them to the `OnboardingForm` where they provide details such as job title, use cases, company name, etc. When they reach this page, `isLoggedIn` is `true`, but they don't have an `organizationId` yet that we can use for org-based feature flags. This was a tricky thing since we otherwise rely on orgId when the user is logged in. We need to make this check so as not to cause it to load endlessly, but still initialize LD in the "anonymous" context

libs/shared-web/src/hooks/useAuthController.ts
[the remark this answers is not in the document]
antonjoel82: We didn't previously have `replace`, so I kept it the same to avoid regressions, but will add it.

libs/shared-web/src/providers/LaunchDarklyProvider.tsx
antonjoel82: Added, thanks! Sentry should is the outermost provider, so I think we're good!

apps/web/src/Providers.tsx
antonjoel82: Sounds good, thanks for the feedback! Updated

libs/shared-web/src/utils/auth-selectors/selectShouldInitializeLaunchDarkly.tsx
SokratisVidros: Sounds good. Thanks.

SokratisVidros (approved):

antonjoel82:
Unfortunately, I am encountering some flakiness with tests, but I have verified manually that the behavior works. After talking with Sok this morning, we don't believe it's worth blocking on them since we'll be replacing them soon. Planning to merge this tomorrow morning after deployment breakfast

PR #5523, opened 2024-05-08
fix: sentry error options.inverse is not a function

ainouzgali:

Fix for sentry errors that are coming from an incorrect use of our handlebar helpers that are expected to be in the syntax of if - else. Like equal, gt, lt etc.
There seems to be a misunderstanding with how people use it vs how we planned people will use it.
We treat it as a condition that encapsulates the if condition and its structure, while people use it under the assumption it will return a boolean value.
For example:
a user does this:
`{{#if (gte step.total_count 3)}}true{{else}}false{{/if}}`
but the correct use would be:
`{{#gte step.total_count 3}}true{{else}}false{{/gte}}`

In this PR, we change it to allow both uses, we check if the user gave a function (the if else block content) and if he didn't then return a boolean value.

### Screenshots

<details>

</details>


scopsy (approved):
Bravo!

jainpawan21 (approved):
🏆

PR #5527, opened 2024-05-08
feat: catch errors of wrong use of handlebars

ainouzgali:

For compile email template, there was no catch block for wrong use of handlebars syntax which resulted in
1.  A lot of sentry errors that only creates clutter 
2. Uninformative message for the  user (showed Internal server error)

I added the try-catch on the actual template compilation - so we won't miss adding it in each place we are using that usecase.

Sentry links:
[1](https://novu-r9.sentry.io/issues/4380720667/), [2](https://novu-r9.sentry.io/issues/4765198103/), [3](https://novu-r9.sentry.io/issues/4109348362/), [4](https://novu-r9.sentry.io/issues/4837740604/), [5](https://novu-r9.sentry.io/issues/5310239789/), [6](https://novu-r9.sentry.io/issues/5310189560/)
### Screenshots

<details>

</details>

### DIFF
--- a/packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
+++ b/packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
@@ -5,6 +5,7 @@ import { HandlebarHelpersEnum } from '@novu/shared';
 
 import { CompileTemplateCommand } from './compile-template.command';
 import * as i18next from 'i18next';
+import { ApiException } from '../../utils/exceptions';
 
 const assertResult = (condition: boolean, options) => {
   const fn = condition ? options.fn : options.inverse;
@@ -208,10 +209,16 @@ Handlebars.registerHelper(
 export class CompileTemplate {
   async execute(command: CompileTemplateCommand): Promise<string> {
     const templateContent = command.template;
+    let result = '';
+    try {
+      const template = Handlebars.compile(templateContent);
 
-    const template = Handlebars.compile(templateContent);
-
-    const result = template(command.data, {});
+      result = template(command.data, {});
+    } catch (e: any) {
+      throw new ApiException(
+        e?.message || `Message content could not be generated`
+      );
+    }
 
     return result.replace(/&#x27;/g, "'");
   }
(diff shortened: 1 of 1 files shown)


BiswaViraj (approved):

packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
```diff
@@ -208,10 +209,16 @@ Handlebars.registerHelper(
...
-
-    const template = Handlebars.compile(templateContent);
-
-    const result = template(command.data, {});
+    let result = '';
+    try {
+      const template = Handlebars.compile(templateContent);
+
+      result = template(command.data, {});
+    } catch (e: any) {
```
BiswaViraj:
```suggestion
    } catch (e: unknown) {
```
Better TS ;)

BiswaViraj (dismissed):

packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
ainouzgali: Unfortunately, it creates a ts error for e.message 😬

ainouzgali: At the moment, thats how all catch errors are typed in the project

antonjoel82: @ainouzgali You can use `checkIsResponseError` as a type-guard here! TypeScript began recommending `unknown` as the type for `catch` in TS 4.4

antonjoel82: @BiswaViraj Side note: I would suggest that we reserve `Request changes` for blocking problems (like `issue` or important `chore`s) 🙂 -- this is probably more of a nitpick until we put it into a coding style guide

(But agree with commenting on it so we have fewer places to change later 🙌 )

antonjoel82 (approved):
🙌

packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
ainouzgali: Cool. Didn't know about it. Thanks!

PR #5529, opened 2024-05-08
guard does not fit handlebar error, revert to show informative error message

ainouzgali:

 The guard `checkIsResponseError` does not satisfy the error returned by the handlebar compile. So, the message shown to user is general, and not specific to their action. 
 
Context [here](https://github.com/novuhq/novu/pull/5527#discussion_r1594298599)
### Screenshots

<details>

</details>

### DIFF
--- a/packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
+++ b/packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@nestjs/common';
 import * as Handlebars from 'handlebars';
 import { format } from 'date-fns';
-import { checkIsResponseError, HandlebarHelpersEnum } from '@novu/shared';
+import { HandlebarHelpersEnum } from '@novu/shared';
 
 import { CompileTemplateCommand } from './compile-template.command';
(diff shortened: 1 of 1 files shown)


packages/application-generic/src/usecases/compile-template/compile-template.usecase.ts
```diff
@@ -214,12 +214,10 @@ export class CompileTemplate {
       const template = Handlebars.compile(templateContent);
 
       result = template(command.data, {});
-    } catch (e: unknown) {
-      let errorMessage = `Message content could not be generated`;
```
ainouzgali, line 218: @antonjoel82 it always shows this message 😔 Which is a shame, as the message returned from the error is quite informative. 

This error structure does not stand to the conditions of this guard

ainouzgali: Lets tackle this TS - error - any/unknown thing at a different time, when we can give it more attention in an organized way. wdyt?

antonjoel82: Yep, sounds good for sure!

antonjoel82 (approved):

PR #5531, opened 2024-05-09
fix: auto-scroll on provider sidebar [NV-3745]

ChmaraX:
* clicking on the provider tile didn't scroll to the selected section in the sidebar
* coming directly from the URL didn't scroll to the section
* in some cases  `scrollTo` was not able to find element in the DOM because it didn't yet exist, since the providers are rendered asynchronously
* `scrollTo` query param was not updating

### Screenshots

**Before:**

https://github.com/novuhq/novu/assets/32128755/ad5955cf-51bd-4cd1-8af3-21a9612575a1

**After:**

https://github.com/novuhq/novu/assets/32128755/7a8ab852-9630-4238-ab74-9acc5771edb4


scopsy (approved):
🚀 Fantastic!

PR #5532, opened 2024-05-09
Fixes for scheduled e2e

SokratisVidros:

1. Scheduled E2E should disable deployment only if the failure happens on prod, next or main branches.
2. Scheduled E2E should disable all affected workflows per E2E suite.


SokratisVidros:
@AliaksandrRyzhou ☝️

LetItRock (approved):

PR #5541, opened 2024-05-10
feat: add ability to set channels during trigger

scopsy:
- Requested by multiple customers
- Currently users are required to have cronjobs that perform 2 API requests: Create or Update Subscriber, and set channel credentials

We already support creating a subscriber during trigger, this change allows creating it with the channels object for push and chat channels similarly to how ew allow doing that with email and phone today.

### Screenshots

<details>

</details>


packages/application-generic/src/usecases/process-subscriber/process-subscriber.usecase.ts
```diff
@@ -47,7 +47,11 @@ export class ProcessSubscriber {
       subscriberId: subscriberPayload.subscriberId,
     });
 
-    if (subscriber && !subscriberNeedUpdate(subscriber, subscriberPayload)) {
+    if (
+      subscriber &&
+      !subscriberNeedUpdate(subscriber, subscriberPayload) &&
+      !subscriberPayload.channels
```
scopsy: I don't entirely like this, but this is for now until we have a better diff logic happening here

.idea/runConfigurations/_template__of_mocha_javascript_test_runner.xml
```diff
@@ -8,10 +8,10 @@
       <env name="E2E_RUNNER" value="true" />
       <env name="NODE_ENV" value="test" />
       <env name="TS_NODE_COMPILER_OPTIONS" value="{&quot;strictNullChecks&quot;: false}" />
-      <env name="NOVU_ENTERPRISE" value="true" />
+      <env name="NOVU_ENTERPRISE" value="true      " />
     </envs>
     <ui />
-    <extra-mocha-options>--timeout 30000 --require ts-node/register --exit --file e2e/setup.ts</extra-mocha-options>
+    <extra-mocha-options>--require ts-node/register --exit --file e2e/setup.ts</extra-mocha-options>
```
djabarovgeorge, line 14: i wonder what is the default timeout, it does not exist it would be better to have it defined here

djabarovgeorge (approved):
LGTM :)

PR #5546, opened 2024-05-10
[New design system pkg] Update `web` imports to point to new Panda

antonjoel82:

To avoid noise in the main PR (#5467), we handle updating all the existing local `styled-system` references in `web` to the design-system


apps/web/package.json
```diff
@@ -169,6 +171,7 @@
     "react-app-rewired": "^2.2.1",
     "react-error-overlay": "6.0.11",
     "react-scripts": "^5.0.1",
+    "sinon": "9.2.4",
```
SokratisVidros: Is this addition required? Note that Cypress is going away later today.

SokratisVidros (approved):

PR #5549, opened 2024-05-12
feat: add invite team members nudge flow

jainpawan21:

In NV-3697, we want to use Novu workflow to send in-app and after some delay email. hubspot will be used to send email 

Novu workflow :- Trigger -> in-app -> delay (24hr) -> email (with condition on webhook)

Flow:- User trigger the workflow, based on all conditions, in-app notification is sent to the user. After 24hr delay, workflow will create post request to webhookUrl (here webhookurl is new /invites/webhook API). If all conditions are satisfied, email is sent using hubspot.

### Screenshots

<details>

</details>


apps/web/src/AppRoutes.tsx
```diff
@@ -124,6 +123,7 @@ export const AppRoutes = () => {
           <Route path="" element={<LayoutsListPage />} />
         </Route>
         <Route path="/translations/*" element={<TranslationRoutes />} />
+        <Route path={ROUTES.ANY} element={<HomePage />} />
```
davidsoderberg, line 130: Why was this one moved?

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -206,4 +221,68 @@ export class ParseEventRequest {
 
     return reservedVariables?.map((reservedVariable) => reservedVariable.type) || [];
   }
+
+  public async sendInAppNudgeForTeamMemberInvite(command: ParseEventRequestCommand) {
+    // check if this is first trigger
+    const notifications = await this.notificationRepository.count({
+      _organizationId: command.organizationId,
+      _environmentId: command.environmentId,
+    });
```
LetItRock: ☑ **chore:** This will be super costly in terms of Mongo request, what we can do instead is take the first five or whatever number of notifications, and if there are then return.

apps/web/src/AppRoutes.tsx
jainpawan21: catch all route should go in last. after login, it was redirecting to `/workflows` page if user came from `/team ` route

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -206,4 +221,68 @@ export class ParseEventRequest {
 
     return reservedVariables?.map((reservedVariable) => reservedVariable.type) || [];
   }
+
+  public async sendInAppNudgeForTeamMemberInvite(command: ParseEventRequestCommand) {
```
LetItRock, line 241: ☑ **chore:** we don't need to do this for self-hosted

LetItRock, line 241: 🤓 **nitpick (non-blocking):** can we please also cover this logic with e2e tests?

apps/api/src/app/invites/usecases/invite-nudge-webhook/invite-nudge-usecase.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (hmacHash !== hmacHashFromWebhook) {
+        throw new Error('Unauthorized request');
+      }
+
+      const membersCount = await this.memberRepository.count({
+        _organizationId: command?.body?.subscriber?._organizationId,
+      });
+
+      if (membersCount === 1) {
+        return await axiosInstance.post(
```
davidsoderberg: can we return `{send: true}` instead of the axios response here?

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
LetItRock, line 241: ☑ **chore:** and let's maybe also have this under the feature flag so we can disable if something is odd

apps/api/src/app/invites/usecases/invite-nudge-webhook/invite-nudge-usecase.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const axiosInstance = axios.create();
+
+@Injectable({
+  scope: Scope.REQUEST,
+})
+export class InviteNudgeWebhook {
+  constructor(private memberRepository: MemberRepository) {}
+
+  async execute(command: InviteNudgeWebhookCommand) {
```
LetItRock, line 17: ☑ **chore:** same behind the feature flag

apps/web/src/AppRoutes.tsx
davidsoderberg: Ah okay yes makes sense :)

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -206,4 +221,68 @@ export class ParseEventRequest {
...
+          process.env.NOVU_INVITE_TEAM_MEMBER_NUDGE_TRIGGER_IDENTIFIER || 'in-app-invite-team-member-nudge',
+          {
+            to: {
+              subscriberId: command.userId,
+              email: user?.email as string,
+            },
+            payload: {
+              INVITE_TEAM_MEMBER_NUDGE_PAYLOAD_KEY: true,
+              webhookUrl: `${process.env.API_ROOT_URL}/v1/invites/webhook`,
+            },
```
LetItRock: ❓ **question:**shouldn't this be like:
```suggestion
            payload: {
              [INVITE_TEAM_MEMBER_NUDGE_PAYLOAD_KEY]: true,
              ...
            },
```

apps/api/src/app/widgets/usecases/mark-message-as/mark-message-as.usecase.ts
```diff
@@ -56,6 +56,7 @@ export class MarkMessageAs {
     });
 
     if (command.mark.seen != null) {
+      await this.sendAnalyticsEventForInviteTeamNudge(messages);
```
LetItRock: ☑ **chore:** we need to update also the other use-case `MarkMessageAsByMark`

apps/web/src/AppRoutes.tsx
LetItRock, line 130: ❓ **question:** why do we need to move this?

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -22,6 +21,8 @@ type LoginFormProps = {
 
 export function LoginForm({ email, invitationToken }: LoginFormProps) {
   const navigate = useNavigate();
+  const location = useLocation();
+  const state = location.state as { redirectTo?: { pathname?: string } };
```
LetItRock: 🤓 **nitpick (non-blocking):** please put the appropriate types on the `useLocation` hook

apps/web/src/components/layout/RequiredAuth.tsx
```diff
@@ -34,7 +34,7 @@ export function RequiredAuth({ children }: any) {
   }
 
   if (!getToken()) {
-    return <Navigate to={ROUTES.AUTH_LOGIN} replace />;
+    return <Navigate to={ROUTES.AUTH_LOGIN} replace state={{ redirectTo: location }} />;
```
LetItRock, line 37: 👏 **praise:** nice and simple!

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
jainpawan21: Shall I add limit `1` ?

LetItRock: yes, but maybe better would be to do that with `findOne`

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
jainpawan21: I have [this](https://github.com/novuhq/novu/blob/NV-3697/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts#L248) condition before `novu.trigger`.
Wdyt? I should add similar conditions [here](https://github.com/novuhq/novu/blob/NV-3697/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts#L153)

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
jainpawan21: I feel it should work without [], because `INVITE_TEAM_MEMBER_NUDGE_PAYLOAD_KEY` is of type string.
I will test it again. 🙂

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
jainpawan21: Can I copy this line of code after changing the key
https://github.com/novuhq/novu/blob/NV-3697/apps/api/src/app/messages/usecases/get-messages/get-messages.usecase.ts#L61C5-L68C7 ?

apps/api/src/app/widgets/usecases/mark-message-as/mark-message-as.usecase.ts
jainpawan21: for `read` state, I have added condition in [web](https://github.com/novuhq/novu/blob/NV-3697/apps/web/src/components/layout/components/NotificationCenterWidget.tsx#L46)

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
LetItRock: yes, but all code above it is not needed to be executed

LetItRock: yes, but with the new feature flag created in LD and as well as constant in the code

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
LetItRock: it won't, it will create an object like this: `{ INVITE_TEAM_MEMBER_NUDGE_PAYLOAD_KEY: true }` but we want it to be like `{ nv-type-team-member-invite-nudge: true }`

apps/api/src/app/widgets/usecases/mark-message-as/mark-message-as.usecase.ts
LetItRock: ahh ok, got it... I was thinking that we want to call it whenever someone marks the message as seen... but we want this to be called in case of clicking in our NC Widget

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -20,8 +19,16 @@ type LoginFormProps = {
   email?: string;
 };
 
+export interface LocationState {
+  redirectTo: {
+    pathname: string;
+  };
+}
+
```
antonjoel82, line 27: ❓ **question:** Why are we defining the type for the return of the `useLocation` hook here?‏

antonjoel82: ☑ **chore:** We should avoid exporting a generic type from here as we may unknowingly cause circular dependencies‏

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -71,7 +78,12 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
       }
 
       setToken(token);
-      navigate(ROUTES.WORKFLOWS);
+
+      if (state?.redirectTo?.pathname) {
+        navigate(state?.redirectTo?.pathname);
+      } else {
+        navigate(ROUTES.WORKFLOWS);
+      }
```
antonjoel82: ✏ **suggestion:** ‏
```suggestion
      navigate(state?.redirectTo?.pathname || ROUTES.WORKFLOWS);
```

antonjoel82 (commented):
Left a couple notes on the `web` side!

LetItRock (approved):

PR #5554, opened 2024-05-13
Refactor [@novu/client] : remove axios dependency and use fetch instead

BiswaViraj:

- Updated the HttpClient to use fetch
- removed axios dependency
- Moved the http-client to @novu/client
### Screenshots

<details>

</details>


libs/shared/src/services/http-client/api.client.ts
```diff
@@ -14,47 +12,89 @@ export interface IPaginatedResponse<T = unknown> {
 }
 
 export class HttpClient {
```
SokratisVidros: @BiswaViraj Since these `/lib/shared` files are just used by `@novu/client` package, let's move them under `packages/client` to simplify the SDK setup and have all its code collocated. Then, we will have a single, isomorphic HTTP client for Novu API, and we will be golden!

libs/shared/src/services/http-client/api.client.ts
```diff
@@ -14,47 +12,89 @@ export interface IPaginatedResponse<T = unknown> {
...
-  private axiosClient: AxiosInstance;
+  private backendUrl: string;
+  private apiVersion = 'v1';
```
SokratisVidros: Should we make this a public configurable option field to support future versions as well?

libs/shared/src/services/http-client/api.client.ts
```diff
@@ -14,47 +12,89 @@ export interface IPaginatedResponse<T = unknown> {
...
+      method: 'DELETE',
+      headers: this.headers,
+      body: JSON.stringify(body),
+    });
+    const data = await response.json();
+
+    return data.data;
+  }
+
+  private getQueryString(params?: IParamObject) {
```
SokratisVidros: @BiswaViraj we can use [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) API that works in all JS runtimes.

SokratisVidros (changes requested):

libs/shared/src/services/http-client/api.client.ts
BiswaViraj: Tried using the URLSearchParams but it throws an typescript error due to the interface IParamObject.

packages/client/src/http-client/http-client.ts
```diff
@@ -0,0 +1,89 @@
...
+import { CustomDataType } from '@novu/shared';
+
+export class HttpClient {
+  private backendUrl: string;
+  private apiVersion = 'v1';
+  private headers: Record<string, string> = {
+    'Content-Type': 'application/json',
+  };
+
+  constructor(backendUrl: string, options?: ApiOptions) {
```
BiswaViraj: 🗒 **note (non-blocking):** Accept options object to keep it extensible.
In future we can easily add support for timeout, custom headers etc easily without breaking changes.‏

packages/client/src/http-client/http-client.ts
```diff
@@ -0,0 +1,89 @@
...
+  setAuthorizationToken(token: string) {
+    this.headers.Authorization = `Bearer ${token}`;
+  }
+
+  disposeAuthorizationToken() {
+    delete this.headers.Authorization;
+  }
+
+  async getFullResponse(url: string, params?: CustomDataType) {
+    const response = await fetch(
```
SokratisVidros: @BiswaViraj Should we add some error handling such as `response.ok` and a try/catch?

packages/client/src/http-client/http-client.ts
```diff
@@ -0,0 +1,89 @@
...
+        headers: this.headers,
+      }
+    );
+    const data = await response.json();
+
+    return data.data;
+  }
+
+  async post(url: string, body = {}) {
+    const response = await fetch(this.backendUrl + url, {
```
SokratisVidros: @BiswaViraj We can factor all invocations to fetch in a `doFetch` method and only pass the necessary RequestInit and Body parameters to make it more dry.

packages/client/src/http-client/http-client.ts
SokratisVidros: Let's also add `backendUrl` inside options and default it to `api.novu.co`.

SokratisVidros (changes requested):

packages/client/src/http-client/http-client.ts
BiswaViraj: Should we also make this change to the `ApiService`? It will be a breaking change
https://github.com/novuhq/novu/blob/1f49b9b96bf459916588482e3489d86bca8ef1f2/packages/client/src/api/api.service.ts
@SokratisVidros

SokratisVidros: We can be backward compatible. PTAL at a similar pattern in [@novu/node](https://github.com/novuhq/novu/blob/next/packages/node/src/lib/novu.ts#L44).

SokratisVidros (approved):
💯

PR #5568, opened 2024-05-14
fix(web): Stop fetching context data on window focus

SokratisVidros:
Every time the window gets focused, six requests are triggered to fetch orgs, users, changes, subscriptions, etc...

### Screenshots

<img width="624" alt="Screenshot 2024-05-14 at 18 37 02" src="https://github.com/novuhq/novu/assets/1352422/64260d02-622a-4910-b3e9-0e3a6584f500"

### DIFF
--- a/apps/web/src/Providers.tsx
+++ b/apps/web/src/Providers.tsx
@@ -20,7 +20,6 @@ const queryClient = new QueryClient({
     queries: {
       queryFn: defaultQueryFn as any,
       refetchOnWindowFocus: false,
-      retry: false,
     },
   },
 });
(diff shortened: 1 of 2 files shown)


apps/web/src/Providers.tsx
```diff
@@ -20,7 +20,6 @@ const queryClient = new QueryClient({
     queries: {
       queryFn: defaultQueryFn as any,
       refetchOnWindowFocus: false,
```
rifont, line 22: Refetch on window focus is already disabled on this line, it’s not clear what this change is achieving by removing the retry option. Could you elaborate?

SokratisVidros: Please ignore. This change was pushed as part of the PW migration.

PR #5570, opened 2024-05-14
perf(web): Convert image assets to `webp` & skip flaky tests

antonjoel82:
### What changed? Why was the change needed?

- Skip the flaky logout test since we're switching to Playwright soon
- Convert non-provider image assets in `public/static/images` to `webp` to massively reduce size

### Validation

- Please do a visual skim to make sure I didn't miss replacing any paths. Primarily-affected areas:
   - login
   - quickstart
   - nav bar (avatar)

PR #5575, opened 2024-05-15
fix(web): fixed the flaky failing e2e tests after the changes we introduced with the async launch darkly

LetItRock:
Fixed a couple of failing e2e tests in the Web application. 

### Screenshots

<details>

</details>

### DIFF
--- a/apps/web/cypress/tests/auth.spec.ts
+++ b/apps/web/cypress/tests/auth.spec.ts
@@ -223,13 +223,16 @@ describe('User Sign-up and Login', function () {
 
     it('should logout user when auth token is expired', function () {
       // login the user
-      cy.visit('/auth/login');
+      cy.waitLoadFeatureFlags(() => {
+        cy.visit('/auth/login');
+      });
+
       cy.getByTestId('email').type('test-user-1@example.com');
       cy.getByTestId('password').type('123qwe!@#');
       cy.getByTestId('submit-btn').click();
 
       cy.waitLoadFeatureFlags();
-
+      cy.waitForNetworkIdle(500);
       cy.location('pathname').should('equal', '/workflows');
 
       // setting current time in future, to simulate expired token
@@ -238,9 +241,7 @@ describe('User Sign-up and Login', function () {
       const date = new Date(Date.now() + THIRTY_DAYS + ONE_MINUTE);
       cy.clock(date);
 
-      cy.visit('/subscribers');
-
-      cy.waitLoadFeatureFlags();
+      cy.getByTestId('side-nav-subscribers-link').click();
 
       // checking if token is removed from local storage
       cy.getLocalStorage('auth_token').should('be.null');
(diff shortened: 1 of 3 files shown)


apps/web/cypress/tests/auth.spec.ts
```diff
@@ -223,13 +223,16 @@ describe('User Sign-up and Login', function () {
...
+        cy.visit('/auth/login');
+      });
+
       cy.getByTestId('email').type('test-user-1@example.com');
       cy.getByTestId('password').type('123qwe!@#');
       cy.getByTestId('submit-btn').click();
 
       cy.waitLoadFeatureFlags();
-
+      cy.waitForNetworkIdle(500);
```
LetItRock, line 235: Seems like there were some race conditions between the network and logic to mock the launch darkly requests, but not 100% sure.

apps/web/cypress/tests/auth.spec.ts
```diff
@@ -238,9 +241,7 @@ describe('User Sign-up and Login', function () {
       const date = new Date(Date.now() + THIRTY_DAYS + ONE_MINUTE);
       cy.clock(date);
 
-      cy.visit('/subscribers');
-
-      cy.waitLoadFeatureFlags();
+      cy.getByTestId('side-nav-subscribers-link').click();
```
LetItRock, line 244: when using the `cy.visit('/subscribers');` it looks like that application is hanging with the loader... @antonjoel82 we have to verify that when the token is expired and the user refreshes the page he will be navigated to the login page...

apps/web/cypress/tests/auth.spec.ts
```diff
@@ -223,13 +223,16 @@ describe('User Sign-up and Login', function () {
 
     it('should logout user when auth token is expired', function () {
       // login the user
-      cy.visit('/auth/login');
+      cy.waitLoadFeatureFlags(() => {
+        cy.visit('/auth/login');
+      });
```
LetItRock, line 228: wait for LD mocked request to finish

SokratisVidros (approved):

apps/web/cypress/tests/auth.spec.ts
antonjoel82: I don't think this is necessary because we don't block at `/auth/login` anyway

apps/web/cypress/tests/auth.spec.ts
antonjoel82: Validated this manually, but as mentioned via slack and the original PR, we deemed it wasn't worthwhile to try to fix and fight with all these tests if we're going to delete them this week anyway.

I elected to skip in this PR (https://github.com/novuhq/novu/pull/5570)

apps/web/cypress/tests/auth.spec.ts
LetItRock: all the routes are wrapped in LD provider... but I'll check it, because I've added all the changes at the same time which helped

SokratisVidros (approved):

PR #5578, opened 2024-05-15
feat(ci): Community releases

AliaksandrRyzhou:
A separate workflow to build Community Docker images that should be a multi-architecture.

### DIFF
--- a/.github/workflows/community-releases.yml
+++ b/.github/workflows/community-releases.yml
@@ -38,7 +38,7 @@ jobs:
         run: |
           service=${{ matrix.name }}
           echo "SERVICE_NAME=$(basename "${service//-/-}")" >> $GITHUB_ENV
-          echo "REGISTRY_OWNER=aliaksandrryzhou" >> $GITHUB_ENV
+          echo "REGISTRY_OWNER=novuhq" >> $GITHUB_ENV
 
       - name: Install pnpm
         uses: pnpm/action-setup@v3
--- a/.github/workflows/community-releases.yml
+++ b/.github/workflows/community-releases.yml
@@ -1,4 +1,4 @@
-name: Docker Community Release
+name: Prepare Self-hosted Release
 
 on:
   workflow_dispatch:
--- a/.github/workflows/community-releases.yml
+++ b/.github/workflows/community-releases.yml
@@ -1,0 +1,98 @@
+name: Docker Community Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The version to tag docker images'
+        required: true
+        type: string
+
+permissions:
+  contents: write
+  packages: write
+  deployments: write
+  id-token: write
+
+jobs:
+  prepare:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/community-releases.yml
```diff
@@ -0,0 +1,98 @@
+name: Docker Community Release
```
SokratisVidros:
```suggestion
name: Prepare Self-hosted Release
```

SokratisVidros: Let's also name it `prepare-self-hosted-release.yml` to match the prepare cloud release naming.

.github/workflows/community-releases.yml
```diff
@@ -0,0 +1,98 @@
...
+          DOCKER_BUILD_ARGUMENTS: >
+            --cache-from type=registry,ref=ghcr.io/${{ env.REGISTRY_OWNER }}/cache:build-cache-${{ env.SERVICE_NAME }}-community
+            --cache-to type=registry,ref=ghcr.io/${{ env.REGISTRY_OWNER }}/cache:build-cache-${{ env.SERVICE_NAME }}-community,mode=max
+            --platform=linux/amd64,linux/arm64 --provenance=false
+            --output=type=image,name=ghcr.io/${{ env.REGISTRY_OWNER }}/${{ env.SERVICE_NAME }},push-by-digest=true,name-canonical=true
+        run: |
+          cd apps/$SERVICE_NAME && pnpm run docker:build
+          docker images
+
+      - name: Tag and Push docker image
```
SokratisVidros, line 77: @AliaksandrRyzhou Should we also add a [health-check step](https://github.com/novuhq/novu/blob/f16076233ae6319c24989ec11b09ed8849ee48ee/.github/workflows/reusable-docker.yml#L162)?

SokratisVidros (commented):
@AliaksandrRyzhou 😍 starting fresh!!!

.github/workflows/community-releases.yml
```diff
@@ -0,0 +1,98 @@
...
+
+  docker_build:
+    runs-on: ubuntu-latest
+    needs: prepare
+    timeout-minutes: 80
+    environment: Production
+    strategy:
+      fail-fast: false
+      matrix:
+        name: ['novu/api','novu/worker','novu/inbound-mail','novu/web','novu/webhook','novu/widget','novu/ws']
```
LetItRock, line 32: missing embed

SokratisVidros (approved):

.github/workflows/community-releases.yml
```diff
@@ -0,0 +1,98 @@
+name: Docker Community Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'The version to tag docker images'
+        required: true
+        type: string
```
LetItRock, line 9: I think that it would be nice to have an option to build and release just specified images. I can imagine that when some images fail we would need to have an option just to build for example two `novu/api, novu/worker` after we have applied the fix. WDYT?

PR #5579, opened 2024-05-15
[Design System] Deprecations, linting, and helpers

antonjoel82:

- Deprecate old `design-system` concepts that should now be done through design tokens
- Add helpers for getting the color scheme
- Add linting for direct imports of `@mantine/core` in `web`
- Add Netlify configurations to the `toml` --  we were implicitly linting (all of web, not just changed files) during `deploy-preview` builds, which I don't think we should be doing. This also has the added benefit of giving us version control to these commands


apps/web/.eslintrc.js
```diff
@@ -22,8 +22,16 @@ module.exports = {
...
       'error',
       {
-        paths: ['src'],
-        patterns: ['@novu/dal'],
+        paths: [
+          {
+            name: '@novu/dal',
+          },
+          {
+            name: '@mantine/core',
```
antonjoel82, line 30: 🗒 **note (non-blocking):** I think this will be the best way to prevent people from creating more debt of direct imports from Mantine. If they are truly needed, an `eslint-disable` will suffice‏

apps/web/netlify.toml
```diff
@@ -1,3 +1,13 @@
+[build.environment]
+  DISABLE_ESLINT_PLUGIN="true"
```
antonjoel82, line 2: 🗒 **note (non-blocking):** This disables linting when performing Netlify builds -- this was built-in to `react-app-rewired`. 

In general, I don't believe Netlify should be concerned with linting, and it should be done in a separate, modular CI action. For this specific PR though, it was required because it _lints the entire package‏_, which will fail with the new rule.

apps/web/.eslintrc.js
SokratisVidros: Nice!

SokratisVidros: Nice!

apps/web/.eslintrc.js
```diff
@@ -22,8 +22,16 @@ module.exports = {
...
+            message:
+              'Please avoid referencing @mantine/core directly in new or updated code. Instead, import from @novu/novui',
```
SokratisVidros, line 32: Is it `@novu/ui` or is it a word play?

apps/web/netlify.toml
SokratisVidros: I agree 100%. Linting is a local, pre-commit concern.

SokratisVidros (approved):

PR #5589, opened 2024-05-16
Update github actions

Cliftonz:
In this PR we update Github actions to ensure that the standard github actions leveraged in our pipeline do not stop everything from working on May 27th


SokratisVidros (approved):

AliaksandrRyzhou (approved):
@Cliftonz what do you think about an update of the following actions too?

- pnpm/action-setup@v2
- actions/cache@v3
- hashicorp/setup-terraform@v1
- aws-actions/configure-aws-credentials@v1

We can find warnings about it [here](https://github.com/novuhq/novu/actions/runs/9112838926) 

Also, I think it's important to update `set-output` [here](https://github.com/novuhq/novu/blob/58fe1e321b4765dd6dafe23d35b4c01737871d3d/.github/workflows/reusable-api-e2e.yml#L39-L41) because this command is deprecated and will be disabled soon. New format you can find [here](https://github.com/novuhq/novu/blob/58fe1e321b4765dd6dafe23d35b4c01737871d3d/.github/workflows/reusable-widget-e2e.yml#L32-L34)

SokratisVidros:
@Cliftonz what's the latest update on this PR? Should we treat it as a priority? If

PR #5592, opened 2024-05-17
[Design System] Remove Panda from old Design System & migrate component audit

antonjoel82:

- Remove Panda from old Design System & migrate component audit
- Deletes a bunch of files, and ensures that we don't look in the wrong place -- `novui` should be the new source of truth for all things Panda

_Note: this will fail building until #5546 is merged with import changes_


libs/novui/src/component-audit/ComponentAuditTable.tsx
```diff
@@ -68,48 +68,111 @@ export const ComponentAuditTable: React.FC<ComponentAuditTableProps> = ({ data:
     : data;
 
   return (
-    <table className={`${css({ width: '100%', borderCollapse: 'collapse' })} ${className ?? ''}`}>
+    <table className={`${css({ width: '[100%]', borderCollapse: 'collapse' })} ${className ?? ''}`}>
```
antonjoel82: 🗒 note (non-blocking): "strict" tokens really hold you accountable 😅 ‏

SokratisVidros (approved):
🧹

PR #5610, opened 2024-05-20
fix(web): Email add-logo to brand page redirect

ChmaraX:
Going to `Workflow -> Add Email channel -> Add logo` would redirect to `/brand` path, which doesn't exist when information architecture is enabled and as a result would default `/workflows`.


apps/web/src/pages/templates/components/email-editor/EmailMessageEditor.tsx
```diff
@@ -83,7 +87,7 @@ export function EmailMessageEditor({
   }
 
   function getBrandSettingsUrl(): string {
-    return '/brand';
+    return isInformationArchitectureEnabled ? ROUTES.BRAND_SETTINGS : ROUTES.BRAND;
```
scopsy, line 90: cc. @antonjoel82

SokratisVidros (approved):

PR #5612, opened 2024-05-20
chore(root): Move @novu/application generic under libs

SokratisVidros:
This package contains shared code for Novu services. It is considered an internal library and should be deprecated from NPM as it can't be used independently from Novu services powered by Nest.js.


libs/application-generic/package.json
```diff
@@ -5,11 +5,7 @@
   "main": "build/main/index.js",
   "typings": "build/main/index.d.ts",
   "module": "build/module/index.js",
-  "private": false,
-  "repository": {
-    "url": "https://github.com/novuhq/novu",
-    "directory": "packages/nest"
-  },
+  "private": true,
```
SokratisVidros, line 8: All the other files are moved under the `libs` folder. Package.json was changed to make this package private.

libs/application-generic/package.json
```diff
@@ -138,18 +130,6 @@
     "LICENSE",
     "README.md"
   ],
-  "ava": {
```
SokratisVidros, line 141: I guess we are not using ava anymore :)

rifont: 🧹

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.command.ts
```diff
@@ -16,7 +16,7 @@ import {
 } from '@novu/shared';
 
 import { EnvironmentWithUserCommand } from '../../../commands';
-import { NotificationStep } from '../../../usecases';
+import { NotificationStep } from '../..';
```
rifont, line 19:
```suggestion
import { NotificationStep } from '../../usecases';
```

I often find these implicit imports from the index file to be troublesome for introduction of cyclic dependencies. Can we revert?

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
```diff
@@ -37,7 +37,7 @@ import {
   CreateMessageTemplateCommand,
   NotificationStep,
   NotificationStepVariantCommand,
-} from '../../../usecases';
+} from '../..';
```
rifont, line 40: Same as above, risk of cyclic dependency
```suggestion
} from '../../usecases';
```

rifont (approved):
Awesome! Nice tidy up too.

rifont:
The PR appears to need another install, the [Get Affected Packages](https://github.com/novuhq/novu/actions/runs/9165935028/job/25200346545?pr=5612#step:5:183) workflow is complaining about lockfile changes.

Got a few pesky cspell [errors](https://github.com/novuhq/novu/actions/runs/9165935028/job/25200346827?pr=5612) too.

SokratisVidros:
> The PR appears to need another install, the [Get Affected Packages](https://github.com/novuhq/novu/actions/runs/9165935028/job/25200346545?pr=5612#step:5:183) workflow is complaining about lockfile changes.
> 
> Got a few pesky cspell [errors](https://github.com/novuhq/novu/actions/runs/9165935028/job/25200346827?pr=5612) too.

👍  The first one with the lockfile is trickier than it looks as all the pnpm-lock changes are committed but pnpm still refuses to cooperate. The good news is that it's reproducible locally so I will spend some time to figure out it today.

PR #5615, opened 2024-05-21
chore(root): Fix for prepare-cloud-release checkout

SokratisVidros:
Checkout only the last commit.  0 indicates all history for all branches and tags creating the wrong prod history.


rifont (approved):
Nice one, just the PR title to fix.

PR #5619, opened 2024-05-21
feat(echo): add skip support

djabarovgeorge:
Add support for skip result to be returned to Novu Cloud during Echo Step execution.
Update the Resonate use-case to use the skip result to conditionally skip the execution of the Echo step

### Screenshots

image

image

<details>

### Related enterprise PR
[EE-PR](https://github.com/novuhq/packages-enterprise/pull/137)

</details>


apps/worker/src/app/workflow/usecases/add-job/add-job.usecase.ts
djabarovgeorge, line 1: needed because I separated variable normalize from filter logic

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
djabarovgeorge, line 1: bug-fix: we were overring the canceled status (by filter) to complete.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -74,19 +77,27 @@ export class SendMessage {
...
   @InstrumentUsecase()
-  public async execute(command: SendMessageCommand) {
+  public async execute(command: SendMessageCommand): Promise<{ status: 'success' | 'canceled' }> {
     const payload = await this.buildCompileContext(command);
 
-    const [shouldRun, preferred] = await Promise.all([
-      this.filter(command, payload),
-      this.filterPreferredChannels(command.job),
-    ]);
+    const variables = await this.normalizeVariablesUsecase.execute(
```
djabarovgeorge, line 90: the reason for the separation of variable normalize and filter.
we need to pass the normalized variables to `resonate` use case, if the step is skipped we have no reason to execute the `filter` function. by that we will reduce the possible compute and IO operation.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -97,60 +108,47 @@ export class SendMessage {
...
+      );
     }
 
-    if (!shouldRun?.passed || !preferred) {
+    if (!filterResult?.passed || !channelPreferenceResult || resonateResponse?.options?.skip) {
       await this.jobRepository.updateStatus(command.environmentId, command.jobId, JobStatusEnum.CANCELED);
 
-      return;
+      const echoSkip = resonateResponse?.options?.skip;
+      await this.executionLogRoute.execute(
```
djabarovgeorge, line 130: before we created an execution log only if the filter did not pass, now we will log execution detail on every step halt.

packages/application-generic/src/usecases/conditions-filter/conditions-filter.usecase.ts
djabarovgeorge, line 1: I separated variable normalize from filter logic here.

packages/application-generic/src/usecases/select-integration/select-integration.usecase.ts
djabarovgeorge, line 1: needed because I separated variable normalize from filter logic

packages/echo/src/client.ts
djabarovgeorge, line 1: Will love to hear the opinions of people that familiar with Echo SDK regarding the changes here

packages/echo/src/types/step.types.ts
```diff
@@ -90,7 +90,7 @@ export type CustomStep = <
     /**
      * Skip the step. If the skip function returns true, the step will be skipped.
      *
-     * @param payload The payload for the step.
+     * @param inputs The inputs for the step.
```
djabarovgeorge, line 92: typo?

packages/echo/src/client.ts
```diff
@@ -496,6 +497,13 @@ export class Echo {
 
   private executeStepFactory<T, U>(event: IEvent, setResult: (result: any) => void): ActionStep<T, U> {
     return async (stepId, stepResolve, options) => {
+      if (await this.shouldSkip(options?.skip, event.data)) {
```
djabarovgeorge: i don't see a reason to execute the step is we will skip the step.

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
```diff
@@ -99,4 +100,98 @@ describe('Echo Trigger ', async () => {
...
+    expect(executedMessage2.length).to.be.eq(1);
+    expect(executedMessage2[0].status).to.be.eq(JobStatusEnum.CANCELED);
+  });
 });
+
+async function syncWorkflow(session) {
+  const resultDiscover = await axios.get(echoServer.serverPath + '/echo?action=discover');
+
+  await session.testAgent.post(`/v1/echo/sync`).send({
+    chimeraUrl: echoServer.serverPath + '/echo',
```
rifont: ⚠ **issue:** ‏address removal of `chimera` prefix

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -97,60 +108,47 @@ export class SendMessage {
         ExecuteOutput<IChimeraChannelResponse> | null
       >({
         ...command,
-        variables: shouldRun.variables,
+        variables: variables,
```
rifont:
```suggestion
        variables,
```

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -173,55 +171,127 @@ export class SendMessage {
...
+      case StepTypeEnum.EMAIL: {
+        await this.sendMessageEmail.execute(sendMessageCommand);
+
+        break;
+      }
+      case StepTypeEnum.CHAT: {
+        await this.sendMessageChat.execute(sendMessageCommand);
+
+        break;
+      }
```
rifont: Line-break consistency.
```suggestion
      case StepTypeEnum.EMAIL: {
        await this.sendMessageEmail.execute(sendMessageCommand);
        break;
      }
      case StepTypeEnum.CHAT: {
        await this.sendMessageChat.execute(sendMessageCommand);
        break;
      }
```

packages/echo/src/types/step.types.ts
rifont: Yep, good catch.

libs/application-generic/src/usecases/select-variant/select-variant.usecase.ts
```diff
@@ -43,7 +60,7 @@ export class SelectVariant {
           userId: command.userId,
           step: command.step,
           job: command.job,
-          variables: command.filterData,
+          variables: variables,
```
rifont:
```suggestion
          variables,
```

libs/application-generic/src/usecases/select-integration/select-integration.spec.ts
```diff
@@ -117,7 +117,10 @@ describe('select integration', function () {
           new ExecutionLogQueueService(new WorkflowInMemoryProviderService()),
           new GetFeatureFlag(new FeatureFlagsService())
         ),
-        new CompileTemplate()
+        // @ts-ignore
+        new CompileTemplate(),
+        // @ts-ignore
+        new NormalizeVariables()
       ),
```
rifont: If it's really necessary to override type safety, prefer to use `ts-expect-error` with a trailing `- <reason>` instead of `ts-ignore`. `ts-expect-error` will fail if the types are fixed, whilst `ts-ignore` will not. This helps us to keep on top of type-safety.

Preferably, can we fix the types here? If not possible, replace with `ts-expect-error`

```suggestion
        // @ts-expect-error - <insert reason here>
        new CompileTemplate(),
        // @ts-expect-error - <insert reason here>
        new NormalizeVariables()
      ),
```

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
```diff
@@ -0,0 +1,117 @@
...
+  TenantRepository,
+} from '@novu/dal';
+import { IMessageFilter } from '@novu/shared';
+import { IFilterVariables } from '../../utils/filter-processing-details';
+import { CachedEntity } from '../../services/cache/interceptors/cached-entity.interceptor';
+import { buildSubscriberKey } from '../../services/cache/key-builders/entities';
+import { ConditionsFilterCommand } from '../conditions-filter';
+
+@Injectable()
+export class NormalizeVariables {
```
rifont, line 26: Can we add a typedoc to the top of this file explaining the purpose of the use-case? `Normalize` can take a few different meanings, it will be helpful to our future selves to have a quick explanation.

packages/echo/src/client.ts
rifont: This interface looks great - it's possible we'll add other meta-options in the future, such as retry policy per step. This is good start.

Here are other options GPT suggested we could add in the future:
1. Retry Policy: Define how many times and under what conditions a step should be retried if it fails.
2. Timeout: Specify a maximum duration for the step execution.
3. Condition: Execute the step only if certain conditions are met.
4. Priority: Assign a priority to the step to influence the order of execution.
5. Rate Limiting: Control the rate at which the step can be executed.
6. Logging: Enable or configure logging for the step execution.
7. Error Handling: Define custom error handling strategies for the step.
Dependencies: Specify other steps that must be completed before this step can execute.
9. Notifications: Send notifications on step completion, failure, or other events.
10. Audit: Enable auditing to track changes and executions of the step.

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
djabarovgeorge: Totally agree, I will add more info there.

libs/application-generic/src/usecases/select-integration/select-integration.spec.ts
djabarovgeorge: thanks, will fix this one

packages/echo/src/client.ts
denis-kralj-novu: 💭 **thought (non-blocking):** ‏ 8. sounds like an option that would spawn an infinite loop 🤣

packages/echo/src/client.ts
```diff
@@ -49,6 +49,7 @@ import {
 import { Schema } from './types/schema.types';
 import { EMOJI, log } from './utils';
 import { VERSION } from './version';
+import { Skip } from './types/skip.types';
```
denis-kralj-novu, line 52: ❓ **question:** ‏ Type import was needed? I don't see it popping up in the additions below

packages/echo/src/types/skip.types.ts
```diff
@@ -1 +1 @@
-export type Skip = (payload: any) => boolean | Promise<boolean>;
+export type Skip = boolean;
```
denis-kralj-novu: ❓ **question:** ‏ Why is there a need to formalize this boolean flag as its own type? I am unsure of what the function if that is

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
```diff
@@ -0,0 +1,128 @@
...
+  private async fetchSubscriberIfMissing(
+    command: ConditionsFilterCommand,
+    filters: IMessageFilter[]
+  ): Promise<SubscriberEntity | undefined> {
+    if (command.variables?.subscriber) {
+      return command.variables.subscriber;
+    }
+
+    const subscriberFilterExist = filters?.find((filter) => {
+      return filter?.children?.find((item) => item?.on === 'subscriber');
```
denis-kralj-novu: ✏ **suggestion:** ‏ we could exchange the magic string for the enum `FilterPartTypeEnum.SUBSCRIBER`

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
```diff
@@ -0,0 +1,128 @@
...
+          subscriberId: command.job.subscriberId,
+          _environmentId: command.environmentId,
+        })) ?? undefined
+      );
+    }
+
+    return undefined;
+  }
+
+  private async fetchTenantIfMissing(
```
denis-kralj-novu, line 83: 🤓 **nitpick (non-blocking):** ‏ Is the `**IfMissing()` pattern of naming precise enough? It kind of implies that if it is not missing it won't be fetched. Maybe just keep the `fetch<THING>()` as the name (or exchange `fetch` for `get` or `find`)

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
```diff
@@ -0,0 +1,128 @@
...
+    command: ConditionsFilterCommand,
+    filters: IMessageFilter[]
+  ): Promise<TenantEntity | undefined> {
+    if (command.variables?.tenant) {
+      return command.variables.tenant;
+    }
+
+    const tenantIdentifier =
+      typeof command.job?.tenant === 'string'
```
denis-kralj-novu, line 92: ❓ **question:** ‏ when do we get into a situation that the tenant isn't a `ITenantDefine` but a string?

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
denis-kralj-novu, line 1: 🗒 **note (non-blocking):** ‏ I've noticed that this is simply moving a helper method logic into a dedicated use-case. If the goal is just to move without refactoring than questions/suggestions in this file can be ignored

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
djabarovgeorge: good catch, strange looks like the chimera url refactor i did in prague was lost, i am not sure why and when ... i will rewrite it..

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
djabarovgeorge: yes, it was moved to a dedicated use case without changes in order to reuse this logic when needed.

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
djabarovgeorge: good catch! :)

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
djabarovgeorge: hmm well if `command.variables?.tenant` exists we will return it and won't execute any retrievals.

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
djabarovgeorge: good questions, my assumption will be that it is possible backward compatibility.

packages/echo/src/client.ts
djabarovgeorge: good catch! It looks like leftovers to me 🧹

packages/echo/src/types/skip.types.ts
djabarovgeorge: initially, we created the Skip option as a function, however after a small overview with Richard we decided that we could simplify this option for now.
instead of: `skip :() => param.skip`
pass: `skip : param.skip`

I left it as a type so if in the future we will change this behavior it will be easy to change in one place.

let me know what you think.

denis-kralj-novu: Got it, wrap it in a type so that when we extend, we only need to update the definition and places that need the new behavior, other places would stay the same

libs/application-generic/src/usecases/normalize-variables/normalize-variables.usecase.ts
denis-kralj-novu: true, but this exact element of the process is missing in the function name, the name only concerns itself with the situation when `command.variables?.tenant` is missing.

Not much value in dying on this hill though, in general I think the name is acceptable

apps/api/e2e/utils.ts
```diff
@@ -0,0 +1,122 @@
+//# cSpell:disable
```
denis-kralj-novu: Is there a particular reason we moved away from [fakerjs](https://fakerjs.dev/)?

djabarovgeorge: [REF ](https://novu.slack.com/archives/C06DU7A7BPV/p1717411734557139)

denis-kralj-novu: thank you :)

djabarovgeorge:
@rifont I reverted the custom faker logic and updated the Skip option to be a function and a bit more generic. let's see if we can merge this one.

apps/api/migrations/normalize-message-template-cta-action/normalize-message-template-cta-action-migration.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { expect } from 'chai';
-import { faker } from '@faker-js/faker';
+import faker from '@faker-js/faker';
```
rifont: I don't think this should have changed, right?

```suggestion
import { faker } from '@faker-js/faker';
```

libs/application-generic/src/usecases/select-variant/select-variant.spec.ts
```diff
@@ -28,8 +29,9 @@ describe('select variant', function () {
     selectVariantUsecase = new SelectVariant(
       // @ts-ignore
       new ConditionsFilter(),
-      new MessageTemplateRepository()
+      new MessageTemplateRepository(),
       // @ts-ignore
+      new NormalizeVariables()
```
rifont: The `@ts-ignore` here seems to have been applied to nothing before this diff. Can we remove it?

```suggestion
      new NormalizeVariables()
```

packages/echo/src/client.ts
```diff
@@ -503,6 +511,14 @@ export class Echo {
...
   }
 
+  private async shouldSkip(skip: Skip | undefined, payload: Record<string, unknown>): Promise<boolean> {
+    if (!skip) {
+      return false;
+    }
+
+    return skip(payload);
+  }
+
```
rifont: The `payload` is freely usable in the skip function after hydration occurs. It's the `inputs` which are contextually available only for this step. We need to update naming to reflect this.

```suggestion
  private async shouldSkip(skip: Skip | undefined, inputs: Record<string, unknown>): Promise<boolean> {
    if (!skip) {
      return false;
    }

    return skip(inputs);
  }

```

packages/echo/src/client.ts
rifont: There's a good reason not to execute the step - Developers may have side effects as part of Custom steps, or indeed any steps, that they would not like to have executed when they intended to skip the step execution altogether.

apps/api/migrations/encrypt-api-keys/encrypt-api-keys-migration.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { expect } from 'chai';
-import { faker } from '@faker-js/faker';
 import { createHash } from 'crypto';
+import faker from '@faker-js/faker';
```
SokratisVidros: Where did we land on faker and migrations? I was under the impression we don’t need it here to avoid the docker issue Alex faced as migration code is part of API code.

rifont (commented):
Pushing my recent comments, I still need to test this locally.

packages/echo/src/client.ts
djabarovgeorge: Just to confirm, if a step is marked as skipped, should we also skip its execution?

packages/echo/src/types/skip.types.ts
djabarovgeorge: In the end, we decided to bring back the function type to provide the Product Person the ability to pass inputs to manipulate the skip. 

``` skip :(inputs) => inputs.skip ```

File: apps/api/migrations/normalize-message-template-cta-action/normalize-message-template-cta-action-migration.spec.ts (hunk 17 above)
Comment (djabarovgeorge) [reply apps/api/migrations/normalize-message-template-cta-action/normalize-message-template-cta-action-migration.spec.ts]: true it was created like this by the idea, i will revert it.
thanks for noticing!!

**Comment by denis-kralj-novu** [comment]
Want to point out that all the echo related tests are failing with a timeout 🙁 . This happens on my branches as well however, but given that you added a test f or the skip wanted to mention it.

File: packages/echo/src/client.ts (hunk 4 above)
Comment (rifont) [reply packages/echo/src/client.ts]: Yes, sorry if my comment caused confusion - we're aligned that we should skip execution `skip` function returns `true`.

File: apps/api/migrations/normalize-message-template-cta-action/normalize-message-template-cta-action-migration.spec.ts (hunk 17 above)
Comment (rifont) [reply apps/api/migrations/normalize-message-template-cta-action/normalize-message-template-cta-action-migration.spec.ts]: @djabarovgeorge we still have a number of occurrences of these changes we need to revert.

Comment (djabarovgeorge) [reply apps/api/migrations/normalize-message-template-cta-action/normalize-message-template-cta-action-migration.spec.ts]: I think we are ok. i updated those cases.

File: apps/api/src/app/testing/echo/echo-sync.e2e-ee.ts [hunk 21]
```diff
@@ -260,7 +260,7 @@ describe('Echo Sync - /echo/sync (POST)', async () => {
     expect(messageTemplatesToTest.output).to.equal(undefined);
 
     await session.testAgent.post(`/v1/echo/sync`).send({
-      chimeraUrl: 'https://chimera.novu.com',
```
SokratisVidros: 👋 👋  => 🐉

packages/echo/src/client.ts
```diff
@@ -473,6 +474,15 @@ export class Echo {
   private executeStepFactory<T, U>(event: IEvent, setResult: (result: any) => void): ActionStep<T, U> {
     return async (stepId, stepResolve, options) => {
       const step = this.getStep(event.workflowId, stepId);
+      const eventClone = clone<IEvent>(event);
```
SokratisVidros, line 477: ❓ Why is cloning necessary at this point? I'd suggest we opt-in for a more type-safe cloning function if we have to.

packages/echo/src/types/skip.types.ts
```diff
@@ -1 +1,3 @@
-export type Skip = (payload: any) => boolean | Promise<boolean>;
+import { MaybePromise } from './util.types';
```
SokratisVidros: Can we please rename this to [Awaitable](https://github.com/microsoft/TypeScript/issues/31394)? This is a public type so a good naming is crucial!

packages/echo/src/types/skip.types.ts
```diff
@@ -1 +1,3 @@
...
+
+export type Skip<T_Inputs = any> = (inputs: T_Inputs) => MaybePromise<boolean>;
```
SokratisVidros:
```suggestion
export type Skip<T> = (inputs: T) => Awaitable<boolean>;
```

SokratisVidros (changes requested):

packages/echo/src/client.ts
djabarovgeorge: The main reason is because ajv.validate is updating inputs by ref which causes unpredicted behavior. 

I'd like to refactor our Echo.validate method to use a deep clone and return the inputs as the return value and not by ref, so the code will be more readable and safe. at the moment every change in the client that is related to ajv validation can cause potential bugs. I will do the refactor in a separate PR.

apps/api/src/app/testing/echo/echo-sync.e2e-ee.ts
djabarovgeorge: good point, will rename all of them.

apps/api/migrations/encrypt-api-keys/encrypt-api-keys-migration.spec.ts
djabarovgeorge: it issue resolved once we added the faker package in the devDependencies section on the package json file.

packages/echo/src/types/skip.types.ts
djabarovgeorge: Should we keep the MaybePromise as backward compatibility for a couple of versions?

packages/echo/src/client.ts
```diff
@@ -855,3 +879,7 @@ export class Echo {
     return getCodeResult;
   }
 }
+
+function clone<Result>(data: unknown) {
+  return JSON.parse(JSON.stringify(data)) as Result;
+}
```
rifont, line 886: Let's move this to a standalone `json.utils.ts` file.

packages/echo/src/types/skip.types.ts
SokratisVidros: It's not needed. This is an alpha SDK, so far, and there are no guarantees for backward compatibility.

SokratisVidros (approved):

scopsy:
Just an update, found an issue during testing that skip was applied during a preview which broke the web preview for skipped steps. 

Updated the code and added a unit test for this scenario.

packages/echo/src/client.ts
djabarovgeorge: Agree I will refactor it as part of the Echo.validate refactor

PR #5624, opened 2024-05-22
refactor(web): Single theme entry point & ensure Mantine singletons

rifont:
* Move `ThemeProvider` to root of component tree. This ensures that everything has theming context available
* pin `@types/react` version via PNPM `overrides` to ensure that `@mantine/*` dependencies resolve to a singleton in the PNPM virtual filesystem
  * Pnpm v9 automatically installs optional deps by default when they match an existing semver, and we have other packages declaring `@types/react` as an optional dependency which results in multiple PNPM virtual fileystem paths for Mantine imports.
  * This resolves an issue where the color scheme couldn't be resolved due to lack of a `@mantine/core` singleton, which provides necessary context
* Use `AuthLayout` in routes to provide a single entry point for styling all unauthenticated pages.
* Declare `@mantine/core` as a `peerDependency` and `devDependency` in all packages except `@novu/web`

### Screenshots

https://github.com/novuhq/novu/assets/32132657/1d7ae862-e922-43dd-a131-b98a250a99df

<details>

</details>


apps/web/src/AppRoutes.tsx
```diff
@@ -53,28 +54,30 @@ export const AppRoutes = () => {
...
-      />
-      <Route
-        path={ROUTES.PARTNER_INTEGRATIONS_VERCEL_LINK_PROJECTS_EDIT}
-        element={
-          <EnsureOnboardingComplete>
-            <LinkVercelProjectPage type="edit" />
-          </EnsureOnboardingComplete>
-        }
-      />
+      <Route element={<AuthLayout />}>
```
rifont: Wrap all of these paths with `AuthLayout` with an `<Outlet/>` added to `AuthLayout` to render children. We now remove `AuthLayout` being duplicated in each of these child routes.

apps/web/src/Providers.tsx
```diff
@@ -60,7 +60,7 @@ const Providers: React.FC<PropsWithChildren<{}>> = ({ children }) => {
           </BrowserRouter>
         </QueryClientProvider>
       </SegmentProvider>
-    </ColorSchemeProvider>
+    </ThemeProvider>
```
rifont, line 60: ThemeProvider now lives at the top of the tree.

apps/web/src/components/layout/components/AuthLayout.tsx
```diff
@@ -1,52 +1,49 @@
...
+              display: 'none',
+            },
+          }}
+        >
           <img
-            src={CONTEXT_PATH + '/static/images/logo-formerly-dark-bg.png'}
+            src={CONTEXT_PATH + '/static/images/notifications/notification_01.png'}
             alt="logo"
-            style={{ alignSelf: 'flex-start', maxWidth: 150, marginTop: 5, marginLeft: 5 }}
+            style={{ maxWidth: 400 }}
```
rifont: Bad diff, there's no changes here.

libs/novui/package.json
```diff
@@ -80,7 +80,7 @@
     "@testing-library/react-hooks": "^8.0.1",
     "@testing-library/user-event": "^14.5.2",
     "@types/node": "^18.11.12",
-    "@types/react": "^17.0.1",
+    "@types/react": "^17.0.0",
```
rifont, line 83: Fix for consistency.

apps/web/src/AppRoutes.tsx
SokratisVidros: How about we call the `AuthLayout`, `PublicPageLayout` to make it more clear as Auth layout doesn't tell us if this is for authenticated or unauthenticated context.

apps/web/src/Providers.tsx
SokratisVidros: 💯

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -56,7 +55,7 @@ export default function InvitationPage() {
   }, [queryClient]);
 
   return (
-    <AuthLayout>
+    <>
```
SokratisVidros: This fragment can be removed. We can return AuthContainer or null.

apps/web/src/pages/auth/PasswordResetPage.tsx
```diff
@@ -24,7 +22,7 @@ export function PasswordResetPage({}: Props) {
   }
 
   return (
-    <AuthLayout>
+    <>
```
SokratisVidros: Ditto

SokratisVidros (approved):
💯

apps/web/src/AppRoutes.tsx
rifont: Great suggestion. I'll change this.

PR #5625, opened 2024-05-22
feat(echo): sync version with package json

djabarovgeorge:
What?

We are currently sending a hardcoded 0.0.1 version, we need to update the implementation to read from the package.json version instead.

Why? (Context)

We need to share the version in use to alert Developers when they are using an outdated version of the Echo SDK

Definition of Done

Health check endpoint returns a version that is dynamically populated from the package.json version.

### Screenshots

<details>

</details>


rifont (approved):
Let's gooo!

PR #5626, opened 2024-05-22
chore(web): Refactor auth logic

SokratisVidros:

1. Introduce useAuth() to simplify auth state management for the web-app. This work is necessary to support multiple auth providers for SSO features.
2. Apply a minor refactoring to LaunchDarkly handling, aiming for a simple implementation in the next commit.


apps/web/src/hooks/useBlueprint.ts
```diff
@@ -12,10 +12,10 @@ export const useBlueprint = () => {
   const { pathname } = useLocation();
   const segment = useSegment();
   const id = localStorage.getItem('blueprintId');
-  const { token } = useAuthController();
+  const { token } = useAuth();
 
   useEffect(() => {
-    if (id && !!token) {
+    if (id) {
```
antonjoel82, line 18: ❓ **question:** Should we remove `token` from this altogether?‏

apps/web/src/pages/auth/InvitationPage.tsx
```diff
@@ -8,15 +8,16 @@ import { getInviteTokenData } from '../../api/invitation';
...
 import { useAcceptInvite } from './components/useAcceptInvite';
 import { LoginForm } from './components/LoginForm';
 
 export default function InvitationPage() {
   const queryClient = useQueryClient();
   const navigate = useNavigate();
-  const { token, logout, currentUser } = useAuthContext();
+  const { token, logout, currentUser } = useAuth();
   const location = useLocation();
+  // TODO: Replace token check with currentUser check
```
antonjoel82, line 20: ❓ **question:** What does this mean? Could we do this now?‏

apps/web/src/pages/auth/components/SignUpForm.tsx
```diff
@@ -62,15 +62,15 @@ export function SignUpForm({ invitationToken, email }: SignUpFormProps) {
      * To get the correct token when sending the request
      */
     const token = (response as any).token;
-    applyToken(token);
+    // login(token);
```
antonjoel82, line 65: ❓ **question:** Why is this commented? Do we need this?
‏

libs/shared/src/types/auth/auth.types.ts
```diff
@@ -10,6 +10,9 @@ export interface IJwtPayload {
   exp: number;
 }
 
+// @deprecated Use IJwtClaims instead
```
antonjoel82, line 13:
```suggestion
/**
* @deprecated Use IJwtClaims instead
*/
```

antonjoel82: ❓ **question:** Why this naming change?‏

libs/shared-web/src/hooks/useFeatureFlags.ts
antonjoel82, line 1: 🤓 **nitpick (non-blocking):** Should rename this file to `useFeatureFlag.ts` then‏

libs/shared-web/src/hooks/useEnvController.ts
```diff
@@ -64,19 +64,20 @@ export const useEnvController = (
       }
 
       setIsLoading(true);
+      // TODO: Do we need to get the token from this endpoint?
```
antonjoel82, line 67: 🗒 **note (non-blocking):** This seems necessary based on Richard mentioning that we extract `environment` and `orgId` from the token‏

apps/web/src/Providers.tsx
```diff
@@ -26,7 +24,7 @@ const queryClient = new QueryClient({
 });
 
 /** Full-page loader that uses color-preferences for background */
-const fallbackDisplay = (
+const Fallback = (
```
antonjoel82, line 27: ❓ **question:** Can we delete this then?‏

libs/shared-web/src/constants/routes.enum.ts
```diff
@@ -47,3 +47,10 @@ export enum ROUTES {
   SECURITY = '/settings/security',
   BILLING = '/settings/billing',
 }
+
+export const PUBLIC_ROUTES = [
```
antonjoel82:
```suggestion
export const PUBLIC_ROUTES = new Set<ROUTES>([
```

apps/web/src/Gate.tsx
```diff
@@ -0,0 +1,41 @@
+import { type ReactNode, useEffect, useLayoutEffect } from 'react';
+import { useLDClient } from 'launchdarkly-react-client-sdk';
+import { useAuth } from '@novu/shared-web';
+
+export function Gate({ children }: { children: ReactNode }) {
```
antonjoel82: ✏ **suggestion:** Rename more semantically and add comment about the purpose fo the function‏

antonjoel82:
```suggestion
export function Gate({ children }: PropsWithChildren<{}>) {
```

apps/web/src/Gate.tsx
```diff
@@ -0,0 +1,41 @@
...
+        anonymous: true,
+      });
+    }
+  }, [ldClient, currentOrganization]);
+
+  useLayoutEffect(() => {
+    if (inPublicRoute || !isLoading) {
+      document.getElementById('loader')?.remove();
+    }
+  }, [inPublicRoute, isLoading]);
```
antonjoel82: ✏ **suggestion:** IMO should be a separate hook‏

apps/web/public/index.html
```diff
@@ -42,6 +42,27 @@
...
+          background-color: #EDF0F2;
+        }
+      }
+      #loader {
+          display: grid;
+          width: 100vw;
+          height: 100vh;
+          place-items: center;
+        }
+    </style>
```
antonjoel82: ✏ **suggestion:** Prefer in different file for keeping this top-level file cleaner‏

apps/web/public/index.html
```diff
@@ -42,6 +42,27 @@
...
+        }
+
+        body {
+          background-color: #EDF0F2;
+        }
+      }
+      #loader {
+          display: grid;
+          width: 100vw;
+          height: 100vh;
```
antonjoel82:
```suggestion
          width: 100dvw;
          height: 100dvh;
```

apps/web/public/index.html
```diff
@@ -42,6 +42,27 @@
...
+      @layer critical {
+        /* Override Mantine injected CSS layers to prevent minor jumps */
+        *,::before,::after,::backdrop,::file-selector-button {
+          margin: 0px;
+          padding: 0px;
+          box-sizing: border-box;
+        }
+
+        body {
+          background-color: #EDF0F2;
```
antonjoel82:
```suggestion
          /* surface.page (light mode) */
          background-color: #EDF0F2;
```

apps/web/src/index.tsx
```diff
@@ -1,19 +1,45 @@
...
+        kind: 'user',
+        anonymous: true,
+      },
+      options: {
+        // eslint-disable-next-line max-len
+        // https://docs.launchdarkly.com/sdk/features/bootstrapping#:~:text=Alternatively%2C%20you%20can%20bootstrap%20feature%20flags%20from%20local%20storage%3A
+        bootstrap: 'localStorage',
+      },
+    });
+  }
```
antonjoel82, line 32: ✏ **suggestion:** Prefer this extracted to an aptly named file‏

libs/shared/src/types/auth/auth.types.ts
SokratisVidros: It's a standard naming convention when it comes to JWTs https://www.iana.org/assignments/jwt/jwt.xhtml#claims

libs/shared-web/src/constants/routes.enum.ts
SokratisVidros: Done.

apps/web/src/Providers.tsx
SokratisVidros: Done

PR #5630, opened 2024-05-24
fix(web): Build error correction for `useAuth`

antonjoel82:

- Changed to accommodate `Set` instead of array


djabarovgeorge (approved):
LGTM!

djabarovgeorge (approved):
LGTM!

PR #5633, opened 2024-05-27
ci(root): Run CSpell only on community PRs

Cliftonz:


AliaksandrRyzhou (approved):

scopsy:
@Cliftonz why did we disabled this? Will we still be running cspell for our team members somehow?

The problem might happen when we introduce a spelling mistake, and than a community member PR will be broken because of our mistake.

Cliftonz:
@scopsy This is defined [here](https://www.notion.so/Inner-DX-action-plan-ae5b1c49effe471293212fdd461c62cd?pvs=4#ef2a9bc5e9d4433ba7c7944059cbcc28) in the inner dx project. This was to remove blockers on us. 

I will be moving cspell to the web app only going forward.

PR #5636, opened 2024-05-27
chore(web): More auth refactoring

SokratisVidros:
- Simplify useAuth storage and only use local storage
- Load useBlueprint only in one place


apps/web/src/components/utils/Spotlight.tsx
```diff
@@ -91,16 +89,8 @@ export const SpotLight = ({ children }) => {
           toggleColorScheme();
         },
       },
-      {
```
SokratisVidros, line 94: This removes logout from ctrl+K to prevent large number of rerenders.

rifont (approved):
Let's go!

PR #5651, opened 2024-05-30
feat(novui,web): Setup Mantine v7 and demonstrate ability to style in lib and app

antonjoel82:

- Install and setup Mantine v7
- Add a basic wrapper component and utility type

### Screenshots

Gorgeous, definitely not temporary example of:

- Mantine component in `novui` styled with Panda
- Style extended in web via Panda

Screenshot 2024-05-30 at 11 14 47 AM


apps/web/src/pages/settings/SettingsPage.tsx
```diff
@@ -54,6 +56,11 @@ export function SettingsPage() {
 
   return (
     <SettingsPageWrapper>
+      <Test
```
antonjoel82: ☑ **chore:** Remove before merge‏

SokratisVidros (approved):
🎨

libs/novui/src/types/CoreProps.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Defines the foundational props from which all Novu components should inherit.
+ */
+export interface CoreProps {
+  className?: string;
+  'data-test-id'?: string;
```
SokratisVidros: I'd like to suggest removing `data-test-id` and promote accessible selectors. Mantine v7 is stronger in that compared to v5. The team should strive to eliminate the need for data-test-id.

libs/novui/postcss.config.cjs
```diff
@@ -1,7 +1,20 @@
...
+   * Must use array + require syntax for vite + postcss & panda
+   * https://vitejs.dev/config/shared-options.html#css-postcss
    * https://panda-css.com/docs/installation/storybook#install-panda
    */
-  plugins: [require('@pandacss/dev/postcss')()],
+  plugins: [
+    require('@pandacss/dev/postcss')(),
+    require('postcss-preset-mantine')(),
+    require('postcss-simple-vars')({
+      variables: {
```
SokratisVidros, line 11: @antonjoel82 can you please elaborate why these variables are not part of Panda and are set separately on mantine?

libs/novui/src/types/CoreProps.ts
antonjoel82: Sounds good to me -- I have not been a fan of our use of it either

libs/novui/postcss.config.cjs
antonjoel82: These were part of the [Mantine setup](https://mantine.dev/getting-started/#:~:text=Create%20postcss.config.cjs%20file%20at%20the%20root%20of%20your%20application%20with%20the%20following%20content%3A). Given that we haven't defined our own in our Design System and to avoid any slowdowns in the event that these are somehow used in the internals of v7, I elected to keep them. We can always remove them down the line.

SokratisVidros: Sounds reasonable. Thanks for explaining!

PR #5653, opened 2024-05-31
chore(root): Bump pnpm to its latest version

SokratisVidros:

This is a small quality of life change for contributors. 

Since the introduction of the [conventional commit PR title Github](https://github.com/novuhq/novu/blob/14b6e3bc6528428457e28d9fb2b5a7b8b9fc218f/.github/workflows/conventional-commit.yml) action and the introduction of squash-merge in our main repos, there is no need to enforce conventional commits in our every day commits before pushing them to the branch.

This simplifies the monorepo and gives a small speed boost to the work done by husky locally on each commit.


rifont (approved):
🧹

libs/embed/tools/semantic-release-prepare.ts
```diff
@@ -8,7 +8,6 @@ const pkg = JSON.parse(
 )
 
 pkg.scripts.prepush = "pnpm test:prod && pnpm build"
```
antonjoel82, line 10: ☑ **chore:** Well this is annoying... but please add to `cspell` or ignore above to fix this check‏

antonjoel82 (approved):
🙏

PR #5654, opened 2024-05-31
feat(js): the base js sdk package scaffolding

LetItRock:
The base JS SDK package configuration and classes:
- builds to ESM, CJS, UMD
- the base classes: Novu, NovuEventEmitter, Session, Feeds, Preferences (implementation in the upcoming PRs)
- the package is private for now - won't be pushed to NPM

### Screenshots

Screenshot 2024-05-31 at 16 48 24

<details>

</details>


apps/api/src/app/widgets/dtos/message-response.dto.ts
```diff
@@ -85,7 +86,7 @@ class MessageCTA {
 }
 
 @ApiExtraModels(EmailBlock, MessageCTA)
-export class MessageResponseDto {
+export class MessageResponseDto implements INotificationDto {
```
LetItRock, line 89: created the new `INotificationDto` interface that will be used in the `@novu/js`

libs/shared/src/dto/subscriber/subscriber.dto.ts
```diff
@@ -26,3 +32,23 @@ export class SubscriberDto {
   channels?: IChannelSettings[];
   deleted: boolean;
 }
+
+export interface ISubscriberResponseDto {
```
LetItRock, line 36: used by `INotificationDto`

libs/shared/src/dto/widget/notification.dto.ts
```diff
@@ -0,0 +1,64 @@
...
+  url?: string;
+}
+
+interface IMessageCTA {
+  type?: ChannelCTATypeEnum;
+  data: IMessageCTAData;
+  action?: IMessageAction;
+}
+
+export interface INotificationDto {
```
LetItRock, line 33: the new Notification DTO interface

packages/js/.eslintrc.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
+  extends: ['../../.eslintrc.js'],
+  rules: {
+    '@typescript-eslint/explicit-module-boundary-types': 'error',
```
LetItRock: no explicit types, all should be defined

packages/js/README.md
```diff
@@ -0,0 +1,4 @@
+# Novu's JavaScript SDK
+
+The `@novu/js` package provides a JavaScript SDK for building custom inbox notification experiences.
+The package provides a low-level API for interacting with the Novu platform In-App notifications.
```
LetItRock, line 4: we will improve it later

packages/js/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@novu/js",
+  "version": "0.24.2",
+  "repository": "https://github.com/novuhq/novu",
+  "description": "Novu's JavaScript SDK for building custom inbox notification experiences",
+  "author": "",
+  "license": "ISC",
+  "main": "dist/index.js",
```
LetItRock: cjs

packages/js/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "module": "dist/index.mjs",
```
LetItRock: esm

packages/js/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "files": [
+    "dist/*"
+  ],
+  "sideEffects": false,
+  "private": true,
```
LetItRock, line 29: won't be published to npm for now

packages/js/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "types": "dist/index.d.ts",
```
LetItRock: TS types

packages/js/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "@types/node": "^18.11.12",
+    "jest": "^29.3.1",
+    "ts-jest": "^29.0.3",
+    "tsup": "^8.0.2",
+    "typescript": "4.9.5"
+  },
+  "dependencies": {
+    "@novu/client": "workspace:*",
+    "@novu/shared": "workspace:*",
+    "mitt": "^3.0.1"
```
LetItRock, line 54: event emitter library that is tiny ~200 bytes

packages/js/src/Novu.ts
```diff
@@ -0,0 +1,39 @@
...
+
+interface NovuOptions {
+  applicationIdentifier: string;
+  subscriberId: string;
+  subscriberHash?: string;
+}
+
+export class Novu implements Pick<NovuEventEmitter, 'on' | 'off'> {
+  #emitter: NovuEventEmitter;
+  #session: Session;
```
LetItRock, line 15: The "private properties", we are not using the `private` TS fields because after the transpilation the fields would be publicly accessible in JS code (mainly in the UMD).

More about it here: https://stackoverflow.com/questions/59641564/what-are-the-differences-between-the-private-keyword-and-private-fields-in-types/59641565#59641565

packages/js/src/event-emitter/NovuEventEmitter.ts
```diff
@@ -0,0 +1,23 @@
+import mitt, { Emitter } from 'mitt';
+
+import { Events, EventHandler, EventNames } from './types';
+
+export class NovuEventEmitter {
```
LetItRock, line 5: our own wrapper over the emitter

packages/js/src/event-emitter/types.ts
```diff
@@ -0,0 +1,14 @@
+import { INotificationDto } from '@novu/shared';
+
+import { PaginatedResponse } from '../types';
+
+export type Events = {
+  'feeds.fetch.*': undefined;
+  'feeds.fetch.pending': undefined;
+  'feeds.fetch.success': { res: PaginatedResponse<INotificationDto> };
+  'feeds.fetch.error': { error: unknown };
+};
```
LetItRock, line 10: created just for the sake of checking the types, we will revisit it later

packages/js/tsup.config.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const baseConfig: Options = {
+  splitting: false,
+  sourcemap: false,
+  clean: true,
+  dts: true,
+};
+
+export default defineConfig([
+  { ...baseConfig, entry: ['src/index.ts'], format: ['esm', 'cjs'] },
```
LetItRock: the TSUP config to build the ESM, CJS

packages/js/tsup.config.ts
```diff
@@ -0,0 +1,24 @@
...
+    ...baseConfig,
+    entry: { novu: 'src/umd.ts' },
+    format: ['iife'],
+    minify: true,
+    dts: false,
+    outExtension: () => {
+      return {
+        js: '.min.js',
+      };
+    },
```
LetItRock, line 36: The bundled and minified JS file `novu.min.js` with IIFE that exposes the `Novu` to the `window` object

libs/shared/src/dto/widget/notification.dto.ts
SokratisVidros: @LetItRock I am skeptical whether we need separate notification definitions for the frontend and the backend. Most of the fields in this interface are 100% relevant in the server but seem a DX overkill for the new, frontend, JS SDK.

Is this meant to be reused?

Sidenote: we should stop using the `I` prefix for interfaces and switch to `types` as the `I` interface gives a more old-school DX touch to the SDK.

packages/js/package.json
SokratisVidros: I'd suggest using conditional exports instead of `main`, `module`, `type`. They provide more control, and they have become the [new defacto standard](https://stackoverflow.com/questions/68572936/what-is-the-difference-between-main-and-module-vs-exports-in-package-json). 

The assumption is that we will also include them to re-export the JS UI part.

packages/js/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "@types/jest": "^29.2.3",
+    "@types/node": "^18.11.12",
+    "jest": "^29.3.1",
+    "ts-jest": "^29.0.3",
+    "tsup": "^8.0.2",
+    "typescript": "4.9.5"
+  },
+  "dependencies": {
+    "@novu/client": "workspace:*",
+    "@novu/shared": "workspace:*",
```
SokratisVidros, line 53: This feels dangerous dependency. Is @novu/shared treeshable?

 Like I said before, starting from a clean slate and defining the new JS SDK with only the necessary, frontend fields. This exercise will also dictate the necessary Novu API changes to make it leaner.
 
 I am happy to discuss this offline on Monday.

packages/js/src/Novu.ts
```diff
@@ -0,0 +1,39 @@
+import { NovuEventEmitter } from './event-emitter';
+import type { EventHandler, EventNames, Events } from './event-emitter';
+import { Feeds } from './feeds';
+import { Session } from './session';
+import { Preferences } from './preferences';
+
+interface NovuOptions {
```
SokratisVidros, line 7:
```suggestion
type NovuOptions {
```

packages/js/src/Novu.ts
```diff
@@ -0,0 +1,39 @@
...
+  applicationIdentifier: string;
```
SokratisVidros, line 8: This is a good opportunity for a DX improvement. We can encode the `applicationId` in the API key and remove one option.

packages/js/src/Novu.ts
```diff
@@ -0,0 +1,39 @@
...
+export class Novu implements Pick<NovuEventEmitter, 'on' | 'off'> {
+  #emitter: NovuEventEmitter;
+  #session: Session;
+
+  public readonly feeds: Feeds;
+  public readonly preferences: Preferences;
+
+  constructor(options: NovuOptions) {
+    this.#emitter = new NovuEventEmitter();
+    this.#session = new Session(this.#emitter, {
```
SokratisVidros, line 22:
```suggestion
    this.#session = new Session(this.#emitter, options)
```

packages/js/src/event-emitter/NovuEventEmitter.ts
SokratisVidros: Is this wrapper required for type safety? Can we ensure type safety without it?

packages/js/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "target": "ESNext",
+    "module": "ESNext",
+    "strict": true,
+    "skipLibCheck": true,
+    "typeRoots": ["./node_modules/@types"],
+    "sourceMap": true,
+    "declaration": false,
```
SokratisVidros:
```suggestion
    "declaration": true,
```

packages/js/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "declarationMap": false,
```
SokratisVidros:
```suggestion
    "declarationMap": true,
```

packages/js/tsconfig.json
SokratisVidros: Types need to be generated for DX.

packages/js/tsup.config.ts
SokratisVidros, line 36: Let's also add the following:

// @ts-ignore
import { name, version } from './package.json';

1. Trick to inject version and name in the source code without importing the full package.json
```
  define: {
      PACKAGE_NAME: `"${name}"`,
      PACKAGE_VERSION: `"${version}"`,
      __DEV__: `${!isProd}`,
    },
```

2. To fully support `ESM/CJS` we need to copy a package.json with the necessary `ESM/CJS` attributes into each dist folder [See example](https://github.com/clerk/javascript/blob/main/packages/nextjs/tsup.config.ts#L40).

SokratisVidros (changes requested):
📥

libs/shared/src/dto/widget/notification.dto.ts
LetItRock: > Is this meant to be reused?

yes, this is meant to be reused for the new SDK, and it's the first step to clear the interface and fields... we will make the improvements to it and reduce the fields amount when we refactor the widget API's...

> Sidenote: we should stop using the I prefix for interfaces and switch to types as the I interface gives a more old-school DX touch to the SDK.

yes I agree, I can change that... but I would say we should discuss that as the team

packages/js/src/event-emitter/NovuEventEmitter.ts
LetItRock: it's just the wrapper for us to be able to switch to any other implementation

packages/js/package.json
LetItRock: yes it should be... and we only use types with `import type` statement
agree, ideally we want to have a shared interface between FE and BE

packages/js/src/Novu.ts
LetItRock: yes, we can, but this requires many other changes... let's leave it for now and improve it later

packages/js/src/Novu.ts
LetItRock: there will be more properties and session doesn't need to know about them

packages/js/tsconfig.json
LetItRock: the types are generated by tsup ;) but this option is misleading

packages/js/tsconfig.json
LetItRock: same... but let's discuss it on Monday

packages/js/tsup.config.ts
LetItRock: 1. Ok
2. Not sure I understand why we need this, and we have all esm/cjs files in one dist folder

libs/shared/src/dto/widget/notification.dto.ts
SokratisVidros: We resolved this offline: 
1. We won't be using any `I` naming moving forward as part of the public API
2. We will reuse the DTOs internally for now to drive the new public API classes and we will note the fields we need from those DTOs so as to update the frontend API endpoints at a later stage.

packages/js/tsup.config.ts
SokratisVidros: 💯

SokratisVidros:
@LetItRock let's add the check on bundle size as part of the build process and set it to 10KB for now.

SokratisVidros (approved):

SokratisVidros (approved):

PR #5659, opened 2024-05-31
build(root): Update Jarvis to run docker infra instead of through brew

Cliftonz:


scripts/dev-environment-setup.sh
```diff
@@ -383,43 +389,29 @@ install_aws_cli () {
     fi
 }
 
-install_databases () {
-    SKIP="$(check_homebrew)"
+start_database() {
+  # Check if mongodb is installed
+  brew ls --versions mongodb > /dev/null
```
rifont: ❓ **question:** What is the behaviour when `brew` is not installed on the developer's system? I think we should handle cases where it is not in a silent way.‏

scripts/dev-environment-setup.sh
```diff
@@ -383,43 +389,29 @@ install_aws_cli () {
...
+  if [ $? -eq 0 ]; then
+    echo "Error: MongoDB is already installed via brew. Please uninstall it first."
+    exit 1
```
rifont: ⚠ **issue:** I think throwing an error here could be a bit heavy-handed. I think there are valid cases for having Mongo/Redis installed in Brew for use outside of the Novu project. Perhaps we can log a warning here instead?‏

rifont (commented):
When running `pnpm dev-environment-setup`, I am getting a prompt to enter my password after the Redis installation check occurs, but I'm not made aware of why I need to enter my password. Is there a step in the setup sequence that we either need to:
1. Remove to prevent asking for password
2. Add an addition `echo` to give info to the User on _why_ they need to enter their password

```bash
# ...
✅ NVM has been installed
 
 
✅ Node.js v20.8.1 is already installed
 
 
✅ PNPM 8.9.0 is already installed
 
 
✅ Docker is already installed
 
 
✅ AWS CLI is already installed
 
Warning: Redis is already installed via brew. Please uninstall it first.
Password:
```

rifont (commented):
⚡🚀👨‍💻

rifont (approved):
⚡🚀👨‍💻

PR #5661, opened 2024-06-01
fix(node): Fixing chat missing in events interface

ameensom:
What changed? Why was the change needed?
Included 'chat' into overrides trigger enum, as it's mentioned in the docs for sending whatsapp business but shows an error

When using novu/node package and using method trigger, the overrides object does not recognize child object 'chat' and returns type error, and it appears that it's not defined in the events types interface.


packages/node/src/lib/events/events.interface.ts
```diff
@@ -49,6 +49,7 @@ export type ITriggerOverrides = {
     | 'plivo'
     | 'postmark'
     | 'sendgrid'
+    | 'chat'
```
rifont: We are seeking to introduce stronger typing across out SDKs. This fallback into `object` creates a weak typing for `chat`.

Can we follow a similar pattern as `email` below, creating an `IChatOverrides` interface and listing the necessary properties to expose in a strongly typed way?

rifont (commented):
Thanks for highlighting this missing override. I added a comment to address @ameensom 🙏

packages/node/src/lib/events/events.interface.ts
ameensom: totally agree,
i've pushed a minor change that adds the expected keys inside template object, however I think we need to add more types, hence we have media sending and interactive messages, it does have lots of variety in it, i'll try to add more soon,
Let me know if the last change is good so I can move on with the other message types

rifont (approved):
Excellent @ameensom , this is exactly what we're after. Merging now.

PR #5662, opened 2024-06-02
refactor(root): Nv 3846 consolidate env file behavior

denis-kralj-novu:
web and widget were the 2 apps that had their `.env` file included in the repository.  api, ws and worker have a procedure in place for how to setup the initial `.env` state via setup script, so this change aims to consolidate the behavior and make it as similar as possible without affecting local development or deployed behavior


.gitignore
```diff
@@ -97,7 +97,7 @@ fabric.properties
 # Module directory
 .terraform/
 
-.env
+**/.env
```
denis-kralj-novu, line 100: All of the removal in other .gitignore files are related to consolidation done here on the top level

apps/web/.example.env
denis-kralj-novu, line 1: shows up like a move since the content is the same

apps/widget/.example.env
denis-kralj-novu, line 1: shows up like a move since the content is the same

scripts/setup-env-files.js
```diff
@@ -15,14 +16,10 @@ const prePopulateEnv = (folders, folderBasePath, exampleEnvFilePath = 'src/.exam
...
-  const apps = ['api', 'ws', 'worker'];
   const appsBasePath = `${__dirname}/../apps`;
-
   console.log('----------------------------------------');
-  console.log('Pre-populating .env files from .example.env');
-
-  prePopulateEnv(apps, appsBasePath);
-
+  prePopulateEnv(['api', 'ws', 'worker'], appsBasePath);
+  prePopulateEnv(['web', 'widget'], appsBasePath, '.example.env', '.env');
```
denis-kralj-novu, line 22: web and widget don't have their `.env` files in `/src` but in the parent directory, so this adjustment is needed

.gitignore
rifont: 🧹 nice one!

rifont (commented):
Looks great, nice tidy up and move to consistency.

@LetItRock , do you see any issues with this simplification? The main point to verify is that the population to the `window` object with Web Env-vars is completed. Tagging you as you have familiarity with this config step.

scripts/setup-env-files.js
SokratisVidros: 😮‍💨

SokratisVidros (approved):

denis-kralj-novu:
@SokratisVidros once this is merged it will deploy to `dev.web.novu.co` correct? Do we have some way to deploy to a temp env to ensure I haven't missed anything in this env shuffling? I don't want to get into a situation where I break dev 😁

SokratisVidros:
@denis-kralj-novu You can use a Netlify preview deployment for Web app changes. We don't have preview deployment infra on AWS yet. So let's merge it but be ready to rollback [see relevant workflow](https://github.com/novuhq/novu/actions/workflows/rollback.yml).

denis-kralj-novu:
> @denis-kralj-novu You can use a Netlify preview deployment for Web app changes. We don't have preview deployment infra on AWS yet. So let's merge it but be ready to rollback [see relevant workflow](https://github.com/novuhq/novu/actions/workflows/rollback.yml).

I will merge first thing tomorrow morning then.

LetItRock:
> Looks great, nice tidy up and move to consistency.
> 
> @LetItRock , do you see any issues with this simplification? The main point to verify is that the population to the `window` object with Web Env-vars is completed. Tagging you as you have familiarity with this config step.

@rifont @denis-kralj-novu 
The `.env` file for the Web application is created during the cloud deployment to prod or dev env: https://github.com/novuhq/novu/blob/e246a7539f8487d1b873597b410a7c50806981c7/.github/workflows/reusable-web-deploy.yml#L75

Then we create the `env-config.js` based on `.env` in the `public` folder which is imported as a script by `index.html`.

A similar process we do for the docker images, but we read from the local environment variables.

---

During the local development, the fallback values are used as the configuration: https://github.com/novuhq/novu/blob/e246a7539f8487d1b873597b410a7c50806981c7/libs/shared-web/src/config.ts#L18
For the future: if we would like to get rid of fallback values we need to make changes in the `config.ts` and fill the `.env` file.

---

But overall, it looks good to me 😄 😛

LetItRock (approved):

PR #5664, opened 2024-06-03
feat(root): NV-3847 include create-novu-app to monorepo

denis-kralj-novu:
- Include create-novu-app command into the packages directory (gotten from [here](https://github.com/novuhq/create-echo-app))
- change legacy name to `create-novu-app`
- adjust readme
- include a bunch of `/* eslint-disable */` to enable commit, should be addressed as developers work on the files


packages/create-novu-app/create-app.ts
```diff
@@ -0,0 +1,231 @@
+/* eslint-disable */
```
denis-kralj-novu, line 1: In general I needed to add this in a lot of places to allow the commit to pass, they should be removed as devs work on given files

packages/create-novu-app/README.md
```diff
@@ -0,0 +1,7 @@
+# Create Novu App
```
denis-kralj-novu, line 1: Minimal readme that isn't the out of the box next.js variant

packages/create-novu-app/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "create-novu-app",
+  "version": "0.0.1-alpha.4",
```
denis-kralj-novu: We should probably agree on what the correct value is to put into the version

packages/create-novu-app/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "create-novu-app",
```
denis-kralj-novu, line 2: I've exchanged instances of "echo" for "novu" in this file

packages/create-novu-app/helpers/examples.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      const info = await infoResponse.json();
+      return { username, name, branch: info['default_branch'], filePath };
+    } catch {
+      return;
+    }
+  }
+
+  // If examplePath is available, the branch name takes the entire path
+  const branch = examplePath ? `${_branch}/${file.join('/')}`.replace(new RegExp(`/${filePath}|/$`), '') : _branch;
```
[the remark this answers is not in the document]
denis-kralj-novu: I wouldn't say this is a consideration for us, I don't even see how this parameter would be sent from the instructions. Even if it is, I'm unsure how a malicious actor would exploit this, as it will only ask GitHub for file info on something that doesn't exist.

Feedback anyone? I might not be seeing the full picture ☹️

packages/create-novu-app/package.json
rifont: Let's go with the same version as other monorepo packages for now.

```suggestion
  "version": "0.24.2",
```

rifont (approved):
Awesome! Just a few minor details to address

packages/create-novu-app/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "prompts": "2.4.2",
+    "tar": "6.1.15",
+    "update-check": "1.5.4",
+    "validate-npm-package-name": "3.0.0",
+    "typescript": "5.2.2",
+    "async-sema": "3.0.1"
+  },
+  "engines": {
+    "node": ">=18.17.0"
+  }
```
rifont: No need for engines here, it's added at the monorepo root

```suggestion
```

denis-kralj-novu: done

packages/create-novu-app/package.json
denis-kralj-novu: done

PR #5668, opened 2024-06-03
fix(web): add redirect to page after login

jainpawan21:
1. added logger in case of hubspot failed to add email to list
2. add `_organizationId: subscriberSession._organizationId,` for widgets analytics event
3. redirect to required page after login

Ref:- [NV-3697](https://linear.app/novu/issue/NV-3697/invite-team-member-nudge)

### Screenshots

<details>

</details>


apps/api/src/app/widgets/widgets.controller.ts
```diff
@@ -476,6 +476,7 @@ export class WidgetsController {
   ): Promise<LogUsageResponseDto> {
     this.analyticsService.track(body.name, subscriberSession._organizationId, {
       environmentId: subscriberSession._environmentId,
+      _organizationId: subscriberSession._organizationId,
```
jainpawan21: doubt: it should be `_organizationId` or `organization` ?

apps/api/src/app/widgets/widgets.controller.ts
```diff
@@ -476,6 +476,7 @@ export class WidgetsController {
   ): Promise<LogUsageResponseDto> {
     this.analyticsService.track(body.name, subscriberSession._organizationId, {
```
jainpawan21, line 477: suggestion : `subscriberSession._organizationId` should be changed with `subscriberSession._userId` as per `track` method params

apps/api/src/app/widgets/widgets.controller.ts
LetItRock: I see that we are not consistent with the pattern, but let's say do `_organization` as I see it's used in payments as well which is pretty new code :D

apps/api/src/app/widgets/widgets.controller.ts
rifont: Nice catch

rifont (approved):

PR #5670, opened 2024-06-04
feat(echo): add regular delay step feature parity

djabarovgeorge:
Using the DX guide, add support to the Echo Resonate use-case for all features supported in Novu Cloud workflow editor to bring the Echo Regular Delay experience to feature parity.

We must establish the DX guide before we implement the feature, to ensure that all features are accounted for.

### Definition of Done

Delay Step DX guide completed and peer reviewed
### Screenshots

<details>

</details>


libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
djabarovgeorge, line 1: Renamed the CalculateDelayService to ComputeJobWaitDurationService to be more explicit about the responsibility of this service, which is calculating the time we will postpone the job execution for both Delay and Digest

packages/echo/src/schemas/steps/actions/delay.schema.ts
djabarovgeorge, line 1: The main change in order to fully support regular delay, we need to add the type on out output

packages/echo/src/client.test.ts
```diff
@@ -356,24 +358,52 @@ describe('Echo Client', () => {
...
+          },
+        ],
+        inputs: {},
+      };
+
+      const delayExecutionResult = await echo.executeWorkflow(delayEvent);
+
+      expect(delayExecutionResult).toBeDefined();
+      expect(delayExecutionResult.outputs).toBeDefined();
+      if (!delayExecutionResult.outputs) throw new Error('executionResult.outputs is undefined');
```
denis-kralj-novu, line 407: ✏ **suggestion (non-blocking):** ‏ I would imagine only one of the 2 lines is required. I see this is the same as for the email step, if we choose to consolidate to one line I would do it in both places

packages/echo/src/client.test.ts
```diff
@@ -344,9 +345,10 @@ describe('Echo Client', () => {
     it('should execute workflow successfully when action is execute and data is provided', async () => {
       await echo.workflow('test-workflow', async ({ step }) => {
         await step.email('send-email', async () => ({ body: 'Test Body', subject: 'Subject' }));
+        await step.delay('delay', async () => ({ type: 'regular', unit: 'seconds', amount: 1 }));
```
denis-kralj-novu: 💭 **thought (non-blocking):** ‏ maybe pull out the configuration of the delay step to a const above so that it can be used in asserts below

denis-kralj-novu (approved):

packages/echo/src/client.test.ts
djabarovgeorge: True only one is required the `if` statement is just a trick to help TypeScript understand that the variable has a value and won't be `null` later on. This lets us avoid adding question marks all over the code, making it cleaner to read (at least in my opinion).

packages/echo/src/client.test.ts
djabarovgeorge: done 🔥

packages/echo/src/client.test.ts
denis-kralj-novu: fair enough :)

SokratisVidros (approved):

scopsy:
Found a very strange behaviour when an email is after a delay step in the studio, preview is broken. Need to investigate

scopsy:
Okey, really weird it's also happening on `next` branch. Only happens when there are no inputSchema on the step above the one we want to preview. 

This will not work:
```
await step.delay('delay', async () {
  return {
    .....
  }
})

await step.email('delay', async () {
  return {
    .....
  }
}, {
  inputSchema: {
    ....
  }
})
```

While this will:
```
await step.delay('delay', async () {
  return {
    .....
  }
}, {
  inputSchema: {

  }
})

await step.email('delay', async () {
  return {
    .....
  }
}, {
  inputSchema: {
    ....
  }
})
```

scopsy (approved):

PR #5672, opened 2024-06-04
feat(echo): add scheduled delay step feature parity 

djabarovgeorge:
 
Using the guide from [NV-3802](https://linear.app/novu/issue/NV-3802/%E2%9A%A1-delay-step-dx-guide) , add support to the Echo Resonate use-case for all features supported in Novu Cloud workflow editor to bring the Echo Scheduled Delay experience to feature parity.

We need to support existing features in Echo

### Definition of Done

I can use Echo to perform a scheduled delay
### Screenshots

<details>

</details>


libs/application-generic/src/utils/require-inject.ts
```diff
@@ -47,18 +53,27 @@ enum RequireInjectEnum {
 
 export interface IChimeraDigestResponse {
   amount: number;
-  unit: 'seconds' | 'minutes' | 'hours' | 'days' | 'weeks' | 'months';
+  unit: DigestUnitEnum;
```
denis-kralj-novu, line 56: 👏 **praise:** ‏ great idea to use existing enum!

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -82,6 +95,29 @@ export class ComputeJobWaitDurationService {
...
+    }
+
+    return;
+  }
+
+  private getUserScheduledDelayMetadata(
+    chimeraResponse: IChimeraDigestResponse | IDelayOutput
+  ): { date: string } | undefined {
+    if (isUserScheduled(chimeraResponse)) {
+      return { date: chimeraResponse?.date };
```
denis-kralj-novu: 🤓 **nitpick (non-blocking):** ‏ since `isUserScheduled()` has a type predicate, the `?` operator isn't needed in the scope of the if, because the chimeraResponse will be treated as the proper type

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -82,6 +95,29 @@ export class ComputeJobWaitDurationService {
     return 0;
   }
 
+  private getUserRegularDelayMetadata(
+    chimeraResponse: IChimeraDigestResponse | IDelayOutput
+  ): { amount: number; unit: DigestUnitEnum } | undefined {
+    if (isUserRegular(chimeraResponse)) {
```
denis-kralj-novu: 🤓 **nitpick (non-blocking):** ‏ since `isUserRegular(`) has a type predicate, the `?` operator isn't needed in the scope of the if, because the chimeraResponse will be treated as the proper type

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -37,35 +39,46 @@ export class ComputeJobWaitDurationService {
...
-      const delay = differenceInMilliseconds(new Date(delayDate), new Date());
-
-      if (delay < 0) {
-        throw new ApiException(
-          `Delay date at path ${delayPath} must be a future date`
-        );
+      const userMetadata = this.getUserScheduledDelayMetadata(chimeraResponse);
+      let delay = 0;
+
+      if (userMetadata.date) {
```
denis-kralj-novu: ❓ **question:** ‏ will this work if the return of the `this.getUserScheduledDelayMetadata(chimeraResponse);` ends up being `undefined`?

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
```diff
@@ -389,6 +390,55 @@ describe('Echo Trigger ', async () => {
     expect(messagesAfter.length).to.be.eq(1);
     expect(messagesAfter[0].content).to.match(/people waited for \d+ seconds/);
   });
+
+  it('should trigger the scheduled delay with invalid date', async () => {
```
denis-kralj-novu: ✏ **suggestion:** ‏ Maybe a better name would be 'should report error on date that is in the past from now'

denis-kralj-novu (commented):
Let me know what you thing on my comments and I think we can get this approved quickly :)

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
djabarovgeorge: Good point for some reason ts is not able to infer the undefined value, any idea why?
I will add a question mark as an extra guard.

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
djabarovgeorge: love it ! :)

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
denis-kralj-novu: could it be a difference between `return;` and `return undefined;`?

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -14,8 +14,10 @@ import { ApiException } from '../../utils/exceptions';
 import { isRegularDigest } from '../../utils/digest';
 import { TimedDigestDelayService } from './timed-digest-delay.service';
 import {
-  IChimeraDelayResponse,
   IChimeraDigestResponse,
```
SokratisVidros: Let's remove the Chimera part in a separate PR as well. Right now, we have Chimera, Echo and it feels messy.

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -28,7 +30,7 @@ export class ComputeJobWaitDurationService {
     stepMetadata?: IWorkflowStepMetadata;
     payload: any;
     overrides: any;
-    chimeraResponse?: IChimeraDigestResponse | IChimeraDelayResponse;
+    chimeraResponse?: IChimeraDigestResponse | IDelayOutput;
```
SokratisVidros:
```suggestion
    response?: IDigestResponse | IDelayOutput;
```

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -82,6 +95,29 @@ export class ComputeJobWaitDurationService {
     return 0;
   }
 
+  private getUserRegularDelayMetadata(
+    chimeraResponse: IChimeraDigestResponse | IDelayOutput
```
SokratisVidros:
```suggestion
    response: IChimeraDigestResponse | IDelayOutput
```

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
djabarovgeorge: I think it could be because of  "strictNullChecks": false, in application-generic, i always forget we have it there under false. it will be amazing to switch it to true 😅

djabarovgeorge: George from the past created [this one](https://linear.app/novu/issue/NV-3701/application-generic-set-strictnullchecks-as-true)

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
djabarovgeorge: I think it was already addressed by Denis in another PR.

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
djabarovgeorge: Yeah i thought to rename it as part of the digest work 😅

djabarovgeorge: done here anyways :)

PR #5673, opened 2024-06-04
feat(web): Local sidebar

antonjoel82:

- Create the new top-level directory for Novu V2
- New sidebar for `Local` env (currently displays any time you visit `/studio/*` routes)
  - New version of the EnvironmentSelect
- Begin setup for eventual color theming

### Screenshots

Screenshot 2024-06-04 at 6 16 15 PM


ainouzgali (approved):
Looks good ⭐

apps/web/src/studio/components/EnvironmentSelect/EnvironmentPopover.styles.ts
ainouzgali, line 1: It seems you are not using these styles :) as you are using the EnvironmentPopover from the regular component folder

antonjoel82: Good catch, thanks! Removed.

PR #5674, opened 2024-06-04
fix(web): Hubspot onboarding invite member flow

jainpawan21:

1. Fix hubspot form field for invitation cases
2. Fix the code at few places to support these cases
Screenshot 2024-06-07 at 9 18 13 PM

### Screenshots

<details>

</details>


antonjoel82:
Hey @jainpawan21! Can you please fill-out the PR description with what the changes are and why they're necessary?

jainpawan21:
> Hey @jainpawan21! Can you please fill-out the PR description with what the changes are and why they're necessary?

Changed it to draft PR as I am working on some bugs

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -48,40 +51,59 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
     }
   >((data) => api.post('/v1/auth/login', data));
 
-  useEffect(() => {
-    (async () => {
-      if (!tokenInQuery) {
-        return;
-      }
+  const handleLoginInUseEffect = async () => {
+    // if currentUser is true, it means user exists, then while accepting invitation, InvitationPage will handle accept this case
```
SokratisVidros, line 55: This indicates a routing issue as the invitation handling should be done in one place, not between pages. It's OK for now but we will have to get back to this.

libs/application-generic/src/services/auth/shared.ts
```diff
@@ -48,6 +48,10 @@ export const buildOauthRedirectUrl = (request): string => {
   if (invitationToken) {
     url += `&invitationToken=${invitationToken}`;
   }
+  const isLoginPage = JSON.parse(request.query.state).isLoginPage;
```
SokratisVidros, line 51: Instead of marking whether OAuth comes from loginPage, we can add a `redirect_url` parameter set from the front end, which would simplify the FE logic.

apps/web/src/pages/auth/components/gitHubUtils.ts
```diff
@@ -2,12 +2,21 @@ import { SignUpOriginEnum } from '@novu/shared';
...
+}: {
+  invitationToken?: string;
+  isLoginPage?: boolean;
+}) => {
   const queryParams = new URLSearchParams();
   queryParams.append('source', SignUpOriginEnum.WEB);
   if (invitationToken) {
     queryParams.append('invitationToken', invitationToken);
   }
+  if (isLoginPage) {
```
SokratisVidros, line 17: Ditto about `redirect_url`

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -48,40 +51,59 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
...
+    if (currentUser) {
+      return;
+    }
 
-      if (!invitationToken && (!organizationId || !environmentId)) {
-        await login(tokenInQuery, ROUTES.AUTH_APPLICATION);
+    // if token from github is not present
```
SokratisVidros:
```suggestion
    // if token from OAuth or CLI is not present
```

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -48,40 +51,59 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
...
+    // handle github login after invitation
+    if (invitationTokenFromGithub) {
+      await login(tokenInQuery);
+      const updatedToken = await acceptInvite(invitationTokenFromGithub);
 
-      if (isFromVercel) {
-        await login(tokenInQuery);
-        startVercelSetup();
+      if (updatedToken) {
+        await login(updatedToken, isRedirectedFromLoginPage === 'true' ? ROUTES.WORKFLOWS : ROUTES.AUTH_APPLICATION);
```
SokratisVidros, line 71: Ouch. This has to be done this way as the accept invitation endpoint is [authenticated](https://github.com/novuhq/novu/blob/17230b1232f0cb19a7c697372ad7876c674cef51/apps/api/src/app/invites/invites.controller.ts#L66) but it shouldn't.

Removing the `UserGuard` is a breaking change, but I'd say the current need for an authentication token is wrong, so I feel confident making the change in a follow-up PR. 

What do you think?

SokratisVidros (changes requested):

apps/web/src/pages/auth/components/LoginForm.tsx
jainpawan21: I think, the  user should exist before accepting invitation. So current `UserGuard` is correct

libs/application-generic/src/services/auth/shared.ts
jainpawan21: @SokratisVidros which means instead of hardcoed `/auth/login `at [this](https://github.com/novuhq/novu/blob/fix/hubspot-ques-form-revert/libs/application-generic/src/services/auth/shared.ts#L2) line, `redirect_url` query param value should be used?

SokratisVidros (approved):
I am approving this PR for now to release to production and we will get back to the `redirect_url` issue soon.

PR #5678, opened 2024-06-04
fix(echo): Replace faulty instanceof check with static code check

rifont:
* The `error instanceof EchoError` check was evaluating to `false` even when passed a valid `EchoError`. The issue only appears with the CommonJS bundle, not with ESModules. The decision was made to avoid `instanceof` altogether and replace with a static check to identify if the `Error.code` matches an expected code
* This fixes [this failing test](https://github.com/novuhq/novu/actions/runs/9367163429/job/25786344943?pr=5675#step:10:448) 

### Screenshots

<details>

</details>


packages/create-novu-app/package.json
```diff
@@ -20,7 +20,7 @@
   "scripts": {
     "start": "node dist/index.js",
     "dev": "ncc build ./index.ts -w -o dist/",
-    "prerelease": "node ../../scripts/rm.mjs dist",
+    "prerelease": "rimraf dist",
```
rifont, line 23: Small fix to ensure installation passes without error.

packages/echo/src/handler.ts
```diff
@@ -261,8 +262,12 @@ export class EchoRequestHandler<Input extends any[] = any[], Output = any> {
     }
   }
 
-  private handleError(error: any): IActionResponse {
-    if (error instanceof EchoError) {
```
djabarovgeorge, line 265: 🔥 the instanceof in cross bundles

djabarovgeorge (approved):
🚀

docker/local/docker-compose.e2e.yml
rifont, line 1: This addition was needed to fix the failing CI pipeline, it was accidentally removed in a previous PR.

PR #5681, opened 2024-06-05
feat(create-novu-app): add local tunnel capability

denis-kralj-novu:
Add local tunnel to novu app
Add dotenv to novu app
Add new launch scripts
Rename echo to novu


apps/api/src/.env.test
```diff
@@ -110,3 +110,5 @@ IS_USE_MERGED_DIGEST_ID_ENABLED=true
 
 HUBSPOT_INVITE_NUDGE_EMAIL_USER_LIST_ID=
 HUBSPOT_PRIVATE_APP_ACCESS_TOKEN=
+
+NOVU_ENTERPRISE=true
```
denis-kralj-novu: Would it make sense to include this in the repository? The tests expect this flag if I'm not mistaking, so it makes sense to me to keep this in

apps/web/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "pnpm panda --watch & cross-env NODE_OPTIONS=--max_old_space_size=8192 PORT=4200 react-app-rewired start",
     "prebuild": "rimraf build",
-    "build": "pnpm panda && cross-env NODE_OPTIONS=--max_old_space_size=4096 GENERATE_SOURCEMAP=false react-app-rewired --max_old_space_size=4096 build",
+    "build": "pnpm panda && cross-env NODE_OPTIONS=--max_old_space_size=8192 GENERATE_SOURCEMAP=false react-app-rewired --max_old_space_size=8192 build",
```
denis-kralj-novu, line 8: Without this the build would fail with the following error:
```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```
Since this setting is at the value of `8192` in the start command, I adjusted it here and then the build passed.

package.json
```diff
@@ -13,7 +13,7 @@
     "clean": "lerna clean --yes && rimraf node_modules",
     "commit": "cz",
     "nx": "nx",
-    "lint-staged": "lint-staged",
+    "lint-staged": "NODE_OPTIONS=--max_old_space_size=8192 lint-staged",
```
denis-kralj-novu, line 16: Without this committing would fail with the following error:
```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```
I aligned this value with the ones present in `apps/web`

apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
```diff
@@ -32,4 +41,24 @@ export class UpdateEnvironment {
       { $set: updatePayload }
     );
   }
+  async shouldUpdateEchoConfiguration(command: UpdateEnvironmentCommand): Promise<boolean> {
+    if (process.env.NOVU_ENTERPRISE === 'true' || process.env.CI_EE_TEST === 'true') {
```
denis-kralj-novu, line 45: I initially thought that I could use the command that I defined in the ee package for updating the endpoint. However, this would then lead to a situation where that command would return the results of the update, wheres the previous set of adjustments in the caller would have its own result. I could do consolidation logic since it is aiming via `environmentId`, but that seemed a bit messy. Let me know if that would still be preferable to this way of updating the data.

packages/create-novu-app/index.ts
```diff
@@ -19,6 +19,8 @@ let projectPath = '';
 
 const handleSigTerm = () => process.exit(0);
 
+const defaultTunnelHost = 'https://localtunnel.me';
```
denis-kralj-novu, line 22: for some reason the default value in the option doesn't work so I needed to place it here to have access to it.

packages/create-novu-app/templates/app-react-email/ts/scripts/tunnel.mjs
denis-kralj-novu, line 1: There is no error handling or retries or metric gathering right now. We can discuss if we want it at this time or if we should focus on more critical areas.

I'm also a bit bothered that for this script I wasn't enabled to do a typescript file. Would appreciate an example of how to achieve this to make coding in this space more reasonable.

denis-kralj-novu: I will also invest a bit of time to see if I can remove the duplication and have the 2 script files in one place instead of 2

PR #5682, opened 2024-06-05
feat(novui,web): V2 flow list

antonjoel82:

- Create table in `novui` with newest version of react-table
  - Add stories
- Add design tokens from Figma
  - _Note: I'm not thrilled with how some of them are done, but I've kept them as-is to match Figma for the time being_
- Setup test example of new Workflows page in `web`

### Validation

1. Run `pnpm storybook` in `novui` and look for Table
2. Visit https://deploy-preview-5682--dev-web-novu.netlify.app/studio/flows to see in web

### Screenshots

Screenshot 2024-06-06 at 11 22 56 AM
Screenshot 2024-06-06 at 11 25 47 AM

### DIFF
--- a/apps/web/src/studio/components/workflows/table/WorkflowsTable.tsx
+++ b/apps/web/src/studio/components/workflows/table/WorkflowsTable.tsx
@@ -1,4 +1,5 @@
 import { createColumnHelper, Table } from '@novu/novui';
+import { css } from '@novu/novui/css';
 import format from 'date-fns/format';
 import { FC } from 'react';
 import { WorkflowTableRow } from './WorkflowsTable.types';
@@ -97,6 +98,10 @@ const WORKFLOW_COLUMNS = [
   }),
 ];
 
-export const WorkflowsTable: FC<IWorkflowsTableProps> = (props) => {
-  return <Table<typeof TEST_FLOW> columns={WORKFLOW_COLUMNS} data={[TEST_FLOW]} />;
+export const WorkflowsTable: FC<IWorkflowsTableProps> = () => {
+  return (
+    <div className={css({ display: 'flex', flex: '1' })}>
+      <Table<typeof TEST_FLOW> columns={WORKFLOW_COLUMNS} data={[TEST_FLOW]} className={css({ w: '100%' })} />
+    </div>
+  );
 };
--- a/apps/web/src/studio/components/workflows/table/WorkflowsTable.tsx
+++ b/apps/web/src/studio/components/workflows/table/WorkflowsTable.tsx
@@ -1,3 +1,5 @@
+/* cspell:disable */
+
 import { createColumnHelper, Table } from '@novu/novui';
 import { css } from '@novu/novui/css';
 import format from 'date-fns/format';
@@ -5,6 +7,7 @@ import { FC } from 'react';
 import { WorkflowTableRow } from './WorkflowsTable.types';
 import { GroupCell, NameCell, StatusCell } from './WorkflowsTableCellRenderers';
 
+// TODO: remove this test data
 const TEST_FLOW = {
   _id: 'adaewr',
   name: 'test naming',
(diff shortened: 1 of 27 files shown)


apps/web/src/studio/components/workflows/table/WorkflowsTable.tsx
```diff
@@ -0,0 +1,110 @@
...
+      },
+      push: {
+        hasActiveIntegrations: false,
+      },
+    },
+  },
+};
+
+interface IWorkflowsTableProps {
+  temp?: string;
```
antonjoel82, line 80: 🗒 note (non-blocking): This is a placeholder for other props‏

libs/novui/src/components/table/Table.tsx
```diff
@@ -0,0 +1,107 @@
...
+}: ITableProps<TRow>) {
+  const columns = useMemo(() => userColumns?.map((col) => ({ ...col })), [userColumns]);
+  const data = useMemo(() => (userData || [])?.map((row) => ({ ...row })), [userData]);
+  const fakeData = useMemo(() => Array.from({ length: loadingItems }).map((_, index) => ({ index })), [loadingItems]);
+
+  const table = useReactTable<TRow>({
+    columns,
+    data: isLoading ? (fakeData as TRow[]) : data,
+    getCoreRowModel: getCoreRowModel(),
+    // FIXME: remove this
```
rifont: ❓ **question:** Was the intention to remove this one?‏

libs/novui/src/tokens/semanticSpacing.tokens.ts
```diff
@@ -15,4 +15,282 @@ export const SEMANTIC_SPACING_TOKENS = defineSemanticTokens.spacing({
...
+        horizontal: {
+          type: 'spacing',
+          value: '{spacing.50}',
+        },
+        vertical: {
+          type: 'spacing',
+          value: '{spacing.50}',
+        },
+      },
+      futer: {
```
rifont: ❓ **question:** ‏Is this a legitimate semantic name, or a typo of `footer`?

libs/novui/src/tokens/semanticSpacing.tokens.ts
```diff
@@ -15,4 +15,282 @@ export const SEMANTIC_SPACING_TOKENS = defineSemanticTokens.spacing({
...
+          type: 'spacing',
+          value: '{spacing.50}',
+        },
+        bottom: {
+          type: 'spacing',
+          value: '{spacing.100}',
+        },
+      },
+    },
+    WF: {
```
rifont, line 91: ❓ **question:** ‏I wonder if we should adhere to a semantic property naming convention here? We have a mixture of different cases right now (`camelCase`, `PascalCase`, `Upper-Kebab-Case`, `lower-kebab-case`), the pattern is difficult to assess and follow. My suggestion would be to `camelCase` everything.

rifont (approved):
Awesome! Never had Storybook load so quickly in my life 🚀

libs/novui/src/tokens/semanticSpacing.tokens.ts
antonjoel82: @rifont I believe it is a typo -- but it's how it's currently done in Figma. I'm not happy with a number of the tokens and how they're written, but I elected to keep everything to match Nik's designs for now...

libs/novui/src/tokens/semanticSpacing.tokens.ts
antonjoel82: @rifont I agree, but these are copied directly from Figma. I'm not sure if it's better to try to 

1. standardize them in code but have them differ from Figma
2. Or, keep all of them the same as Figma for consistency with designs (unless they actually break panda)

I opted for #2 now since I'd rather have discussions about how to make our system better when Nik is back. What do you think?

libs/novui/src/tokens/semanticSpacing.tokens.ts
antonjoel82: Actually -- decided to fix this because it's a mistake -- not a convention

PR #5665, opened 2024-06-03
feat(js): lazy session initialization and interface fixes

LetItRock:
### What changed? Why was the change needed?
Changes:
- JS SDK lazy session initialization with tests
- feed/notification interface improvements

### Screenshots

<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #5688, opened 2024-06-06
feat(js): js sdk feeds module

LetItRock:
The changes include:
- `@novu/client` ApiService types improvements
- The build configuration for ESM, CJS - preserving folders structure, UMD - minification and gzip
- Improvements to the size limits checker script
- The Feeds module
- The Notification ORM pattern
- Event Emitter types improvements

#### The UMD size limits checker script
Screenshot 2024-06-06 at 17 29 52

#### Testing the JS SDK
Screenshot 2024-06-06 at 17 29 31

<details>

</details>


packages/client/src/api/api.service.ts
```diff
@@ -43,6 +44,12 @@ export class ApiService {
     }
   }
 
+  private removeNullUndefined(obj) {
+    return Object.fromEntries(
+      Object.entries(obj).filter(([_, value]) => value != null)
+    );
+  }
```
LetItRock, line 51: the helper function that is used on the URL query params to remove the keys with `null` or `undefined` values

packages/client/src/api/api.service.ts
```diff
@@ -57,10 +64,10 @@ export class ApiService {
 
   async updateAction(
     messageId: string,
-    executedType: ButtonTypeEnum,
-    status: MessageActionStatusEnum,
+    executedType: `${ButtonTypeEnum}`,
+    status: `${MessageActionStatusEnum}`,
```
LetItRock, line 68: this is the way to convert enum keys to union, this way we don't require using the enum from `@novu/shared`

libs/shared/src/entities/messages/action.enum.ts
```diff
@@ -1,5 +1,4 @@
 export enum ButtonTypeEnum {
   PRIMARY = 'primary',
   SECONDARY = 'secondary',
-  CLICKED = 'clicked',
```
LetItRock, line 4: not used

packages/client/src/api/api.service.ts
```diff
@@ -71,6 +78,9 @@ export class ApiService {
     );
   }
 
+  /**
+   * @deprecated use markMessagesAs instead
+   */
```
LetItRock, line 83: the endpoint is deprecated so this function as well

packages/client/src/api/api.service.ts
```diff
@@ -86,6 +96,19 @@ export class ApiService {
...
+    markAs,
+  }: {
+    messageId: string | string[];
+    markAs: `${MarkMessagesAsEnum}`;
+  }): Promise<INotificationDto[]> {
+    return await this.httpClient.post(`/widgets/messages/mark-as`, {
+      messageId,
+      markAs,
+    });
+  }
```
LetItRock, line 110: the new function that should be used instead

packages/client/src/api/api.service.ts
```diff
@@ -154,17 +177,21 @@ export class ApiService {
...
   }
 
-  async getUnseenCount(query: IUnseenCountQuery = {}) {
+  async getUnseenCount(
+    query: IUnseenCountQuery = {}
+  ): Promise<{ count: number }> {
     return await this.httpClient.get(
       '/widgets/notifications/unseen',
-      query as unknown as CustomDataType
+      this.removeNullUndefined(query) as unknown as CustomDataType
```
LetItRock, line 185: remove `null or undefined` from query params

packages/js/package.json
```diff
@@ -5,25 +5,27 @@
   "description": "Novu's JavaScript SDK for building custom inbox notification experiences",
   "author": "",
   "license": "ISC",
-  "main": "dist/index.cjs",
-  "module": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "main": "dist/cjs/index.cjs",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
```
LetItRock, line 10: build the ESM and CJS modules with preserving the folders structure

packages/js/package.json
```diff
@@ -35,22 +37,19 @@
...
     "jest": "^29.3.1",
-    "size-limit": "^11.1.4",
+    "tiny-glob": "^0.2.9",
     "ts-jest": "^29.0.3",
     "tsup": "^8.0.2",
     "typescript": "4.9.5"
   },
   "dependencies": {
     "@novu/client": "workspace:*",
-    "@novu/shared": "workspace:*",
```
LetItRock, line 53: removed `@novu/shared` as it brings a lot of unneeded code to the bundle

packages/js/scripts/size-limit.js
```diff
@@ -1,75 +1,68 @@
+import fs from 'fs/promises';
```
LetItRock, line 1: improved the calculations of this script as sometimes it didn't provide the correct results

packages/js/scripts/size-limit.js
```diff
@@ -1,75 +1,68 @@
...
   {
-    name: 'UMD',
-    path: umdPath,
-    limit: LIMIT,
-    files: [umdPath],
-    sizeLimit: LIMIT_IN_BYTES,
+    name: 'UMD gzip',
+    filePath: umdGzipPath,
+    limit: '10 kb',
+    limitInBytes: 10_000,
```
LetItRock, line 25: check only UMD minified and gzip files

packages/js/src/base-module.ts
```diff
@@ -14,9 +16,9 @@ export class BaseModule {
   #callsQueue: CallQueueItem[] = [];
   #sessionError: unknown;
 
-  constructor(emitter: NovuEventEmitter, apiService: ApiService) {
-    this._emitter = emitter;
-    this._apiService = apiService;
+  constructor() {
+    this._emitter = NovuEventEmitter.getInstance();
+    this._apiService = ApiServiceSingleton.getInstance();
```
LetItRock, line 21: use the singleton pattern for the event emitter and api service, otherwise, I would have to pass these two across all classes

packages/js/src/event-emitter/types.ts
```diff
@@ -1,19 +1,120 @@
...
+  optimistic?: O;
+};
+type NovuSuccessEvent<A, R> = Pick<NovuPendingEvent<A>, 'args'> & {
+  result: R;
+};
+type NovuErrorEvent<A, F = undefined> = Pick<NovuPendingEvent<A>, 'args'> & {
+  error: unknown;
+  fallback?: F;
+};
+// three possible status of the event: pending, success, error
```
LetItRock, line 27: There are three types of emitted events: pending, success, and error.
Each event has an associated object that is passed with it:
- pending event - object with args and optional optimistic result
- success event - object with args and result
- error event - object with args, error, and optional fallback value

packages/js/src/event-emitter/types.ts
```diff
@@ -1,19 +1,120 @@
...
+type EventName<T extends string> = `${T}.pending` | `${T}.success` | `${T}.error`;
```
LetItRock, line 28: Event name consist of three parts: module.action.status
Module: session, feeds, notification
Action: module actions
Status: status of action

packages/js/src/event-emitter/types.ts
```diff
@@ -1,19 +1,120 @@
...
-  'session.initialize.*': undefined;
-  'session.initialize.pending': undefined;
-  'session.initialize.success': ISessionDto;
-  'session.initialize.error': { error: unknown };
-  'feeds.fetch.*': undefined;
-  'feeds.fetch.pending': undefined;
-  'feeds.fetch.success': { response: PaginatedResponse<Notification> };
-  'feeds.fetch.error': { error: unknown };
+type BaseEvents<T extends string, ARGS, RESULT, OPTIMISTIC = undefined, FALLBACK = undefined> = {
+  [key in `${EventName<T>}`]: EventObject<key, ARGS, RESULT, OPTIMISTIC, FALLBACK>;
```
LetItRock, line 46: the helped type that will create three event names with associated objects, for example:
```
type SessionInitializeEvents = BaseEvents<'session.initialize', InitializeSessionArgs, Session>;
// is actually
type SessionInitializeEvents = {
   'session.initialize.pending': NovuPendingEvent<InitializeSessionArgs>,
   'session.initialize.success': NovuSuccessEvent<InitializeSessionArgs, Session>,
   'session.initialize.error': NovuErrorEvent<InitializeSessionArgs>
}


```

packages/js/src/event-emitter/types.ts
```diff
@@ -1,19 +1,120 @@
...
+export type Events = SessionInitializeEvents &
+  FeedFetchEvents &
+  FeedFetchCountEvents &
+  FeedMarkNotificationsAsEvents &
+  FeedMarkAllNotificationsAsEvents &
+  FeedRemoveNotificationsEvents &
+  FeedRemoveAllNotificationsEvents &
+  NotificationMarkAsEvents &
+  NotificationMarkActionAsEvents &
+  NotificationRemoveEvents;
```
LetItRock, line 116: all possible events of the event emitter

packages/js/src/feeds/feeds.ts
```diff
@@ -1,33 +1,187 @@
...
+  MarkNotificationActionAsByInstanceArgs,
+} from './types';
+import { READ_OR_UNREAD, SEEN_OR_UNSEEN } from '../utils/notification-utils';
+import { markActionAs, markNotificationAs, markNotificationsAs, remove, removeNotifications } from './helpers';
 
 export class Feeds extends BaseModule {
-  async fetch({ page = 1, ...restOptions }: FetchFeedOptions): Promise<PaginatedResponse<Notification>> {
+  async fetch({ page = 0, status, ...restOptions }: FetchFeedArgs = {}): Promise<PaginatedResponse<Notification>> {
     return this.callWithSession(async () => {
+      const args = { page, status, ...restOptions };
```
LetItRock, line 31: args with default values

packages/js/src/feeds/feeds.ts
```diff
@@ -1,33 +1,187 @@
...
+  async markNotificationAs(args: MarkNotificationAsByInstanceArgs): Promise<Notification>;
+  async markNotificationAs(args: MarkNotificationAsArgs): Promise<Notification> {
+    return this.callWithSession(async () =>
+      markNotificationAs({
+        emitter: this._emitter,
+        apiService: this._apiService,
+        args,
+      })
+    );
+  }
```
LetItRock, line 94: allows for the usage:
```
novu.feeds.markNotificationAs({ id: 'id', status: ... });
novu.feeds.markNotificationAs({ notification, status: ... });
```
similar pattern is used in other functions

LetItRock: btw all functions that are based on notification are having the optimistic updates logic

LetItRock: The Feeds module has the similar functionality like Notification has, that's why it reuses the same logic `markNotificationAs`

packages/js/src/feeds/helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+  apiService: ApiService;
+  args: MarkNotificationAsArgs;
+}): Promise<Notification> => {
+  const isNotification = typeof notification !== 'undefined';
+  const notificationId = isNotification ? notification._id : id ?? '';
+  const args = { id, notification, status };
+  try {
+    emitter.emit('notification.mark_as.pending', {
+      args,
+      optimistic: isNotification ? new Notification({ ...notification, ...getOptimisticMarkAs(status) }) : undefined,
```
LetItRock, line 57: Optimistic result. 
In case if function is called with `id` we are not able to build optimistic result.

packages/js/src/feeds/helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+    emitter.emit('notification.mark_as.success', { args, result: updatedNotification });
+
+    return updatedNotification;
+  } catch (error) {
+    emitter.emit('notification.mark_as.error', {
+      args,
+      error,
+      fallback: isNotification
+        ? new Notification({ ...notification, ...getFallbackMarkAs(status, notification) })
+        : undefined,
```
LetItRock, line 75: the fallback value

packages/js/src/feeds/notification.ts
```diff
@@ -1,23 +1,63 @@
...
-export class Notification {
   _id: string;
   _subscriberId: string;
   _feedId?: string | null;
   createdAt: string;
   updatedAt: string;
-  actor?: IActor;
-  subscriber?: ISubscriberResponseDto;
+  actor?: Actor;
+  subscriber?: Subscriber;
```
LetItRock, line 43: I decided to provide a better interface for these fields and not use the interface from `@novu/shared`. The downside of this is that we have a duplicated types, but we will change it in the future when we will switch to autogenerated sdk for the API.

packages/js/src/feeds/notification.ts
```diff
@@ -30,8 +70,95 @@ export class Notification {
     this.content = notification.content;
     this.read = notification.read;
     this.seen = notification.seen;
+    this.deleted = notification.deleted;
     this.cta = notification.cta;
     this.payload = notification.payload;
     this.overrides = notification.overrides;
   }
+
+  markAsRead(): Promise<Notification> {
```
LetItRock, line 73: the interface for some of these functions is simpler and doesn't require passing any args

packages/js/src/feeds/types.ts
```diff
@@ -0,0 +1,86 @@
+import type { NotificationActionStatus, NotificationButton, NotificationStatus } from '../types';
+import { Notification } from './notification';
+
+export type FetchFeedArgs = {
```
LetItRock, line 4: types for all function args from feeds module

packages/js/src/novu.ts
```diff
@@ -1,39 +1,35 @@
...
 
   public readonly feeds: Feeds;
   public readonly preferences: Preferences;
 
   constructor(options: NovuOptions) {
-    this.#apiService = new ApiService(options.backendUrl ?? PRODUCTION_BACKEND_URL);
-    this.#emitter = new NovuEventEmitter();
-    this.#session = new Session(this.#emitter, this.#apiService, {
+    ApiServiceSingleton.getInstance({ backendUrl: options.backendUrl });
+    this.#emitter = NovuEventEmitter.getInstance({ recreate: true });
```
LetItRock, line 24: the singleton are created with this class initialization

packages/js/test-sdk.ts
```diff
@@ -0,0 +1,43 @@
+import { Novu } from './src';
```
LetItRock, line 1: the test file to verify the sdk, I use it with command `tsx ./test-sdk.ts`

packages/js/tsconfig.json
```diff
@@ -5,6 +5,8 @@
     "target": "ESNext",
     "module": "ESNext",
     "strict": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true,
```
LetItRock, line 9: stricter ts checks

packages/js/src/base-module.ts
SokratisVidros: Alternatively we can use a factory pattern and get the event emitter from the closure. Testing would be slightly easier in that case due to dependency injection (DI).

packages/js/src/feeds/feeds.ts
```diff
@@ -1,33 +1,187 @@
...
       } catch (error) {
-        this._emitter.emit('feeds.fetch.error', { error });
+        this._emitter.emit('feeds.mark_all_notifications_as.error', { args, error });
+        throw error;
+      }
+    });
+  }
+
+  async markNotificationActionAs(args: MarkNotificationActionAsByIdArgs): Promise<Notification>;
+  async markNotificationActionAs(args: MarkNotificationActionAsByInstanceArgs): Promise<Notification>;
```
SokratisVidros, line 138: Let’s add specific method per action for better DX.

packages/js/src/types.ts
```diff
@@ -1,7 +1,111 @@
...
+  PRIMARY = 'primary',
+  SECONDARY = 'secondary',
+}
+
+export enum NotificationActionStatus {
+  PENDING = 'pending',
+  DONE = 'done',
+}
+
+export enum ActorType {
```
SokratisVidros, line 18: This enum is convoluted. Is its purpose to drive the icon selection?

the none status should be replaced with the null type. The as far as I can tell there is a user vs system actor. 

Id also argue that it might be more helpful to replace it with an icon property.

packages/js/src/types.ts
```diff
@@ -1,7 +1,111 @@
...
+  SYSTEM_CUSTOM = 'system_custom',
+}
+
+export enum CtaType {
+  REDIRECT = 'redirect',
+}
+
+export type Session = {
+  token: string;
+  profile: {
```
SokratisVidros, line 31:
```suggestion
  subscriber:
```

SokratisVidros (changes requested):

packages/js/src/base-module.ts
LetItRock: It might be, but in the tests, we will be mocking the `ApiService` class, not the singleton.

packages/js/src/types.ts
LetItRock: I'm not sure either if it's useful to the users at all, like they are just interested in the actor icon URL and nothing more.
But that's just what we have rn and we can improve later.

packages/js/src/types.ts
LetItRock: I responded in another PR: https://github.com/novuhq/novu/pull/5665/files#r1630782252. I'll create a ticket for that.

LetItRock: https://linear.app/novu/issue/COM-33/refactor-the-session-interface

packages/js/src/feeds/feeds.ts
LetItRock: Then we should do the same for the other functions as well, like the same way I did in the Notification class:
- markNotificationAsRead, markNotificationAsSeen, markNotificationAsUnseen, markNotificationAsUnread
- markNotificationPrimaryActionAsDone, markNotificationPrimaryActionAsPending, markNotificationSecondaryActionAsDone, markNotificationSecondaryActionAsPending

But then the question would be whether we should also emit separate events?

packages/js/src/types.ts
LetItRock: https://linear.app/novu/issue/COM-34/improve-the-actor-interface

packages/js/src/feeds/helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+const getOptimisticMarkAs = (status: NotificationStatus): Partial<Notification> => {
+  switch (status) {
+    case NotificationStatus.READ:
+      return { read: true, seen: true };
+    case NotificationStatus.UNREAD:
+      return { read: false, seen: true };
+    case NotificationStatus.SEEN:
+      return { seen: true };
+    case NotificationStatus.UNSEEN:
+      return { seen: false };
```
BiswaViraj, line 23:
```suggestion
      return { seen: false, read: false };
```

packages/js/src/feeds/helpers.ts
```diff
@@ -0,0 +1,263 @@
...
+};
+
+const getFallbackMarkAs = (status: NotificationStatus, notification: Notification): Partial<Notification> => {
+  switch (status) {
+    case NotificationStatus.READ:
+    case NotificationStatus.UNREAD:
+      return { read: notification.read, seen: notification.seen };
+    case NotificationStatus.SEEN:
+    case NotificationStatus.UNSEEN:
+      return { seen: notification.seen };
```
BiswaViraj, line 36:
```suggestion
      return { seen: notification.seen, read: notification.read };
```

packages/js/src/feeds/feeds.ts
```diff
@@ -1,33 +1,187 @@
...
+  RemoveNotificationsByInstancesArgs,
+  MarkNotificationActionAsByIdArgs,
+  MarkNotificationActionAsByInstanceArgs,
+} from './types';
+import { READ_OR_UNREAD, SEEN_OR_UNSEEN } from '../utils/notification-utils';
+import { markActionAs, markNotificationAs, markNotificationsAs, remove, removeNotifications } from './helpers';
 
 export class Feeds extends BaseModule {
-  async fetch({ page = 1, ...restOptions }: FetchFeedOptions): Promise<PaginatedResponse<Notification>> {
+  async fetch({ page = 0, status, ...restOptions }: FetchFeedArgs = {}): Promise<PaginatedResponse<Notification>> {
```
BiswaViraj, line 29: 💭 **thought (non-blocking):** I think we should start page numbers from 1
and handle internally by page = --page;‏

BiswaViraj (commented):
Looks good,
Have added couple of small comments

packages/js/src/feeds/helpers.ts
LetItRock: I didn't add that just because we haven't fixed that on the server side. Like to not confuse the users with an optimistic state different from the result (server) state.

packages/js/src/feeds/feeds.ts
LetItRock: yeah, everywhere in our APIs we are using a page from 0, and in sdks as well :(

packages/client/src/api/api.service.ts
SokratisVidros: nit-pick comment: Alternatively we can use the URLSearchParams interface.

packages/client/src/api/api.service.ts
```diff
@@ -86,6 +96,19 @@ export class ApiService {
...
   }
 
+  async markMessagesAs({
+    messageId,
+    markAs,
+  }: {
+    messageId: string | string[];
+    markAs: `${MarkMessagesAsEnum}`;
+  }): Promise<INotificationDto[]> {
+    return await this.httpClient.post(`/widgets/messages/mark-as`, {
```
SokratisVidros, line 106: In the new API, I'd like to suggest having a dedicated method per action in a RESTful way. `/v1/inbox/messages/:id/mark-as-read` without any body. This is not actionable for now.

packages/js/src/base-module.ts
SokratisVidros: Correct, that's where DI helps.

packages/js/src/event-emitter/types.ts
SokratisVidros: I am torn about this one. It feels a bit convoluted if we start from the event naming convention, deconstruct the parts, and then define the payload based on that. How about the opposite: start with an event map and build types from it?

For [example](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAogbhAdsAzlAvFA3gKCvqAcggWQEZCAubKAEwENh7qVgAnAS0QHMoBfANxQA9MNilgUAO4dgACyj0oYeiAA2Ae3q08BYhIBMVGgGMNAV2TVE5gLYAjCG35DRUAIKIN8p1BJJJGXlFOg4AMzCnAOVVTW1dfH0AgGZjOA0OWlcxeGiguQtJJRV1LR0CIn9kABZjGzU1bI8vH2cpVShgDToIMK5oekQ-CWlZAvMimNL4wRwcUEgoABkOViQnAB4AFQA+DCgtvwAPYCRaNHTMqAB+KAAKAEoMPcvaKGo7krjaai2n9BeGSycxManoKDQuWQMFsslObG2x1OiHOUAAShAzGxaBtWJweAAaRSIEA7Pa4CpgThwRjQNSrZFOFDULBQADaAGkoFwoABrCAgDRhA4AXWu1BWa0Qmy2nJFOzZIv4+ywswSUA0iA2XIgJzOaH5guFuzuVWA1A5RPpUqcEoZ6wRso58oe1Hkq2w6oq4XuAEJ3SgAHTWxlsFBss0ip4Uiqx-AB4P26VhiMSJWYRUCL0EPjZ+NyVaJm0pyO+wNgcwoOR3EMOh5ZuP4NgQYDmNhDAMNnNzCpCsLapH6vkCoUHHamiQWq1J23LGeOuU7V2dAtoGNxn13f2rouh8ORp7N1vtlerLtxhO15P7tP7S-zm8BEVlvpqeE155QNRQX3oTBXpx6zzKAjzbDtV3PfgewICBYWAAddWRVFDVHE0zSnKZvl+OVBxRC4gRuKBLFoXp+jebDnWXANPUbblhS3e9i0fZAoxAlswNPFBIIqMINGcO4zEQVgv3nDVhUYvdUyfX1o2AioALYT5YjKIDaNzWjQJPTt1VzXMcDcABVFB6G4aBdXoWwwDUCBKBBTVhNguFfEwaUpHEAIYSchEoVQccgP0sQAGVzHsFATE4ewuF4LphgCFAcEc4B4UDTU7iSchCCJJTplof49kElANGs4MNG4bLvkDBgmAeIDEuS1L0uAIwsq+FTPwKoqIBKsrWu0QMzEsYAaqzOqnBSxA0rNVIsryqAOuKzQysIHyoGSTpOG4Uzm1oQhhoSuD6omxrahm9r7M67q0pW6p1o4Taoh2vaAtgA6oti5B4tGthA0SyaJAoIlWSq5giAACQgBoNEIfhaoOsbfsa5rTAsKwoGqAwYZGuHvoRqbdqxryfrgv6AhOqB6jUesgA).

packages/js/src/feeds/feeds.ts
```diff
@@ -1,33 +1,187 @@
...
     return this.callWithSession(async () => {
+      const args = { page, status, ...restOptions };
       try {
-        this._emitter.emit('feeds.fetch.pending');
+        this._emitter.emit('feeds.fetch.pending', { args });
 
-        const response = await this._apiService.getNotificationsList(page, restOptions);
+        const response = await this._apiService.getNotificationsList(page, {
+          ...restOptions,
+          ...(status && SEEN_OR_UNSEEN.includes(status) && { seen: status === NotificationStatus.SEEN }),
```
SokratisVidros, line 37: This conversion of status between the SDK and the apiService seems unnecessary. Is it mandatory to keep the apiService backward compatible?

packages/js/src/feeds/feeds.ts
```diff
@@ -1,33 +1,187 @@
...
+    return this.callWithSession(async () =>
+      markNotificationAs({
+        emitter: this._emitter,
+        apiService: this._apiService,
+        args,
+      })
+    );
+  }
+
+  async markNotificationsAs(args: MarkNotificationsAsByIdsArgs): Promise<Notification[]>;
```
SokratisVidros, line 96: I suggest we go with a specific method on this public API for better DX, such as `novu.feeds.markAllAsRead()`.

SokratisVidros (changes requested):

packages/client/src/api/api.service.ts
LetItRock: it doesn't work :( 
Screenshot 2024-06-12 at 18 21 40

packages/client/src/api/api.service.ts
LetItRock: I'll create a ticket.

packages/js/src/base-module.ts
LetItRock: Can you please show with the example?

packages/js/src/feeds/feeds.ts
LetItRock: Unfortunately, it is a public package :( but I think that we would need to deprecate it and suggest using the new SDK instead

packages/js/src/feeds/feeds.ts
LetItRock: If you are saying to go additionally with these then yes we can introduce this granularity, but need to be consistent with that pattern and implement also for other methods:
- for single notification: `markNotificationAsRead, markNotificationAsUnread, markNotificationAsSeen, markNotificationAsUnseen`
- for selected notifications: `markNotificationsAsRead, markNotificationsAsUnread, markNotificationsAsSeen, markNotificationsAsUnseen`
- for all notifications: `markAllAsRead, markAllAsUnread, markAllAsSeen, markAllAsUnseen`
- similar pattern to actions

packages/js/src/event-emitter/types.ts
LetItRock: https://novu.slack.com/archives/C06TPLSQGUT/p1718208263023249?thread_ts=1718120928.044529&cid=C06TPLSQGUT

packages/js/src/feeds/feeds.ts
LetItRock: And then the question is whether all these should be treated as a separate events

PR #5692, opened 2024-06-06
feat(novui): Map panda values to mantine theme

antonjoel82:

- Create mapping layer between Panda tokens and Mantine
  - _This will not be a silver bullet, but should provide some stronger consistency from the get-go_
- Experiment with a Button to visualize differences (this is a WIP) and will be cleaned-up & improved in later PRs

### Notes

- Using `gradient` as our primary color works for "fills" / backgrounds, but doesn't work for `borders` so `outline` variants do not work out of the box. This will have to be addressed manually.

### Screenshots

Screenshot 2024-06-07 at 9 25 12 AM

### DIFF
--- a/libs/novui/src/components/button/Button.tsx
+++ b/libs/novui/src/components/button/Button.tsx
@@ -1,0 +1,25 @@
+import { FC } from 'react';
+import { CorePropsWithChildren } from '../../types';
+import { Button as ExternalButton, ButtonProps } from '@mantine/core';
+import { IconType } from '../../icons';
+import { css } from '../../../styled-system/css';
+
+export interface IButtonProps
+  extends CorePropsWithChildren,
+    React.ButtonHTMLAttributes<HTMLButtonElement>,
+    Pick<ButtonProps, 'size' | 'variant'> {
+  Icon?: IconType;
+}
+
+export const Button: FC<IButtonProps> = ({ children, Icon, size = 'md', variant = 'filled', ...buttonProps }) => {
+  return (
+    <ExternalButton
+      size={size}
+      leftSection={Icon ? <Icon title="button-icon" size="16" className={css({ color: 'legacy.white' })} /> : undefined}
+      variant={variant}
+      {...buttonProps}
+    >
+      {children}
+    </ExternalButton>
+  );
+};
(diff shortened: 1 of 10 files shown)


libs/novui/src/components/button/Button.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <ExternalButton
+      size={size}
+      leftSection={Icon ? <Icon title="button-icon" size="16" className={css({ color: 'legacy.white' })} /> : undefined}
+      variant={variant}
+      {...buttonProps}
+    >
+      {children}
+    </ExternalButton>
+  );
+};
```
rifont, line 25: 👏 **praise:** Super clean!‏

libs/novui/src/components/button/Button.tsx
```diff
@@ -0,0 +1,25 @@
...
+    React.ButtonHTMLAttributes<HTMLButtonElement>,
+    Pick<ButtonProps, 'size' | 'variant'> {
+  Icon?: IconType;
+}
+
+export const Button: FC<IButtonProps> = ({ children, Icon, size = 'md', variant = 'filled', ...buttonProps }) => {
+  return (
+    <ExternalButton
+      size={size}
+      leftSection={Icon ? <Icon title="button-icon" size="16" className={css({ color: 'legacy.white' })} /> : undefined}
```
rifont, line 18: 💭 **thought (non-blocking):** Is there a semantic token we can use to replace `legacy.white` for the color here? Not a major, but would be good to reach a point where we can stop using legacy stuff.‏‏

libs/novui/src/components/mantine-theme.config.ts
```diff
@@ -0,0 +1,87 @@
...
+  black: token('colors.legacy.black'),
+  primaryColor: 'gradient',
+  primaryShade: 6,
+  colors: {
+    gray: generateMantineColorTokens('mauve'),
+    yellow: generateMantineColorTokens('amber'),
+    blue: generateMantineColorTokens('blue'),
+    green: generateMantineColorTokens('green'),
+    red: generateMantineColorTokens('red'),
+    // must have a tuple of 10 strings, but replace the value at primaryShade with our gradient
```
rifont, line 27: 👏 **praise:** ‏Nice comment 🚀

rifont (approved):
Awesome, this will save us a lot of time. Let's get it shipped and we can refine as needed later 🚀🚢

libs/novui/src/components/button/Button.tsx
antonjoel82: Sorry, that is just a placeholder for testing. We're going to need quite a bit more customization for our buttons.

I'm going to have to meet with Nik when he's back though because the colors are in a confusing state right now, and it's difficult to discern from designs what maps to what.

PR #5694, opened 2024-06-06
ci(root): Update new cloud-deploy workflow

Cliftonz:
- Slack notifications `slackapi/slack-github-action`
- ECS services matrix fix for worker service
- Workflow fix
- Image fix
- Only manual run for testing mode
- The Docker build process will be updated a bit later


.github/workflows/cloud-deploy-embed.yml
```diff
@@ -0,0 +1,51 @@
...
+
+name: Deploy EMBED
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  workflow_call:
+    inputs:
+      env:
+        description: 'Deployment environment - choose between dev and prod'
```
SokratisVidros, line 11:
```suggestion
        description: 'Deployment environment - choose between staging and production'
```

.github/workflows/cloud-deploy-embed.yml
```diff
@@ -0,0 +1,51 @@
...
+        required: true
+        type: string
```
SokratisVidros, line 13: Please default this to staging

.github/workflows/cloud-deploy-embed.yml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  # This workflow contains a single job called "build"
+  deploy_embed:
+    if: ${{ contains(inputs.env, 'dev') }}
```
SokratisVidros, line 17:
```suggestion
    if: ${{ contains(inputs.env, 'staging') }}
```

.github/workflows/cloud-deploy-embed.yml
SokratisVidros: @AliaksandrRyzhou how about adding three options here, US/EU and Staging, considering each environment differently, this will make the task definition simpler as we will avoid the if conditions for every environment, later on.

As for the values per environment such as the widget_api_url we can fetch them from a predefined map of such values.

.github/workflows/cloud-deploy-reusable-app-service-deploy.yml
```diff
@@ -0,0 +1,229 @@
...
+        run: |
+          echo "ecs_container_name=$(terraform output -json ${{ env.SERVICE_NAME }}_ecs_container_name | jq -r .)" >> $GITHUB_OUTPUT
+          echo "ecs_service=$(terraform output -json ${{ env.SERVICE_NAME }}_ecs_service | jq -r .)" >> $GITHUB_OUTPUT
+          echo "ecs_cluster=$(terraform output -json ${{ env.SERVICE_NAME }}_ecs_cluster | jq -r .)" >> $GITHUB_OUTPUT
+          echo "task_name=$(terraform output -json ${{ env.SERVICE_NAME }}_task_name | jq -r .)" >> $GITHUB_OUTPUT
+          echo "aws_region=$(terraform output -json aws_region | jq -r .)" >> $GITHUB_OUTPUT
+          echo "queue_workers_services=$(terraform output -json queue_workers_services)" >> $GITHUB_OUTPUT
+
+  deploy_service_worker:
+    if: inputs.service_name == 'worker'
```
SokratisVidros, line 88: What is the difference between worker and other services?

.github/workflows/cloud-deploy-reusable-docker.yml
```diff
@@ -0,0 +1,204 @@
...
+        with:
+          version: v24.0.6
+          daemon-config: |
+            {
+              "features": {
+                "containerd-snapshotter": true
+              }
+            }
+
+      - name: Setup QEMU
```
SokratisVidros, line 110: @AliaksandrRyzhou Since this is about cloud deployment, why do we need QEMU?

.github/workflows/cloud-deploy-reusable-docker.yml
```diff
@@ -0,0 +1,204 @@
...
+          platforms: ${{ needs.job-setup.outputs.platforms }}
+
+      - name: Set Up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        with:
+          driver-opts: 'image=moby/buildkit:v0.13.1'
+
+      - name: Build, tag, and push image to ghcr.io
+        id: build-image
+        if: ${{ inputs.env_tag == 'dev' || inputs.env_tag == 'stg' }}
```
SokratisVidros, line 122: If we pass environment as an option, building should be per environment so there shouldn't be any need for conditionals here.

.github/workflows/cloud-deploy-services.yml
```diff
@@ -0,0 +1,100 @@
...
+jobs:
+  set_environment:
+    runs-on: ubuntu-latest
+    outputs:
+      environment: ${{ steps.set_env.outputs.environment }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: set_env
+        run: |
```
SokratisVidros, line 40: Same applies here. We should eliminate the per environment conditionals as much as possible.

SokratisVidros (changes requested):

.github/workflows/cloud-deploy-reusable-app-service-deploy.yml
Cliftonz: @SokratisVidros The difference is the worker service is not one service like the rest of the services. It's deployed as five different worker sets one for each queue in the system.
This is also something you will want to keep in mind if you do add a new queue/process to Novu as it will require a new worker service to be set up.

Thus the worker is different, and in all honestly, all other services should deploy like the worker service to keep it cleanly organized, however, the terraform query will still be different.

SokratisVidros: If not mistaken, this should be part of the ECS (or TS definition), not part of the CI. It's a smell that the CI need to change everytime we change a queue.

.github/workflows/cloud-deploy-reusable-docker.yml
AliaksandrRyzhou: I've left it as an opportunity to manage platforms in the future. It didn't overload CICD significantly. I'll delete it, I agree

SokratisVidros: Thank you. Let's only keep what's needed for now for cloud deployments. Every second counts.

.github/workflows/cloud-deploy-embed.yml
AliaksandrRyzhou: If we create the three options here it means that in `Production` deployment we have to call twice `Deploy EMBED WF` in `cloud-deploy.yml` because the `prod` has two regions which in turn means that we will have to add extra jobs. In this WF we use conditions only to create a few jobs instead of one in the production deployment case, because to make matrix element conditionals won't look greate. All this  jobs are equal for both environments beacause in the end they use the same `cloud-deploy-reusable-embed-deploy.yml` only with different variables. I see what you mean to create something like here:
```
name: Deploy EMBED

on:
  workflow_call:
    inputs:
      env:
        description: 'Deployment environment - choose between staging, production-eu, and production-us'
        required: true
        type: string
        default: 'staging'

jobs:
  deploy_embed:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        environment: [staging, production-eu, production-us]

    steps:
      - name: Set Environment Variables
        id: set-env-vars
        run: |
          if [ "${{ matrix.environment }}" == "staging" ]; then
            echo "env=Development" >> $GITHUB_ENV
            echo "widget_url=https://dev.widget.novu.co" >> $GITHUB_ENV
            echo "netlify_deploy_message=Dev deployment" >> $GITHUB_ENV
            echo "netlify_alias=dev" >> $GITHUB_ENV
            echo "netlify_gh_env=development" >> $GITHUB_ENV
            echo "netlify_site_id=22682666-3a8d-40be-af26-017bfadf5ae9" >> $GITHUB_ENV
          elif [ "${{ matrix.environment }}" == "production-eu" ]; then
            echo "env=Production" >> $GITHUB_ENV
            echo "widget_url=https://eu.widget.novu.co" >> $GITHUB_ENV
            echo "netlify_deploy_message=Production deployment" >> $GITHUB_ENV
            echo "netlify_alias=prod" >> $GITHUB_ENV
            echo "netlify_gh_env=Production" >> $GITHUB_ENV
            echo "netlify_site_id=0c830b50-df83-480b-ba36-a7f3176efcc8" >> $GITHUB_ENV
          elif [ "${{ matrix.environment }}" == "production-us" ]; then
            echo "env=Production" >> $GITHUB_ENV
            echo "widget_url=https://widget.novu.co" >> $GITHUB_ENV
            echo "netlify_deploy_message=Production deployment" >> $GITHUB_ENV
            echo "netlify_alias=prod" >> $GITHUB_ENV
            echo "netlify_gh_env=Production" >> $GITHUB_ENV
            echo "netlify_site_id=0689c015-fca0-4940-a26d-3e33f561bc48" >> $GITHUB_ENV
          fi

      - name: Deploy
        uses: ./.github/workflows/cloud-deploy-reusable-embed-deploy.yml
        with:
          environment: ${{ env.env }}
          widget_url: ${{ env.widget_url }}
          netlify_deploy_message: ${{ env.netlify_deploy_message }}
          netlify_alias: ${{ env.netlify_alias }}
          netlify_gh_env: ${{ env.netlify_gh_env }}
          netlify_site_id: ${{ env.netlify_site_id }}
        secrets: inherit

```
But as you can see and as I said we we should call it a few times. We will receive the same difference between environments on a level of the workflows higher. @SokratisVidros Please let me know your opinion.

.github/workflows/cloud-deploy-embed.yml
AliaksandrRyzhou: @SokratisVidros Only have one concern about it. Are we going to rename `dev` to `staging` inside of all cloud deployment files? Because we have a lot of code pieces where we use environment GitHub, Netlify. If we change this we have to add one more variable gh_environment/nf_environment to continue using these environments within these systems. Or we should create for instance Staging environment within GitHub and fill it in the same secrets and variables as the Development environment already has. Because the most part of jobs use `env: ${{ needs.pre-release.outputs.environment }}` and in the lowest level this variable is a 
```
jobs:
  reusable_embed_deploy:
    runs-on: ubuntu-latest
    timeout-minutes: 80
    environment: ${{ inputs.environment }}
```
or
```
jobs:
  reusable_web_deploy:
    runs-on: ubuntu-latest
    timeout-minutes: 80
    environment: ${{ inputs.environment }}
```
or
```
  reusable_docker:
    needs: [job-setup]
    runs-on: ubuntu-latest
    timeout-minutes: 80
    environment: ${{ inputs.environment }}
```
or
```
  infrastructure_data:
    runs-on: ubuntu-latest
    timeout-minutes: 80
    environment: ${{ inputs.environment }}
```

.github/workflows/cloud-deploy-embed.yml
SokratisVidros: For posterity, this was resolved offline.

SokratisVidros:
The team is now working on a new direction for a simplified CICD.

PR #5697, opened 2024-06-07
ci(root): Add tests back to each PR

Cliftonz:


.github/workflows/on-pr.yml
```diff
@@ -18,6 +23,49 @@ jobs:
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@v4
 
+  test_web:
```
SokratisVidros, line 26: This workflow channel already contains a `test_e2e_web` step that uses the get-affected step as an input. We should consolidate the steps and make them all mandatory for a PR to be merged.

SokratisVidros (changes requested):

.github/workflows/on-pr.yml
Cliftonz: @SokratisVidros Can you clarify what your looking for when you say make them all mandatory?

They are already all mandatory by github branch rules?

SokratisVidros: Yes. Mandatory as rules. I am unsure if that will work, though, as some steps might be skipped due to [this limitation](https://github.com/orgs/community/discussions/13690).

An interesting solution to the problem was proposed by [Mixpanel](https://engineering.mixpanel.com/enforcing-required-checks-on-conditional-ci-jobs-in-a-github-monorepo-8d4949694340).

PR #5698, opened 2024-06-07
feat(root): include local tunnel work

denis-kralj-novu:

The `create-novu-app` package now has the ability to open a localtunnel to your app on `npm run dev`

For local testing of behavior, the project can be built with `npm run dev`. It can then be triggered via `npm run start -- --api-key=[YOUR_API_KEY]`. using pnpm is slightly different with `pnpm run start --api-key=[YOUR_API_KEY]`

### With enterprise
- on user registration, the bridge endpoint url is also stored in the environment configuration
- the bridge endpoint url gets updated when the development environment api key is regenerated
- the environments endpoint allows to update the bridge endpoint url on PUT
## Related enterprise PR
This PR depends on [this enterprise PR](https://github.com/novuhq/packages-enterprise/pull/139)


apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
```diff
@@ -32,4 +41,24 @@ export class UpdateEnvironment {
       { $set: updatePayload }
     );
   }
+  async shouldUpdateEchoConfiguration(command: UpdateEnvironmentCommand): Promise<boolean> {
+    if (process.env.NOVU_ENTERPRISE === 'true' || process.env.CI_EE_TEST === 'true') {
```
denis-kralj-novu, line 45: I initially thought that I could use the command that I defined in the ee package for updating the endpoint. However, this would then lead to a situation where that command would return the results of the update, wheres the previous set of adjustments in the caller would have its own result. I could do consolidation logic since it is aiming via `environmentId`, but that seemed a bit messy. Let me know if that would still be preferable to this way of updating the data.

apps/web/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "pnpm panda --watch & cross-env NODE_OPTIONS=--max_old_space_size=8192 PORT=4200 react-app-rewired start",
     "prebuild": "rimraf build",
-    "build": "pnpm panda && cross-env NODE_OPTIONS=--max_old_space_size=4096 GENERATE_SOURCEMAP=false react-app-rewired --max_old_space_size=4096 build",
+    "build": "pnpm panda && cross-env NODE_OPTIONS=--max_old_space_size=8192 GENERATE_SOURCEMAP=false react-app-rewired --max_old_space_size=8192 build",
```
denis-kralj-novu: Without this the build would fail with the following error:
```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```
Since this setting is at the value of `8192` in the start command, I adjusted it here and then the build passed.

package.json
```diff
@@ -13,7 +13,7 @@
     "clean": "lerna clean --yes && rimraf node_modules",
     "commit": "cz",
     "nx": "nx",
-    "lint-staged": "lint-staged",
+    "lint-staged": "NODE_OPTIONS=--max_old_space_size=8192 lint-staged",
```
denis-kralj-novu: Without this committing would fail with the following error:
```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```
I aligned this value with the ones present in `apps/web`

packages/create-novu-app/index.ts
```diff
@@ -19,6 +19,8 @@ let projectPath = '';
 
 const handleSigTerm = () => process.exit(0);
 
+const defaultTunnelHost = 'https://localtunnel.me';
```
denis-kralj-novu, line 22: for some reason the default value in the option doesn't work so I needed to place it here to have access to it.

packages/create-novu-app/templates/index.ts
```diff
@@ -166,6 +173,21 @@ export const installTemplate = async ({
...
+  const host = tunnelHost;
+
+  await fs.writeFile(
+    path.join(root, ".env"),
+    `PORT=${port}` +
+      os.EOL +
+      `TUNNEL_SUBDOMAIN=${subdomain}` +
+      os.EOL +
+      `TUNNEL_HOST=${host}` +
+      os.EOL,
```
rifont: 💭 **thought (non-blocking):** Consider `Object.entries(obj).reduce(...)` with `obj` being a map of env var key to value.‏

packages/create-novu-app/templates/index.ts
```diff
@@ -175,7 +197,10 @@ export const installTemplate = async ({
     version: "0.1.0",
     private: true,
     scripts: {
-      dev: "next dev --port=4000",
+      tunnel:
+        "node --loader ts-node/esm --no-warnings=ExperimentalWarning scripts/tunnel.mts",
+      "next-dev": `next dev --port=${port}`,
+      dev: "concurrently 'npm run tunnel' 'npm run next-dev'",
```
rifont: ⚠ **issue:** It feels a bit heavy-handed and opinionated to add a library to concurrently run the tunnel with Dev server. Are there `next dev` scripts we can hook into to setup the tunnel in a more transparent way?‏

package.json
rifont: This could be an indication of a memory leak, we shouldn't need this much memory to run a linting job. In the past when this has occurred, restarting my machine fixes the issue. I'm hesitant to make this change because it may mask further memory leaks. I'd suggest we revert for now.

```suggestion
    "lint-staged": "lint-staged",
```

packages/create-novu-app/index.ts
```diff
@@ -41,6 +43,22 @@ const program = new Commander.Command(packageJson.name)
...
+  .option(
+    '-k, --api-key [apiKey]',
+    `
+
+  Your Novu Development environment apiKey. Note that your novu app won't
+  work with a non development environment apiKey.
+`
+  )
+  .option(
+    '-th, --tunnel-host',
```
rifont: I find it a bit non-standard to have a shorthand flag variant with 2 characters. What do we think of shortening to `-t`? This interface shouldn't change very frequently, if we have another `t` character we can find something else. It's better to optimize for what we have today given that this is a critical part of the onboarding flow right now.

```suggestion
    '-t, --tunnel-host',
```

libs/application-generic/src/utils/buildBridgeEndpointUrl.ts
```diff
@@ -0,0 +1,19 @@
...
+): string => {
+  return `${buildBridgeSubdomain(apiKey)}.${baseAddress}`;
+};
+
+/*
+ * Creates a bridge subdomain based on the apiKey provided
+ */
+export const buildBridgeSubdomain = (apiKey: string): string => {
+  return createHash('md5').update(apiKey).digest('hex');
+};
```
rifont, line 22: ⚠ **issue:** This hashing function is simple enough for us to duplicate across both places we need it, adding a dependency on `@novu/application-generic` will significantly bloat the bundle size of `create-novu-app`. Let's duplicate for now until we have a better home for these types of functions, accepting double maintenance cost.‏

packages/create-novu-app/templates/app-react-email/ts/scripts/tunnel.mjs
rifont, line 1: ⚠ **issue:** My thinking is that we would generate the tunnel URL at runtime on the `create-novu-app` client to support the use-case of rolling the API key to a new value. Currently, if the developer rolls the API-key ("rolls" meaning they invalidate the old key and generate a new one), the tunnel URL becomes invalid. We want to transparently support providing a new API key without needing to regenerate the tunnel.‏

packages/create-novu-app/templates/app/ts/scripts/tunnel.mjs
```diff
@@ -0,0 +1,34 @@
...
+
+dotenv.config();
+
+(async () => {
+  if (
+    process.env.PORT === undefined ||
+    Number.isNaN(parseInt(process.env.PORT))
+  ) {
+    throw new Error(
+      "POST environment variable value is required and should be a positive integer",
```
rifont: Typo
```suggestion
      "PORT environment variable value is required and should be a positive integer",
```

packages/create-novu-app/index.ts
denis-kralj-novu: Yep, that makes sense :)

libs/application-generic/src/utils/buildBridgeEndpointUrl.ts
denis-kralj-novu: Right, agree, should I add a note somewhere to say that there should be parity between the 2 places?

packages/create-novu-app/templates/app-react-email/ts/scripts/tunnel.mjs
denis-kralj-novu: Right, the subdomain gets calculated when `create-novu-app` is run, so we have that today. The scenario with the rolling, I'm thinking to just add the regeneration of the url as well to the process of rolling the API key as well. The roll use-case would call the regenerate URL use-case, what do you think?

package.json
denis-kralj-novu: reverted back

packages/create-novu-app/templates/app-react-email/ts/scripts/tunnel.mjs
denis-kralj-novu: subdomain now generated based on api key on `npm run dev`

apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
```diff
@@ -20,10 +20,19 @@ export class UpdateEnvironment {
       updatePayload.identifier = command.identifier;
     }
 
-    if (command.dns && command.dns.inboundParseDomain !== '') {
+    if (command.dns && command.dns.inboundParseDomain && command.dns.inboundParseDomain !== '') {
```
denis-kralj-novu, line 23: Added check here to also ensure that the value isn't falsy

apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
denis-kralj-novu: @rifont let me hear your thoughts on this as well

packages/create-novu-app/templates/index.ts
denis-kralj-novu: done

packages/create-novu-app/templates/index.ts
denis-kralj-novu: discussed 1on1, decided to go with `concurrently`

apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
rifont: I think this is a suitable approach for the time being 👍

libs/application-generic/src/utils/buildBridgeEndpointUrl.ts
```diff
@@ -0,0 +1,22 @@
...
+): string => {
+  return `${buildBridgeSubdomain(apiKey)}.${baseAddress}`;
+};
+
+/*
+ * Creates a bridge subdomain based on the apiKey provided. This function is used in several
+ * places, including packages/create-novu-app/templates/index.ts when generating the
+ * subdomain in the bridge application. Developers should take care to keep changes
+ * in sync.
+ */
```
rifont, line 19: 👏 **praise:** ‏Nice comment 🙏

packages/create-novu-app/templates/index.ts
```diff
@@ -175,7 +193,9 @@ export const installTemplate = async ({
     version: "0.1.0",
     private: true,
     scripts: {
-      dev: "next dev --port=4000",
+      tunnel: "tsx scripts/tunnel.mts",
+      "next-dev": `next dev --port=${port}`,
+      dev: 'concurrently -k --restart-tries 5 --restart-after 1000 --names "📡 TUNNEL,🖥️  SERVER" -c "bgBlue.bold,bgMagenta.bold" "npm:tunnel" "npm:next-dev"',
```
rifont, line 198: 👏 **praise:** 😍 love the emojis‏

apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
rifont: 🚀

rifont (approved):
Awesome work! It's a very clean DX. Let's ship it! 🚀

PR #5701, opened 2024-06-10
feat(js): js sdk preferences

LetItRock:
The Preferences module for JS SDK.

### Screenshots

<details>

</details>


packages/client/src/api/api.service.ts
```diff
@@ -206,14 +207,28 @@ export class ApiService {
...
   }
 
+  async getPreferences({
+    level = PreferenceLevelEnum.TEMPLATE,
+  }: {
+    level?: `${PreferenceLevelEnum}`;
+  }): Promise<Array<IUserPreferenceSettings | IUserGlobalPreferenceSettings>> {
+    return this.httpClient.get(`/widgets/preferences/${level}`);
+  }
+
```
LetItRock, line 231: We provide newer APIs but it wasn't exposed with this ApiService class.

packages/js/src/preferences/helpers.ts
```diff
@@ -0,0 +1,64 @@
...
+  const level = hasWorkflow ? PreferenceLevel.TEMPLATE : PreferenceLevel.GLOBAL;
+
+  return new Preference({
+    level,
+    enabled: preference.enabled,
+    channels: preference.channels,
+    workflow,
+    overrides: preference.overrides,
+  });
+};
```
LetItRock: Map server preference to the one that we use in the module with a better interface.

packages/js/src/preferences/helpers.ts
```diff
@@ -0,0 +1,64 @@
...
+  const { workflowId, enabled, channel } = args;
+  try {
+    emitter.emit('preferences.update.pending', { args });
+
+    let response;
+    if (workflowId) {
+      response = await apiService.updateSubscriberPreference(workflowId, channel, enabled);
+    } else {
+      response = await apiService.updateSubscriberGlobalPreference([{ channelType: channel, enabled }]);
+    }
```
LetItRock, line 60: It seems that we can unify this in the future under one endpoint.

packages/js/src/feeds/notification.ts
```diff
@@ -35,20 +35,20 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'> {
...
-  subscriber?: Subscriber;
-  transactionId: string;
-  content: string;
-  read: boolean;
-  seen: boolean;
-  deleted: boolean;
-  cta: Cta;
-  payload: Record<string, unknown>;
-  overrides: Record<string, unknown>;
+  readonly _id: string;
```
SokratisVidros: Let's remove all the fields we don't need on the FE from this Class for now and we will clean them from the API at a later stage.

SokratisVidros:
```suggestion
  readonly id: string;
```

packages/js/src/feeds/notification.ts
```diff
@@ -35,20 +35,20 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'> {
...
+  readonly _feedId?: string | null;
```
SokratisVidros: We can remove this, notifications will always belong to a feed array.

packages/js/src/feeds/notification.ts
```diff
@@ -35,20 +35,20 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'> {
...
+  readonly createdAt: string;
+  readonly updatedAt: string;
+  readonly actor?: Actor;
+  readonly subscriber?: Subscriber;
```
SokratisVidros: Is this necessary for Inbox? If not please remove.

packages/js/src/feeds/notification.ts
```diff
@@ -35,20 +35,20 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'> {
...
+  readonly transactionId: string;
```
SokratisVidros: Ditto

packages/js/src/feeds/notification.ts
```diff
@@ -35,20 +35,20 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'> {
...
+  readonly content: string;
+  readonly read: boolean;
+  readonly seen: boolean;
+  readonly deleted: boolean;
```
SokratisVidros, line 24: Ditto

packages/js/src/feeds/notification.ts
```diff
@@ -35,20 +35,20 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'> {
...
+  readonly cta: Cta;
+  readonly payload: Record<string, unknown>;
+  readonly overrides: Record<string, unknown>;
```
SokratisVidros: Ditto.

packages/js/src/preferences/helpers.ts
SokratisVidros: Correct. Let's add a comment in the code about that.

packages/js/src/preferences/preference.ts
```diff
@@ -0,0 +1,38 @@
+import { ApiService } from '@novu/client';
+
+import { NovuEventEmitter } from '../event-emitter';
+import { ChannelPreference, ChannelPreferenceOverride, ChannelType, PreferenceLevel, WorkflowInfo } from '../types';
+import { ApiServiceSingleton } from '../utils/api-service-signleton';
+import { updatePreference } from './helpers';
```
SokratisVidros, line 6: Nit-picky suggestion: Move the helpers into the Preference class to have everything in one place and and avoid an extra import. Every byte counts ;)

packages/js/src/preferences/preferences.ts
```diff
@@ -1,3 +1,32 @@
 import { BaseModule } from '../base-module';
```
SokratisVidros, line 1: For a future PR: The naming is a bit off. Is this a `BaseModule` or `BaseResource`, or `BaseModel`?

packages/js/src/types.ts
```diff
@@ -99,6 +118,29 @@ export type Cta = {
   action?: MessageAction;
 };
 
+export type WorkflowInfo = {
```
SokratisVidros:
```suggestion
export type Workflow = {
```

packages/js/src/types.ts
```diff
@@ -99,6 +118,29 @@ export type Cta = {
...
+  _id: string;
+  name: string;
+  critical: boolean;
+  tags?: string[];
```
SokratisVidros: Do we need this at this stage? If no, please remove.

packages/js/src/types.ts
```diff
@@ -99,6 +118,29 @@ export type Cta = {
...
+  identifier: string;
```
SokratisVidros, line 125: What's the difference between an `identifier` and the `_id`?

packages/js/src/types.ts
```diff
@@ -99,6 +118,29 @@ export type Cta = {
   action?: MessageAction;
 };
 
+export type WorkflowInfo = {
+  _id: string;
```
SokratisVidros:
```suggestion
  id: string;
```

packages/js/src/types.ts
```diff
@@ -99,6 +118,29 @@ export type Cta = {
...
+  tags?: string[];
+  identifier: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  data?: Record<string, any>;
+};
+
+export type ChannelPreference = {
+  email?: boolean;
+  sms?: boolean;
+  in_app?: boolean;
```
SokratisVidros, line 133:
```suggestion
  inApp?: boolean;
```

packages/js/src/types.ts
```diff
@@ -99,6 +118,29 @@ export type Cta = {
...
+  chat?: boolean;
+  push?: boolean;
+};
+
+export type ChannelPreferenceOverride = {
```
SokratisVidros: This seems overly complicated for the current frontend needs and is driven by the current state of the API. Let's build the DX on an ideal scenario and hide the API tech-debt with internal helper methods. Then we will update the API.

SokratisVidros (changes requested):

packages/js/src/feeds/notification.ts
LetItRock: but you might want to know to which feed it belongs when using individually

packages/js/src/preferences/helpers.ts
LetItRock: https://linear.app/novu/issue/COM-60/consolidate-the-update-preferences-global-template-under-one-endpoint

packages/js/src/preferences/preference.ts
LetItRock: can't do that as it's reused in two places: single Preference class and Preferences module

packages/js/src/preferences/preferences.ts
LetItRock: yes I was thinking what could be a better name here, what you would suggest instead?

packages/js/src/types.ts
LetItRock: `identifier` is used for the trigger, `_id` is internal Mongo id

packages/js/src/types.ts
LetItRock: this is due to internal implementation and will require changes across the stack :(

packages/js/src/types.ts
LetItRock: I think that we can remove it from the new Preference class for now, not even sure if that is used often by clients.

packages/js/src/event-emitter/novu-event-emitter.ts
```diff
@@ -1,23 +1,34 @@
...
   }
 
   on<Key extends EventNames>(eventName: Key, listener: EventHandler<Events[Key]>): void {
-    this.emitter.on(eventName, listener);
+    this.#mittEmitter.on(eventName, listener);
   }
 
   off<Key extends EventNames>(eventName: Key, listener: EventHandler<Events[Key]>): void {
-    this.emitter.on(eventName, listener);
+    this.#mittEmitter.on(eventName, listener);
```
BiswaViraj:
```suggestion
    this.#mittEmitter.off(eventName, listener);
```

BiswaViraj: Sorry, noticed it is already fixed in the other PR 👌🏼

BiswaViraj (approved):

PR #5704, opened 2024-06-10
feat(js): handling the web socket connection and events

LetItRock:
The changes:
- lazily connect to the WebSocket
- allow to listen to WebSocket events
- build UMD module with Webpack

### Screenshots

<details>

</details>


packages/client/src/api/api.service.ts
```diff
@@ -1,4 +1,4 @@
-import {
+import type {
```
LetItRock, line 1: use `import type` everywhere in the `@novu/client` so that `@novu/shared` won't be included in the bundle

packages/js/package.json
```diff
@@ -31,7 +31,8 @@
   "private": true,
   "scripts": {
     "start": "pnpm run build -- --watch --sourcemap",
-    "build": "tsup && pnpm run post:build",
+    "build": "tsup && pnpm run build:umd && pnpm run post:build",
+    "build:umd": "webpack --config webpack.config.cjs",
```
LetItRock, line 36: build UMD with Webpack

packages/js/src/base-module.ts
```diff
@@ -34,6 +37,10 @@ export class BaseModule {
     });
   }
 
+  protected onSessionSuccess(_: Session): void {}
+
+  protected onSessionError(_: unknown): void {}
```
LetItRock, line 42: These two functions will be overridden by the class that extends this one.

packages/js/src/event-emitter/novu-event-emitter.ts
```diff
@@ -25,7 +25,7 @@ export class NovuEventEmitter {
   }
 
   off<Key extends EventNames>(eventName: Key, listener: EventHandler<Events[Key]>): void {
-    this.#mittEmitter.on(eventName, listener);
+    this.#mittEmitter.off(eventName, listener);
```
LetItRock, line 28: my fault 😅

packages/js/src/novu.ts
```diff
@@ -30,13 +33,17 @@ export class Novu implements Pick<NovuEventEmitter, 'on' | 'off'> {
     this.#session.initialize();
     this.feeds = new Feeds();
     this.preferences = new Preferences();
+    this.#socket = new Socket({ socketUrl: options.socketUrl });
   }
 
   on<Key extends EventNames>(eventName: Key, listener: EventHandler<Events[Key]>): void {
+    if (this.#socket.isSocketEvent(eventName)) {
+      this.#socket.initialize();
+    }
```
LetItRock, line 44: When someone tries to listen to one of the WS events then initialize the WebSocket connection and add listeners on events.

packages/js/src/socket/socket.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.#emitter.emit(UNSEEN_COUNT_CHANGED, {
+      result: unseenCount,
+    });
+  };
+
+  #unreadCountChanged = ({ unreadCount }: { unreadCount: number }) => {
+    this.#emitter.emit(UNREAD_COUNT_CHANGED, {
+      result: unreadCount,
+    });
+  };
```
LetItRock: Handle different types of WebSocket events and just call our emitter.

packages/js/src/socket/socket.ts
```diff
@@ -0,0 +1,99 @@
...
+      this.#emitter.emit('socket.connect.success', { args, result: undefined });
+    });
+
+    this.#socketIo.on('connect_error', (error) => {
+      this.#emitter.emit('socket.connect.error', { args, error });
+    });
+
+    this.#socketIo?.on(WebSocketEvent.RECEIVED, this.#notificationReceived);
+    this.#socketIo?.on(WebSocketEvent.UNSEEN, this.#unseenCountChanged);
+    this.#socketIo?.on(WebSocketEvent.UNREAD, this.#unreadCountChanged);
```
LetItRock: Create the socket connection and listen on the events.

packages/js/src/socket/socket.ts
```diff
@@ -0,0 +1,99 @@
...
+  initialize(): void {
+    if (this.#token) {
+      this.#initializeSocket().then((error) => console.error(error));
+
+      return;
+    }
+
+    this.callWithSession(async () => {
+      this.#initializeSocket().then((error) => console.error(error));
+    });
```
LetItRock: When there is a session just initialize the socket, otherwise, initialize it when the session is created (lazily).

packages/js/tsup.config.ts
```diff
@@ -39,24 +39,24 @@ export default defineConfig([
...
+  //   },
+  //   esbuildPlugins: [
+  //     compress({
+  //       gzip: true,
+  //       brotli: false,
+  //       outputDir: '.',
+  //       exclude: ['**/*.map'],
+  //     }),
+  //   ],
+  // },
```
LetItRock: Commented out this code temporarily until we investigate why Rollup includes in UMD build some additional code that results in bigger bundle size.

packages/client/src/api/api.service.ts
BiswaViraj: 💭 **thought (non-blocking):** If we are just using the types, maybe we can move the `@novu/shared` under the `devDependency` as well.‏

LetItRock: We can't because of the types generated for ESM, they are not "copied" during the build time.

BiswaViraj (approved):

packages/js/package.json
SokratisVidros: Let's add a task to review this before release to have a more predictable and simpler building pipeline.

packages/js/scripts/size-limit.mjs
```diff
@@ -15,14 +15,12 @@ const modules = [
   {
     name: 'UMD minified',
     filePath: umdPath,
-    limit: '10 kb',
-    limitInBytes: 20_000,
+    limitInBytes: 70_000,
```
SokratisVidros, line 18: What's the current size?

```suggestion
    limitInBytes: 30_000,
```

packages/js/src/socket/socket.ts
```diff
@@ -0,0 +1,99 @@
+import io, { Socket as SocketIO } from 'socket.io-client';
```
SokratisVidros: @LetItRock It feels safer to add WebSocket handling under a `/realtime` subpath to isolate it from the root export for react-native as I am afraid it won't work out of the box.

This goes hand-in-hand with the plugin-based architecture I described in the latest DX guide. The Novu object gets plugins, one of them is the components, the other is for WS realtime updates.

packages/js/src/socket/socket.ts
SokratisVidros: What type of information do we get from WS currently, except unread_count? I'd like to suggest to have one updated event for all WS signals for easier scalability in the future.

packages/js/src/socket/socket.ts
```diff
@@ -0,0 +1,99 @@
...
+    });
+  };
+
+  #unreadCountChanged = ({ unreadCount }: { unreadCount: number }) => {
+    this.#emitter.emit(UNREAD_COUNT_CHANGED, {
+      result: unreadCount,
+    });
+  };
+
+  async #initializeSocket(): Promise<void> {
```
SokratisVidros: @LetItRock We must add the leader election logic across tabs to maintain only one WS per browser. Cross-tab communication can happen with Broadcast API or its polyfill based on local storage events.

PTAL https://greenvitriol.com/posts/browser-leader

SokratisVidros (changes requested):

packages/js/package.json
LetItRock: https://linear.app/novu/issue/COM-58/give-a-second-chance-to-tsup-to-build-the-umd-files

packages/js/src/socket/socket.ts
LetItRock: you can check the three above functions and their interfaces, that's what is called and returned

packages/js/scripts/size-limit.mjs
LetItRock: Screenshot 2024-06-20 at 16 59 26

packages/js/src/socket/socket.ts
LetItRock: I remember, we have a ticket for that: https://linear.app/novu/issue/COM-55/single-websocket-connection-across-tabs

packages/js/src/socket/socket.ts
SokratisVidros: For posterity, this was resolved offline.

SokratisVidros (approved):

PR #5711, opened 2024-06-12
chore(root): rename chimera to bridge

denis-kralj-novu:
rename chimera to bridge. We want to move away from the prototype names to agreed names.

### Related enterprise PR
This PR depends on **[this EE PR](https://github.com/novuhq/packages-enterprise/pull/142)**


SokratisVidros (approved):

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
```diff
@@ -320,7 +320,7 @@ async function discoverAndSyncEcho(session: UserSession) {
   const resultDiscover = await axios.get(echoServer.serverPath + '/echo?action=discover');
 
   await session.testAgent.post(`/v1/echo/sync`).send({
-    chimeraUrl: echoServer.serverPath + '/echo',
+    bridgeUrl: echoServer.serverPath + '/echo',
```
scopsy, line 323: @denis-kralj-novu I'm about to merge this PR: https://github.com/novuhq/novu/pull/5619/files and it was decided to use just `url` there. I believe we should go with it?

scopsy:
@SokratisVidros @denis-kralj-novu how are we with a breaking change here? We currently have around 8 active organizations using v2. My concern is that after the deploy it will suddenly stop working for them regardless of their sdk version

scopsy (approved):

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
rifont: I think we should be more explicit in case we add support for any other type of `url` in the future, so I'm in favour of `bridgeUrl`

apps/api/src/app/events/e2e/echo-trigger.e2e-ee.ts
```diff
@@ -320,7 +320,7 @@ async function discoverAndSyncEcho(session: UserSession) {
   const resultDiscover = await axios.get(echoServer.serverPath + '/echo?action=discover');
 
   await session.testAgent.post(`/v1/echo/sync`).send({
```
rifont, line 322: Let's also handle a backwards compatible rename from `/v1/echo` -> `/v1/bridge` in this PR.

apps/api/src/app/events/usecases/send-test-email/send-test-email.usecase.ts
```diff
@@ -87,10 +87,10 @@ export class SendTestEmail {
       subject = template.subject;
     }
 
-    if (command.chimera) {
+    if (command.bridge) {
       if (process.env.NOVU_ENTERPRISE === 'true' || process.env.CI_EE_TEST === 'true') {
         if (!require('@novu/ee-echo-api')?.PreviewStep) {
```
rifont, line 92: Similarly, let's rename `@novu/ee-echo-api` -> `@novu/ee-bridge-api`

rifont (commented):
We should handle all renaming of `echo` -> `bridge` in this PR too.

Please disregard what I'd mentioned about not renaming the SDK, let's also handle it in this PR and get the migration out of the way. This means:
* renaming `@novu/echo` -> `@novu/framework`
* updating the `create-novu-app` endpoint from a default of `/api/echo` -> `/api/novu`
* updating the SDK client construction in templates from `Echo` -> `Novu` and the instance itself from `echo` -> `novu` (so we get `novu.workflows(...)`)

PR #5712, opened 2024-06-12
chore(web): More e2e stability fixes

SokratisVidros:

This is a huge amount of work aiming at E2E stability:

1. Do not clear the DB from any e2e tests, as it kills the parallelization.
2. Stabilize auth and invitation tests after the latest changes in the invitation flow.
3. Clean up test password logic in @novu/testing to have more predictable passwords
4. Reuse the same email normalizer across the codebase and avoid bugs in testing.
5. Stop using Launch Darkly in E2E tests because it causes flakiness. Use feature flags via `window._env` instead.
6. Remove isPlaywright configuration logic in favor of env vars.
7. Lint and clean up e2e files according to the repo ESLint config.


.github/workflows/reusable-web-e2e.yml
```diff
@@ -97,13 +98,13 @@ jobs:
 
       - name: Playwright Install
         working-directory: apps/web
-        run: pnpm playwright:install
+        run: pnpm test:e2e:install
```
rifont, line 106: Nice one 🦾

rifont (approved):
🧹🚀🎉

PR #5717, opened 2024-06-12
refactor(echo): introduce declaring workflow

denis-kralj-novu:
- Add workflow function
- Adjust tests to work with new paradigm
- Extract independent tests to separate suite
- Adjust handler and serve methods

### Enterprise PR dependence

[✅ MERGED] This PR depends on [this EE PR](https://github.com/novuhq/packages-enterprise/pull/145)


packages/echo/src/express.ts
```diff
@@ -8,9 +9,11 @@ import { type SupportedFrameworkName } from './types';
 export const frameworkName: SupportedFrameworkName = 'express';
 
 export const serve = (options: ServeHandlerOptions): any => {
+  const client = options.client ? options.client : new Echo();
```
rifont: I think we can handle this instantiation inside the `EchoRequestHandler`, the `...options` spread will ensure the client is injected if available, and we won't need to modify any framework files.

denis-kralj-novu: Yep, much better approach

packages/echo/src/handler.ts
```diff
@@ -64,9 +66,10 @@ export class EchoRequestHandler<Input extends any[] = any[], Output = any> {
 
   constructor(options: IEchoRequestHandlerOptions<Input, Output>) {
     this.handler = options.handler;
-    this.client = options.client;
+    this.client = options.client ? options.client : new Echo();
+    this.client.addWorkflows(options.workflow);
```
denis-kralj-novu: @rifont do we concern ourselves that devs might use this method as well? We don't say it it in our docs (note to self: update readme with how to use the framework) but there is still a chance devs might see this and use it

SokratisVidros: Do we have to expose it? I see it's public, but we can make it private.

packages/echo/src/workflow.ts
```diff
@@ -0,0 +1,203 @@
...
+          channel as ChannelStepEnum,
+          ajv,
+          schemas.output,
+          schemas.result
+        );
+
+        return acc;
+      }, {} as Record<ChannelStepEnum, ActionStep<any, any>>),
+      /*
+       * Temporary workaround for inApp, which has snake_case step type
```
SokratisVidros, line 69: Is the `in_app` customer facing? If it is, let's change it to `inApp` and internally change it again until we fix the API.

packages/echo/src/client.ts
```diff
@@ -72,15 +72,26 @@ export class Echo {
...
 
     const ajv = new Ajv({ useDefaults: true });
     addFormats(ajv);
     this.ajv = ajv;
   }
 
+  public addWorkflows(workflows: Array<DiscoverWorkflowOutput>) {
+    for (const workflow of workflows) {
+      if (this.discoveredWorkflows.some((existing) => existing.workflowId === workflow.workflowId)) {
+        throw new WorkflowAlreadyExistsError(workflow.workflowId);
```
SokratisVidros, line 106: Should we throw an error or discard duplicates? What's better DX in this case?

packages/echo/src/client.ts
```diff
@@ -72,15 +72,26 @@ export class Echo {
   public static NOVU_SIGNATURE_HEADER = HttpHeaderKeysEnum.SIGNATURE;
 
   constructor(config?: ClientConfig) {
-    this.apiKey = config?.apiKey;
+    // TODO: transform this as per notion DX guide
+    this.apiKey = config?.apiKey ?? process.env.NOVU_API_KEY;
     this.backendUrl = config?.backendUrl ?? DEFAULT_NOVU_API_BASE_URL;
-    this.devModeBypassAuthentication = config?.devModeBypassAuthentication || false;
+    this.devModeBypassAuthentication = config?.devModeBypassAuthentication || process.env.NODE_ENV === 'development';
```
SokratisVidros:
```suggestion
    this.strictAuthentication = config?.strictAuthentication || process.env.NODE_ENV === 'development';
```

packages/echo/src/client.ts
```diff
@@ -72,15 +72,26 @@ export class Echo {
   public static NOVU_SIGNATURE_HEADER = HttpHeaderKeysEnum.SIGNATURE;
 
   constructor(config?: ClientConfig) {
-    this.apiKey = config?.apiKey;
+    // TODO: transform this as per notion DX guide
+    this.apiKey = config?.apiKey ?? process.env.NOVU_API_KEY;
```
SokratisVidros: I suggest moving the config(options) parsing to a separate method that will handle the default, env var fallback logic in one place and enforce naming convention on options and env vars.

packages/echo/src/client.ts
```diff
@@ -72,15 +72,26 @@ export class Echo {
...
     this.backendUrl = config?.backendUrl ?? DEFAULT_NOVU_API_BASE_URL;
```
SokratisVidros:
```suggestion
    this.apiUrl = config?.apiUrl ?? NOVU_API_URL;
```

packages/echo/src/client.ts
```diff
@@ -72,15 +72,26 @@ export class Echo {
   public static NOVU_SIGNATURE_HEADER = HttpHeaderKeysEnum.SIGNATURE;
 
   constructor(config?: ClientConfig) {
```
SokratisVidros:
```suggestion
  constructor(options?: ClientOptions) {
```

SokratisVidros (changes requested):

packages/echo/src/handler.ts
denis-kralj-novu: if I make it private then it is not accessible in the handler either. I can move all the logic associated with persisting the workflows in the handler but it is more work.

Is there a way in TS to make a class method available within the package but not outside of it?

packages/echo/src/client.ts
denis-kralj-novu: I don't think implicitly discarding duplicates it the way to go. How would we know which copy is the "correct" one? the one defined first or last? Throwing explains to the user that the identifiers have to be unique

SokratisVidros: Thanks for the clarification.

packages/echo/src/client.ts
```diff
@@ -72,15 +72,47 @@ export class Echo {
...
+      return '';
+    }, '');
+
+    return builtConfiguration;
+  }
+
+  public addWorkflows(workflows: Array<DiscoverWorkflowOutput>) {
+    for (const workflow of workflows) {
+      if (this.discoveredWorkflows.some((existing) => existing.workflowId === workflow.workflowId)) {
+        throw new WorkflowAlreadyExistsError(workflow.workflowId);
```
SokratisVidros, line 106: Having said that, instead of an error class per error type it's better DX to have a NovuFrameworkError and add messages and reason. That way it's easier for consumers to consume a single type and for us it creates less code.

packages/echo/src/handler.ts
SokratisVidros: I see. Maybe `this.client.workflows = workflows` then and use getter setters?

packages/echo/src/client.ts
denis-kralj-novu: you think that if the user wants to react on specific error states that matching message content is the way to go? in my experience, we do a catch all to log and then re throw, while particular errors that interest us are better targeted by matching types as opposed to message content

packages/echo/src/client.ts
denis-kralj-novu: if I understand correctly, your name adjustment means we are checking the inverse of what was there before (should be bypass auth because dev vs. enable strict auth). that would mean that the `process.env.NODE_ENV === 'development'` should also be the opposite

packages/echo/src/client.ts
SokratisVidros: In TS, every new type is an extra import. So the catch-all code becomes cumbersome. My suggestion wasn't just a message. It combines message, reason, and code in a `NovuError` class. The code is what helps you dictate logic. Let me know if this helps.

packages/echo/src/client.ts
denis-kralj-novu: ✔️

packages/echo/src/client.ts
denis-kralj-novu: yep, that makes more sense.

packages/echo/src/client.ts
denis-kralj-novu: done

packages/echo/src/client.test.ts
```diff
@@ -1,98 +1,39 @@
...
-    echo.workflow('setup-workflow', async ({ step }) => {
+  beforeEach(async () => {
+    const newWorkflow = await workflow('setup-workflow', async ({ step }) => {
       await step.email('send-email', async () => ({
         body: 'Test Body',
         subject: 'Subject',
       }));
     });
+
+    echo = new Client();
```
SokratisVidros:
```suggestion
    client = new Client();
```

packages/echo/src/client.ts
```diff
@@ -71,16 +71,48 @@ export class Echo {
 
   public static NOVU_SIGNATURE_HEADER = HttpHeaderKeysEnum.SIGNATURE;
 
-  constructor(config?: ClientConfig) {
-    this.apiKey = config?.apiKey;
-    this.backendUrl = config?.backendUrl ?? DEFAULT_NOVU_API_BASE_URL;
-    this.devModeBypassAuthentication = config?.devModeBypassAuthentication || false;
+  constructor(options?: ClientOptions) {
+    const builtConfiguration = this.buildConfiguration(options);
```
SokratisVidros: 🤓 

```suggestion
    const opts = this.buildOptions(options);
```

SokratisVidros (approved):

packages/echo/src/client.test.ts
denis-kralj-novu: fixed here and file wide

packages/echo/src/client.ts
denis-kralj-novu: good eye

.github/workflows/reusable-api-e2e.yml
```diff
@@ -90,6 +90,7 @@ jobs:
         run: CI='' pnpm build:api --skip-nx-cache
 
       - name: Run E2E tests
+        if: ${{ !inputs.ee }}
```
denis-kralj-novu, line 93: don't run these tests if we are only interested in the ee tests

packages/echo/src/handler.ts
```diff
@@ -23,16 +23,18 @@ import {
...
+  client?: Client;
+  workflow: Array<DiscoverWorkflowOutput>;
 }
 
-interface IEchoRequestHandlerOptions<Input extends any[] = any[], Output = any> extends ServeHandlerOptions {
+interface INovuRequestHandlerOptions<Input extends any[] = any[], Output = any> extends ServeHandlerOptions {
   frameworkName: string;
-  client: Echo;
+  client?: Client;
+  workflow: Array<DiscoverWorkflowOutput>;
```
denis-kralj-novu: I've seen this in the discussion in slack and in the DX guide. any particular reason why we are using a singular for a property that is containing an array?

packages/echo/src/handler.ts
denis-kralj-novu: let's not worry about that now. The docs talk about how we want to use it so if a developer uses the other methods without understanding how it works it is on them. We have provided a good DX, but we cannot force peoples hand. We can circle back to this if it becomes a problem

packages/echo/src/workflow.ts
denis-kralj-novu: added to my todo, won't target it in this PR

packages/echo/src/client.ts
denis-kralj-novu: updated

packages/echo/src/client.ts
denis-kralj-novu: updated

packages/echo/src/client.ts
denis-kralj-novu: Ok so I took the time to look at the situation more. Basically all errors inherit up the tree the `NovuError` base class, and this class includes a http status code as well as a code that describes what happened from the framework perspective.

GIven how try/catch works, with this inheritance approach the user can aim for exactly the specific error type they want with `instanceof`. If they want to work with multiple error type imports and handle each case that way, they can. On the other hand, they can simply aim for `NovuError` and see its code to decide what to do next.

Does this make sense @SokratisVidros ?

packages/echo/src/workflow.ts
rifont: `inApp` is the customer facing interface here, no further action needed.

packages/echo/src/client.ts
rifont: Exactly right @denis-kralj-novu , the `NovuError` is a generic Error interface, the specific Errors are just sugar for explicitly declaring errors throughout the framework with the `code` embedded. Some errors also have some sugar in the constructor to provide a nice error message where they are used in multiple places, to ensure consistent formatting.

In a future iteration, we can add an [Addressable Error](https://rauchg.com/2016/addressable-errors) to each of the error classes for better DX.

This approach grants SDK consumers the flexibility to handle different types of errors with `instanceof`, whilst also granting flexibility to use the `NovuError.code` to discriminate on the error if their preference is for error object access instead.

.github/workflows/reusable-api-e2e.yml
rifont: Nice! Been meaning to address this one 🙏

packages/echo/src/handler.ts
rifont: Good catch, that was a typo! It should be plural `workflows`

rifont:
```suggestion
  workflows: Array<DiscoverWorkflowOutput>;
```

packages/echo/src/handler.ts
rifont: With the switch to implicitly instantiate the client on `serve`, I think most Developers _won't_ see the client methods anymore, I'd expect less than 5% will need to explicitly use the `Client`.

We could take a good variety of approaches to not expose it, but I also like the idea of the Framework providing escape hatches for Developers to build custom things on top of. So long as we choose appropriate naming, which I believe `addWorkflows` is, we should grant Developers flexibility to go wild with this. We might learn something from their creativity that we can build out as a first-class capability later.

apps/api/e2e/echo.server.ts
```diff
@@ -2,22 +2,22 @@ import * as http from 'http';
...
   private port = 9999;
-  public echo = new Echo({ devModeBypassAuthentication: true });
+  public echo = new Client({ strictAuthentication: false });
 
   get serverPath() {
     return `http://localhost:${this.port}`;
   }
 
-  async start() {
+  async start(workflows: Array<DiscoverWorkflowOutput> = []) {
```
rifont: ✏ **suggestion:** I'd suggest making this an object containing `{workflows}` as we'll soon add more domains.‏

packages/create-novu-app/templates/app-react-email/ts/app/echo/client.ts
```diff
@@ -5,7 +5,7 @@ export const echo = new Echo({
   /**
    * Enable this flag only during local development
    */
-  devModeBypassAuthentication: process.env.NODE_ENV === "development",
+  strictAuthentication: process.env.NODE_ENV !== "development",
```
rifont, line 8: 👏 **praise:** Nice, much more readable.‏

packages/echo/src/client.ts
```diff
@@ -57,33 +57,75 @@ JSONSchemaFaker.option({
...
+        builtConfiguration[prop] = providedOptions[prop];
+      } else if (process.env[envVar]) {
+        builtConfiguration[prop] = process.env[envVar];
+      }
+
+      return '';
+    }, '');
+
+    return builtConfiguration;
+  }
```
rifont, line 101: The client is getting a bit bloated with "useless" methods, methods that don't actually need to access instance properties (i.e. `this.<property>`) and can therefore be refactored to stateless functions.

I don't think we should action now, but let's keep this in mind for future refactoring - we should be able to reduce the size of the `Client` quite significantly.

packages/echo/src/client.ts
```diff
@@ -57,33 +57,75 @@ JSONSchemaFaker.option({
...
+      }
+      const envVar = `NOVU_${toConstantCase(prop)}`;
+      if (providedOptions && providedOptions[prop]) {
+        builtConfiguration[prop] = providedOptions[prop];
+      } else if (process.env[envVar]) {
+        builtConfiguration[prop] = process.env[envVar];
+      }
+
+      return '';
+    }, '');
```
rifont: ✏ **suggestion:** Readability is suffering a bit here. Give the small number of config options today, I'll suggest we simplify this to a more classical style of `if (config.option !== undefined) { // handle option parsing/setting }`. We'll seek to use Zod across this codebase eventually and we can replace with a Zod schema with built-in parsing later. Wdyt?‏

packages/echo/src/utils/string.utils.test.ts
```diff
@@ -0,0 +1,20 @@
+import { toConstantCase } from './string.utils';
+
+describe('convert to snake case', () => {
```
rifont:
```suggestion
describe('convert to constant case', () => {
```

rifont (approved):
Awesome work! Left a few comments, most don't need to be addressed right now but are for future reference.

apps/api/e2e/echo.server.ts
denis-kralj-novu: done

packages/echo/src/client.ts
denis-kralj-novu: agree that it is an eyesore, The suggestion is valid and I was struggling with myself if to just do it that way :)

packages/echo/src/client.test.ts
```diff
@@ -166,60 +107,65 @@ describe('Echo Client', () => {
         }));
       });
 
-      const discovery = echo.discover();
+      client.addWorkflows([newWorkflow]);
+
+      // wait for discovery to finish
+      await new Promise((resolve) => setTimeout(resolve, 1));
```
denis-kralj-novu, line 113: @rifont so here is a problem with the non awaited workflow method that needs to be resolved in a timely manner. Because this test defines a pretty complex workflow, not all the async methods have time to finish before the `discover()` method below is called. Waiting even 1 millisecond 'fixes' this. 

I've confirmed that this is the case in next as well with the old way of defining workflows. If I remove the await from the `client.workflow()` calls, I still need this wait time to have this particular test pass.

I've tried to play around with the code but no real resolution came to my mind. In reality, the user can do operations in the workflow definition that take an arbitrary amount of time, and discovery won't pick up depending steps until those promises are fulfilled.

How would we target this?

PR #5719, opened 2024-06-12
feat(web,novui): Button, Title, Text, `colorPalette` experiments

antonjoel82:

- Refine `Button` appearance further
- Add `Text` and `Title` components
- Begin working with solutions using "virtual color" / `colorPalette` to make mode switching easier

### Screenshots

Screenshot 2024-06-12 at 3 07 00 PM
Screenshot 2024-06-12 at 3 07 04 PM


libs/novui/src/components/button/Button.tsx
```diff
@@ -1,22 +1,106 @@
...
-    Pick<ButtonProps, 'size' | 'variant'> {
+    Pick<ButtonProps, 'size'> {
   Icon?: IconType;
+  variant?: ButtonVariant;
 }
 
-export const Button: FC<IButtonProps> = ({ children, Icon, size = 'md', variant = 'filled', ...buttonProps }) => {
+const SLOTS: ButtonStylesNames[] = ['root', 'inner', 'label', 'loader', 'section'];
+
+const buttonRecipe = sva({
```
antonjoel82: 🗒 **note (non-blocking):** Unfortunately layer styles with mantine are not working when I try them, so for now the `important` below are necessary. I will dig back into it further, but didn't want to get blocked on it while defining components that everyone else will likely need‏

SokratisVidros (approved):

PR #5721, opened 2024-06-13
fix(novui): semantic colors tokens for table

davidsoderberg:


rifont (approved):

libs/novui/src/tokens/semanticColors.tokens.ts
```diff
@@ -16,20 +16,6 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
       type: 'color',
     },
   },
-  table: {
```
davidsoderberg, line 19: This was duplicate see line 76

davidsoderberg: Change was needed because of failing of here: https://github.com/novuhq/novu/actions/runs/9495865804/job/26169025237#step:6:53

PR #5731, opened 2024-06-14
feat(echo): Cors specific origin selection for prod environments - DRAFT - Do not Merge

scopsy:
Waiting for dev deployment for testing

### DIFF
--- a/packages/echo/src/handler.ts
+++ b/packages/echo/src/handler.ts
@@ -179,15 +179,31 @@ export class EchoRequestHandler<Input extends any[] = any[], Output = any> {
     const { originHeader, anonymousHeader } = headers;
     const isProduction = process.env.NODE_ENV === 'production';
     const isValidOrigin =
-      originHeader.includes('https://web.novu.co') ||
+      (originHeader && originHeader.includes('https://web.novu.co')) ||
       originHeader.includes('https://eu.web.novu.co') ||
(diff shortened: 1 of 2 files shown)


packages/echo/src/handler.ts
```diff
@@ -168,9 +173,39 @@ export class EchoRequestHandler<Input extends any[] = any[], Output = any> {
     workflowId: string,
     stepId: string,
     action: string,
-    anonymousHeader: string,
-    source: string
+    source: string,
+    headers: { originHeader: string; anonymousHeader: string }
```
scopsy, line 177: Refactored to easily allowing to add more headers in the future

packages/echo/src/handler.ts
```diff
@@ -168,9 +173,39 @@ export class EchoRequestHandler<Input extends any[] = any[], Output = any> {
...
   ): Record<PostActionEnum, () => Promise<IActionResponse>> {
+    const { originHeader, anonymousHeader } = headers;
+    const isProduction = process.env.NODE_ENV === 'production';
+    const isValidOrigin =
+      (originHeader && originHeader.includes('https://web.novu.co')) ||
```
rifont, line 182: I suggest we create a `NovuWebUrlEnum` with all options, then we can simplify this check to:
```typescript
const isValidOrigin = Object.values(NovuApiUrlEnum).includes(originHeader);
```

PR #5738, opened 2024-06-14
feat(novui,web): V2 step editor. Tabs

antonjoel82:

_Sorry, this PR got quite big as I made a push to have others able to start dev on Sunday!_

- Create Tabs component
- Fix color mode issues with Button and IconButton
- Make Button Polymorphic
- Scaffold pages so others can begin development

### Screenshots

https://www.loom.com/share/ab93d8d6b13c46baab8e0a1fbf8585df?sid=40ea1db1-f2a1-457a-a324-01e8d337244d


apps/web/src/AppRoutes.tsx
```diff
@@ -48,8 +48,12 @@ import { TenantsPage } from './pages/tenants/TenantsPage';
 import { UpdateTenantPage } from './pages/tenants/UpdateTenantPage';
 import { TranslationRoutes } from './pages/TranslationPages';
 import { useSettingsRoutes } from './SettingsRoutes';
-import { WorkflowsListPage } from './studio/components/workflows/WorkflowsListPage';
-import { WorkflowsDetailPage } from './studio/components/workflows/WorkflowsDetailPage';
+import {
```
SokratisVidros, line 51: 😍 Nice and tidy!

libs/novui/src/types/LocalizedMessage.ts
```diff
@@ -0,0 +1,13 @@
+/**
```
SokratisVidros, line 1: Can you elaborate on how this definition will help us with i18 in the future? Is it like a marker for places that need to be replaced with i18n-aware strings?

I'd also argue whether it belongs to the design system package or the web app. 

If that's the case, I'd suggest instead of using that to use a `t("this is a string")` function, that lives outside the design system that currently is just a unit function `f(x) => x`/

libs/novui/src/recipes/tabs.recipe.ts
```diff
@@ -0,0 +1,67 @@
...
+      _disabled: {
+        opacity: 'disabled',
+      },
+      '&:hover:not(:disabled)': {
+        opacity: 'hover',
+      },
+    },
+    list: {
+      color: 'typography.text.main',
+      gap: 'margins.layout.tabs.tab-tab !important',
```
SokratisVidros, line 23: Why are these `!important` required? Is it due to the old and new stylings living together in the new app?

libs/novui/src/components/tabs/Tabs.tsx
```diff
@@ -0,0 +1,59 @@
+import { Tabs as ExternalTabs, type TabsProps as ExternalTabsProps } from '@mantine/core';
+import { forwardRef, ReactNode } from 'react';
+import { CoreProps, LocalizedMessage } from 'src/types';
+import { css, cx } from '../../../styled-system/css';
```
SokratisVidros, line 4: I suggest adding a path alias to the novui tsconfig for `@/styled-system` access in a separate PR.

SokratisVidros (changes requested):

libs/novui/src/panda-preset.ts
```diff
@@ -85,7 +86,18 @@ export const novuPandaPreset = definePreset({
...
+        tabs: TABS_RECIPE,
       },
     },
   },
+  staticCss: {
+    css: [
+      {
+        properties: {
+          // Must generate color modes statically to ensure they're available
+          colorPalette: ['mode.local', 'model.cloud'],
```
ainouzgali:
```suggestion
          colorPalette: ['mode.local', 'mode.cloud'],
```

libs/novui/src/recipes/tabs.recipe.ts
rifont: @antonjoel82 , let's circle back on this later. Can you please create a ticket with a detail on why we needed this, and what steps we need to take to fix this.

libs/novui/src/types/LocalizedMessage.ts
rifont: @antonjoel82 , similarly, can you please respond in this thread and we'll continue the conversation here. A subsequent PR can be used to modify this proposal if necessary. @SokratisVidros , is that agreeable?

SokratisVidros (approved):
@antonjoel82 I approved this PR for now to unblock the rest of the team but let's talk about the comments later today or this week.

libs/novui/src/types/LocalizedMessage.ts
SokratisVidros: Of course!

PR #5743, opened 2024-06-16
feat(framework): add auto deterministic preview for required payload variables

scopsy:

The framework SDK will now use the `[placeholder]` value when previewing required values. This is helpful for when a user wants to see any workflow or preview it without getting an error first.

### Screenshots

CleanShot 2024-06-16 at 13 32 30@2x

<details>

</details>


packages/echo/src/client.ts
```diff
@@ -626,11 +633,16 @@ export class Echo {
   }
 
   private createExecutionInputs(event: IEvent, workflow: DiscoverWorkflowOutput): Record<string, unknown> {
-    const executionData = event.data;
+    let payload = event.data;
+    if (event.action === 'preview') {
+      const mockResult = this.mock(workflow.data.schema);
+
+      payload = Object.assign(mockResult, payload);
```
scopsy: Wonder if we should do something more smart here like recursive replace?

rifont: I think we should go one way or the other:
* No replacement at all
* Recursive replacement

If we're halfway, it's likely we'll end up the unexpected results when it comes to nested objects.

scopsy: Any suggestions of how to achieve this? Can install something like lodash deep merge perhaps?

rifont: Does this suggestion work? I'm not sure I understand why `Object.assign` is used
```suggestion
      payload = this.mock(workflow.data.schema);
```

rifont: I have a strong preference to keep this library very lightweight for fast execution on serverless/low-compute environments. Adding lodash feels like a bit of an overkill for now. I added a suggested below.

scopsy: The reason is that I want to polyfill all the values except the ones the user has provided in the inputs field. So what we need is merging the user provided inputs and the mocked object

PR #5744, opened 2024-06-16
feat(echo): introduce options method for cors config with nextjs

scopsy:
## Why?
This is needed to support CORS config with Next.js, and other frameworks when the v2 studio attempts to communicate with the handler.


djabarovgeorge (approved):
Looks good to me :)

PR #5746, opened 2024-06-17
feat(js): Introduce UI

BiswaViraj:
### Screenshots

<details>

</details>


packages/js/src/ui/context/appearance-context.tsx
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    //handle color variables
+    for (const key in props.variables?.colors) {
+      const colors = props.variables?.colors;
+
+      if (colors && colors.hasOwnProperty(key)) {
+        const value = colors[key as keyof Variables['colors']];
+
+        const shades = [50, 100, 200, 300, 400, 500, 600, 700, 800, 900];
```
SokratisVidros: Will we keep such calculations inside this context? As the project grows I feel we need an abstraction to calculate css variables in a unit-testable way.

packages/js/src/ui/index.tsx
```diff
@@ -1,7 +1,8 @@
 import { render } from 'solid-js/web';
 import Inbox from './Inbox';
 import type { NovuOptions } from '../novu';
-
+import './index.css';
```
SokratisVidros, line 4: What's the bundle size after all these additions?

packages/js/src/ui/helpers/utils.ts
```diff
@@ -0,0 +1,26 @@
...
+      return `${kebabKey}: ${value};`;
+    })
+    .join(' ');
+}
+
+export function createClassFromCssString(styleElement: HTMLStyleElement, styles: string) {
+  const index = styleElement.sheet?.cssRules.length ?? 0;
+  const className = `nv-css-${index}`;
+  const rule = `.${className} { ${styles} }`;
+  styleElement.sheet?.insertRule(rule, index);
```
SokratisVidros: Is there a way to do this in bulk? Should we think about browser performance considerations?

packages/js/src/ui/context/appearance-context.types.ts
```diff
@@ -0,0 +1,22 @@
+export type CSSProperties = {
```
SokratisVidros: I am not a big fan of having types outside of provides in separate types files. Shall we still use a single file across this package? cc: @LetItRock what's your take?

SokratisVidros (changes requested):

packages/js/src/ui/context/appearance-context.types.ts
desiprisg: I prefer a single file as well.

packages/js/src/ui/context/appearance-context.tsx
desiprisg: I'll refactor into some utilities. This is straight from the POC.

packages/js/src/ui/helpers/utils.ts
desiprisg: We should be able to do it in bulk.

packages/js/src/ui/index.tsx
BiswaViraj: Should we have UMD build as well?

SokratisVidros: Yes.

packages/js/.eslintrc.cjs
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   extends: ['../../.eslintrc.js'],
   rules: {
-    '@typescript-eslint/explicit-module-boundary-types': 'error',
+    '@typescript-eslint/explicit-module-boundary-types': 'off',
```
LetItRock, line 4: 💭 **thought (non-blocking):** this is useful for the SDK interface... I know that the TS can infer the types but it's a good practice to define it in the SDKs

packages/js/package.json
```diff
@@ -61,8 +78,10 @@
   },
   "dependencies": {
     "@novu/client": "workspace:*",
+    "clsx": "^2.1.1",
     "mitt": "^3.0.1",
     "socket.io-client": "4.7.2",
-    "solid-js": "^1.8.11"
+    "solid-js": "^1.8.11",
+    "tailwind-merge": "^2.3.0"
```
LetItRock: ❓ **question:** do we need this package? https://github.com/dcastil/tailwind-merge/blob/v2.3.0/docs/when-and-how-to-use-it.md#alternatives

LetItRock: https://bundlephobia.com/package/tailwind-merge@2.3.0

packages/js/src/ui/helpers/use-style.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    return cn(
+      ...descriptors,
+      className, // default styles
+      appearanceClassname, // overrides via appearance prop classes
+      ...cssInJsClasses
+    );
+  });
+
+  return styleFuncMemo();
```
LetItRock, line 31: ❓ **question:** do we need to call it? cant we just remove the first arrow fn in line 13?

packages/js/src/ui/helpers/use-style.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Elements, useAppearance } from '../context';
+import { cn } from './utils';
+
+export const useStyle = () => {
+  const appearance = useAppearance();
+  const [isServer, setIsServer] = createSignal(true);
+
+  onMount(() => {
+    setIsServer(false);
+  });
```
LetItRock, line 11: ✏ **suggestion:** no need for the state, we can have a reusable const `const isBrowser = typeof window !== 'undefined'`

packages/js/src/ui/Inbox.tsx
```diff
@@ -2,10 +2,13 @@ import { For, createSignal, onMount, type Component } from 'solid-js';
 import { Notification } from '../feeds';
 import { Novu } from '../novu';
 import type { NovuOptions } from '../novu';
+import { AppearanceContextType, AppearanceProvider } from './context';
+import { useStyle } from './helpers';
 
 const Inbox: Component<{
   name: string;
   options: NovuOptions;
+  appearance?: Pick<AppearanceContextType, 'elements' | 'variables'>;
```
LetItRock: ✏ **suggestion:** please create a separate type for this as it's used in a couple places

packages/js/src/ui/config/default-appearance.ts
```diff
@@ -0,0 +1,13 @@
...
+  colorPrimary: '#0081F1',
+  colorPrimaryForeground: 'white',
+  colorSecondary: '#F3F3F3',
+  colorSecondaryForeground: '#1A1523',
+  colorBackground: '#FFFFFF',
+  colorForeground: '#1A1523',
+  colorNeutral: 'black',
+  fontSize: 'inherit',
+  borderRadius: '0.375rem',
+};
```
BiswaViraj, line 13: ❓ **question:** ‏@desiprisg How will we handle the hover color and icon colors

packages/js/src/ui/context/appearance-context.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [rules, setRules] = createSignal<string[]>([]);
+
+  //place style element on HEAD. Placing in body is available for HTML 5.2 onward.
+  onMount(() => {
+    const styleEl = document.createElement('style');
+    styleEl.id = NOVU_CSS_IN_JS_STYLESHEET_ID;
+    document.head.appendChild(styleEl);
+
+    setStyleElement(styleEl);
+  });
```
LetItRock: ☑ **chore:** we should reuse the same style element between re-mounts

packages/js/src/ui/context/appearance-context.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  //handle variables
+  createEffect(() => {
+    const styleEl = styleElement();
+
+    if (!styleEl) {
+      return;
+    }
+
+    const variableRules = parseVariables({ ...defaultVariables, ...(props.variables || ({} as Variables)) });
```
LetItRock: ✏ **suggestion:** please think about a better name for this function, because it doesn't tell what it does, maybe something similar to `generateColorShades`

LetItRock: ❓ **question:** btw what is the idea behind having many color shades?

packages/js/src/ui/config/default-appearance.ts
desiprisg: Icon colors would be a shade of `foreground`. Hover is going to be a shade of `colorNeutralAlpha` for dropdowns and shades of the underlying color for buttons.

packages/js/src/ui/context/appearance-context.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const elementsStyleData = parseElements(props.elements || {});
+    setStore('descriptorToCssInJsClass', (obj) => ({
+      ...obj,
+      ...elementsStyleData.reduce<Record<string, string>>((acc, item) => {
+        acc[item.key] = item.className;
+
+        return acc;
+      }, {}),
+    }));
+    setRules((oldRules) => [...oldRules, ...elementsStyleData.map((el) => el.rule)]);
```
LetItRock: ☑ **chore:** There will be a problem with rules getting duplicated between re-renders. With the current code we do not support re-rendering, but with only mount/unmount. But we will be. 
So I think the easiest solution will be to recalculate and override all the rules in this case.

packages/js/src/ui/helpers/use-style.tsx
```diff
@@ -0,0 +1,32 @@
...
+        ? (appearance.elements?.[descriptor] as string) || ''
+        : '';
+
+    const descriptors = (Array.isArray(descriptor) ? descriptor : [descriptor]).map((des) => `nv-${des}`);
+    const cssInJsClasses =
+      !!descriptors.length && !isServer() ? descriptors.map((des) => appearance.descriptorToCssInJsClass[des]) : [];
+
+    return cn(
+      ...descriptors,
+      className, // default styles
```
LetItRock, line 25: ❓ **question:** Shouldn't the `descriptors` be after the `className` so the clients will be able to use them to override the default styles?

packages/js/tailwind.config.js
```diff
@@ -0,0 +1,46 @@
...
+module.exports = {
+  content: ['./src/**/*.{js,jsx,ts,tsx}'],
+  prefix: 'nt-',
+  corePlugins: {
+    preflight: false,
+  },
+  theme: {
+    extend: {
+      colors: {
+        primary: { DEFAULT: defaultColor('nv-color-primary'), ...generateColorShades('nv-color-primary') },
```
LetItRock, line 30: 💭 **thought (non-blocking):** the `DEFAULT` could be a part of the `generateColorShades` function

packages/js/src/ui/index.css
```diff
@@ -0,0 +1,469 @@
+/*
```
LetItRock, line 1: ✏ **suggestion:** Please prettify this file because the code and comments look odd

packages/js/src/ui/index.css
```diff
@@ -0,0 +1,469 @@
...
+! tailwindcss v3.4.4 | MIT License | https://tailwindcss.com
+*/ /*
+1. Prevent padding and border from affecting element width. (https://github.com/mozdevs/cssremedy/issues/4)
+2. Allow adding a border to an element by just adding a border-width. (https://github.com/tailwindcss/tailwindcss/pull/116)
+*/
+
+.novu {
```
LetItRock, line 8: ❓ **question:** Where did we get this CSS reset code from?

LetItRock (changes requested):
hey 👋 I left a couple suggestions and questions ;) let me know if something is not clear

packages/js/src/ui/context/appearance-context.types.ts
LetItRock: Yes, I prefer the same to have it in the file where the context is defined, usually because when checking the types I'm also interested really often to see the implementation details.

packages/js/package.json
desiprisg: I'll remove it for now. If we notice any conflict problems that are hard to fix, we can see about reintroducing. Thanks for looking out for the bundle here!

packages/js/src/ui/helpers/use-style.tsx
desiprisg: We noticed that this was necessary for everything to re-run. However, I want to take a closer look at all this logic at some point.

packages/js/src/ui/config/constants.ts
```diff
@@ -0,0 +1 @@
+export const NOVU_CSS_IN_JS_STYLESHEET_ID = 'novu-css-in-js-appearance-styles';
```
SokratisVidros, line 1: Can we make this ID shorter? Do we add it to the DOM?

packages/js/src/ui/helpers/use-style.tsx
desiprisg: This is just the order that the classes will be shown on the element. We want the descriptors first so that it's the first thing the dev sees. Order doesn't really matter.

packages/js/tailwind.config.js
desiprisg: We don't want a default for every usage of that function, just because we generate solid AND alpha shades, and that would mean two DEFAULTs.

packages/js/src/ui/index.css
desiprisg: This is just the default tailwind reset, but we're using the where selector to get specificity down to 0 to allow overriding.

packages/js/src/ui/helpers/use-style.tsx
desiprisg: We noticed that this was necessary for everything to re-run. However, I want to take a closer look at all this logic at some point.

packages/js/src/ui/config/default-appearance.ts
```diff
@@ -0,0 +1,13 @@
+import { Variables } from '../context';
+
+export const defaultVariables: Required<Variables> = {
+  colorPrimary: '#0081F1',
+  colorPrimaryForeground: 'white',
+  colorSecondary: '#F3F3F3',
+  colorSecondaryForeground: '#1A1523',
```
SokratisVidros, line 7: I am not a big fan of this after all as it doesn't tell me immediately what to do. Text on primary might be better after all.

SokratisVidros (approved):

LetItRock (approved):

packages/js/src/ui/context/appearance-context.tsx
desiprisg: Fixed!

PR #5750, opened 2024-06-17
chore(echo): rename @novu/echo to @novu/framework

denis-kralj-novu:
- Update package json
- Update package folder name
- Update readme details


packages/framework/README.md
denis-kralj-novu, line 1: First place with actual changes, the consideration is that we are updating the docs URL while we still don't have them updated, should this be included now or after we update the docs?

packages/framework/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@novu/echo",
-  "version": "0.24.3-alpha.4",
+  "name": "@novu/framework",
+  "version": "0.24.3-alpha.5",
```
denis-kralj-novu, line 3: updated alpha version here

denis-kralj-novu: Also, the author is set as "Athena", we wan't to change that to "Novu" maybe?

rifont: Yes, let's make it `Novu`

rifont (approved):
🪟🏗️ 🚀

PR #5755, opened 2024-06-17
feat(ee-auth): Use clerk in the enterprise version 

ChmaraX:

Notion Document explaning the change in detail:
https://www.notion.so/novuhq/Migration-to-Clerk-Auth-API-e097841461d14e51922cc9074b706695?pm=c

TODO:
- [x] add new e2e tests for the replaced implementation
- [x] revise existing e2e tests
- [ ] add new e2e tests for clerk webhooks
- [ ] solve environment switch on the web side
- [x] add another feature flag to be able to disable this change while having EE = true
- [x] there are new e2e-ee tests from https://github.com/novuhq/novu/pull/5698 which are now redundant after this change - we can probably delete them here

### Related enterprise PR
https://github.com/novuhq/packages-enterprise/pull/147

</details>


apps/api/src/app/auth/community.auth.module.config.ts
```diff
@@ -0,0 +1,95 @@
...
+import { injectRepositories } from '@novu/application-generic';
+
+const AUTH_STRATEGIES: Provider[] = [JwtStrategy, ApiKeyStrategy, JwtSubscriberStrategy];
+
+if (process.env.GITHUB_OAUTH_CLIENT_ID) {
+  AUTH_STRATEGIES.push(GitHubStrategy);
+}
+
+const authServiceProvider = {
+  provide: 'AUTH_SERVICE',
```
LetItRock: ✏ **suggestion:** it would be nice to use some constants for all these named injections

apps/api/src/app/auth/community.auth.module.config.ts
```diff
@@ -0,0 +1,95 @@
...
+    ],
+    exports: [
+      RolesGuard,
+      RootEnvironmentGuard,
+      AuthService,
+      'AUTH_SERVICE',
+      'USER_AUTH_GUARD',
+      'USER_REPOSITORY',
+      'MEMBER_REPOSITORY',
+      'ORGANIZATION_REPOSITORY',
```
LetItRock, line 63: ✏ **suggestion:** some of the exports and imports are shared between community and ee, maybe we can have a shared config?

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.usecase.ts
```diff
@@ -10,7 +10,7 @@ import { EnvironmentResponseDto } from '../../dtos/environment-response.dto';
   scope: Scope.REQUEST,
 })
 export class GetMyEnvironments {
-  constructor(private environmentRepository: EnvironmentRepository) {}
+  constructor(private environmentRepository: EnvironmentRepository, private authService: AuthService) {}
```
LetItRock: ❓ **question:** not used?

apps/api/src/app/organization/ee.organization.controller.ts
```diff
@@ -0,0 +1,120 @@
...
+      })
+    );
+  }
+
+  @Get('/me')
+  @ExternalApiAccessible()
+  @ApiResponse(OrganizationResponseDto)
+  @ApiOperation({
+    summary: 'Fetch current organization details',
+  })
```
LetItRock, line 46: ❓ **question:** left for backward compatibility and will be removed later right?

apps/api/src/app/organization/usecases/create-organization/sync-external-organization/sync-external-organization.usecase.ts
```diff
@@ -0,0 +1,131 @@
...
+import { ApiException } from '../../../../shared/exceptions/api.exception';
+import { CreateNovuIntegrations } from '../../../../integrations/usecases/create-novu-integrations/create-novu-integrations.usecase';
+import { CreateNovuIntegrationsCommand } from '../../../../integrations/usecases/create-novu-integrations/create-novu-integrations.command';
+import { ModuleRef } from '@nestjs/core';
+import { SyncExternalOrganizationCommand } from './sync-external-organization.command';
+
+@Injectable({
+  scope: Scope.REQUEST,
+})
+export class SyncExternalOrganization {
```
LetItRock, line 20: ☑ **chore:** I'm not sure if we need to have a separate usecase if under the hood we are injecting different repository implementations. I see it problematic when we would need to introduce some changes, we would have to remember to do that in two places.
So instead we can do an if statement before the `addMemberUsecase.execute` call to check if it's not enterprise; or inject the void `AddMemberUsecase`.

apps/api/package.json
```diff
@@ -22,8 +22,9 @@
...
-    "test:e2e": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true mocha --require ts-node/register --exit --file e2e/setup.ts src/**/*.e2e.ts",
-    "test:e2e:ee": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true CI_EE_TEST=true mocha --require ts-node/register --exit --file e2e/setup.ts src/**/*.e2e-ee.ts",
+    "test:e2e": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true mocha --require ts-node/register --exit --file e2e/setup.ts src/**/*.e2e.ts ",
+    "test:e2e:ee": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true mocha --grep @skip-in-ee --invert --require ts-node/register --exit --file e2e/setup.ts src/**/*.e2e.ts ",
+    "test:e2e:ee-only": "cross-env TS_NODE_COMPILER_OPTIONS='{\"strictNullChecks\": false}' TZ=UTC NODE_ENV=test E2E_RUNNER=true CI_EE_TEST=true mocha --require ts-node/register --exit --file e2e/setup.ts src/**/*.e2e-ee.ts",
```
LetItRock: ❓ **question:** what's the idea here?

apps/api/src/app/user/ee.user.controller.ts
```diff
@@ -0,0 +1,30 @@
...
+  })
+  async syncExternalUser(@UserSession() user: UserSessionData): Promise<UserEntity> {
+    Logger.verbose('Syncing external Clerk user', user._id);
+
+    return this.syncExternalUserUsecase.execute(
+      SyncExternalUserCommand.create({
+        userId: user._id,
+      })
+    );
+  }
```
LetItRock: ❓ **question:** I guess we want to leave the old functionality for backward compatibility until we have the Web changes implemented? 
Just asking because I'm not sure how you decided to deal with the required changes in the Web app.

libs/dal/package.json
```diff
@@ -20,6 +20,9 @@
   "license": "ISC",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
+  "peerDependencies": {
+    "@nestjs/common": "10.2.2"
+  },
```
LetItRock, line 25: 🗒 **note (non-blocking):** dal used to be independent package :(

LetItRock (commented):
@ChmaraX amazing work on this PR 🙌 
I left you a few questions and some small suggestions ;)

libs/dal/package.json
ChmaraX: The reason is that I need `@Inject` to change the implementation of the affected repositories e.g. `UserRepository` based on the EE flag while maintaining the original location of the UserRepository to avoid changing imports in 100s files:

```typescript
import { Inject } from '@nestjs/common';
import { Types } from 'mongoose';

export class UserRepository implements IUserRepository {
  constructor(@Inject('USER_REPOSITORY') private userRepository: IUserRepository) {}
```

I'm adding this to a TODO list and will think about ways to keep `@novu/dal` independent

apps/api/src/app/auth/community.auth.module.config.ts
ChmaraX: I agree - I'm planning to rework how the ee-auth is injected, more here: https://github.com/novuhq/packages-enterprise/pull/147#discussion_r1667712525

apps/api/src/app/organization/ee.organization.controller.ts
ChmaraX: Yes - currently the organization data resides in 2 places - Clerk and our internal MongoDB. Specifically, `apiServiceLevel`, `branding` and `partnerConfiguration` resides in MongoDB, rest is in Clerk. For that reason I left this to provide a way how to retrieve these data on frontend from API - if we confirm its not needed on frontend, we can remove this.

apps/api/src/app/organization/usecases/create-organization/sync-external-organization/sync-external-organization.usecase.ts
ChmaraX: Good point - adding to TODO list - will open in new PR

apps/api/src/.env.development
```diff
@@ -86,3 +86,6 @@ PR_PREVIEW_ROOT_URL=dev-web-novu.netlify.app
 
 HUBSPOT_INVITE_NUDGE_EMAIL_USER_LIST_ID=
 HUBSPOT_PRIVATE_APP_ACCESS_TOKEN=
+
+CLERK_ISSUER_URL=
+CLERK_LONG_LIVED_TOKEN=
```
SokratisVidros: If not mistaken this is needed for test. Do we have to add it in every `.env` file?

apps/api/src/app/auth/auth.module.ts
```diff
@@ -1,65 +1,24 @@
...
-          })
-        )
-        .forRoutes({
-          path: '/auth/github',
-          method: RequestMethod.GET,
-        });
+    if (isClerkEnabled()) {
+      configureEE(consumer);
+    } else {
+      configureCommunity(consumer);
```
SokratisVidros, line 21: 💯

apps/api/src/app/user/ee.user.controller.ts
ChmaraX: This got replaced by `webhooks-clerk.controller.ts` 

All these Clerk changes will live under a standalone feature flag, therefore everything including web should work as it was until the flag is removed.
When `IS_CLERK_ENABLED = false` the original user controller will be used

apps/api/src/app/auth/ee.auth.module.config.ts
```diff
@@ -0,0 +1,55 @@
...
+      'USER_REPOSITORY',
+      'MEMBER_REPOSITORY',
+      'ORGANIZATION_REPOSITORY',
+    ],
+  };
+}
+
+export function configure(consumer: MiddlewareConsumer) {
+  const eeAuthPackage = require('@novu/ee-auth');
+  if (eeAuthPackage?.configure) {
```
SokratisVidros: Should we throw an error in line 54 if require fails so that the app fail fast?

apps/api/src/app/events/e2e/trigger-multicast.e2e.ts
```diff
@@ -81,14 +81,27 @@ function expectBulkSingleSubscriberStub(
   }
 }
 
+function getEchoGatewayModule() {
```
SokratisVidros, line 84:
```suggestion
function getBridgeGatewayModule() {
```

apps/api/src/app/organization/e2e/sync-organization.e2e-ee.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  it('should create internal organization with externalId of given existing Clerk organization and metadata', async () => {
+    const updatedClerkOrganization: Organization = await clerkClientMock.organizations.getOrganization({
+      organizationId: clerkOrganization.id,
+    });
+
+    expect(internalOrganization._id).to.equal(updatedClerkOrganization.publicMetadata.externalOrgId);
+    expect(internalOrganization.externalId).to.equal(updatedClerkOrganization.id);
+
+    // these are stored in the Clerk organization only and then concated to the response
```
SokratisVidros:
```suggestion
    // these are stored in the Clerk organization only and then concatenated with the response
```

apps/api/src/app/organization/usecases/create-organization/sync-external-organization/sync-external-organization.command.ts
```diff
@@ -0,0 +1,8 @@
+import { IsString, IsDefined, IsEnum, IsOptional } from 'class-validator';
+import { AuthenticatedCommand } from '@novu/application-generic';
+
+export class SyncExternalOrganizationCommand extends AuthenticatedCommand {
+  @IsDefined()
+  @IsString()
+  externalId: string;
```
SokratisVidros, line 7:
```suggestion
  externalOrganizationId: string;
```

libs/application-generic/src/utils/inject-repositories.ts
```diff
@@ -0,0 +1,228 @@
...
+    return new clerkClientMock();
+  }
+}
+
+export function injectRepositories(
+  { repositoriesOnly }: { repositoriesOnly?: boolean } = {
+    repositoriesOnly: true,
+  }
+) {
+  if (!isClerkEnabled()) {
```
SokratisVidros, line 36: **nitpick (non-blocking):** ‏ 

We can start with the `isClerkEnabled` scenario and return early for better readability.

libs/dal/src/repositories/member/member.repository.ts
```diff
@@ -13,193 +10,129 @@ export interface IAddMemberData {
   memberStatus: MemberStatusEnum;
 }
 
-type MemberQuery = FilterQuery<MemberDBModel> & EnforceOrgId;
+export class MemberRepository implements IMemberRepository {
```
SokratisVidros, line 13: To avoid tainting @novu/dal with an extra dependency, I think we can remove this class that picks the underlying repo implementationwe can work with the IMemberRepository in the rest of the code and and replace it with a factory method that exports EE or Community repository.

apps/api/package.json
ChmaraX: Before we had 2 types of tests - "regular" and ee, EE flag didnt modify or swap anything in the regular version so there was no need to run them together

Now we have 3 types - regular,  regular with EE flag enabled = swapped functionality with Clerk, and EE only

We could potentially combine the ee and ee-only together when we remove the feature flag.


image

libs/dal/src/repositories/organization/organization.repository.ts
```diff
@@ -1,118 +1,97 @@
...
-          filter: { _id: orgId, 'partnerConfigurations.configurationId': configurationId },
-          update: {
-            'partnerConfigurations.$.projectIds': unusedOrgIds.includes(orgId) ? [] : data[orgId],
-          },
-        },
-      };
-    });
-
-    return await this.bulkWrite(bulkWriteOps);
+export class OrganizationRepository implements IOrganizationRepository {
```
SokratisVidros, line 6: Ditto about the wrapping class

libs/shared/src/types/auth/auth.types.ts
```diff
@@ -11,6 +13,18 @@ export interface IJwtClaims {
   iss?: string;
 }
 
+// JWT payload + custom claims
+export type ClerkJwtPayload = JwtPayload & {
```
SokratisVidros, line 17: This type definition is a bit suspicious. We have the community and the Clerk JWT payload I'd expect to have either a common interface or transform one to the other. At this point, I'd suggest making the clerk JWT the source of truth and adopt the rest accordingly.

libs/shared/src/types/clerk/clerk.user.types.ts
```diff
@@ -0,0 +1,93 @@
+import { User as ClerkUser } from '@clerk/clerk-sdk-node';
+import { ClerkPaginationRequest } from '@clerk/types';
+import { IServicesHashes } from '../../entities/user';
+import { JobTitleEnum } from '../organization';
+
+export type User = ClerkUser & UserMetadataParams;
+
+export type UpdateUserParams = {
```
SokratisVidros, line 8: Can we get these types from Clerk?

libs/testing/src/user.session.ts
```diff
@@ -281,14 +400,19 @@ export class UserSession {
...
+  // TODO: create EE version
   async switchEnvironment(environmentId: string) {
     const environment = await this.environmentService.getEnvironment(environmentId);
 
     if (environment) {
       this.environment = environment;
       await this.testAgent.post(`/v1/auth/environments/${environmentId}/switch`);
 
-      await this.fetchJWT();
+      if (isClerkEnabled()) {
```
SokratisVidros, line 411: We will merge this PR after the Env work is released so some adaptation is necessary.

SokratisVidros (changes requested):

libs/dal/src/repositories/member/member.repository.ts
ChmaraX: I think this would require changing every reference of `MemberRepository` throughout the code. 
Either way I will try to change how this works by removing the `@Inject` requirement therefore removing `@nestjs/common` dependency from `@novu/dal` and possibly move the affected repositories to `@novu/ee-dal` and create 2x `MemberRepository` with same name one in dal and one in ee-dal - will see how that would work.

libs/shared/src/types/auth/auth.types.ts
ChmaraX: This type will eventually be not needed here - after we implement the JWT env switch it will live only in `clerk.strategy.ts` and not in shared package. 

The only 2 places where this type is used is:
-  `clerk.strategy.ts -> validate(payload: ClerkJwtPayload)` - its what the validate function gets right after decoding the JWT so its a correct type here, can't be transformed before this
- `user.session.ts` -> updateEETokenClaims()` - which is used only for testing when switching tokens and will be eventually removed

libs/shared/src/types/clerk/clerk.user.types.ts
ChmaraX: It seems like the type exported from Clerk is not the same as the update() function accepts:

Whats exported:

```typescript
type UpdateUserParams = {
    username?: string | undefined;
    firstName?: string | undefined;
    lastName?: string | undefined;
    primaryEmailAddressId?: string | undefined;
    primaryPhoneNumberId?: string | undefined;
    primaryWeb3WalletId?: string | undefined;
    unsafeMetadata?: UserUnsafeMetadata | undefined;
```

What the `clerkClient.users.updateUser()` accepts:

```typescript
type UpdateUserParams = {
    firstName?: string;
    lastName?: string;
    username?: string;
    password?: string;
    skipPasswordChecks?: boolean;
    signOutOfOtherSessions?: boolean;
    primaryEmailAddressID?: string;
    primaryPhoneNumberID?: string;
    primaryWeb3WalletID?: string;
    profileImageID?: string;
    totpSecret?: string;
    backupCodes?: string[];
    externalId?: string;
    createdAt?: Date;
    createOrganizationEnabled?: boolean;
} & UserMetadataParams & (UserPasswordHashingParams | object);
```

SokratisVidros: We can report that, but try reusing types as much as possible. Note that, this is not a blocker for merging.

libs/shared/src/types/auth/auth.types.ts
SokratisVidros: Awesome.

apps/api/e2e/setup.ts
```diff
@@ -4,9 +4,23 @@ import sinon from 'sinon';
 import chai from 'chai';
 
 import { bootstrap } from '../src/bootstrap';
+import { isClerkEnabled } from '@novu/shared';
 
 const dalService = new DalService();
 
+async function seedClerkMongo() {
+  if (isClerkEnabled()) {
```
SokratisVidros, line 12: Consider adding a console.log here in case the feature flag is off.

apps/api/package.json
```diff
@@ -71,8 +74,10 @@
     "nanoid": "^3.1.20",
     "nest-raven": "^10.0.0",
     "newrelic": "^9.15.0",
+    "nimma": "^0.6.0",
     "passport": "0.6.0",
     "passport-github2": "^0.1.12",
+    "passport-google-oauth": "^2.0.0",
```
SokratisVidros, line 80: Do we need this package at this point?

apps/api/src/app.module.ts
```diff
@@ -57,6 +61,12 @@ const enterpriseImports = (): Array<Type | DynamicModule | Promise<DynamicModule
     }
   }
 
+  if (isClerkEnabled()) {
```
SokratisVidros, line 64: Sprinkle it with some white labelling

```suggestion
  if (isEEAuthEnabled()) {
```

apps/api/src/app/auth/legacy-ee-auth/auth.controller.ts
```diff
@@ -0,0 +1,42 @@
...
+  Logger,
+} from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+import { ApiException, buildOauthRedirectUrl } from '@novu/application-generic';
+
+@Controller('/auth')
+@UseInterceptors(ClassSerializerInterceptor)
+export class AuthController {
+  @Get('/google')
```
SokratisVidros, line 18: We can remove the google auth completely as it is not officially supported in the community edition. In fact we can remove the whole legacy-ee-auth folder.

.github/workflows/on-pr.yml
```diff
@@ -237,10 +237,11 @@ jobs:
     strategy:
       # The order is important for ee to be first, otherwise outputs not work correctly
       matrix:
-        name: ['novu/api-ee', 'novu/api']
+        name: ['novu/api-ee', 'novu/api', 'novu/api-ee-clerk']
```
SokratisVidros, line 240:
```suggestion
        name: ['novu/api-ee', 'novu/api', 'novu/api-ee-with-cloud-auth']
```

SokratisVidros (approved):

.github/workflows/on-pr.yml
ChmaraX: When Clerk is fully released we should be able to merge it into single `novu/api-ee` tests

PR #5756, opened 2024-06-17
chore(web): Nv 3810 remove information architecture ff

SokratisVidros:

Remove the legacy feature flag and clean up the branching logic in the web app after the latest release.


apps/web/src/components/nav/Sidebar.tsx
```diff
@@ -0,0 +1,37 @@
+import { useLocation } from 'react-router-dom';
```
SokratisVidros, line 1: This is a revised Sidebar component for the post IA era.

apps/web/src/constants/routes.ts
```diff
@@ -42,7 +42,9 @@ export enum ROUTES {
   SETTINGS = '/settings',
   PROFILE = '/settings/profile',
   TEAM_SETTINGS = '/settings/team',
-  BRAND_SETTINGS = '/settings/brand',
+  /* Deprecated */
+  BRAND_SETTINGS_DEPRECATED = '/settings/brand',
```
SokratisVidros, line 46: This is just to align URLs with the copywriting on the sidebar.

apps/web/src/components/docs/docs.const.ts
```diff
@@ -2,6 +2,7 @@ import { ROUTES } from '../../constants/routes';
 
 export const PATHS: { [key in ROUTES]?: string } = {
   [ROUTES.ACTIVITIES]: 'activity-feed/introduction',
+  // TODO: Where is this endpoint after IA?
   [`${ROUTES.BRAND}/layouts`]: 'content-creation-design/layouts',
```
SokratisVidros: @antonjoel82 @BiswaViraj is `/brand/layout` page still around post IA? My understanding is that it's not but I want to double-check with you.

davidsoderberg: This should be removable now as the mapping on the row below it should be the new router for layouts.

apps/web/src/constants/routes.ts
```diff
@@ -42,7 +42,9 @@ export enum ROUTES {
   SETTINGS = '/settings',
   PROFILE = '/settings/profile',
   TEAM_SETTINGS = '/settings/team',
-  BRAND_SETTINGS = '/settings/brand',
+  /* Deprecated */
```
antonjoel82, line 45:
```suggestion
  /* @deprecated */
```

apps/web/src/components/nav/Sidebar.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const Sidebar = () => {
+  const { pathname } = useLocation();
+
+  let Variant = RootNavMenu;
+
+  if (pathname.startsWith(ROUTES.SETTINGS)) {
+    Variant = SettingsNavMenu;
+  } else if (pathname.startsWith(ROUTES.STUDIO)) {
+    Variant = LocalNavMenu;
+  }
```
antonjoel82, line 30: 💭 **thought (non-blocking):** I'm not sure that this revision is an improvement over the original and will just add verbosity through a chain of `if / else if`‏

apps/web/src/pages/settings/SettingsPage.tsx
antonjoel82, line 1: I believe this whole page can be deleted since it's the old, tabs-based settings page

antonjoel82 (approved):
Looks good, thanks!

apps/web/src/pages/settings/SettingsPage.tsx
SokratisVidros: Done

SokratisVidros: Removed

apps/web/src/components/nav/Sidebar.tsx
SokratisVidros: I hear you. I considered adding a local Route component to handle the complicity but it felt too much for now. Settings will change very soon and I don't foresee more nesting in our sidebar but it's definitely something we need to refactor if sidebar becomes more complicated.

PR #5764, opened 2024-06-18
chore(web): Restore layouts page

SokratisVidros:
It was accidentally removed at 2da2cd213051e02fbfd28081ae0ca0f8d269984a. An e2e spec has been also added to prevent this from happening again.


apps/web/src/pages/layouts/v2/LayoutsPage.tsx
antonjoel82, line 1: ❓ **question:** To confirm are we sure we want to remove the `v2` page?‏

antonjoel82 (approved):
Confirming about v2, but otherwise looks good after a quick smoke test

apps/web/src/pages/layouts/v2/LayoutsPage.tsx
SokratisVidros: Yes, why no? Please let me know if I am missing any context.

PR #5768, opened 2024-06-18
chore(api): Remove environmentId from auth JWT

SokratisVidros:
This is preparatory work for the cloud auth provider migration, which will preserve user and organization-only data inside the JWT.

The current environment selection is handled separately and is persisted in local storage. As part of future work, we can store the last environment ID on the user object in Novu to enable cross-device memory.


apps/api/src/app/auth/auth.controller.ts
```diff
@@ -167,6 +167,7 @@ export class AuthController {
     return await this.switchOrganizationUsecase.execute(command);
   }
 
+  // *DEPRECATED*
   @Post('/environments/:environmentId/switch')
```
SokratisVidros, line 167: This endpoint is not necessary as switching is a frontend concern from no one, and no new JWT auth_token is needed on environment switching.

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -204,18 +205,14 @@ export class AuthController {
...
+  async authenticateTest(@Param('userId') userId: string, @Query('organizationId') organizationId: string) {
     if (process.env.NODE_ENV !== 'test') throw new NotFoundException();
 
     const user = await this.userRepository.findById(userId);
     if (!user) throw new BadRequestException('No user found');
 
     const member = organizationId ? await this.memberRepository.findMemberByUserId(organizationId, user._id) : null;
 
-    return await this.authService.getSignedToken(user, organizationId, member as MemberEntity, environmentId);
+    return await this.authService.getUserJWT(user, organizationId, member as MemberEntity);
```
SokratisVidros: Just a renaming to make sure this is not the subscriber JWT used in the notification center

apps/api/src/app/auth/services/passport/jwt.strategy.ts
```diff
@@ -1,25 +1,30 @@
...
   @Instrument()
-  async validate(payload: UserSessionData) {
+  async validate(req: http.IncomingMessage, payload: UserSessionData) {
     const user = await this.authService.validateUser(payload);
     if (!user) {
       throw new UnauthorizedException();
     }
 
-    return payload;
+    const environmentId = req.headers[HttpRequestHeaderKeysEnum.NOVU_ENVIRONMENT_ID.toLowerCase()];
```
SokratisVidros: Read environment id from the corresponding header

apps/api/src/app/environments/environments.controller.ts
```diff
@@ -36,6 +36,7 @@ export class EnvironmentsController {
     private getMyEnvironmentsUsecase: GetMyEnvironments
   ) {}
 
+  // TODO: To be removed.
```
SokratisVidros: Fetching all environments for the current user and org should be enough. This will be removed after a week or so to ensure that all web apps have refreshed their source code.

apps/web/src/components/layout/components/EnsureOnboardingComplete.tsx
```diff
@@ -6,9 +6,14 @@ import { useBlueprint } from '../../../hooks/index';
 export function EnsureOnboardingComplete({ children }: any) {
   useBlueprint();
   const location = useLocation();
-  const { currentOrganization, environmentId } = useAuth();
+  const { currentUser, currentOrganization } = useAuth();
 
-  if ((!currentOrganization || !environmentId) && location.pathname !== ROUTES.AUTH_APPLICATION) {
+  if (!currentUser) {
```
SokratisVidros, line 35: Minor auth fix, in case something goes wrong with the fetch API requests, go to sign-in not /auth/application

libs/dal/src/repositories/environment/environment.repository.ts
```diff
@@ -30,9 +30,12 @@ export class EnvironmentRepository extends BaseRepository<EnvironmentDBModel, En
...
 
   async findOrganizationEnvironments(organizationId: string) {
-    return this.find({
-      _organizationId: organizationId,
-    });
+    return this.find(
+      {
+        _organizationId: organizationId,
+      },
+      { apiKeys: 0 }
```
SokratisVidros: I need to figure out why this still returns apiKeys attributes from Mongo.

apps/web/src/api/api.client.ts
```diff
@@ -84,10 +84,12 @@ function buildUrl(url: string, absoluteUrl: boolean) {
 
 function getHeaders() {
   const token = localStorage.getItem('auth_token');
+  const lastEnvironmentId = localStorage.getItem('novu_last_environment_id');
```
scopsy: Can we throw an error or have a guard that will prevent a situation where environmentId header is missing? Ideally also in the backend. This is extremely important that no API request will be made without the environmentId header to the backend

apps/api/src/app/auth/services/passport/jwt.strategy.ts
scopsy: We should throw an error incase the environmentId is missing here

SokratisVidros: Done.

apps/web/src/api/api.client.ts
SokratisVidros: If not mistaken the EnvironmentCommand requires the environment ID using class validator. If not provided, a 400 will be raised for all controllers that need it. 

[See here](https://github.com/novuhq/novu/pull/5768/files#diff-1b0e3f7797aad06baa5dea4639298a417be5d4b075bb84017056e6430c627b9fR6)

apps/api/src/app/auth/services/passport/jwt.strategy.ts
```diff
@@ -1,25 +1,35 @@
...
-  async validate(payload: UserSessionData) {
-    const user = await this.authService.validateUser(payload);
+  async validate(req: http.IncomingMessage, session: UserSessionData) {
+    const user = await this.authService.validateUser(session);
     if (!user) {
       throw new UnauthorizedException();
     }
 
-    return payload;
+    session.scheme = ApiAuthSchemeEnum.BEARER;
```
SokratisVidros, line 21: The scheme property is used to detect if the API is accessed from JWT or APIKey.

apps/api/src/app/shared/commands/project.command.ts
```diff
@@ -3,18 +3,20 @@ import { BaseCommand } from './base.command';
 
 export { BaseCommand };
 
-export abstract class EnvironmentWithUserCommand extends BaseCommand {
+export abstract class EnvironmentCommand extends BaseCommand {
```
SokratisVidros, line 6: Making the command definitions dryer. The key point here is that the `EnvironmentWithUserCommand` requires environmentId presense and raises an error otherwise.

apps/api/src/app/auth/services/passport/jwt.strategy.ts
SokratisVidros: If not mistaken the EnvironmentCommand requires the environment ID using class validator. If not provided, a 400 will be raised for all controllers that need it. 

[See here](https://github.com/novuhq/novu/pull/5768/files#diff-1b0e3f7797aad06baa5dea4639298a417be5d4b075bb84017056e6430c627b9fR6)

Moreover, we can't throw an error at such a centralized place, as not all API routes require environment ID. That is, to add it here, we need a list of endpoints that need it or not needed and act accordingly. But that's a maintenance headache.

apps/web/src/components/nav/EnvironmentSelect/EnvironmentSelect.tsx
```diff
@@ -1,55 +1,52 @@
...
-export const EnvironmentSelectRenderer: React.FC<ReturnType<typeof useEnvironmentSelect>> = ({
-  icon,
-  isPopoverOpened,
-  setIsPopoverOpened,
-  handlePopoverLinkClick,
-  readonly,
-  ...selectProps
-}) => {
-  return (
-    <EnvironmentPopover
```
SokratisVidros, line 16: The popover about changes is not relevant in V2 anymore and was a bit buggy in V1. So, it was removed.

apps/web/src/components/nav/Sidebar.tsx
```diff
@@ -16,17 +15,5 @@ export const Sidebar = () => {
     return null;
   }
 
-  let Variant = RootNavMenu;
-
-  if (pathname.startsWith(ROUTES.SETTINGS)) {
-    Variant = SettingsNavMenu;
-  } else if (pathname.startsWith(ROUTES.STUDIO)) {
```
SokratisVidros, line 23: The env switcher is not relevant in Studio anymore.

apps/api/src/app/auth/auth.controller.ts
```diff
@@ -156,16 +156,18 @@ export class AuthController {
   @Header('Cache-Control', 'no-store')
   async organizationSwitch(
     @UserSession() user: UserSessionData,
-    @Param('organizationId') organizationId: string
-  ): Promise<string> {
+    @Param('organizationId') organizationId: string,
+    @Req() request: Request
```
denis-kralj-novu: ☑ **chore:** ‏ added parameter is not used

apps/api/src/app/auth/auth.controller.ts
denis-kralj-novu: ❓ **question:** ‏ How would removal of this affect community version?

.source
denis-kralj-novu, line 1: ☑ **chore:** ‏ Please link the relevant EE MR, merge that first and then update the reference in this PR to avoid having devs pull latest dev with EE link to a feature branch commit

apps/api/src/app/auth/services/passport/jwt.strategy.ts
```diff
@@ -10,16 +11,31 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
...
   @Instrument()
-  async validate(payload: UserSessionData) {
-    const user = await this.authService.validateUser(payload);
+  async validate(req: http.IncomingMessage, session: UserSessionData) {
+    // Set the scheme to Bearer, meaning the user is authenticated via a JWT coming from Dashboard
+    session.scheme = ApiAuthSchemeEnum.BEARER;
+
+    // Fetch the environmentId from the request header
+    const environmentIdFromHeader =
+      (req.headers[HttpRequestHeaderKeysEnum.NOVU_ENVIRONMENT_ID.toLowerCase()] as string) || '';
```
denis-kralj-novu, line 25: ✏ **suggestion:** ‏ if the header needs to go `toLowerCase()`, maybe save it as lower case instead? This is to avoid constant conversion whenever we call this particular header.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.command.ts
```diff
@@ -1,7 +1,13 @@
-import { IsNotEmpty } from 'class-validator';
-import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
+import { IsNotEmpty, IsOptional } from 'class-validator';
+import { BaseCommand } from '../../../shared/commands/base.command';
 
-export class GetMyEnvironmentsCommand extends EnvironmentWithUserCommand {
+export class GetMyEnvironmentsCommand extends BaseCommand {
```
denis-kralj-novu, line 4: ✏ **suggestion:** ‏ the `EnvironmentCommand` would be a better fit given the requirements here instead of `BaseCommand`

apps/web/tests/subscribers.spec.ts
```diff
@@ -18,5 +18,5 @@ test('should display subscribers list', async ({ page }) => {
     'Created At',
     'Data',
   ]);
-  await subscribersPage.assertSubscribersTableRowCount(1);
+  await subscribersPage.assertSubscribersTableRowCount(2);
```
denis-kralj-novu, line 21: ❓ **question:** ‏ How did the nature of this test change?

apps/api/src/app/auth/auth.controller.ts
SokratisVidros: Removed

.source
SokratisVidros: Correct. This is right before we merge

apps/api/src/app/auth/services/passport/jwt.strategy.ts
SokratisVidros: This is true and has to be applied across headers. Browser sets it titleized, Nest does the normalization so in some cases we need both versions.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.command.ts
SokratisVidros: The environmentId has a required environment id, whereas the `GetMyEnvironmentsCommand` has an optional environmentId. Is this what you are talking about?

apps/web/tests/subscribers.spec.ts
SokratisVidros: That's a great question. I will delve deeper.

apps/api/src/app/environments/usecases/get-my-environments/get-my-environments.command.ts
denis-kralj-novu: Ah, I didn't notice that the optional detail, thanks for pointing that out

libs/shared/src/types/http/utils.types.spec.ts
```diff
@@ -1,6 +1,16 @@
...
+describe('HTTP headers', () => {
+  /**
+   * This describe block resolves the Jest error of a test suite not having any tests.
+   * It has no other purpose.
+   */
+  it('tests the Typescript compiler errors below', () => {
+    expect(true).toEqual(true);
+  });
+});
+
```
denis-kralj-novu, line 13: ✏ **suggestion:** ‏ We could rename the file to not end in `.spec.ts`.

apps/api/src/app/auth/services/passport/jwt.strategy.ts
```diff
@@ -10,16 +11,31 @@ export class JwtStrategy extends PassportStrategy(Strategy) {
...
       jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
       secretOrKey: process.env.JWT_SECRET,
+      passReqToCallback: true,
     });
   }
 
   @Instrument()
-  async validate(payload: UserSessionData) {
-    const user = await this.authService.validateUser(payload);
+  async validate(req: http.IncomingMessage, session: UserSessionData) {
```
ChmaraX, line 19: 🤓 **nitpick (non-blocking):** ‏We should probably standardize on using a single type for the request object. It appears that Request from the global namespace is the most commonly used type.

libs/shared/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "@clerk/clerk-sdk-node": "^5.0.9",
+    "@nestjs/swagger": "^7.1.8",
```
denis-kralj-novu, line 42: ❓ **question:** How did this even work with the import missing? I'm guessing some other package pulled it in :/

libs/application-generic/src/usecases/switch-organization/switch-organization.command.ts
```diff
@@ -1,4 +1,4 @@
-import { IsNotEmpty } from 'class-validator';
+import { IsNotEmpty, IsOptional } from 'class-validator';
```
denis-kralj-novu: ☑ **chore:** ‏ Import not needed

apps/api/src/app/environments/environments.controller.ts
```diff
@@ -20,6 +20,7 @@ import { UpdateEnvironmentRequestDto } from './dtos/update-environment-request.d
 import { ApiCommonResponses, ApiResponse } from '../shared/framework/response.decorator';
 import { UserAuthentication } from '../shared/framework/swagger/api.key.security';
 import { SdkGroupName, SdkMethodName } from '../shared/framework/swagger/sdk.decorators';
+import Api from 'twilio/lib/rest/Api';
```
ChmaraX: ❓ **question:** It doesn't seem to be used?‏

```suggestion
```

apps/web/src/api/api.client.ts
```diff
@@ -87,8 +87,14 @@ function buildUrl(url: string, absoluteUrl: boolean) {
...
   // TODO: change the way we get the clerk token
   const token = getToken();
-
-  return { Authorization: `Bearer ${token}` };
+  const lastEnvironmentId = getEnvironmentId();
+
+  return token
+    ? {
+        Authorization: `Bearer ${token}`,
+        'Novu-Environment-Id': lastEnvironmentId || '',
```
ChmaraX, line 95: ✏ **suggestion:** `getToken()` function returns either token or empty string - we could unify these two functions

libs/shared/package.json
SokratisVidros: Correct. It's because I moved header utils into shared.

apps/api/src/app/environments/environments.controller.ts
SokratisVidros: removed

apps/web/src/components/providers/CommunityAuthProvider.tsx
```diff
@@ -1,23 +1,278 @@
...
+import { getUser } from '../../api/user';
+import { switchOrganization as apiSwitchOrganization, getOrganizations } from '../../api/organization';
+import { type AuthContextValue } from './AuthProvider';
 
-export const CommunityAuthContext = createContext<ReturnType<typeof useCreateAuthCommunity>>({
+// TODO: Add a novu prefix to the local storage key
+export const LOCAL_STORAGE_AUTH_TOKEN_KEY = 'auth_token';
+
+const noop = () => {};
+const asyncNoop = async () => {};
```
ChmaraX, line 22: 👏 **praise:** Good idea‏

apps/web/src/components/providers/EnvironmentProvider.tsx
```diff
@@ -0,0 +1,178 @@
...
+        await switchEnvironment({
+          environmentId,
+          redirectUrl,
+        });
+      } else {
+        throw new Error('Development environment not found');
+      }
+    },
+    [environments, switchEnvironment]
+  );
```
ChmaraX, line 143: ✏ **suggestion:** ‏I think we could generalize this, even if the intent is to have two separate functions for specific env switch, e.g. something like:

```typescript
const switchToEnvironment = useCallback(
  async (environmentName: BaseEnvironmentEnum, redirectUrl?: string) => {
    const environmentId = environments?.find((env) => env.name === environmentName)?._id;

    if (environmentId) {
      await switchEnvironment({
        environmentId,
        redirectUrl,
      });
    } else {
      throw new Error(`${capitalize(environmentName.toLowerCase())} environment not found`);
    }
  },
  [environments, switchEnvironment]
);


const switchToProductionEnvironment = (redirectUrl?: string) => switchToEnvironment(BaseEnvironmentEnum.PRODUCTION, redirectUrl);
const switchToDevelopmentEnvironment = (redirectUrl?: string) => switchToEnvironment(BaseEnvironmentEnum.DEVELOPMENT, redirectUrl);
```

libs/testing/src/user.session.ts
```diff
@@ -241,13 +239,16 @@ export class UserSession {
...
+      .set('Authorization', this.token)
+      .set('Novu-Environment-Id', this.environment._id);
   }
 
   private async fetchJwtEE() {
     await this.updateEETokenClaims({
       externalId: this.user ? this.user._id : '',
       externalOrgId: this.organization ? this.organization._id : '',
+      // TODO: @adamchmara, is this necessary anymore?
       environmentId: this.environment ? this.environment._id : '',
```
ChmaraX: ☑ **chore:** the `environmentId` line shouldn't be needed anymore if its also removed from community.

ChmaraX (approved):

PR #5769, opened 2024-06-18
feat(novui): Select & Checkbox components

antonjoel82:

- Add Select and Checkbox components based on Input
- Add global `colorPalette` arg to Storybook
- Styling tweaks

### Screenshots

https://github.com/novuhq/novu/assets/47441786/ab7382cf-a3bf-4217-b39e-02f5ffd1d1b4

https://github.com/novuhq/novu/assets/47441786/de873fd3-65d2-4856-9b76-99f3ff844b4f


libs/novui/src/components/button/Button.tsx
```diff
@@ -21,7 +20,7 @@ type IconButtonDefaultElement = 'button';
 
 export type IButtonProps<C extends React.ElementType = IconButtonDefaultElement> = PolymorphicComponentPropWithRef<
```
SokratisVidros: Since this is a type and not an interface, there is no need for the `I` prefix.
```suggestion
export type ButtonProps<C extends React.ElementType = IconButtonDefaultElement> = PolymorphicComponentPropWithRef<
```

libs/novui/src/components/checkbox/Checkbox.tsx
```diff
@@ -1 +1,41 @@
...
+import { Checkbox as ExternalCheckbox } from '@mantine/core';
+import { ChangeEventHandler, forwardRef } from 'react';
+import { PolymorphicRef } from '../../types/props-helpers';
+import { css, cx } from '../../../styled-system/css';
+import { splitCssProps } from '../../../styled-system/jsx';
+import { checkbox } from '../../../styled-system/recipes';
+import { JsxStyleProps } from '../../../styled-system/types';
+import { CoreProps, LocalizedMessage, LocalizedString } from '../../types';
+
+export interface ICheckboxProps extends JsxStyleProps, CoreProps {
```
SokratisVidros: Are we using interfaces or types for props? We need to stick to one.

libs/novui/src/components/select/Select.types.ts
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
SokratisVidros: Let's move the types into the Select.tsx file. There is no benefit from having them in a separate file.

SokratisVidros (approved):

libs/novui/src/components/select/Select.tsx
```diff
@@ -1 +1,39 @@
-export { Select } from '@mantine/core';
+import { Loader, Select as ExternalSelect } from '@mantine/core';
+import { forwardRef } from 'react';
+import { IconArrowDropDown } from 'src/icons';
```
ainouzgali: ☑ **chore:** there are imports of src/.. instead of ../icons/.. , it fails to build :)‏

libs/novui/src/recipes/checkbox.recipe.ts
```diff
@@ -0,0 +1,59 @@
+import { CheckboxStylesNames } from '@mantine/core';
+import { defineSlotRecipe } from '@pandacss/dev';
+import { colorPaletteGradientHorizontal } from 'src/ingredients';
```
ainouzgali: Here as well

ainouzgali (approved):
Thanks Joel!!

PR #5770, opened 2024-06-18
feat(node): add support for bridge URL submission and default value

denis-kralj-novu:
add support for bridge URL submission and default value


packages/node/src/lib/events/events.ts
```diff
@@ -8,6 +8,10 @@ import { WithHttp } from '../novu.interface';
 
 export class Events extends WithHttp implements IEvents {
   async trigger(workflowIdentifier: string, data: ITriggerPayloadOptions) {
+    if (!data.bridgeUrl && process.env.NEXT_PUBLIC_VERCEL_URL) {
+      data.bridgeUrl = `https://${process.env.NEXT_PUBLIC_VERCEL_URL}/api/echo`;
```
rifont: Echo -> novu

denis-kralj-novu: whoops, what a mistake 😅

denis-kralj-novu: fixed

SokratisVidros (approved):

PR #5779, opened 2024-06-19
feat(web): add docs for step types in workflow floating menu

davidsoderberg:


apps/web/src/ee/translations/pages/TranslationGroupsPage.tsx
```diff
@@ -45,9 +47,12 @@ export const TranslationGroupsPage = () => {
...
     <PageContainer>
       <Group position="apart" mx={24} mt={24} mb={16} align="center">
-        <Title size={2} data-test-id="translation-title">
-          Translations
-        </Title>
+        <Flex align="center" gap="75">
+          <Title size={2} data-test-id="translation-title">
+            Translations
+          </Title>
+          <DocsButton />
```
davidsoderberg, line 54: This one was not added during when I created the docs button but now when ee web is moved here this is easy and possible.

apps/web/src/studio/components/workflows/node-view/WorkflowFloatingMenu.tsx
```diff
@@ -14,63 +14,74 @@ import {
...
-    alert('This should open docs for ' + stepType);
+  const [docsOpen, setDocsOpen] = useState<boolean>(false);
+  const [path, setPath] = useState<string>('');
+
+  const toggle = () => {
+    setDocsOpen((prevOpen) => !prevOpen);
+  };
+
+  const handleClick = (pathToSet: string) => () => {
+    setPath('echo/steps/' + pathToSet);
```
davidsoderberg, line 31: To make it easy for use to switch when we have changed echo in our docs path

SokratisVidros: What does this `setPath` do? Is it a navigation?

davidsoderberg: It sets the path of what docs the modal should render.

SokratisVidros (approved):

PR #5783, opened 2024-06-19
fix(api): create in-app integrations when new org is created

jainpawan21:
In-app integrations should be created and active by default when user sign up and new org is created

### Screenshots

<details>

</details>


apps/api/src/app/integrations/usecases/create-novu-integrations/create-novu-integrations.usecase.ts
```diff
@@ -89,8 +89,24 @@ export class CreateNovuIntegrations {
...
+        name: 'Novu In-App',
+        active: true,
+        check: false,
+        userId: command.userId,
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+      })
+    );
+  }
+
```
ainouzgali: Let's also add a count check to be extra sure there is no existing in app integration (only one is allowed per environment)

ainouzgali (commented):
@jainpawan21 left you a comment :) 
Please also run the web tests, to see none are failing there. Maybe there is a test that adds an in-app (which they won't be able to do anymore)

apps/api/src/app/integrations/usecases/create-novu-integrations/create-novu-integrations.usecase.ts
SokratisVidros: Good call Gali, we need to ensure one in-app integration.

SokratisVidros (approved):

ainouzgali (approved):
🚀

PR #5784, opened 2024-06-19
feat(novui): JsonSchemaForm styling

antonjoel82:

- Style JsonSchemaForm
- Add some other small styling tweaks

**Note**: _Designs do not provide comprehensive examples, nor use typical strategies for spacing in some places, so this is a best attempt. We should pass this by Nikolay for further review._

### Screenshots

Screenshot 2024-06-19 at 4 14 50 PM

Screenshot 2024-06-19 at 4 09 30 PM


libs/novui/src/tokens/semanticColors.tokens.ts
```diff
@@ -189,21 +189,21 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
   mode: {
     cloud: {
       start: {
-        value: { base: '{colors.legacy.gradientStart}', _dark: '{colors.legacy.gradientStart}' },
+        value: { base: '{colors.legacy.gradientEnd}', _dark: '{colors.legacy.gradientEnd}' },
         type: 'color',
       },
       end: {
-        value: { base: '{colors.legacy.gradientEnd}', _dark: '{colors.legacy.gradientEnd}' },
+        value: { base: '{colors.legacy.gradientStart}', _dark: '{colors.legacy.gradientStart}' },
```
antonjoel82, line 196: 🗒 note (non-blocking): These are flipped this way in our primary gradient 🙃 ‏

scopsy (approved):
Fantastic Joel! Thank you. Will be playing with more realistic customer usecases to see how design fits. And will collaborate with Nik in case we need to support some usecases

ainouzgali (approved):
Amazing @antonjoel82, it looks great!!

SokratisVidros (approved):
💯

libs/novui/src/json-schema-components/JsonSchemaForm.tsx
```diff
@@ -1,68 +1,72 @@
-import React from 'react';
-import { RJSFSchema } from '@rjsf/utils';
+import Form, { FormProps } from '@rjsf/core';
```
SokratisVidros, line 1: I'd argue that the JSONSchemaForm should be part of our UI library, as it is very specific to the studio part of the web app but we can debate on that later.

antonjoel82: Sounds good!

PR #5793, opened 2024-06-20
feat(web): add mixpanel session recording

davidsoderberg:


apps/web/src/components/providers/SegmentProvider.tsx
```diff
@@ -9,8 +12,27 @@ type Props = {
...
 export const SegmentProvider = ({ children }: Props) => {
+  const isV2ExperienceEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V2_EXPERIENCE_ENABLED);
   const segment = React.useMemo(() => new SegmentService(), []);
 
+  useEffect(() => {
+    if (!segment._mixpanelEnabled) {
+      return;
+    }
+
+    if (isV2ExperienceEnabled) {
```
scopsy: This can be enabled for all regarding v2 feature flag :)

apps/web/src/pages/studio-onboarding/index.tsx
```diff
@@ -24,6 +24,7 @@ export const StudioOnboarding = () => {
   }, [url, testEndpoint]);
 
   useEffect(() => {
+    mixpanel.start_session_recording();
```
scopsy: I think this can just happen on app bootstrap, not specific to onboarding :)

scopsy:
@davidsoderberg did you had a chance of how this work with combination of the Segment client?
I'm just concerned they will be conflicting with each other

davidsoderberg:
> @davidsoderberg did you had a chance of how this work with combination of the Segment client? I'm just concerned they will be conflicting with each other

I have tested it with identify and track and both seems to work.

scopsy:
Awesome!

scopsy (approved):

PR #5798, opened 2024-06-20
feat(web): add sync and update bridge button

denis-kralj-novu:

- Remove unused usecases
- Remove unused code
- Update naming where needed
- Add validate bridge endpoint
- Add buttons and relevant modals
- Add Code component to novui
- Update .source reference to latest next

## Joel's additions

- Add status indicator with 3 states and hover state
- Add hook for connection status
- Add opt-in solution for avoiding layout shift with inputs
- Separate modal content from triggers
- Fix a couple styling issues in web and novui components 
- Added work-in-progress `CodeBlock` component
  - _After brief discussion with Dima, we decided we can use the existing one from web for now since Mantine's new code block does not have line numbers built-in and would require time-consuming style tweaks to work across dark and light mdoe_

### Screenshots

Screenshot 2024-06-25 at 11 12 52 PM

Screenshot 2024-06-25 at 11 13 20 PM

Screenshot 2024-06-25 at 11 13 30 PM

### Related enterprise PR
[Dependency PR](https://github.com/novuhq/packages-enterprise/pull/152)


apps/api/src/app/validate/usecases/validate-bridge-url/validate-bridge-url.usecase.ts
denis-kralj-novu, line 1: server-side novu endpoint validation

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+          label={'Novu Endpoint URL'}
+          description={'Specify the full URL including the protocol and novu endpoint path'}
+          onChange={onBridgeUrlChange}
+          value={inputUrl}
+          disabled={isLoading}
+          error={urlError}
+        />
+        <HStack justify={'space-between'}>
+          <Flex align="center">
+            {/* TODO: how do I set the path of the docs? */}
```
denis-kralj-novu: I'm not really sure how to configure the docs button to send me to the correct location

apps/web/src/components/layout/components/v2/HeaderNav.tsx
```diff
@@ -33,6 +35,7 @@ export function HeaderNav() {
     >
       {/* TODO: Change position: right to space-between for breadcrumbs */}
       <HStack flexWrap={'nowrap'} justifyContent="flex-end" gap={'100'}>
+        {isV2ExperienceEnabled && <BridgeMenuItems />}
```
denis-kralj-novu, line 38: hide this behind feature flag

apps/web/src/components/layout/components/v2/SyncInfoModal.tsx
```diff
@@ -0,0 +1,63 @@
+import { Modal, Title, colors } from '@novu/design-system';
+import { Button, Code } from '@novu/novui';
+import { Tabs, Text } from '@novu/novui';
+
+import { IconOutlineCloudUpload, IconPencil, IconLink } from '@novu/novui/icons';
+import { useState } from 'react';
+import { useEnvironment } from '../../../../hooks/useEnvironment';
+import { useApiKeysPage } from '../../../../pages/settings/ApiKeysPage/useApiKeysPage';
+import { getBridgeUrl } from './utils';
+// TODO: make code copy paste
```
denis-kralj-novu: The content of the code block is copy pastable, but a convenience button might be wanted as well

apps/web/src/components/layout/components/v2/SyncInfoModal.tsx
denis-kralj-novu, line 1: In general, this Sync info modal seems a bit strange when working in local, because the info shared there isn't actually applicable for the local environment (there isn't an API key, the novu endpoint only exists while the local tunnel is open. I'm wondering if this button + modal should even exist in local?

When in dev and prod, it pulls in the correct apiKey and bridge url into the code from the environment

libs/novui/src/components/code/Code.tsx
denis-kralj-novu, line 1: quick exposing of code component

apps/api/src/app/validate/validate.controller.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ExternalApiAccessible } from '../auth/framework/external-api.decorator';
+import { ApiCommonResponses, ApiResponse } from '../shared/framework/response.decorator';
+import { UserAuthentication } from '../shared/framework/swagger/api.key.security';
+import { ValidateBridgeUrlRequestDto } from './dtos/validate-bridge-url-request.dto';
+import { ValidateBridgeUrlResponseDto } from './dtos/validate-bridge-url-response.dto';
+import { ValidateBridgeUrlCommand } from './usecases/validate-bridge-url/validate-bridge-url.command';
+import { ValidateBridgeUrlUsecase } from './usecases/validate-bridge-url/validate-bridge-url.usecase';
+
+@ApiCommonResponses()
+@Controller('/validate')
```
scopsy: Should we maybe just have this on the existing bridge controller instead of a new resourcE?

denis-kralj-novu: makes sense, didn't think to look at the ee controllers, will move it

scopsy: CleanShot 2024-06-24 at 09 25 23@2x
There is an existing `status` endpoint 😅

denis-kralj-novu: ye, but this one comes from the environment, the case I have supplies the endpoint. I will see if there is a a way to consolidate if needed

scopsy:
CleanShot 2024-06-24 at 09 48 10@2x

A few things I've noticed:
- Modal background is not covering the top navbar
- Buttons are not matching design (color and margins)
- Text colors do not match the design

This is probably due to a mistake in the figma and Linear ticket specification, but the "Update endpoint" should only appear on hover over the "Connected" status icon.
CleanShot 2024-06-24 at 09 51 36@2x

On hover state should also match the design:
CleanShot 2024-06-24 at 09 51 49@2x

scopsy:
CleanShot 2024-06-24 at 10 05 40@2x
After clicking on update, I think we can close the modal and have a `showNotification` with success saying that the URl was updated. Wdyt?

scopsy:
CleanShot 2024-06-24 at 10 18 59@2x
Sync changes also appear to not match the UI on Figma. (Colors, spacing, background, etc...)

apps/web/src/components/layout/components/v2/SyncInfoModal.tsx
denis-kralj-novu: @scopsy  would like to hear your thoughts on this

apps/web/src/components/layout/components/v2/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+export const getBridgeUrl = (env: IEnvironment | undefined, pathname: string) => {
+  let bridgeUrl: string | null | undefined;
+  if (isLocalEnv(pathname)) {
+    bridgeUrl = getTunnelUrl();
+  } else {
+    bridgeUrl = env?.echo?.url;
+  }
+
+  return bridgeUrl;
+};
```
antonjoel82:
```suggestion
export const getBridgeUrl = (env: IEnvironment | undefined, pathname: string) => {
  return isLocalEnv(pathname) ? getTunnelUrl() : env?.echo?.url;
};
```

apps/web/src/components/layout/components/v2/utils.ts
```diff
@@ -0,0 +1,15 @@
+import { IEnvironment } from '@novu/shared';
+import { getTunnelUrl } from '../../../../api/bridge/utils';
+
+export const isLocalEnv = (pathname: string) => pathname.includes('/studio');
```
antonjoel82: ☑ **chore:** We already have a util for this, so please use that -- I believe it's called `isStudioRoute`‏

libs/novui/src/components/code/Code.tsx
```diff
@@ -0,0 +1,7 @@
+import { CodeProps as ExternalCodeProps, Code as ExternalCode } from '@mantine/core';
```
antonjoel82: 👏 **praise:** Nice, thank you!‏

libs/novui/src/icons/icon-registry.ts
```diff
@@ -4344,6 +4344,8 @@ import {
   MdOutlineToggleOff as IconOutlineToggleOff,
 } from 'react-icons/md';
 
+import { HiPencil as IconPencil } from 'react-icons/hi2';
```
antonjoel82: ☑ **chore:** Please remove this. We should only use Icons included by default. I think you may be looking for `IconOutlineEdit` or `IconEdit`‏

apps/web/src/components/layout/components/v2/SyncInfoModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const title = () => {
+    return (
+      <>
+        <Title size={2}>Sync changes to the Environment</Title>
+        <Text variant="secondary" className={css({ marginBottom: '16px' })}>
+          Run the following command to publish changes to the desired environment:
+        </Text>
+      </>
+    );
+  };
```
antonjoel82:
```suggestion
  const title = (
      <Stack gap="100">
        <Title size={2}>Sync changes to the Environment</Title>
        <Text variant="secondary">
          Run the following command to publish changes to the desired environment:
        </Text>
      </Stack>
    );
  };
```

apps/web/src/components/layout/components/v2/SyncInfoModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const githubCodeBlock = (
+    <Prism className={codeBlockCss} withLineNumbers language="yaml">
+      {githubYamlContent}
+    </Prism>
+  );
+  const otherCICodeBlock = (
+    <Prism className={codeBlockCss} withLineNumbers language="bash">
+      {otherCICodeContent}
+    </Prism>
+  );
```
antonjoel82: ☑ **chore:** Extract outside of component‏

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+        onClose={toggleBridgeUpdateModalShow}
+      >
+        {/* TODO: is there a better way to add empty space for the error message (so the modal doesn't resize) */}
+        <Input
+          label={'Endpoint URL'}
+          onChange={onBridgeUrlChange}
+          value={inputUrl}
+          disabled={isLoading}
+          error={urlError}
+          className={css({ marginBottom: '16px' })}
```
antonjoel82: ☑ **chore:** Change how we do this to ensure proper spacing and token usage‏

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const buildBridgeUpdateButtonTooltipText = (env: IEnvironment | undefined) => {
+    const bridgeUrl = getBridgeUrl(env, location.pathname);
+
+    if (!bridgeUrl) {
+      return 'No Novu endpoint defined!';
+    }
+
+    return (
+      <HStack>
+        <IconLink></IconLink>
```
antonjoel82:
```suggestion
        <IconLink />
```

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { setTunnelUrl } from '../../../../api/bridge/utils';
+import { useEnvironment } from '../../../../hooks/useEnvironment';
+import { IconPencil, IconLink } from '@novu/novui/icons';
+import { useLocation } from 'react-router-dom';
+
+export function BridgeUpdateModal() {
+  const [inputUrl, setInputUrl] = useState('');
+  const [isUpdating, setIsUpdating] = useState(false);
+  const [urlError, setUrlError] = useState('');
+  const [showBridgeUpdateModal, setShowBridgeUpdateModal] = useState(false);
```
antonjoel82:
```suggestion
  const [inputUrl, setInputUrl] = useState<string>('');
  const [isUpdating, setIsUpdating] = useState<boolean>(false);
  const [urlError, setUrlError] = useState<string>('');
  const [showBridgeUpdateModal, setShowBridgeUpdateModal] = useState<boolean>(false);
```

antonjoel82 (commented):
Leaving comments before I take the reins!

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
antonjoel82: @davidsoderberg Sounds like we could use your help here please!

apps/web/src/components/layout/components/v2/BridgeUpdateModalTrigger.tsx
antonjoel82, line 1: @scopsy @denis-kralj-novu Will need to connect with y'all (and possibly Nik) to determine how to fill-in the FIXMEs here

libs/novui/src/components/code-block/CodeBlock.tsx
```diff
@@ -0,0 +1,54 @@
...
+        ref={ref}
+        component={Component}
+        classNames={classNames}
+        className={cx(css(cssProps), className)}
+        code={code}
+        {...otherProps}
+      />
+    );
+  }
+);
```
antonjoel82, line 54: 🗒 **note (non-blocking):** Unfortunately this seemed to be a bit of a frustrating time sink for the time being because Mantine's new supported syntax-highlighted code block library (Highlight) does not have quite the same functionality as the old Prism -- at least in terms of line numbers. For now, I added a deprecated tag so that we don't use it yet, but most of the structure and composition of it will remain the same regardless of how we choose to show these code blocks.

libs/novui/src/recipes/input.recipe.ts
```diff
@@ -97,6 +99,30 @@ export const INPUT_RECIPE = defineSlotRecipe({
...
-    wrapper: {},
+    wrapper: {
+      // prevent Mantine from injecting 2px margin that causes layout shift
+      marginBottom: '0 !important',
+    },
+  },
+  variants: {
+    variant: {
+      // TODO: determine if we want this built-in! Prevents layout shift with error states
+      preventLayoutShift: {
```
antonjoel82, line 110: 🗒 **note (non-blocking):** This addresses a UX concern with Inputs shifting when an error message appears / disappears. However, it requires saving space for the error to accomplish this, so it adds some additional white space below each input if this variant is used. I chose to do this as a variant for the time being because it's "opt-in" so we can pick and choose for the time being where we're most concerned about the layout shift‏

PR #5808, opened 2024-06-24
feat(web, novui): add modal component novui

jainpawan21:

Add Modal component in novui
### Screenshots

<details>

</details>


libs/novui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,26 @@
+import { Modal as ExternalModal, ModalProps as ExternalModalProps } from '@mantine/core';
+import { CoreProps } from '../../types';
+import { modal } from '../../../styled-system/recipes';
+import { splitCssProps } from '../../../styled-system/jsx';
+import { css, cx } from '../../../styled-system/css';
+
+interface IModalProps extends ExternalModalProps, CoreProps {
```
antonjoel82: ☑ **chore:** ‏Please be careful about directly extending Mantine props, and instead use `Pick<>` to select specific ones. Otherwise we couple ourselves and expose too many options that allow for deviating from the design system

libs/novui/src/components/modal/Modal.tsx
```diff
@@ -0,0 +1,26 @@
...
+  opened: boolean;
+  onClose: () => void;
+}
+
+export const Modal = ({ children, ...props }: IModalProps) => {
```
antonjoel82, line 14:
```suggestion
/**
* Base modal container. Please use Modal.Header and Modal.Body for the component's children.
*/
export const Modal = ({ children, ...props }: IModalProps) => {
```

libs/novui/src/recipes/modal.recipe.ts
```diff
@@ -0,0 +1,50 @@
...
+    root: {
+      colorPalette: 'mode.cloud',
+      borderRadius: '75',
+      position: 'relative',
+      padding: '50',
+      color: 'typography.text.main',
+      width: 'modal.width !important',
+    },
+    inner: {
+      width: '100%',
```
antonjoel82:
```suggestion
      width: 'full',
```

libs/novui/src/recipes/modal.recipe.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+    header: {
+      background: 'modal.background !important',
```
antonjoel82, line 23: 👏 **praise:** Nice job with using a new token here!‏

libs/novui/src/tokens/semanticRadius.tokens.ts
```diff
@@ -1,9 +1,9 @@
 import { defineSemanticTokens } from '@pandacss/dev';
 
 /**
- * Represents the size of an element.
+ * Represents the size of an radius.
  *
- * Used for properties like width and height.
+ * Used for properties like radius.
```
antonjoel82, line 6: 👏 **praise:** Good catch, thank you!‏

libs/novui/src/tokens/semanticSizes.tokens.ts
```diff
@@ -15,6 +15,28 @@ export const SEMANTIC_SIZES_TOKENS = defineSemanticTokens.sizes({
...
     },
   },
+  modal: {
+    width: {
+      DEFAULT: {
+        value: '800px',
+        type: 'sizes',
+      },
+      max: {
+        value: '800px',
```
antonjoel82, line 25: ❓ **question:** Do we want both of these values to be the same and do we need both?‏

libs/novui/src/recipes/modal.recipe.ts
```diff
@@ -0,0 +1,50 @@
...
+      lineHeight: '125',
+      marginBottom: '25',
+    },
+    content: {
+      background: 'modal.background !important',
+      boxShadow: 'medium !important',
+    },
+    overlay: {
+      background: 'surface.page !important',
+      opacity: '40 !important',
```
antonjoel82: ☑ **chore:** Please extract to a `semanticOpacity` token for `overlay`‏

antonjoel82 (commented):
Great progress, Pawan! Could you please add some Storybook stories as we've done with other components? This is the recommended way to develop these in isolation for quick dev and to be able to demo them.

libs/novui/src/tokens/semanticSizes.tokens.ts
jainpawan21: We can have both values but different. Currently these values are kept same for simplicity. Later I wil handle modal size

antonjoel82 (commented):
Hey Pawan, thanks for adding the Storybook and other updates!

Based on the Storybook, things don't quite match designs yet. I think we need some more changes here to make this helpful for devs.

Screenshot 2024-07-03 at 7 07 43 AM
Screenshot 2024-07-03 at 7 08 13 AM


Some things to note or think about:

1. Title text doesn't match
2. Padding is too small
3. To provide value, we may want to add some other things based on our sample designs that build behavior in for developers such as:
  - a subtitle prop that has opinionated styling within the component
  - a way to add buttons (that conforms to designs)

Let me know if you have any questions!

PR #5820, opened 2024-06-25
feat(js): Scope variables under class

desiprisg:

We need to scope the css variables under a unique class for each `InboxUI` instance. This might not be necessary for most simple use cases.

We also make sure that one stylesheet per `InboxUI` instance is generated.

### Screenshots

<details>

</details>


SokratisVidros (approved):

LetItRock (approved):
nice! 🙌

PR #5823, opened 2024-06-25
feat(api): inbox - get notifications count endpoint

LetItRock:
The new Inbox endpoint to return the notifications count: `GET /inbox/notifications/count`.
Filtering with query params:
```
`?&tags=[newsletter]&read=true&archived=false`

tags?: string[]; // filter by workflow tags
read?: boolean; // read/unread, nothing means both states [true,false]
archived?: boolean; // archived/unarchived, nothing means both states [true,false]
```

Response JSON:
```
{ 
   "data": { count: 99 }, // the notifications count
   "filter": {
      "tags": [],
      "read": true,
      "archived": false
   }
}
```

### Screenshots

Screenshot 2024-06-25 at 15 02 18
Screenshot 2024-06-25 at 15 02 28
Screenshot 2024-06-25 at 15 02 41


BiswaViraj (approved):

SokratisVidros (approved):
💯

PR #5827, opened 2024-06-26
feat(web): add segment tracking

davidsoderberg:


apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorInputsPanel.tsx
```diff
@@ -53,6 +55,9 @@ export const WorkflowStepEditorInputsPanel: FC<IWorkflowStepEditorInputsPanelPro
                         size={'sm'}
                         Icon={IconOutlineSave}
                         onClick={() => {
+                          segment.track('Step controls saved', {
+                            step: step?.type,
+                          });
```
ainouzgali: ❓ **question:** Should we call it channel or channel type? I think thats the convention in other places‏

apps/web/src/studio/components/workflows/step-editor/WorkflowTestStepButton.tsx
```diff
@@ -19,10 +21,16 @@ export const WorkflowTestStepButton = ({
...
+  const segment = useSegment();
+  const isStudio = useIsStudio();
   const { currentUser } = useAuth();
   const { mutateAsync: testSendEmailEvent, isLoading: isTestingEmail } = useMutation(testSendEmailMessage);
 
   const handleTestClick = async () => {
+    segment.track('Step test ran', {
+      step: ChannelTypeEnum.EMAIL,
+      env: isStudio ? 'local' : 'cloud',
+    });
```
ainouzgali: ✏ **suggestion:** maybe add the context, like Email step test run - [Workflows Test Page]‏

apps/web/src/studio/hooks/useIsStudio.ts
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from 'react';
+import { useLocation } from 'react-router-dom';
+import { ROUTES } from '../../constants/routes';
+
+export const useIsStudio = () => {
+  const { pathname } = useLocation();
+
+  return useMemo(() => pathname.startsWith(ROUTES.STUDIO), [pathname]);
+};
```
ainouzgali, line 9: ☑ **chore:** We already have a util for it, `isStudioRoute`‏

davidsoderberg: I know the problem for me with the util is that path can contain /studio and do not need to start with it so if a workflow is named studio it will return true and can be problematic. Should we change it instead?

apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorInputsPanel.tsx
davidsoderberg: I talked with Radek about this one as it is specified in the issue, his response was:
"To me a channel is a high-level bucket that we support. here, we're in a specific context of a workflow, and our workflows have steps - and steps have types. yes, they are connected to channels, but they are not channels per se."

ainouzgali (approved):

apps/web/src/studio/hooks/useIsStudio.ts
antonjoel82: @davidsoderberg Yes, please centralize this implementation so we don't repeat ourselves or diverge

PR #5828, opened 2024-06-26
feat(js): Com 38 inbox popover

BiswaViraj:

- Popover shell component
### Screenshots

image

<details>

</details>


packages/js/src/ui/components/Bell.tsx
```diff
@@ -1,9 +1,9 @@
-import { createSignal, onCleanup, onMount, Show } from 'solid-js';
+import { Component, createSignal, onCleanup, onMount, Show } from 'solid-js';
 import { useNovu } from '../context';
 import { useStyle } from '../helpers';
 import { BellIcon } from '../icons';
 
-export function Bell() {
+export const Bell: Component = () => {
```
desiprisg: Can we not use the `Component` type in general? It seems reduntant.

packages/js/src/ui/components/PopoverWithBell.tsx
```diff
@@ -0,0 +1,20 @@
+import { createSignal, useContext } from 'solid-js';
+import { Bell, Popover } from '.';
+
+export const PopoverWithBell = (props: any) => {
```
desiprisg: What are the props for here?

packages/js/src/ui/components/PopoverWithBell.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const [bellRef, setBellRef] = createSignal<HTMLDivElement | null>(null);
+  const [isOpen, setIsOpen] = createSignal(false);
+
+  const handlePopover = () => {
+    setIsOpen((prev) => !prev);
+  };
+
+  return (
+    <>
+      <div class="nt-relative" ref={setBellRef} onClick={handlePopover}>
```
desiprisg: This should be a `button`.

packages/js/src/ui/components/PopoverWithBell.tsx
```diff
@@ -0,0 +1,20 @@
...
+        <Bell />
+        <Popover isOpen={isOpen()} targetRef={bellRef()} placement="bottom"></Popover>
```
desiprisg: We need to make sure this is accessible as well. It needs to close via `esc` and by clicking outside the popover.  Also, focus needs to "lock" in the popover. This logic should be in the popover implementation itself.

@SokratisVidros Does that sound feasible to you or should we just use `floating-ui` and see if we can replace later on?

packages/js/src/ui/components/PopoverWithBell.tsx
desiprisg: I generally don't think we should have internal reusable components. I think that the logic here will just be in the component that we export from the package itself. Will we be exporing a `<Popover />` ? If so, would the API be :
```
<Inbox>
  <Popover>
    <PopoverTrigger>
      <Bell />
    </PopoverTrigger>
    <PopoverContent>
      <Foo/>
    </PopoverContent>
  </Popover>
</Inbox>
  ```
  or something similar?

Comment (desiprisg) [reply packages/js/src/ui/components/PopoverWithBell.tsx]: The reason to not have internal reusable components is because we will have to prop drill a lot of descriptors in a big reusable internal component. It's better to have components that map to one element in the DOM, or compound components for internal consumption.

Comment (SokratisVidros) [reply packages/js/src/ui/components/PopoverWithBell.tsx]: Yes, composition only, avoid props.

File: packages/js/src/ui/api/feeds.ts [hunk 5]
```diff
@@ -0,0 +1,26 @@
+import type { FetchCountArgs } from '../../feeds';
+import { useNovu } from '../context';
+
+export const fetchFeeds = async () => {
```
LetItRock: 🗒 **note (non-blocking):** I saw this file in the other PR, not sure why it's showing here again

packages/js/src/ui/components/Bell.tsx
LetItRock: The `Component` type specifically tells that this component doesn't accept children or any other props. 
Here is the ref: https://www.solidjs.com/guides/typescript#component-types

packages/js/src/ui/components/Popover.tsx
```diff
@@ -0,0 +1,82 @@
...
+  offset?: number;
+};
+
+export const Popover: Component<PopoverProps> = (props) => {
+  const style = useStyle();
+
+  const placementwithOffsetClass = placementwithOffset(props.placement, props.offset || 0);
+
+  return (
+    <Show when={props.isOpen && props.targetRef !== null}>
```
LetItRock: 💭 **thought (non-blocking):** this code will break the show/hide animation

packages/js/src/ui/components/Popover.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <Portal mount={props.targetRef as HTMLDivElement}>
+        <div
+          class={style(
+            `nt-w-[400px] nt-h-[600px] nt-rounded-xl nt-bg-background
+             nt-shadow-[0_5px_15px_0_rgba(122,133,153,0.25)] nt-absolute nt-z-[9999]
+             ${placementwithOffsetClass}`,
+            'popover'
+          )}
```
LetItRock: ☑ **chore:** I think that we should use one of the existing libraries for the popover (like floating-ui), because it's not easy to cover all the corner cases with positioning, when no room for the content, viewport size updates, etc.

LetItRock (changes requested):

packages/js/src/ui/components/PopoverTarget.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const PopoverTarget: ParentComponent = (props) => {
+  const { setTargetRef, onToggle } = usePopover();
+  const style = useStyle();
+
+  return (
+    <button
+      ref={setTargetRef}
+      onClick={onToggle}
+      class={style(
```
BiswaViraj: 🗒 **note (non-blocking):** Decided to have the PopoverTarget to have the bellContainer styles.‏

packages/js/src/ui/components/Popover.tsx
BiswaViraj: In most of the scenarios we wont need those functionality. Floating-ui will add ≅7kb of bundle size :(

packages/js/src/ui/Inbox.tsx
```diff
@@ -31,7 +31,21 @@ const InternalInbox = () => {
   return (
     <div class={style('novu', 'root')}>
       <div class="nt-text-2xl nt-font-bold">{t('inbox.title')}</div>
-      <BellContainer />
+      <div class="nt-absolute nt-top-1/2 nt-left-1/2 nt-transform -nt-translate-x-1/2 -nt-translate-y-1/2">
```
desiprisg: ✏️ Maybe `nt-inset-0 nt-m-auto` here instead of top,left and translate ?

packages/js/src/ui/components/BellContainer.tsx
```diff
@@ -22,20 +20,12 @@ export function BellContainer() {
...
-        'bellContainer'
-      )}
-    >
+    <span>
       <BellIcon />
       <Show when={unreadCount() > 0}>
         <span
           class="nt-absolute nt-top-2 nt-right-2 nt-block nt-w-2 nt-h-2 nt-transform nt-translate-x-1/2
-            nt--translate-y-1/2 nt-bg-primary nt-rounded-full"
+            -nt-translate-y-1/2 nt-bg-primary nt-rounded-full"
```
desiprisg: ✏️ Ditto :)

packages/js/src/ui/components/Popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+type Direction = 'top' | 'bottom' | 'left' | 'right';
+type AnchorPosition = 'start' | 'end';
+type Position = Direction | `${Direction}-${AnchorPosition}`;
+type Offset = {
+  x: number | string;
+  y: number | string;
+};
+
+export type PopoverProps = {
+  opened?: boolean;
```
desiprisg: ✏️ Isn't `open` the norm here?

packages/js/src/ui/components/Popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+  onChange?: (value: boolean) => void;
+  position?: Position;
+  offset?: Offset;
+  children?: JSX.Element;
+};
+
+type PopoverContextType = {
+  targetRef: Accessor<HTMLElement | null>;
+  contentRef: Accessor<HTMLElement | null>;
+  opened: Accessor<boolean>;
```
desiprisg: ⚠️ Why does this receive an `Accessor`  instead of the value ? Props are reactive.

packages/js/src/ui/components/PopoverContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+export const PopoverContent: ParentComponent = (props) => {
+  const style = useStyle();
+  const { setContentRef, opened, targetRef, position, offset, onClose, contentRef } = usePopover();
+
+  const handleClickOutside = (e: any) => {
+    if (contentRef()?.contains(e.target)) return;
+    onClose();
+  };
+
+  const handleEscapeKey = (e: any) => {
```
desiprisg: ❓ Why is this and outside click handling needed? https://developer.mozilla.org/en-US/docs/Web/API/Popover_API/Using#auto_state_and_light_dismiss

packages/js/src/ui/components/PopoverTarget.tsx
desiprisg: ⚠️ We shouldn't apply the styles here. Especially, the `bellContainer` class. Ideally, this should receive an `asChild` prop.

Edit: I mean that the styles need to be passed to `PopoverTarget` when we use it. Let's leave the primitive unstyled.

packages/js/src/ui/components/Popover.tsx
desiprisg, line 1: Can we make a `popover` directory since there are three different files for popover?

packages/js/src/ui/helpers/popover-utils.ts
desiprisg, line 1: Let's rename to `popoverUtils` since we decided that's the casing we'll be using. 

Also, do we need all those different positioning states? Have we handled auto-adjustment here? What if the popover can't open towards the left because there isn't enough space?

packages/js/src/ui/helpers/use-uncontrolled-state.tsx
desiprisg, line 1: Let's rename to `useUncontrolledState` . I also think the `.tsx` extension is unnecessary.

packages/js/src/ui/Inbox.tsx
SokratisVidros: Yes!

packages/js/src/ui/Inbox.tsx
```diff
@@ -31,7 +31,21 @@ const InternalInbox = () => {
...
+        <div>
```
SokratisVidros: Is this extra div required? If not, let's remove it. We need to keep HTML as clean as possible and to the absolute minimum. Every extra HTML node, makes customization and the appearance prop DX handler.

packages/js/src/ui/Inbox.tsx
```diff
@@ -31,7 +31,21 @@ const InternalInbox = () => {
...
+          <Popover>
+            <Popover.Target>
+              <BellContainer />
+            </Popover.Target>
+            <Popover.Content>
+              <div>
```
SokratisVidros: Ditto

packages/js/src/ui/components/Popover.tsx
SokratisVidros: I've seen libraries using both, let's go with `open` for now, it's shorter.

packages/js/src/ui/components/Popover.tsx
```diff
@@ -0,0 +1,88 @@
...
+type AnchorPosition = 'start' | 'end';
+type Position = Direction | `${Direction}-${AnchorPosition}`;
+type Offset = {
+  x: number | string;
+  y: number | string;
+};
+
+export type PopoverProps = {
+  opened?: boolean;
+  defaultOpened?: boolean;
```
SokratisVidros:
```suggestion
  defaultOpen?: boolean;
```

packages/js/src/ui/components/PopoverContent.tsx
SokratisVidros: Good call, let's remove if not necessary. Does it maybe have to do with Solid, knowing the state of the popover div?

SokratisVidros (changes requested):

packages/js/src/ui/Inbox.tsx
desiprisg: Additionally, we should be adding a descriptor to every element, unless it something we specifically don't want customized (like a Novu banner).

packages/js/src/ui/components/Bell/Bell.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  onCleanup(() => {
+    novu.off('notifications.unread_count_changed', updateReadCount);
+  });
+
+  if (props.children) {
+    return props.children({ unreadCount: unreadCount() });
+  }
+
+  return <DefaultBellContainer unreadCount={unreadCount()} />;
```
BiswaViraj: Can be used as

```

<Bell>
  {({ unreadCount })=> return <CustomBell/>}
</Bell>

```

packages/js/src/ui/components/Popover.tsx
BiswaViraj: It didnt worked in the context without the accessor.

packages/js/src/ui/components/Popover.tsx
BiswaViraj: We will use [Solid motion](https://github.com/solidjs-community/solid-motionone?tab=readme-ov-file#exit-animations) library for animations and it supports Show component 🎉

packages/js/src/ui/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { setContentRef, opened, targetRef, onClose, contentRef } = usePopover();
+
+  const position = useFloating(targetRef, contentRef, {
+    placement: 'bottom-end',
+    whileElementsMounted: autoUpdate,
+
+    middleware: [offset(10), flip(), shift()],
+  });
+
+  const handleClickOutside = (e: any) => {
```
desiprisg, line 19: Doesn't floating ui handle this ?

packages/js/src/ui/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,63 @@
...
+    if (contentRef()?.contains(e.target)) return;
+    onClose();
+  };
+
+  const handleEscapeKey = (e: any) => {
```
desiprisg, line 24: Ditto

packages/js/src/ui/components/Popover/PopoverTarget.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { usePopover } from '.';
+
+export const PopoverTarget: ParentComponent = (props) => {
+  const { setTargetRef, onToggle } = usePopover();
+
+  return (
+    <button
+      ref={setTargetRef}
+      onClick={onToggle}
+      class="nt-h-6 nt-w-6 nt-flex nt-justify-center nt-items-center nt-outline-none"
```
desiprisg: ⚠️  This should use `style` and have a descriptor as well. Maybe `popoverTrigger` ?

packages/js/src/ui/components/Popover/PopoverContent.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <Show when={opened() && targetRef()}>
+      <Portal mount={targetRef() as HTMLElement}>
+        <div
+          id="novu-popover-content"
+          ref={setContentRef}
+          class={style(
+            `nt-w-[400px] nt-h-[600px] nt-rounded-xl nt-bg-background nt-translate-y-0
+         nt-shadow-[0_5px_15px_0_rgba(122,133,153,0.25)] nt-z-[9999]
+       `,
+            'popover'
```
desiprisg: Let's name this `popoverContent` to make it more clear ?

packages/js/src/ui/helpers/useUncontrolledState.ts
```diff
@@ -0,0 +1,33 @@
...
+  if (value !== undefined) {
+    const accessor: Accessor<boolean> = () => value;
+
+    return [accessor, handleUncontrolledChange];
+  }
+
+  /**
+   * If value is not provided, return uncontrolled state
+   */
+  return [uncontrolledValue, handleUncontrolledChange];
```
desiprisg: Is returning accessor a pattern with solid? cc: @SokratisVidros

BiswaViraj: @desiprisg, To have reactivity, we need to pass around accessors
Here is a small snippet to rest reactivity with and without accessors
https://stackblitz.com/edit/github-9nx6bn?file=src%2FApp.tsx

packages/js/src/ui/components/Popover/PopoverContent.tsx
BiswaViraj: No, I checked and it doesnt support it, they have this feature in their framework packages (react, vue)

packages/js/src/ui/components/Bell/Bell.tsx
desiprisg: Do we also allow
```
<Bell> <CustomBell/> </Bell>
```
?
Would maybe this 
```<Bell> foo </Bell>```
error ?

File: packages/js/src/ui/components/Bell/DefaultBellContainer.tsx [hunk 23]
```diff
@@ -0,0 +1,31 @@
+import { Show } from 'solid-js';
+import { useStyle } from '../../helpers';
+import { BellIcon } from '../../icons';
+
+type DefaultBellContainerProps = {
+  unreadCount: number;
+};
+
+export const DefaultBellContainer = (props: DefaultBellContainerProps) => {
```
desiprisg: Let's just name this `BellContainer` ?

packages/js/src/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,68 @@
...
+    setIsOpen(false);
+  };
+
+  const onToggle = () => {
+    if (isOpen()) {
+      onClose();
+    } else {
+      setIsOpen(true);
+    }
+  };
```
desiprisg:
```suggestion
  const onToggle = () => {
    setIsOpen(o => !o);
  };
```

packages/js/src/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const context = useContext(PopoverContext);
+  if (!context) {
+    throw new Error('usePopover must be used within Popover component');
+  }
+
+  return context;
+}
+
+Popover.Trigger = PopoverTrigger;
+Popover.Content = PopoverContent;
```
desiprisg, line 64:
```suggestion
export const Popover = {
  Root: PopoverRoot,
  Trigger: PopoverTrigger,
  Content: PopoverContent
}
```

desiprisg: Since we use `Popover.Trigger/Content` I'd prefer a `Popover.Root` here for consistency.

packages/js/src/ui/helpers/useUncontrolledState.ts
desiprisg: Ah yes this acts like a `createSignal` itself, right.

packages/js/src/ui/helpers/useUncontrolledState.ts
```diff
@@ -0,0 +1,33 @@
+import { Accessor, createSignal } from 'solid-js';
+
+type UseUncontrolledState = {
+  value?: boolean;
+  fallbackValue?: boolean;
+};
+
+type UseUncontrolledStateOutput = [Accessor<boolean>, (value: boolean) => void];
```
desiprisg: Do we need the output type? Can we just let it auto infer?

packages/js/src/ui/components/Bell/Bell.tsx
BiswaViraj: We don't allow 
```
<Bell> <CustomBell/> </Bell>
```
We decided to have 
```
<Bell>
  {({ unreadCount })=> return <CustomBell unreadCount={unreadcount} />}
</Bell>
```

packages/js/src/ui/helpers/useUncontrolledState.ts
```diff
@@ -0,0 +1,33 @@
...
+  const [uncontrolledValue, setUncontrolledValue] = createSignal(!!fallbackValue);
+
+  const handleUncontrolledChange = (val: boolean) => {
+    setUncontrolledValue(val);
+  };
+  /**
+   * If value is provided, return controlled state
+   */
+  if (value !== undefined) {
+    const accessor: Accessor<boolean> = () => value;
```
desiprisg, line 17: I think you can just do a memo here to get the type. Otherwise, no reason to type this as `Accessor`.

packages/js/src/ui/components/Popover/Popover.tsx
BiswaViraj: in general the norm is to have the Parent/root name be of the particular component and the child components to have the `.Child`

for example:
```
<Accordion>
  <Accordion.Panel/>
</Accordion>

<Menu>
  <Menu.item />
</Menu>
```

packages/js/src/ui/helpers/useUncontrolledState.ts
BiswaViraj: it infers wrong return types due to being a conditional tuple return types.
So, we need the Output type

packages/js/src/ui/helpers/useUncontrolledState.ts
BiswaViraj: CreateMemo also returns an Accessor.
Accessor Type is just a simple Generic
`type Accessor<T> = () => T`
same as writing `() => boolean`
I can update the type to 
`() => boolean` if we decide to use that

packages/js/src/ui/components/Popover/Popover.tsx
desiprisg: Radix generally does `.Root`. But since this is internal it doesn't really matter.

packages/js/src/ui/helpers/useUncontrolledState.ts
desiprisg: I was just saying, there's no need to type it out here. you can just do
`const accessor = () => value;` or `createMemo()`

desiprisg (approved):

SokratisVidros (approved):

PR #5829, opened 2024-06-26
feat(web): add checks if org have v2 enabled or not

davidsoderberg:


scopsy (approved):

scopsy:
@davidsoderberg after merge let's please test this on dev for the 2 user groups. Just to make sure we are not breaking something for existing vcustomers

apps/web/src/pages/templates/components/CreateWorkflowDropdown.tsx
```diff
@@ -38,8 +45,29 @@ export const CreateWorkflowDropdown = ({
...
   const { item: templateId, onMouseEnter, onMouseLeave } = useHoverOverItem<string>();
+  const { toggle, setPath, Component } = useDocsModal();
+
+  useEffect(() => {
+    if (!isV2Enabled) {
+      return;
+    }
+    setPath(PATHS[ROUTES.STUDIO_FLOWS] as string);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isV2Enabled]);
```
antonjoel82, line 59: ❓ **question:** Is there a reason we omitted `setPath`? We should avoid omitting dependencies unless there's a specific concern‏

libs/shared/src/types/feature-flags/feature-flags.ts
```diff
@@ -14,4 +14,5 @@ export enum FeatureFlagsKeysEnum {
   IS_V2_EXPERIENCE_ENABLED = 'IS_V2_EXPERIENCE_ENABLED',
   IS_MIXPANEL_RECORDING_ENABLED = 'IS_MIXPANEL_RECORDING_ENABLED',
   IS_INTEGRATION_INVALIDATION_DISABLED = 'IS_INTEGRATION_INVALIDATION_DISABLED',
+  IS_V2_ENABLED = 'IS_V2_ENABLED',
```
antonjoel82, line 17: ❓ **question:** Should we remove `IS_V2_EXPERIENCE_ENABLED` then?‏

PR #5834, opened 2024-06-26
feat(web): fix so scroll in onboarding is correct

davidsoderberg:


ainouzgali (approved):

apps/web/src/pages/studio-onboarding/components/Wrapper.tsx
```diff
@@ -1,14 +1,19 @@
 import { css, cx } from '@novu/novui/css';
-import { ReactNode } from 'react';
+import { ReactNode, useEffect } from 'react';
 
 export const Wrapper = ({ children, className }: { children: ReactNode; className?: string }) => {
+  useEffect(() => {
+    document.body.setAttribute('style', 'overflow: auto');
+  }, []);
+
```
antonjoel82, line 8: ❓ **question:** Why did we need this?‏

apps/web/src/pages/studio-onboarding/components/Wrapper.tsx
```diff
@@ -1,14 +1,19 @@
...
   return (
     <div
       className={cx(
         css({
           width: '100dvw',
-          height: '100dvh',
+          minHeight: '100dvh',
           colorPalette: 'mode.cloud',
+          paddingBottom: '4rem',
```
antonjoel82, line 16: ☑ **chore:** If we're going to use a non-DS value, please make into a semantic token so we only change it in one place‏

PR #5837, opened 2024-06-26
feat(framework): update novu framework headers

denis-kralj-novu:
update novu framework headers

### Related enterprise PR
[**Related PR**](https://github.com/novuhq/packages-enterprise/pull/154)


packages/framework/src/version.ts
```diff
@@ -1,3 +1,4 @@
 import packageJson from '../package.json';
 
 export const VERSION = packageJson.version;
```
rifont: Suggest to rename this to be more explicit.

```suggestion
export const SDK_VERSION = packageJson.version;
```

rifont (approved):

packages/framework/src/constants/http-headers.constants.ts
```diff
@@ -1,15 +1,15 @@
 export enum HttpHeaderKeysEnum {
-  SIGNATURE = 'x-novu-signature',
-  ANONYMOUS = 'x-novu-anonymous',
-  EXECUTION_DURATION = 'x-novu-execution-duration',
-  SDK_VERSION = 'x-novu-sdk',
-  FRAMEWORK = 'x-novu-framework',
+  SIGNATURE = 'novu-signature',
+  ANONYMOUS = 'novu-anonymous',
+  SIGNATURE_DEPRECATED = 'x-novu-signature',
```
SokratisVidros: Made my day! 💯

SokratisVidros (approved):
Very well done! 👏 👏 👏

packages/framework/src/constants/http-headers.constants.ts
```diff
@@ -1,15 +1,15 @@
...
+  ANONYMOUS_DEPRECATED = 'x-novu-anonymous',
+  NOVU_FRAMEWORK_SDK = 'novu-framework-sdk',
+  NOVU_FRAMEWORK_API = 'novu-framework-api',
+  FRAMEWORK_VERSION = 'novu-framework-version',
```
rifont: 🤓 **nitpick (non-blocking):** Suggest to rename all the enum keys (CONSTANT_CASE) to align with the enum values (kebab-case) for maximum explicitness and readability.

This is the pattern we are following and enforcing with some [clever typescript typings](https://github.com/novuhq/novu/blob/next/apps/api/src/app/shared/framework/types/headers.types.ts#L23) in the API application.‏

denis-kralj-novu: I like this pattern and will apply it. I will also create a task to follow up and move the helper on a shared location so that it can be reused in more locations

denis-kralj-novu: https://linear.app/novu/issue/NV-3955/extract-enum-key-value-match-testing-function-to-shared-location-and

PR #5838, opened 2024-06-26
feat(novui): Add LoadingOverlay component

antonjoel82:

- Add `LoadingOverlay` component
  - Works best for full-page, but can be used in a component by adding `position: relative` to the parent
- Fix Panda build issue and re-instate input padding 

### Screenshots

https://github.com/novuhq/novu/assets/47441786/82212648-392a-4528-bc1a-b0e24310b166


libs/novui/src/recipes/loading-overlay.recipe.ts
```diff
@@ -0,0 +1,33 @@
+import { LoadingOverlayStylesNames } from '@mantine/core';
+import { defineSlotRecipe } from '@pandacss/dev';
+import { colorPaletteGradientHorizontal } from 'src/ingredients';
```
ainouzgali: this will not compile I believe :) should be `../ingredients`

antonjoel82: Thank you! I hate that my editor always does this, and haven't found a solution that consistently addresses this

SokratisVidros (approved):

PR #5839, opened 2024-06-26
feat(framework): add trigger action

djabarovgeorge:
### Screenshots
image

image

<details>

</details>


packages/framework/src/handler.ts
```diff
@@ -57,21 +57,29 @@ interface IActionResponse<TBody extends string = string> {
...
     this.handler = options.handler;
     this.client = options.client ? options.client : new Client();
     this.client.addWorkflows(options.workflows);
+    this.http = this.initHttpClient();
     this.frameworkName = options.frameworkName;
     this.hmacEnabled = this.client.strictAuthentication;
   }
 
+  private initHttpClient() {
+    return axios.create({
```
SokratisVidros: There is no need for an extra dependency on the framework package that increases it's bundle and runtime size. Let's use fetch.

packages/framework/src/handler.ts
```diff
@@ -212,6 +221,29 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
     };
   }
 
+  private triggerAction(triggerEvent: ITriggerEvent) {
+    return async () => {
+      if (!triggerEvent.bridgeUrl && process.env.NEXT_PUBLIC_VERCEL_URL) {
```
SokratisVidros: This should be provided in framework-specific handlers or `trigger` function re-exports. This is reusable logic across all frameworks so I'd remove it and respect the arguments of the `triggerEvent`.

SokratisVidros (changes requested):

packages/framework/package.json
```diff
@@ -104,6 +104,7 @@
     "zod-to-json-schema": "^3.23.0"
   },
   "dependencies": {
+    "@novu/shared": "workspace:*",
```
rifont: ☑ **chore:** move this one to `devDependencies`‏

packages/framework/src/client.test.ts
rifont, line 1: ⚠ **issue:** Let's also add a small unit test to verify that `fetch` is called with the expected data.‏

packages/framework/src/handler.ts
```diff
@@ -262,6 +297,9 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
       }
 
       return this.createError(error);
+    }
+    if ((error as any).response.status === HttpStatusEnum.BAD_REQUEST) {
```
rifont: ✏ **suggestion:** Could we chain this into an `else if` for consistency with the existing implementation?


```suggestion
    } else if ((error as any).response.status === HttpStatusEnum.BAD_REQUEST) {
```

djabarovgeorge: Definitely, it should have been else if my miss, thanks for rasing it up!

packages/framework/package.json
djabarovgeorge: We will keep it like this for now as decided with Richard, and revisit it later.

SokratisVidros (approved):

PR #5843, opened 2024-06-26
feat(novu): add cli loading states

scopsy:
### Screenshots

CleanShot 2024-06-27 at 00 49 34

This is visible after 10 seconds of failed attempts
CleanShot 2024-06-27 at 00 49 37

Successful connection
CleanShot 2024-06-27 at 00 49 56
<details>

</details>


packages/cli/src/commands/dev.ts
```diff
@@ -30,7 +34,59 @@ export async function devCommand(options: DevCommandOptions) {
...
+  }
+
+  let healthy = false;
+  const endpointText = `Looking for the Novu Endpoint at ${parsedOptions.origin}${NOVU_ENDPOINT_PATH}. Ensure your application is configured and running locally.`;
+  const endpointSpinner = ora(endpointText).start();
+
+  let counter = 0;
+  while (!healthy) {
+    try {
+      const response = await fetch(`${parsedOptions.origin}${NOVU_ENDPOINT_PATH}?action=health-check`, {
```
SokratisVidros: Nit pick: we should prefer fetch over axios, so we can remove axios in a follow up PR.

packages/cli/src/commands/dev.ts
```diff
@@ -30,7 +34,59 @@ export async function devCommand(options: DevCommandOptions) {
...
+        headers: {
+          accept: 'application/json',
+          'Content-Type': 'application/json',
+        },
+      });
+      const healthResponse = await response.json();
+      healthy = healthResponse.status === 'ok';
+
+      if (healthy) {
+        endpointSpinner.succeed(`Endpoint properly configured: ${parsedOptions.origin}${NOVU_ENDPOINT_PATH}`);
```
SokratisVidros:
```suggestion
        endpointSpinner.succeed(`Novu Endpoint detected: ${parsedOptions.origin}${NOVU_ENDPOINT_PATH}`);
```

SokratisVidros (approved):

packages/cli/src/index.ts
```diff
@@ -18,6 +18,7 @@ program
   .command('dev')
   .description('Start a Novu Dev Studio and a localtunnel')
   .option('-p, --port <value>', 'Set the local port for the Novu endpoint, defaults to 4000', '4000')
+  .option('--path <value>', 'Set the Novu Endpoint mounted path, defaults to 4000', '/api/novu')
```
SokratisVidros: Can users override that now in @novu/framework?

packages/cli/src/commands/dev.ts
scopsy: Im pretty sure i used fetch here 😀

SokratisVidros: Correct, but axios is already in CLI. This comment was for a follow up PR.

packages/cli/src/index.ts
scopsy: This is not really a novu/framework concern, as the user can mount the endpoint anywhere he really wants. We just need to make that the Local Studio knows which path to access

CleanShot 2024-06-27 at 09 10 26@2x

packages/cli/src/index.ts
```diff
@@ -18,6 +18,7 @@ program
   .command('dev')
   .description('Start a Novu Dev Studio and a localtunnel')
   .option('-p, --port <value>', 'Set the local port for the Novu endpoint, defaults to 4000', '4000')
+  .option('-er, --endpoint-route <value>', 'Set the Novu Endpoint mounted route, defaults to /api/novu', '/api/novu')
```
scopsy: @SokratisVidros @rifont what do you think on the naming here?

PR #5845, opened 2024-06-27
feat(web): hide more for v2 enabled feature flag

davidsoderberg:
Screenshot 2024-06-27 at 11 27 08

### DIFF
--- a/apps/web/src/pages/templates/TemplatesListNoData.tsx
+++ b/apps/web/src/pages/templates/TemplatesListNoData.tsx
@@ -8,8 +8,9 @@ import { CardTile, colors, Popover } from '@novu/design-system';
 import { useSegment } from '../../components/providers/SegmentProvider';
 import { IBlueprintTemplate } from '../../api/types';
 import { TemplateCreationSourceEnum } from './shared';
-import { useHoverOverItem } from '../../hooks';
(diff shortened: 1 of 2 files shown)


apps/web/src/pages/templates/TemplatesListNoData.tsx
```diff
@@ -70,6 +71,11 @@ export const TemplatesListNoData = ({
 }) => {
   const segment = useSegment();
   const { item: templateId, onMouseEnter, onMouseLeave } = useHoverOverItem<string>();
+  const isV2Enabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V2_ENABLED);
```
scopsy, line 75: @davidsoderberg maybe we should only show the         <EchoProjectCardTile /> here?

davidsoderberg: Not sure I follow

scopsy: In the empty state components we have 3 cards (Blank, Echo, and some from the template store)

I'm suggesting to still render the empty state component, but only have the Echo card there.

scopsy (approved):

PR #5851, opened 2024-06-27
feat(js): Introduce baseTheme prop and theme merging

desiprisg:

Allow passsing a `baseTheme` on `appearance`. The `baseTheme` can be a single `Theme` (meaning `elements` and `variables`) or an array. All the base themes and `elements`, `variables` are merged in the order they are provided (in the case of `baseTheme` array and the `elements`, `variables` are added last, meaning they override.

Also created a dark theme that can be used like so:
```
<Inbox 
  appearance={{
    baseTheme: dark, //or [dark, ...moreThemes]
    variables: {...} //overrides base theme variables
    elements: {...} //overrides base theme elements
  }}
/>
```
### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/ui/context/AppearanceContext.tsx
+++ b/packages/js/src/ui/context/AppearanceContext.tsx
@@ -1,4 +1,13 @@
-import { ParentProps, createContext, createEffect, createSignal, onCleanup, onMount, useContext } from 'solid-js';
+import {
+  ParentProps,
+  createContext,
+  createEffect,
+  createSignal,
+  onCleanup,
+  onMount,
+  useContext,
+  createMemo,
+} from 'solid-js';
 import { createStore } from 'solid-js/store';
 import { defaultVariables } from '../config';
 import { parseElements, parseVariables } from '../helpers';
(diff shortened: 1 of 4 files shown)


packages/js/src/ui/context/AppearanceContext.tsx
```diff
@@ -46,6 +56,9 @@ export const AppearanceProvider = (props: AppearanceProviderProps) => {
   const [styleElement, setStyleElement] = createSignal<HTMLStyleElement | null>(null);
   const [elementRules, setElementRules] = createSignal<string[]>([]);
   const [variableRules, setVariableRules] = createSignal<string[]>([]);
+  const themes = createMemo(() =>
+    Array.isArray(props.appearance?.baseTheme) ? props.appearance?.baseTheme || [] : [props.appearance?.baseTheme || {}]
+  );
```
LetItRock, line 63: ✏ **suggestion:** no need for memoization

```
const themes =
    props.appearance?.baseTheme && Array.isArray(props.appearance.baseTheme)
      ? props.appearance.baseTheme
      : [props.appearance.baseTheme];
```

Btw I'm not sure why it's not picking the appearance prop type.

desiprisg: This still needs to be a function though. Otherwise, it is not reactive.
From docs: `In Solid, you often don't need to wrap functions in memos; you can alternatively just define and call a regular function to get similar reactive behavior.`

SokratisVidros (approved):

packages/js/src/ui/context/AppearanceContext.tsx
desiprisg: Since we're calling `themes()` twice here, the memo is preventing it from executing two times.

PR #5854, opened 2024-06-27
refactor(web): studio onboarding to use same bridge hook as rest of studio

davidsoderberg:


SokratisVidros (approved):

apps/web/src/pages/studio-onboarding/index.tsx
```diff
@@ -48,7 +43,7 @@ export const StudioOnboarding = () => {
             To start sending your first workflows, you first need to connect Novu to your Bridge Endpoint. This setup
             will create a sample Next.js project and pre-configured the @novu/framework client for you.
           </Text>
-          <SetupTimeline testResponse={{ data, isLoading }} />
+          <SetupTimeline testResponse={{ data: data as BridgeStatus, isLoading }} />
```
SokratisVidros, line 46: I am a bit surprised we need this assertion here.

davidsoderberg: I was to but typescript refused to let it pass without this :(

PR #5857, opened 2024-06-27
chore(framework,node): Rename NOVU_API_KEY to NOVU_SECRET_KEY

SokratisVidros:
Secret key makes it clear that it must be kept a secret.

### DIFF
--- a/packages/create-novu-app/index.ts
+++ b/packages/create-novu-app/index.ts
@@ -42,11 +42,11 @@ const program = new Commander.Command(packageJson.name)
     projectPath = name;
   })
   .option(
-    '-k, --api-key [apiKey]',
+    '-k, --secret-key [secretKey]',
     `
 
-  Your Novu Development environment apiKey. Note that your novu app won't
-  work outside of local mode without an apiKey.
+  Your Novu development environment secret key. Note that your novu app won't
+  work outside of local mode without it.
 `
   )
   .option(
--- a/packages/framework/src/client.ts
+++ b/packages/framework/src/client.ts
@@ -47,12 +47,16 @@ JSONSchemaFaker.option({
   alwaysFakeOptionals: true,
 });
 
+function isRuntimeInDevelopment() {
+  return process.env.NODE_ENV === 'development';
+}
+
 export class Client {
   private discoveredWorkflows: Array<DiscoverWorkflowOutput> = [];
 
   private templateEngine = new Liquid();
 
-  public apiKey?: string;
+  public secretKey?: string;
 
   public version: string = SDK_VERSION;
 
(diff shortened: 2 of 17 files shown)


packages/create-novu-app/index.ts
```diff
@@ -42,11 +42,11 @@ const program = new Commander.Command(packageJson.name)
     projectPath = name;
   })
   .option(
-    '-k, --api-key [apiKey]',
+    '-k, --secret-key [secretKey]',
     `
 
-  Your Novu Development environment apiKey. Note that your novu app won't
-  work outside of local mode without an apiKey.
+  Your Novu development environment secret key. Note that your novu app won't
```
denis-kralj-novu, line 48: 🗒 **note (non-blocking):** ‏ this can also be the prod secret key, maybe remove the name of the env

packages/framework/src/client.ts
```diff
@@ -70,10 +70,12 @@ export class Client {
...
 
-    if (providedOptions?.apiKey !== undefined) {
-      builtConfiguration.apiKey = providedOptions.apiKey;
-    } else if (process.env.NOVU_API_KEY !== undefined) {
-      builtConfiguration.apiKey = process.env.NOVU_API_KEY;
+    builtConfiguration.apiKey = providedOptions?.apiKey || process.env.NOVU_SECRET_KEY || process.env.NOVU_API_KEY;
+
+    if (!builtConfiguration.apiKey) {
+      throw new Error(
+        'Missing secret key. Set the NOVU_SECRET_KEY environment variable or pass an apiKey to the client options.'
```
denis-kralj-novu: ✏ **suggestion:** ‏ 
```suggestion
        'Missing secret key. Set the NOVU_SECRET_KEY environment variable or pass a secret key to the client options.'
```

packages/node/src/lib/novu.ts
```diff
@@ -61,7 +61,14 @@ export class Novu extends EventEmitter {
...
       }
     } else {
-      apiKey = getEnvVariable('NOVU_API_KEY');
+      apiKey =
+        getEnvVariable('NOVU_SECRET_KEY') || getEnvVariable('NOVU_API_KEY');
+    }
+
+    if (!apiKey) {
+      throw new Error(
+        'Missing secret key. Set the NOVU_SECRET_KEY environment variable or pass an apiKey to new Novu(apiKey) constructor.'
```
denis-kralj-novu: ✏ **suggestion:** ‏ WE could also rename the `apiKey` parameter, otherwise it is confusing to use secret key and apiKey in one sentence

denis-kralj-novu (approved):

packages/create-novu-app/index.ts
SokratisVidros: We shouldn't promote the use of the prod key at this point. Production keys should be used only in env vars (unless the setup is more complicated) in the production infrastructure. 

In a future iteration, we will add prefixed identifiers to all keys, in order to understand easily if they are dev or prod keys.

packages/node/src/lib/novu.ts
SokratisVidros: Done!

PR #5864, opened 2024-06-28
refactor(framework): move types to shared

djabarovgeorge:
We need to externalize these types and reuse them on the API, the framework, the Web and Worker.

### Screenshots

<details>

### Related enterprise PR
[EE-PR](https://github.com/novuhq/packages-enterprise/pull/156)

</details>


packages/framework/package.json
```diff
@@ -101,6 +101,7 @@
     "typescript": "^5.3.3",
     "vitest": "^1.2.1",
     "zod": "^3.23.0",
+    "json-schema-to-ts": "^3.0.0",
```
djabarovgeorge, line 104: Is there any reason to keep it in the `dependencies` section?

packages/framework/package.json
djabarovgeorge, line 1: Do we need the actual version under `"@novu/shared": "workspace:*"` instead of the `workspace:*`?

libs/shared/package.json
```diff
@@ -39,13 +39,17 @@
   },
   "dependencies": {
     "class-transformer": "0.5.1",
-    "class-validator": "0.14.0"
+    "class-validator": "0.14.0",
+    "ajv": "^8.12.0"
```
djabarovgeorge, line 43: ajv is safe for both node and browser, however, i wonder if it should be part of the `shared` lib

libs/shared/src/consts/framework/resource.constants.ts
djabarovgeorge, line 1: ResourceEnum was duplicated and this `@novu/shared.ResourceEnum` was merged with the resource of `@novu/framework.ResourceEnum`

packages/framework/src/constants/step.constants.ts
```diff
@@ -1,13 +0,0 @@
-export enum ChannelStepEnum {
-  EMAIL = 'email',
-  SMS = 'sms',
-  PUSH = 'push',
-  CHAT = 'chat',
-  IN_APP = 'in_app',
-}
```
djabarovgeorge, line 7: is a duplication with `@novu/shared.ChannelTypeEnum` therefore was deleted.

packages/framework/src/constants/step.constants.ts
```diff
@@ -1,13 +0,0 @@
...
-
-export enum ActionStepEnum {
-  DIGEST = 'digest',
-  DELAY = 'delay',
-  CUSTOM = 'custom',
-}
```
djabarovgeorge, line 13: is a duplication with @novu/shared.ActionTypeEnum therefore was deleted.

libs/shared/src/types/framework/discover.types.ts
```diff
@@ -1,10 +1,10 @@
-import { ActionStepEnum, ChannelStepEnum } from '../constants';
 import { JsonSchema, Schema } from './schema.types';
 import { StepOptions } from './step.types';
 import { Execute, WorkflowOptions } from './workflow.types';
 import { Awaitable } from './util.types';
+import { ActionTypeEnum, ChannelTypeEnum } from '../channel';
 
-export type StepType = `${ChannelStepEnum | ActionStepEnum}`;
+export type StepTypeString = `${ChannelTypeEnum | ActionTypeEnum}`;
```
djabarovgeorge, line 7: we had duplication of type `StepType` which were:
1. StepType = ChannelTypeEnum | ActionTypeEnum;
2. StepType = `${ChannelTypeEnum | ActionTypeEnum}`;

As a temporary solution in order to not break the interface i kept both and renamed the template literal (the second StepType) to StepTypeString, i would like to hear your opinion on this matter.

libs/shared/src/types/framework/step.types.ts
```diff
@@ -22,18 +23,6 @@ export type StepOptions = {
   providers?: Record<string, (payload: any) => Awaitable<any>>;
 };
 
-export enum JobStatusEnum {
```
djabarovgeorge, line 25: this one was deleted because it was duplicated.

packages/framework/package.json
rifont: Checked in code should use `workspace:*`, and we modify to the actual version during publish only. After publish, we revert back to `workspace:*`. `pnpm version` handles all this trickiness for us.

packages/framework/package.json
rifont: We need to include it in `dependencies` as JSON-schema is the "default", barebones validation tool we're using in the framework. This ensures that Dev's can use the Framework without installing anything else.

SokratisVidros (changes requested):
As discussed offline, the source of truth for types should be the @novu/framework.

djabarovgeorge:
Closing this one as we won't relocate the types from framework to shared package at the momnet.

PR #5868, opened 2024-06-28
fix(web): Strengthen security posture against click jacking

SokratisVidros:
Be more specific in Netlify and allow only /studio paths to be embedded in an Iframe coming from localhost


rifont (approved):
Awesome! 🦾

PR #5871, opened 2024-06-28
fix(novu): Allow popups so we can open discord link

antonjoel82:

We currently get:

```
Blocked opening 'https://discord.gg/novu' in a new window because the request was made in a sandboxed frame whose 'allow-popups' permission is not set.
```

When clicking on the `?` button that opens the discord invite URL from local. This should address that!

### DIFF
--- a/packages/cli/src/dev-server/http-server.ts
+++ b/packages/cli/src/dev-server/http-server.ts
@@ -78,7 +78,7 @@ export class DevServer {
 
           function injectIframe(src) {
             const iframe = window.document.createElement('iframe');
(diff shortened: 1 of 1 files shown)


antonjoel82:
Screenshot 2024-06-28 at 8 07 52 AM
While this allows opening a new tab for the link, it still fails on the Discord side. Created this ticket to follow-up: https://linear.app/novu/issue/NV-3979/determine-what-to-do-on-help-icon-click-in-local

antonjoel82:
Closed because it's handled here #5873

PR #5876, opened 2024-06-29
refactor(root): Rename from `web.novu.co` to `dashboard.novu.co`

rifont:

* Add Netlify 301 redirect for transparent redirect on web server for all environments
* Replace all occurrences of `web.novu.co` with `dashboard.novu.co`

**Deployment sequence (per `<environment>`):**
1. Merge [this Cloud-Infra PR](https://github.com/novuhq/cloud-infra/pull/132) first, which modifies CORS config.
2. Apply the changes from Cloud Infra to the relevant `<environment>`
3. Deploy modified ECS task definitions to API & Worker
4. After the ECS rolling restart concludes, merge this PR, then immediately trigger a new build from Netlify for the relevant `<environment>`. There is likely to be a small time window where access to `web.novu.co` results in CORS errors without due to the redirect not taking place, corresponding to the delta between when CORS changes from `#2` above are deployed and the Netlify redirects take effect. This delta should be less than 10 minutes, which has the effect of downtime for new logins to the Web interface.

### Screenshots

<details>

</details>


apps/web/netlify.toml
```diff
@@ -8,10 +8,29 @@
...
+[[redirects]]
+  from = "https://web.novu.co/*"
+  to = "https://dashboard.novu.co/:splat"
+  status = 301
+  force = true
+
 [[redirects]]
   from = "/*"
   to = "/index.html"
   status = 200
```
rifont, line 32: Important to keep this redirect last, as Netlify stops evaluating redirects after the first `from` match - keeping this last ensures that the subdomain redirects occur first, stop matching after redirect, then start to match the greedy `/*` redirect lastly.

apps/web/netlify.toml
```diff
@@ -8,10 +8,29 @@
 [context.deploy-preview]
   command = "cd apps/web && pnpm run envsetup && cd ../../ && pnpm run build:web --skip-nx-cache"
 
+[[redirects]]
+  from = "https://dev.web.novu.co/*"
+  to = "https://dev.dashboard.novu.co/:splat"
```
rifont, line 13: See Netlify [Splat docs](https://docs.netlify.com/routing/redirects/redirect-options/#splats) for more info.

The asterisk `*` converts to `:splat` in their syntax, such that the path is maintained during redirects.

SokratisVidros (approved):

antonjoel82 (approved):
Looks good!


💭 **thought (non-blocking):** While we have this changeset, it may be worth substituting all these raw strings with an env variable (or other constant) to avoid having to find all instances in the future.‏
‏

PR #5877, opened 2024-06-29
feat(framework): Add trigger capability to defined workflows

rifont:
* Add `workflow.trigger(...)` capability for triggering defined workflows
* Add `(await workflow.trigger(...)).cancel()` capability for cancelling trigger of defined workflows
* Add `getBridgeUrl` util to resolve correct bridge URL in framework agnostic way

**NB:** See this stacked follow up PR (https://github.com/novuhq/novu/pull/5880) which replaces use of the internal HTTP client with `@novu/api` bindings

### Screenshots
_Tested working in the AI Digest Repo_
https://github.com/rifont/lightonia/pull/1
image

_`workflow.trigger()` intellisense with strongly typed `payload`_
image

_`(await workflow.trigger()).cancel()` available after using the trigger (no need to pass `transactionId`)_
image

<details>

</details>


packages/framework/src/client.test.ts
```diff
@@ -47,7 +47,7 @@ describe('Novu Client', () => {
 
     it('should throw an error when secretKey is not provided', () => {
       expect(() => new Client({ secretKey: undefined })).toThrow(
-        'Missing secret key. Set the NOVU_SECRET_KEY environment variable or pass `secretKey` to the client options.'
+        'Missing secret key. Set the `NOVU_SECRET_KEY` environment variable or pass `secretKey` to the client options.'
```
rifont, line 50: Small DX fix for better readability.

packages/framework/src/errors/execution.errors.ts
```diff
@@ -74,3 +74,12 @@ export class ExecutionProviderOutputInvalidError extends BadRequestError {
...
 }
+
+export class WorkflowPayloadInvalidError extends BadRequestError {
+  code = ErrorCodeEnum.WORKFLOW_PAYLOAD_INVALID_ERROR;
+
+  constructor(workflowId: string, data: any) {
+    super(`Workflow with id: \`${workflowId}\` has invalid \`payload\`. Please provide the correct payload.`);
+    this.data = data;
+  }
+}
```
rifont, line 85: New runtime error thrown when providing invalid payload during the `workflow.trigger(...)` validation

packages/framework/src/handler.ts
```diff
@@ -133,11 +133,6 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
       (await actions.headers(HttpHeaderKeysEnum.NOVU_SIGNATURE)) ||
       (await actions.headers(HttpHeaderKeysEnum.NOVU_SIGNATURE_DEPRECATED)) ||
       '';
-    const anonymousHeader =
-      (await actions.headers(HttpHeaderKeysEnum.NOVU_ANONYMOUS)) ||
-      (await actions.headers(HttpHeaderKeysEnum.NOVU_ANONYMOUS_DEPRECATED)) ||
-      '';
-    const source = url.searchParams.get(HttpQueryKeysEnum.SOURCE) || '';
```
rifont, line 140: No longer used.

packages/framework/src/types/discover.types.ts
```diff
@@ -67,6 +68,13 @@ export type DiscoverWorkflowOutput = {
   };
 };
 
+export type Workflow<T_Payload> = {
+  trigger: (
+    event: Prettify<Omit<EventTriggerParams<T_Payload>, 'workflowId' | 'bridgeUrl'>>
+  ) => Promise<EventTriggerResult>;
+  definition: DiscoverWorkflowOutput;
+};
```
rifont, line 76: A wrapper around the defined workflow typings, enabling us to add usable methods like `trigger` to the `const myWorkflow.<stuff>` in a less cluttered namespace than `DiscoverWorkflowOutput`, which isn't stuff that the Developer will typically use.

I've chosen the `definition` naming on the thing we use for discovery as it conceptually makes sense as the noun for the workflow you defined in code. Additionally, `definition` will work well for other domains we add in the future. I would like to hear thoughts on this matter to verify it makes sense.

packages/framework/src/types/execution.types.ts
```diff
@@ -15,17 +14,6 @@ export type Event = {
...
-  workflowId: string;
-  to: TriggerRecipientsPayload;
-  actor?: TriggerRecipientSubscriber;
-  bridgeUrl?: string;
-  payload: ITriggerPayload;
-  tenant?: ITenantDefine;
-  transactionId?: string;
-  overrides?: Record<string, unknown>;
-};
-
```
rifont, line 29: These don't belong in execution types, they are moved to `event.types`.

packages/framework/src/types/index.ts
```diff
@@ -1,14 +1,15 @@
...
+export * from './skip.types';
 export * from './step.types';
 export * from './subscriber.types';
 export * from './util.types';
-export * from './workflow.types';
-export * from './schema.types';
-export * from './cron.types';
-export * from './skip.types';
 export * from './validator.types';
+export * from './workflow.types';
```
rifont, line 15: Some reordering alphabetically.

packages/framework/src/types/util.types.ts
```diff
@@ -8,3 +8,10 @@ export type Either<A, B> = Partial<A> & Partial<B> & (A | B);
  * A type that represents a value that may be a promise or a regular value.
  */
 export type Awaitable<T> = T | Promise<T>;
+
+/**
+ * A type that represents a type that is a prettified version of the original type.
+ * The prettified type has any generics removed from intellisense and displays a flat object.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type Prettify<T> = { [K in keyof T]: T[K] } & {};
```
rifont, line 17: How about that, @SokratisVidros ? 😁

packages/framework/src/workflow.test.ts
```diff
@@ -1,10 +1,11 @@
-import { it, describe } from 'vitest';
+import { it, describe, beforeEach, expect, vi, afterEach } from 'vitest';
+import { MissingSecretKeyError, WorkflowPayloadInvalidError } from './errors';
 import { workflow } from './workflow';
 
 describe('workflow function', () => {
   describe('Type tests', () => {
     it('should not compile when the channel output is incorrect', async () => {
-      await workflow('setup-workflow', async ({ step }) => {
+      workflow('setup-workflow', async ({ step }) => {
```
rifont, line 8: Redundant `await`.

packages/framework/src/workflow.test.ts
```diff
@@ -62,4 +63,182 @@ describe('workflow function', () => {
...
+        {
+          payloadSchema: testPayloadSchema,
+        }
+      );
+
+      // Capture in a test function to avoid throwing execution errors
+      const testFn = () =>
+        testWorkflow.trigger({
+          // @ts-expect-error - foo is missing from the payload
+          payload: {},
```
rifont, line 102: Verifying that the payload is strongly typed here 🦾

packages/framework/src/workflow.ts
```diff
@@ -27,9 +31,51 @@ export function workflow<
...
+      ...(event.actor && { actor: event.actor }),
+      ...(event.tenant && { tenant: event.tenant }),
+      ...(bridgeUrl && { bridgeUrl }),
+    };
+
+    const result = await apiClient.post<EventTriggerResponse>('/events/trigger', requestPayload);
+
+    const cancel = async () => {
+      return apiClient.delete<CancelEventTriggerResponse>(`/events/trigger/${result.transactionId}`);
+    };
```
rifont, line 71: Adding [cancel](https://docs.novu.co/api-reference/events/cancel-triggered-event) for a complete E2E story on the `trigger`.

packages/framework/src/workflow.ts
```diff
@@ -27,9 +31,51 @@ export function workflow<
   workflowId: string,
   execute: Execute<T_Payload, T_Control>,
   workflowOptions?: WorkflowOptions<T_PayloadSchema, T_ControlSchema>
-): DiscoverWorkflowOutput {
+): Workflow<T_Payload> {
   const options = workflowOptions ? workflowOptions : {};
 
+  const apiClient = initApiClient(process.env.NOVU_SECRET_KEY as string);
```
rifont, line 37: We are using the API client here in quite a dumb way.

I'm not particularly happy about this for a couple reasons:
1. Each workflow will have a client instantiated
2. It doesn't use the `secretKey` provided via the client options

What we really need is some sort of "global" API client that respects all of:
* `secretKey` provided to the client
* `NOVU_SECRET_KEY` in the environment

Would love to hear any suggestions for improvement here.

rifont: Additionally, I chose not to use the `@novu/node` package right now and use `fetch` directly instead, because using the former felt like quite a sledgehammer for what we need right now.

When we look at consolidating the `@novu/framework` and `@novu/node` packages into a single package, we should revisit this.

packages/framework/src/types/discover.types.ts
```diff
@@ -67,6 +68,13 @@ export type DiscoverWorkflowOutput = {
   };
 };
 
+export type Workflow<T_Payload = any> = {
+  trigger: (
+    event: Prettify<Omit<EventTriggerParams<T_Payload>, 'workflowId' | 'bridgeUrl'>>
+  ) => Promise<EventTriggerResult>;
```
rifont, line 74: I thought about adding [`triggerBulk`](https://docs.novu.co/api-reference/events/bulk-trigger-event) and [`broadcast`](https://docs.novu.co/api-reference/events/broadcast-event-to-all) here.

`broadcast` is straightforward to add.

`triggerBulk` needs some more thinking about error handling. If one of the bulk events has an invalid payload as determined with schema validation, there's a question of how that should be surfaced on the method results. Should it do something like a `Promise.allSettled(...)` and potentially have a mixture of successful and failed triggers, or should we throw a failure state if at least 1 of the payloads is invalid, which would stop trigger of all bulk events on the client?

rifont:
I'm thinking I'll replace all the api HTTP client with `@novu/api` usage instead. We can start dogfooding it for all resources exposed code-first.

Where should the instantiated `new (await import('@novu/api')).Client` live? And how should it be accessed?

rifont:
> I'm thinking I'll replace all the api HTTP client with `@novu/api` usage instead. We can start dogfooding it for all resources exposed code-first.
> 
> Where should the instantiated `new (await import('@novu/api')).Client` live? And how should it be accessed?

See https://github.com/novuhq/novu/pull/5880

packages/framework/src/errors/execution.errors.ts
```diff
@@ -75,6 +75,15 @@ export class ExecutionProviderOutputInvalidError extends BadRequestError {
   }
 }
 
+export class WorkflowPayloadInvalidError extends BadRequestError {
```
SokratisVidros, line 78: This is not for this PR, but we need to consolidate all errors into a single class with different codes and error messages to save some bytes from the framework SDK.

packages/framework/src/types/discover.types.ts
SokratisVidros: Good call. TriggerBulk response requires some API work for the error payload. We would also need some indication of progress on that payload.

Let's add it as a post-release improvement.

packages/framework/src/types/discover.types.ts
SokratisVidros: I really like using the term definition. 💯

packages/framework/src/types/event.types.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  workflowId: string;
+  /**
+   * Recipients to trigger the workflow to
+   */
+  to: TriggerRecipientsPayload;
+  /**
+   * Actor to trigger the workflow from
+   */
+  actor?: TriggerRecipientSubscriber;
```
SokratisVidros: I am not a big fan of the `to` and `actor` nomenclature. We had the same concerns while designing the notification DSL. We don't have a good solution yet, but I'd consider renaming TriggerRecipientSubscriber type to just `Actor` or `Subscriber` payload for now.

packages/framework/src/types/util.types.ts
SokratisVidros: 😍

packages/framework/src/workflow.ts
SokratisVidros: I agree 100%. 

1. As suggested, I'd like to use a basic fetch at this point. 
2. I like the factory pattern for this kind of this such as function `buildAPIClient({ apiKey, ...options })`. You can call that once and assign it to the current framework `Client` object as a private prop.

packages/framework/src/workflow.ts
```diff
@@ -27,9 +31,51 @@ export function workflow<
...
+
+  const trigger: Workflow<T_Payload>['trigger'] = async (event) => {
+    if (!process.env.NOVU_SECRET_KEY) {
+      throw new MissingSecretKeyError();
+    }
+
+    let validatedData: T_Payload;
+    if (options.payloadSchema) {
+      const validationResult = await validateData(options.payloadSchema, event.payload);
```
SokratisVidros, line 46: We've discussed this separately regarding DX, and I think it's a great opportunity to use the `const {data, error }` signature for this function instead of throwing an error. 

I'd like to hear your thoughts.

SokratisVidros (changes requested):

packages/framework/src/utils/env.utils.ts
```diff
@@ -7,3 +7,32 @@ export const getResponse = (): typeof Response => {
...
+
+  // Vercel preview deployments
+  if (process.env.NEXT_PUBLIC_VERCEL_URL) {
+    return `https://${process.env.NEXT_PUBLIC_VERCEL_URL}/api/novu`;
+  }
+
+  // Local environments
+  try {
+    if (process.env.NODE_ENV === 'development') {
+      const response = await fetch('http://localhost:2022/.well-known/novu');
```
SokratisVidros, line 28: This is fine for now. In the original doc, since @novu/framework can't tell where Studio is running, there was a best-effort algorithm to use a fixed set of 5 predefined ports and try all of them until one replies. We can leave it as such and get back to it after the release.

packages/framework/src/errors/execution.errors.ts
djabarovgeorge: Agree, in addition, we could use this type for error from external errors like the Novu API error when we want to forward them to the user.

packages/framework/src/utils/env.utils.ts
rifont: Yep definitely, sounds good to address later.

packages/framework/src/types/event.types.ts
rifont: Sounds good, I agree. Will update

packages/framework/src/workflow.ts
rifont: If I'm consuming this `workflow.trigger()` method in my Typescript application, I would already have validated that my integration is correct via strong `payload` typings. Therefore if we're entering a state of an invalid `payload` being present, to me that's more of an exceptional state that warrants an error being thrown.

This is a wider conversation about client error handling that we need to look at. Right now, our `@novu/api` Speakeasy SDK is throwing an error here whenever invalid parameters are provided to a REST-API SDK method, I think we should maintain consistency with the opionated Speakeasy implementation.

packages/framework/src/errors/execution.errors.ts
rifont: This makes total sense. I raised https://linear.app/novu/issue/NV-4060/consolidate-errors-into-a-single-class-to-reduce-novuframework-bundle for us to review later.

packages/framework/src/types/discover.types.ts
rifont: Raised https://linear.app/novu/issue/NV-4061/add-triggerbulk-and-triggerbroadcast-to-defined-novuframework toa ddress this.

packages/framework/src/utils/env.utils.ts
rifont: Raised https://linear.app/novu/issue/NV-4062/refine-the-novuframework-well-knownnovu-discovery-algorithm

packages/framework/src/types/event.types.ts
rifont: The proper fix is to modify our API DTO naming. For now, I am using this Typescript trick:
```typescript
// eslint-disable-next-line @typescript-eslint/ban-types
type Actor = TriggerRecipientSubscriber & {};
```

is this acceptable?

packages/framework/src/workflow.ts
SokratisVidros: You are right about consistency. I added the Speakeasy team in the loop. We will probably release as such.

packages/framework/src/types/event.types.ts
rifont: I have taken this approach and added the `type` import directive for `@novu/shared` rather than declare all the types duplicately in `@novu/framework` as the nested typings are numerous.

packages/framework/src/workflow.ts
rifont: The `workflow()` doesn't have access to the `Client.apiClient` as the Workflows come before the Client, not after. The current approach of instantiating a global client here isn't ideal, but it saves us from having a global client living in the installed NPM package.

@SokratisVidros , do we want to move ahead with this for now and revisit as part of adding the `@novu/api` REST-API client?

SokratisVidros: Sure.

SokratisVidros (approved):

PR #5880, opened 2024-06-30
refactor(framework): Use `@novu/api` Client for Workflow trigger and typings

rifont:
* Use `@novu/api` Client for Workflow trigger and inherit typings for (`workflow.trigger()`)
* Remove `@novu/shared`, as necessary typings all come from `@novu/api`
* Rename colliding `@novu/api` (the API Service) to `@novu/api-server`

### Screenshots

<details>

</details>


packages/framework/src/workflow.test.ts
```diff
@@ -100,7 +104,7 @@ describe('workflow function', () => {
         testWorkflow.trigger({
           // @ts-expect-error - foo is missing from the payload
           payload: {},
-          to: 'test@test.com',
+          to: ['test@test.com'],
```
rifont, line 107: Adjusting per the typing on the new SDK

rifont:
With the current changes, I experienced issues `serve`ing the Workflows. Nextjs appears to silently fail when rendering the needed routes in `serve`, resulting in all health check calls to `http://localhost:4000/api/novu?action=health-check` to fail with a `404` not found error.

I have narrowed the issue down to the new `@novu/api`.`Novu` client being used - when this class instantiation is removed across `handler.ts` and `workflow.ts`, the route is served as normal. Further debugging needed.

PR #5889, opened 2024-07-01
feat(api, worker): add support for stateless step controls

djabarovgeorge:

Currently there is no way to test a workflow with modified controls in the Local Studio.
Add Backend and API Support

### Screenshots

<details>

### Related enterprise PR
[EE-PR](https://github.com/novuhq/packages-enterprise/pull/158)

</details>


apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
```diff
@@ -95,6 +90,7 @@ export class SubscriberJobBound {
       source: command.payload.__source || 'api',
       subscriberSource: _subscriberSource || null,
       requestCategory: requestCategory || null,
+      statelessWorkflow: !!command.bridge?.url,
```
djabarovgeorge, line 93: added indication for us to use in the metrics so would know if the trigger was stateless or not.

apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
```diff
@@ -108,11 +104,6 @@ export class SubscriberJobBound {
 
     // If no subscriber makes no sense to try to create notification
     if (!subscriberProcessed) {
-      /**
```
djabarovgeorge, line 111: removing this one as we could potentially have a lot more comments like those, we should revisit the flow until the creation of the job once we won't be limited by the job entity, and add execution details in the way.

libs/dal/src/repositories/notification-template/notification-template.entity.ts
```diff
@@ -127,6 +128,8 @@ export class StepVariantEntity implements IStepVariant {
   shouldStopOnFail?: boolean;
 
   bridgeUrl?: string;
+
+  controlVariables?: ControlsDto;
```
djabarovgeorge, line 132: called as `controlVariables` and not `controls` because we already have the `controls` on the step which is the step schema.

apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
```diff
@@ -172,8 +162,14 @@ export class SubscriberJobBound {
       ...bridgeWorkflow,
       type: 'ECHO',
       steps: bridgeWorkflow.steps.map((step) => {
+        const stepControlVariables = command.payload.controls?.[ControlVariablesLevelEnum.STEP_CONTROLS].find(
```
scopsy, line 165: Let's add a "?" here?

scopsy (approved):

PR #5892, opened 2024-07-01
fix(web): set cookie as secure

ainouzgali:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/web/src/utils/cookies.ts
+++ b/apps/web/src/utils/cookies.ts
@@ -42,7 +42,7 @@ const ONBOARDING_COOKIE_EXPIRY_DAYS = 10 * 365;
 export function setNovuOnboardingStepCookie() {
   return novuOnboardedCookie.set('1', {
     expires: ONBOARDING_COOKIE_EXPIRY_DAYS,
-    sameSite: 'none',
-    secure: true,
+    sameSite: 'None',
+    secure: window.location.protocol === 'https:',
   });
 }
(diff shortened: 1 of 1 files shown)


apps/web/src/utils/cookies.ts
```diff
@@ -43,6 +43,6 @@ export function setNovuOnboardingStepCookie() {
   return novuOnboardedCookie.set('1', {
     expires: ONBOARDING_COOKIE_EXPIRY_DAYS,
     sameSite: 'none',
-    secure: window.location.protocol === 'https',
+    secure: true,
```
SokratisVidros: @ainouzgali `true` doesn't work for localhost environments when server over http. 

Actually the previous check was wrong. An `:` was missing.
`secure: window.location.protocol === 'https:'`

apps/web/src/utils/cookies.ts
```diff
@@ -43,6 +43,6 @@ export function setNovuOnboardingStepCookie() {
   return novuOnboardedCookie.set('1', {
     expires: ONBOARDING_COOKIE_EXPIRY_DAYS,
     sameSite: 'none',
```
SokratisVidros: There is one more type here:

```suggestion
    sameSite: 'None',
```

SokratisVidros (changes requested):

SokratisVidros (approved):

PR #5895, opened 2024-07-01
feat(web): Add Clerk components for the EE version

ChmaraX:
## What changed? Why was the change needed?
We want to use Clerk components in the enterprise version to simplify authN and authZ processes.

TODO:
- [ ] implement environment switch changes from https://github.com/novuhq/novu/pull/5768
- [ ] skip tests that are replaced with Clerk components (we probably don't want to test those)
- [x] change API client - the current way of getting token/cookies is not ideal
- [x] proper styling/localisation of new components (can be done later)

### Screenshots

https://github.com/novuhq/novu/assets/32128755/e3941bef-63f5-4132-958f-006741b50478

https://github.com/novuhq/novu/assets/32128755/72ac3523-86ea-42e8-81dd-44e10105493f

https://github.com/novuhq/novu/assets/32128755/c94add79-25d0-4640-99f8-d11eb9084e11

### Related PRs
https://github.com/novuhq/novu/pull/5755

https://github.com/novuhq/packages-enterprise/pull/147

### Clerk components

Added Clerk components which are toggled using feature flag / env var:
- SignIn component
- SignUp component
- UserProfile/UserButton
- OrganizationSwitcher

### EnterpriseAuthProvider
Auth provider that handles Clerk related auth processes. It's equivalent of original `useAuth` hook, but its reworked to be a provider / singleton instead of a hook, now the side effects are not triggered on each hook use resulting in much better performance. I think this could be done in the community version as well.

### Changes in registration flow:
There is a change in how the registration flow is being executed.

**Before:** 
the entire registration including user + organization + questionnaire was done in a single step and API call

**Now:** 
First the user is created in Clerk and synced with internal DB, then the organization is created and synced, and finally the organization is updated with questionnaire data. 

The diffence is that this can potentially cause incomplete registration, where e.g. user stops after user creation, so we will end up with user in database but no organization. 
 
### User profile / settings
The aim is to combine organization and user settings to single modal, which are originally two separate pages/modals in Clerk and there seems to be no "official" way to combine them.  

The way how its done now is a hack, where we hide the inner navbar and change the pre-selected inner navbar item by reordering the items:
image
image

**Addional notes**
The original user profile items `Profile` and `Security` probably cannot be renamed and the icon can't be changed aswell:
image


apps/web/src/AppRoutes.tsx
```diff
@@ -58,20 +58,26 @@ import {
   WorkflowsTestPage,
 } from './studio/components/workflows';
 import { WorkflowsStepEditorPageV2 } from './pages/templates/editor_v2/TemplateStepEditorV2';
+import { IS_CLERK_AUTH_ENABLED } from './config/index';
```
SokratisVidros: Let's make this vendor agnostic.

```suggestion
import { IS_EE_AUTH_ENABLED } from './config/index';
```

apps/web/src/components/layout/components/PrivatePageLayout.tsx
```diff
@@ -45,7 +47,22 @@ export function PrivatePageLayout() {
...
 
-  return (
+  function clerkPrivateWrapper(children: React.ReactNode) {
+    if (!IS_CLERK_AUTH_ENABLED) {
+      return <>children</>;
+    }
+
+    return (
+      <>
+        <SignedIn>{children}</SignedIn>
```
SokratisVidros: Is it possible to add these in the AppRoutes files or does react-router-dom complain about <Route> not being a direct children of <Router/>?

apps/web/src/components/nav/RootNavMenu.tsx
```diff
@@ -44,7 +46,7 @@ export const RootNavMenu: React.FC = () => {
   return (
     <NavMenu variant="root">
       <NavMenuSection>
-        <OrganizationSelect />
+        {IS_CLERK_AUTH_ENABLED ? <OrganizationSwitcher /> : <OrganizationSelect />}
```
SokratisVidros: How can we remove this ternary duplication in the codebase? Should we use a facade component such ash <OrganizationSwitcher/> that does the ternary inside, or maybe a react higher-order component (i.e. `withAuthComponent(OrgSwitcher, OrgSelect)`  to do the feature flag check in there and return argument 1 or 2?

apps/web/src/config/index.ts
```diff
@@ -68,3 +68,13 @@ export const FEATURE_FLAGS = Object.values(FeatureFlagsKeysEnum).reduce((acc, ke
...
 export const IS_EU_ENV = (ENV === 'production' || ENV === 'prod') && API_ROOT.includes('eu.api.novu.co');
+
+export const IS_CLERK_AUTH_ENABLED =
+  window._env_.REACT_APP_IS_CLERK_AUTH_ENABLED === 'true' || process.env.REACT_APP_IS_CLERK_AUTH_ENABLED === 'true';
+
+export const CLERK_PUBLISHABLE_KEY =
+  window._env_.REACT_APP_CLERK_PUBLISHABLE_KEY || process.env.REACT_APP_CLERK_PUBLISHABLE_KEY;
+
+if (!CLERK_PUBLISHABLE_KEY) {
+  throw new Error('Missing Clerk Publishable Key');
```
SokratisVidros, line 79: 💯

apps/web/src/config/index.ts
```diff
@@ -68,3 +68,13 @@ export const FEATURE_FLAGS = Object.values(FeatureFlagsKeysEnum).reduce((acc, ke
 export const HUBSPOT_PORTAL_ID = window._env_.REACT_APP_HUBSPOT_EMBED || process.env.REACT_APP_HUBSPOT_EMBED;
 
 export const IS_EU_ENV = (ENV === 'production' || ENV === 'prod') && API_ROOT.includes('eu.api.novu.co');
+
+export const IS_CLERK_AUTH_ENABLED =
```
SokratisVidros: You can replace this feature flag by the presence of `CLERK_PUBLISHABLE_KEY`

apps/web/src/ee/clerk/ClerkRoutes.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <>
+      <Route
+        element={
+          <SignedOut>
+            <PublicPageLayout />
+          </SignedOut>
+        }
+      >
+        <Route path={ROUTES.AUTH_SIGNUP + '/*'} element={<SignUpPage />} />
```
SokratisVidros: Let's replace these with /sign-in, and /sign-up

apps/web/src/ee/clerk/ClerkRoutes.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <Route path={ROUTES.AUTH_LOGIN + '/*'} element={<SignInPage />} />
+      </Route>
+      <Route
+        element={
+          <SignedIn>
+            <PublicPageLayout />
+          </SignedIn>
+        }
+      >
+        <Route path={ROUTES.AUTH_SIGNUP_ORGANIZATION_LIST} element={<OrganizationListPage />} />
```
SokratisVidros: Why is this route necessary? The SignUp component should display this page since organizations are enabled for our instance.

apps/web/src/ee/clerk/components/UserProfileButton.tsx
```diff
@@ -0,0 +1,27 @@
+import { UserButton, UserProfile } from '@clerk/clerk-react';
+import { IconCreditCard, IconGroup, IconRoomPreferences, IconWorkspacePremium } from '@novu/design-system';
+import { ROUTES } from '../../../constants/routes';
+import { BrandingPage } from '../../../pages/brand/BrandingPage';
+import { BillingPage } from '../../billing/pages/BillingPage';
+import { CustomOrganizationProfile } from './CustomOrganizationProfile';
+
+export function UserProfileButton() {
```
SokratisVidros, line 8: 💯

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+export const useAuthEnterpriseContext = () => {
+  const context = useContext(AuthEnterpriseContext);
+  if (context === undefined) {
+    throw new Error('useAuthEnterpriseContext must be used within an AuthEnterpriseProvider');
+  }
+
+  return context;
+};
+
+const _AuthEnterpriseProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
SokratisVidros: Technically this is the community auth provider. The enterprise is the Clerk one. So I'd wrap the application with an AuthProvider component or use the HOC I mentioned before depending if the app loads in EE or community edition.

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+      });
+    } else {
+      ldClient.identify({
+        kind: 'user',
+        anonymous: true,
+      });
+    }
+  }, [ldClient, organization]);
+
+  const memoizedValues = useMemo(
```
SokratisVidros: Good idea :)

apps/web/src/utils/utils.ts
```diff
@@ -26,3 +26,18 @@ export function parsePayload(payload: string) {
     return {};
   }
 }
+
+export function getCookies(cookieName: string): string | null {
```
SokratisVidros: The repo includes js-cookies lib so you can use that. Having said that, eventually you shouldn't need it as you need `await Clerk.session.getToken()` to get the latest session token.

SokratisVidros (changes requested):

apps/web/src/components/layout/components/PrivatePageLayout.tsx
```diff
@@ -45,7 +47,22 @@ export function PrivatePageLayout() {
     [environment]
   );
 
-  return (
+  function clerkPrivateWrapper(children: React.ReactNode) {
```
antonjoel82: ⚠ **issue:** I believe there are performance concerns with this since it will re-render _the entire app / component tree‏_ on every re-render. https://stackoverflow.com/a/59776934

apps/web/src/config/index.ts
antonjoel82: ❓ **question:** How will this affect self-hosted users?‏

apps/web/src/ee/clerk/components/OrganizationSwitcher.tsx
```diff
@@ -0,0 +1,21 @@
+import { OrganizationSwitcher as ClerkOrganizationSwitcher } from '@clerk/clerk-react';
+
+// TODO: this is tmp. styling
+const OrganizationSwitcherAppearance = {
+  elements: {
+    organizationSwitcherTrigger: {
+      paddingInline: 'var(--nv-spacing-125)',
```
antonjoel82: 👏 **praise:** Nice! Can also use 'token(spacing.125)'‏

apps/web/src/ee/clerk/components/OrganizationSwitcher.tsx
```diff
@@ -0,0 +1,21 @@
...
+      fontSize: '14px',
+      width: '240px',
+      height: '52px',
```
antonjoel82: ☑ **chore:** Please use tokens for the rest of these, especially font size‏

apps/web/src/ee/clerk/components/UserProfileButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <UserButton afterSignOutUrl={ROUTES.AUTH_LOGIN}>
+      <UserButton.UserProfilePage label="account" />
+      <UserButton.UserProfilePage label="Organisation" url={ROUTES.ORGANIZATION} labelIcon={<IconRoomPreferences />}>
```
antonjoel82:
```suggestion
      <UserButton.UserProfilePage label="Organization" url={ROUTES.ORGANIZATION} labelIcon={<IconRoomPreferences />}>
```

apps/web/src/ee/clerk/components/CustomOrganizationProfile.tsx
```diff
@@ -0,0 +1,46 @@
...
+    );
+  }
+
+  return (
+    <OrganizationProfile appearance={OrganizationProfileAppearance}>
+      <OrganizationProfile.Page label="members" />
+      <OrganizationProfile.Page label="general" />
+    </OrganizationProfile>
+  );
+};
```
antonjoel82, line 46: ✏ **suggestion:** Just do two separate components that use the same appearance‏

apps/web/src/ee/clerk/components/UserProfileButton.tsx
```diff
@@ -0,0 +1,27 @@
+import { UserButton, UserProfile } from '@clerk/clerk-react';
+import { IconCreditCard, IconGroup, IconRoomPreferences, IconWorkspacePremium } from '@novu/design-system';
```
antonjoel82:
```suggestion
import { IconCreditCard, IconGroup, IconRoomPreferences, IconWorkspacePremium } from '@novu/novui/icons';
```

apps/web/src/ee/clerk/components/UserProfileButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { CustomOrganizationProfile } from './CustomOrganizationProfile';
+
+export function UserProfileButton() {
+  return (
+    <UserButton afterSignOutUrl={ROUTES.AUTH_LOGIN}>
+      <UserButton.UserProfilePage label="account" />
+      <UserButton.UserProfilePage label="Organisation" url={ROUTES.ORGANIZATION} labelIcon={<IconRoomPreferences />}>
+        <CustomOrganizationProfile firstItem="general" />
+      </UserButton.UserProfilePage>
+      <UserButton.UserProfilePage label="security" />
```
antonjoel82, line 15:
```suggestion
      <UserButton.UserProfilePage label="Security" />
```

apps/web/src/ee/clerk/components/UserProfileButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { IconCreditCard, IconGroup, IconRoomPreferences, IconWorkspacePremium } from '@novu/design-system';
+import { ROUTES } from '../../../constants/routes';
+import { BrandingPage } from '../../../pages/brand/BrandingPage';
+import { BillingPage } from '../../billing/pages/BillingPage';
+import { CustomOrganizationProfile } from './CustomOrganizationProfile';
+
+export function UserProfileButton() {
+  return (
+    <UserButton afterSignOutUrl={ROUTES.AUTH_LOGIN}>
+      <UserButton.UserProfilePage label="account" />
```
antonjoel82, line 11:
```suggestion
      <UserButton.UserProfilePage label="Account" />
```

apps/web/src/ee/clerk/pages/OrganizationListPage.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { ROUTES } from '../../../constants/routes';
+
+export default function OrganizationListPage() {
+  return (
+    <AuthLayout
+      title="Select or create organization"
+      description="Please select or create an organization to continue."
+    >
+      <OrganizationList
+        hidePersonal={true}
```
antonjoel82:
```suggestion
        hidePersonal
```

apps/web/src/ee/clerk/pages/QuestionnairePage.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { HubspotSignupForm } from '../../../pages/auth/components/HubspotSignupForm';
+
+export default function QuestionnairePage() {
+  // TODO: Remove vercel integration logic from this page
+  const { isLoading } = useVercelIntegration();
+  const isHubspotFormFeatureFlagEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_HUBSPOT_ONBOARDING_ENABLED);
+  const isHubspotEnabled = HUBSPOT_PORTAL_ID && isHubspotFormFeatureFlagEnabled;
+
+  if (isLoading) {
+    <SetupLoader title="Loading..." />;
```
antonjoel82:
```suggestion
    return <SetupLoader title="Loading..." />;
```

apps/web/src/ee/clerk/pages/QuestionnairePage.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  const renderQuestionnaireForm = () => {
+    if (isHubspotEnabled) {
+      return <HubspotSignupForm />;
+    }
+
+    return <QuestionnaireForm />;
+  };
+
+  return <AuthLayout title="Tell us more about you">{renderQuestionnaireForm()}</AuthLayout>;
```
antonjoel82:
```suggestion
  return <AuthLayout title="Tell us more about you">{isHubspotEnabled ? <HubspotSignupForm /> : <QuestionnaireForm />}</AuthLayout>;
```

apps/web/src/ee/clerk/pages/SignUpPage.tsx
```diff
@@ -0,0 +1,15 @@
+import { SignUp } from '@clerk/clerk-react';
+import AuthLayout from '../../../components/layout/components/AuthLayout';
+import { ROUTES } from '../../../constants/routes';
+
+export default function SignUpPage() {
+  return (
+    <AuthLayout title="Sign Up" description="Hello and welcome! Sign up to the best notifications platform ever">
```
antonjoel82, line 7: ❓ **question:** Is this a temporary description? We may want marketing's input here‏

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
```diff
@@ -0,0 +1,52 @@
+import { PropsWithChildren } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { CLERK_PUBLISHABLE_KEY, IS_CLERK_AUTH_ENABLED } from '../../../config/index';
+import { ClerkProvider as _ClerkProvider } from '@clerk/clerk-react';
+import { useColorScheme } from '@novu/design-system';
+import { dark } from '@clerk/themes';
+
+const ClerkModalElement = {
+  modalContent: {
+    width: '80rem',
```
antonjoel82, line 10: ☑ **chore:** Use token or add TODO for using token‏

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+const localization = {
+  userProfile: {
+    navbar: {
+      title: 'Settings',
+      description: 'Manage your account settings',
+    },
+  },
+};
+
+export const ClerkProvider: React.FC<PropsWithChildren<{}>> = ({ children }) => {
```
antonjoel82, line 30:
```suggestion
export const ExternalAuthProvider: React.FC<PropsWithChildren<{}>> = ({ children }) => {
```
Or something else non-vendor specific

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+  defaultLocale: clerkOrganization.publicMetadata.defaultLocale,
+  domain: clerkOrganization.publicMetadata.domain,
+  productUseCases: clerkOrganization.publicMetadata.productUseCases,
+});
+
+const AuthEnterpriseContext = createContext<AuthEnterpriseContextProps | undefined>(undefined);
+
+export const useAuthEnterpriseContext = () => {
+  const context = useContext(AuthEnterpriseContext);
+  if (context === undefined) {
```
antonjoel82:
```suggestion
  if (!context) {
```

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+  const logout = useCallback(() => {
+    queryClient.clear();
+    segment.reset();
+    navigate(ROUTES.AUTH_LOGIN);
+    signOut();
+  }, [navigate, queryClient, segment, signOut]);
+
+  useEffect(() => {
+    if (orgId || !userId) return;
```
antonjoel82: ❓ **question:** Why this check?‏

antonjoel82:
```suggestion
    if (orgId || !userId) {
      return;
    }
```

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
antonjoel82, line 1: ✏ **suggestion:** With a file this large, more comments would be helpful!‏

antonjoel82 (commented):
Thanks for your hard work on this! Just left some comments / suggestions

apps/web/src/AppRoutes.tsx
```diff
@@ -58,20 +58,26 @@ import {
...
-        <Route path={ROUTES.AUTH_SIGNUP} element={<SignUpPage />} />
-        <Route path={ROUTES.AUTH_LOGIN} element={<LoginPage />} />
-        <Route path={ROUTES.AUTH_RESET_REQUEST} element={<PasswordResetPage />} />
-        <Route path={ROUTES.AUTH_RESET_TOKEN} element={<PasswordResetPage />} />
-        <Route path={ROUTES.AUTH_INVITATION_TOKEN} element={<InvitationPage />} />
-        <Route path={ROUTES.AUTH_APPLICATION} element={<QuestionnairePage />} />
-      </Route>
+      {!IS_CLERK_AUTH_ENABLED ? (
+        <Route element={<PublicPageLayout />}>
+          <Route path={ROUTES.AUTH_SIGNUP} element={<SignUpPage />} />
```
SokratisVidros: If we move the new endpoints under /sign-up and /sign-in there is no need of a feature flag. It also makes the URL schema much simpler.

apps/web/src/AppRoutes.tsx
```diff
@@ -58,20 +58,26 @@ import {
...
+          <Route path={ROUTES.AUTH_LOGIN} element={<LoginPage />} />
+          <Route path={ROUTES.AUTH_RESET_REQUEST} element={<PasswordResetPage />} />
```
SokratisVidros: @ChmaraX Do we still need these paths if we set Clerk managing all `/sign-in/*` and `/sign-up/*` catch all routes?

apps/web/src/AppRoutes.tsx
```diff
@@ -114,28 +120,30 @@ export const AppRoutes = () => {
...
-          <Route path={`${ROUTES.BILLING}/*`} element={<BillingPage />} />
-          <Route path={ROUTES.SECURITY} element={<AccessSecurityPage />} />
-          <Route path={`${ROUTES.SETTINGS}`} element={<Navigate to={ROUTES.PROFILE} replace />} />
-          <Route path="permissions" element={<Navigate to={ROUTES.SECURITY} replace />} />
-          <Route path="sso" element={<Navigate to={ROUTES.SETTINGS} replace />} />
-          <Route path="data-integrations" element={<Navigate to={ROUTES.SETTINGS} replace />} />
-          <Route path={ROUTES.PROFILE} element={<UserProfilePage />} />
-          <Route path={`${ROUTES.SETTINGS}/*`} element={<Navigate to={ROUTES.SETTINGS} replace />} />
-        </Route>
+        {!IS_CLERK_AUTH_ENABLED ? (
```
SokratisVidros: @ChmaraX can you elaborate on why we need this change?

apps/web/src/ee/clerk/ClerkRoutes.tsx
ChmaraX: I will replace the paths in a separate PR cause there are a lot of hardcoded references for this even outside of web, just to name few:
- various env vars in API (VERCEL_REDICERT_URI, CLIENT_SUCCESS_AUTH_REDIRECT...)
- tests in api
- places in CLI
- ...

apps/web/src/ee/clerk/components/OrganizationSwitcher.tsx
ChmaraX: I've changed the spacing to use tokens and removed the fontSize -- This is just temporary so it roughly fits and we can start testing it - we will open new PR with custom styling by Nikolay for all the Clerk components.

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
ChmaraX: I see the point but I'm wondering if renaming this wouldn't cause more confusion since this is specifically ClerkProvider same as e.g. HelmetProvider, it providers just Clerk and nothing else. Because we also already have `EnterpriseAuthProvider`, adding `ExternalAuthProvider` could be confusing.

ChmaraX: I moved the `ClerkProvider` inside `AuthEnterpriseProvider` - this way we have just one auth provider for EE so it should be less confusing and its also vendor agnostic.

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
ChmaraX: - removed the `userId` one cause its redundant
- `orgId` is used to check if user has already active organization, in such case we dont need to choose any

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
ChmaraX: I added bunch of comments to the useEffects - will try to revise this class again after we test/release everything to see if there is a place to simplify things

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
ChmaraX: This is actually the enterprise one, the community one is `useAuthCommunity` hook - but I've moved the ClerkProvider inside this class so we have just one provider for EE.

apps/web/src/components/nav/RootNavMenu.tsx
ChmaraX: Both options are feasible, I think the questions is just how nested will the if/else be. For now I tried to follow existing conventions and keep it as simple as possible. 

What could also matter is how different the EE version will be - e.g. if we plan to change the `RootNavMenu` more, it could make more sense to have `EERootNavMenu` cause we are hiding the settings nav item here.

apps/web/src/utils/utils.ts
ChmaraX: it looks like `js-cookie` is just a dependency of some other library, we don't have it as direct dependency, so I thought I would avoid installing it for few lines of code which will get eventually changed anyway

apps/web/src/ee/clerk/components/UserProfileButton.tsx
ChmaraX: this one unfortunately cannot be changed (or I haven't yet found a way how)

```typescript
type UserProfilePageProps = PageProps<'account' | 'security'>;
```

apps/web/src/ee/clerk/components/UserProfileButton.tsx
ChmaraX: this one unfortunately cannot be changed (or I haven't yet found a way how)

```typescript
type UserProfilePageProps = PageProps<'account' | 'security'>;
```

apps/web/src/ee/clerk/pages/SignUpPage.tsx
ChmaraX: Yes, all descriptions and styles on the Clerk components are just temporary and will change.

ChmaraX:
TODO: will rebase on top of next when the related PRs get merged

apps/web/src/ee/clerk/ClerkRoutes.tsx
ChmaraX: The `OrganizationListPage` is necessary otherwise it would open Clerk site with the component, the route is probably not not necessary but when I do `ROUTES.AUTH_SIGNUP + '/*'` (or `'auth/signup/*'` it doesn't work for some unknown reason, the only format it works is `'/auth/singup/:something`, otherwise it wont display the org select list at all

apps/web/src/auth/getToken.ts
```diff
@@ -1,5 +1,13 @@
+import { IS_EE_AUTH_ENABLED } from '../config/index';
+import { getCookies } from '../utils/utils';
 import { LOCAL_STORAGE_AUTH_TOKEN_KEY } from './auth.const';
 
 export function getToken(): string {
-  return localStorage.getItem(LOCAL_STORAGE_AUTH_TOKEN_KEY) || '';
+  const token = IS_EE_AUTH_ENABLED ? getCookies('__session') : localStorage.getItem(LOCAL_STORAGE_AUTH_TOKEN_KEY);
```
SokratisVidros: The JWT token from Clerk should always be fetched from `const { getToken} = useAuth()` or `window.Clerk.session.getToken()`.

apps/web/src/auth/CommunityAuthContextProvider.tsx
```diff
@@ -0,0 +1,24 @@
+import { createContext } from 'react';
```
SokratisVidros: Can we please move these under web/src/providers and rename to CommunityAuthProvider like all the other providers?

apps/web/src/auth/useCreateAuthContextCommunity.ts
```diff
@@ -44,7 +46,7 @@ function inIframe() {
 /**
  * TODO: this function should be decomposed into smaller, more focused pieces of functionality.
  */
-export function useCreateAuthContext() {
+export function useCreateAuthContextCommunity() {
```
SokratisVidros:
```suggestion
export function useCreateAuthCommunity() {
```

apps/web/src/ee/clerk/EnterpriseAuthRoutes.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <SignedOut>
+        <Navigate to={ROUTES.AUTH_LOGIN} replace />
+      </SignedOut>
+    </>
+  );
+
+  return (
+    <>
+      <Route element={<EnterprisePublicAuthLayout />}>
+        <Route path={ROUTES.AUTH_SIGNUP + '/*'} element={<SignUpPage />} />
```
SokratisVidros, line 32: Let's use `/sign-in` and `/sign-up` in the ee edition.

apps/web/src/ee/clerk/providers/useCreateAuthContextEnterprise.ts
```diff
@@ -0,0 +1,205 @@
...
+  const inPrivateRoute = !inPublicRoute;
+
+  const logout = useCallback(() => {
+    queryClient.clear();
+    segment.reset();
+    navigate(ROUTES.AUTH_LOGIN);
+    signOut();
+  }, [navigate, queryClient, segment, signOut]);
+
+  const redirectTo = useCallback(
```
SokratisVidros: This method shouldn't be needed in the Clerk realm as the components support redirect_url parameters. We can leave it for now but let's add a todo to get back to this.

apps/web/src/ee/clerk/providers/useCreateAuthContextEnterprise.ts
```diff
@@ -0,0 +1,205 @@
...
+    await queryClient.refetchQueries();
+  }, [queryClient]);
+
+  // check if user has active organization
+  useEffect(() => {
+    if (orgId) {
+      return;
+    }
+
+    if (isOrgListLoaded && clerkUser) {
```
SokratisVidros: I am not so sure why we need this callback and this logic. All the org and user logic should be handled by Clerk components.

apps/web/src/utils/utils.ts
SokratisVidros: It's already used in web here https://github.com/novuhq/novu/blob/d9814af178ce4466329c1293575593e8e40d7928/apps/web/src/utils/cookies.ts#L1

SokratisVidros (changes requested):

apps/web/src/AppRoutes.tsx
ChmaraX: The `/auth/login` URL is used and often times hardcoded on multiple places outside of web - various redirect uris, env variables, CLI commands - I would rather not do this change as part of this PR since it would increase the scope and it would be harder to track the issues related to this path change - lets do it in a separate one.

apps/web/src/AppRoutes.tsx
ChmaraX: These are the community ones

apps/web/src/ee/clerk/EnterpriseAuthRoutes.tsx
ChmaraX: The `/auth/login` URL is used and often times hardcoded on multiple places outside of web - various redirect uris, env variables, CLI commands - I would rather not do this change as part of this PR since it would increase the scope and it would be harder to track the issues related to this path change - lets do it in a separate one.

apps/web/src/ee/clerk/providers/useCreateAuthContextEnterprise.ts
ChmaraX: Yes I thought so as well, but It was added recently and I was not 100% sure so I kept it - we can remove this later when we confirm its not required.

apps/web/src/AppRoutes.tsx
ChmaraX: This comment is on outdated part of code, but we need this to render different set of routes based on FF:

```typescript
const AuthRoutes = () => {
  const CommunityAuthRoutes = () => (
    <Route element={<PublicPageLayout />}>
      <Route path={ROUTES.AUTH_SIGNUP} element={<SignUpPage />} />
      <Route path={ROUTES.AUTH_LOGIN} element={<LoginPage />} />
      <Route path={ROUTES.AUTH_RESET_REQUEST} element={<PasswordResetPage />} />
      <Route path={ROUTES.AUTH_RESET_TOKEN} element={<PasswordResetPage />} />
      <Route path={ROUTES.AUTH_INVITATION_TOKEN} element={<InvitationPage />} />
      <Route path={ROUTES.AUTH_APPLICATION} element={<QuestionnairePage />} />
    </Route>
  );

  return IS_EE_AUTH_ENABLED ? EnterpriseAuthRoutes() : CommunityAuthRoutes();
};
```

```typescript
export const EnterpriseAuthRoutes = () => {
  const EnterprisePublicAuthLayout = () => (
    <SignedOut>
      <PublicPageLayout />
    </SignedOut>
  );

  // private but we want appearance of public layout
  const EnterprisePrivateAuthLayout = () => (
    <>
      <SignedIn>
        <PublicPageLayout />
      </SignedIn>
      <SignedOut>
        <Navigate to={ROUTES.AUTH_LOGIN} replace />
      </SignedOut>
    </>
  );

  return (
    <>
      <Route element={<EnterprisePublicAuthLayout />}>
        <Route path={ROUTES.AUTH_SIGNUP + '/*'} element={<SignUpPage />} />
        <Route path={ROUTES.AUTH_LOGIN + '/*'} element={<SignInPage />} />
      </Route>
      <Route element={<EnterprisePrivateAuthLayout />}>
        <Route path={ROUTES.AUTH_SIGNUP_ORGANIZATION_LIST} element={<OrganizationListPage />} />
        <Route path={ROUTES.AUTH_APPLICATION} element={<QuestionnairePage />} />
      </Route>
    </>
  );
};
```

apps/web/src/auth/getToken.ts
ChmaraX: I agree but `getToken()` from `useAuth()` is async and we call `getToken()` synchronously inside the now legacy api.client inside `getHeader()` - we previously agreed to change this after the API client gets reworked 

```typescript
export const api = {
  get(url: string, options: IOptions = { absoluteUrl: false }) {
    return axios
      .get(buildUrl(url, options.absoluteUrl), {
        headers: getHeaders(),
      })
      .then((response) => {
      })
      .catch((error) => {
      });
  },
```

apps/web/src/AppRoutes.tsx
SokratisVidros: Do you have these places in mind? Are you talking about the marketing site? The CLI part is gone. We can keep the routes and redirect to sign-in or sign-up.

apps/web/src/auth/getToken.ts
SokratisVidros: Can we please try with Axios async interceptors? See here https://github.com/axios/axios/issues/3870

apps/web/src/ee/clerk/providers/useCreateAuthContextEnterprise.ts
ChmaraX: It is a guard that checks if user has an active organization. Maybe I'm missing something but by default when I sign up or sign in as a user, the Clerk JWT doesn't contain orgId. 

Without this check you would be able to access other private routes without an active organization / creating the organization first. So we need to eitehr force creation of an org or if user is already a member of some orgs, we chose the first one as default - alternatively we could display org select after login but that can be done now in the org switch selector inside an app as well.

I mostly followed this guide: https://clerk.com/docs/guides/force-organizations

apps/web/src/AppRoutes.tsx
ChmaraX: When I search for the path, there are 35 refs in 19 files in the project + maybe something outside of project like the marketing site you mention which we are not aware of. 

Changing this means also changing the .env vars which possibly includes changing of CI, I can also see some vercel env vars there, maybe something in vercel needs to be changed - no idea just thinking out loud but I personally wouldn't mix it up with this change.

SokratisVidros (approved):

apps/web/src/AppRoutes.tsx
SokratisVidros: I see. Let's leave it for this PR and we will address it separately.

PR #5897, opened 2024-07-01
fix(web): tracking events in studio

davidsoderberg:


apps/web/src/api/api.client.ts
```diff
@@ -128,6 +128,17 @@ export function buildApiHttpClient({
     }
   };
 
+  const post = async (url, body?: Record<string, unknown>) => {
```
antonjoel82:
```suggestion
  async function post<TBody = Record<string, unknown>> (url: string, body?: TBody) {
```

apps/web/src/utils/segment.ts
```diff
@@ -1,14 +1,26 @@
...
+
+type TrackFunction = UseMutateFunction<
+  any,
+  unknown,
+  {
+    event: string;
+    data?: Record<string, unknown> | undefined;
+  },
+  unknown
+>;
```
antonjoel82: 💭 **thought (non-blocking):** We should type this more intentionally‏

apps/web/src/components/providers/SegmentProvider.tsx
```diff
@@ -9,7 +10,8 @@ type Props = {
 const SegmentContext = React.createContext<SegmentService>(undefined as any);
 
 export const SegmentProvider = ({ children }: Props) => {
-  const segment = React.useMemo(() => new SegmentService(), []);
+  const { mutate: track } = useTelemetry();
```
rifont, line 13: ❓ **question:** Can we modify `useTelemetry` so we can do `const { track } = useTelemetry();` instead? This is a common pattern and object destructure renaming will be cumbersome + bad inner-DX‏

rifont (approved):
Nice work! 🚀

apps/web/src/utils/segment.ts
```diff
@@ -1,14 +1,26 @@
...
 
 export class SegmentService {
   private _segment: AnalyticsBrowser | null = null;
+  private readonly _trackFunction: TrackFunction;
   private _segmentEnabled: boolean;
   public _mixpanelEnabled: boolean;
 
-  constructor() {
+  constructor(trackFunction: TrackFunction) {
```
antonjoel82: ✏ **suggestion:** Convert this to an object to make it easier to extend with a smaller changeset in the future‏

scopsy (approved):

apps/web/src/hooks/useNovuAPI.ts
```diff
@@ -33,3 +34,28 @@ export const useApiKeys = (options?: any) => {
...
+      mixpanel.register({ $device_id: segmentDeviceId });
+      const sessionReplayProperties = mixpanel.get_session_recording_properties();
+
+      data = {
+        ...(data || {}),
+        ...sessionReplayProperties,
+      };
+    }
+
+    return mutate({ event: event + ' - [WEB]', data });
```
SokratisVidros, line 59:
```suggestion
    return mutate({ event: event + ' - [Dashboard]', data });
```

SokratisVidros (approved):

PR #5899, opened 2024-07-01
fix(web): Remove faulty cross-iframe link for Discord invite

rifont:
* Remove faulty cross-iframe link for Discord invite (we will come back to address support via Local Studio later)

### Screenshots

_Help button gone_
image

<details>

</details>

### DIFF
--- a/apps/web/src/components/layout/components/LocalStudioHeader/LocalStudioHeader.tsx
+++ b/apps/web/src/components/layout/components/LocalStudioHeader/LocalStudioHeader.tsx
@@ -1,10 +1,9 @@
 import { Header } from '@mantine/core';
 import { IconButton } from '@novu/novui';
 import { css } from '@novu/novui/css';
-import { IconHelpOutline, IconOutlineMenuBook } from '@novu/novui/icons';
+import { IconOutlineMenuBook } from '@novu/novui/icons';
 import { HStack } from '@novu/novui/jsx';
 import { FC } from 'react';
-import { discordInviteUrl } from '../../../../pages/quick-start/consts';
 import { useStudioWorkflowsNavigation } from '../../../../studio/hooks';
 import { HEADER_NAV_HEIGHT } from '../../constants';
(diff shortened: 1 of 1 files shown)


ainouzgali (approved):

apps/web/src/components/layout/components/LocalStudioHeader/LocalStudioHeader.tsx
```diff
@@ -35,10 +35,6 @@ export const LocalStudioHeader: FC = () => {
             target="_blank"
             rel="noopener noreferrer"
           />
-
-          {/* This doesn't work because of Discord's popup blocker via the response header:
-          Cross-Origin-Opener-Policy: same-origin-allow-popups. We will likely need a Javascript workaround for Discord's popup blocker. */}
-          <IconButton Icon={IconHelpOutline} as="a" href={discordInviteUrl} target="_blank" rel="noopener noreferrer" />
```
antonjoel82, line 41: 🤓 **nitpick (non-blocking):** Please remove associated unused imports‏

antonjoel82 (approved):
Looks good, but confirming that we time-boxed this and couldn't find a pleasing resolution?

rifont:
@antonjoel82 I looked at this problem in more detail yesterday. The main problem is that we're working in an iFramed application, and 3rd parties don't like to receive requests from such applications because they can be vulnerable to cross-origin attacks.

One of the suggested resolutions is to serve some html on a custom endpoint (outside of the SPA), link to that html, the HTML has some Javascript which redirects to an app. This is quite messy and will take a fair bit of time to get working with our current SPA setup.

My suggestion is that we circle back to this once we stop using the iFrame hack and instead use Webpack Module Federation or otherwise to serve the Local Studio component.

apps/web/src/components/layout/components/LocalStudioHeader/LocalStudioHeader.tsx
rifont: Good catch! - done

PR #5906, opened 2024-07-01
fix(web): store localhost url

denis-kralj-novu:


apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -33,14 +34,38 @@ export const BridgeUpdateModal: FC<BridgeUpdateModalProps> = ({ isOpen, toggleOp
...
+
+      const result = { isValid: response.status === 'ok' };
+
+      return result;
+    } catch {}
+
+    return { isValid: false };
+  };
+  const localDomains = ['localhost', '127.0.0.1'];
+  const isLocalAddress = () => {
```
denis-kralj-novu, line 50: I've added this to check if the address is local, but this only works for conventional address. We cannot really target host file modifications. I'm not sure if this is really a clean solution.

scopsy: I think this can work for now

scopsy (approved):

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -33,14 +34,38 @@ export const BridgeUpdateModal: FC<BridgeUpdateModalProps> = ({ isOpen, toggleOp
...
   const onUpdateClick = async () => {
     setUrlError('');
     setIsUpdating(true);
     try {
       if (url) {
         validateURL(url);
 
-        const result = await validateBridgeUrl({ bridgeUrl: url });
+        let result =
+          isLocalStudio && isLocalAddress()
```
denis-kralj-novu, line 62: we also need to support supplying the local tunnel address here as well as the local address

PR #5914, opened 2024-07-02
feat(web,novui): Studio <-> Dashboard Button, other small fixes

antonjoel82:
### Complete

- [Show test workflow button in Cloud](https://linear.app/novu/issue/NV-4024/test-workflow-should-be-visible-on-step-page-like-in-studio) 
- [Add button for navigating between modes](https://linear.app/novu/issue/NV-4004/no-formal-way-to-navigate-from-local-studio-to-the-dashboard-in-novu)
- Styling tweaks
  - Modal subtitle text
  - Full width buttons

https://www.loom.com/share/8ba89f5722204aacba9ab93aba686f0b?sid=36cdafc7-3ce4-411f-8fd1-f37776411b8e


scopsy:
@antonjoel82 I think we can update the dev-server.ts file with the following header, should resolve our cors issue there I believe:

```
  private serveWellKnownPath(req: http.IncomingMessage, res: http.ServerResponse) {
    res.setHeader('Content-Type', 'application/json');
    res.setHeader('Access-Control-Allow-Origin', '*');

    res.end(JSON.stringify(this.options));
  }
```

apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebar.tsx
```diff
@@ -1,33 +1,26 @@
...
+import { SidebarFooterButton } from './SidebarFooterButton';
 
 export const LocalStudioSidebar: WithLoadingSkeleton = () => {
   const { isLoading, data } = useDiscover();
 
+  const navigate = useStudioNavigate();
+
+  const goToCloudDashboard = () => {
+    navigate(window.location.origin, {});
+  };
```
scopsy: @antonjoel82 I think here we can just do `window.open(FRONT_END_URL (https://dashboard.novu.co)` in a new tab?

I want to make sure that the user understands it's a different application

apps/web/src/components/nav/RootNavMenu.tsx
```diff
@@ -148,7 +156,17 @@ export const RootNavMenu: React.FC = () => {
         ></NavMenuLinkButton>
       </NavMenuSection>
       <FreeTrialSidebarWidget />
-      <RootNavMenuFooter />
+      {isV2Enabled ? (
```
scopsy: @antonjoel82 for this particular condition can we use IS_V2_EXPERIENCE_ENABLED FF? instead of IS_V2_ENABLED?

The rest of the conditions should stay the same using the IS_V2_ENABLED condition

apps/web/src/studio/hooks/useNavigateToLocalStudio.ts
```diff
@@ -0,0 +1,63 @@
...
+ * TODO: repeatedly getting CORS error since origin appears as :4200 when trying to access :2022
+ *
+ * Get response, and navigate to the appropriate Local Studio URL. If no well-known response is available,
+ * open a modal.
+ */
+export const useNavigateToLocalStudio = ({ fallbackFn }: UseNavigateToLocalStudio = {}) => {
+  const navigate = useNavigate();
+
+  const forceStudioNavigation = () => {
+    navigate(DEFAULT_STUDIO_URL);
```
scopsy: Can we do a window.open here instead of just internal navigate?

antonjoel82:
> @antonjoel82 I think we can update the dev-server.ts file with the following header, should resolve our cors issue there I believe:
> 
> ```
>   private serveWellKnownPath(req: http.IncomingMessage, res: http.ServerResponse) {
>     res.setHeader('Content-Type', 'application/json');
>     res.setHeader('Access-Control-Allow-Origin', '*');
> 
>     res.end(JSON.stringify(this.options));
>   }
> ```

Yep, thank you! Denis and I updated it with a slightly more restrictive header

apps/web/src/components/nav/RootNavMenu.tsx
antonjoel82: Sure. Could you please explain the difference? This is where I think using selectors is important so we centralize logical distinctions with helpful comments

apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebar.tsx
antonjoel82: Sure, but what is `FRONT_END_URL`? I don't see anything similar in `web`

apps/web/src/components/nav/RootNavMenu.tsx
scopsy: Yes, the name is slightly confusing. But:
- is v2 enabled - Should only be enabled for new signups after date X (this hides some existing pages like changes page, layout page, etc....)
- is v2 experience enabled - Will be enabled for all users regarding their signup date

scopsy (approved):

PR #5919, opened 2024-07-02
fix(api): Ensure both web and dashboard subdomains work in parallel

SokratisVidros:
This is required for a smooth migration process and to avoid signing out existing users.


rifont (approved):
🚀

PR #5921, opened 2024-07-02
fix(framework): bug with duplicate step names

denis-kralj-novu:
- Move error detection to action endpoints
- Remove throwing on ID collision, log error instead 
- Add tests

### DIFF
--- a/packages/framework/src/client.ts
+++ b/packages/framework/src/client.ts
@@ -153,7 +153,7 @@ export class Client {
   public getErrors(): string[] {
     const errors: string[] = [];
 
-    const workflowIds = this.discoveredWorkflows.map((w) => w.workflowId);
+    const workflowIds = this.discoveredWorkflows.map((discoveredWorkflow) => discoveredWorkflow.workflowId);
 
     const workflowOccuranceMap = this.getOccuranceMap(workflowIds);
 
@@ -164,7 +164,7 @@ export class Client {
     }
 
     for (const workflow of this.discoveredWorkflows) {
-      const stepIds = workflow.steps.map((s) => s.stepId);
+      const stepIds = workflow.steps.map((discoveredStep) => discoveredStep.stepId);
 
       const stepOccuranceMap = this.getOccuranceMap(stepIds);
 
--- a/packages/framework/src/client.ts
+++ b/packages/framework/src/client.ts
@@ -95,10 +95,11 @@ export class Client {
   public addWorkflows(workflows: Array<Workflow>) {
     for (const workflow of workflows) {
       if (this.discoveredWorkflows.some((existing) => existing.workflowId === workflow.definition.workflowId)) {
-        throw new WorkflowAlreadyExistsError(workflow.definition.workflowId);
-      } else {
-        this.discoveredWorkflows.push(workflow.definition);
+        console.error(
+          `Workflow with id '${workflow.definition.workflowId}' defined more than once, workflowId needs to be unique`
+        );
       }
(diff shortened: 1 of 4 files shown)


packages/framework/src/handler.ts
```diff
@@ -176,8 +176,21 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
...
+        const errors = this.client.getErrors();
+
+        if (errors.length !== 0) {
+          return this.createResponse(HttpStatusEnum.UNPROCESSABLE_ENTITY, errors);
+        }
+
+        return this.triggerAction({ workflowId, ...body })();
+      },
       [PostActionEnum.EXECUTE]: async () => {
+        const errors = this.client.getErrors();
```
denis-kralj-novu, line 189: ❓ **question:** ‏ what is the usecase for execute? I'm not sure how it gets used in the studio and the bridge api client doesn't seem to use the action.

packages/framework/src/handler.ts
```diff
@@ -223,17 +241,27 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
...
       },
       [GetActionEnum.HEALTH_CHECK]: async () => {
-        const result = await this.client.healthCheck();
+        const result = this.client.healthCheck();
 
         return this.createResponse(HttpStatusEnum.OK, result);
       },
       [GetActionEnum.CODE]: async () => {
-        const result = await this.client.getCode(workflowId, stepId);
+        const errors = this.client.getErrors();
```
denis-kralj-novu, line 259: ❓ **question:** ‏ same as execute, what is the usecase for code? I'm not sure how it gets used in the studio and the bridge api client doesn't seem to use the action.

packages/framework/src/client.ts
```diff
@@ -149,6 +150,47 @@ export class Client {
     };
   }
 
+  public getErrors(): string[] {
+    const errors: string[] = [];
+
+    const workflowIds = this.discoveredWorkflows.map((discoveredWorkflow) => discoveredWorkflow.workflowId);
+
+    const workflowOccuranceMap = this.getOccuranceMap(workflowIds);
```
SokratisVidros, line 158: **suggestion:** ‏ We can do this check in a single pass with an assert function that will throw an error in the first duplicate detection using a reduce function. The accumulator is the occuranceMap.

packages/framework/src/handler.ts
```diff
@@ -176,8 +176,21 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
...
+        if (errors.length !== 0) {
+          return this.createResponse(HttpStatusEnum.UNPROCESSABLE_ENTITY, { errors });
+        }
+
+        return this.triggerAction({ workflowId, ...body })();
+      },
       [PostActionEnum.EXECUTE]: async () => {
+        const errors = this.client.getErrors();
+
+        if (errors.length !== 0) {
```
SokratisVidros, line 191: Let's generalize error handling across all action types to make the code dryer. Every byte counts in the framework SDK.

packages/framework/src/handler.ts
SokratisVidros: cc: @rifont ☝️

packages/framework/src/handler.ts
SokratisVidros: cc: @rifont ☝️

SokratisVidros (changes requested):

PR #5931, opened 2024-07-02
feat(root): persistent tunnel urls

djabarovgeorge:
Try to keep same tunnel url when relaunching the cli to avoid broken UI states using old values
We can save the random tunnel id in Config Store and reuse for subsequent times.

### Screenshots

<details>

</details>


packages/cli/src/commands/dev.ts
```diff
@@ -143,7 +144,45 @@ type LocalTunnelResponse = {
...
+async function createTunnel(localOrigin: string): Promise<string> {
+  const delay = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
+
+  const configTunnelUrl = config.getValue('tunnelUrl');
+  const storeUrl = configTunnelUrl ? new URL(configTunnelUrl) : null;
+  const originUrl = new URL(localOrigin);
+
+  if (storeUrl) {
+    await connectToTunnel(storeUrl, originUrl);
+    await delay(100);
```
djabarovgeorge: there is a race with the time it takes to establish a connection with the socket.
initially, I created a retry mechanism but in the end, i refactored to 100ms await to simplify the flow.

packages/cli/src/commands/dev.ts
```diff
@@ -143,7 +144,45 @@ type LocalTunnelResponse = {
...
+    const healthCheck = await tunnelHealthCheck(configTunnelUrl);
```
djabarovgeorge: the connection won't throw any error, therefore in order to check if we have a healthy tunnel connection we execute a health check request to the tunnel URL.

packages/cli/src/commands/dev/dev.ts
```diff
@@ -108,42 +82,43 @@ async function endpointHealthChecker(parsedOptions: DevCommandOptions, endpointR
...
+      return storeUrl.origin;
+    }
   }
-}
 
-type LocalTunnelResponse = {
-  id: string;
-  url: string;
-};
+  return await connectToNewTunnel(originUrl);
```
djabarovgeorge: 👀 this is the most important addition to the code base 👀

packages/cli/src/commands/dev/dev.ts
```diff
@@ -108,42 +82,43 @@ async function endpointHealthChecker(parsedOptions: DevCommandOptions, endpointR
...
-      return DashboardUrlEnum.EU;
-    case CloudRegionEnum.STAGING:
-      return DashboardUrlEnum.STAGING;
-    case CloudRegionEnum.US:
-    default:
-      return DashboardUrlEnum.US;
+  if (storeUrl) {
+    await connectToTunnel(storeUrl, originUrl);
+    await wait(100);
+    const healthCheck = await tunnelHealthCheck(configTunnelUrl + endpointRoute);
```
scopsy: The main problem I'm seeing here, is that we are not checking the health of the tunnel here, but the health of the bridge app in the tunnel. 

If my local app will be down for some reason, and the tunnel is completely healthy this will fail here. 

Can we maybe expose some sort of a `health` attribute returned by the NtfrClient?

djabarovgeorge: Totally agree this was my first approach however I was not able to expose any data from the connection. Let me do a quick check again :)

packages/cli/src/commands/dev/consts.ts
```diff
@@ -0,0 +1 @@
+export const TUNNEL_URL = 'https://novu.sh/api/tunnels';
```
SokratisVidros: 🤓 It's not worth the extra import and file for just this. Consider moving it to the dev.ts file.

packages/cli/src/commands/dev/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function wait(ms: number) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function getDefaultOrigin(port: string) {
+  return `http://${SERVER_HOST}:${port}`;
+}
+
+function getDefaultDashboardUrl(region: string) {
```
SokratisVidros, line 13: Replacing the enum with an object const will also simplify this logic to `CloudRegion[region]`.

packages/cli/src/commands/dev/dev.ts
```diff
@@ -108,42 +82,43 @@ async function endpointHealthChecker(parsedOptions: DevCommandOptions, endpointR
...
-function parseOptions(options: DevCommandOptions) {
-  const { origin, port, region } = options || {};
+async function tunnelHealthCheck(configTunnelUrl: string): Promise<boolean> {
+  try {
+    const res = await (
+      await fetch(configTunnelUrl + '?action=health-check', {
+        method: 'GET',
+        headers: {
+          accept: 'application/json',
+          'Content-Type': 'application/json',
```
SokratisVidros, line 95: For debugging: 

```suggestion
          'Content-Type': 'application/json',
          'User-Agent': 'novu@version-goes-here',
``

SokratisVidros (commented):
Nice work and structural clean up. Since we converted the dev command to a folder, consider moving the `dev-server` under the `dev` folder.

packages/cli/src/commands/dev/consts.ts
djabarovgeorge: yep agree i was extracting one by one, initially, i was not aware how many would go for each file 😅

packages/cli/src/commands/dev/dev.ts
djabarovgeorge: Done :)

packages/cli/src/commands/dev/utils.ts
djabarovgeorge: we are using the CloudRegionEnum as template literal types as well, I will keep it like this for now

scopsy (approved):

packages/cli/src/commands/dev/enums.ts
rifont, line 1: Nice!

SokratisVidros (approved):

PR #5932, opened 2024-07-02
fix(framework): Remove compile time secret key check

rifont:
* Remove compile time secret key check. This is necessary to ensure that the application can still build without a secret key being present in the environment
* Add missing `type` import directives to further reduce bundle size
* Convert remaining `interface`s to `type`s
* Publish @novu/framework@[0.24.3-alpha.16](https://www.npmjs.com/package/@novu/framework/v/0.24.3-alpha.16)

### Screenshots

_No critical dep issue during Next.js app startup `npm run dev`_
image

<details>

</details>

### DIFF
--- a/packages/framework/src/types/subscriber.types.ts
+++ b/packages/framework/src/types/subscriber.types.ts
@@ -1,4 +1,8 @@
 export type Subscriber = {
+  subscriberId?: string;
   firstName?: string;
   lastName?: string;
+  email?: string;
+  phone?: string;
+  avatar?: string;
 };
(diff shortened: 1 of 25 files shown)


packages/framework/src/types/subscriber.types.ts
```diff
@@ -1,4 +1,8 @@
 export type Subscriber = {
+  subscriberId?: string;
   firstName?: string;
   lastName?: string;
+  email?: string;
+  phone?: string;
+  avatar?: string;
```
rifont, line 7: Adding some additional subscriber fields

scopsy (approved):

PR #5933, opened 2024-07-02
fix(web): Automatically switch to dev env when navigating to local studio

antonjoel82:

https://linear.app/novu/issue/NV-4074/error-local-studio-works-only-with-development-api-keys

- Automatically switch to dev env when navigating to local studio -- opted for two prong approach:
  - Set env on "Open Local Studio" button as this should handle majority of use cases and avoids the "Cloud Dashboard" env select from getting out of sync
  - Set env during redirect -- this works (and will catch the occasional direct access, but causes the original "Cloud Dashboard" tab to not properly reflect the environment
- Fixed spacing on Sidebar between Free Trial widget and button 

### Questions for Reviewers

- Are there any risks associated with this?
- Do we need to handle the 2nd case when the env change happens during redirect and the original tab is (visually -- but not actually) out of sync?

### Screenshots

https://www.loom.com/share/c5295bbb1ef94b7ea094c72af3c0db2a?sid=e370930c-06b7-4bfe-b09a-a9091491910d

### DIFF
--- a/apps/web/src/components/layout/components/LocalStudioSidebar/SidebarFooter.tsx
+++ b/apps/web/src/components/layout/components/LocalStudioSidebar/SidebarFooter.tsx
@@ -18,7 +18,7 @@ export const SidebarFooter: FC<SidebarFooterProps> = ({ children, className }) =
         gradientTo={'surface.panel/100'}
         gradientToPosition={'80%'}
       />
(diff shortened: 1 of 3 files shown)


SokratisVidros:
@antonjoel82 we tried an alternative approach that saves us an extra fetch request. See https://github.com/novuhq/novu/pull/5938

PR #5938, opened 2024-07-03
fix(web): Always pick the secret key from development environment

SokratisVidros:
Always pick the secret key from development environment

This way, users won't get an error during their local studio authentication if they work on production environment in the cloud.

### DIFF
--- a/apps/web/src/studio/LocalStudioAuthenticator.tsx
+++ b/apps/web/src/studio/LocalStudioAuthenticator.tsx
@@ -1,5 +1,5 @@
 import { useEffect } from 'react';
-import { useAuth, useAPIKeys, useEnvironment } from '../hooks';
+import { useAuth, useEnvironment } from '../hooks';
 import { ROUTES } from '../constants/routes';
 import { assertProtocol } from '../utils/url';
 import { encodeBase64 } from './utils/base64';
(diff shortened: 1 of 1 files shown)


scopsy (approved):

apps/web/src/studio/LocalStudioAuthenticator.tsx
```diff
@@ -110,9 +105,14 @@ export function LocalStudioAuthenticator() {
     const localBridgeURL = buildBridgeURL(parsedApplicationOrigin.origin, tunnelPath);
     const tunnelBridgeURL = buildBridgeURL(tunnelOrigin, tunnelPath);
 
+    // TODO: Add apiKeys to the IEnvironment interface as they exist in the response
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    const devSecretKey = environments.find((env) => env.name.toLowerCase() === 'development')?.apiKey;
```
scopsy, line 111: @SokratisVidros just to clarify, is the apiKey here is defined on the environments object?  I remember it was an array of `apiKeys[0].key`

SokratisVidros: It does on the API layer, but the Interface in @novu/shared is not up to date. However this change will happen in my PR that removes envId from the JWT.

PR #5939, opened 2024-07-03
feat(novu): add analytics when dev command is running

davidsoderberg:


packages/cli/src/commands/dev.ts
```diff
@@ -34,9 +42,20 @@ export type DevCommandOptions = {
...
+
 export async function devCommand(options: DevCommandOptions) {
   await showWelcomeScreen();
 
+  analytics.track({
+    identity: {
+      anonymousId: anonymousId,
+    },
+    data: {},
+    event: 'Studio started - [Studio]',
```
scopsy:
```suggestion
    event: 'Studio started - [CLI]',
```

packages/cli/src/commands/dev.ts
```diff
@@ -5,6 +5,14 @@ import * as ora from 'ora';
 import * as open from 'open';
 import * as chalk from 'chalk';
 import { SERVER_HOST } from '../constants';
+import { v4 as uuidv4 } from 'uuid';
+import { AnalyticService, ConfigService } from '../services';
+
+const analytics = new AnalyticService();
+const config = new ConfigService();
+if (process.env.NODE_ENV === 'development') {
```
SokratisVidros, line 13: NODE_ENV won't be set in the separate TTY the developers use to run this command. When do we want to trigger the config clearance?

packages/cli/src/commands/dev.ts
```diff
@@ -34,9 +42,20 @@ export type DevCommandOptions = {
   route: string;
 };
 
+const anonymousIdLocalState = config.getValue('anonymousId');
+const anonymousId = anonymousIdLocalState || uuidv4();
```
SokratisVidros, line 46: This UUID is not correlated to anything else in the system. It would only demonstrate if there is a level of stickiness in a developer using the `dev` command. 

However, I still find that metric weak. I'd wait till the next iteration and do this with the actual user id.

PR #5965, opened 2024-07-03
chore(root): Only build and deploy in production

SokratisVidros:

Tests run on every PR in prod and next branches. To speed up release time, we removed them from production deployment github actions.

### DIFF
--- a/.github/workflows/prod-deploy-web.yml
+++ b/.github/workflows/prod-deploy-web.yml
@@ -9,14 +9,7 @@ on:
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
-  test_web:
-    uses: ./.github/workflows/reusable-web-e2e.yml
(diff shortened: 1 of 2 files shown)


.github/workflows/prod-deploy-web.yml
```diff
@@ -35,9 +28,6 @@ jobs:
     secrets: inherit
 
   deploy_web_us:
```
SokratisVidros, line 30: Run US/EU deployment in parallel. Let me know if there was a reason we were doing them sequentially.

davidsoderberg (approved):

.github/workflows/prod-deploy-web.yml
rifont: There is no good reason to do them separately. This is a good change.

rifont (approved):

PR #5969, opened 2024-07-03
fix(novu): From 60kb to 30KB

SokratisVidros:

1. Source maps are not needed yet
2. Declarations are not needed in a CLI
3. Do not include source code in the bundle
4. Tidy tsconfig files


packages/cli/tsconfig.json
```diff
@@ -1,12 +1,21 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "module": "commonjs",
+    "declaration": false,
+    "declarationMap": false,
+    "noImplicitAny": false,
```
denis-kralj-novu: ❓ **question:** ‏ Is there a good reason for setting `noImplicitAny` to false?

packages/cli/tsconfig.json
```diff
@@ -1,12 +1,21 @@
...
+    "removeComments": true,
     "allowSyntheticDefaultImports": true,
-    "types": ["node", "mocha", "chai", "sinon"],
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
```
denis-kralj-novu: 🤓 **nitpick (non-blocking):** ‏ We don't seem to use these, and it is still experimental, I would say remove it

packages/cli/tsconfig.spec.json
denis-kralj-novu, line 1: ❓ **question:** ‏ I don't understand why we have to have 2  different permutations of tsconfig, this variant seems to only be used in the `.selintrc.js` file. Apart from the `types` list and `allowJs` flag all the content is the same.

denis-kralj-novu (approved):
Few questions but in general good job

packages/cli/tsconfig.json
SokratisVidros: No reason at all. Great catch.

packages/cli/tsconfig.json
SokratisVidros: Done.

packages/cli/tsconfig.spec.json
SokratisVidros: I agree, we used to have three, not we have two, and we want to avoid adding specs to dist. I am happy to do another pass on this on all our packages at a later stage.

PR #5975, opened 2024-07-04
feat(web): add v2 components for docs

davidsoderberg:


apps/web/src/components/docs/Docs.tsx
```diff
@@ -81,6 +97,10 @@ export const Docs = ({ code = '', description = '', title = '', isLoading, child
   }
 
   if (Component === null) {
+    if (isChildDocs) {
+      return <Text>We could not load this part of the documentation for you. Please try again.</Text>;
```
scopsy: In what scenario will this happen? And what will change if the user will try again?

apps/web/src/components/docs/Docs.tsx
```diff
@@ -25,21 +27,35 @@ const DOCS_WRAPPER_ELEMENT_ID = 'embedded-docs';
...
     if (code.length === 0) {
       return null;
     }
 
     return getMDXComponent(code, {
+      framework: () => null,
+      NextStepsStep: () => <ChildDocs path="snippets/quickstart/next-steps" />,
+      TestStep: () => <ChildDocs path="snippets/quickstart/test" />,
+      DeployApp: () => <ChildDocs path="snippets/quickstart/deploy" />,
+      LocalStudio: () => <ChildDocs path="snippets/quickstart/start-studio" />,
```
scopsy: Will we need now for every snippet we are adding to modify the application? This might be happening frequently

apps/web/src/components/docs/Docs.tsx
davidsoderberg: If the docs could not be fetched from github for some reason. It could if it is github that is the issue. For me it is more a being nice to the user thing.

apps/web/src/components/docs/Docs.tsx
davidsoderberg: We will need to do this each time we import a new file that is not already mapped here.

rifont: @davidsoderberg , we need to find a better solution here. It's not feasible to need changes to our web-app every time a new Snippet is added.

We need to scrape the MDX content for snippets imports and populate this list dynamically to remove this tight coupling.

davidsoderberg: I agree will look into if we can auto resolve these on the backend

apps/web/src/components/docs/ChildDocs.tsx
```diff
@@ -0,0 +1,8 @@
+import { Docs } from './Docs';
+import { useLoadDocs } from './useLoadDocs';
+
+export const ChildDocs = ({ path }: { path: string }) => {
+  const { isLoading, data } = useLoadDocs({ path, isEnabled: true });
+
+  return <Docs isLoading={isLoading} isChildDocs={true} {...data}></Docs>;
```
antonjoel82:
```suggestion
  return <Docs isLoading={isLoading} isChildDocs {...data}></Docs>;
```

apps/web/src/components/docs/Docs.tsx
```diff
@@ -22,29 +24,69 @@ const getMDXComponent = mdxBundler.getMDXComponent;
...
 
+const globalOverrides = {
+  frameworkterminal: {
+    FrameworkTerminal: () => <nv-framework-terminal></nv-framework-terminal>,
+  },
+  framework: () => null,
+  echoterminal: {
+    EchoTerminal: () => <nv-framework-terminal></nv-framework-terminal>,
+  },
+};
```
antonjoel82: ☑ **chore:** Please add a comment explaining what these do / why they are needed‏

antonjoel82: ❓ **question:** Is 'echo' here our own keyword? If so, let's update it to match our new terminology‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -22,29 +24,69 @@ const getMDXComponent = mdxBundler.getMDXComponent;
...
+
+      return prev;
+    }, {} as Record<string, Record<string, any> | any>);
+
     return getMDXComponent(code, {
-      frameworkterminal: {
-        FrameworkTerminal: () => <nv-framework-terminal></nv-framework-terminal>,
-      },
+      ...globals,
+      ...globalOverrides,
```
antonjoel82: 💭 **thought (non-blocking):** This is slightly confusing to me -- so we are always "overriding" a few specific properties? Would it be better to just switch the order of how these are spread so that they are used more as "defaults"?‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -105,15 +151,38 @@ export const Docs = ({ code = '', description = '', title = '', isLoading, child
...
-            {actions}
-          </Flex>
-          <Text>{description}</Text>
-        </div>
+        <When truthy={!isChildDocs}>
+          <div>
+            <Flex justify="space-between" align="center">
+              <TitleH1
+                className={css({
+                  width: '99%',
```
antonjoel82, line 62: ❓ **question:** Why do we need this?‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -105,15 +151,38 @@ export const Docs = ({ code = '', description = '', title = '', isLoading, child
...
-          <Flex justify="space-between" align="center">
-            <TitleH1>{title}</TitleH1>
-            {actions}
-          </Flex>
-          <Text>{description}</Text>
-        </div>
+        <When truthy={!isChildDocs}>
+          <div>
+            <Flex justify="space-between" align="center">
+              <TitleH1
```
antonjoel82: 🤓 **nitpick (non-blocking):** Can just use Title component with variants now‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -105,15 +151,38 @@ export const Docs = ({ code = '', description = '', title = '', isLoading, child
...
+                {title}
+              </TitleH1>
+              {actions}
+            </Flex>
+            <Text>{description}</Text>
+          </div>
+        </When>
         <Component
           components={{
+            CodeGroup: ({ children, ...props }: { children: ReactNode }) => {
```
antonjoel82:
```suggestion
            CodeGroup: ({ children }: { children: ReactNode }) => {
```

apps/web/src/components/docs/Docs.tsx
```diff
@@ -141,9 +210,13 @@ export const Docs = ({ code = '', description = '', title = '', isLoading, child
...
             },
             Frame: ({ className, ...props }: any) => {
+              const src = Array.isArray(props.children)
+                ? props.children.find((child) => child.type === 'img')?.props.src
+                : props.children.props?.src;
+
               return (
                 <div {...props}>
-                  <img alt="" src={`${MINTLIFY_IMAGE_URL}${props.children.props.src}`} />
+                  <img alt="" src={`${MINTLIFY_IMAGE_URL}${src}`} />
```
antonjoel82: ❓ **question:** What happens here if src is undefined?‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -105,15 +151,38 @@ export const Docs = ({ code = '', description = '', title = '', isLoading, child
...
+            ResponseExample: ({ children }: { children: ReactNode }) => {
+              return <div>{children}</div>;
+            },
+            RequestExample: ({ children }: { children: ReactNode }) => {
+              return <div>{children}</div>;
+            },
+            ParamField: ({ children }: { children: ReactNode }) => {
+              return <div>{children}</div>;
+            },
+            Expandable: ({ children }: { children: ReactNode }) => {
```
antonjoel82: ✏ **suggestion:** Consider a 'DefaultComponent" that we can use for each of these (and others) to centralize the implementation.‏

antonjoel82: ❓ **question:** Also, do we need the wrapping div?‏

apps/web/src/components/layout/components/v2/BridgeUpdateModal.tsx
```diff
@@ -124,7 +123,8 @@ export const BridgeUpdateModal: FC<BridgeUpdateModalProps> = ({ isOpen, toggleOp
                 <LinkText
                   onClick={(e) => {
                     e.preventDefault();
-                    window.open(`${DOCS_URL}/concepts/endpoint`, '_blank');
+                    setPath('concepts/endpoint');
```
antonjoel82: ✏ **suggestion:** We should centralize all paths like this in one file and refer to them by key. This will make it easier to reuse or modify them, and will enable us to use more type-safe params (than a string)‏

apps/web/src/components/docs/Docs.tsx
```diff
@@ -22,29 +24,69 @@ const getMDXComponent = mdxBundler.getMDXComponent;
...
+  actions = null,
+  isChildDocs = false,
+  mappings = {},
+}: DocsProps) => {
   const Component = useMemo(() => {
     if (code.length === 0) {
       return null;
     }
 
+    const globals = Object.keys(mappings).reduce((prev: Record<string, Record<string, any> | any>, key) => {
```
antonjoel82: ✏ **suggestion:** I recommend "Component" to be a separate, named component outside of the Docs scope that accepts props, and this reduce should be a separate helper function that lives outside the component and _could‏_ be tested in isolation

apps/web/src/components/docs/Docs.tsx
```diff
@@ -22,29 +24,69 @@ const getMDXComponent = mdxBundler.getMDXComponent;
...
+  },
+};
+
 type DocsProps = PropsWithChildren<
   DocsQueryResults & {
     isLoading?: boolean;
-    actions: ReactNode;
+    actions?: ReactNode;
+    isChildDocs?: boolean;
+    mappings: Record<string, Record<string, string> | string>;
```
antonjoel82: ☑ **chore:** Please add /** */ comment explaining what mappings represents / how callers should use it‏

apps/web/src/components/docs/ChildDocs.tsx
```diff
@@ -0,0 +1,8 @@
+import { Docs } from './Docs';
+import { useLoadDocs } from './useLoadDocs';
+
+export const ChildDocs = ({ path }: { path: string }) => {
```
antonjoel82, line 8: ☑ **chore:** Please add /** */ comment explaining what "ChildDocs" are / how they are different, and how they are used‏

antonjoel82 (commented):
Thanks for the updates, David! Left some feedback for things to help future maintainers

apps/web/src/components/docs/Docs.tsx
davidsoderberg: Right now we need 'echo' as that file in our docs repo have not been renamed yet

apps/web/src/components/docs/Docs.tsx
davidsoderberg: Because otherwise the floating close icon and external link will be hiding part of the title

apps/web/src/components/docs/Docs.tsx
davidsoderberg: It is there to make sure it ends up as it owns block

apps/web/src/components/docs/Docs.tsx
antonjoel82: Can we please add a TODO in the code then that mentions the legacy name and why it's still there?

apps/web/src/components/docs/Docs.tsx
antonjoel82: Hmm, then there's probably a solution that may address it more directly that would be better to aim for

apps/web/src/components/docs/globals.tsx
antonjoel82, line 1: 👏 **praise:** Great, thank you for separating this!‏

apps/web/src/components/docs/docs.const.ts
```diff
@@ -1,12 +1,20 @@
 import { ROUTES } from '../../constants/routes';
 
+export const CONCEPT_PATHS = {
+  CONTROLS: 'concepts/controls',
+  ENDPOINT: 'concepts/endpoint',
+  WORKFLOWS: 'concepts/workflows',
+  TENANTS: 'concepts/tenants',
+  SUBSCRIBERS: 'concepts/subscribers',
+} as const;
```
antonjoel82: 👏 **praise:** 🔥 ‏

antonjoel82 (approved):
Thanks for the updates!

.cspell.json
```diff
@@ -655,6 +655,7 @@
     "focusable",
     "textareas",
     "frameworkterminal",
+    "echoterminal",
```
rifont: Not needed

```suggestion
```

rifont (approved):
Awesome work! 🚀

rifont:
@davidsoderberg can we get this merged?

PR #5976, opened 2024-07-04
refactor(web): Use context instead of non-singleton Auth hook

antonjoel82:

[Ref](https://novu.slack.com/archives/C06UMR2M5U5/p1720029289379419)

- Fixes large amount of unnecessary identify calls to Segment (34 -> 1 on page refresh!)

### Issues

- [x]  **RESOLVED** In testing, I observed that when changing an Organization, the data on the screen changes, but the value in the Select does not. Furthermore, identify is not called properly. 

### Notes

- This solution could have potential performance concerns if unmemoized data causes the whole constituent component tree to re-render


davidsoderberg (approved):
Looks good, great job ⭐

apps/web/src/ee/translations/components/EditGroupSidebar.tsx
```diff
@@ -115,7 +115,7 @@ export const EditGroupSidebar = ({
       }
     >
       <FormProvider {...methods}>
-        <GroupFormCommonFields edit={true} control={control} readonly={readonly} />
+        <GroupFormCommonFields edit={true} control={control as Control<ICreateGroup>} readonly={readonly} />
```
antonjoel82, line 118: 🗒 **note (non-blocking):** I'm not sure why this appeared now, but my build kept failing locally without this.‏

apps/web/src/ee/translations/components/CreateGroupSidebar.tsx
```diff
@@ -116,7 +116,7 @@ export const CreateGroupSidebar = ({
       }
     >
       <FormProvider {...methods}>
-        <GroupFormCommonFields control={control} readonly={readonly} />
+        <GroupFormCommonFields control={control as Control<ICreateGroup>} readonly={readonly} />
```
antonjoel82, line 119: 🗒 **note (non-blocking):** I'm not sure why this appeared now, but my build kept failing locally without this.‏

SokratisVidros:
@antonjoel82 Let's touch base on this as it is strongly lated to the work done for Clerk. It's definitely in the right direction but we need to make sure we are not doing duplicate work.

SokratisVidros (commented):
@antonjoel82 what's failing in identify? Is it only called once now before the promises resolve?

As soon as we resolve the known issues we can proceed with merging the improvement.

antonjoel82:
@SokratisVidros, thanks to @davidsoderberg's change, we should be good to go!

apps/web/src/auth/auth.const.ts
```diff
@@ -0,0 +1,2 @@
+// TODO: Add a novu prefix to the local storage key
```
denis-kralj-novu, line 1: ❓ **question:** ‏ Is this something that will be addressed in this PR or separately?

apps/web/src/auth/useCreateAuthContext.ts
```diff
@@ -34,16 +33,6 @@ function saveToken(token: string | null) {
...
-export function getToken(): string {
-  return localStorage.getItem(LOCAL_STORAGE_AUTH_TOKEN_KEY) || '';
-}
-
-export function getTokenClaims(): IJwtClaims | null {
-  const token = getToken();
-
-  return token ? jwtDecode<IJwtClaims>(token) : null;
-}
-
```
denis-kralj-novu, line 46: 👏 **praise:** ‏ Nice job moving helpers to more appropriate places <3

denis-kralj-novu (approved):
Good job, hope the N+1 madness stops here

apps/web/src/auth/auth.const.ts
antonjoel82: I moved this comment along with the constant, so would likely aim for a future PR so that the original commenter can do so or explain why this hasn't been done already. I figure it's a matter of needing to modify the token retrieval method to be backwards compatible for the max duration that the `auth_token` could be alive, so it may be:
a change --> wait 1 month --> remove the old `auth_token` retrieval

PR #5978, opened 2024-07-04
fix(web): so buttons are working correctly on get started

davidsoderberg:
Screenshot 2024-07-04 at 09 53 50
Screenshot 2024-07-04 at 09 53 57

### DIFF
--- a/apps/web/src/studio/components/GetStartedPageV2/index.tsx
+++ b/apps/web/src/studio/components/GetStartedPageV2/index.tsx
@@ -19,7 +19,6 @@ import { PageContainer } from '../../layout/PageContainer';
 import { Development } from './Development';
 import { GithubAction } from './GithubAction';
 import { Ide } from './ide';
-import { Studio } from './Studio';
 
 const Link = styled('a', text);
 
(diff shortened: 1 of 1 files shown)


apps/web/src/studio/components/GetStartedPageV2/index.tsx
```diff
@@ -199,7 +203,16 @@ export const GetStartedPageV2 = () => {
...
+                <IconFolderOpen
+                  size={16}
+                  className={css({
+                    color: {
+                      _dark: 'legacy.BGLight !important',
+                      base: 'legacy.B20 !important',
+                    },
+                  })}
+                />{' '}
+                <span>Discover examples</span>
```
ainouzgali, line 215: Is there an hover state for the buttons? like, a color change?

ainouzgali (approved):

apps/web/src/studio/components/GetStartedPageV2/index.tsx
davidsoderberg: No

PR #5988, opened 2024-07-04
feat(root): Fix JS build and introduce playground applications

desiprisg:

I've attempted to fix the build in this PR. I've also worked around our css problem for now with a file rename. The `index.directcss` file shouldn't really change often, if not ever. With a .css suffix, the inline import plugin doesn't seem to work as expected.
For now, I've used the plugin to inject the css in HEAD on our `<Renderer />` .
Note: In the playground app, I'm lazy loading `@novu/js` as the solid code can't run on the server. This is something the `@novu/nextjs` package will be doing when we introduce it, but I'm doing in right now to allow us to use that app to test the package instead of web.
### Screenshots

<details>

</details>


packages/js/package.json
```diff
@@ -33,7 +37,7 @@
   "private": true,
   "scripts": {
     "start": "pnpm run build -- --watch --sourcemap",
-    "build": "tsup && pnpm run build:umd && pnpm run post:build",
```
BiswaViraj, line 36: ❓ **question:** can we revert this change?‏

packages/js/package.json
```diff
@@ -16,12 +16,16 @@
...
+      "require": "./dist/cjs/ui/index.cjs",
+      "import": "./dist/esm/ui/index.mjs",
+      "default": "./dist/esm/ui/index.mjs"
     },
-    "./ui/index.css": "./dist/ui/index.css"
+    "./ui/index.css": {
+      "require": "./dist/cjs/ui/index.css",
+      "import": "./dist/esm/ui/index.css",
+      "default": "./dist/esm/ui/index.css"
+    }
```
BiswaViraj: ☑ **chore:** This can be removed as you're using .directcss now‏

playground/nextjs/.eslintignore
```diff
@@ -0,0 +1 @@
+/
```
BiswaViraj, line 1:
```suggestion

```

❓ **question:** Maybe a typo‏

packages/js/src/ui/novuUI.tsx
```diff
@@ -1,12 +1,13 @@
...
 import { Appearance } from './context';
-import './index.css';
 import { generateRandomString } from './helpers';
 import { Localization } from './context/LocalizationContext';
-import { NovuComponent, NovuComponentControls, NovuComponentName, Renderer } from './components/Renderer';
+import { NovuComponent, NovuComponentName, Renderer } from './components/Renderer';
 import { NovuProviderProps } from './types';
 import { createSignal } from 'solid-js';
+//@ts-expect-error inline import esbuild syntax
+import css from 'directcss:./index.directcss';
```
BiswaViraj, line 10: 🗒 **note (non-blocking):** Not a fan of this hacky approach. But it's fine for now. We should revisit this later for a better solution.
‏

BiswaViraj (approved):
Great work on this

left a couple of comments

playground/nextjs/.eslintignore
desiprisg: No this is just to ignore everything, as the root one didn't work.

packages/js/src/ui/novuUI.tsx
desiprisg: We either will use a CDN for this or find another solution altogether. This is to unblock us for now.

packages/js/package.json
desiprisg: I'm letting tsup generate UMD now. We can revisit this though. Didn't touch the webpack config at all.

PR #6002, opened 2024-07-05
feat(js): Notification list

desiprisg:
### Screenshots

https://github.com/user-attachments/assets/6bd2b39e-5e7e-4f23-8b64-b73408a72934
Edit: Replaces 'Loading...' with 3 notification skeletons

<details>

</details>


apps/web/package.json
```diff
@@ -130,7 +130,8 @@
     "uniqid": "^5.3.0",
     "uuid": "8.3.2",
     "web-vitals": "^1.0.1",
-    "zod": "^3.22.4"
+    "zod": "^3.22.4",
+    "@novu/js": "workspace:*"
```
SokratisVidros: Let's try it in a playground app, not in web until we release.

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -14,6 +15,7 @@ export const Inbox = (props: InboxProps) => {
   const style = useStyle();
   const { id } = useAppearance();
   const { t } = useLocalization();
+  const [feeds] = createResource(fetchFeeds);
```
SokratisVidros: This should come from Novu/js headless. Same for mutations.

packages/js/src/ui/icons/EmptyIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import { useStyle } from '../helpers';
+
+export function EmptyIcon() {
+  const style = useStyle();
+
+  return (
+    <svg
```
SokratisVidros, line 7: Did we pass this from an SVG optimizer?

apps/web/package.json
desiprisg: Didn't mean to commit that

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -10,9 +11,44 @@ type InboxProps = {
...
+  });
+
+  return (
+    <>
+      <Switch>
+        <Match when={currentScreen() === 'inbox'}>
+          <Header updateScreen={setCurrentScreen} />
+          {t('inbox.title')}
+
+          <Show when={!initialLoading()} fallback={<>Loading</>}>
```
LetItRock: ❓ **question:** the loading state is part of the ticket, are you planning to add it or do it in a separate ticket?

packages/js/src/ui/helpers/createInfiniteScroll.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+    });
+    onCleanup(() => io.disconnect());
+    add = (el: Element) => {
+      io.observe(el);
+      onCleanup(() => io.unobserve(el));
+    };
+  }
+
+  const [contents] = createResource(page, fetcher);
```
LetItRock: ✏ **suggestion:** please move this line higher in the code as you are referencing contents in the intersection observer code.

packages/js/src/ui/helpers/createInfiniteScroll.ts
```diff
@@ -0,0 +1,59 @@
...
+    page: Accessor<number>;
+    setPage: Setter<number>;
+    setPages: Setter<T[]>;
+    end: Accessor<boolean>;
+    setEnd: Setter<boolean>;
+  }
+] {
+  const [pages, setPages] = createSignal<T[]>([]);
+  const [initialLoading, setInitialLoading] = createSignal(true);
+  const [page, setPage] = createSignal(0);
```
LetItRock: 🗒 **note (non-blocking):** the code of this hook would have to be updated after [this PR](https://github.com/novuhq/novu/pull/6086) gets merged

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -10,9 +11,44 @@ type InboxProps = {
...
+            <NotificationList notifications={pages()} />
+
+            <Show when={!end()}>
+              <p ref={setEl}>Loading...</p>
```
LetItRock: ☑ **chore:** This should be part of the list I think and rendered at the bottom, because rn it's outside of the list and forces to load straight the second page as it always visible.

We agreed with Nikolay to render the three list item skeletons when loading the next page.

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -10,9 +11,44 @@ type InboxProps = {
...
-export const Inbox = (props: InboxProps) => {
+const InboxContent = () => {
   const [currentScreen, setCurrentScreen] = createSignal<'inbox' | 'settings'>('inbox');
   const { t } = useLocalization();
+
+  const novu = useNovu();
+
+  const [pages, initialLoading, setEl, { end }] = createInfiniteScroll(async (page) => {
+    return (await novu.feeds.fetch({ page })).data;
+  });
```
LetItRock: ☑ **chore:** this hook and some other conditional rendering/loading logic should be a part of the `Notifications` component.

LetItRock (commented):
Something needs to be polished here with the infinite scroll as it fetches for me the two pages even without scrolling.

https://github.com/user-attachments/assets/d852af84-ef1c-446a-b051-0fdd0b0fc0bf

packages/js/src/ui/components/Inbox.tsx
desiprisg: Planning to add it!

packages/js/src/ui/components/Inbox.tsx
desiprisg: yeap

packages/js/src/ui/helpers/createInfiniteScroll.ts
```diff
@@ -0,0 +1,52 @@
...
+  const [initialLoading, setInitialLoading] = createSignal(true);
+  const [offset, setOffset] = createSignal(0);
+  const [end, setEnd] = createSignal(false);
+  const [contents] = createResource(offset, fetcher);
+
+  let setEl: (el: Element) => void = () => {};
+  if (!isServer) {
+    const io = new IntersectionObserver((e) => {
+      if (e.length > 0 && e[0]!.isIntersecting && !end() && !contents.loading) {
+        setOffset((o) => o + e.length);
```
LetItRock: ☑ **chore:** The offset is calculated here in the wrong way, it always increases the value for me by one. It should be something like `setOffset(data().length)` I guess.

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -1,18 +1,39 @@
...
+const InboxContent = () => {
   const [currentScreen, setCurrentScreen] = createSignal<'inbox' | 'settings'>('inbox');
   const { t } = useLocalization();
+
+  return (
+    <>
+      <Switch>
+        <Match when={currentScreen() === 'inbox'}>
+          <Header updateScreen={setCurrentScreen} />
+
```
LetItRock: ✏ **suggestion:** empty lines

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { useLocalization } from '../../context';
+import { useStyle } from '../../helpers';
+import { EmptyIcon } from '../../icons/EmptyIcon';
+import { SkeletonAvatar, SkeletonText } from '../primitives/Skeleton';
+
+const NotificationListContainer = (props: ParentProps) => {
+  const style = useStyle();
+
+  return (
+    <div class={style('notificationListContainer', 'nt-w-[25rem] nt-h-[37.5rem] nt-overflow-auto')}>
```
LetItRock:
```suggestion
    <div class={style('notificationListContainer', 'nt-w-[25rem] nt-h-full nt-overflow-auto')}>
```

`37.5rem` is too much, but because it's in the flex container the height is restricted to available space

LetItRock: and the `width` should not be set here but defined by the list item content, otherwise, if someone creates the custom list item they will be restricted by the width of the container and would have to override it

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+      {props.children}
+    </div>
+  );
+};
+
+const NotificationColumn = (props: ParentProps) => {
+  const style = useStyle();
+
+  return (
+    <div class={style('notificationList', 'nt-flex nt-flex-col nt-min-h-full nt-w-full relative')}>
```
LetItRock: ☑ **chore:** I don't see any reason to have two separate containers `NotificationListContainer` and `NotificationColumn`, please merge them into one.

LetItRock: `relative` not used.

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+    </div>
+  );
+};
+export const EmptyNotificationList = () => {
+  const style = useStyle();
+  const { t } = useLocalization();
+
+  return (
+    <NotificationListContainer>
+      <NotificationColumn data-empty>
```
LetItRock: ❓ **question:** what is this `data-empty`?

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </div>
+      </NotificationColumn>
+    </NotificationListContainer>
+  );
+};
+
+const NotificationSkeleton = () => {
+  return (
+    <>
+      {/* eslint-disable-next-line local-rules/no-class-without-style */}
```
LetItRock: ✏ **suggestion:** if moved to a separate file you can disable this rule at the top of the file

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+    </NotificationListContainer>
+  );
+};
+
+type NotificationListProps = {
+  options?: FetchFeedArgs;
+};
+
+export const NotificationList = (props: NotificationListProps) => {
+  const [data, { initialLoading, setEl, end }] = useFeedInfiniteScroll({ options: props.options });
```
LetItRock, line 44: 🗒 **note (non-blocking):** I noticed that when switching between settings and the notifications we are refetching everything from the ground, maybe we can improve that in a separate PR with storing under the context or smth

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <Show when={!initialLoading()} fallback={<NotificationListSkeleton count={8} />}>
+      <Show when={data().length !== 0} fallback={<EmptyNotificationList />}>
```
LetItRock: I love these solid helper classes <3

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+        </div>
+      </div>
+    </>
+  );
+};
+
+type NotificationListSkeletonProps = {
+  count: number;
+};
+export const NotificationListSkeleton = (props: NotificationListSkeletonProps) => {
```
LetItRock: ✏ **suggestion:** no need in exporting

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
+import { For, ParentProps, Show } from 'solid-js';
```
LetItRock, line 1: ✏ **suggestion:** please split this file to the separate components as it's hard to read

LetItRock (changes requested):
fantastic work George! 🙌 finally we have a notifications list! 🔥 
left you a couple suggestions, let me know if something needs clarification ;)

packages/js/eslint-local-rules.js
```diff
@@ -14,6 +14,15 @@ module.exports = {
       return {
         JSXAttribute(node) {
           if (node.name && node.name.name === 'class') {
+            const parentElement = node.parent;
```
SokratisVidros, line 17: 💯

packages/js/src/ui/api/hooks/useFeed.ts
```diff
@@ -33,3 +34,16 @@ export const useFeed = (props: {
...
   return { feed, fetchFeed, hasMore: hasMore };
 };
+
+type UseFeedInfiniteScrollProps = {
+  options?: Exclude<FetchFeedArgs, 'offset'>;
+};
+export const useFeedInfiniteScroll = (props?: UseFeedInfiniteScrollProps) => {
+  const novu = useNovu();
+
+  const infiniteScroll = createInfiniteScroll(async (offset) => {
```
SokratisVidros:
```suggestion
  return createInfiniteScroll(async (offset) => {
```

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -1,18 +1,39 @@
...
-import { Settings } from './Settings';
 
 type InboxProps = {
   open?: boolean;
   renderBell?: ({ unreadCount }: { unreadCount: number }) => JSX.Element;
 };
 
-export const Inbox = (props: InboxProps) => {
+const InboxContent = () => {
   const [currentScreen, setCurrentScreen] = createSignal<'inbox' | 'settings'>('inbox');
```
SokratisVidros: Please add these in an const object or an string enum to save some bytes instead of repeating the strings.

packages/js/src/ui/components/Notification/NotificationList.tsx
SokratisVidros: I agree. We need to generate as minimal HTML as possible.

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </NotificationColumn>
+    </NotificationListContainer>
+  );
+};
+
+const NotificationSkeleton = () => {
+  return (
+    <>
+      {/* eslint-disable-next-line local-rules/no-class-without-style */}
+      <div class="nt-flex nt-gap-2 nt-p-4">
```
SokratisVidros: Why don't we have appearance in the Skeleton component?

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+  );
+};
+
+type NotificationListSkeletonProps = {
+  count: number;
+};
+export const NotificationListSkeleton = (props: NotificationListSkeletonProps) => {
+  return (
+    <NotificationListContainer>
+      <NotificationColumn data-loading>
```
SokratisVidros: What is this `data-loading`?

packages/js/src/ui/components/Notification/NotificationList.tsx
SokratisVidros: Good call, but as Pawel said, it can be an improvement we do in another PR.

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+type NotificationListProps = {
+  options?: FetchFeedArgs;
+};
+
+export const NotificationList = (props: NotificationListProps) => {
+  const [data, { initialLoading, setEl, end }] = useFeedInfiniteScroll({ options: props.options });
+
+  return (
+    <Show when={!initialLoading()} fallback={<NotificationListSkeleton count={8} />}>
```
SokratisVidros, line 47: Do we have to make the count dynamic based on the current height of the container element? I'd like to avoid generating unnecessary scrolling for the skeleton UI that then goes away when the actual data loads, especially if the data is less than the skeleton.

packages/js/src/ui/components/Notification/NotificationList.tsx
SokratisVidros: 🤓 

```suggestion
      <Show when={data().length > 0} fallback={<EmptyNotificationList />}>
```

packages/js/src/ui/config/defaultLocalization.ts
```diff
@@ -2,4 +2,5 @@ import { Localization } from '../context/LocalizationContext';
 
 export const defaultLocalization: Required<Localization> = {
   'inbox.title': 'Inbox',
+  'notifications.emptyNotice': 'No messages',
```
SokratisVidros:
```suggestion
  'notifications.emptyNotice': 'No notifications',
```

packages/js/src/ui/context/LocalizationContext.tsx
```diff
@@ -1,9 +1,10 @@
 import { ParentProps, createContext, createMemo, useContext } from 'solid-js';
-import { defaultLocalization } from '../config/default-localization';
+import { defaultLocalization } from '../config/defaultLocalization';
 import { Path } from '../helpers/types';
 
 export type Localization = {
   'inbox.title': string;
+  'notifications.emptyNotice': string;
```
SokratisVidros: @desiprisg to avoid adding a type key every time we add a new localization key, let's generate the type from the typeof the Localization object.

packages/js/src/ui/icons/Email.tsx
```diff
@@ -2,7 +2,7 @@ export const Email = () => {
   return (
     <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" viewBox="0 0 20 20">
       <path
-        fill="#6F6E77"
+        fill="currentColor"
```
SokratisVidros, line 5: 💯

SokratisVidros (changes requested):

packages/js/src/ui/components/Notification/NotificationList.tsx
desiprisg: This is to allow targeting this element (which has a descriptor) in it's empty state only.

packages/js/src/ui/helpers/createInfiniteScroll.ts
desiprisg: Ah right 🤦 . We will switch to cursor here though, right? It will have to change then as well.

packages/js/src/ui/components/Notification/NotificationList.tsx
desiprisg: I didn't want to add keys to the containers just yet. There are a lot of them because we are rendering a slightly different skeleton for settings. I want to see if we can make these more similar or implement them in another way before adding that many keys.

packages/js/src/ui/components/Notification/NotificationList.tsx
desiprisg: Allows you to target this element via CSS, only in the loading state, since `nv-notificationList` is the same element in all three states (loading, empty, full).

desiprisg: Seemed reasonable at the time but it's a little premature, might remove

packages/js/src/ui/components/Notification/NotificationList.tsx
desiprisg: Sounds like a good idea, but I'd rather do this improvement in another PR.

packages/js/src/ui/components/Notification/NotificationList.tsx
desiprisg: Have removed this.

packages/js/src/ui/helpers/createInfiniteScroll.ts
LetItRock: yes, we will update it later to the cursor approach

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { useStyle } from '../../helpers';
+import { EmptyIcon } from '../../icons/EmptyIcon';
+import { NotificationListSkeleton, NotificationSkeleton } from './NotificationListSkeleton';
+
+export const NotificationListContainer = (props: ParentProps) => {
+  const style = useStyle();
+
+  return (
+    <div
+      class={style('notificationList', 'nt-flex nt-flex-col nt-min-h-full nt-w-full nt-h-[37.5rem] nt-overflow-auto')}
```
LetItRock, line 14:
```suggestion
      class={style('notificationList', 'nt-flex nt-flex-col nt-flex-1 nt-relative nt-w-full nt-overflow-auto')}
```
The min-height is 600px, which is the total height of the Inbox, we don't need to set it in the flex container, just `flex: 1` will expand it to available space.
The `relative` is needed for the `EmptyNotificationList` because it's positioned absolutely, otherwise, it will be positioned relative to the top popover container.

packages/js/src/ui/helpers/createInfiniteScroll.ts
```diff
@@ -0,0 +1,52 @@
...
+  const [initialLoading, setInitialLoading] = createSignal(true);
+  const [offset, setOffset] = createSignal(0);
+  const [end, setEnd] = createSignal(false);
+  const [contents] = createResource(offset, fetcher);
+
+  let setEl: (el: Element) => void = () => {};
+  if (!isServer) {
+    const io = new IntersectionObserver((e) => {
+      if (e.length > 0 && e[0]!.isIntersecting && !end() && !contents.loading) {
+        setOffset(setOffset(data().length));
```
LetItRock, line 23:
```suggestion
        setOffset(data().length);
```

LetItRock (approved):
awesome work! 🙌

SokratisVidros (approved):

PR #6005, opened 2024-07-07
feat(web): refetch discovery after bridge url update

djabarovgeorge:
### Screenshots

<details>

</details>


apps/api/src/app/environments/usecases/update-environment/update-environment.usecase.ts
djabarovgeorge, line 1: does not really relate to this task, however, we encourage Boy Scout rule so 😃

rifont: Good catch! This is needed for the next migration step to be successful. I will have a PR soon to remove `echo.url` from the Environment entity for good as the next step.

Then, we will reserve the use of `echo` for Bash only 😉

rifont (approved):
🚀🧹

scopsy (approved):

PR #6008, opened 2024-07-08
Com 41 implement inbox header and footer components

BiswaViraj:

- hooks for feeds and read all
- Novu footer
- Icons
- Header Notification status dropdown
- More options dropdown
### Screenshots

image
image
image

<details>

</details>


packages/js/src/ui/components/Header/StatusDropdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+    ],
+  });
+
+  const handleClickOutside = (e: any) => {
+    if (contentRef()?.contains(e.target)) return;
+    setIsOpen(false);
+  };
+
+  onMount(() => {
+    document.body.addEventListener('click', handleClickOutside);
```
desiprisg: ✏️ A good improvement here is to create a custom `useFloating` hook that adds these on and any other functionality we want to have.

packages/js/src/ui/api/feeds.ts
desiprisg, line 1: I don't see any reason in having this file. Why don't we move the logic in the hooks ?

packages/js/src/ui/components/Footer.tsx
```diff
@@ -0,0 +1,10 @@
+import { Novu } from '../icons';
+
+export const Footer = () => {
+  return (
+    <div class="nt-flex nt-justify-center nt-items-center nt-gap-1 nt-mt-auto nt-py-3 nt-text-foreground-alpha-200">
```
desiprisg, line 6: Eslint should complain here for not using `style()`. However, it is the single place we don't want to add descriptors, so let's add an exception to the eslint rule for these two elements.

packages/js/src/ui/components/Header/ActionsContainer.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { cn, useStyle } from 'src/ui/helpers';
+import { Settings } from '../../icons';
+import { MoreActionsDropdown } from './MoreActionsDropdown';
+
+export const ActionsContainer = () => {
+  const style = useStyle();
+  const { id } = useAppearance();
+
+  return (
+    <div class="nt-flex nt-gap-2">
```
desiprisg: ⚠️ Please use `style()` here.

packages/js/src/ui/components/Header/MoreActionsDropdown.tsx
```diff
@@ -0,0 +1,91 @@
...
+        position: position.strategy,
+        top: `${position.y ?? 0}px`,
+        left: `${position.x ?? 0}px`,
+      }}
+      class={style(
+        'moreActionsDropdownContent',
+        cn(id, 'nt-w-max nt-rounded-lg nt-shadow-[0_5px_20px_0_rgba(0,0,0,0.20)] nt-z-10 nt-bg-background nt-py-2')
+      )}
+    >
+      <ul class="nt-list-none">
```
desiprisg: ⚠️ Please use `style()` here.

packages/js/src/ui/components/Header/StatusDropdown.tsx
```diff
@@ -0,0 +1,165 @@
...
+        position: position.strategy,
+        top: `${position.y ?? 0}px`,
+        left: `${position.x ?? 0}px`,
+      }}
+      class={style(
+        'inboxStatusDropdownContent',
+        cn(id, 'nt-w-max nt-rounded-lg nt-shadow-[0_5px_20px_0_rgba(0,0,0,0.20)] nt-z-10 nt-bg-background nt-py-2')
+      )}
+    >
+      <ul class="nt-list-none">
```
desiprisg: ⚠️ Please use style() here.

packages/js/src/ui/components/common/Dropdown.tsx
desiprisg, line 1: Why not have the same format as Popover here? 
`Dropdown`, `Dropdown.Trigger` and `DropdownContent`.

packages/js/src/ui/components/common/DropdownItem.tsx
desiprisg, line 1: Can we place this and `Dropdown.tsx` in a `Dropdown` directory?

packages/js/src/ui/components/common/DropdownItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useAppearance } from '../../context';
+import { AppearanceKey, cn, useStyle } from '../../helpers';
+import { Check } from '../../icons';
+
+export const DropdownItem = (props: {
+  label: string;
+  onClick: () => void;
+  isSelected?: boolean;
+  icon: () => JSX.Element;
+  appearanceKeyPrefix: string;
```
desiprisg: ⚠️ We shouldn't be passing appearance keys around in general. We need to make `DropdownItem` headless.

packages/js/src/ui/context/AppearanceContext.tsx
```diff
@@ -25,6 +25,20 @@ export type Elements = {
...
+  inboxStatusDropdownItem?: ElementStyles;
+  inboxStatusDropdownItemLabel?: ElementStyles;
+  inboxStatusDropdownItemLeftIcon?: ElementStyles;
+  inboxStatusDropdownItemRightIcon?: ElementStyles;
+  inboxStatusTitle?: ElementStyles;
+  moreActionsDropdownTrigger?: ElementStyles;
+  moreActionsDropdownContent?: ElementStyles;
+  moreActionsDropdownItem?: ElementStyles;
+  moreActionsDropdownItemLabel?: ElementStyles;
+  moreActionsDropdownItemLeftIcon?: ElementStyles;
```
desiprisg, line 53: We are going to be adding more appearance keys and I think we need to revise the naming here.
We should have a `dropdownTrigger`, `dropdownContent` etc. appearance keys for targeting all dropdowns
as well as `dropdownTrigger__moreActions` for targeting that specific dropdown. 
Does that make sense? Same for `inboxStatus`. 
In general, the element should come first in the name, like `dropdown`.

packages/js/src/ui/context/FeedsContext.tsx
desiprisg, line 1: Why is the `FeedsContext` needed?

packages/js/src/ui/components/common/DropdownItem.tsx
desiprisg: Feel free to split this up into more components like it's done in radix: https://www.radix-ui.com/primitives/docs/components/dropdown-menu

I'd say `isSelected` can be done with a data attribute instead and just use css, and the children can be the label. Wdyt?
Final API could be
```
<Dropdown>
  <DropdownTrigger>
    //label and icon div
  </DropdownTrigger>
  <DropdownContent>
    <DropdownItem onClick class={/*add class and appearance keys as needed */}>
      //label and icon
    </DropdownItem>
  <DropdownContent/>
</Dropdown>
```
If we need to reuse the same styles multiple times, I'd say just export a class string from this file, and use it when necessary.
I.e.
```
<DropdownItem class={style(['dropdownItem', 'dropdownItem__moreActions'], cn(dropdownItemDefaultStyles, your_extra_classes))} />
```

packages/js/src/ui/context/FeedsContext.tsx
desiprisg, line 1: Let's rename this to `StatusContext` or `InboxStatusContext` to make it more clear. Same for the providers.

packages/js/src/ui/components/Header/ActionsContainer.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { MoreActionsDropdown } from './MoreActionsDropdown';
+
+export const ActionsContainer = () => {
+  const style = useStyle();
+  return (
+    <div class={style('moreActionsContainer', 'nt-flex nt-gap-2')}>
+      <MoreActionsDropdown />
+      <button
+        class={style(
+          'settingsIconContainer',
```
desiprisg: This is also a `button`. I would do `['button', 'settingsButton']` here.

packages/js/src/ui/api/hooks/useFeed.ts
```diff
@@ -0,0 +1,40 @@
...
+        hasMore: response.hasMore,
+      });
+      props.onSuccess?.(response);
+    } catch (error) {
+      console.error('Error fetching feeds:', error);
+      props.onError?.(error);
+    }
+  };
+
+  createEffect(() => {
```
desiprisg, line 34: I think this needs to be a `createResource`, but we'll refactor with `createInfiniteScroll` anyway.

BiswaViraj: Yes, was not sure how to add infinite data using createResource
but yes will be refactored to `createInfiniteScroll`

desiprisg (approved):

PR #6011, opened 2024-07-08
feat: convert sendgrid output schema to zod schema

davidsoderberg:


packages/framework/src/schemas/providers/email/sendgrid.ts
```diff
@@ -1,18 +1,20 @@
...
-  type: 'object',
-  properties: {
-    ipPoolName: { type: 'string' },
-  },
-  required: ['ipPoolName'],
-  additionalProperties: false,
-} as const satisfies Schema;
+export const sendgridOutputSchema = z.object({
+  ipPoolName: z.string(),
+  customData: z
```
davidsoderberg, line 11: I did not find how we map these from the bridge into the provider

rifont: We don't currently. Let's discuss how we can marry them up.

PR #6018, opened 2024-07-09
redactor(cli): move create-novu-app to cli [GRW-1]

scopsy:

- Reduce onboarding time by avoiding installing another CLI via npx. (novu cli is already cached after launching studio)
- Prepare for future DX to allow either recognize an existing package or creating a new one

```bash
npx novu@latest init --secret-key 12345
```
Docs: https://github.com/novuhq/docs/pull/645


.vscode/launch.json
```diff
@@ -229,11 +229,11 @@
   ],
   "compounds": [
     {
-      "name": "-- RUN FULL ENV - Local",
-      "configurations": ["API", "DAL", "EMBED", "SHARED", "TESTING LIB", "WS", "WEB", "WIDGET", "worker"]
+      "name": "-- RUN ENV - Local",
+      "configurations": ["API", "DAL", "SHARED", "TESTING LIB", "WEB", "worker"]
```
scopsy: Remove some legacy services

packages/cli/src/commands/init/index.ts
```diff
@@ -0,0 +1,177 @@
...
+    (validation as any).problems.forEach((problem: string) => console.error(`    ${red(bold('*'))} ${problem}`));
+    process.exit(1);
+  }
+
+  let userId: string;
+  // if no secret key is supplied set to empty string
+  if (!program.secretKey) {
+    program.secretKey = '';
+  } else {
+    try {
```
scopsy, line 101: Small addition to help with telemetry

packages/cli/src/index.ts
```diff
@@ -79,4 +80,16 @@ program
...
+  .description(`Scaffold a new Novu application`)
+  .option(
+    '-s, --secret-key <secret-key>',
+    `Your Novu development environment secret key. Note that your novu app won't work outside of local mode without it.`
+  )
+  .option('-a, --api-url <url>', 'The Novu Cloud API URL', 'https://api.novu.co')
+  .action(async (options: IInitCommandOptions) => {
+    return await init(options, anonymousId);
+  });
+
```
rifont: ✏ **suggestion:** Small casing changes to maintain consistency, and replacing `Scaffold` with `create` as it's aligned with the previous `create-novu-app` naming. Should we also add a `.description` for the command to maintain consistency with the other commands having a description?‏

```suggestion
program
  .command('init')
  .description(`Create a new Novu application`)
  .option(
    '-s, --secret-key <secret-key>',
    `The Novu development environment Secret Key. Note that your Novu app won't work outside of local mode without it.`
  )
  .option('-a, --api-url <url>', 'The Novu Cloud API URL', 'https://api.novu.co')
  .action(async (options: IInitCommandOptions) => {
    return await init(options, anonymousId);
  });

```

.vscode/launch.json
djabarovgeorge: WS is legacy as well?

packages/cli/package.json
```diff
@@ -21,6 +21,7 @@
     "start:dev": "cross-env NODE_ENV=dev NOVU_EMBED_PATH=http://127.0.0.1:4701/embed.umd.min.js NOVU_API_ADDRESS=http://127.0.0.1:3000 NOVU_CLIENT_LOGIN=http://127.0.0.1:4200/auth/login CLI_SEGMENT_WRITE_KEY=GdQ594CEBj4pU6RFldDOjKJwZjxZOsIj nodemon init",
     "start:mode": "cross-env NODE_ENV=dev CLI_SEGMENT_WRITE_KEY=GdQ594CEBj4pU6RFldDOjKJwZjxZOsIj nodemon",
     "start:dev:mode": "cross-env NODE_ENV=dev CLI_SEGMENT_WRITE_KEY=GdQ594CEBj4pU6RFldDOjKJwZjxZOsIj nodemon dev --dashboard-url http://localhost:4200",
+    "start:init": "cross-env NODE_ENV=dev nodemon init",
```
djabarovgeorge: not sure what we have the mode suffix, but should we keep it to be more consistent?

```suggestion
    "start:init:mode": "cross-env NODE_ENV=dev nodemon init",
```

packages/cli/src/commands/init/create-app.ts
```diff
@@ -59,7 +58,7 @@ export async function createApp({
 
   process.chdir(root);
 
-  let hasPackageJson = false;
+  const hasPackageJson = false;
```
djabarovgeorge: 🔨

packages/cli/src/commands/init/index.ts
djabarovgeorge: It is not a big improvement but what do you think about adding it above before "What is your project named?" question, fetch it without waiting and here we can await for the response.
in cases of load, it could reduce a couple of milliseconds 😅

packages/cli/src/commands/init/index.ts
```diff
@@ -0,0 +1,177 @@
...
+
+      const user = await response.json();
+      userId = user.data._id;
+
+      analytics.alias({
+        previousId: anonymousId,
+        userId: user.data._id,
+      });
+    } catch (error) {
+      console.error('Error fetching environment details:', error);
```
djabarovgeorge: Should we throw an exception here as well? 

```suggestion
console.error('The API key you provided could not be found in our system. Please verify your key or contact support.');
```

```suggestion
console.error('The provided API key seems to be invalid. Please check your key and try again.');
```

djabarovgeorge (commented):
Looks like awesome work, i left couple of comments for your review

.vscode/launch.json
scopsy: Good point

packages/cli/src/commands/init/index.ts
scopsy: Yes, was also thinking about it 🤔

PR #6021, opened 2024-07-09
feat(framework): add generic support for providers

davidsoderberg:
Screens:
_Permissive schemas:_
image

_Demo_
https://www.loom.com/share/07a5fe3d078f4429835420810714648e


packages/framework/src/schemas/providers/chat/index.ts
```diff
@@ -1,5 +1,15 @@
...
   slack: slackProviderSchemas,
+  discord: genericProviderSchemas,
+  getstream: genericProviderSchemas,
+  grafanaOnCall: genericProviderSchemas,
+  mattermost: genericProviderSchemas,
+  msTeams: genericProviderSchemas,
+  rocketChat: genericProviderSchemas,
+  ryver: genericProviderSchemas,
+  whatsappBusiness: genericProviderSchemas,
+  zulip: genericProviderSchemas,
```
rifont: Let's import and use the ProviderId enums from here https://github.com/novuhq/novu/blob/next
/libs/shared/src/consts/providers/provider.enum.ts to ensure that we always add the generic schema here and don't regress the provider experience when new providers are added

packages/framework/src/schemas/providers/push/index.ts
```diff
@@ -1 +1,11 @@
...
+
+export const pushProviderSchemas = {
+  fcm: genericProviderSchemas,
+  apns: genericProviderSchemas,
+  expo: genericProviderSchemas,
+  oneSignal: genericProviderSchemas,
+  pushWebhook: genericProviderSchemas,
+  pusherBeams: genericProviderSchemas,
+  pushpad: genericProviderSchemas,
+};
```
rifont:
```suggestion
import { PushProviderIdEnum } from '@novu/shared';
import { Schema } from '../../../types';
import { genericProviderSchemas } from '../generic';

export const pushProviderSchemas: Record<PushProviderIdEnum, { output: Schema }> = {
  [PushProviderIdEnum.FCM]: genericProviderSchemas,
  apns: genericProviderSchemas, //etc
  expo: genericProviderSchemas,
  oneSignal: genericProviderSchemas,
  pushWebhook: genericProviderSchemas,
  pusherBeams: genericProviderSchemas,
  pushpad: genericProviderSchemas,
};
```

packages/providers/src/lib/brevo/brevo.provider.ts
```diff
@@ -10,9 +10,10 @@ import {
   IEmailProvider,
   ISendMessageSuccessResponse,
 } from '@novu/stateless';
+import { EmailProviderIdEnum } from '@novu/shared';
 
 export class BrevoEmailProvider implements IEmailProvider {
-  id = 'sendinblue';
+  id = EmailProviderIdEnum.Sendinblue; // brevo changed name from sendinblue.
```
rifont, line 16: Should we update the Provider filename here in that case?

davidsoderberg: No as Brevo is the new name. We can't change id because of people might already have providers that have sendinblue as id.

scopsy:
@davidsoderberg @rifont is the idea of adding the actual overrides passing to the provider SDK in another PR? Just from the complete user journey perspective to be able to override any native SDK property from the framework SDK.

Or maybe just missed it in this PR diff

davidsoderberg:
> @davidsoderberg @rifont is the idea of adding the actual overrides passing to the provider SDK in another PR? Just from the complete user journey perspective to be able to override any native SDK property from the framework SDK.
> 
> Or maybe just missed it in this PR diff

Yes that is a part of https://linear.app/novu/issue/NV-3974/⚡-add-schemas-for-top-10-provider-schemas-in-novuframework and not NV-3623 as this pr is for :)

scopsy:
Oh! Completly makes sense 💪

packages/providers/src/lib/brevo/brevo.provider.ts
rifont: Got it, thanks!

rifont (approved):
Awesome work! 🧹

PR #6023, opened 2024-07-09
chore(api,worker): Just deploy Novu Cloud to staging

SokratisVidros:

When merging to next branch, there is no need to build a community docker image.

### DIFF
--- a/.github/workflows/dev-deploy-api.yml
+++ b/.github/workflows/dev-deploy-api.yml
@@ -30,7 +30,6 @@ jobs:
     secrets: inherit
 
   deploy_dev_api:
-    if: "!contains(github.event.head_commit.message, 'ci skip')"
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
     needs: test_api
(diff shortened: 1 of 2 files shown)


.github/workflows/dev-deploy-api.yml
```diff
@@ -41,59 +40,51 @@ jobs:
...
       deployments: write
       id-token: write
-    strategy:
-      matrix:
-        name: ['novu/api-ee', 'novu/api']
     steps:
       - uses: actions/checkout@v4
         with:
-          submodules: ${{ contains (matrix.name,'-ee') }}
+          submodules: true
```
scopsy, line 46: @SokratisVidros just to clarify, the goal with this PR is to have a separate pipeline for Community Docker Images?

SokratisVidros: Yes. See [here](https://github.com/novuhq/novu/blob/5edb4dabe8c3522da722dd68f8033dd3f97348cb/.github/workflows/prepare-self-hosted-release.yml#L1). The `next` branch is our trunk, and it should only be deployed to staging ASAP for testing.

rifont (approved):
🚀

SokratisVidros:
For the record, this change should also be applied to the remaining services including ws, webhooks and inbound.

PR #6025, opened 2024-07-09
feat(worker): extract i18n instance from command serialization

djabarovgeorge:
### Screenshots

<details>

</details>


libs/application-generic/src/usecases/compile-email-template/compile-email-template.usecase.ts
```diff
@@ -30,26 +26,12 @@ export class CompileEmailTemplate extends CompileTemplateBase {
 
   public async execute(
     command: CompileEmailTemplateCommand,
-    initiateTranslations?: (
-      environmentId: string,
-      organizationId,
-      locale: string
-    ) => Promise<void>
+    // we need i18nInstance outside the command on order to avoid command serialization on it.
+    i18nInstance?: any
```
djabarovgeorge, line 30: I don't really like the duplicated comments. I thought about extracting to an object type and adding a comment there, but the flattened input feels more user-friendly.

To make it generic with flattened input parameters, one solution could be:

```typescript
export type CompileTemplateContainer<T_Command> = [
  command: T_Command,
  // We need i18nInstance outside the command to avoid command serialization in it.
  i18nInstance?: any
];

public async execute(...[command, i18nInstance]: CompileTemplateContainer<CompileTemplateCommand>)
```

However, this approach seems to add more complexity, and I'm not sure I want to implement it.

apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -1,6 +1,6 @@
-import { Injectable, Logger } from '@nestjs/common';
+import { Injectable, Logger, NotFoundException } from '@nestjs/common';
 import { ModuleRef } from '@nestjs/core';
-import * as inlineCss from 'inline-css';
+import inlineCss from 'inline-css';
```
scopsy, line 3: @rifont, is that related to your recent changes?

scopsy (approved):

PR #6031, opened 2024-07-10
feat(framework): add subscriber data mapping GRW-11

scopsy:

https://github.com/novuhq/packages-enterprise/pull/164
### Screenshots

<details>

</details>

### DIFF
--- a/packages/framework/src/types/subscriber.types.ts
+++ b/packages/framework/src/types/subscriber.types.ts
@@ -5,4 +5,7 @@ export type Subscriber = {
   email?: string;
   phone?: string;
   avatar?: string;
+  data?: Record<string, any>;
+  isOnline?: boolean;
+  lastOnlineAt?: string;
 };


packages/framework/src/types/subscriber.types.ts
```diff
@@ -5,4 +5,7 @@ export type Subscriber = {
   email?: string;
   phone?: string;
   avatar?: string;
+  data?: Record<string, any>;
```
rifont, line 8: Can we strongly type `data`? I don't think we shouldn't have `any` interfaces exposed to the Framework consumer, it weakens the type safeness of their Novu integration. We should probably explore type declaration merging on this `data` object instead.

As a related follow-up, what is the origin of `data` on the subscriber?

rifont: It appears to be this `data` property on the Subscriber's [POST /v1/subscriber request](https://docs.novu.co/api-reference/subscribers/create-subscriber#:~:text=string-,data,-object). Given that such `data` lookups should now happen during the code-first Workflow execution to ensure maximum data freshness, I think we can remove this attribute entirely.

scopsy: I do think it can serve a different purpose with v2, but let's touch base on this over a coffee later. Want to brainstorm a little on this

scopsy: So citcle back on this,
I do still see some usecase where users might want to persist some subscriber entity related data that is part of the notifications domain.

Some examples currently where customer have some complex preference logic and they want to persist it inside of the subscriber and later use to manipulate the workflow.

I think it still very usefull ability, to not force our customers worry about storing this notification related custom data somewhere on their servers

scopsy: Regarding the any, it makes sense. Where would you provide such a global type?

PR #6035, opened 2024-07-10
Com 50 implement the settings component

BiswaViraj:
### Screenshots

https://github.com/novuhq/novu/assets/29251776/486bf1fa-0cb6-44ed-b4fe-a377c0a4769c

https://github.com/novuhq/novu/assets/29251776/276966ab-12bc-43cb-a02c-2ce76c477222

<details>

</details>


packages/js/postcss.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     'postcss-prefix-selector': {
       transform: function (_, selector) {
         // Prefix each class selector with :where(.class)
-        if (selector.startsWith('.')) {
+        if (selector.startsWith('.') && !selector.includes('::after') && !selector.includes('::before')) {
```
BiswaViraj: .where(.class::after) doesnt renders the after element
had to add this to make sure ::after element renders properly

desiprisg: Nice find. We can still override the `after` with `elements` after this, right?

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -1,17 +1,19 @@
...
 import { Popover, popoverContentClasses, popoverTriggerClasses } from './Popover';
+import { Settings } from './Settings';
 
 type InboxProps = {
   open?: boolean;
   renderBell?: ({ unreadCount }: { unreadCount: number }) => JSX.Element;
 };
 
 export const Inbox = (props: InboxProps) => {
+  const [isSettingsOpen, setIsSettingsOpen] = createSignal(false);
```
desiprisg: Can we do `string` instead of `boolean` here. Like `currentScreen` that has values `inbox` or `settings` ?

packages/js/src/ui/components/Settings/LoadingScreen.tsx
```diff
@@ -0,0 +1,38 @@
+import { useStyle } from '../../helpers';
+import { ArrowDropDown } from '../../icons';
+
+export const LoadingScreen = () => {
+  return (
+    <>
+      <LoadingSkeleton />
```
desiprisg: Map ? :P

packages/js/src/ui/icons/LeftArrow.tsx
desiprisg, line 1: Can we rename to `ArrowLeft` ?

packages/js/postcss.config.js
desiprisg: Actually, I don't think this is correct, `.class::after` needs to transform to `:where(.class)::after`, right ?

desiprisg: We just need to make sure we are wrapping just the class with `:where`. Seems feasible.

BiswaViraj: Ah right, `:where(.class)::after` is valid css.
Will update it.

packages/js/src/ui/components/Settings/Settings.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const { preferences, mutate } = usePreferences();
+
+  const allPreferences = createMemo(() => {
+    const globalPreference = preferences()?.find((preference) => preference.level === PreferenceLevel.GLOBAL);
+    const workflowPreferences = preferences()?.filter((preference) => preference.level === PreferenceLevel.TEMPLATE);
+
+    return { globalPreference, workflowPreferences };
+  });
+
+  const optimisticUpdate = ({
```
desiprisg, line 24: This seems like something we should abstract a bit. Any ideas?

desiprisg (approved):

PR #6041, opened 2024-07-11
feat(js): Auto apply generic appearance keys via style()

desiprisg:

* Appearance keys are now stored in an array to enable runtime checks
* Appearance keys are formatted to allow instant knowledge about them being a primitive, or an extension of a primitive for a more specific usecase. Specifically, `extension__primitive` extends `primitive`.
* `style()` automatically applies primitives based on this format. Use the more specific appearance key when creating a new element.
* `style()` no longer receives an array of keys as an argument, in order to force the syntax to be the defining factor.
### Screenshots

<details>

</details>


BiswaViraj (approved):
🚀

PR #6043, opened 2024-07-11
chore(root): update staging URLs

denis-kralj-novu:
- Updated staging URLs
- added resolution algorithm for API url value

### DIFF
--- a/.github/workflows/dev-deploy-widget.yml
+++ b/.github/workflows/dev-deploy-widget.yml
@@ -29,7 +29,7 @@ jobs:
     uses: ./.github/workflows/reusable-widget-deploy.yml
     with:
       environment: Development
-      react_app_api_url: https://dev.api.novu.co
+      react_app_api_url: https://api.novu-staging.co
       react_app_ws_url: https://dev.ws.novu.co
       react_app_webhook_url: https://dev.webhook.novu.co
       react_app_sentry_dsn: https://02189965b1bb4cf8bb4776f417f80b92@o1161119.ingest.sentry.io/625116
--- a/apps/web/src/config/index.ts
+++ b/apps/web/src/config/index.ts
@@ -4,6 +4,15 @@ function isBrowser() {
   return typeof window !== 'undefined';
 }
 
+function autodetectApiRoot() {
+  const origin = window.location.origin;
+  const matcher = new RegExp(/web|dashboard/);
+
+  const isValidTargetForReplace = !origin.includes('localhost') && matcher.test(origin);
+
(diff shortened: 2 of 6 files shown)


apps/web/src/config/index.ts
```diff
@@ -4,6 +4,15 @@ function isBrowser() {
   return typeof window !== 'undefined';
 }
 
+function autodetectApiRoot() {
```
denis-kralj-novu, line 7: 🗒 **note (non-blocking):** ‏ Personally I don't see value in doing this. The config value in env will always override this because of precedence on lines 26 and 27, so the config in the deployment yml needs to be updated as well. 

We could make the config non mandatory, but I am against this as well. I realize that this boilerplate can be viewed as cumbersome, but it is also easy to understand. Not have to dig into the code of the web app in order to see how it concludes what the api endpoint is is a positive in my eyes as well.

packages/cli/src/commands/dev/enums.ts
```diff
@@ -7,5 +7,5 @@ export enum CloudRegionEnum {
 export enum DashboardUrlEnum {
   US = 'https://dashboard.novu.co',
   EU = 'https://eu.dashboard.novu.co',
-  STAGING = 'https://dev.dashboard.novu.co',
+  STAGING = 'https://dashboard.novu-staging.co',
```
denis-kralj-novu, line 10: I think this would need a publish of the cli again yes?

SokratisVidros: Correct.

.github/workflows/dev-deploy-widget.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     uses: ./.github/workflows/reusable-widget-deploy.yml
     with:
       environment: Development
-      react_app_api_url: https://dev.api.novu.co
+      react_app_api_url: https://api.novu-staging.co
```
SokratisVidros, line 32: If we remove this line from dev deployments, we can make both `novu-staging` and `dev.dashboard.novu.co` work in parallel for the next couple of days that we need to test the migration.

apps/web/src/config/index.ts
SokratisVidros: See my previous comment about the env var.

SokratisVidros (approved):

.github/workflows/dev-deploy-widget.yml
denis-kralj-novu: this is in widget, did you mean in the config above for web?

SokratisVidros: 🤦 yes!

PR #6046, opened 2024-07-11
fix(ee-auth): onboarding check

ChmaraX:

Fixed onboarding check to work for both enterprise and community versions. The organization in the enterprise version exists before the questionnaire is filled therefore the original check was not sufficient.

The `currentOrganization?.productUseCases !== undefined` is temporary, because in Clerk we store just `domain` and `productUseCases` while `domain` being optional. 

We will need proper check by introducing new property to Clerk metadata, e.g. `isOnboardingCompeted: boolean`


denis-kralj-novu (approved):
sure, ship it

PR #6048, opened 2024-07-11
fix(ee-auth): update session when org metadata are updated

ChmaraX:

When Clerk organization metadata are updated after questionnaire is submitted, we need to reload session in order to get these new data, otherwise we wouldn't be able to check if the onboarding is complete.


denis-kralj-novu (approved):

SokratisVidros (approved):
👏

PR #6049, opened 2024-07-11
fix(js): Popover cleanups

desiprisg:

* Use portals to render popover content outside the normal flow
* Introduce a `<Root />` component which applies the `novu` and `${id}` classes. Should be used for portals and any element we allow mounting for. I have added it to the `<Renderer />`
* Re-introduce index.css and generate a dummy file on build to use the inline import plugin.
* Focus trapping and dismissing in popovers. Only the outer popover closes and focus is trapped when opening a popover.
### Screenshots

<details>

</details>


packages/js/src/ui/components/Header/Header.tsx
```diff
@@ -1,13 +1,12 @@
...
+import { cn, useStyle } from '../../helpers';
 import { ActionsContainer } from './ActionsContainer';
 import { StatusDropdown } from './StatusDropdown';
 
 export const Header = () => {
   const style = useStyle();
 
   return (
-    <div class={style('inboxHeader', cn('nt-flex nt-justify-between nt-items-center nt-self-stretch nt-py-5 nt-px-6'))}>
+    <div class={style('inboxHeader', cn('nt-flex nt-justify-between nt-items-center nt-w-full'))}>
```
BiswaViraj:
```suggestion
    <div class={style('inboxHeader', cn('nt-flex nt-justify-between nt-items-center nt-w-full  nt-py-5 nt-px-6'))}>
```
Add back the padding

packages/js/src/ui/components/Header/StatusOptions.tsx
```diff
@@ -61,18 +60,19 @@ export const StatusItem = (props: {
...
-        <span class={style('inboxStatus__dropdownItemLabelContainer', dropdownItemLabelContainerClasses())}>
-          <span class={style('inboxStatus__dropdownItemLeftIcon', '')}>{props.icon()}</span>
-          <span class={style('inboxStatus__dropdownItemLabel', dropdownItemLabelClasses())}>{props.label}</span>
+    <Dropdown.Item
+      class={style('inboxStatus__dropdownItem', cn(dropdownItemVariants(), 'nt-flex nt-gap-8'))}
+      onClick={props.onClick}
+    >
+      <span class={style('inboxStatus__dropdownItemLabelContainer', 'nt-flex nt-gap-2 nt-items-center')}>
+        <span class={style('inboxStatus__dropdownItemLeftIcon')}>{props.icon()}</span>
+        <span class={style('inboxStatus__dropdownItemLabel')}>{props.label}</span>
```
BiswaViraj, line 69: ⚠ **issue:** The label renders in white color, add `nt-foreground` class to this‏

BiswaViraj (changes requested):

packages/js/src/ui/components/Header/StatusOptions.tsx
desiprisg: We need to add `nt-foreground` top level, so everything inherits, no ?

BiswaViraj (changes requested):
Nice work on this.

Just a final fix and then we can merge this.

On opening the popover the inbox status dropdown is focussed by default and has focus styles.

https://github.com/user-attachments/assets/7c415966-589d-4348-b502-be72ef844407

desiprisg:
> Nice work on this.
> 
> Just a final fix and then we can merge this.
> 
> On opening the popover the inbox status dropdown is focussed by default and has focus styles.
> 
>  Screen.Recording.2024-07-12.at.6.07.55.PM.mov

Will see how this behaves once I'm done with the button styles in the next PR and make the whole div focusable if it's needed.

BiswaViraj (approved):

PR #6050, opened 2024-07-11
feat(novui,web): Apply CSS Layers

antonjoel82:

https://linear.app/novu/issue/NV-3902/use-css-layer-styles-for-novui

- Use CSS Layers to apply styles from Mantine and Panda
- Remove all now-obsolete `!important` from recipes!
- Keep Panda up-to-date

### Validation

- Given the possibility for visual regressions in all dependencies of `novui`, please conduct a smoke test to help verify these changes 🙏 

_Fixed onboarding bg color_
Screenshot 2024-07-16 at 3 33 37 PM

https://www.loom.com/share/1f249cba936b4ac686b25eb8f836fb1a?sid=0d7bbbaf-3520-46fc-afc6-2bb7f12c5831


apps/web/src/Providers.tsx
```diff
@@ -28,7 +28,7 @@ const queryClient = new QueryClient({
 
 const Providers: React.FC<PropsWithChildren<{}>> = ({ children }) => {
   return (
-    <ThemeProvider>
+    <ThemeProvider shouldDisableGlobals>
```
antonjoel82, line 32: 🗒 note (non-blocking): This is one place that I'm worried about visual regressions, but without it, I'm seeing styles breaking in web. This is because the emotion globals / reset are not wrapped in layers, and thus take precedent over all the novui styles

rifont (commented):
I'm seeing a few regressions of these css properties:
* font weight
* font size
* line height
* color

It will be easiest to compare the before/after images by downloading them and using `command-tilde` (`⌘+~`) to cycle between them in the MacOS Preview App.

Do we need to adjust our base token values to support this shift to use layers?

_**Local Before/After layers**_
_Before_
local-layers-before
_After_
local-layers-after
_Local Comparison video_
https://github.com/user-attachments/assets/26073c6d-79e3-4894-9571-e1a05007071b

_**Cloud Before/After layers**_
_Before_
cloud-layers-before
_After_
cloud-layers-after
_Cloud Comparison video_
https://github.com/user-attachments/assets/671f2690-64c6-4ef2-8759-8ce7d6a1309d

antonjoel82:
> I'm seeing a few regressions of these css properties:
> 
> * font weight
> * font size
> * line height
> * color
> 
> It will be easiest to compare the before/after images by downloading them and using `command-tilde` (`⌘+~`) to cycle between them in the MacOS Preview App.
> 
> Do we need to adjust our base token values to support this shift to use layers?
> 
> _**Local Before/After layers**_ _Before_ local-layers-before _After_ local-layers-after _Local Comparison video_ https://github.com/user-attachments/assets/26073c6d-79e3-4894-9571-e1a05007071b
> 
> _**Cloud Before/After layers**_ _Before_ cloud-layers-before _After_ cloud-layers-after _Cloud Comparison video_ https://github.com/user-attachments/assets/671f2690-64c6-4ef2-8759-8ce7d6a1309d

Awesome, thank you for your diligent testing! My current thinking is to test out Panda's global CSS (that we define) to see if that is an efficient way to set things while still being overridden easily.

rifont:
> > I'm seeing a few regressions of these css properties:
> > 
> > * font weight
> > * font size
> > * line height
> > * color
> > 
> > It will be easiest to compare the before/after images by downloading them and using `command-tilde` (`⌘+~`) to cycle between them in the MacOS Preview App.
> > Do we need to adjust our base token values to support this shift to use layers?
> > _**Local Before/After layers**_ _Before_ local-layers-before _After_ local-layers-after _Local Comparison video_ https://github.com/user-attachments/assets/26073c6d-79e3-4894-9571-e1a05007071b
> > _**Cloud Before/After layers**_ _Before_ cloud-layers-before _After_ cloud-layers-after _Cloud Comparison video_ https://github.com/user-attachments/assets/671f2690-64c6-4ef2-8759-8ce7d6a1309d
> 
> Awesome, thank you for your diligent testing! My current thinking is to test out Panda's global CSS (that we define) to see if that is an efficient way to set things while still being overridden easily.

Exploring [Panda Global CSS](https://panda-css.com/docs/concepts/writing-styles#global-styles) sounds like a reasonable next step - my guess is that the default `reset` @layer styles applied by PandaCSS is different from the user-agent defined defaults, which is what is causing the regression on a number of critical styles. So long as we stick to using semantic token references for the global css to allow inheritance from the design token system, I think we're golden with this direction. Let's try the least touch approach to get us there.

antonjoel82:
> > > I'm seeing a few regressions of these css properties:
> > > 
> > > * font weight
> > > * font size
> > > * line height
> > > * color
> > > 
> > > It will be easiest to compare the before/after images by downloading them and using `command-tilde` (`⌘+~`) to cycle between them in the MacOS Preview App.
> > > Do we need to adjust our base token values to support this shift to use layers?
> > > _**Local Before/After layers**_ _Before_ local-layers-before _After_ local-layers-after _Local Comparison video_ https://github.com/user-attachments/assets/26073c6d-79e3-4894-9571-e1a05007071b
> > > _**Cloud Before/After layers**_ _Before_ cloud-layers-before _After_ cloud-layers-after _Cloud Comparison video_ https://github.com/user-attachments/assets/671f2690-64c6-4ef2-8759-8ce7d6a1309d
> > 
> > 
> > Awesome, thank you for your diligent testing! My current thinking is to test out Panda's global CSS (that we define) to see if that is an efficient way to set things while still being overridden easily.
> 
> Exploring [Panda Global CSS](https://panda-css.com/docs/concepts/writing-styles#global-styles) sounds like a reasonable next step - my guess is that the default `reset` @layer styles applied by PandaCSS is different from the user-agent defined defaults, which is what is causing the regression on a number of critical styles. So long as we stick to using semantic token references for the global css to allow inheritance from the design token system, I think we're golden with this direction. Let's try the least touch approach to get us there.

The issue as far as I could tell is with the `design-system/ThemeProvider` global styles not being wrapped in a CSS layer. As a result, they are applied after _all_ the layers, and thus much of the styling done in `novui` is being overridden.

antonjoel82:
Added Global CSS via Panda today, and everything seemed to be working well, but then started seeing some weird things in Studio. While using browser tools, what I was seeing didn't make much sense, so decided to pull in changes from `next` and do a clean build, but unfortunately am now stuck with build errors.

rifont (approved):
🐼🚀

rifont:
Awesome work @antonjoel82 ! It's looking brilliant!

PR #6052, opened 2024-07-12
Plat 2 migrate novu staging environment to novu stagingcom

denis-kralj-novu:
Added environment variable to ensure CORS doesn't break on old URLs when a new deployment happens


ChmaraX (approved):
LGTM

PR #6054, opened 2024-07-12
fix(ee-auth): retrieve user organizations list from clerk

ChmaraX:
There is one use case where we need list of user organizations in EE => `partner-integrations/vercel/link-projects`

The Clerk organizations resource offers pagination but we can't use it because the rest of the app expects `organizations` to be a full list. 
Since both ee and community auth providers need to return the same `organizations` property, In order to implement pagination we would need to change all the components which are using the `organizations` property to have some sort of pagination mechanism (incl. some design probably) and then also change the community hook to return paginated list.

image


apps/web/src/ee/clerk/providers/useCreateAuthEnterprise.ts
```diff
@@ -38,7 +38,11 @@ export function useCreateAuthEnterprise() {
   const { signOut, orgId } = useAuth();
   const { user: clerkUser, isLoaded: isUserLoaded } = useUser();
   const { organization: clerkOrganization, isLoaded: isOrganizationLoaded } = useOrganization();
-  const { setActive, isLoaded: isOrgListLoaded } = useOrganizationList();
+  const {
+    setActive,
+    isLoaded: isOrgListLoaded,
+    userMemberships,
+  } = useOrganizationList({ userMemberships: { infinite: true } });
```
ChmaraX, line 45: `infinite: true` means that when `fetchNext()` is called, it gets appended to existing `userMemberships.data` property

https://clerk.com/docs/references/react/use-organization-list

apps/web/src/pages/auth/components/LoginForm.tsx
```diff
@@ -79,7 +79,7 @@ export function LoginForm({ email, invitationToken }: LoginFormProps) {
       }
     }
 
-    if (organizations) {
```
ChmaraX, line 82: Not sure why `organizations` (all user orgs) was being checked here - `currentOrganization` is more fitting, since thats basically a subset of `organizations` so if that doesn't exist, currentOrganization also doesn't exist and vice versa

SokratisVidros (approved):

PR #6058, opened 2024-07-14
feat(web): GRW-12 - create a bridge app with sample workflows for new signups

scopsy:
- Syncs new signups to a sample bridge endpoint with 5 workflows
- Shows a top level banner on all workflows page indicating that you are connected to a sample bridge endpoint with CTA to complete onboarding
- minor copy updates


apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -57,13 +57,15 @@ export class Sync {
...
     }
 
-    this.analyticsService.track('Sync Request - [Bridge API]', command.userId, {
-      _organization: command.organizationId,
-      _environment: command.environmentId,
-      workflowsCount: discover.workflows?.length || 0,
-      localEnvironment: !!command.bridgeUrl?.includes('novu.sh') ? true : undefined,
-      source: command.source,
-    });
+    if (command.source !== 'sample-workspace') {
```
scopsy, line 60: Don't fire the event for sample sync

apps/web/src/AppRoutes.tsx
```diff
@@ -119,8 +120,8 @@ export const AppRoutes = () => {
           <Route path="create" element={<CreateTenantPage />} />
           <Route path=":identifier" element={<UpdateTenantPage />} />
         </Route>
-        {isImprovedOnboardingEnabled ? (
-          <Route path={ROUTES.GET_STARTED} element={<GetStartedPage />} />
+        {isV2Enabled ? (
+          <Route path={ROUTES.GET_STARTED} element={<GetStartedPageV2 location="get-started" />} />
```
scopsy, line 124: Restored the Getting Started Page

apps/web/src/components/layout/components/PrivatePageLayout.tsx
```diff
@@ -74,6 +78,9 @@ export function PrivatePageLayout() {
               <AppShell className={css({ '& *': { colorPalette: isLocalEnv ? 'mode.local' : 'mode.cloud' } })}>
                 <Sidebar />
                 <ContentShell>
+                  {location.pathname !== '/get-started' && environment?.bridge?.url === BRIDGE_SYNC_SAMPLE_ENDPOINT && (
```
scopsy: Only visible when the synced url is equal to our sample endpoint

djabarovgeorge:
Found this white like on the top, could it be related to this PR changes?
image

apps/web/src/components/layout/components/v2/SampleWorkflowsBanner.tsx
djabarovgeorge, line 1: Should we convert the styles to use panda styles instead here?

apps/web/src/pages/auth/components/QuestionnaireForm.tsx
```diff
@@ -75,6 +65,18 @@ export function QuestionnaireForm() {
...
     await refetchEnvironments();
 
+    try {
+      await api.post(`/v1/bridge/sync`, {
+        bridgeUrl: BRIDGE_SYNC_SAMPLE_ENDPOINT,
+        source: 'sample-workspace',
+      });
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.error(e);
```
djabarovgeorge, line 75: Do we need this log?

scopsy: Yes, I think it makes sense to log it to console, otherwise will be silent

scopsy:
@djabarovgeorge good catch on the top menu clipping, fixed it.

djabarovgeorge (approved):
Looks good to me, tested it locally workers as expected, please note the comments i left.
In addition i wonder if we should keep the sample workflow in case the user initialized his environment, but then either closed it. and then he or the product would like to play with Novu but would not be able to because they wont have running Bridge.



Everything seems to be working! I've tested it locally and it functions as intended. There are a few comments I've added for your reference.

On a separate note, I have a question. Should we keep the sample workflow in case a user sets up their environment, but then later closes it? This way, if he (or the product team members) want to use Novu again for testing, they wouldn't need a running Bridge because they already have the sample workflow.

PR #6062, opened 2024-07-15
feat(api): add usage of bridge provider options in send message usecases a…

davidsoderberg:
**What?**

When delivering a message to a 3rd party provider in the Worker engine, we need to ensure that the Bridge provider data is given to the Delivery Provider resolver during the delivery flow.

**Why? (Context)**

Bridge provider data should be used when delivery messages via a Delivery Provider.

**Screens**
_Precedence of provider data resolution_
image

_Demonstration of precedence in Provider dashboard (Twilio)_
image


apps/api/src/app/events/usecases/send-test-email/send-test-email.usecase.ts
```diff
@@ -111,6 +112,10 @@ export class SendTestEmail {
 
       html = data.outputs.body;
       subject = data.outputs.subject;
+
+      if (data.providers && typeof data.providers === 'object') {
```
rifont, line 116: What is the type of `data` from `this.previewStep.execute` right now? I think we can improve the typing such that `data.providers` is always of type `object`, and more specifically it can be `Record<string, unknown>`.

If this needs to happen in the Framework typings, let's do it.

apps/api/src/app/events/usecases/send-test-email/send-test-email.usecase.ts
```diff
@@ -111,6 +112,10 @@ export class SendTestEmail {
...
+        bridgeProviderOptions = data.providers[integration.providerId] || {};
```
rifont: I like using the naming of `bridgeProviderData` everywhere (as you had below), it's a bit more explicit.

packages/providers/src/base.provider.ts
```diff
@@ -0,0 +1,106 @@
...
+  query: Record<string, string>;
+};
+
+export abstract class BaseProvider {
+  protected casing: CasingEnum | undefined;
+
+  protected keyCaseObject: Record<string, string> = {};
+
+  protected transform<
+    Output = Record<string, unknown>,
```
rifont: I usually prefer to prefix generic parameters with `T_` to make it very clear that they are generic params, otherwise they can be confused with `TrainCased` class names.

```suggestion
    T_Output = Record<string, unknown>,
```

apps/worker/src/app/workflow/usecases/send-message/send-message-chat.usecase.ts
```diff
@@ -397,14 +397,18 @@ export class SendMessageChat extends SendMessageBase {
...
+
+      const result = await chatHandler.send(
+        {
+          phoneNumber,
+          customData: overrides,
+          webhookUrl: chatWebhookUrl,
+          channel: channelSpecification,
+          ...(bridgeContent?.content ? bridgeContent : { content }),
+        },
+        bridgeProviderData
```
rifont: ❓ **question:** Why are we adding a new positional argument here? Can we add `bridgeProviderData` as an optional property on the first positional argument object instead?

New positional arguments complicate extension, usually required to pass a default object like `{}` to make Typescript whilst adding redundant code. We should prefer to extend objects instead.‏

The same goes for all the other message handlers.

packages/framework/src/schemas/providers/email/mailjet.schema.ts
```diff
@@ -42,61 +42,61 @@ const inlineAttatchment = {
...
-    CustomID: { type: 'string' },
-    EventPayload: { type: 'string' },
-    URLTags: { type: 'string' },
-    Headers: { type: 'object', additionalProperties: true },
-    Variables: { type: 'object', additionalProperties: true },
+    customId: { type: 'string' },
+    eventPayload: { type: 'string' },
+    urlTags: { type: 'string' },
+    headers: { type: 'object', additionalProperties: true },
+    variables: { type: 'object', additionalProperties: true },
```
rifont: 👏 **praise:** Excellent! "The unified Notification framework for developers."‏

packages/providers/src/lib/africas-talking/africas-talking.provider.spec.ts
rifont, line 1: Whilst the value added by these tests is minimal, I think it can be helpful to keep them here as a scaffold for when we improve them. It's also cluttering the PR. Can we add them back?

packages/providers/src/utils/change-case/functions.ts
```diff
@@ -0,0 +1,298 @@
+// Regexps involved with splitting words in various case formats.
+const SPLIT_LOWER_UPPER_RE = /([\p{Ll}\d])(\p{Lu})/gu;
+const SPLIT_UPPER_UPPER_RE = /(\p{Lu})([\p{Lu}][\p{Ll}])/gu;
```
rifont: It looks like the names of these got clipped, let's add back the end for consistency.

```suggestion
const SPLIT_LOWER_UPPER_REGEX = /([\p{Ll}\d])(\p{Lu})/gu;
const SPLIT_UPPER_UPPER_REGEX = /(\p{Lu})([\p{Lu}][\p{Ll}])/gu;
```

packages/providers/src/utils/change-case/functions.ts
```diff
@@ -0,0 +1,298 @@
...
+export interface IOptions {
+  locale?: Locale;
+  split?: (value: string) => string[];
+  /** @deprecated Pass `split: splitSeparateNumbers` instead. */
+  separateNumbers?: boolean;
+  delimiter?: string;
+  prefixCharacters?: string;
+  suffixCharacters?: string;
+  keyCaseTransformer?: (key: string) => string;
+}
```
rifont: ☑ **chore:** This interface should be declared above `IPascalCaseOptions` so that the types read like a book from top to bottom. I had to scroll back up to the top of the file looking for imports before I realised the type was below.

This makes for easier to read and more maintainable code.‏

packages/providers/src/utils/change-case/functions.ts
rifont, line 1: ☑ **chore:** Please add unit tests for all of these utils (change-case & deep-merge)‏

packages/providers/src/utils/change-case/functions.ts
```diff
@@ -0,0 +1,298 @@
...
+    const char = input.charAt(index);
+    if (!suffixCharacters.includes(char)) break;
+    suffixIndex = index;
+  }
+
+  return [
+    input.slice(0, prefixIndex),
+    splitFn(input.slice(prefixIndex, suffixIndex)),
+    input.slice(suffixIndex),
+  ];
```
rifont: ☑ **chore:** Rather than return a tuple which obfuscates the output variable intention, please return an object `{prefix, words, suffix}` which describes the interface fully for better encapsulation of logic (so that the consumer doesn't need to read through the function to understand what the outputs mean)‏

packages/providers/src/utils/change-case/functions.ts
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/**
+ * Options used for converting strings to any case.
+ */
+export interface IOptions {
+  locale?: Locale;
+  split?: (value: string) => string[];
+  /** @deprecated Pass `split: splitSeparateNumbers` instead. */
+  separateNumbers?: boolean;
```
rifont: ☑ **chore:** A copy-paste artifact? I think we can remove this line and implementation usage and go with the deprecation recommendation.‏

packages/providers/src/utils/change-case/functions.ts
```diff
@@ -0,0 +1,298 @@
...
+  delimiter?: string;
+  prefixCharacters?: string;
+  suffixCharacters?: string;
+  keyCaseTransformer?: (key: string) => string;
+}
+
+/**
+ * Split any cased input strings into an array of words.
+ */
+export function split(value: string) {
```
rifont: ☑ **chore:** Please explicitly declare return typings for all functions to avoid implicit typings which can expose the function to future return type breakage. Explicit typings also ensure future maintainers know what the function should return without needing to read through the function, providing better encapsulation.‏
```suggestion
export function split(value: string): string {
```

packages/providers/src/utils/types.ts
```diff
@@ -0,0 +1,14 @@
...
+};
+
+export type WithPassthrough<T> = Prettify<T & { _passthrough?: Passthrough }>;
+
+/**
+ * A type that represents a type that is a prettified version of the original type.
+ * The prettified type has all generics removed from intellisense and displays a flat object.
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type Prettify<T> = { [K in keyof T]: T[K] } & {};
```
rifont: ☑ **chore:** `Prettify` isn't necessary here, it's only useful for external/public SDK consumers to beautify the output. As this type is used internally by Providers, we don't need it.‏
```suggestion
export type WithPassthrough<T> = T & { _passthrough?: Passthrough };
```

packages/providers/src/base.provider.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  protected keyCaseObject: Record<string, string> = {};
+
+  protected transform<
+    T_Output = Record<string, unknown>,
+    T_Input = Record<string, unknown>,
+    T_Data = Record<string, unknown>
+  >(
+    bridgeProviderData: WithPassthrough<T_Input>,
+    providerdata: T_Data
```
rifont: I'd suggest `triggerProviderData` @davidsoderberg, wdyt?

packages/providers/src/base.provider.ts
```diff
@@ -0,0 +1,95 @@
...
+    T_Data = Record<string, unknown>
+  >(
+    bridgeProviderData: WithPassthrough<T_Input>,
+    triggerProviderData: T_Data
+  ): TransformOutput<T_Output> {
+    const {
+      _passthrough = {
+        body: {},
+        headers: {},
+        query: {},
```
rifont: I think we need to add a fallback value of empty object for each of these properties. Without this fallback, if I specify a `_passthrough = { body: { key: 'value' } }`, the `headers` and `query` will not be populated with a default value and will be `undefined`.

rifont: Should we just do a deepmerge of the fallback `_passthrough` and the provided `_passthrough` to handle these fallbacks?

davidsoderberg: A yes I missed it for body lower down

packages/providers/src/utils/change-case/change-case.spec.ts
```diff
@@ -0,0 +1,559 @@
...
+  dotCase,
+  kebabCase,
+  noCase,
+  pascalCase,
+  pathCase,
+  sentenceCase,
+  snakeCase,
+  trainCase,
+} from './index';
+const studPrimitiveObject = {
```
rifont: I think this is a typo?

```suggestion
const stubPrimitiveObject = {
```

packages/providers/src/utils/change-case/change-case.spec.ts
```diff
@@ -0,0 +1,559 @@
...
+
+const stud = {
+  ...studPrimitiveObject,
+  listOfObjects: [studPrimitiveObject, studPrimitiveObject],
+  nestedObject: {
+    ...studPrimitiveObject,
+    listOfObjects: [studPrimitiveObject, studPrimitiveObject],
+  },
+  listOfListOfObjects: [[studPrimitiveObject], [studPrimitiveObject]],
+};
```
rifont: Should it be `stub` instead of `stud`?

apps/worker/src/app/workflow/usecases/send-message/send-message-sms.usecase.ts
rifont, line 1: Commenting here as there is no file to comment on. We have handled Provider data injection for Chat, Email, Push, and SMS. We are missing In-App. We should handle it in this PR even though there is no Provider using this data currently, otherwise we risk regressing in the future.

packages/providers/src/utils/deepmerge.utils.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+    const result = deepmergeAll([obj1, obj2]);
+
+    expect(result).toEqual({
+      a: {
+        b: 1,
+        c: 2,
+        d: {
+          a: 1,
+        },
+        x: [1, 2, 3, 3, 4, 5],
```
rifont: 👏 **praise:** ‏Great test! Very easy to read.

packages/providers/src/base.provider.ts
```diff
@@ -0,0 +1,94 @@
...
+    triggerProviderData: T_Data
+  ): TransformOutput<T_Output> {
+    const { _passthrough = {}, ...bridgeData } = bridgeProviderData;
+    const result = this.casingTransform(bridgeData);
+
+    const fallbackPassthrough = { body: {}, headers: {}, query: {} };
+    const mergedPassthrough = deepmerge<Passthrough>(
+      fallbackPassthrough,
+      _passthrough
+    );
```
rifont: On second thought, I think this should be renamed to `default` -

```suggestion
    const defaultPassthrough = { body: {}, headers: {}, query: {} };
    const mergedPassthrough = deepmerge<Passthrough>(
      defaultPassthrough,
      _passthrough
    );
```

rifont (approved):
Amazing work!

PR #6064, opened 2024-07-15
fix(js): Fix build types

desiprisg:
### Screenshots

<details>

</details>


packages/js/package.json
```diff
@@ -10,30 +10,31 @@
...
     "dist/cjs",
     "dist/esm",
-    "dist/types",
-    "dist/ui"
+    "dist/types"
   ],
   "sideEffects": false,
   "private": true,
   "scripts": {
+    "clean": "rimraf ./dist",
```
BiswaViraj, line 33: You can pass `clean: true` to the tsup config and it will delete the dist folder before building.
[Ref](https://tsup.egoist.dev/#typescript--javascript:~:text=%3A%20true%2C-,clean%3A%20true%2C,-%7D)

playground/nextjs/src/pages/index.tsx
```diff
@@ -1,21 +1,25 @@
...
 
 export default function Home() {
   return (
     <div className="h-screen w-full bg-neutral-100">
       <Mounter
         mount={(el) => {
-          const NovuUI = require('@novu/js/ui').NovuUI;
-          const ui = new NovuUI({
+          //require here as it won't work in SSR.
+          const { NovuUI } = require('@novu/js/ui');
```
BiswaViraj, line 10: Duplicate import, NovuUI is imported at the top

desiprisg: I am just importing the type so I can cast it! We can't import in SSR so it will error if we do it on top. This will be handled by the react/nextjs package once we make those.

packages/js/package.json
desiprisg: I think it doesn't clean the whole folder, but just what it generates.

BiswaViraj (approved):

PR #6066, opened 2024-07-15
fix(web): ensure onboarding flow for cli

scopsy:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/web/src/pages/auth/onboarding/Onboarding.tsx
+++ b/apps/web/src/pages/auth/onboarding/Onboarding.tsx
@@ -3,6 +3,7 @@ import { css } from '@novu/novui/css';
 import { IconClose, IconCloseFullscreen } from '@novu/novui/icons';
 import { HStack, LinkOverlay } from '@novu/novui/jsx';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


apps/web/src/pages/auth/onboarding/Onboarding.tsx
```diff
@@ -13,40 +14,42 @@ export function OnboardingPage() {
   const { colorScheme } = useColorScheme();
 
   return (
-    <div
-      className={css({
-        bg: 'surface.page',
-      })}
-    >
-      <HStack
+    <EnsureOnboardingComplete>
```
scopsy, line 17: Diff is a bit strange, but that's the only change

SokratisVidros: [This is the reason](https://github.com/novuhq/novu/blob/3ea3c7699a7c00cbc0ace22f0cde1d49c0a4c0ec/apps/web/src/AppRoutes.tsx#L90). So far all post-auth screens were handled by the PrivatePageLayout. The naming if that guard is off, but we will fix it later.

SokratisVidros (approved):

PR #6068, opened 2024-07-15
fix(web): translation e2e tests

ChmaraX:
`currentOrganization` object is not being properly refreshed when running tests PW when `useQuery` is used - replaced with specific `reloadOrganization` to match EE implementation.

### DIFF
--- a/apps/web/src/ee/translations/components/DefaultLocaleModal/index.tsx
+++ b/apps/web/src/ee/translations/components/DefaultLocaleModal/index.tsx
@@ -38,8 +38,8 @@ export const DefaultLocaleModal = ({
   const { mutateAsync: saveDefaultLocale, isLoading: isSaving } = useMutation<any, any, any>(
     (args) => api.patch('/v1/translations/language', args),
     {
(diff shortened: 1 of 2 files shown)


SokratisVidros (approved):

apps/web/src/ee/translations/components/DefaultLocaleModal/index.tsx
```diff
@@ -33,13 +33,13 @@ export const DefaultLocaleModal = ({
...
-  const { currentOrganization } = useAuth();
+  const { currentOrganization, reloadOrganization } = useAuth();
 
   const { mutateAsync: saveDefaultLocale, isLoading: isSaving } = useMutation<any, any, any>(
     (args) => api.patch('/v1/translations/language', args),
     {
-      onSuccess: () => {
-        queryClient.refetchQueries([`/v1/organizations`]);
+      onSuccess: async () => {
+        await reloadOrganization();
```
SokratisVidros, line 42: 💯

SokratisVidros (approved):

PR #6069, opened 2024-07-15
fix(api): replace axios instance with test agent in events e2e

ChmaraX:
In some tests we are creating a new `axiosInstance` and in some tests we are using `session.testAgent` - this leads to various issues and weird behaviour - we should probably use `session.testAgent` everywhere since thats being initialized in the testing session.


SokratisVidros (approved):
💯

PR #6072, opened 2024-07-15
fix(api): get organizations e2e

ChmaraX:
Remove redundant steps from the `get-organizations` e2e test


denis-kralj-novu (approved):
🚀

PR #6073, opened 2024-07-15
feat(worker): Disable inline-css behind feature flag

antonjoel82:

https://linear.app/novu/issue/NV-4117/email-html-content-issue-in-mobile-devices

- `inline-css` call is only used in `trigger` use cases (not send-test-email), so this logic was causing divergence in how we send emails
- Calling `inlineCss` caused the styles to omit media queries that resulted in incorrect styling in mobile clients
- _Flag will be disabled in prod until post-deploy, at which time we will carefully choose to enable and monitor closely_

Follow-up ticket: https://linear.app/novu/issue/NV-4120/delete-is-email-inline-css-disabled-feature-flag-and-remove-from-code

### Validation

- Ran full app (worker, web, API) locally and verified feature flag working, and resulting in desired email rendering

### Screenshots

Original:
IMG_1816 (1)

Corrected: 
B743C55B-86C5-4C6E-A19E-8EEDA263AE2D


rifont (approved):
💯

PR #6075, opened 2024-07-15
feat(framework): Add new Inbox properties to `step.inApp` schema

rifont:

* New `step.inApp` properties are needed to support the inbox component.

### Screenshots

_Everything specified_
```typescript
export const inAppWorkflow = workflow('in-app-workflow', async ({ step }) => {
  await step.inApp('send-in-app', (controls) => ({
    subject: `Welcome!`,
    body: 'Hello there',
    avatar: 'https://static.vecteezy.com/system/resources/thumbnails/006/487/917/small_2x/man-avatar-icon-free-vector.jpg',
    primaryAction: {
      label: 'Primary',
      url: 'https://google.com',
    },
    secondaryAction: {
      label: 'Secondary',
      url: 'https://google.com',
    }
  }));
```

_Studio Preview_
image

_Dashboard Preview_
image

_V1 In-App Delivered Notification_
image

_V2 Inbox Delivered Notification_
image

<details>

</details>


BiswaViraj (approved):

BiswaViraj:
>  /* 👇is this property meant to be specifiable in the Framework? There's some duplication with `avatar` */

Not sure, if this is clear now, but still want to address here for visibility.
In the example, it does seem like there's some duplication in `avatar` and `actor` as we are using `user` type
But in the scenario, where `avatar` is a `system icon` or `custom`
The `actor` will be `undefined or null`.
@rifont

rifont:
> > /* 👇is this property meant to be specifiable in the Framework? There's some duplication with `avatar` */
> 
> Not sure, if this is clear now, but still want to address here for visibility. In the example, it does seem like there's some duplication in `avatar` and `actor` as we are using `user` type But in the scenario, where `avatar` is a `system icon` or `custom` The `actor` will be `undefined or null`. @rifont

Thanks for the update @BiswaViraj . In that case, given that the presence of `actor` in dependent on the presence of `avatar`, does it make more sense to more closely co-locate the attributes? I am thinking we can merge the `avatar` and `actor` property, so we have:

_`actor.type==='user'`_
```typescript
const inAppRes = await step.inApp('send-in-app', async () => ({
  body: 'Test Body',
  actor: {
    type: 'user',
    subscriberId: 'joe@acme.com', // required when type is user
    firstName: 'Joe', // optional when type is user, overrides default subscriber first name
    lastName: 'Smith', // optional when type is user, overrides default subscriber last name
    avatar: 'https://novu.co/image', // optional when type is user, overrides the default subscriber avatar
  },
}));
```

_`actor.type==='custom'`_
```typescript
const inAppRes = await step.inApp('send-in-app', async () => ({
  body: 'Test Body',
  actor: {
    type: 'custom',
    avatar: 'https://novu.co/image', // type: URL - required when type is `custom`
  },
}));
```

_`actor.type==='system'`_
```ts
const inAppRes = await step.inApp('send-in-app', async () => ({
  body: 'Test Body',
  actor: {
    type: 'system',
    avatar: 'robot', // type: IconName - required when type is system
  },
}));
```

Wdyt?

packages/framework/src/schemas/steps/channels/in-app.schema.ts
```diff
@@ -1,9 +1,69 @@
...
+  },
+  required: ['label'],
+  additionalProperties: false,
+} as const satisfies Schema;
+
+const actorSchema = {
+  type: 'object',
+  properties: {
+    subscriberId: { type: 'string' },
+    firstName: { type: 'string' },
```
SokratisVidros: Actor won't always be a human. Would it make sense to concat first and last name into a displayName property and move control to the implementor to populate it correctly?

SokratisVidros (approved):

packages/framework/src/schemas/steps/channels/in-app.schema.ts
rifont: I think such concatenation is best left to the presentation layer (i.e. to Inbox), folks might like to do things per custom inbox implementation. For example, take a subscriber with subset of `{ firstName: 'John', lastName: 'Doe' }`, presentation layer implementation might look like `displayName: ${firstName} ${lastName.chatAt(0)}.`-> `displayName: 'John D.'`.

I'd suggest we could introduce a `displayName` for `actor.type==='system'` to handle naming for a non-Human actor. That would come at a cost of a less unified API where both `displayName` and `firstName+lastName` must be handled in the Inbox for each `actor.type`

rifont: There's also an argument that Framework can function as the presentation layer as Sok mentioned, such that `displayName` is constructed on the Framework and we have a unified `displayName` interface on the `actor`.

BiswaViraj (approved):

packages/framework/src/schemas/steps/channels/in-app.schema.ts
rifont: We will defer `displayName` for now and revisit in a future iteraiton.

rifont:
Converted to draft as we still need to wire-up the Worker implementation to support these new attributes in Framework.

apps/web/src/components/workflow/preview/in-app/Content.styles.ts
```diff
@@ -69,6 +69,14 @@ export const NotificationTextStyled = styled.div<{ isExampleNotification: boolea
   }}
 `;
 
+export const SubjectTextStyled = styled.p`
+  font-size: 16px;
+  font-style: normal;
+  font-weight: 700;
+  line-height: 20px;
+  margin: 0;
```
antonjoel82:
```suggestion
  font-size: 16px;
  font-style: normal;
  font-weight: 600;
  line-height: 20px;
  margin: 0;
```

This doesn't conform to the Design System value

apps/web/src/components/workflow/preview/in-app/InAppPreview.tsx
```diff
@@ -39,8 +50,29 @@ export function InAppPreview({ showVariables = true }: { showVariables?: boolean
...
+            ...(data.outputs.secondaryAction
+              ? [
+                  {
+                    type: ButtonTypeEnum.SECONDARY,
+                    content: data.outputs.secondaryAction.label,
+                    url: data.outputs.secondaryAction.url,
+                  },
+                ]
+              : []),
+          ],
```
antonjoel82: ✏ **suggestion:** This mutation seems worthwhile to extract to a hook with refined types and logic‏

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -180,7 +182,35 @@ export class SendMessageInApp extends SendMessageBase {
...
+          ...(bridgeOutputs.secondaryAction
+            ? [
+                {
+                  type: ButtonTypeEnum.SECONDARY,
+                  content: bridgeOutputs.secondaryAction.label,
+                  url: bridgeOutputs.secondaryAction.url,
+                },
+              ]
+            : []),
+        ],
```
antonjoel82: 💭 **thought (non-blocking):** This verbose logic is duplicated 3 times, so likely a good candidate for DRY extraction‏

apps/web/src/components/workflow/preview/in-app/Content.styles.ts
rifont: Good catch, thanks!

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
rifont: I agree. I've extracted to a shared util.

PR #6082, opened 2024-07-16
feat(framework): add sanitize html to step output

davidsoderberg:
**What?**
* Sanitize the step output for all channel steps to eliminate XSS attacks via Studio and downstream content renderers such as Inbox.

**Screenshots**
_Step result is sanitized for email_
image


packages/framework/src/client.ts
```diff
@@ -291,7 +292,15 @@ export class Client {
         setResult(stepResult);
       }
 
-      return stepResult.outputs as any;
+      if (
+        (
+          [ChannelStepEnum.CHAT, ChannelStepEnum.EMAIL, ChannelStepEnum.SMS, ChannelStepEnum.IN_APP] as string[]
```
rifont: Need `PUSH` too

packages/framework/src/client.ts
```diff
@@ -35,6 +35,7 @@ import type {
 } from './types';
 import { EMOJI, log } from './utils';
 import { transformSchema, validateData } from './validators';
+import { sanitizeHTML, sanitizeHtmlInObject } from './utils/sanitiez-html.utils';
```
rifont: `sanitizeHTML` appears to be unused

packages/framework/src/utils/sanitiez-html.utils.ts
rifont, line 1: I think we need to add some tests for:
1. unit test - verify that nested objects are sanitized
2. Integration test - verify that channel steps are sanitized, whilst other steps like digest/delay/custom are not sanitized

packages/framework/package.json
```diff
@@ -132,25 +132,26 @@
...
+    "@apidevtools/json-schema-ref-parser": "11.6.4",
     "@sveltejs/kit": "^1.27.3",
+    "@types/aws-lambda": "^8.10.141",
     "@types/express": "^4.17.13",
     "@types/node": "^20.14.10",
-    "@types/aws-lambda": "^8.10.141",
     "@types/sanitize-html": "2.11.0",
     "@vercel/node": "^2.15.9",
+    "aws-lambda": "^1.0.7",
+    "eslint-plugin-unused-imports": "^4.0.0",
```
rifont: What if we moved this to the top-level package.json, disabled the rule by default, then just enabled the rule for this package (given that we extend the top-level eslint config). That way other packages can slowly opt-in and we keep a consistent eslint plugin version through the codebase.

packages/framework/src/client.ts
```diff
@@ -35,6 +35,7 @@ import type {
 } from './types';
 import { EMOJI, log } from './utils';
 import { transformSchema, validateData } from './validators';
+import { sanitizeHtmlInObject } from './utils/sanitiez-html.utils';
```
rifont: typo:
```suggestion
import { sanitizeHtmlInObject } from './utils/sanitize-html.utils';
```

packages/framework/src/client.ts
```diff
@@ -291,7 +292,21 @@ export class Client {
         setResult(stepResult);
       }
 
-      return stepResult.outputs as any;
+      if (
```
SokratisVidros: Instead of checking on specific channels, should we check if `stepResult.outputs` is typeof `string`?

SokratisVidros (approved):

packages/framework/src/client.ts
```diff
@@ -291,7 +292,21 @@ export class Client {
...
+        (
+          [
+            ChannelStepEnum.CHAT,
+            ChannelStepEnum.EMAIL,
+            ChannelStepEnum.SMS,
+            ChannelStepEnum.IN_APP,
+            ChannelStepEnum.PUSH,
+          ] as string[]
+        ).includes(step.type)
```
rifont:
```suggestion
        Object.values(ChannelStepEnum).includes(step.type as ChannelStepEnum)
```

packages/framework/src/client.ts
davidsoderberg: I think I like the check to be a bit more verbose so any future of us will easier understand what the check is for :)

PR #6083, opened 2024-07-16
chore(web): ensure onboarding for v2 works

denis-kralj-novu:
- remove force redirect on signup, overrides any other redirect
- add redirect allowed origins to provider
- wait for user data to show up when token is present

### DIFF
--- a/apps/web/src/ee/clerk/providers/ClerkProvider.tsx
+++ b/apps/web/src/ee/clerk/providers/ClerkProvider.tsx
@@ -27,6 +27,8 @@ const localization = {
   },
 };
 
+const ALLOWED_REDIRECT_ORIGINS = ['http://localhost:2022', new URL(location.href).origin];
+
 export const ClerkProvider: React.FC<PropsWithChildren<{}>> = ({ children }) => {
   const navigate = useNavigate();
   const { colorScheme } = useColorScheme();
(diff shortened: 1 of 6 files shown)


denis-kralj-novu:
Cannot test the new user registration flow while **[this error](https://novu.slack.com/archives/C06DU7A7BPV/p1721131701635689)** still happens

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
```diff
@@ -27,6 +27,8 @@ const localization = {
   },
 };
 
+const ALLOWED_REDIRECT_ORIGINS = ['http://localhost:2022', new URL(location.href).origin];
```
SokratisVidros: Can you please try this and see if it works?

```suggestion
const ALLOWED_REDIRECT_ORIGINS = ['http://localhost:*'];
```

SokratisVidros (changes requested):

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
denis-kralj-novu: great suggestion, applied

SokratisVidros (approved):

SokratisVidros (approved):

PR #6085, opened 2024-07-16
fix(web): get new envs directly instead of via usequery update

ChmaraX:
If we call `refetchEnvironments` from `useQuery` we can't guarantee the env-id will be in localStorage since its being set as side effect.

If we get the env and also set the localStorage env id inside single function, we are able to await both.

Ideal case would be if `useQuery` could both `getEnvironments()` and  `selectEnvironment()` inside single `refetchEnvironments()` call, but that doesn't seem to work even when trying to call `selectEnvironment` inside useQuery's `onSuccess`.

### DIFF
--- a/apps/web/src/components/providers/EnvironmentProvider.tsx
+++ b/apps/web/src/components/providers/EnvironmentProvider.tsx
@@ -1,12 +1,6 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
-import {
-  useQuery,
-  useQueryClient,
-  type QueryObserverResult,
-  type RefetchOptions,
-  type RefetchQueryFilters,
-} from '@tanstack/react-query';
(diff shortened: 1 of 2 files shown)


apps/web/src/components/providers/EnvironmentProvider.tsx
```diff
@@ -143,25 +135,38 @@ export function EnvironmentProvider({ children }: { children: React.ReactNode })
...
-    (async () => {
-      if (environments) {
-        await switchEnvironment({ environmentId: getEnvironmentId() });
-      }
-    })();
-  }, [environments, switchEnvironment]);
+    if (environments && environments.length > 0 && !currentEnvironment) {
+      switchEnvironment({ environmentId: getEnvironmentId() });
+    }
+  }, [currentEnvironment, environments, switchEnvironment]);
```
ChmaraX, line 141: This change is to avoid calling switchEnvironment on every path change - now its being called just once when `currentEnvironment` doesn't exist which is only upon registration - in other cases we are switching env directly using switchEnvironment functions.

SokratisVidros (approved):

PR #6090, opened 2024-07-17
chore(client): add the novu api version header

LetItRock:
Changes:
- added the Novu API Version header to the `HttpClient` from `@novu/client` package
- removed unused file

### Screenshots

<details>

</details>


packages/client/src/http-client/http-client.ts
```diff
@@ -4,8 +4,9 @@ import type { CustomDataType } from '@novu/shared';
 const DEFAULT_API_VERSION = 'v1';
 const DEFAULT_BACKEND_URL = 'https://api.novu.co';
 const PACKAGE_NAME = '@novu/client';
-const PACKAGE_VERSION = '0.42.0';
+const PACKAGE_VERSION = '2.0.0-canary.0';
 const DEFAULT_USER_AGENT = `${PACKAGE_NAME}-${PACKAGE_VERSION}`;
+const NOVU_API_VERSION = '2024-06-26';
```
BiswaViraj: ❓ **question:** ‏How will we track this date? is it related to the date of new `@novu/api` release?

packages/client/src/http-client/http-client.ts
```diff
@@ -22,6 +23,7 @@ export class HttpClient {
     this.headers = {
       'Content-Type': 'application/json',
       'User-Agent': userAgent,
+      'Novu-API-Version': NOVU_API_VERSION,
```
BiswaViraj: 💭 **thought (non-blocking):** I think we don't need to add this, if someone passes a custom backend url‏

packages/client/src/http-client/http-client.ts
LetItRock: @BiswaViraj For now it's hardcoded and we just want to track post v2 requests, it's the date that Sokratis suggested

packages/client/src/http-client/http-client.ts
LetItRock: We would end up writing the comparing logic to check if it's a US or EU instance URL. I think we can do that...

packages/client/src/http-client/http-client.ts
SokratisVidros: Since we need the date mostly in Novu/js, should we allow the option on the client level and pass the date from the client instance used in `@novu/js`?

SokratisVidros (changes requested):

packages/client/src/http-client/http-client.ts
LetItRock: If we need that date only in the `@novu/js` I would say to update the headers just in that package and not here.

packages/client/src/http-client/http-client.ts
SokratisVidros: Correct. No need to hardcode on the client level

SokratisVidros (approved):

PR #6097, opened 2024-07-17
feat(web,novui): initial implementation of var autocomplete in controls 

ainouzgali:
### Screenshots
Screenshot 2024-07-22 at 18 38 34
Screenshot 2024-07-22 at 18 39 20

<details>

</details>


libs/novui/package.json
```diff
@@ -129,13 +129,22 @@
     }
   },
   "dependencies": {
-    "@mantine/code-highlight": "^7.10.2",
-    "@mantine/core": "^7.10.0",
-    "@mantine/hooks": "^7.10.0",
+    "@mantine/code-highlight": "^7.11.2",
+    "@mantine/core": "^7.11.2",
+    "@mantine/hooks": "^7.11.2",
+    "@mantine/tiptap": "^7.11.2",
```
antonjoel82, line 135: @ainouzgali FYI I updated all the Mantine packages to have the same version since it can cause issues if they get mismatched!

libs/novui/src/tokens/semanticColors.tokens.ts
```diff
@@ -203,6 +203,16 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
...
+  suggestion: {
+    surface: {
+      value: { base: '{colors.legacy.BGLight}', _dark: '{colors.legacy.B20}' },
+      type: 'color',
+    },
+    text: {
+      value: { base: '#B743FF', _dark: '#AD74FF' },
+      type: 'color',
+    },
+  },
```
ainouzgali: @antonjoel82 should I add this color, that is only used here to the color tokens? I wasn't sure, because for every palette there, there are shades for each of them (example: amber.40, amber.60 etc)

ainouzgali: [Figma](https://www.figma.com/design/ybKRCB2ZO5DqICxjxzZGnW/Echo?node-id=3519-33552&m=dev) for context

libs/novui/src/json-schema-components/widgets/InputEditorWidget.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+      lineHeight: '125',
+      borderColor: 'input.border !important',
+      border: 'solid',
+    },
+  }),
+  content: css({
+    background: 'input.surface !important',
+    borderColor: 'input.border !important',
+    borderRadius: 'input !important',
+    lineHeight: '125 !important',
```
antonjoel82: 💭 **thought (non-blocking):** I don't believe most of these `!important` should be necessary any more after the CSS Layers PR got merged!‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const { variables = [] } = formContext;
+  const reactRenderer = useRef<ReactRenderer<SuggestionListRef>>(null);
+
+  const variablesList = useMemo<VariableItem[]>(() => {
+    return variables?.map((variable: string) => {
+      return { label: variable, id: variable };
+    });
+  }, [variables]);
+
+  const baseExtensions: Extensions = [Document, Paragraph, Text];
```
antonjoel82, line 43: ✏ **suggestion:** If it works, please declare outside of the component so as not to define on every render‏

libs/novui/src/tokens/semanticColors.tokens.ts
antonjoel82: @ainouzgali I would consider this a "design smell" -- we should ask design to stay within our color palettes or add these colors fully. I commented on the Figma :)

libs/novui/src/tokens/semanticColors.tokens.ts
```diff
@@ -203,6 +203,16 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
       type: 'color',
     },
   },
+  suggestion: {
+    surface: {
+      value: { base: '{colors.legacy.BGLight}', _dark: '{colors.legacy.B20}' },
+      type: 'color',
+    },
```
antonjoel82, line 210: ❓ **question:** ‏‏Do we have designs for these values? Can we match these to other things we already have (i.e. existing menus)? If not, it be worthwhile to move this to the top-level `surfaces` tokens as `menu`.

apps/web/src/studio/utils/variables.ts
```diff
@@ -0,0 +1,14 @@
+import { SystemVariablesWithTypes } from '@novu/shared';
+
+export function getSuggestionVariables(schemaObject, namespace: string) {
```
antonjoel82, line 3: ❓ **question:** Do we have more explicit types for `schemaObject`? Please add if so‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,132 @@
...
+import Mention from '@tiptap/extension-mention';
+
+import { css, cx } from '../../../styled-system/css';
+import { input } from '../../../styled-system/recipes';
+import { splitCssProps } from '../../../styled-system/jsx';
+import classes from './InputEditorWidget.styles';
+
+import { VariableSuggestionList, SuggestionListRef, VariableItem } from './VariableSuggestionList';
+
+import '@mantine/tiptap/styles.css';
```
antonjoel82:
```suggestion
import '@mantine/tiptap/styles.layer.css';
```

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.styles.ts
```diff
@@ -0,0 +1,36 @@
...
+    padding: '25',
+    marginY: '25',
+    border: 'none !important',
+    boxShadow: 'medium !important',
+    color: 'typography.text.main',
+    maxHeight: '[200px]',
+    overflow: 'none !important',
+    overflowY: 'auto',
+    textOverflow: 'ellipsis',
+  }),
```
antonjoel82: 💭 **thought (non-blocking):** Same note about `!important` here‏

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.tsx
```diff
@@ -0,0 +1,89 @@
...
+      command(item);
+    };
+
+    return createPortal(
+      <Menu
+        opened={opened}
+        closeOnEscape
+        classNames={classes}
+        position="bottom-start"
+        width={200}
```
antonjoel82:
```suggestion
```
Please define a semantic size token and use CSS instead if possible. If not, please use `token()` here to use the new token

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.tsx
```diff
@@ -0,0 +1,89 @@
...
+import classes from './VariableSuggestionList.styles';
+
+export type VariableItem = {
+  id: string;
+  label: string;
+};
+
+export type SuggestionListRef = {
+  onKeyDown: (props: SuggestionKeyDownProps) => boolean;
+  focus: () => void;
```
antonjoel82, line 15:
```suggestion
  focus: () => void;
  close: () => void;
```

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+type SuggestionListProps = SuggestionProps<VariableItem>;
+
+export const VariableSuggestionList = forwardRef<SuggestionListRef, SuggestionListProps>(
+  ({ clientRect, command, query, items }, ref) => {
+    const [opened, { close: closeSuggestionList, open: openSuggestionList }] = useDisclosure(true);
```
antonjoel82, line 25: 👏 **praise:** I like being verbose here to make the code easier to read elsewhere -- well done!‏

antonjoel82 (commented):
Nice job, and great work on the clean-up!

One main note: I apologize for the reversal, but I do actually think we should use `defineSlotRecipe` for each of these widget pieces -- they could even be combined into one recipe. The reason being that it will clean up the CSS (both in the classNames added to the elements in the app but also in the generated CSS). If for some reason styles don't seem to be generating properly, we can use the `staticCss` property in the recipe definition as we did in `json-schema-form.recipe.ts`

Lastly, I think we may need to refine some of the styles for long strings and the styles that appear when using the keyboard to navigate the suggestion list:
Screenshot 2024-07-22 at 10 22 07 AM

P.S. I'd suggest developing in Storybook to develop more quickly and see things in isolation. I just added autocomplete to the stories

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
ainouzgali: I tried multiple things to imitate what you did for the `core/styles.layer.css` but couldn't get it to work. It loses all styles when I change it to the layer 
I assume because its a different mantine package (tiptap), I tried adding it to the exports in package.json and adding it to imports in index.css but couldn't get it to work 🤦‍♀️

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
ainouzgali: Unfortunately it doesn't work

libs/novui/src/tokens/semanticColors.tokens.ts
ainouzgali: It is not for the menu, it is for the variable code that shows inside the input `bla bla {{var}}` , so I feel it is not generic enough to be moved to the top level. Wdyt?
But I do see now that naming it suggestion is confusing, I'll change it either way.

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.styles.ts
ainouzgali: I'll remove the `!important` after I get the layer issue working 😬

apps/web/src/studio/utils/variables.ts
ainouzgali: Technically its type should be the `RJSFSchema` that comes from the`@rjsf/utils` package that we don't import in web.

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.tsx
ainouzgali: Let me know if I did it ok, we didn't have any size tokens for bigger sizes

libs/novui/src/tokens/semanticColors.tokens.ts
antonjoel82: Sounds reasonable, not top level makes sense then, thanks!

apps/web/src/studio/utils/variables.ts
antonjoel82: Gotcha -- then it sounds to me like this helper should live in `novui`!

ainouzgali: I thought about it, but it feels like we are already adding too much logic to novui, which we shouldn't do in a design system. No?

ainouzgali:
@antonjoel82 for the recipe for the `InputEditorWidget` I would have liked to use parts of the `INPUT_RECIPE` in the new recipe- like we should as we  mostly want the same design -  but because I didn't manage to add the layer styles, it will not work as I supposedly still need to have `!important` for those styles.  
So, wondering what you think about either insisting on the layer vs duplicating the styles

ainouzgali:
@antonjoel82 I don't know, I thought about giving up on adding the layer for now, because it was just not working. 

But, now, with no layer and all !important added, the only change from before (where it all worked) is the change [from the classname](https://github.com/novuhq/novu/pull/6097/files/ec774c09d3a3cfb2d7858584a75592804209c168#diff-98d34ad09b65898eb4d5ae1bc801f34ecb9851a04ffb7a7593935a9d7bf131b0)s (wrapped in css) [to use recipes instead](https://github.com/novuhq/novu/pull/6097#pullrequestreview-2192080747).  Its fine in storybook but web is not getting the right styles. I cleaned several times and rebuilt, I just don't know... 
Should there be a difference between recipes vs classnames styles that should have a different affect like that? I figured it was some build or cache but I really cleaned everything  

Let me know what you think. Otherwise I am leaning to just returning the styles as it was before even though it will be less clean in the elements and the generated css.

ainouzgali:
@antonjoel82 thanks for the style fixes! It seems to work fine now in web. Will appreciate your final overview and approval so we can merge this 💪

libs/novui/src/recipes/input-editor-widget.recipe.ts
```diff
@@ -0,0 +1,50 @@
...
+        borderRadius: 'xs',
+        background: 'variable.surface',
+        p: '25',
+        lineHeight: '125',
+        borderColor: 'variable.border',
+        border: 'solid',
+      },
+    },
+  },
+  staticCss: ['*'],
```
antonjoel82, line 49: 🗒 **note (non-blocking):** This tells Panda to generate styles for all variants and slots, regardless of which ones are actually used in code. When using `defineSlotRecipe`, Panda normally only generates classes and styles for recipes that the static checker detects are used in code.‏

libs/novui/src/tokens/semanticColors.tokens.ts
```diff
@@ -203,6 +203,20 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
       type: 'color',
     },
   },
+  variable: {
+    surface: {
+      value: { base: '{colors.legacy.BGLight}', _dark: '{colors.legacy.B20}' },
+      type: 'color',
+    },
+    text: {
+      value: { base: '#B743FF', _dark: '#AD74FF' },
```
antonjoel82, line 213:
```suggestion
      // TODO: design debt -- waiting for a proper color palette / solution from design
      value: { base: '#B743FF', _dark: '#AD74FF' },
```

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,134 @@
...
+      const newValue = value
+        ?.toString()
+        .replace(
+          AUTOCOMPLETE_REGEX,
+          '<span data-id="$1" contenteditable="false" class="suggestion" data-type="mention">$1</span>'
+        );
+
+      editor.commands.setContent(newValue);
+    }
+  }, []);
```
antonjoel82, line 118: ❓ **question:** Should we include `editor` here as a dependency? If it breaks things, then do we need an eslint-disable here for dependencies here? ‏

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.tsx
```diff
@@ -0,0 +1,90 @@
...
+        // for some reason these don't seem to work, so we use clickOutRef on the dropdown
+        closeOnClickOutside
+        clickOutsideEvents={['click', 'mousedown', 'touchstart']}
+      >
+        <Menu.Target>
+          <div
+            style={{
+              position: 'absolute',
+              top: clientRect?.()?.bottom,
+              left: clientRect?.()?.left,
```
antonjoel82, line 73:
```suggestion
              // inline styles must be used here (instead of Panda) for dynamic / conditional behavior
              top: clientRect?.()?.bottom,
              left: clientRect?.()?.left,
```

.source
antonjoel82, line 1: 🗒 **note (non-blocking):** I have accidentally included this while merging `next` into the branch. If this isn't necessary, can you please reset it to the `.source` from next?‏

antonjoel82 (approved):
Awesome work!

PR #6098, opened 2024-07-17
fix(web): ensure we wait for loads to happen

denis-kralj-novu:
Ensure we wait for the environment request to happen before we say we are done loading


apps/web/src/components/providers/EnvironmentProvider.tsx
```diff
@@ -69,15 +69,19 @@ function selectEnvironment(environments: IEnvironment[] | undefined | null, sele
...
   const {
     data: environments,
-    isLoading,
+    isLoading: isLoadingEnvironments,
     refetch: refetchEnvironments,
   } = useQuery<IEnvironment[]>([QueryKeys.myEnvironments, currentOrganization?._id], getEnvironments, {
     enabled: !!currentOrganization,
     retry: false,
     staleTime: Infinity,
+    onSettled: (data, error) => {
```
ChmaraX, line 82: 🗒 **note (non-blocking):** ‏ `onSettled` seems [to be deprecated](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery) due to unstable behaviour 

maybe `onSuccess` would be sufficient replacement?

ChmaraX (approved):

apps/web/src/components/providers/EnvironmentProvider.tsx
denis-kralj-novu: but I also want it to trigger on error, so an alternative would be to have both the onSuccess and onFailure handler. But the hooks need to be

PR #6099, opened 2024-07-17
feat(js): inbox - single websocket connection across tabs

LetItRock:
In this PR:
- Implemented the locking mechanism across multiple tabs with `Web Locks API`. The locking mechanism establishes the single WebSocket connection across multiple tabs.
- Single WebSocket event broadcasting with `BroadcastChannel`

### Screenshots

https://github.com/user-attachments/assets/af95955d-1abc-4f31-bb82-ee0612ccbc7e

### DIFF
--- a/packages/js/src/ui/components/elements/Bell/Bell.tsx
+++ b/packages/js/src/ui/components/elements/Bell/Bell.tsx
@@ -1,4 +1,6 @@
 import { createSignal, JSX, onCleanup, onMount } from 'solid-js';
+import { requestLock } from '../../../helpers/browser';
+import { NV_INBOX_WEBSOCKET } from '../../../helpers/constants';
 import { useNovu } from '../../../context';
 import { BellContainer } from './DefaultBellContainer';
 
@@ -14,11 +16,14 @@ export function Bell(props: BellProps) {
   };
 
   onMount(() => {
-    novu.on('notifications.unread_count_changed', updateReadCount);
-  });
+    const resolveLock = requestLock(NV_INBOX_WEBSOCKET, () => {
+      novu.on('notifications.unread_count_changed', updateReadCount);
+    });
 
-  onCleanup(() => {
-    novu.off('notifications.unread_count_changed', updateReadCount);
+    onCleanup(() => {
+      novu.off('notifications.unread_count_changed', updateReadCount);
+      resolveLock();
+    });
   });
 
   if (props.children) {
--- a/packages/js/src/ui/components/elements/Bell/Bell.tsx
+++ b/packages/js/src/ui/components/elements/Bell/Bell.tsx
@@ -1,37 +1,16 @@
-import { createSignal, JSX, onCleanup, onMount } from 'solid-js';
-import { requestLock } from '../../../helpers/browser';
-import { NV_INBOX_WEBSOCKET_LOCK } from '../../../helpers/constants';
-import { useNovu } from '../../../context';
+import { Accessor, JSX } from 'solid-js';
 import { BellContainer } from './DefaultBellContainer';
-import { useTabsChannel } from '../../../helpers/useTabsChannel';
+import { useUnreadCount } from '../../../context/UnreadCountContext';
 
 type BellProps = {
-  children?: ({ unreadCount }: { unreadCount: number }) => JSX.Element;
+  children?: ({ unreadCount }: { unreadCount: Accessor<number> }) => JSX.Element;
 };
 export function Bell(props: BellProps) {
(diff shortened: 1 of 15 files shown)


packages/js/src/ui/components/elements/Bell/Bell.tsx
```diff
@@ -1,24 +1,32 @@
...
   const updateReadCount = (data: { result: number }) => {
     setUnreadCount(data.result);
+    postMessage(data.result);
   };
 
   onMount(() => {
-    novu.on('notifications.unread_count_changed', updateReadCount);
-  });
+    const resolveLock = requestLock(NV_INBOX_WEBSOCKET_LOCK, () => {
+      novu.on('notifications.unread_count_changed', updateReadCount);
```
LetItRock: The WebSocket connection is established only when starting to listen on the one of WebSocket events. That's why wrapping it with Web Lock API is safe.

packages/js/src/ui/components/elements/Bell/Bell.tsx
```diff
@@ -1,24 +1,32 @@
...
 
   const updateReadCount = (data: { result: number }) => {
     setUnreadCount(data.result);
+    postMessage(data.result);
   };
 
   onMount(() => {
-    novu.on('notifications.unread_count_changed', updateReadCount);
-  });
+    const resolveLock = requestLock(NV_INBOX_WEBSOCKET_LOCK, () => {
```
LetItRock: The helper function uses the Web Lock API to get the single lock between multiple tabs. Only one tab will get it.

packages/js/src/ui/components/elements/Bell/Bell.tsx
```diff
@@ -1,24 +1,32 @@
...
 import { BellContainer } from './DefaultBellContainer';
+import { useTabsChannel } from '../../../helpers/useTabsChannel';
 
 type BellProps = {
   children?: ({ unreadCount }: { unreadCount: number }) => JSX.Element;
 };
 export function Bell(props: BellProps) {
   const [unreadCount, setUnreadCount] = createSignal(0);
   const novu = useNovu();
+  const { postMessage } = useTabsChannel<number>({ onMessage: setUnreadCount });
```
LetItRock: The helper hook allows sending a message across multiple tabs. It's based on the `BroadcastChannel` API. The message will be only received by the other tabs but not the sender.

packages/js/src/ui/helpers/browser.ts
```diff
@@ -0,0 +1,21 @@
...
+  const promise = new Promise<void>((res) => {
+    resolve = res;
+  });
+
+  navigator.locks.request(id, () => {
+    if (!isFullfilled) {
+      cb(id);
+    }
+
+    return promise;
```
LetItRock, line 14: hold the lock until the promise is resolved

packages/js/src/ws/socket.ts
```diff
@@ -87,13 +87,17 @@ export class Socket extends BaseModule {
 
   initialize(): void {
     if (this.#token) {
-      this.#initializeSocket().then((error) => console.error(error));
+      this.#initializeSocket().catch((error) => {
```
LetItRock, line 90: noticed this small mistake that was causing the `undefined` error in the console

packages/js/src/ui/components/elements/Bell/Bell.tsx
SokratisVidros: Nice, we can create one more level of abstraction and introduce a `requestWSLock` in helper and remove the `NV_INBOX_WEBSOCKET_LOCK` import from this file.

SokratisVidros (changes requested):
@LetItRock the single websocket connection across tabs is a headless concern. That is, customer who use @novu/js headless without any UI, should be able to experience the same functionality and have one connection across all tabs in the same domain. 

I suggest we move this logic to the lower level and then have the UI consume only the Novu specific events of unread count changes.

LetItRock:
> @LetItRock the single websocket connection across tabs is a headless concern. That is, customer who use @novu/js headless without any UI, should be able to experience the same functionality and have one connection across all tabs in the same domain.
> 
> I suggest we move this logic to the lower level and then have the UI consume only the Novu specific events of unread count changes.

@SokratisVidros This code is strongly based on the Web APIs, if we move it to the lower level it would mean that we have to exclude it from executing in the mobile, and server environments.
The Socket interface would have to be improved as well to support subscribing to events without the establishing connection.

SokratisVidros (approved):

PR #6108, opened 2024-07-18
refactor(web): simplify and break down auth providers 

ChmaraX:
Generally cleaned up the auth providers to increase the code readability, since they packed too much logic it was hard to navigate

* extracted shared auth logic to `useCommonAuth` hook (auth version specific side effects can be passed as arguments, e.g. logout)
* refactored some of the methods and constants
* moved EE organization guard useEffect to standalone `withOrganizationGuard` 
* used `createContextAndHook` utility for auth providers similarly as in `EnterpriseProvider`

### DIFF
--- a/apps/web/src/components/providers/AuthProvider.tsx
+++ b/apps/web/src/components/providers/AuthProvider.tsx
@@ -1,7 +1,8 @@
 import { IS_EE_AUTH_ENABLED } from '../../config/index';
 import { IOrganizationEntity, IUserEntity } from '@novu/shared';
 import { CommunityAuthCtx, CommunityAuthProvider } from './CommunityAuthProvider';
-import { EnterpriseAuthCtx, EnterpriseAuthProvider } from '../../ee/clerk/providers/EnterpriseAuthProvider';
+import EnterpriseAuthProvider, { EnterpriseAuthCtx } from '../../ee/clerk/providers/EnterpriseAuthProvider';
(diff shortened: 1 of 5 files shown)


apps/web/src/components/providers/AuthProvider.tsx
```diff
@@ -25,16 +25,16 @@ export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
...
 
-  // eslint-disable-next-line react-hooks/rules-of-hooks
   return <CommunityAuthProvider>{children}</CommunityAuthProvider>;
 };
 
 export const useAuth = () => {
-  const context = IS_EE_AUTH_ENABLED ? EnterpriseAuthContext : CommunityAuthContext;
-  const value = useContext(context);
+  const ctx = IS_EE_AUTH_ENABLED ? EnterpriseAuthCtx : CommunityAuthCtx;
+  const context = useContext(ctx);
```
ChmaraX, line 33: Can't use `useEnterpriseAuthCtx` from the `createContextAndHook` utility here because that would result in usage like this:

```typescript
const value = IS_EE_AUTH_ENABLED ? useEnterpriseAuthCtx() : useCommunityAuthCtx()
```

Which triggers both hooks and violates ESLint rule of hooks

SokratisVidros:
@ChmaraX PTAL https://github.com/novuhq/novu/pull/6123

PR #6114, opened 2024-07-18
fix(web): V2 minor fixes

antonjoel82:

- Make Studio nav buttons collapsible, but only the active one open ([ticket](https://linear.app/novu/issue/NV-4130/arrow-on-workflows-list-doesnt-collapse-when-clicked-should-have-a))
- Fix Step Preview error display to include relevant information and change the aesthetic
- Fix cli reference to `web` (--> `dashboard`)

https://www.loom.com/share/68ac81703ecf4d269bca71793a7d39f2?sid=c5e00b28-dd9f-493f-82dd-63149e2d4e62

### DIFF
--- a/apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarToggleButton.tsx
+++ b/apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarToggleButton.tsx
@@ -8,7 +8,7 @@ import {
   getStudioWorkflowStepLink,
   getStudioWorkflowTestLink,
 } from '../../../../studio/utils/routing';
-import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton/index';
+import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton';
 
 type LocalStudioSidebarToggleButtonProps = {
   workflow: IBridgeWorkflow;
--- a/apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarToggleButton.tsx
+++ b/apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarToggleButton.tsx
@@ -8,7 +8,7 @@ import {
   getStudioWorkflowStepLink,
   getStudioWorkflowTestLink,
 } from '../../../../studio/utils/routing';
-import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton';
+import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton/index';
 
 type LocalStudioSidebarToggleButtonProps = {
   workflow: IBridgeWorkflow;
--- a/apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
+++ b/apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
@@ -1,37 +1,74 @@
 import { css } from '@novu/novui/css';
-import { Text } from '@novu/novui';
-import { Center } from '@novu/novui/jsx';
-import { IconErrorOutline } from '@novu/novui/icons';
+import { Button, Text, Title } from '@novu/novui';
+import { Center, Stack, VStack } from '@novu/novui/jsx';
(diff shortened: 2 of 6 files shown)


scopsy (approved):

apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarToggleButton.tsx
```diff
@@ -8,7 +8,7 @@ import {
   getStudioWorkflowStepLink,
   getStudioWorkflowTestLink,
 } from '../../../../studio/utils/routing';
-import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton';
+import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton/index';
```
rifont: ❓ **question:** Is there a specific reason we needed to add this import from `index`? In my experience, it's usually a symptom of a circular dependency in the tree.‏

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
```diff
@@ -1,37 +1,74 @@
 import { css } from '@novu/novui/css';
-import { Text } from '@novu/novui';
-import { Center } from '@novu/novui/jsx';
-import { IconErrorOutline } from '@novu/novui/icons';
+import { Button, Text, Title } from '@novu/novui';
+import { Center, Stack, VStack } from '@novu/novui/jsx';
+import { IconErrorOutline, IconExpandLess, IconExpandMore } from '@novu/novui/icons';
+import { useDisclosure } from '@mantine/hooks';
+import { hstack } from '@novu/novui/patterns';
```
rifont, line 6: 💭 **thought (non-blocking):** Is it worthwhile to add an `HSack` component into `novui`? I've seen a few places we use it.‏

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
```diff
@@ -1,37 +1,74 @@
...
         bg: 'input.border.error/20',
-        borderColor: 'input.border.error !important',
-        border: '1px solid !important',
+        border: 'solid',
+        borderColor: 'input.border.error',
         width: '100%',
-        borderRadius: 'input !important',
-        my: 'margins.layout.tabs.bottom !important',
+        borderRadius: 'input',
+        my: 'margins.layout.tabs.bottom',
```
rifont, line 21: 👏 **praise:** 🧹😍‏

packages/cli/src/commands/init/templates/github/workflows/novu.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         uses: novuhq/actions-novu-sync@v2
         with:
           # The secret key used to authenticate with Novu Cloud
-          # To get the secret key, go to https://web.novu.co/api-keys.
+          # To get the secret key, go to https://dashboard.novu.co/api-keys.
```
rifont, line 15: 👏 **praise:** Nice catch!‏

rifont (approved):
Awesome work Joel!

apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarToggleButton.tsx
antonjoel82: Nope -- good catch! I'm not sure why this is happening suddenly with VSCode, but I've seen this in my changes and other peoples' recently. That's an interesting note on circular dependencies though -- I'll keep an eye out

antonjoel82:
```suggestion
import { NavMenuLinkButton, NavMenuToggleButton } from '../../../nav/NavMenuButton';
```

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
antonjoel82: @rifont It is a pre-built component by Panda, but the beauty of it is that it's also a `pattern` that generates a `className`! This allows us to style semantic elements (button in this case) without having to use a polymorphic component (aka adding `as="button"`), which tends to help TypeScript in my experience

PR #6121, opened 2024-07-19
feat(api): change sync org request scope to default (singleton)

ChmaraX:
- there is no need for this to be request scoped since it doesn't hold any state
- this change is also required because we need to inject this usecase to passport strategy which can inject only singletons 

> Websocket Gateways should not use request-scoped providers because they must act as singletons. Each gateway encapsulates a real socket and cannot be instantiated multiple times. The limitation also applies to some other providers, like [Passport strategies](https://docs.nestjs.com/security/authentication#request-scoped-strategies) or Cron controllers.

https://github.com/novuhq/packages-enterprise/pull/173


denis-kralj-novu (approved):
🚀

PR #6125, opened 2024-07-21
feat(js): Implement the renderNotification prop

desiprisg:

We need a way to allow for the react package to render custom elements in the `@novu/js` package. A way to do this, is to have a renderer component in the react package, that is responsible for rendering all these elements. This includes a map of all elements that are rendered.

A render prop is defined in the react package. This needs to be transformed to a `mount` prop for `@novu/js` to be able to use it as it has no way to render JSX from react. The transformation is as simple as appending a new element to the renderer's map, with the render prop as the data (the key is the `HTMLElement`). This triggers a rerender for the `<Renderer />` in `@novu/react` and the component will be mounted. In this case, the components that are mounted are the notifications, using the `renderNotification` prop.

We will need to make some performance improvements here around the mounts/unmounts that happen.

Essentially the flow is:
* `@novu/react` wraps everything with a `<Renderer />` component. This is also a context, so our custom components are able to mutate it's state.
* Developer uses our `<Inbox />` component and defines a `renderNotification` prop.
* This `renderNotification` prop is converted to a function that mutates the `<Renderer />` state and adds a new element (a custom notification).
* This `mountNotification` function that was created is passed to `@novu/js` as a prop for the Inbox.
* Every time a notification is rendered by `@novu/js`, this function is triggered, given an `HTMLElement` that we render on `@novu/js`(this is the element that the custom notification will be portaled onto).
* `@novu/react`'s `<Renderer />` is rerendered and portals the custom notifications on the `div`s that `@novu/js` placed in the DOM.

An example would be :
```
      <Inbox
        options={{
          ...
        }}
        renderNotification={(notification) => {
          return <button className="p-10 bg-red-300 w-full">{notification.body}</button>;
        }}
      />
```
that results in: 

Screenshot 2024-07-21 at 1 35 52 PM

### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/ExternalElementMounter.tsx
+++ b/packages/js/src/ui/components/ExternalElementMounter.tsx
@@ -1,0 +1,26 @@
+import { onCleanup, onMount, ParentProps } from 'solid-js';
+
+type ExternalElementMounterProps = ParentProps<{
+  mount: (el: HTMLDivElement) => () => void;
+}>;
+
+export const ExternalElementMounter = ({ mount, ...rest }: ExternalElementMounterProps) => {
+  let ref: HTMLDivElement;
+
+  onMount(() => {
+    const unmount = mount(ref);
+
+    onCleanup(() => {
+      unmount();
+    });
+  });
+
+  return (
+    <div
+      ref={(el) => {
+        ref = el;
+      }}
+      {...rest}
+    />
+  );
+};
--- a/packages/js/src/ui/components/Inbox.tsx
+++ b/packages/js/src/ui/components/Inbox.tsx
@@ -1,11 +1,13 @@
 import { Accessor, createSignal, JSX, Match, Switch } from 'solid-js';
 import { useStyle } from '../helpers';
+import { NotificationMounter } from '../types';
 import { Bell, Footer, Header, Settings, SettingsHeader } from './elements';
 import { NotificationList } from './Notification';
 import { Button, Popover } from './primitives';
 
-type InboxProps = {
+export type InboxProps = {
   open?: boolean;
+  mountNotification?: NotificationMounter;
   renderBell?: ({ unreadCount }: { unreadCount: Accessor<number> }) => JSX.Element;
 };
 
@@ -13,7 +15,11 @@ enum Screen {
   Inbox = 'inbox',
   Settings = 'settings',
 }
-const InboxContent = () => {
+
+type InboxContentProps = {
+  mountNotification?: NotificationMounter;
+};
+const InboxContent = (props: InboxContentProps) => {
   const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
 
   return (
@@ -21,7 +27,7 @@ const InboxContent = () => {
       <Switch>
         <Match when={currentScreen() === Screen.Inbox}>
           <Header updateScreen={setCurrentScreen} />
-          <NotificationList />
+          <NotificationList mountNotification={props.mountNotification} />
         </Match>
         <Match when={currentScreen() === Screen.Settings}>
           <SettingsHeader backAction={() => setCurrentScreen(Screen.Inbox)} />
@@ -46,7 +52,7 @@ export const Inbox = (props: InboxProps) => {
         )}
       />
       <Popover.Content appearanceKey="inbox__popoverContent">
-        <InboxContent />
+        <InboxContent mountNotification={props.mountNotification} />
       </Popover.Content>
     </Popover.Root>
   );
--- a/packages/js/src/ui/components/Notification/DefaultNotification.tsx
+++ b/packages/js/src/ui/components/Notification/DefaultNotification.tsx
@@ -1,0 +1,30 @@
+import { Show } from 'solid-js';
+import { InboxNotification } from '../../../types';
+import { useStyle } from '../../helpers';
+
+type DefaultNotificationProps = {
+  notification: InboxNotification;
+};
+
+//TODO: Complete the implementation
+export const DefaultNotification = (props: DefaultNotificationProps) => {
+  const style = useStyle();
+
+  return (
+    <div class={style('notificationContainer', 'nt-w-full nt-p-6')}>
+      <div class={style('notificationSubjectContainer', 'nt-flex nt-items-center nt-gap-2 nt-relative')}>
+        <span
+          class={style(
+            'notificationDot',
+            'nt-absolute -nt-left-3 nt-top-0 -nt-translate-x-1/2 nt-translate-y-1/2 nt-size-3 nt-bg-primary nt-rounded-full nt-border'
+          )}
+        />
+        <Show when={props.notification.subject}>
+          <span class={style('notificationSubject', 'nt-w-full nt-font-semibold')}>{props.notification.subject}</span>
+        </Show>
+        <span>{props.notification.createdAt}</span>
+      </div>
+      {props.notification.body}
+    </div>
+  );
+};
--- a/packages/js/src/ui/components/Notification/Notification.tsx
+++ b/packages/js/src/ui/components/Notification/Notification.tsx
@@ -1,0 +1,18 @@
+import { Show } from 'solid-js';
+import { InboxNotification } from '../../../types';
+import { NotificationMounter } from '../../types';
+import { ExternalElementMounter } from '../ExternalElementMounter';
+import { DefaultNotification } from './DefaultNotification';
+
+type NotificationProps = {
+  notification: InboxNotification;
+  mountNotification?: NotificationMounter;
+};
+
+export const Notification = (props: NotificationProps) => {
+  return (
+    <Show when={props.mountNotification} fallback={<DefaultNotification notification={props.notification} />}>
+      <ExternalElementMounter mount={(el) => props.mountNotification!(el, { notification: props.notification })} />
+    </Show>
+  );
+};
--- a/packages/js/src/ui/components/Notification/NotificationList.tsx
+++ b/packages/js/src/ui/components/Notification/NotificationList.tsx
@@ -4,6 +4,8 @@ import { useFeedInfiniteScroll } from '../../api';
 import { useLocalization } from '../../context';
 import { useStyle } from '../../helpers';
 import { EmptyIcon } from '../../icons/EmptyIcon';
+import { NotificationMounter } from '../../types';
+import { Notification } from './Notification';
 import { NotificationListSkeleton, NotificationSkeleton } from './NotificationListSkeleton';
 
 export const NotificationListContainer = (props: ParentProps) => {
@@ -38,8 +40,10 @@ const EmptyNotificationList = () => {
 };
 
 type NotificationListProps = {
+  mountNotification?: NotificationMounter;
   options?: FetchFeedArgs;
 };
+/* This is also going to be exported as a separate component. Keep it pure. */
 export const NotificationList = (props: NotificationListProps) => {
   const [data, { initialLoading, setEl, end }] = useFeedInfiniteScroll({ options: props.options });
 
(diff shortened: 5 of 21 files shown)


packages/js/src/api/inbox-service.ts
```diff
@@ -12,6 +12,7 @@ export class InboxService {
 
   constructor(options: InboxServiceOptions = {}) {
     this.#httpClient = new HttpClient(options);
+    //@ts-ignore
```
desiprisg: Ignore this, removing..

packages/js/src/ui/components/ExternalElementMounter.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const ExternalElementMounter = ({ mount, ...rest }: ExternalElementMounterProps) => {
+  let ref: HTMLDivElement;
+
+  onMount(() => {
+    const unmount = mount(ref);
+
+    return () => {
+      unmount();
+    };
```
BiswaViraj: You will need to use `onCleanup` method
This wont work in Solid :(

BiswaViraj (changes requested):

packages/js/src/ui/components/ExternalElementMounter.tsx
desiprisg: indeed 😆

playground/nextjs/src/components/Inbox.tsx
```diff
@@ -0,0 +1,57 @@
...
+    return <Renderer options={props} />;
+  }
+
+  const { renderNotification, ...options } = props as BaseNovuUIOptions & AIOInboxProps;
+  return (
+    <Renderer options={options}>
+      <AIOInbox renderNotification={renderNotification} />
+    </Renderer>
+  );
+};
```
LetItRock: ☑ **chore:** It doesn't work now without the `renderNotification`:
```
<Inbox />
```

playground/nextjs/src/components/Renderer.tsx
```diff
@@ -0,0 +1,55 @@
...
+      setMountedElements((mountedElements) => {
+        const newMountedElements = new Map(mountedElements);
+        newMountedElements.set(el, mountedElement);
+        return newMountedElements;
+      });
+
+      return () => {
+        setMountedElements((mountedElements) => {
+          const newMountedElements = new Map(mountedElements);
+          newMountedElements.set(el, mountedElement);
```
LetItRock: ☑ **chore:** not clearing it appropriately

playground/nextjs/src/components/Inbox.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useRenderer } from '@/context/RendererContext';
+import { InboxNotification } from '@novu/js/dist/types/types';
+import type { BaseNovuUIOptions } from '@novu/js/ui';
+import { ReactNode, useCallback, useEffect } from 'react';
+import { Renderer } from './Renderer';
+
+type AIOInboxProps = {
+  renderNotification?: (notification: InboxNotification) => ReactNode;
+};
+const AIOInbox = (props: AIOInboxProps) => {
```
LetItRock: ✏ **suggestion:** What does the AIO stand for? Maybe just call it InboxMounter or InboxRenderer?

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -47,8 +50,9 @@ export const NotificationList = (props: NotificationListProps) => {
...
       <Show when={data().length > 0} fallback={<EmptyNotificationList />}>
         <NotificationListContainer>
-          {/* eslint-disable-next-line local-rules/no-class-without-style */}
-          <For each={data()}>{(notification) => <p class="nt-my-10">{notification.body}</p>}</For>
+          <For each={data()}>
+            {(notification) => <Notification notification={notification} mountNotification={props.mountNotification} />}
+          </For>
           <Show when={!end()}>
             <div ref={setEl}>
               <For each={Array.from({ length: 3 })}>{() => <NotificationSkeleton />}</For>
```
LetItRock, line 59: ✏ **suggestion:** I think we should provide a way to render also the custom skeleton

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -1,27 +1,32 @@
...
 };
 
 enum Screen {
   Inbox = 'inbox',
   Settings = 'settings',
 }
+
 const InboxContent = () => {
   const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
+  const { mountNotification } = useAIOInbox();
```
LetItRock: ✏ **suggestion:** please move it under the `NotificationList`

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -37,17 +42,19 @@ export const Inbox = (props: InboxProps) => {
...
-          <Button class={style('inbox__popoverTrigger')} variant="ghost" size="icon" {...triggerProps}>
-            <Bell>{props.renderBell}</Bell>
-          </Button>
-        )}
-      />
-      <Popover.Content appearanceKey="inbox__popoverContent">
-        <InboxContent />
-      </Popover.Content>
-    </Popover.Root>
+    <AIOInboxProvider {...props}>
```
LetItRock: ❓ **question:** what's the idea behind this provider? we only use one prop as far as I see...

packages/js/src/ui/components/Renderer.tsx
```diff
@@ -67,11 +67,13 @@ export const Renderer = (props: RendererProps) => {
...
               <InboxNotificationStatusProvider>
-                {[...props.nodes].map(([node, component]) => (
-                  <Portal mount={node}>
-                    <Root>{NovuComponents[component.name](component.props || {})}</Root>
-                  </Portal>
-                ))}
+                <For each={[...props.nodes]}>
+                  {([node, component]) => (
+                    <Portal mount={node}>
+                      <Root>{novuComponents[component.name](component.props || {})}</Root>
```
LetItRock, line 73: ❓ **question:** shouldn't this be like:
```
const Component = novuComponents[component.name];
...
<Root><Component {...props} /></Root>
```

because currently it works like a function invocation

LetItRock: nvm, I'll handle this case in my PR

packages/js/src/ui/components/Renderer.tsx
```diff
@@ -15,7 +15,7 @@ import { UnreadCountProvider } from '../context/UnreadCountContext';
 import { Bell, Root } from './elements';
 import { Inbox } from './Inbox';
 
-const NovuComponents = {
+export const novuComponents = {
   Inbox,
   Bell,
```
BiswaViraj, line 20: We need to export Settings and Notification list components too

playground/nextjs/src/components/Inbox.tsx
desiprisg: AIO stands for All in One

packages/js/src/ui/components/Renderer.tsx
desiprisg: We will :)

playground/nextjs/src/components/Inbox.tsx
desiprisg: The default notification item will have to be implemented as well.

packages/js/src/ui/components/Inbox.tsx
desiprisg: Ideally, `NotificationList` should not use the AIOInbox context, as it will also be used to export the `<Notifications />` component.

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -0,0 +1,9 @@
+import { InboxNotification } from '../../../types';
+
+type DefaultNotificationProps = {
+  notification: InboxNotification;
+};
+
+export const DefaultNotification = (props: DefaultNotificationProps) => {
+  return <>{props.notification.body}</>;
```
SokratisVidros: ❓ Is this where our default look-n-feel for notification item will live?

packages/js/src/ui/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { DefaultNotification } from './DefaultNotification';
+
+type NotificationProps = {
+  notification: InboxNotification;
+  mountNotification?: MountNotification;
+};
+
+export const Notification = (props: NotificationProps) => {
+  return (
+    <Show when={props.mountNotification} fallback={<DefaultNotification notification={props.notification} />}>
```
SokratisVidros, line 14: I ❤️ this DX.

packages/js/src/ui/components/Notification/Notification.tsx
```diff
@@ -0,0 +1,22 @@
...
+      <ExternalElementMounter
+        mount={(el) => {
+          return props.mountNotification!(el, { notification: props.notification });
```
SokratisVidros: 🤓  Make it an oneliner
```suggestion
  props.mountNotification!(el, { notification: props.notification });
```

packages/js/src/ui/components/Notification/NotificationList.tsx
SokratisVidros: Great idea! I added a linear ticket post release.

playground/nextjs/src/components/Inbox.tsx
SokratisVidros: @desiprisg Please rename it, it doesn't resonate very well.

packages/js/src/ui/components/Renderer.tsx
SokratisVidros: I'd also want to remind you that we need to export a Notifications components that also accepts `renderNotification` prop.

packages/js/src/ui/types.ts
```diff
@@ -1,12 +1,19 @@
 import { NovuOptions } from '../novu';
+import { InboxNotification } from '../types';
 import { Appearance, Localization } from './context';
 
-export type NovuProviderProps = {
+export type MountNotification = (el: HTMLDivElement, options: { notification: InboxNotification }) => () => void;
```
SokratisVidros:
```suggestion
export type NotificationMounter = (el: HTMLDivElement, options: { notification: InboxNotification }) => () => void;
```

SokratisVidros (changes requested):

packages/js/src/ui/components/Notification/DefaultNotification.tsx
desiprisg: Yes, creating this now

SokratisVidros: Should we merge this PR and start a new one?

playground/nextjs/src/components/Inbox.tsx
LetItRock: are you going to do that in this PR or separately?

playground/nextjs/src/components/Inbox.tsx
LetItRock: 🙃

packages/js/src/ui/components/Inbox.tsx
LetItRock: Are you saying that it won't be under that context?

playground/nextjs/src/components/Inbox.tsx
desiprisg: I'll do it separately.

packages/js/src/ui/components/Inbox.tsx
desiprisg: I removed the context as it's not a good idea. It was specific to the all in one inbox, so we can't really use it in the other components that we will export separately.

packages/js/src/ui/components/Renderer.tsx
desiprisg: Yeap, this is PR is like an MVP to show how we will be doing this back and forth.

packages/js/src/ui/components/Notification/DefaultNotification.tsx
desiprisg: Yeah, we'll do this since we're waiting on some possible design changes.

SokratisVidros (approved):

BiswaViraj (approved):

LetItRock (approved):

PR #6134, opened 2024-07-23
fix(worker): Use susbcriber locale for in-app and email steps

rifont:

* In-App and Email steps were not using the Subscriber's locale and were always falling back to using the Organization's default locale. This fix ensures that the Organization's default locale resolution remains with the Translations service and the Subscribers locale is used for translations of In-App and Email channel, aligning with the SMS, Push, and Chat channel implementation.

### Screenshots
_Email received using Susbcriber locale_
german-email-fix

_In-App received using Susbcriber locale_
image

<details>

</details>

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
@@ -53,7 +53,6 @@ export class SendMessageEmail extends SendMessageBase {
 
   constructor(
     protected environmentRepository: EnvironmentRepository,
-    protected organizationRepository: OrganizationRepository,
     protected subscriberRepository: SubscriberRepository,
     protected messageRepository: MessageRepository,
     protected layoutRepository: LayoutRepository,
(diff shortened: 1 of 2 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -212,12 +211,10 @@ export class SendMessageEmail extends SendMessageBase {
     }
 
     try {
-      const organization = await this.getOrganization(command.organizationId);
-
       const i18nInstance = await this.initiateTranslations(
         command.environmentId,
         command.organizationId,
-        command.payload.subscriber?.locale || organization?.defaultLocale
+        subscriber.locale
```
rifont, line 217: Aligning the logic here with what we have in [Push](https://github.com/novuhq/novu/blob/next/apps/worker/src/app/workflow/usecases/send-message/send-message-push.usecase.ts#L86), SMS, and Chat channel send-message use-cases.

The Translations service already handles the defaultLocale resolution of the Organization, so we remove the Organization fetch from these use-cases.

djabarovgeorge (approved):
🐏🐏🐏

PR #6136, opened 2024-07-23
refactor(worker): use 'system' for environmentId and userId for inline-css flag evaluation

rifont:

* use `'system'` for environmentId and userId for inline-css flag evaluation

### Screenshots

<details>

</details>

### DIFF
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
@@ -247,9 +247,9 @@ export class SendMessageEmail extends SendMessageBase {
         const shouldDisableInlineCss = await this.getFeatureFlag.execute(
           GetFeatureFlagCommand.create({
             key: FeatureFlagsKeysEnum.IS_EMAIL_INLINE_CSS_DISABLED,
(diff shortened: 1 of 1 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-email.usecase.ts
```diff
@@ -247,9 +247,9 @@ export class SendMessageEmail extends SendMessageBase {
         const shouldDisableInlineCss = await this.getFeatureFlag.execute(
           GetFeatureFlagCommand.create({
             key: FeatureFlagsKeysEnum.IS_EMAIL_INLINE_CSS_DISABLED,
-            environmentId: command.environmentId,
+            environmentId: 'system',
             organizationId: command.organizationId,
-            userId: command.userId,
+            userId: 'system',
```
rifont, line 252: Applying the same flag evaluation context as we have [here](https://github.com/novuhq/packages-enterprise/blob/next/billing/src/guards/quota-throttler.guard.ts#L65) for quota-throttler guard to attempt a fix at rule evaluation.

antonjoel82 (approved):
Verified locally!

PR #6137, opened 2024-07-23
feat(web): normalize user emails in ee

ChmaraX:
* added custom `onBeforeRequest` to Clerk instance, so we are able to login and repair users for which we store just normalized email 

### Screenshots

<details>

</details>


apps/web/src/ee/clerk/providers/clerk-singleton.ts
```diff
@@ -0,0 +1,45 @@
...
+export async function buildClerk({ publishableKey }: BuildClerkOptions): Promise<ClerkProp> {
+  if (clerk) {
+    return clerk as ClerkProp;
+  }
+
+  clerk = new Clerk(publishableKey);
+
+  clerk.__unstable__onBeforeRequest(async (requestInit) => {
+    const { path, method, body } = requestInit;
+    const isSignIn = path === '/client/sign_ins' && method === 'POST';
```
SokratisVidros, line 21: Should we maybe check with startsWith or path never includes the query string?

SokratisVidros (approved):

apps/web/src/ee/clerk/providers/clerk-singleton.ts
ChmaraX: updated to trigger only upon regular email/pass sign-in

PR #6145, opened 2024-07-24
chore(novui): Setup Panda-learning resources

antonjoel82:

- Setup a sandbox in Novui Storybook for trying basic Panda concepts
- Improve the Storybook preview height


libs/novui/src/panda-sandbox/css.stories.tsx
```diff
@@ -0,0 +1,46 @@
+import { Meta, StoryFn } from '@storybook/react/*';
+import React from 'react';
+import { css, cx } from '../../styled-system/css';
+
+export default {
+  title: 'Panda Sandbox/css',
+} as Meta;
+
+// css simply outputs a className string
```
rifont, line 9: Could we give an example of what the output string could look like?

libs/novui/src/panda-sandbox/css.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return <div className={className}>Hello Panda 🐼! css()</div>;
+};
+
+export const Standard = StandardTemplate.bind({});
+Standard.args = {};
+
+/**
+ * cx is a Panda-generated function for merging classNames.
+ * Like in normal css, later-defined properties take precedence.
+ */
```
rifont, line 40: A question that might arise is "Why can't we just string concatenate the classnames instead of using a function?"

I think we should answer that in this comment.

libs/novui/src/panda-sandbox/recipes/cva.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { css, cva, cx } from '../../../styled-system/css';
+import { Stack, styled, Wrap } from '../../../styled-system/jsx';
+
+export default {
+  title: 'Panda Sandbox/recipes/cva',
+} as Meta;
+
+/**
+ * cva is a way to create Atomic Recipes
+ * https://panda-css.com/docs/concepts/recipes#atomic-recipe-or-cva
```
rifont, line 13: For each of these, I think we should elaborate and give a concrete example of when we might use `cva`, for example:

"We use cva when we need to define multiple states for a single atomic component. A great example is for a status indicator where we have a mixture of common styling, and variant styling."

libs/novui/src/panda-sandbox/recipes/cva.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+      </Wrap>
+    </Stack>
+  );
+};
+
+export const ClassName = StandardTemplate.bind({});
+ClassName.args = {};
+
+/**
+ * Here we're using the `styled` function from Panda to create a type-safe JSX element based on the same recipe!
```
rifont, line 74: Good start, I'd add something like:

"This provides a better DX when the component needs to be consumed in multiple places because we only need to import the JSX component, rather than both the recipe and the component we'd like to apply styles to."

libs/novui/src/panda-sandbox/recipes/sva.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Stack } from '../../../styled-system/jsx';
+
+export default {
+  title: 'Panda Sandbox/recipes/sva',
+} as Meta;
+
+/**
+ * sva is a function to create recipes that target specific named parts / elements ("slots")
+ * of a component -- it returns an object of classNames.
+ * https://panda-css.com/docs/concepts/slot-recipes#atomic-slot-recipe-or-sva
```
rifont, line 14: This is great, however I'd consider using a more familiar form input example that everyone will have seen before. Then, I'd add something like:

"sva is similar to cva, but is typically used when we have a composition of components (i.e. a composite component) that contains multiple atomic components that we would like to style independently. A good example of where we would use sva is for a form input component recipe, where we have a label, text input, and helper text that each have their own styles."

libs/novui/src/panda-sandbox/css.stories.tsx
antonjoel82: I'm not really sure that that adds any benefit, and may actually just add unnecessary info

libs/novui/src/panda-sandbox/css.stories.tsx
antonjoel82: You can, this is just a nicer way to do it (like the well known `classnames` package`)

libs/novui/src/panda-sandbox/css.stories.tsx
rifont: Fair point, it's probably irrelevant and subject to changes as Panda matures.

rifont (approved):
Awesome work!

PR #6147, opened 2024-07-24
fix(web): make clerk redirect properly

denis-kralj-novu:


apps/web/src/ee/clerk/pages/OrganizationListPage.tsx
```diff
@@ -1,8 +1,15 @@
...
+import { useEffect } from 'react';
 import AuthLayout from '../../../components/layout/components/AuthLayout';
 import { ROUTES } from '../../../constants/routes';
+import { useRedirectURL } from '../../../hooks/useRedirectURL';
 
 export default function OrganizationListPage() {
+  const { setRedirectURL } = useRedirectURL();
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => setRedirectURL(), []);
```
SokratisVidros, line 11: Was this required for redirects to work? We have the same thing in Private layout that is triggered when Clerk is done, so I am curious which case foes the private layout misses?

SokratisVidros (approved):

apps/web/src/ee/clerk/pages/OrganizationListPage.tsx
denis-kralj-novu: Yes, this covers the new sign up case. Without it, the redirect url gets lost because the signup page has several steps where the page changes, and some of them are done via forceRedirectUrl. Adding this call fixed this. I could look at maybe using different props (e.g. `afterSignUpUrl` in the sign up component), maybe then this wouldn't be needed. Should I invest more time into this @SokratisVidros ?

SokratisVidros: Thanks for the explanation. Not for now, but let's add a Linear ticket to get back to this.

denis-kralj-novu: will do

PR #6149, opened 2024-07-24
feat(js): inbox tabs

LetItRock:
The Inbox tabs UI component implementation:
- the tabs prop drilling from the `NovuUI`
- the tabs primitive component
- dynamic visible tabs rendering logic based on available parent container space

Known limitations:
- ~it doesn't support keyboard navigation~
- it doesn't recalculate the "visible tabs" when the container size changes

### Screenshots

Screenshot 2024-07-24 at 17 35 33

Screenshot 2024-07-24 at 17 35 24

Screenshot 2024-07-24 at 17 52 11

https://github.com/user-attachments/assets/51ea626b-92de-4632-be1c-dbf0514ddb78

https://github.com/user-attachments/assets/48f4f83a-ff7a-47b7-b715-03ef4cf8dfb7


packages/js/src/ui/components/InboxTabs/InboxTab.tsx
```diff
@@ -0,0 +1,19 @@
+import { useStyle } from 'src/ui/helpers';
+import { Tabs } from '../primitives';
+
+export const InboxTab = (props: { label: string; class?: string }) => {
+  const style = useStyle();
+
+  // TODO: Replace with actual count from API
```
LetItRock, line 9: the fetching logic for the counts will be in the separate PR/ticket

packages/js/src/ui/components/InboxTabs/InboxTab.tsx
```diff
@@ -0,0 +1,19 @@
...
+  const count = Math.floor(Math.random() * 120 + 1);
+  return (
+    <Tabs.Tab value={props.label} class={`nt-flex nt-gap-2 ${props.class ?? ''}`}>
+      <span class={style('tabsTabLabel', 'nt-text-sm nt-font-medium')}>{props.label}</span>
+      <span
+        class={style('tabsTabCount', 'nt-rounded-full nt-bg-primary nt-px-[6px] nt-text-primary-foreground nt-text-sm')}
+      >
+        {count >= 100 ? '99+' : count}
+      </span>
```
LetItRock: not sure if these should be part of the `Tabs.Tab` or not

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { Tabs } from '../primitives';
+import { NotificationList } from '../Notification';
+import { MoreTabsDropdown } from './MoreTabsDropdown';
+import { InboxTab } from './InboxTab';
+
+type InboxTabsProps = {
+  tabs: Array<{ label: string; value: Array<string> }>;
+};
+
+export const InboxTabs = (props: InboxTabsProps) => {
```
LetItRock, line 19: the stateful component for tabs

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
```diff
@@ -0,0 +1,76 @@
...
+        setDropdownTabs(props.tabs.filter((tab) => !visibleTabIds.includes(tab.label)));
+        observer.disconnect();
+      },
+      { root: tabsListEl }
+    );
+
+    for (const tabElement of tabs) {
+      observer.observe(tabElement);
+    }
+  });
```
LetItRock: Calculate which tabs are fully visible in the parent container. The ones that are not visible will be shown in the dropdown.
To have enough space to render the dropdown we remove one tab from the visible tabs list in line 35;

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <Tabs class="nt-flex-1 nt-overflow-hidden" value={activeTab} onChange={setActiveTab}>
+      <Show
+        when={visibleTabs().length > 0}
+        fallback={
+          <Tabs.List ref={setTabsList}>
+            {props.tabs.map((tab) => (
+              <InboxTab label={tab.label} class="nt-invisible" />
+            ))}
+          </Tabs.List>
```
LetItRock: These tabs are rendered hidden at the beginning (otherwise there would be a blink). Then we calculate which ones are fully visible and don't overflow in the parent container. Later we render the visible tabs.

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
```diff
@@ -0,0 +1,76 @@
...
+          ))}
+          <Show when={dropdownTabs().length > 0}>
+            <MoreTabsDropdown dropdownTabs={dropdownTabs} activeTab={activeTab} setActiveTab={setActiveTab} />
+          </Show>
+        </Tabs.List>
+      </Show>
+      {props.tabs.map((tab) => (
+        <Tabs.Panel value={tab.label} class="nt-flex-1 nt-overflow-hidden">
+          <NotificationList options={{ tags: tab.value }} />
+        </Tabs.Panel>
```
LetItRock: Question: should we render only the active tab? currently, we do render all which results in X requests.

packages/js/src/ui/components/InboxTabs/MoreTabsDropdown.tsx
```diff
@@ -0,0 +1,43 @@
...
+  dropdownTabs: Accessor<Array<{ label: string; value: Array<string> }>>;
+  activeTab: Accessor<string>;
+  setActiveTab: Setter<string>;
+}) => {
+  const style = useStyle();
+
+  return (
+    <Dropdown.Root fallbackPlacements={['bottom', 'top']} placement="bottom-start">
+      <Dropdown.Trigger
+        class={style('actions__dropdownTrigger')}
```
LetItRock: I've renamed the `moreActions` to `actions` just to be able to reuse the same appearance keys. 
IDK if it's ok, ideally we should have a flexible component that we can reuse in all places, but that would require more refactoring.

packages/js/src/ui/components/primitives/Popover/PopoverContent.tsx
```diff
@@ -6,7 +6,7 @@ import { Root } from '../../elements';
 import { usePopover } from './PopoverRoot';
 
 export const popoverContentVariants = () =>
-  'nt-flex-col nt-gap-4 nt-h-[600px] nt-min-w-[400px] nt-rounded-xl nt-bg-background nt-shadow-[0_5px_15px_0_rgba(122,133,153,0.25)] nt-z-10 nt-cursor-default nt-flex nt-flex-col nt-overflow-hidden';
+  'nt-flex-col nt-h-[600px] nt-w-[400px] nt-rounded-xl nt-bg-background nt-shadow-[0_5px_15px_0_rgba(122,133,153,0.25)] nt-z-10 nt-cursor-default nt-flex nt-flex-col nt-overflow-hidden';
```
LetItRock: From one side the width should be restricted to 400px, from another angle it should be dictated by the content of the notification list item. I'm not sure 🤷‍♂️

packages/js/src/ui/context/AppearanceContext.tsx
```diff
@@ -69,14 +75,13 @@ export const appearanceKeys = [
...
-  //More actions
-  'moreActionsContainer',
-  'moreActions__dropdownTrigger',
-  'moreActions__dropdownContent',
-  'moreActions__dropdownItem',
-  'moreActions__dropdownItemLabel',
-  'moreActions__dropdownItemLabelContainer',
-  'moreActions__dropdownItemLeftIcon',
+  //Actions dropdown
+  'actionsContainer',
```
LetItRock: check the comment in `MoreTabsDropdown`

packages/js/src/ui/components/primitives/Tabs/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,61 @@
...
+        return;
+      }
+
+      setKeyboardNavigation(Array.from(tabs).includes(document.activeElement));
+    };
+
+    document.addEventListener('keyup', handleTabKey);
+
+    return onCleanup(() => document.removeEventListener('keyup', handleTabKey));
+  });
```
LetItRock, line 31: listen to the `tab` keyboard events and check if `document.activeElement` is the "tab dom element", if yes allow keyboard navigation with `ArrowLeft` or `ArrowRight`

packages/js/src/ui/components/primitives/Tabs/useKeyboardNavigation.ts
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      tabElements[activeIndex].focus();
+      setActiveTab(newTab);
+    };
+
+    document.addEventListener('keydown', handleArrowKeys);
+
+    return onCleanup(() => document.removeEventListener('keydown', handleArrowKeys));
+  });
```
LetItRock: the keyboard arrow keys navigation between the tabs; calculations for the next tab to focus and set active tab;

packages/js/src/ui/components/InboxTabs/MoreTabsDropdown.tsx
desiprisg: ⚠️ This is not ok, `moreActions__dropdownTrigger` is a more specific descriptor reserved for that dropdown trigger. You should use something like `moreTabs__dropdownTrigger` here. You can target both with `dropdownTrigger`, as that will be applied to both.

Also, you should use the `appearanceKey` prop here as this `class` attribute will be overriden by the class of the button that you specified in `asChild`. Not 100% sure, but make sure to double check that the correct appearance keys and styles are applied.

packages/js/src/ui/components/InboxTabs/MoreTabsDropdown.tsx
```diff
@@ -0,0 +1,50 @@
+import { Accessor, Setter } from 'solid-js';
+import { cn, useStyle } from 'src/ui/helpers';
+import { Check, DotsMenu } from '../../icons';
+import { Button, Dropdown, dropdownItemVariants, Tabs } from '../primitives';
+
+export const MoreTabsDropdown = (props: {
+  dropdownTabs: Accessor<Array<{ label: string; value: Array<string> }>>;
```
desiprisg: ☑️ No need for this to be an `Accessor`. Props are reactive by default.

packages/js/src/ui/components/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,62 @@
+import { Accessor, createContext, createEffect, createSignal, ParentProps, Setter, useContext } from 'solid-js';
+import { useStyle } from '../../../helpers';
+import { TabsList } from './TabsList';
+import { TabsPanel } from './TabsPanel';
+import { TabsTab } from './TabsTab';
+import { useKeyboardNavigation } from './useKeyboardNavigation';
+
+type TabsProps = ParentProps & {
+  defaultValue?: string;
+  value?: Accessor<string>;
```
desiprisg: ☑️ No need for this to be an `Accessor`.

packages/js/src/ui/components/primitives/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,62 @@
...
+  onChange?: (value: string) => void;
+  class?: string;
+};
+
+type TabsContextProps = {
+  activeTab: Accessor<string>;
+  setActiveTab: Setter<string>;
+  visibleTabs: Accessor<string[]>;
+  setVisibleTabs: Setter<string[]>;
+};
```
desiprisg: ☑️ No need for Accessors here, as these are props.

packages/js/src/ui/components/primitives/Tabs/TabsPanel.tsx
```diff
@@ -0,0 +1,24 @@
...
+  class?: string;
+};
+
+export const TabsPanel = (props: TabsPanelProps) => {
+  const style = useStyle();
+  const { activeTab } = useTabsContext();
+
+  return (
+    <div
+      class={style('tabsPanel', `${activeTab() === props.value ? 'nt-block' : 'nt-hidden'} ${props.class ?? ''}`)}
```
desiprisg: ☑️Please make use of the `cn` helper here to make this more readable.
https://www.npmjs.com/package/clsx

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { activeTab, setActiveTab } = useTabsContext();
+  const clickHandler = () => setActiveTab(props.value);
+
+  return (
+    <Button
+      variant="unstyled"
+      size="none"
+      ref={props.ref}
+      id={props.value}
+      class={style(
```
desiprisg: ☑️ `cn` will clear this up a lot as well.

packages/js/src/ui/icons/Check.tsx
```diff
@@ -1,6 +1,6 @@
-export const Check = () => {
+export const Check = (props: { class?: string }) => {
   return (
-    <svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" viewBox="0 0 20 20">
+    <svg class={props.class} xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" viewBox="0 0 20 20">
```
desiprisg: 🤓 You can just spread the props here I think.

packages/js/src/ui/novuUI.tsx
```diff
@@ -91,8 +92,8 @@ export class NovuUI {
   }
 
   //All in one <Inbox />
-  mountInbox(element: MountableElement) {
-    this.#mountComponent({ name: 'Inbox', element });
+  mountInbox(element: MountableElement, props?: Pick<InboxProps, 'tabs'>) {
```
desiprisg: ❓ Don't we need all the props here?

desiprisg (commented):
Generally, I'd prefer it if we implemented the tabs more the `<Popover>`, as we could set the default tabs styles in a function (like `popoverContentVariants()`), and default to those, unless another `class` prop is specified. See `PopoverContent.tsx` for reference.

We only need one instance of Tabs for now, so I'd say this could also be a cleanup ticket post-release.

packages/js/src/ui/components/InboxTabs/MoreTabsDropdown.tsx
LetItRock: @desiprisg Visually these two dropdowns look the same that's why I thought about using the same appearance key. We may need to extract it to a separate component and reuse it in both places, then it will have a better sense.

desiprisg: Appearance keys are not really about things looking the same, but more about allowing developers to target them specifically. It does not need to be a different component for us. `actions` does not describe the element here. `moreActions` was about the extra actions you can access via the first dropdown that we added. This is about tabs.

packages/js/src/ui/components/primitives/Popover/PopoverContent.tsx
desiprisg: I'd say content of list

packages/js/src/ui/components/InboxTabs/MoreTabsDropdown.tsx
LetItRock: But you want to target the same components (visually same) with one appearance prop. I want to adjust all my "..." dropdown styles to look the same. That's why I considered extracting it as a separate component and allowing styles to be adjusted separately. 
Like think about three levels here: 
1) dropdown
2) specific "..." dropdown
3) more specific "more actions" / "tabs actions" dropdowns

desiprisg: Yeah you can do that, you always separate those with `__` and pass the most specific to the element. Just make sure the names make sense.

packages/js/src/ui/components/primitives/Popover/PopoverContent.tsx
LetItRock: yes, I thought the same, but without the restrictions on the Inbox Popover width, the tabs component will enlarge its width :/

packages/js/src/ui/components/primitives/Tabs/Tabs.tsx
LetItRock: in terms of context value, you need to pass the reactive props like that, otherwise it doesn't work

packages/js/.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   extends: ['../../.eslintrc.js'],
-  plugins: ['local-rules'],
+  plugins: ['import', 'promise', '@typescript-eslint', 'prettier', 'local-rules'],
```
LetItRock, line 3: fixed the eslint + prettier configuration for the package, had to move local-rules file also to root

packages/js/src/ui/components/InboxTabs/InboxTab.tsx
desiprisg: Generally, they should, but in our case, we'd only do it for the appearance keys. When we add an element, we are forced to add an appearance key anyway, and more specific that the primitive one. This means that we would redeclare one anyway when using `Tab.Label` for example. I'm fine with them not being actual components.

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export const InboxTabs = (props: InboxTabsProps) => {
+  const [activeTab, setActiveTab] = createSignal<string>((props.tabs[0] && props.tabs[0].label) ?? '');
+  const [tabsList, setTabsList] = createSignal<HTMLDivElement>();
+  const [visibleTabs, setVisibleTabs] = createSignal<InboxTabsProps['tabs']>([]);
+  const [dropdownTabs, setDropdownTabs] = createSignal<InboxTabsProps['tabs']>([]);
+
+  onMount(() => {
+    const tabsListEl = tabsList();
+    if (!tabsListEl) return;
```
desiprisg: I think `tabsListEl` is set here https://docs.solidjs.com/concepts/refs#refs-in-solid

packages/js/src/ui/components/elements/InboxStatus/InboxStatusOptions.tsx
```diff
@@ -53,10 +53,7 @@ export const StatusItem = (props: {
   const style = useStyle();
 
   return (
-    <Dropdown.Item
-      class={style('inboxStatus__dropdownItem', cn(dropdownItemVariants(), 'nt-flex nt-gap-8'))}
-      onClick={props.onClick}
-    >
+    <Dropdown.Item appearanceKey="inboxStatus__dropdownItem" class="nt-flex nt-gap-8" onClick={props.onClick}>
```
desiprisg: How come we don't need the `dropdownItemVariants()`  here anymore?

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
```diff
@@ -0,0 +1,52 @@
+import { JSX, For } from 'solid-js';
+import { cn, useStyle } from '../../../helpers';
+import { DotsMenu } from '../../../icons';
+import { Button, Dropdown } from '../../primitives';
+
+export type Option = { label: string; leftIcon?: JSX.Element; rightIcon?: JSX.Element };
+
+const dropdownTriggerButtonVariants = () =>
```
desiprisg: Shouldn't this be in `DropdownTrigger.tsx` ?

packages/js/src/ui/components/primitives/Tabs/Tabs.tsx
desiprisg: I thought these were props that the provider accepts, can we rename this to `TabsContextValue` instead so it's more clear?

packages/js/src/ui/components/primitives/Tabs/TabsPanel.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const TabsPanel = (props: TabsPanelProps) => {
+  const style = useStyle();
+  const { activeTab } = useTabsContext();
+
+  return (
+    <Show when={activeTab() === props.value}>
+      <div
+        class={style(
+          props.appearanceKey ?? 'tabsPanel',
+          cn(activeTab() === props.value ? 'nt-block' : 'nt-hidden', props.class)
```
desiprisg: Adding a class prop to this means that another appearance key will be added from this component's usage. However, that extra key will not be placed in the correct spot in the class string.

```
<TabsPanel class={style('test', 'nt-p-10')}
```
would result in `tabsPanel nt-block test nt-p-10`. This means that we need to only use the appearanceKey prop for this but it's easy to use this the wrong way. Maybe it's better to introduce a utility inside `style()` which will sort the classnames based on which classes exist on `appearanceKeys` and get rid of the `appearanceKey` props altogether. Let's leave this for another PR, just sharing my thoughts.

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <Button
+      variant="unstyled"
+      size="none"
+      ref={props.ref}
+      id={props.value}
+      appearanceKey={props.appearanceKey ?? 'tabsTab'}
+      class={cn(
+        `nt-relative nt-transition nt-outline-none focus-visible:nt-outline-none focus-visible:nt-ring-2 focus-visible:nt-ring-primary focus-visible:nt-ring-offset-2 nt-pb-[0.625rem] after:nt-absolute after:nt-content-[''] after:nt-bottom-0 after:nt-left-0 after:nt-w-full after:nt-h-[2px] after:nt-border-b-2`,
+        activeTab() === props.value
```
desiprisg: Instead of using a signal here, let's just add a data attribute (`data-selected`) when this is the active tab and use the selector to style! This will also enable developer to target just the active tab via CSS.

apps/api/src/metadata.ts
desiprisg, line 1: I think this is not meant to be commited

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
LetItRock: I'm not sure what you are trying to say :/

apps/api/src/metadata.ts
LetItRock: It is not when we don't touch any API app files, but I don't know why it regenerated this file for me when running the server locally.

packages/js/src/ui/components/elements/InboxStatus/InboxStatusOptions.tsx
LetItRock: It's an internal detail of the `Dropdown.Item` component. And you can see that we are now using the `appearanceKey` and `class` props separately.

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
LetItRock: It makes sense, but we are using here the `asChild` prop so I would need to export it and use here as well

packages/js/src/ui/components/primitives/Tabs/Tabs.tsx
LetItRock: sure ;)

packages/js/src/ui/components/primitives/Tabs/TabsPanel.tsx
LetItRock: I think that in the case when we override some component styles we should use the `appearanceKey` and `class` props separately. This will ensure that they are put in the right order.

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
LetItRock: There is the `aria-selected` attribute that could be used for that.

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
desiprisg: Yes

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
desiprisg: I think there's not need to check for it being undefined

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
desiprisg: I think we need to be using data attributes, as that's what's most commonly used in these cases:
Take this as an example: https://www.radix-ui.com/primitives/docs/components/tabs#trigger
`data-state` could work for both trigger and content.

We should also be using the `data` attribute in our CSS to enforce this.

`aria` attributes are an implementation detail of our components and should not be the attributes customers use to style.

packages/js/src/ui/components/elements/InboxStatus/InboxStatusOptions.tsx
desiprisg: It's not an internal detail. It's the default styles for the dropdown item. You can import them and use them if you want to extend its styles, or omit it if you want to use it unstyled.

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
desiprisg: Not sure what you mean, but the idea behind this function is that you can use it when you want to apply the base styles for that and extend.

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
LetItRock: the type is `HTMLDivElement | undefined` so I am forced to do that ;)

packages/js/src/ui/components/elements/InboxStatus/InboxStatusOptions.tsx
LetItRock: Why do I have to do that if I just want to update some styles?

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
LetItRock: What I'm saying is that I would have to apply it in the dropdown trigger, right? 
And then because we render the `Button` component with `asChild` prop I would have to update the `Button class` prop here as well. Check how the code looks now.

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
LetItRock: yeah for the better DX, we can support both ;)

packages/js/src/ui/components/primitives/Popover/PopoverContent.tsx
BiswaViraj: I feel we should have width constraints on the Popover and the let the notification content inherit the parents width

BiswaViraj: Letting content dictate the popover's width can make the popover grow out of bounds of the viewport. ideally content should be  responsive enough to be able render in all sizes of screen and also as full page solutions.
which will be possible if we inherit width from parent and grow/shrink with parent width changes

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
```diff
@@ -0,0 +1,49 @@
+import { JSX, For } from 'solid-js';
+import { cn, useStyle } from '../../../helpers';
+import { DotsMenu } from '../../../icons';
+import { Button, Dropdown, dropdownTriggerButtonVariants } from '../../primitives';
+
+export type Option = { label: string; leftIcon?: JSX.Element; rightIcon?: JSX.Element };
+
+export const OptionsDropdown = (props: {
```
desiprisg: In general, I would avoid this type of component here. If we need variations of dropdowns, we should be defining those where we need them. Why make a large and complicated component?

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <Dropdown.Root fallbackPlacements={['bottom', 'top']} placement={props.placement ?? 'bottom-start'}>
+      <Dropdown.Trigger
+        appearanceKey="optionsDropdown__dropdownTrigger"
+        asChild={(triggerProps) => (
+          <Button
+            variant="unstyled"
+            size="none"
+            appearanceKey="optionsDropdown__button"
+            {...triggerProps}
+            class={cn(dropdownTriggerButtonVariants(), props.buttonClass)}
```
desiprisg: Continuing the other discussion here. Is `dropdownTriggerButtonVariants()` not applied by `...triggerProps` ? Do we do this to just extend with `props.buttonClass`. In my mind, this isn't a good practice. Why have this component at all? We should just use the primitive where we need it.

packages/js/src/ui/components/InboxTabs/InboxTabs.tsx
SokratisVidros: Yes, only the active tab the first time

packages/js/src/ui/components/Renderer.tsx
```diff
@@ -20,7 +20,7 @@ export const novuComponents = {
   Bell,
 };
 
-export type NovuComponent = { name: NovuComponentName; props?: unknown };
+export type NovuComponent = { name: NovuComponentName; props?: any };
```
SokratisVidros, line 23: Please add stronger types as the union of the props of all available components.

packages/js/src/ui/components/elements/InboxStatus/InboxStatusOptions.tsx
SokratisVidros: If not mistaken customization comes from `style` function. Without it everything else is hardcoded, right?

packages/js/src/ui/components/elements/OptionsDropdown/OptionsDropdown.tsx
SokratisVidros: I agree with @desiprisg. We must leverage composition at its maximum. This is the main problem we have with the current design system in the dashboard, making it very hard to use in the long run. In practise, we should follow the Radix/ShadUI pattern for the internal UI library of the Inbox component.

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
```diff
@@ -0,0 +1,43 @@
+import { JSX, ParentProps, Ref } from 'solid-js';
```
SokratisVidros: Maybe just `Tab.tsx`? The stuttering feels a bit weird.

SokratisVidros (changes requested):

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
desiprisg: We can do `TabRoot`, `TabList`, `TabTrigger` and `TabContent`

packages/js/src/ui/components/Renderer.tsx
LetItRock: Unfortunately, this can't be determined by the places where this type is used. Other types don't work well.

packages/js/src/ui/components/primitives/Tabs/TabsTab.tsx
LetItRock: I'll follow Radix pattern: 
```
<Tabs.Root>
    <Tabs.List>
      <Tabs.Trigger />
    </Tabs.List>
    <Tabs.Content />
  </Tabs.Root>
```

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -51,11 +60,8 @@ export const Inbox = (props: InboxProps) => {
           </Button>
         )}
       />
-      <Popover.Content
-        appearanceKey="inbox__popoverContent"
-        class={cn(popoverContentVariants(), 'nt-max-w-96 nt-w-full')}
```
LetItRock: moved to the variants function

SokratisVidros (approved):

PR #6150, opened 2024-07-24
fix(web): Fix monitoring ids

denis-kralj-novu:

- Map mongo identifiers to user and org entities when possible
- Expose shouldMonitor flag for use cases when we want to choose not to
- Remove unused imports
- Remove unreferenced pages
- Add explanation of why the mapping happens


apps/web/src/components/providers/AuthProvider.tsx
```diff
@@ -32,6 +32,7 @@ type OrganizationState =
 
 export type AuthContextValue = UserState &
   OrganizationState & {
+    shouldMonitor: boolean;
```
rifont: ❓ **question:** The name `shouldMonitor` is lacking a bit of description for me right now - what we we monitoring? Can we make it `shouldMonitor<Something>`? ‏

denis-kralj-novu: This value is now only used in the `useMonitoring()` hook that based on the value identifies users/orgs with segment/launchDarkly.

In reality I don't expect this value to be used anywhere else, as it is false only in one case (when a user is in the process of signing up), so I could move the logic for its calculation directly in the hook that uses it, wdyt?

rifont: I agree, moving the logic to be co-located with it's consumer makes the most sense given that the use is limited.

denis-kralj-novu: addressed

apps/web/src/hooks/useMonitoring.ts
```diff
@@ -9,8 +9,18 @@ export function useMonitoring() {
   const ldClient = useLDClient();
   const segment = useSegment();
 
+  const isNovuUser = currentUser && currentUser._id && !currentUser._id.startsWith('user_');
```
denis-kralj-novu, line 12: checking `currentUser._id` and`currentOrganization._id` for undefined seems to be needed for e2e tests :/

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -147,27 +147,55 @@ export const EnterpriseAuthProvider = ({ children }: { children: React.ReactNode
...
+   * the internal Novu id is present.
+   */
+
+  return {
+    _id: clerkUser.externalId ?? clerkUser.id,
+    firstName: clerkUser.firstName,
+    lastName: clerkUser.lastName,
+    email: clerkUser.emailAddresses[0].emailAddress,
+    profilePicture: clerkUser.imageUrl,
+    createdAt: clerkUser.createdAt?.toString() ?? '',
```
rifont, line 170: ❓ **question:** The mapper below doesn't include the nullish coalescing operator (`??`) for the `createdAt`, is it really necessary here? When would a `clerkUser` exist without a `createdAt` property?‏

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -147,27 +147,55 @@ export const EnterpriseAuthProvider = ({ children }: { children: React.ReactNode
...
+   * When mapping to IOrganizationEntity, we have 2 cases:
+   *  - user exists and has signed in
+   *  - user is signing up
+   *
+   * In the case where the user is still signing up, we are using the clerk identifier for the id.
+   * This however quickly gets update to the .publicMetadata.externalOrgId (which is actually the novu
+   * internal entity identifier) that gets used further in the app. A few consumers that want to user
+   * this identifier before it is set to the internal value monitor the flag 'shouldMonitor' and only
+   * use them when the internal Novu id is present.
+   */
```
rifont, line 192: 👏 **praise:** Great commentary!‏

rifont (approved):

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
denis-kralj-novu: that is a good question, i will check in the docs and remove if it isn't needed

denis-kralj-novu: Indeed, there is a difference in the types, for the `UserResource` it can be set to `null`. I've asked the Clerk guys if they have an explanation for this.

PR #6156, opened 2024-07-25
chore(root): New cloud deployment CI

AliaksandrRyzhou:
- Slack notifications slack `api/slack-github-action`.
- ECS services matrix fix for worker service.
- Only manual run for testing mode.
- The number of conditions for different environments was decreased.
- Most of the variables moved to the GitHub settings for environments.
- Tree environments were created within GitHub settings, `Staging`, `Production-EU`, and `Production-US`. Before you run the deployment for the Prod you have to create similar variables and secrets as inside of the `Staging` environment.
- Some part of variables can be simplified but I leaved it for a flexibility in first version of cloud deployment.


.github/workflows/cloud-deploy-embed.yml
```diff
@@ -0,0 +1,51 @@
...
+      packages: write
+      deployments: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-project
+        with:
+          slim: 'true'
+
+      - name: Build
```
SokratisVidros: ❓ Do we need both build:embed and build:prod steps? Are they doing different things? cc: @LetItRock

.github/workflows/cloud-deploy-reusable-app-service-deploy.yml
```diff
@@ -0,0 +1,217 @@
...
+        run: |
+          echo "ecs_container_name=$(terraform output -json ${{ env.SERVICE_NAME }}_ecs_container_name | jq -r .)" >> $GITHUB_OUTPUT
+          echo "ecs_service=$(terraform output -json ${{ env.SERVICE_NAME }}_ecs_service | jq -r .)" >> $GITHUB_OUTPUT
+          echo "ecs_cluster=$(terraform output -json ${{ env.SERVICE_NAME }}_ecs_cluster | jq -r .)" >> $GITHUB_OUTPUT
+          echo "task_name=$(terraform output -json ${{ env.SERVICE_NAME }}_task_name | jq -r .)" >> $GITHUB_OUTPUT
+          echo "aws_region=$(terraform output -json aws_region | jq -r .)" >> $GITHUB_OUTPUT
+          echo "queue_workers_services=$(terraform output -json queue_workers_services)" >> $GITHUB_OUTPUT
+
+  deploy_service_worker:
+    if: inputs.service_name == 'worker'
```
SokratisVidros: @AliaksandrRyzhou Please add a comment why workers need to be treated separately for now. Actually, is this still the case or can we remove this conditional since we fetch the latest ECS definition from AW?S

.github/workflows/cloud-deploy-reusable-app-service-deploy.yml
```diff
@@ -0,0 +1,217 @@
...
+          environment: prod
+          ignore_empty: true
+          ignore_missing: true
+
+      - name: Set New Relic Region
+        if: ${{ inputs.New_Relic_GUID }}
+        id: set_region
+        run: |
+          if [[ "${{ env.aws_region }}" == *"us"* ]]; then
+            echo "New_Relic_Region=US" >> ${GITHUB_ENV}
```
SokratisVidros: Please convert this to a NEW_RELIC_REGION env variable to remove this conditional based on region

.github/workflows/cloud-deploy-services.yml
```diff
@@ -0,0 +1,53 @@
...
+        description: 'Service port for testing'
+        required: true
+        type: string
+      local_tag:
+        description: 'Local Docker image tag'
+        required: true
+        type: string
+
+jobs:
+  build_image:
```
SokratisVidros:
```suggestion
  build_docker_image:
```

.github/workflows/cloud-deploy-services.yml
```diff
@@ -0,0 +1,53 @@
...
+    with:
+      environment: ${{ inputs.environment }}
+      package_name: ${{ inputs.package_name }}
+      project_path: ${{ inputs.project_path }}
+      test_port: ${{ inputs.test_port }}
+      health_check: true
+      local_tag: ${{ inputs.local_tag }}
+      enterprise: true
+
+  deploy_service:
```
SokratisVidros:
```suggestion
  deploy_to_aws:
```

SokratisVidros (approved):
This is the final round of comments. Good work Alex!

.github/workflows/cloud-deploy-reusable-app-service-deploy.yml
AliaksandrRyzhou: If we take a look a difference is not too big here. 
Screenshot from 2024-07-26 15-52-04

The worker service only has a few services inside of ECS cluster and so that to get a task definition for each of them and deploy each of them we have to receive all these names to build a matrix by using the terraform the output values of `queue_workers_services`. We get the services, build matrix for each of them and update the TD for each of them. So that we can avoid using a two jobs we just need to create a terraform output for each service and leave the job with matrix based on a queue of each service, in any another case we use conditions somewher in GitHub Actions.

PR #6160, opened 2024-07-25
fix(framework): Remove only failing validation properties and simplify Slack schema

rifont:

* Add tests for polymorphic schemas
* Remove only failing validation properties in JsonSchema validator
* simplify Slack schema to reduce bundle size.
  * Aside from runtime validation, these Slack blocks definitions were not actually providing any DX value due to use of the `if`/`then` style of JsonSchem syntax. We will need to revisit adding type support for Slack blocks by using explicitly generated Typescript typings from the Slack blocks JsonSchema to reduce the burden on Typescripts inference engine.

### Screenshots

<details>

</details>


packages/framework/src/validators/json-schema.validator.ts
```diff
@@ -20,7 +20,7 @@ export class JsonSchemaValidator implements Validator<JsonSchema> {
       // https://ajv.js.org/options.html#usedefaults
       useDefaults: true,
       // https://ajv.js.org/options.html#removeadditional
-      removeAdditional: 'all',
+      removeAdditional: 'failing',
```
rifont, line 23: This was the important change to ensure that specifying `additionalProperties: true` on the schema is acted upon, maintaining the additional properties. Previously, all additional properties were being removed, ignoring this schema property.

packages/framework/src/types/execution.types.ts
```diff
@@ -36,7 +37,7 @@ export type ExecuteOutputOptions = {
 
 export type ExecuteOutput = {
   outputs: Record<string, unknown>;
-  providers?: Record<string, Record<string, unknown>>;
+  providers?: Record<string, WithPassthrough<Record<string, unknown>>>;
```
rifont, line 40: Adding a missing type here.

packages/framework/src/validators/validator.test.ts
```diff
@@ -40,7 +40,7 @@ describe('validators', () => {
         title: 'should remove additional properties and successfully validate',
         schemas: {
           zod: z.object({ name: z.string() }),
-          json: { type: 'object', properties: { name: { type: 'string' } } } as const,
+          json: { type: 'object', properties: { name: { type: 'string' } }, additionalProperties: false } as const,
```
rifont, line 43: Necessary to fix the failing test after modifying the `removeAdditional` property in the JsonSchema validator.

davidsoderberg (approved):

PR #6164, opened 2024-07-26
fix(web): remove duplicate title and description

denis-kralj-novu:
### Screenshots
image
left is old state, rights is new

image


denis-kralj-novu:
we can also change the sign up page to look like the example on the right, currently it looks like the variant on the right. Let me know if this is desired
image

ChmaraX (approved):

PR #6167, opened 2024-07-29
feat(react): Com 40 create the novureact package

BiswaViraj:

- Created the `@novu/react` package
- Exported Components
  - Inbox
  - Bell
  - Notifications
  - Settings

### Screenshots

<details>

</details>


packages/js/src/ui/components/elements/Bell/Bell.tsx
```diff
@@ -1,18 +1,19 @@
...
 import { BellContainer } from './DefaultBellContainer';
 
 type BellProps = {
-  //TODO: convert this in a `mountBell` prop like we do for notifications.
-  children?: ({ unreadCount }: { unreadCount: Accessor<number> }) => JSX.Element;
+  mountBell?: BellMounter;
 };
 /* This is also going to be exported as a separate component. Keep it pure. */
-export function Bell(props: BellProps) {
+export const Bell = (props: BellProps) => {
```
BiswaViraj: Having it as a function breaks the TS for some reason.
it was not inferring the props type in the React package

antonjoel82: ✏ **suggestion:** ‏Please consider using explicit types per the [Solid TypeScript guide](https://www.solidjs.com/guides/typescript#component-types)

packages/react/src/components/Renderer.tsx
```diff
@@ -0,0 +1,59 @@
...
+import ReactDOM from 'react-dom';
+import { MountedElement, RendererProvider } from '../context/RenderContext';
+
+type RendererProps = React.PropsWithChildren<{
+  options: NovuUIOptions;
+}>;
+
+export const Renderer = (props: RendererProps) => {
+  const { options, children } = props;
+  const [novuUI, setNovuUI] = React.useState<NovuUI | undefined>();
```
antonjoel82, line 17: ❓ **question:** Should this be a `ref` instead perhaps?‏

antonjoel82 (commented):
Just doing a quick skim :)

I would recommend using types as provided by Solid rather than relying on inference (both for performance and being intentional / explicit).

Additionally, adding `/** */` comments to important constructs with explanations of their purpose and use cases would be great (i.e. Renderer, Mounter, etc)

packages/react/tsup.config.ts
```diff
@@ -0,0 +1,15 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  entry: ['src/index.ts'],
+  treeshake: true,
+  sourcemap: false,
+  minify: true,
```
LetItRock: ☑ **chore:** we need to generate the source maps, but don't have to minify the code...

packages/react/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "include": ["src"],
+  "compilerOptions": {
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "target": "ES6",
```
LetItRock:
```suggestion
    "target": "ESNext",
```

packages/react/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
LetItRock, line 1: 🗒 **note (non-blocking):** I'm not sure if we need to extend from the base like we do in all the other packages :/ wdyt?

packages/react/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+  "include": ["src"],
+  "compilerOptions": {
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "target": "ES6",
+    "allowJs": true,
```
LetItRock:
```suggestion
```

packages/react/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "skipLibCheck": true,
+    "strict": true,
+    "noEmit": true,
```
LetItRock: 🗒 **note (non-blocking):** It doesn't align with the tsup config, I guess that we can remove it from here, but please check

packages/react/tsup.config.ts
```diff
@@ -0,0 +1,15 @@
...
+  clean: true,
+  dts: true,
+  splitting: false,
+  format: ['cjs', 'esm'],
+  external: ['react'],
```
LetItRock:
```suggestion
  external: ['react', 'react-dom'],
```

packages/react/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "esModuleInterop": true,
+    "module": "esnext",
+    "isolatedModules": true,
+    "jsx": "preserve",
+    "forceConsistentCasingInFileNames": true,
+    "moduleResolution": "node"
```
LetItRock: `node` is not suggested for the new projects, [ref](https://www.typescriptlang.org/docs/handbook/modules/theory.html#:~:text=It%20should%20not%20be%20used,is%20scheduled%20for%20deprecation%20in)
I guess we can't use `bundler` in ts < 5 so `nodenext`?

packages/js/src/ui/components/elements/Bell/Bell.tsx
```diff
@@ -1,18 +1,19 @@
-import { Accessor, JSX } from 'solid-js';
+import { Show } from 'solid-js';
+import { BellMounter } from 'src/ui/types';
```
LetItRock: ☑ **chore:** please update the import

packages/react/package.json
```diff
@@ -0,0 +1,55 @@
...
+      "import": "./dist/index.mjs"
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "sideEffects": false,
+  "private": true,
+  "scripts": {
+    "dev": "pnpm build --watch",
```
LetItRock:
```suggestion
    "build:watch": "pnpm build --watch",
```
to be consistent

packages/react/src/components/Notifications.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { useRenderer } from '../context/RenderContext';
+import { Mounter } from './Mounter';
+import type { InboxNotification } from '@novu/js';
```
LetItRock: 🤓 **nitpick (non-blocking):** please update the order of imports

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
+import type { InboxNotification } from '@novu/js';
+import type { BaseNovuUIOptions } from '@novu/js/ui';
+import React from 'react';
+import { useRenderer } from '../context/RenderContext';
+
```
LetItRock: 🤓 **nitpick (non-blocking):** please update the order of imports and remove empty line

playground/nextjs/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@novu/js": "workspace:*",
```
LetItRock: 🤓 **nitpick (non-blocking):** I guess this import is not needed anymore?

playground/nextjs/src/pages/index.tsx
```diff
@@ -1,4 +1,4 @@
-import { Inbox } from '@/components/Inbox';
+import { Bell, Inbox, Settings } from '@novu/react';
```
LetItRock: ✏ **suggestion:** I suggest having a few dedicated pages where we can test different rendering approaches separately. I see that we update this file often with different configs. 
Can we have at least: 
- one liner
- custom bell
- custom notification item

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+        children: React.ReactNode;
+      }
+  );
+
+export const Inbox = (props: InboxProps) => {
+  if (props.children) {
+    return <Renderer options={props}>{props.children}</Renderer>;
+  }
+
+  const { renderNotification, ...options } = props as BaseNovuUIOptions & InboxDefaultProps;
```
LetItRock:
```suggestion
  const { renderNotification, renderBell, ...options } = props;
```

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+      } & InboxDefaultProps)
+    | {
+        children: React.ReactNode;
+      }
+  );
+
+export const Inbox = (props: InboxProps) => {
+  if (props.children) {
+    return <Renderer options={props}>{props.children}</Renderer>;
+  }
```
LetItRock:
```suggestion
   if ('children' in props) {
    const { children, ...options } = props;

    return <Renderer options={options}>{children}</Renderer>;
  }
```

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const { renderNotification, ...options } = props as BaseNovuUIOptions & InboxDefaultProps;
+
+  return (
+    <Renderer options={options}>
+      <InboxDefault renderNotification={renderNotification} renderBell={options.renderBell} />
```
LetItRock:
```suggestion
      <InboxDefault renderNotification={renderNotification} renderBell={renderBell} />
```

packages/js/src/ui/components/Renderer.tsx
```diff
@@ -12,12 +12,15 @@ import {
...
 import { UnreadCountProvider } from '../context/UnreadCountContext';
-import { Bell, Root } from './elements';
+import { Bell, Root, Settings } from './elements';
 import { Inbox } from './Inbox';
+import { NotificationList } from './Notification';
 
 export const novuComponents = {
   Inbox,
   Bell,
+  Settings,
```
SokratisVidros:
```suggestion
  Preferences,
```

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const { renderNotification, renderBell, ...rest } = props;
+  const { novuUI, mountElement } = useRenderer();
+
+  const mount = React.useCallback(
+    (element: HTMLElement) => {
+      return novuUI.mountComponent({
+        name: 'Inbox',
+        props: {
+          ...rest,
+          mountNotification: renderNotification
```
SokratisVidros, line 18: Should this prop be renderNotification?
```suggestion
          mountNotification: renderNotification
```

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+            ? (el, { notification }) => {
+                return mountElement(el, renderNotification(notification));
+              }
+            : undefined,
+          mountBell: renderBell
```
SokratisVidros: Ditto

```suggestion
          renderBell
```

packages/react/src/components/Renderer.tsx
SokratisVidros: I think so too.

SokratisVidros (changes requested):

packages/react/src/components/Renderer.tsx
BiswaViraj: I am dynamically switching between `null` and `components` screen below.
I think using ref will not trigger a re-render and the switching won't work

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+          renderBell: renderBell
+            ? (el, { unreadCount }) => {
+                return mountElement(el, renderBell({ unreadCount }));
+              }
+            : undefined,
+        },
+        element,
+      });
+    },
+    [renderNotification]
```
LetItRock: ✏ **suggestion:** if we care about the `renderNotification` prop changes, we should also probably care about the `renderBell` and other props

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,66 @@
...
+          renderNotification: renderNotification
+            ? (el, { notification }) => {
+                return mountElement(el, renderNotification(notification));
+              }
+            : undefined,
+          renderBell: renderBell
+            ? (el, { unreadCount }) => {
+                return mountElement(el, renderBell({ unreadCount }));
+              }
+            : undefined,
```
LetItRock:
```suggestion
            ? (el, { notification }) => mountElement(el, renderNotification(notification))
            : undefined,
          renderBell: renderBell
            ? (el, { unreadCount }) => mountElement(el, renderBell({ unreadCount }))
            : undefined,
```

packages/react/src/components/Renderer.tsx
```diff
@@ -0,0 +1,59 @@
...
+        });
+      };
+    },
+    [setMountedElements]
+  );
+
+  React.useEffect(() => {
+    const loadNovuUI = async () => {
+      const { NovuUI } = await import('@novu/js/ui');
+      const ui = new (NovuUI as new (novuOptions: NovuUIOptions) => NovuUI)(options);
```
LetItRock:
```suggestion
      const { NovuUI } = await import('@novu/js/ui');
```

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { BellMounter, NotificationMounter } from '../types';
 import { Bell, Footer, Header, Settings, SettingsHeader } from './elements';
 import { NotificationList } from './Notification';
 import { Button, Popover } from './primitives';
 
 export type InboxProps = {
   open?: boolean;
-  mountNotification?: NotificationMounter;
-  renderBell?: ({ unreadCount }: { unreadCount: Accessor<number> }) => JSX.Element;
+  renderNotification?: NotificationMounter;
```
desiprisg: ❓ Why are we changing the name here?

packages/react/src/components/Notifications.tsx
```diff
@@ -0,0 +1,26 @@
...
+type BellProps = {
+  children?: never | NotificationsRenderProps;
+};
+
+export const Notifications = (props: BellProps) => {
+  const { novuUI, mountElement } = useRenderer();
+
+  const mount = React.useCallback((element: HTMLElement) => {
+    return novuUI.mountComponent({
+      name: 'NotificationList',
```
desiprisg: ⚠️ We need to name this `Notifications`

packages/react/tsconfig.json
BiswaViraj: I don't see any value in extending it. It is a different package and will be available publicly.
@SokratisVidros wdyt?

packages/react/tsconfig.json
BiswaViraj: Ah, didn't knew `node` is not suggested.
Thanks, will update it to `nodenext`

packages/react/src/components/Inbox.tsx
BiswaViraj: It fixes the need to type assertion below, but `...options` needs the type assertion now 😞

packages/react/src/components/Renderer.tsx
BiswaViraj: Sorry, i didn't understood what I need to change here

packages/js/src/ui/components/Inbox.tsx
BiswaViraj: Wanted to align the naming throughout the app
let me know if you think we shouldn't update this

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+type WithChildrenProps = BaseProps & {
+  children: React.ReactNode;
+};
+
+type InboxProps = DefaultProps | WithChildrenProps;
+
+export const Inbox = (props: InboxProps) => {
+  if ('children' in props) {
+    const { children, ...options } = props as WithChildrenProps;
```
antonjoel82: ✏ **suggestion:** Can use a [custom type-guard](https://blog.logrocket.com/how-to-use-type-guards-typescript/#:~:text=a%20Boolean%20value.-,Custom%20type%20guards,-TypeScript%20custom%20type) here to avoid type assertion!‏

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,63 @@
...
+    children?: never;
+  };
+
+type WithChildrenProps = BaseProps & {
+  children: React.ReactNode;
+};
+
+type InboxProps = DefaultProps | WithChildrenProps;
+
+export const Inbox = (props: InboxProps) => {
```
antonjoel82:
```suggestion
export const Inbox = React.memo((props: InboxProps) => {
```
We want to ensure that we avoid re-renders if user's props do not change

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,63 @@
...
+type BaseProps = BaseNovuUIOptions;
+
+type DefaultProps = BaseProps &
+  InboxDefaultProps & {
+    children?: never;
+  };
+
+type WithChildrenProps = BaseProps & {
+  children: React.ReactNode;
+};
```
antonjoel82: ☑ **chore:** Please use `React`'s types rather than defining our own. If we need to define our own foundational types, please move them to a separate file so that they can be imported into other files (and this one) without risk of circular dependency‏

playground/nextjs/src/pages/render-notification/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { novuConfig } from '@/utils/config';
+import { Inbox } from '@novu/react';
+
+export default function Home() {
+  return (
+    <div className="text-center flex flex-col gap-4">
+      <p className="text-xl">Render Notification Props</p>
```
antonjoel82: 💭 **thought (non-blocking):** Seems like a `h*` element would be more semantic / accurate here. Same goes for all other playground pages‏

playground/nextjs/src/pages/render-notification/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <Inbox
+        options={novuConfig}
+        renderNotification={(notification) => {
+          return (
+            <div className="flex gap-2 flex-nowrap items-start self-stretch my-1 p-2 hover:bg-slate-200">
+              <div className="rounded-full w-8 h-8 overflow-hidden border border-cyan-200">Avatar</div>
+              <div>
+                <div className="text-xl font-bold">{notification.subject ?? 'Subject'}</div>
```
antonjoel82:
```suggestion
                <div className="text-xl font-bold">{notification.subject || 'Subject'}</div>
```
Unless we want `''` to intentionally render an empty subject

playground/nextjs/src/components/Header.tsx
```diff
@@ -0,0 +1,36 @@
...
+  { href: '/preferences', label: 'Preferences' },
+  { href: '/notifications', label: 'Notifications' },
+];
+
+const NavLink = ({ href, label }: { href: string; label: string }) => {
+  const router = useRouter();
+
+  const pathname = router.pathname;
+
+  const isActive = pathname === href;
```
antonjoel82, line 23: ✏ **suggestion:** Looks like `usePathname` from `next/navigation` may be the [recommended approach](https://blog.logrocket.com/how-to-use-type-guards-typescript/#:~:text=a%20Boolean%20value.-,Custom%20type%20guards,-TypeScript%20custom%20type)‏

playground/nextjs/src/components/Header.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const LINKS = [
+  { href: '/', label: 'Default Inbox' },
+  { href: '/render-bell', label: 'Render Bell' },
+  { href: '/render-notification', label: 'Render Notification' },
+  { href: '/preferences', label: 'Preferences' },
+  { href: '/notifications', label: 'Notifications' },
+];
+
+const NavLink = ({ href, label }: { href: string; label: string }) => {
```
antonjoel82: 🤓 **nitpick (non-blocking):** Extract type and use as typing for `LINKS` array‏

antonjoel82 (commented):
Nice work, thanks for the updates!

Some additional thoughts:
- for external-facing code, we should be diligent about using explicit types (and especially use types built-in to the target library / framework -- such as React)
- I'd suggest we use `React.memo` more liberally in exported components. We want to ensure we're not re-rendering constantly / unnecessarily, especially if the caller memoizes (uses useMemo / useCallback) with their props. I know memoization is not free (costs memory), but if someone uses our components and passes memoized props, they should reasonably expect things not to re-render.

playground/nextjs/src/components/Header.tsx
BiswaViraj: I think it is limited to `app router` only.
The playground here uses the `pages router`

playground/nextjs/src/pages/render-notification/index.tsx
BiswaViraj: Agree 👍🏼

packages/js/src/ui/components/Inbox.tsx
desiprisg: Yeah I think it's better to leave as `mountNotification`, In the JS level, it is a mount method, just like the `mountInbox`.

packages/react/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@novu/react",
+  "version": "2.0.0-canary.0",
```
desiprisg, line 3: We don't want 2.0.0 here, right?

packages/react/src/components/Inbox.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { renderNotification, renderBell, ...options } = props;
+
+  return (
+    <Renderer options={options}>
+      <InboxDefault renderNotification={renderNotification} renderBell={renderBell} />
+    </Renderer>
+  );
+});
+
+function isWithChildrenProps(props: InboxProps): props is WithChildrenProps {
```
desiprisg, line 48: ❓  Is there any need for this function?

packages/react/src/components/Notifications.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+  const mount = React.useCallback((element: HTMLElement) => {
+    return novuUI.mountComponent({
+      name: 'Notifications',
+      element,
+      props: props.children
+        ? { mountNotification: (el, { notification }) => mountElement(el, props.children?.(notification)) }
+        : undefined,
+    });
+  }, []);
```
desiprisg: ❓  Should we have dependencies here? I think we need to.

packages/react/src/components/Preferences.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { useRenderer } from '../context/RenderContext';
+import { Mounter } from './Mounter';
+
+const PreferencesDefault = () => {
```
desiprisg: ❓ Why not just "Preferences" ?

packages/react/src/utils/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { InboxNotification } from '@novu/js';
+import type { BaseNovuUIOptions } from '@novu/js/ui';
+
+export type Notification = InboxNotification;
+export type NotificationsRenderProps = (notification: Notification) => React.ReactNode;
+
+export type InboxDefaultProps = {
```
desiprisg: ❓  Can we name this "DefaultInboxProps" as "Default" specified the inbox and not the props?

playground/nextjs/src/components/Header.tsx
desiprisg: We will be adding a folder for app router as well, so we'll be using it there!

packages/react/src/components/Inbox.tsx
BiswaViraj: This is for type-guard
https://github.com/novuhq/novu/pull/6167#discussion_r1697627966

packages/react/src/utils/types.ts
BiswaViraj: ~Sorry, didn't understood what do you mean by Default~
Got it 👌🏼

packages/react/package.json
BiswaViraj: afaik, it gets overridden by our script during the release and we maintain a same release version throughout the monorepo.
@LetItRock can we have different release number here and will it increment independently of the monorepo version when we release builds?

packages/react/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './components';
+export * from './utils/types';
```
BiswaViraj, line 2: export the types as well

packages/react/src/utils/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { InboxNotification } from '@novu/js';
+import type { BaseNovuUIOptions } from '@novu/js/ui';
+
+export type Notification = InboxNotification;
```
LetItRock: ☑ **chore:** We shouldn't use the `InboxNotification` for the public interface, please update it to the `Notification` class. Please check the other places too.

packages/js/src/ui/components/Inbox.tsx
LetItRock: So what we can do is to have `renderNotification, renderBell` as the props on the `react` package, but map it internally to the `mountNotification, mountBell` functions in the JS SDK.

packages/react/package.json
LetItRock: Lerna allows us to keep different versions independent. But all the packages in our monorepo as synced with the same version and I don't think that we want to break this consistency now.

packages/react/src/components/Renderer.tsx
```diff
@@ -0,0 +1,60 @@
...
+        });
+      };
+    },
+    [setMountedElements]
+  );
+
+  React.useEffect(() => {
+    // Need to use require here to not break the build during SSR
+    const { NovuUI } = require('@novu/js/ui');
+    const ui = new (NovuUI as new (novuOptions: NovuUIOptions) => NovuUI)(options) as NovuUI;
```
LetItRock:
```suggestion
    const ui: NovuUI = new NovuUI(options);
```

packages/react/src/utils/createContextAndHook.ts
```diff
@@ -27,12 +26,14 @@ export const createContextAndHook = <CtxVal>(
 
   const useCtx = () => {
     const ctx = React.useContext(Ctx);
-    assertCtxFn(ctx, `${displayName} not found`);
+    assertCtxFn(ctx, `Component must be used inside r wrapped with the <Inbox /> Component`);
```
LetItRock:
```suggestion
    assertCtxFn(ctx, `Component must be wrapped with the <Inbox />`);
```

packages/js/src/ui/components/Notification/Notification.tsx
```diff
@@ -1,11 +1,11 @@
 import { Show } from 'solid-js';
-import { InboxNotification } from '../../../types';
+import type { Notification as NotificationType } from '../../../notifications/notification';
 import { NotificationMounter } from '../../types';
 import { ExternalElementMounter } from '../ExternalElementMounter';
 import { DefaultNotification } from './DefaultNotification';
 
 type NotificationProps = {
-  notification: InboxNotification;
+  notification: NotificationType;
```
desiprisg, line 8: ❓  I think the previous type was the correct one, no?

desiprisg (approved):

packages/react/src/context/RenderContext.tsx
desiprisg, line 1: Can we name this `RendererContext` to match the component?

packages/js/src/ui/components/Notification/Notification.tsx
LetItRock: @desiprisg The old one is the pure `type`, but we want to provide the type based on `class`

LetItRock (approved):
amazing work @BiswaViraj! 🚀

SokratisVidros (approved):

PR #6170, opened 2024-07-29
feat(js): Improve style() functionality

desiprisg:

Improving the `style()` helper in this PR so that:
* We can "chain" multiple calls of this, without messing up the order of our appearanceKeys in the class list.
* No duplicates will be found in the final class list.

An example of what is now possible is
```
//Foo.tsx
export const Foo = (props) => {
  const style = useStyle();
  return <div class={style('foo', cn('nt-p-10', props.class))} />
}

//Bar.tsx
export const Bar = () => {
  const style = useStyle();
  return <Foo class={style('bar__foo', 'nt-bg-red-500')} />
}
```
The resulting class will be `bar__foo foo nt-p-10 nt-bg-red-500` instead of one appearance key appearing in front and the other one in the middle of the class string.
### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/ui/helpers/useStyle.ts
+++ b/packages/js/src/ui/helpers/useStyle.ts
@@ -12,20 +12,37 @@ export const useStyle = () => {
 
   const styleFuncMemo = createMemo(() => (appearanceKey: AppearanceKey, className?: string) => {
     const appearanceKeyParts = appearanceKey.split('__');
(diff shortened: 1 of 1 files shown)


BiswaViraj (approved):

LetItRock (approved):
looks good! 🙌

SokratisVidros (approved):

PR #6171, opened 2024-07-29
feat(web): update ee questionnaire

ChmaraX:
* added language property to Clerk organization metadata to reflect community questionnaire fields
* removed `name` fields from questionnaire form, since name is filled in step before when org is created
* removed unnecessary refetches and reloads in `onUpdateOrganization` - compared to community version, we don't create organization at this point, we are just updating existing one so the reloads shouldn't be needed  

Related PR: https://github.com/novuhq/packages-enterprise/pull/177

### DIFF
--- a/apps/api/src/app/organization/e2e/sync-organization.e2e-ee.ts
+++ b/apps/api/src/app/organization/e2e/sync-organization.e2e-ee.ts
@@ -26,10 +26,6 @@ describe.skip('Sync Organization - /organizations (POST)', async () => {
   const testOrganization: UpdateExternalOrganizationDto = {
     jobTitle: JobTitleEnum.ENGINEER,
     domain: 'example.com',
-    productUseCases: {
-      in_app: true,
-      delay: true,
-    },
   };
 
   function getRequest(payload: UpdateExternalOrganizationDto, origin: string = process.env.FRONT_BASE_URL) {
@@ -97,13 +93,6 @@ describe.skip('Sync Organization - /organizations (POST)', async () => {
     // these are stored in the Clerk organization only and then concatenated with the response
     expect(internalOrganization.name).to.equal(updatedClerkOrganization.name);
     expect(internalOrganization.logo).to.equal(updatedClerkOrganization.imageUrl);
-  });
-
-  it('should have all internal attributes set correctly', async () => {
-    expect(internalOrganization.domain).to.equal(testOrganization.domain);
-    expect(internalOrganization.productUseCases?.in_app).to.eq(testOrganization.productUseCases?.in_app);
-    expect(internalOrganization.productUseCases?.delay).to.eq(testOrganization.productUseCases?.delay);
-    expect(internalOrganization.apiServiceLevel).to.eq(ApiServiceLevelEnum.FREE);
   });
 
   it('should update user job title on organization creation', async () => {
--- a/apps/web/src/ApplicationReadyGuard.tsx
+++ b/apps/web/src/ApplicationReadyGuard.tsx
@@ -34,7 +34,7 @@ export function ApplicationReadyGuard({ children }: PropsWithChildren<{}>) {
   function isOnboardingComplete() {
     if (IS_EE_AUTH_ENABLED) {
       // TODO: replace with actual check property (e.g. isOnboardingCompleted)
-      return currentOrganization?.productUseCases !== undefined;
+      return currentOrganization?.productUseCases !== undefined || currentOrganization?.language !== undefined;
     }
 
     return currentOrganization;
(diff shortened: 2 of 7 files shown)


apps/api/src/app/organization/e2e/sync-organization.e2e-ee.ts
denis-kralj-novu, line 1: I believe we should also add tests to confirm that the new questioner data is being saved yes?

apps/web/src/ApplicationReadyGuard.tsx
```diff
@@ -34,7 +34,7 @@ export function ApplicationReadyGuard({ children }: PropsWithChildren<{}>) {
   function isOnboardingComplete() {
     if (IS_EE_AUTH_ENABLED) {
       // TODO: replace with actual check property (e.g. isOnboardingCompleted)
-      return currentOrganization?.productUseCases !== undefined;
+      return currentOrganization?.productUseCases !== undefined || currentOrganization?.language !== undefined;
```
denis-kralj-novu, line 37: ❓ **question:** ‏ I'm looking at this and thinking is this actually an organization level question? Should this maybe be asked on a per user basis?

denis-kralj-novu, line 37: 💭 **thought (non-blocking):** ‏ I'm thinking this is where we might want to consider getting this data from all of our users. If we change the logic to only check if the languages data is present, we can show the form to every user logging in. Question is if we want to. @SokratisVidros wdyt?

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -200,5 +200,6 @@ const toOrganizationEntity = (clerkOrganization: OrganizationResource): IOrganiz
     defaultLocale: clerkOrganization.publicMetadata.defaultLocale,
     domain: clerkOrganization.publicMetadata.domain,
     productUseCases: clerkOrganization.publicMetadata.productUseCases,
+    language: clerkOrganization.publicMetadata.language,
```
denis-kralj-novu, line 223: ☑ **chore:** ‏ consider pluralization (just like in ee)

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
```diff
@@ -43,51 +28,74 @@ export function QuestionnaireForm() {
     handleSubmit,
     formState: { errors },
     control,
-  } = useForm<UpdateExternalOrganizationDto>({});
+  } = useForm<IOrganizationUpdateForm>({});
```
denis-kralj-novu: ❓ **question:** ‏ didn't we have a push to remove the `I` from our interface names?

apps/api/src/app/organization/e2e/sync-organization.e2e-ee.ts
ChmaraX: 100% - we need to create test also for the sync/link itself

apps/web/src/ApplicationReadyGuard.tsx
ChmaraX: It looks to me like an org owner question, its filled once when org is created. Imo the goal is to collect languages/technologies that are used within the organization, not sure if it would help us to collect it per user basis.

ChmaraX: Correct, if we remove the language check, the form would re-appear BUT it will reappear for the first user that gets logged in, which might not be necessary an org owner and the user might just click anything to make the from disappear (speaking from personal experience). I assume the owner usually has better incentive to fill it correctly upon org creation.

apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
ChmaraX: I'm not author of this specific change so I just mirrored the questionnaire as is to the EE version.
We are using this data in tracking so maybe this exact key name is used somewhere in tracking to build some graph or maybe even something else I'm not aware of, so I try to avoid making changes that are not 100% necessary to avoid introducing potential bugs that could delay release.

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
ChmaraX: I think it mainly on API due to the SDK autogeneration from swagger, but good point I changed it here as well to be consistent.

apps/web/src/ApplicationReadyGuard.tsx
SokratisVidros: Same applies when a user invited, or auto joins an organization. For now, org level should be enough.

SokratisVidros (approved):

apps/web/src/ApplicationReadyGuard.tsx
denis-kralj-novu: fair enough

denis-kralj-novu: if this stays on the org level then my question doesn't make sense. We can agree to keep it this way and move on

PR #6174, opened 2024-07-29
fix(web): http api client missing env and memoizing jwt

ChmaraX:
* `environmentId` was missing in header therefore API was defaulting to development env
* we can't memoize the entire `useNovuAPI` for 2 reasons:
  1) when JWT changes due to expiration and we are on notification modal which makes refetch every second - the api client will still use the old jwt for all the requests resulting in 401
  2) same applies for environmentId

### DIFF
--- a/apps/web/src/api/api.client.ts
+++ b/apps/web/src/api/api.client.ts
@@ -102,10 +102,12 @@ export function buildApiHttpClient({
   baseURL = API_ROOT || 'https://api.novu.co',
   secretKey,
   jwt,
+  environmentId,
 }: {
   baseURL?: string;
   secretKey?: string;
   jwt?: string;
+  environmentId?: string;
 }) {
   if (!secretKey && !jwt) {
     throw new Error('A secretKey or jwt is required to create a Novu API client.');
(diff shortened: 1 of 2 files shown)


SokratisVidros:
@ChmaraX This hook is only used in Local Studio as far as I know. The bug reported was from activity feed. How does the fix resolves this?

ChmaraX:
@SokratisVidros its used also in the execution details modal reported in the bug:

```typescript
export function ExecutionDetailsModal({
  notificationId,
  modalVisibility,
  onClose,
  onViewDigestExecution,
}: {
  notificationId: string;
  modalVisibility: boolean;
  onClose: () => void;
  onViewDigestExecution?: (digestNotificationId: string) => void;
}) {
  const theme = useMantineTheme();
  const [shouldRefetch, setShouldRefetch] = useState(true);
  const { data: response, isInitialLoading } = useNotification(notificationId, {
    enabled: !!notificationId,
    refetchInterval: shouldRefetch ? 1000 : false,
  });
```

```typescript
export const useNotification = (notificationId: string, options?: any) => {
  const api = useNovuAPI();

  return useQuery(['notifications', notificationId], () => api.getNotification(notificationId), options);
};
```

denis-kralj-novu (approved):

PR #6179, opened 2024-07-30
chore(web): Redirect legacy domain

denis-kralj-novu:


apps/web/netlify.toml
```diff
@@ -20,6 +20,30 @@
...
   force = true
 
+[[redirects]]
+  from = "https://web.novu.co/*"
+  to = "https://dashboard.novu.co/:splat"
+  status = 301
+  force = true
+
+[[redirects]]
+  from = "http://web.novu.co/*"
```
SokratisVidros, line 30: It's interesting/surprising that we need an `http` redirect as well and they don't always redirect to HTTPS by default or using HSTS

SokratisVidros (approved):

apps/web/netlify.toml
denis-kralj-novu: this was mentioned in the netlify docs that http and https need separate rules. I've just followed the instructions without too much thought

PR #6180, opened 2024-07-30
chore(root): move source of clerk config data to workflow params for all envs

denis-kralj-novu:
secrets in 'Development' environment weren't touched and should work as is
4 new secrets added in 'Production' environment, names visible in the PR and **[here](https://github.com/novuhq/novu/settings/environments/303839906/edit)**

### DIFF
--- a/.github/workflows/dev-deploy-web.yml
+++ b/.github/workflows/dev-deploy-web.yml
@@ -40,6 +40,8 @@ jobs:
       netlify_alias: dev
       netlify_gh_env: development
       netlify_site_id: 45396446-dc86-4ad6-81e4-86d3eb78d06f
+      clerk_publishable_key: {{ secrets.CLERK_PUBLISHABLE_KEY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/dev-deploy-web.yml
```diff
@@ -40,6 +40,8 @@ jobs:
       netlify_alias: dev
       netlify_gh_env: development
       netlify_site_id: 45396446-dc86-4ad6-81e4-86d3eb78d06f
+      clerk_publishable_key: {{ secrets.CLERK_PUBLISHABLE_KEY }}
```
SokratisVidros, line 43: Nit-picky one, since this is a hacky naming, let's make it clear.
```suggestion
      clerk_publishable_key: {{ secrets.CLERK_PUBLISHABLE_KEY_STAGING }}
```

SokratisVidros (approved):

.github/workflows/dev-deploy-web.yml
denis-kralj-novu: we will be changing this anyway, let's leave it for now

PR #6185, opened 2024-07-30
fix(root): Put in real values into variables (they are not secret)

denis-kralj-novu:


.github/workflows/prod-deploy-web.yml
```diff
@@ -11,7 +11,6 @@ on:
 jobs:
   deploy_web_eu:
     uses: ./.github/workflows/reusable-web-deploy.yml
-    environment: Development
```
denis-kralj-novu, line 14: also a mistake on copy pasting, but removed now

SokratisVidros (approved):

PR #6206, opened 2024-07-31
feat(js): inbox sdk manage pagination state in cache

LetItRock:

The notifications in-memory cache:
- stores data in memory
- supports multiple filters aka tabs
- is listening to the notification events like mark as read/unread/archive/unarchive, action button events, and mark all events and updates optimistically data in the cache for any/all filters
- emits events for all updated filters whenever a notification is updated
- updates or removes notifications in the cache depending on the "action"

The cache structure:
```
{
  `${JSON.stringify(filter1 + page1)}`: { ..., notification: [] },
  `${JSON.stringify(filter1 + page2)}`: { ..., notification: [] },
  `${JSON.stringify(filter2 + page1)}`: { ..., notification: [] },
  `${JSON.stringify(filter2 + page1)}`: { ..., notification: [] },
}
```

UI changes:
- the infinite scroll updates to listen to the emitted events
- mark as actions assigned

### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/cache/in-memory-cache.ts
+++ b/packages/js/src/cache/in-memory-cache.ts
@@ -15,7 +15,7 @@ export class InMemoryCache<T> implements Cache<T> {
     return Array.from(this.#cache.values());
   }
 
-  pairs(): [string, T][] {
+  entries(): [string, T][] {
     return Array.from(this.#cache.entries());
   }
 
--- a/packages/js/src/cache/in-memory-cache.ts
+++ b/packages/js/src/cache/in-memory-cache.ts
@@ -1,0 +1,37 @@
+import type { Cache } from './types';
+
+export class InMemoryCache<T> implements Cache<T> {
+  #cache: Map<string, T>;
+
+  constructor() {
+    this.#cache = new Map();
+  }
+
+  get(key: string): T | undefined {
+    return this.#cache.get(key);
+  }
+
+  getValues(): T[] {
+    return Array.from(this.#cache.values());
+  }
+
+  pairs(): [string, T][] {
+    return Array.from(this.#cache.entries());
+  }
+
+  keys(): string[] {
+    return Array.from(this.#cache.keys());
+  }
+
+  set(key: string, value: T): void {
+    this.#cache.set(key, value);
+  }
+
+  remove(key: string): void {
+    this.#cache.delete(key);
+  }
+
+  clear(): void {
+    this.#cache.clear();
+  }
+}
--- a/packages/js/src/cache/notifications-cache.test.ts
+++ b/packages/js/src/cache/notifications-cache.test.ts
@@ -2,9 +2,26 @@
 import { NovuEventEmitter } from '../event-emitter';
 import { ListNotificationsArgs, ListNotificationsResponse, Notification } from '../notifications';
 import { NotificationsCache } from './notifications-cache';
+import { ChannelType } from '../types';
 
-const notification1 = { id: '1', body: 'test1' } as Notification;
-const notification2 = { id: '2', body: 'test2' } as Notification;
+const notification1 = new Notification({
+  id: '1',
+  body: 'test1',
+  isRead: false,
+  isArchived: false,
+  to: { id: '1', subscriberId: '1' },
+  createdAt: new Date().toISOString(),
+  channelType: ChannelType.IN_APP,
+});
+const notification2 = new Notification({
+  id: '2',
+  body: 'test2',
+  isRead: false,
+  isArchived: false,
+  to: { id: '2', subscriberId: '2' },
+  createdAt: new Date().toISOString(),
+  channelType: ChannelType.IN_APP,
+});
 
 describe('NotificationsCache', () => {
   let notificationsCache: NotificationsCache;
@@ -122,13 +139,50 @@ describe('NotificationsCache', () => {
     expect(result).toEqual([notification1, notification2]);
   });
 
-  it('should update single notifications and emit single event', () => {
+  it('should get unique read notifications based on tags', () => {
+    const updated1 = new Notification({ ...notification1, isRead: true });
+    const updated2 = new Notification({ ...notification2, isRead: true });
+    const updated3 = new Notification({ ...notification2, id: '3' });
+
+    const args1 = { tags: ['tag1'], limit: 10, offset: 0 };
+    const data1: ListNotificationsResponse = {
+      hasMore: false,
+      filter: {},
+      notifications: [updated1, updated2],
+    };
+
+    const args2 = { tags: ['tag1'], limit: 10, offset: 1 };
+    const data2: ListNotificationsResponse = {
+      hasMore: false,
+      filter: {},
+      notifications: [updated3],
+    };
+
+    const args3 = { tags: ['tag2'], limit: 10, offset: 0 };
+    const data3: ListNotificationsResponse = {
+      hasMore: false,
+      filter: {},
+      notifications: [notification2],
+    };
+
+    notificationsCache.set(args1, data1);
+    notificationsCache.set(args2, data2);
+    notificationsCache.set(args3, data3);
+
+    let result = notificationsCache.getUniqueNotifications({ tags: ['tag1'], read: true });
+    expect(result).toEqual([updated1, updated2]);
+
+    result = notificationsCache.getUniqueNotifications({ tags: ['tag2'] });
+    expect(result).toEqual([notification2]);
+  });
+
+  it('should update notification and emit single event', () => {
     const args: ListNotificationsArgs = { limit: 10, offset: 0, tags: ['tag1'], read: false, archived: false };
-    const updatedNotification = { id: '1', body: 'Updated Notification' } as Notification;
+    const updatedNotification = new Notification({ ...notification1, body: 'Updated Notification' });
     const data: ListNotificationsResponse = { hasMore: false, filter: {}, notifications: [notification1] };
 
     notificationsCache.set(args, data);
-    (notificationsCache as any).handleSingleNotificationEvent({ data: updatedNotification });
+    (notificationsCache as any).handleNotificationEvent()({ data: updatedNotification });
 
     expect(mockEmitter.emit).toHaveBeenCalledWith('notifications.list.updated', {
       data: {
@@ -139,38 +193,83 @@ describe('NotificationsCache', () => {
     });
   });
 
-  it('should update single notifications for different filters and emit two events', () => {
+  it('should remove notification and emit single event', () => {
+    const args: ListNotificationsArgs = { limit: 10, offset: 0, tags: ['tag1'], read: false, archived: false };
+    const updatedNotification = new Notification({ ...notification1, body: 'Updated Notification' });
+    const data: ListNotificationsResponse = { hasMore: false, filter: {}, notifications: [notification1] };
+
+    notificationsCache.set(args, data);
+    (notificationsCache as any).handleNotificationEvent({ remove: true })({ data: updatedNotification });
+
+    expect(mockEmitter.emit).toHaveBeenCalledWith('notifications.list.updated', {
+      data: {
+        hasMore: false,
+        filter: {},
+        notifications: [],
+      },
+    });
+  });
+
+  it('should update notification for different filters and emit two events', () => {
     const filter1 = { tags: ['tag1'], read: false, archived: false };
     const filter2 = { tags: ['tag2'], read: false, archived: false };
     const args1: ListNotificationsArgs = { limit: 10, offset: 0, ...filter1 };
     const args2: ListNotificationsArgs = { limit: 10, offset: 0, ...filter2 };
-    const updatedNotification = { id: '1', body: 'Updated Notification' } as Notification;
+    const updatedNotification = new Notification({ ...notification1, body: 'Updated Notification' });
 
-    notificationsCache.set(args1, { hasMore: false, filter: filter1, notifications: [notification1] });
-    notificationsCache.set(args2, { hasMore: false, filter: filter2, notifications: [notification1] });
-    (notificationsCache as any).handleSingleNotificationEvent({ data: updatedNotification });
+    notificationsCache.set(args1, { hasMore: false, filter: filter1, notifications: [notification1, notification2] });
+    notificationsCache.set(args2, { hasMore: false, filter: filter2, notifications: [notification1, notification2] });
+    (notificationsCache as any).handleNotificationEvent()({ data: updatedNotification });
 
     expect(mockEmitter.emit).toHaveBeenCalledTimes(2);
     expect(mockEmitter.emit).toHaveBeenNthCalledWith(1, 'notifications.list.updated', {
       data: {
         hasMore: false,
         filter: filter1,
-        notifications: [updatedNotification],
+        notifications: [updatedNotification, notification2],
       },
     });
     expect(mockEmitter.emit).toHaveBeenNthCalledWith(2, 'notifications.list.updated', {
       data: {
         hasMore: false,
         filter: filter2,
-        notifications: [updatedNotification],
+        notifications: [updatedNotification, notification2],
+      },
+    });
+  });
+
+  it('should remove notification for different filters and emit two events', () => {
+    const filter1 = { tags: ['tag1'], read: false, archived: false };
+    const filter2 = { tags: ['tag2'], read: false, archived: false };
+    const args1: ListNotificationsArgs = { limit: 10, offset: 0, ...filter1 };
+    const args2: ListNotificationsArgs = { limit: 10, offset: 0, ...filter2 };
+    const updatedNotification = new Notification({ ...notification1, body: 'Updated Notification' });
+
+    notificationsCache.set(args1, { hasMore: false, filter: filter1, notifications: [notification1, notification2] });
+    notificationsCache.set(args2, { hasMore: false, filter: filter2, notifications: [notification1, notification2] });
+    (notificationsCache as any).handleNotificationEvent({ remove: true })({ data: updatedNotification });
+
+    expect(mockEmitter.emit).toHaveBeenCalledTimes(2);
+    expect(mockEmitter.emit).toHaveBeenNthCalledWith(1, 'notifications.list.updated', {
+      data: {
+        hasMore: false,
+        filter: filter1,
+        notifications: [notification2],
+      },
+    });
+    expect(mockEmitter.emit).toHaveBeenNthCalledWith(2, 'notifications.list.updated', {
+      data: {
+        hasMore: false,
+        filter: filter2,
+        notifications: [notification2],
       },
     });
   });
 
   it('should update multiple notifications and emit single event', () => {
     const args: ListNotificationsArgs = { limit: 10, offset: 0, tags: ['tag1'], read: false, archived: false };
-    const updatedNotification1 = { id: '1', body: 'Updated Notification' } as Notification;
-    const updatedNotification2 = { id: '2', body: 'Updated Notification' } as Notification;
+    const updatedNotification1 = new Notification({ ...notification1, body: 'Updated Notification' });
+    const updatedNotification2 = new Notification({ ...notification2, body: 'Updated Notification' });
     const data: ListNotificationsResponse = {
       hasMore: false,
       filter: {},
@@ -178,7 +277,7 @@ describe('NotificationsCache', () => {
     };
 
     notificationsCache.set(args, data);
-    (notificationsCache as any).handleMultipleNotificationsEvent({
+    (notificationsCache as any).handleNotificationsEvent()({
       data: [updatedNotification1, updatedNotification2],
     });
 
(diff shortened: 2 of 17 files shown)


packages/js/src/cache/notifications-cache.test.ts
```diff
@@ -0,0 +1,232 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { NovuEventEmitter } from '../event-emitter';
+import { ListNotificationsArgs, ListNotificationsResponse, Notification } from '../notifications';
+import { NotificationsCache } from './notifications-cache';
+
+const notification1 = { id: '1', body: 'test1' } as Notification;
+const notification2 = { id: '2', body: 'test2' } as Notification;
+
+describe('NotificationsCache', () => {
```
LetItRock, line 26: tests for the notifications cache

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+];
+
+const multipleNotificationEvents: NotificationsEvents[] = [
+  'notifications.read_all.pending',
+  'notifications.read_all.resolved',
+  'notifications.archive_all.pending',
+  'notifications.archive_all.resolved',
+  'notifications.archive_all_read.pending',
+  'notifications.archive_all_read.resolved',
+];
```
LetItRock: These are all the events that the cache is listening to. The `pending` events send the optimistically updated data, the `resolved` events send the data from the server.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+
+export class NotificationsCache {
+  #emitter: NovuEventEmitter;
+  /**
+   * The key is the stringified notifications filter, the value is the pagination cache
+   * where the key is pagination arguments and the value is the list of notifications.
+   */
+  #cache: Cache<Cache<ListNotificationsResponse>>;
```
LetItRock: The cache structure is like this:
```
{
  `${JSON.stringify(filter1)}`: {
      `${JSON.stringify(page1)}`: { ..., notification: [] },
      `${JSON.stringify(page2)}`: { ..., notification: [] }
   },
  `${JSON.stringify(filter2)}`: {
      `${JSON.stringify(page1)}`: { ..., notification: [] },
      `${JSON.stringify(page2)}`: { ..., notification: [] }
   }
}
```

Because of a few reasons:
- to support multiple filters aka tabs
- to have the linear pagination state ordered by requests
- allows separation of filters from pagination

I know it's not ideal, but I tried to simplify it and wasn't able to do so in a short period, mainly the hardest part is sending the appropriate events data whenever the notification is updated (line 104, 131). 

The possible structure could be like: 
```
{
  `${JSON.stringify(filter1 + page1)}`: { ..., notification: [] },
  `${JSON.stringify(filter1 + page2)}`: { ..., notification: [] },
  `${JSON.stringify(filter2 + page1)}`: { ..., notification: [] },
  `${JSON.stringify(filter2 + page1)}`: { ..., notification: [] },
}
```

We can give it a second chance later, especially since refactoring should be safer as we have tests.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * The key is the stringified notifications filter, the value is the set of unique notification ids.
+   */
+  #idsCache: Cache<Set<string>>;
```
LetItRock: The helper cache stores unique ids per filter.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+      }
+
+      const updatedNotifications = [...el.notifications];
+      updatedNotifications[index] = data;
+
+      cache.set(key, { ...el, notifications: updatedNotifications });
+    });
+  };
+
+  private handleSingleNotificationEvent = ({ data }: { data?: Notification }): void => {
```
LetItRock: The callback fired for events like mark as read, unread, etc.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (!data) {
+      return;
+    }
+
+    const filterKeysToUpdate = this.#idsCache.keys().filter((key) => {
+      const cache = this.#idsCache.get(key);
+
+      return cache?.has(data.id);
+    });
```
LetItRock: Find all the filters that have that notification.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+        return;
+      }
+
+      this.updateNotificationInCache(filterCache, data);
+
+      const notificationsResponse = this.getReducedNotifications(filterCache);
+      this.#emitter.emit('notifications.list.updated', {
+        data: notificationsResponse,
+      });
+    });
```
LetItRock: For each filter update the notification in the cache and emit the `notifications.list.updated` event with aggregated notifications from all the pages. 
Then there is the listener that updates the state.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+  };
+
+  private handleMultipleNotificationsEvent = ({ data }: { data?: Notification[] }): void => {
```
LetItRock: The callback fired for events like mark all as read, mark all archived, mark all read archived.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+      return notificationIds.some((id) => idsSet?.has(id));
+    });
+
+    filterKeysToUpdate.forEach((key) => {
+      const filterCache = this.#cache.get(key);
+      if (!filterCache) {
+        return;
+      }
+
+      data.forEach((notification) => this.updateNotificationInCache(filterCache, notification));
```
LetItRock: Similar logic but updates many notifications in the cache.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+    return values.reduce<ListNotificationsResponse>(
+      (acc, el) => {
+        return {
+          hasMore: el.hasMore,
+          filter: el.filter,
+          notifications: [...acc.notifications, ...el.notifications],
+        };
+      },
+      { hasMore: false, filter: {}, notifications: [] }
+    );
```
LetItRock: Returns the last pagination response with all aggregated notifications.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+    const filterKey = getFilterKey(args);
+    const pageKey = getPageKey(args);
+    const filterCache = this.getFilterCache(args) ?? new InMemoryCache();
+    filterCache.set(pageKey, data);
+    this.#cache.set(filterKey, filterCache);
+
+    const uniqueIds = this.#idsCache.get(filterKey) ?? new Set();
+    data.notifications.forEach((el) => uniqueIds.add(el.id));
+    this.#idsCache.set(filterKey, uniqueIds);
+  }
```
LetItRock: Store the pagination response for the specified `filter` and `page` data.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+        filterCache
+          .getValues()
+          .map((el) => el.notifications)
+          .flat()
+          .forEach((notification) => uniqueNotifications.set(notification.id, notification));
+      }
+    });
+
+    return Array.from(uniqueNotifications.values());
+  }
```
LetItRock: This function is used by actions like "mark all as read" etc., where we return all the notifications for the specified optional tags, optimistically update all these notifications, and then update the cache.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+  clear(filter: NotificationFilter): void {
+    const filterKey = getFilterKey(filter);
+    this.#cache.remove(filterKey);
+    this.#idsCache.remove(filterKey);
+  }
+
+  clearAll(): void {
+    this.#cache.clear();
+    this.#idsCache.clear();
+  }
```
LetItRock: Not used currently but might be useful later to clear the cache and to refetch the data.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+    filterKeysToUpdate.forEach((key) => {
+      const filterCache = this.#cache.get(key);
+      if (!filterCache) {
+        return;
+      }
+
+      this.updateNotificationInCache(filterCache, data);
+
+      const notificationsResponse = this.getReducedNotifications(filterCache);
+      this.#emitter.emit('notifications.list.updated', {
```
LetItRock: This is the event that the UI is listening to. Of course, the name is a matter of discussion.

packages/js/src/notifications/helpers.ts
```diff
@@ -261,3 +262,111 @@ const getNotificationDetails = (
     };
   }
 };
+
+export const readAll = async ({
```
LetItRock, line 264: The code was moved from another file but slightly updated with the cache usage.

packages/js/src/notifications/notifications.ts
```diff
@@ -21,28 +31,48 @@ import type {
...
-        });
-
-        const modifiedResponse: ListNotificationsResponse = {
-          hasMore: response.hasMore,
-          filter: response.filter,
-          notifications: response.data.map((el) => new Notification(el)),
-        };
+        let data: ListNotificationsResponse | undefined = options.cache
+          ? this.#notificationsCache.getAll(args)
+          : undefined;
```
LetItRock: If option `cache` is `true` get all the data from cache if exists.

packages/js/src/notifications/notifications.ts
```diff
@@ -21,28 +31,48 @@ import type {
...
+          data = {
+            hasMore: response.hasMore,
+            filter: response.filter,
+            notifications: response.data.map((el) => new Notification(el)),
+          };
+
+          if (options.cache) {
+            this.#notificationsCache.set(args, data);
+            data = this.#notificationsCache.getAll(args);
+          }
```
LetItRock: After we get a response from the server, store the data in the cache and return aggregated notifications.

packages/js/src/ui/api/hooks/useNotifications.ts
```diff
@@ -1,52 +1,36 @@
...
+        return;
+      }
+
+      mutate({ data: data.notifications, hasMore: data.hasMore });
+    };
+
+    novu.on('notifications.list.updated', listener);
+
+    onCleanup(() => novu.off('notifications.list.updated', listener));
+  });
```
LetItRock, line 33: Listen to the cache event and update the data in the signal.

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -58,24 +58,52 @@ export const DefaultNotification = (props: DefaultNotificationProps) => {
           <Show
             when={props.notification.isRead}
             fallback={
-              <Button appearanceKey="notificationRead__button" size="icon" variant="icon">
+              <Button
+                appearanceKey="notificationRead__button"
+                size="icon"
+                variant="icon"
+                onClick={() => {
+                  props.notification.read();
```
LetItRock: I had to test the cache somehow so I added these calls for actions ;)

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,212 @@
...
+    const keys = this.#cache.keys();
+
+    const uniqueNotifications = new Map<string, Notification>();
+    keys.forEach((key) => {
+      const filter: NotificationFilter = JSON.parse(key);
+      if (!tags || tags?.every((tag) => filter.tags?.includes(tag))) {
+        const filterCache = this.#cache.get(key);
+        if (!filterCache) {
+          return;
+        }
```
BiswaViraj: ✏ **suggestion:** Can you please add a code comment here about what it does.
It seems little confusing why we want to have tags in every filter‏

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,254 @@
...
+const notificationsEvents: NotificationsEvents[] = [
+  'notifications.read_all.pending',
+  'notifications.read_all.resolved',
+];
+
+// these events should remove the notifications from the cache
+const notificationsEventsThatRemove: NotificationsEvents[] = [
+  'notifications.archive_all.pending',
+  'notifications.archive_all_read.pending',
+];
```
LetItRock: We decided to have the new Inbox behave similarly to Notion Inbox. Where it should keep the notifications in the list when it's marked read/unread, but when it's archived/unarchived it should be removed. Similarly, it will behave to mark all actions.

This class implements this logic based on the emitted events.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
...
+          uniqueFilterKeys.add(getFilterKey(getFilter(key)));
+        }
+      });
+
+      uniqueFilterKeys.forEach((key) => {
+        const notificationsResponse = this.getAggregated(getFilter(key));
+        this.#emitter.emit('notifications.list.updated', {
+          data: notificationsResponse,
+        });
+      });
```
LetItRock: Depending on the received event we decide whether the notification should be removed from the cache or updated in the cache. 
The notification can belong to many different tabs/filters. We must collect unique filters and emit events for each.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
...
+
+          return {
+            hasMore: el.hasMore,
+            filter: el.filter,
+            notifications: [...acc.notifications, ...el.notifications],
+          };
+        },
+        { hasMore: false, filter: {}, notifications: [] }
+      );
+  }
```
LetItRock, line 170: This function aggregates all the paginated notifications for the specified tab/filter and returns the combined data with the last paginated response information.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        value.notifications
+          .filter((el) => typeof read === 'undefined' || read === el.isRead)
+          .forEach((notification) => uniqueNotifications.set(notification.id, notification));
+      }
+    });
+
+    return Array.from(uniqueNotifications.values());
+  }
```
LetItRock, line 207: This function is used for actions like: mark all read, mark all archived, mark read as archived. 
We need to collect a list of unique notifications for the specified filters and then optimistically update them. Later we send it with the pending event, so the cache can change and update it.

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -33,7 +33,7 @@ const InboxContent = (props: InboxContentProps) => {
           <Header updateScreen={setCurrentScreen} />
           <Show
             when={props.tabs && props.tabs.length > 0}
-            fallback={<NotificationList mountNotification={props.mountNotification} />}
+            fallback={<NotificationList mountNotification={props.mountNotification} options={{ archived: false }} />}
```
LetItRock: This is temporarily we will update it based on the filter in the Inbox header.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
+import { InMemoryCache } from './in-memory-cache';
+import type { Cache } from './types';
+import type { ListNotificationsArgs, ListNotificationsResponse, Notification } from '../notifications';
+import { NotificationEvents, NotificationsEvents, NovuEventEmitter } from '../event-emitter';
+import type { NotificationFilter } from '../types';
+import { areTagsEqual, isSameFilter } from '../utils/notification-utils';
+
+const getCacheKey = (args: ListNotificationsArgs): string => {
+  return JSON.stringify(args);
```
desiprisg: ⚠️ I think we should be doing `{ tags, read, archived }` because `JSON.stringify` preserves the order of the properties. This means that we could get different results for the same filter.

packages/js/src/notifications/helpers.ts
```diff
@@ -261,3 +262,107 @@ const getNotificationDetails = (
...
+
+    return {};
+  } catch (error) {
+    emitter.emit('notifications.archive_all.resolved', { args: { tags }, error });
+
+    return { error: new NovuError('Failed to archive all notifications', error) };
+  }
+};
+
+export const archiveAllRead = async ({
```
desiprisg, line 338: ❓  Why is this function needed? Couldn't we pass a filter to `archiveAll()` and avoid the extra functions ?

packages/js/src/cache/notifications-cache.ts
desiprisg: ❓ Does this really make sense here?

I don't think this is the place to do this. If a notification is marked as read and a new call to the fetch function is made, it should not include the notification that was marked as read. Think of it like switching to another tab and going back again. A call is made again and the notification should disappear, although no network requests were made. At least that's what makes sense to me, dunno how doable it is. Let's discuss.

desiprisg: ❓ Does this really make sense here?

I don't think this is the place to do this. If a notification is marked as read and a new call to the fetch function is made, it should not include the notification that was marked as read. Think of it like switching to another tab and going back again. A call is made again and the notification should disappear, although no network requests were made. At least that's what makes sense to me, dunno how doable it is. Let's discuss.

packages/js/src/cache/notifications-cache.ts
LetItRock: good catch!

packages/js/src/notifications/helpers.ts
LetItRock: these helper functions create different optimistically updated objects, send different events, and call different api... too many if statements

packages/js/src/cache/notifications-cache.ts
LetItRock: @desiprisg 

**Behavior:**
It all depends on the top filter that would be applied in the header we have three: `unread + read`, `unread`, `archived`. From what I talked with Nikolay and @SokratisVidros we agreed and wanted to copy the same behaviour from Notion, meaning:
- read/unread actions keep the notification in the list
  - with the `unread + read` filter you will always see it even after refresh
  - with the `unread` filter it will disappear after refresh
- archive/unarchive actions remove notification from the list
  - no matter what filter is applied 

**Cache:**
We talked about this case in the cache meeting that we had: the `Notification` class provides the functionality to mark it as read/unread/archive/unarchive. For example `notification.read()` will optimistically update the notification fields and the value in the cache and will call API with a response updating the cache again (because API is the source of truth). This is what we have agreed, logic in JS SDK.

We can't do a refetch after it's marked as read, because there might be new notifications that will shift the whole list so we would need to refetch all pages.

SokratisVidros: Let’s stick to Notion-like behavior for now. 

The cache powers the UI. Changes to individual notifications are applied to the list but notifications are not removed from the list unless the cache is invalidated.

Gmail does the same thing, you filter by unread then you mark an email as read, it’s not bold anymore but it is in the unread list until the next reload.

desiprisg: Ok so in essence, the update happens to the notification only, not any lists. Fine by me if that's how we envisioned it. I can see the benefits in both!

Looks good then!

packages/js/src/cache/in-memory-cache.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  get(key: string): T | undefined {
+    return this.#cache.get(key);
+  }
+
+  getValues(): T[] {
+    return Array.from(this.#cache.values());
+  }
+
+  pairs(): [string, T][] {
```
SokratisVidros:
```suggestion
  entries(): [string, T][] {
```

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
+import { InMemoryCache } from './in-memory-cache';
+import type { Cache } from './types';
+import type { ListNotificationsArgs, ListNotificationsResponse, Notification } from '../notifications';
+import { NotificationEvents, NotificationsEvents, NovuEventEmitter } from '../event-emitter';
+import type { NotificationFilter } from '../types';
+import { areTagsEqual, isSameFilter } from '../utils/notification-utils';
+
+const getCacheKey = ({ tags, read, archived, limit, offset, after }: ListNotificationsArgs): string => {
+  return JSON.stringify({ tags, read, archived, limit, offset, after });
```
SokratisVidros: I'd suggest we remove null, undefined and empty-arrays keys before adding them to the cache so that we can have more predictable and easy to debug cache keys.

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
...
+export class NotificationsCache {
+  #emitter: NovuEventEmitter;
+  /**
+   * The key is the stringified notifications filter, the values are the paginated notifications.
+   */
+  #cache: Cache<ListNotificationsResponse>;
+
+  constructor() {
+    this.#emitter = NovuEventEmitter.getInstance();
+    notificationEvents.forEach((event) => {
```
SokratisVidros: Instead of splitting events per resource and making the cache know about the notification types, group them by cache functionality. For example, updateEvents and clearanceEvents. That way, you will have two arrays to iterate and add listeners on.

Moreover, pass these events to the cosntructors so that the cache class is unaware of the events types in the first place. Pure dependency injection ;)

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    const updatedNotifications = [...notificationsResponse.notifications];
+    updatedNotifications[index] = data;
+
+    this.#cache.set(key, { ...notificationsResponse, notifications: updatedNotifications });
+
+    return true;
+  };
+
+  private removeNotificationInCache = (key: string, data: Notification): boolean => {
```
SokratisVidros:
```suggestion
  private removeNotificationFromCache = (key: string, data: Notification): boolean => {
```

Or just `removeNotification`

packages/js/src/cache/notifications-cache.ts
```diff
@@ -0,0 +1,253 @@
...
+    this.#cache.set(key, {
+      ...notificationsResponse,
+      notifications: newNotifications,
+    });
+
+    return true;
+  };
+
+  private handleNotificationEvent =
+    ({ remove }: { remove: boolean } = { remove: false }) =>
```
SokratisVidros, line 111: This boolean is not necessary if the cache holds two agnostic arrays of events that delete and events that update notifications.

SokratisVidros (approved):

packages/js/src/cache/notifications-cache.ts
LetItRock: The thing is that the events are tied to a callback interface implementation, so DI is not suited well here.

packages/js/src/cache/notifications-cache.ts
LetItRock: I'll try to simplify that, but you still need to know in the callback when to remove and when to delete, otherwise, you will need to have two different callbacks.

PR #6211, opened 2024-07-31
fix(api): Remove admin check and track always

SokratisVidros:

Every organization should have at least one admin, so the check shouldn't be necessary.


rifont (approved):
🎉

PR #6227, opened 2024-08-01
chore(js,react): inbox props for the notification item click and action button clicks

LetItRock:
Inbox:
- the handler props for the notification item click and action button clicks
- the default behavior for notification item click
- the default behavior for the action button click
- fixes to the WS event mapping and `useWebsocketEvent` hook

### Screenshots

https://github.com/user-attachments/assets/9a0f3105-161c-48a8-ac94-0e96d388f110


apps/api/src/app/inbox/utils/notification-mapper.ts
```diff
@@ -34,7 +34,7 @@ const mapSingleItem = ({
     id: _id,
     subject,
     body: content as string,
-    to,
+    to: subscriber ? to : undefined,
```
LetItRock: when subscriber is not defined don't build the empty object

apps/api/src/app/inbox/utils/notification-mapper.ts
```diff
@@ -52,6 +51,11 @@ const mapSingleItem = ({
     },
     channelType: channel,
     tags,
+    redirect: cta.data.url
+      ? {
+          url: cta.data.url,
+        }
+      : undefined,
```
LetItRock: we can extend the functionality later when this field is an object, for example, add `target: '_blank'` etc

libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -594,9 +594,26 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
...
     const isUpdatingSecondaryCta = actionType === ButtonTypeEnum.SECONDARY;
-    const updatePayload: Partial<MessageEntity> = {};
+    const updatePayload: FilterQuery<MessageEntity> = !message.read
+      ? {
+          seen: true,
+          lastSeenDate: new Date(),
+          read: true,
+          lastReadDate: new Date(),
+        }
+      : {};
```
LetItRock, line 616: when action is performed mark the notification as read if it's not already read

package.json
```diff
@@ -39,7 +39,7 @@
     "start:docker:embed": "cd libs/embed && npm run start:docker",
     "start:docker:widget": "cross-env nx run-many --target=start:docker --projects=@novu/widget",
     "docker:build": "pnpm -r --if-present --parallel docker:build",
-    "build": "nx run-many --target=build --all",
+    "build": "nx run-many --target=build --all --exclude=nextjs",
```
LetItRock, line 42: exclude building the `playground` nextjs application

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -11,6 +16,8 @@ export type InboxProps = {
   tabs?: Array<{ label: string; value: Array<string> }>;
   mountNotification?: NotificationMounter;
   mountBell?: BellMounter;
+  onNotificationClick?: NotificationClickHandler;
+  onActionClick?: NotificationActionClickHandler;
```
LetItRock: two new props that could be passed on to the `Inbox` or `Notifications` component

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -14,20 +16,49 @@ const NotificationBody = (props: NotificationBodyProps) => {
...
+
+    if (!props.notification.isRead) {
+      props.notification.read();
+    }
+
+    props.onNotificationClick?.({ notification: props.notification });
+    if (props.notification.redirect?.url) {
+      window.open(props.notification.redirect?.url, '_blank', 'noreferrer noopener');
+    }
+  };
```
LetItRock, line 43: when the notification is clicked:
- mark it as read if is not already
- call the `onNotificationClick` prop
- if the `redirect.url` was passed, open the link in a new browser tab

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -14,20 +16,49 @@ const NotificationBody = (props: NotificationBodyProps) => {
...
+  const handleActionButtonClick = (action: ActionTypeEnum, e: MouseEvent) => {
+    e.stopPropagation();
+
+    if (action === ActionTypeEnum.PRIMARY) {
+      props.notification.completePrimary();
+    } else {
+      props.notification.completeSecondary();
+    }
+
+    props.onActionClick?.({ notification: props.notification, action });
```
LetItRock: when action button is clicked:
- complete the action
- call the `onActionClick` prop

packages/js/src/ws/socket.ts
```diff
@@ -34,7 +98,7 @@ export class Socket extends BaseModule {
 
   #notificationReceived = ({ message }: { message: TODO }) => {
     this.#emitter.emit(NOTIFICATION_RECEIVED, {
-      result: new Notification(message),
+      result: new Notification(mapToNotification(message)),
```
LetItRock, line 101: the WS event emits the message entity which should we map to the Notification class object

playground/nextjs/src/pages/render-notification/index.tsx
```diff
@@ -8,7 +8,7 @@ export default function Home() {
       <Title title="Render Notification Props" />
       <Inbox
         options={novuConfig}
-        renderNotification={(notification) => {
+        renderNotification={({ notification }) => {
```
LetItRock, line 11: changed it to this interface as it's easier to work / extend in the future it would be needed

apps/api/src/app/inbox/utils/types.ts
```diff
@@ -10,15 +10,14 @@ export type Subscriber = {
...
-  url?: string;
   isCompleted: boolean;
 };
 
 export type InboxNotification = {
   id: string;
   subject?: string;
   body: string;
-  to: Subscriber;
+  to?: Subscriber;
```
BiswaViraj: ❓ **question:** Is it possible to have a notification without the subscriber?‏

BiswaViraj (approved):

packages/js/src/ui/components/Notification/DefaultNotification.tsx
BiswaViraj: 💭 **thought (non-blocking):** On a second thought, do we really need the redirectUrl feature?
We have the `onNotificationClick`, users can add their own logic to `navigate/redirect` using their own implementation of router. and also will be easier for them to create on demand Urls in their app.‏

apps/api/src/app/inbox/utils/types.ts
```diff
@@ -29,6 +28,9 @@ export type InboxNotification = {
   secondaryAction?: Action;
   channelType: ChannelTypeEnum;
   tags?: string[];
+  redirect?: {
```
desiprisg, line 31: ❓  How often would someone need this property to declare a url from their backend in order to use it in their frontend? Is that more common that handling it on their frontend via the prop ?

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -14,20 +16,49 @@ const NotificationBody = (props: NotificationBodyProps) => {
...
+  };
+
   return (
-    <div
+    <a
       class={style(
         'notification',
-        'nt-w-full nt-text-sm hover:nt-bg-neutral-100 nt-group nt-relative nt-flex nt-px-6 nt-py-4 nt-gap-2'
+        'nt-w-full nt-text-sm hover:nt-bg-neutral-100 nt-group nt-relative nt-flex nt-px-6 nt-py-4 nt-gap-2 nt-cursor-pointer'
```
desiprisg: ❓  Do we want a pointer at all times here since it will mark it as read anyway?

packages/js/src/ui/components/Inbox.tsx
desiprisg: ❓  Shouldn't we have a `onPrimaryActionClick` and one for the secondary as well?

apps/api/src/app/inbox/utils/types.ts
LetItRock: How often? I don't know. But the API/framework provides you more flexibility when it comes to building this URL, you might compose it from the data that is not available in your FE.

packages/js/src/ui/components/Notification/DefaultNotification.tsx
LetItRock: No, we don't, only when the notification is unread or there is a redirect URL.

packages/js/src/ui/components/Inbox.tsx
LetItRock: We can have two `onPrimaryActionClick` and `onSecondaryActionClick`, but it's more props, and if we will support X buttons we would have to add X props :/

desiprisg: We need to support both buttons now though, right? I suppose it's fine to have two props here, since we have 2 properties on the BE as well.

apps/api/src/app/inbox/utils/types.ts
LetItRock: it should never be, but this is how the type on the MessageEntity is defined...

packages/js/src/ui/components/Notification/DefaultNotification.tsx
LetItRock: @BiswaViraj https://github.com/novuhq/novu/pull/6227/files#r1702356457
We provide flexibility in building the dynamic URL based on the trigger payload, system variables etc., and now with the framework they can do that by querying their db or any other system. This way on the FE they don't need to fetch the data from API to build. But also, it might not be possible to recreate the same context/state of the system similarly to which they did the trigger (I mean when fetching the data from API). 

I've checked the DB and half of all redirect URLs use the variables.

packages/js/src/ui/components/Inbox.tsx
LetItRock: yeah, I guess it is better to reflect what we have in the API/SDK :)

apps/api/src/app/inbox/utils/types.ts
LetItRock: https://github.com/novuhq/novu/pull/6227#discussion_r1703387453

desiprisg (approved):
💯

PR #6235, opened 2024-08-02
chore(root): Production deployment fixes

SokratisVidros:
1. Build only EE version for Cloud
2. Do not run test during production deployment, run them before

### DIFF
--- a/.github/workflows/dev-deploy-ws.yml
+++ b/.github/workflows/dev-deploy-ws.yml
@@ -55,11 +55,6 @@ jobs:
               }
             }
 
-      - name: Setup QEMU
-        uses: docker/setup-qemu-action@v3
-        with:
-          platforms: linux/amd64,linux/arm64
-
       - name: Set Up Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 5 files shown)


.github/workflows/dev-deploy-ws.yml
```diff
@@ -55,11 +55,6 @@ jobs:
               }
             }
 
-      - name: Setup QEMU
-        uses: docker/setup-qemu-action@v3
-        with:
-          platforms: linux/amd64,linux/arm64
-
```
rifont, line 62: ❓ **question:** Currently we are using QEMU in 13 files in the repo, this PR removes 3 of them. The question is:
1. What purpose does QEMU serve?
2. Can we remove it in those other places too?
‏

rifont (approved):
Speeeed! 🏎️

.github/workflows/dev-deploy-ws.yml
SokratisVidros: It's a relic from the past when we were doing multi architecture deployments.

PR #6254, opened 2024-08-05
chore(js,react): inbox implemented the filtering logic: unread + read, unread, archived

LetItRock:
Inbox:
- exposed the `tabs, open` props in the react package Inbox component
- exposed the "clear cache" in JS SDK
- created the InboxProvider in Solid
- implemented the filtering logic by: unread + read, unread, archived
- connected the `mark all read, archive all, archive read` buttons with SDK calls
- adjusted the logic when the notification actions should be shown

### Screenshots

https://github.com/user-attachments/assets/4528886d-9ce9-47dc-af0a-ad59e57148cf

https://github.com/user-attachments/assets/9ec4b6d4-84e7-4cfd-aa6f-13a86b094aa8


packages/js/src/notifications/notifications.ts
```diff
@@ -242,4 +242,12 @@ export class Notifications extends BaseModule {
       })
     );
   }
+
+  clearCache({ filter }: { filter?: NotificationFilter } = {}): void {
```
LetItRock, line 246: the clear cache functionality, allows to clear by specific filter (tab) or all

packages/js/src/ui/api/hooks/useNotifications.ts
```diff
@@ -32,5 +32,16 @@ export const useNotificationsInfiniteScroll = (props?: UseNotificationsInfiniteS
...
+  createEffect(() => {
+    const newFilter = { ...props.options() };
+    if (isSameFilter(filter, newFilter)) {
+      return;
+    }
+
+    novu.notifications.clearCache();
+    reset();
+    filter = newFilter;
+  });
```
LetItRock, line 44: when the new filter is changed for example to `archived` we need to clear all cache for all tabs and reset the infinite scroll which will trigger API call

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -36,25 +35,28 @@ type InboxContentProps = {
 
 const InboxContent = (props: InboxContentProps) => {
   const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
+  const { tabs, filter } = useInboxContext();
+  const options = createMemo(() => ({ read: filter().read, archived: filter().archived }));
```
LetItRock: when single notification list is shown we don't care about the `tags` as we query all notifications

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -36,25 +35,28 @@ type InboxContentProps = {
 
 const InboxContent = (props: InboxContentProps) => {
   const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
+  const { tabs, filter } = useInboxContext();
```
LetItRock, line 38: created the new `InboxProvider` and hooks

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -90,47 +90,49 @@ export const DefaultNotification = (props: DefaultNotificationProps) => {
             'nt-gap-3 nt-shrink-0 nt-ml-6 nt-float-right nt-hidden group-hover:nt-flex'
           )}
         >
-          <Show
-            when={props.notification.isRead}
-            fallback={
+          <Show when={status() !== NotificationStatus.ARCHIVED}>
```
LetItRock: the read/unread actions should be only shown for filters: `unread + read, unread`

packages/js/src/ui/components/elements/Header/MoreActionsDropdown.tsx
```diff
@@ -1,24 +1,30 @@
...
-          <Button variant="ghost" size="icon" {...triggerProps}>
-            <DotsMenu />
-          </Button>
-        )}
-      />
-      <Dropdown.Content appearanceKey="moreActions__dropdownContent">
-        <MoreActionsOptions />
-      </Dropdown.Content>
-    </Dropdown.Root>
+    <Show when={status() !== NotificationStatus.ARCHIVED}>
```
LetItRock, line 14: when `archived` filter is shown there is no sense in showing mark all actions

packages/js/src/ui/components/elements/Header/MoreActionsOptions.tsx
```diff
@@ -1,31 +1,21 @@
...
-        label="Archive read"
-        /**
-         * TODO: Implement onClick after Filter is implemented
-         */
-        onClick={() => {}}
-        icon={ArchiveRead}
-      />
+      <ActionsItem label="Mark all as read" onClick={() => readAll({ tags: filter().tags })} icon={ReadAll} />
+      <ActionsItem label="Archive all" onClick={() => archiveAll({ tags: filter().tags })} icon={Archive} />
+      <ActionsItem label="Archive read" onClick={() => archiveAllRead({ tags: filter().tags })} icon={ArchiveRead} />
```
LetItRock, line 18: these actions are called per tab (tags)

packages/js/src/ui/context/InboxContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+  [NotificationStatus.UNREAD_READ]: { archived: false },
+  [NotificationStatus.UNREAD]: { read: false },
+  [NotificationStatus.ARCHIVED]: { archived: true },
+};
+
+type InboxProviderProps = ParentProps<{
+  tabs: Array<Tab>;
+}>;
+
+export const InboxProvider = (props: InboxProviderProps) => {
```
LetItRock, line 26: The `InboxProvider` takes the `tabs` as the props passed, calculates the active tab, and builds the initial filter.

packages/js/src/ui/context/InboxContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const [status, setStatus] = createSignal<NotificationStatus>(NotificationStatus.UNREAD_READ);
+  const [filter, setFilter] = createSignal<NotificationFilter>({
+    ...STATUS_TO_FILTER[NotificationStatus.UNREAD_READ],
+    tags: props.tabs.length > 0 ? props.tabs[0].value : [],
+  });
+
+  const setNewStatus = (newStatus: NotificationStatus) => {
+    setStatus(newStatus);
+    setFilter((old) => ({ ...STATUS_TO_FILTER[newStatus], tags: old.tags }));
+  };
```
LetItRock, line 38: when the filter is changed we want to keep old tags for active tab

packages/js/src/ui/context/InboxContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const setNewActiveTab = (newActiveTab: string) => {
+    const tags = tabs().find((tab) => tab.label === newActiveTab)?.value;
+    if (!tags) {
+      return;
+    }
+
+    setActiveTab(newActiveTab);
+    setFilter((old) => ({ ...old, tags }));
+  };
```
LetItRock, line 48: when the active tab is changed we want to update the filter `tags`

packages/js/src/ui/context/InboxContext.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+type InboxProviderProps = ParentProps<{
+  tabs: Array<Tab>;
+}>;
+
+export const InboxProvider = (props: InboxProviderProps) => {
+  const [tabs] = createSignal<Array<Tab>>(props.tabs);
+  const [activeTab, setActiveTab] = createSignal<string>((props.tabs[0] && props.tabs[0].label) ?? '');
+  const [status, setStatus] = createSignal<NotificationStatus>(NotificationStatus.UNREAD_READ);
+  const [filter, setFilter] = createSignal<NotificationFilter>({
```
LetItRock, line 30: the filter is used to query inbox notifications API

packages/js/src/ui/helpers/createInfiniteScroll.ts
```diff
@@ -48,6 +49,18 @@ export function createInfiniteScroll<T>(fetcher: (page: number) => Promise<{ dat
...
+    setData([]);
+    setInitialLoading(true);
+    setEnd(false);
+
+    if (offset() !== 0) {
+      setOffset(0);
+    } else {
+      refetch();
+    }
+  };
```
LetItRock, line 62: The function allows the infinite scroll state to be reset. When the offset is different than 0 it's enough just to update the offset state so it will trigger the fetcher, otherwise, we need to call `refetch` manually which will call the fetcher.

packages/react/src/components/Inbox.tsx
```diff
@@ -39,16 +40,25 @@ export const Inbox = React.memo((props: InboxProps) => {
...
+    onNotificationClick,
+    onPrimaryActionClick,
+    onSecondaryActionClick,
+    ...options
+  } = props;
 
   return (
     <Renderer options={options}>
       <DefaultInbox
+        open={props.open}
```
BiswaViraj, line 56:
```suggestion
        open={open}
```

packages/react/src/components/Inbox.tsx
```diff
@@ -39,16 +40,25 @@ export const Inbox = React.memo((props: InboxProps) => {
...
         renderNotification={renderNotification}
         renderBell={renderBell}
-        onNotificationClick={props.onNotificationClick}
-        onPrimaryActionClick={props.onPrimaryActionClick}
-        onSecondaryActionClick={props.onSecondaryActionClick}
+        onNotificationClick={onNotificationClick}
+        onPrimaryActionClick={onPrimaryActionClick}
+        onSecondaryActionClick={onSecondaryActionClick}
```
BiswaViraj, line 61: It's missing the tabs props

LetItRock: The tabs are propagated to the Renderer which then sets it on the NovuUI and later on the InboxProvider

packages/js/src/ui/components/elements/Header/MoreActionsOptions.tsx
BiswaViraj: 🤓 **nitpick (non-blocking):** ‏I see the value of having these actions called per tabs.
But I think we should also have an action called globally.

BiswaViraj (approved):

packages/js/src/ui/components/InboxTabs/InboxTab.tsx
```diff
@@ -18,14 +22,16 @@ export const InboxTab = (props: { label: string; class?: string }) => {
...
       <span class={style('notificationsTabsTriggerLabel', 'nt-text-sm nt-font-medium')}>{props.label}</span>
-      <span
-        class={style(
-          'notificationsTabsTriggerCount',
-          'nt-rounded-full nt-bg-primary nt-px-[6px] nt-text-primary-foreground nt-text-sm'
-        )}
-      >
-        {count >= 100 ? '99+' : count}
-      </span>
+      <Show when={status() !== NotificationStatus.ARCHIVED}>
```
LetItRock, line 25: no need to show the unread counts when the archived filter is applied (all notifications are read)

packages/js/src/ui/context/CountContext.tsx
```diff
@@ -41,29 +33,28 @@ export const CountProvider = (props: CountProviderProps) => {
...
-        const filter = props.filters[i];
-        const filterKey = createKey(filter);
-        if (!areTagsEqual(filter.tags, notification.tags)) {
+      const allTabs = tabs();
+      for (let i = 0; i < allTabs.length; i++) {
+        const tab = allTabs[i];
+        const tags = tab.value;
+        const allNotifications = tags.length === 0;
+        const includeTags = notification.tags?.every((tag) => tags.includes(tag));
+        if (!allNotifications && !includeTags) {
```
LetItRock, line 50: when the `tags: []` it means that we need to calculate all new messages count; but when it's not empty then we should compare if every notification tag is included in the tab tags

packages/js/src/notifications/notifications.ts
```diff
@@ -242,4 +242,12 @@ export class Notifications extends BaseModule {
...
+    if (filter) {
+      return this.#notificationsCache.clear(filter ?? {});
```
desiprisg, line 248: Should this mean `{}` or `{archived: false}`? I'm guessing the first but just checking.

packages/react/src/components/Inbox.tsx
desiprisg: Yes, @BiswaViraj , the tabs are a prop on the provider. they won't be part of the `mountInbox` options. Essentially, they're passed when you do `new NovuUI`

desiprisg (approved):

packages/js/src/notifications/notifications.ts
LetItRock: the first ;)

PR #6259, opened 2024-08-06
fix(web): redirect to org creation if org is missing

denis-kralj-novu:
redirect wasn't working as expected, now does


apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -102,11 +102,11 @@ export const EnterpriseAuthProvider = ({ children }: { children: React.ReactNode
       if (hasOrgs) {
         const firstOrg = clerkUser.organizationMemberships[0].organization;
         setActive({ organization: firstOrg });
-      } else {
-        navigate(ROUTES.AUTH_SIGNUP_ORGANIZATION_LIST);
```
denis-kralj-novu, line 106: this piece of code was getting hit, but for some reason it wasn't doing the navigate, I have no idea why :/

SokratisVidros (approved):

PR #6268, opened 2024-08-06
refactor(api,application-generic,ws,worker,webhook): resolve circular dependencies; import refactor

ChmaraX:
Changed described in related PR: https://github.com/novuhq/packages-enterprise/pull/188


libs/testing/src/ee/ee.repository.factory.ts
```diff
@@ -1,6 +1,13 @@
 import { CommunityOrganizationRepository, CommunityUserRepository, CommunityMemberRepository } from '@novu/dal';
 import { isClerkEnabled } from '@novu/shared';
 
+/**
+ * We are using nx-ignore-next-line as a workaround here to avoid following circular dependency error:
+ * @novu/application-generic:build --> @novu/testing:build --> @novu/ee-auth:build --> @novu/application-generic:build
+ *
+ * When revising EE testing, we should consider refactoring the code to potentially avoid this circular dependency.
```
ChmaraX, line 8: Currently, doing this would increase the scope too much and would require change of the tests itself.
That said, having circular dependency or skipping a circ dep check in testing package is most of the time harmless.

enterprise/packages/billing/package.json
```diff
@@ -47,5 +48,8 @@
     "@nestjs/swagger": "^7.1.8",
     "@nestjs/throttler": "^5.0.1",
     "@novu/dal": "workspace:*"
+  },
+  "optionalDependencies": {
+    "@novu/ee-auth": "workspace:*"
```
ChmaraX, line 52: This will be changed to direct dependency when we remove `CLERK_ENABLED` flag, since EE === CLERK_ENABLED

apps/api/src/app.module.ts
```diff
@@ -59,12 +59,6 @@ const enterpriseImports = (): Array<Type | DynamicModule | Promise<DynamicModule
     }
   }
 
-  if (isClerkEnabled()) {
-    if (require('@novu/ee-auth')?.EEAuthModule) {
-      modules.push(require('@novu/ee-auth')?.EEAuthModule);
-    }
-  }
-
```
ChmaraX, line 67: This is not needed here because, we always import `AuthModule` which implementation dynamically changes to either EE or community inside the actual module

denis-kralj-novu (approved):

PR #6282, opened 2024-08-08
fix(api): bring back ee user signup analytics tracking

ChmaraX:
- moved SyncExternalUser to @novu/ee-auth package since its related only to that package

Related PR: https://github.com/novuhq/packages-enterprise/pull/189


denis-kralj-novu (approved):
Don't forget to update the EE reference once the EE PR is merged

SokratisVidros (approved):

PR #6289, opened 2024-08-08
fix(api): bring back storage controller for v1 users

ChmaraX:
Bring back storage module and related e2e test for v1 users to enable them to upload branding logo
image


apps/api/src/app/storage/e2e/get-signed-url.e2e.ts
```diff
@@ -1,7 +1,7 @@
 import { UserSession } from '@novu/testing';
 import { expect } from 'chai';
 
-describe('Get Signed Url - /storage/upload-url (GET) @skip-in-ee', function () {
+describe('Get Signed Url - /storage/upload-url (GET)', function () {
```
denis-kralj-novu, line 4: I dont think that this is a problem in EE enabled, the functionality was missing with community edition no?

SokratisVidros (approved):

apps/api/src/app/storage/e2e/get-signed-url.e2e.ts
SokratisVidros: Correct.

PR #6298, opened 2024-08-09
fix(js): Add a minimum height to notification list

desiprisg:

Fixed the 0 height when rendering the empty notice.
Added missing appearance keys that make a lot of sense for customization.
Adjusted the CTA animation not come from outside the element. This keeps it contained and is better even if we decide to export it as a list component.

### Screenshots

image

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/Inbox.tsx
+++ b/packages/js/src/ui/components/Inbox.tsx
@@ -1,4 +1,4 @@
-import { createMemo, createSignal, Match, Show, Switch } from 'solid-js';
+import { createSignal, Match, Show, Switch } from 'solid-js';
 import { useInboxContext } from '../context';
 import { useStyle } from '../helpers';
 import type {
--- a/packages/js/src/ui/components/Inbox.tsx
+++ b/packages/js/src/ui/components/Inbox.tsx
@@ -36,9 +36,10 @@ type InboxContentProps = {
 const InboxContent = (props: InboxContentProps) => {
   const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
   const { tabs, filter } = useInboxContext();
+  const style = useStyle();
 
   return (
-    <>
+    <div class={style('inboxContent', 'nt-h-full nt-flex nt-flex-col')}>
       <Switch>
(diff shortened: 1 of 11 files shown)


BiswaViraj (approved):

LetItRock (commented):
Screenshot 2024-08-12 at 11 22 12

@desiprisg something looks odd here

desiprisg:
@LetItRock Please take another look.

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -36,9 +36,10 @@ type InboxContentProps = {
 const InboxContent = (props: InboxContentProps) => {
   const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
   const { tabs, filter } = useInboxContext();
+  const style = useStyle();
 
   return (
-    <>
+    <div class={style('inboxContent', 'nt-h-full nt-flex nt-flex-col')}>
```
desiprisg, line 42: Add this because it makes sense appearance wise.

packages/js/src/ui/components/Notification/NotificationList.tsx
```diff
@@ -47,24 +35,31 @@ type NotificationListProps = {
...
   const handleOnNewMessagesClick: JSX.EventHandlerUnion<HTMLButtonElement, MouseEvent> = async (e) => {
     e.stopPropagation();
     resetNewMessagesCount();
     refetch({ filter: props.filter });
+    notificationListElement.scrollTo({ top: 0 });
   };
 
   return (
-    <>
+    <div class={style('notificationListContainer', 'nt-h-full nt-overflow-hidden')}>
```
desiprisg: Add this because it makes sense for the structure and we will need it here if we ever export just a list component for the notifications.

desiprisg: `nt-overflow-hidden` is needed for the list below to be scrollable. It sadly clips the CTA animation (because it comes from the top`. Maybe we can make that just fade in.

LetItRock (approved):
amazing work! 🙌

PR #6303, opened 2024-08-12
refactor(webhook,worker,ws): remove redundant dependencies

ChmaraX:
Remove imports and dependencies that are not needed in the app


apps/webhook/src/webhooks/usecases/webhook/webhook.usecase.ts
```diff
@@ -61,15 +54,12 @@ export class Webhook {
       throw new ApiException(`Integration ${integration._id} doesn't have credentials set up`);
     }
 
-    const member = await this.memberRepository.getOrganizationAdminAccount(command.organizationId);
```
ChmaraX, line 64: Every organization must have an admin, this call is redundant and we should remove it everywhere as we already did in some places 

https://novu.slack.com/archives/C0759LER9QT/p1718796984274459

SokratisVidros (approved):

PR #6310, opened 2024-08-12
chore(root): make clean-build script executable

ChmaraX:
Change the file mode from `100644` to `100755` so it can be executed without running `chmod +x` and producing new git change.


rifont (approved):
💯

PR #6312, opened 2024-08-12
refactor(shared): simplify Clerk imports

ChmaraX:
* import types directly from Clerk package wherever possible
* redefined org and user public metadata in global namespace as recommended by Clerk
* removed/replaced redundant types

Related EE PR: https://github.com/novuhq/packages-enterprise/pull/198


libs/shared/src/types/clerk/types.ts
```diff
@@ -0,0 +1,21 @@
+import { ProductUseCases } from '../../dto';
+import { ApiServiceLevelEnum } from '../organization';
+import { IServicesHashes } from '../../entities/user';
+import { JobTitleEnum } from '../organization';
+
+export type _OrganizationPublicMetadata = {
```
rifont: ❓ **question:** What is the intention behind prefixing these with an underscore? It's a new naming pattern we are introducing and I think we should be careful with such decisions. If there isn't a solid reason, I suggest we leave them unprefixed.‏

rifont (approved):
🧽 🏎️

libs/shared/src/types/clerk/types.ts
ChmaraX: It's because we need to redefine global namespace in `env.d.ts` with these types which have the same name. 

I've changed them to be without prefix and instead I used the alias in the import itself to avoid duplicate name error.

```typescript
import type {
  UserPublicMetadata as _UserPublicMetadata,
  OrganizationPublicMetadata as _OrganizationPublicMetadata,
} from '@novu/shared';

declare global {
  interface UserPublicMetadata extends _UserPublicMetadata {}

  interface OrganizationPublicMetadata extends _OrganizationPublicMetadata {}
}
```

rifont: Perfect!

PR #6314, opened 2024-08-12
feat(worker): add environment name to trigger event

djabarovgeorge:
### Screenshots

<details>

</details>


apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
```diff
@@ -244,4 +257,11 @@ export class SubscriberJobBound {
 
     return providers;
   }
+
+  @CachedEntity({
+    builder: (environmentId: string) => buildEnvironmentById({ environmentId }),
+  })
+  private async getEnvironment(environmentId: string) {
+    return await this.environmentRepository.findOne({ _id: environmentId });
```
SokratisVidros: If not mistaken, this is used for easier analytics. May I suggest that instead of caching the query on every time this use case runs, we can store the environment type when we schedule a job and get its name once? For example adding it to IProcessSubscriberDataDto interface.

This approach is much simpler in terms of engineering and has the benefit that the environment name will be stored in Mongo as well, in the jobs collection.

SokratisVidros (changes requested):

apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
djabarovgeorge: Done, i passed the environment name in the queue dto

djabarovgeorge: I'm unsure about storing the environment name in the job entity. do you think it will be useful for us today?

SokratisVidros (approved):

apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts
SokratisVidros: It's just easier for debugging and analytics. Nothing more than that. At some point we could prefix the environment keys with a something that denotes that.

PR #6317, opened 2024-08-13
feat(api): rate limit throttler dry run mode

ChmaraX:
The purpose of the dry run is to allow us to observe how the rate limiting will behave without actually enforcing it.

Later, when everything is confirmed to be working as we expect, we can just remove switch the FF.


apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -153,6 +153,30 @@ export class ApiRateLimitInterceptor extends ThrottlerGuard implements NestInter
...
+      if (!success) {
+        Logger.warn(
+          `[Dry run] ${THROTTLED_EXCEPTION_MESSAGE} orgId: ${organizationId}, envId: ${environmentId}`,
+          'ApiRateLimitInterceptor'
+        );
+      }
+
+      return true;
+    }
+
```
rifont: I think we should move this above the `res.header` setters to avoid any confusion of headers being sent in the response without being acted upon.

Otherwise, looks great!

rifont (changes requested):

apps/api/src/app/rate-limiting/guards/throttler.guard.ts
ChmaraX: Done - also added a bit more info to the log message for potential debugging purpose , since the headers won't be send now

rifont (approved):
Awesome 🚀

PR #6318, opened 2024-08-13
style(web, novui): Add custom scrollbar styling and remove workflow preview overflow

rifont:

* Remove overflow from `PageContainer` to remove scrolling by default
  * Add `overflowY: auto` to pages requiring scrollable content
* Add Panda Global CSS for scrollbar styling

### Screenshots
_Dark Mode_
image

_Light Mode_
image

_Recording demonstrating correct scrolling across the application_
https://www.loom.com/share/01e3c4e328cc42c489f7ca31343c70eb?sid=416b7fce-083c-4d61-a284-eefbb016cefa

<details>

</details>

### DIFF
--- a/apps/web/src/pages/get-started/GetStartedPage.tsx
+++ b/apps/web/src/pages/get-started/GetStartedPage.tsx
@@ -31,7 +31,7 @@ export function GetStartedPage() {
   };
 
   return (
-    <PageContainer>
+    <PageContainer className={css({ overflowY: 'auto' })}>
       <div
         className={css({
           maxWidth: '1000px',
--- a/apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorControlsPanel.tsx
+++ b/apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorControlsPanel.tsx
@@ -142,7 +142,7 @@ export const WorkflowStepEditorControlsPanel: FC<IWorkflowStepEditorControlsPane
 };
 
 export const formContainerClassName = css({
-  h: '80vh',
-  overflowY: 'auto !important',
+  h: '72vh',
+  overflowY: 'auto',
   scrollbar: 'hidden',
 });
--- a/apps/web/src/studio/layout/PageContainer.tsx
+++ b/apps/web/src/studio/layout/PageContainer.tsx
@@ -10,7 +10,6 @@ export const PageContainer: FC<PropsWithChildren<IPageContainerProps>> = ({ chil
     <Container
       className={cx(
         css({
-          overflowY: 'auto !important',
           borderRadius: '0',
           px: 'paddings.page.horizontal',
           py: 'paddings.page.vertical',
--- a/apps/web/src/studio/layout/PageContainer.tsx
+++ b/apps/web/src/studio/layout/PageContainer.tsx
@@ -10,6 +10,7 @@ export const PageContainer: FC<PropsWithChildren<IPageContainerProps>> = ({ chil
     <Container
       className={cx(
         css({
+          overflowX: 'hidden',
           borderRadius: '0',
           px: 'paddings.page.horizontal',
           py: 'paddings.page.vertical',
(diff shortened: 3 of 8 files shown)


apps/web/src/studio/layout/PageContainer.tsx
```diff
@@ -10,7 +10,6 @@ export const PageContainer: FC<PropsWithChildren<IPageContainerProps>> = ({ chil
     <Container
       className={cx(
         css({
-          overflowY: 'auto !important',
```
rifont, line 13: Removing overflow by default. Component consumers now need to explicitly specify when to overflow.

apps/web/src/pages/get-started/GetStartedPage.tsx
```diff
@@ -31,7 +31,7 @@ export function GetStartedPage() {
   };
 
   return (
-    <PageContainer>
+    <PageContainer className={css({ overflowY: 'auto' })}>
```
rifont, line 34: Here and below x2, explicitly setting overflow to be allowed.

libs/novui/src/global.styles.ts
```diff
@@ -14,6 +15,11 @@ export const GLOBAL_CSS = defineGlobalStyles({
 
     backgroundColor: 'surface.page',
     overflow: 'hidden',
+    scrollbarWidth: 'thin',
+    scrollbarColor: {
+      base: `${LEGACY_COLOR_TOKENS.legacy.B80.value} transparent`,
+      _dark: `${LEGACY_COLOR_TOKENS.legacy.B30.value} transparent`,
```
rifont: Panda doesn't support token evaluation for CSS values requiring 2 or more properties to be specified inside a single string and doesn't have native support for these properties ([ref](https://panda-css.com/docs/utilities/interactivity)), so we need to use template literals instead.

libs/novui/src/global.styles.ts
```diff
@@ -14,6 +16,11 @@ export const GLOBAL_CSS = defineGlobalStyles({
 
     backgroundColor: 'surface.page',
     overflow: 'hidden',
+    scrollbarWidth: 'thin',
+    scrollbarColor: {
+      base: `${token('colors.legacy.B80')} transparent`,
+      _dark: `${token('colors.legacy.B30')} transparent`,
```
rifont: Panda doesn't support token evaluation for CSS values requiring 2 or more properties to be specified inside a single string and doesn't have native support for these properties ([ref](https://panda-css.com/docs/utilities/interactivity)), so we need to use template literals instead.

apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorControlsPanel.tsx
```diff
@@ -142,7 +142,7 @@ export const WorkflowStepEditorControlsPanel: FC<IWorkflowStepEditorControlsPane
 };
 
 export const formContainerClassName = css({
-  h: '80vh',
-  overflowY: 'auto !important',
+  h: '72vh',
```
rifont, line 145: With this height at 100% zoom, the step controls was being clipped at the bottom of the page. This patches that. The real fix is to use `flex-grow` in more places across the app, but that's a job for a different PR.

rifont: Fix applied at 100% zoom:
image

antonjoel82: Agreed -- we have some reused or foundational components that broke true flex layouts at the base level, so unfortunately they propagated all the way through. I fixed them in the side nav, but did not go back and address them yet in the content area

libs/novui/src/global.styles.ts
antonjoel82: ✏ **suggestion:** This is also a good code comment!‏

antonjoel82: Also, we should make B80 / B30 into its own semantic color token for `scrollbar`, and then should be able to just do a single template literal here

antonjoel82 (approved):
Thanks for the changes!

libs/novui/src/global.styles.ts
rifont: Great points, I've created new semantic tokens and updated to use those tokens throughout the global css file.

PR #6319, opened 2024-08-14
fix(dal): messages sort order by createdAt

jainpawan21:

`GET /messages` API response is not in createdAt order.
### Screenshots

<details>

</details>


libs/dal/src/repositories/message/message.repository.ts
```diff
@@ -685,12 +685,16 @@ export class MessageRepository extends BaseRepository<MessageDBModel, MessageEnt
...
+      {
+        transactionId: {
+          $in: query.transactionId,
+        },
+        _environmentId: query._environmentId,
       },
-      _environmentId: query._environmentId,
-    })
+      {},
+      { sort: { createdAt: -1 } }
```
ChmaraX, line 696: ✏ **suggestion:** ‏ `messages` collection has 116_756_136 documents on US prod, we might consider adding index on `createdAt` (or possibly compound index with the other fields)

SokratisVidros (approved):
👏

PR #6320, opened 2024-08-14
feat(js): Recalculate notification date each minute

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/Notification/DefaultNotification.tsx
+++ b/packages/js/src/ui/components/Notification/DefaultNotification.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import { JSX, Show } from 'solid-js';
+import { createEffect, createMemo, createSignal, JSX, Show } from 'solid-js';
 import type { Notification } from '../../../notifications';
 import { ActionTypeEnum } from '../../../types';
 import { useInboxContext, useLocalization } from '../../context';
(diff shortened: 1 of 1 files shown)


packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -21,6 +21,20 @@ export const DefaultNotification = (props: DefaultNotificationProps) => {
...
+  const date = createMemo(() => {
+    minutesPassed(); //register as dep
+
+    return formatToRelativeTime({ fromDate: new Date(props.notification.createdAt), locale: locale() });
+  });
+
+  createEffect(() => {
+    const interval = setInterval(() => {
+      setMinutesPassed((prev) => prev + 1);
+    }, 1000);
```
LetItRock: ❓ **question:** WDYT about making it smart and calculating for example the time passed:
- if it's < 60s then update it every second
- if it's >= 1m but < 60m update it every minute
- if it's >= 1h but < 24h update it every hour
Like to control the interval as well so we don't update the DOM unnecessarily, it shouldn't be hard.

LetItRock (approved):

packages/js/src/ui/components/Notification/DefaultNotification.tsx
desiprisg: I think the interval to update shouldn't be tied to the formatting logic

desiprisg: A recalculation of the date every minute is nothing performance wise

desiprisg: We can just do it every second, but I did once per minute as that's what we decided. Also, showing the seconds instead of just now will be spammy for the first minute.

LetItRock: I see the interval is set for every second.

PR #6323, opened 2024-08-14
feat(js): Include headers and tabs in separate components

desiprisg:

We decided to include the whole content of the pages shown in the inbox for the separate components. Also a one line change is possible by uncommenting `InboxContent` to allow mounting the whole content (with routing) for usage in custom popover.
### Screenshots

Screenshot 2024-08-14 at 5 17 51 PM
Screenshot 2024-08-14 at 5 18 08 PM

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/Inbox.tsx
+++ b/packages/js/src/ui/components/Inbox.tsx
@@ -1,4 +1,4 @@
-import { createSignal, Match, Show, Switch } from 'solid-js';
+import { createMemo, createSignal, Match, Show, Switch } from 'solid-js';
 import { useInboxContext } from '../context';
 import { useStyle } from '../helpers';
 import type {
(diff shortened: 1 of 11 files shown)


packages/js/src/ui/components/Inbox.tsx
```diff
@@ -21,28 +21,40 @@ export type InboxProps = {
...
 };
 
-const InboxContent = (props: InboxContentProps) => {
-  const [currentScreen, setCurrentScreen] = createSignal<Screen>(Screen.Inbox);
+export const InboxContent = (props: InboxContentProps) => {
+  const [currentPage, setCurrentPage] = createSignal<InboxPage>(props.initialPage || InboxPage.Notifications);
   const { tabs, filter } = useInboxContext();
   const style = useStyle();
 
+  const navigateToPage = createMemo(() => (page: InboxPage) => {
```
BiswaViraj, line 43: The extra arrow function can be removed. Is there a reason why it needs to return a function ?

```
const navigateToPage = createMemo(page => {
  if(props.hideNav) {
    return
  }
  setCurrentPage(page)
})
```

packages/js/src/ui/components/elements/Header/PreferencesHeader.tsx
```diff
@@ -12,15 +13,19 @@ export const PreferencesHeader = (props: PreferencesHeaderProps) => {
...
-          'preferencesHeader__back__button',
-          'nt-h-6 nt-w-6 nt-flex nt-justify-center nt-items-center nt-rounded-md nt-relative hover:nt-bg-foreground-alpha-50 focus:nt-bg-foreground-alpha-50 nt-text-foreground-alpha-600'
+      <Show when={props.navigateToNotifications}>
+        {(navigateToNotifications) => (
+          <button
+            class={style(
+              'preferencesHeader__back__button',
+              'nt-h-6 nt-w-6 nt-flex nt-justify-center nt-items-center nt-rounded-md nt-relative hover:nt-bg-foreground-alpha-50 focus:nt-bg-foreground-alpha-50 nt-text-foreground-alpha-600'
+            )}
+            onClick={navigateToNotifications()}
```
LetItRock, line 23:
```suggestion
            onClick={navigateToNotifications}
```
❓ **question:** should it be like that?

LetItRock (approved):
looks good to me, just a two small suggestions from me and Biswa ;)

packages/js/src/ui/components/Inbox.tsx
desiprisg: Yes, it's the equivalent of `useCallback` in react.
Essntially `useCallback` is a `useMemo` with an extra arrow function.

packages/js/src/ui/components/elements/Header/PreferencesHeader.tsx
desiprisg: If we use `createMemo`, I believe so.

PR #6326, opened 2024-08-15
feat(framework): Add `preferences` to `workflow` builder

davidsoderberg:


packages/framework/src/resources/workflow.ts
```diff
@@ -376,3 +382,38 @@ function prettyPrintDiscovery(discoveredWorkflow: DiscoverWorkflowOutput): void
     });
   });
 }
+
+function transformPreference(preference?: WorkflowOptionsPreference): DiscoverWorkflowOutputPreference {
```
davidsoderberg: @rifont what do you think about export these help functions to another file/files?

davidsoderberg: Maybe by creating a workflow folder and then make this file index.ts? And move all help functions to separate files?

davidsoderberg: So it would be 8 files instead of one to reduce the length of this one.

rifont: I like the suggestion, it will also promote stronger unit test coverage by nature of having smaller files that are more easily testable.

libs/dal/src/repositories/notification-template/notification-template.entity.ts
```diff
@@ -25,6 +26,19 @@ import type { EnvironmentId } from '../environment';
 import type { ChangePropsValueType } from '../../types';
 import { JSONSchema } from 'json-schema-to-ts';
 
+type WorkflowOptionsPreference = {
+  workflow: {
+    defaultValue: boolean;
+    readOnly: boolean;
+  };
+  channels: {
+    [key in (typeof ChannelTypeEnum)[keyof typeof ChannelTypeEnum]]?: {
```
rifont: Readability improvement:

```suggestion
    [`$ChannelTypeEnum`]?: {
```

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -152,6 +152,7 @@ export class Sync {
               active: this.castToAnyNotSupportedParam(workflow.options)?.active ?? true,
               critical: this.castToAnyNotSupportedParam(workflow.options)?.critical ?? false,
               preferenceSettings: this.castToAnyNotSupportedParam(workflow.options)?.preferenceSettings,
+              preference: workflow.preference,
```
rifont: We should use the plural `preferences` across the board (aligns with existing `tags`, plus it's preferable to keep domain language pluralized, as we do for REST API resource naming)

packages/framework/src/types/discover.types.ts
```diff
@@ -49,6 +50,18 @@ export type DiscoverStepOutput = {
...
 
+export type ChannelPreference = {
+  defaultValue: boolean;
+  readOnly: boolean;
+};
+
+export type DiscoverWorkflowOutputPreference = {
+  workflow: ChannelPreference;
+  channels: {
+    [key in (typeof ChannelTypeEnum)[keyof typeof ChannelTypeEnum]]: ChannelPreference;
```
rifont, line 61: As above, I'd suggest using a typed template literal for better readability

packages/framework/src/types/discover.types.ts
```diff
@@ -73,6 +86,7 @@ export type DiscoverWorkflowOutput = {
     schema: JsonSchema;
     unknownSchema: Schema;
   };
+  preference: DiscoverWorkflowOutputPreference;
```
rifont: Needs to be pluralized too

libs/dal/src/repositories/notification-template/notification-template.entity.ts
davidsoderberg: I tried this but got:
```
A computed property name in a type literal must refer to an expression whose type is a literal type or a 'unique symbol' type.ts(1170)
'ChannelTypeEnum' will evaluate to '[object Object]' when stringified.
```

packages/framework/src/resources/workflow/transform-preference.ts
rifont, line 1: "transform" is a bit misleading as a name, I would suggest renaming to `buildPreferences`

packages/framework/src/types/discover.types.ts
davidsoderberg: See above

libs/dal/src/repositories/notification-template/notification-template.entity.ts
rifont: Interesting, I would've expected Typescript to handle that! Let's go with the original then.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
```diff
@@ -79,7 +79,9 @@ export class UpdateWorkflow {
         `Notification template with id ${command.id} not found`
       );
 
-    let updatePayload: Partial<NotificationTemplateEntity> = {};
+    let updatePayload: Partial<NotificationTemplateEntity> = {
+      preferences: command.preferences,
+    };
```
rifont: It's possible that older `@novu/framework` versions are used during sync that don't have `command.preferences`, therefore we should follow the existing pattern of checking whether the property exists and add the property to the payload conditionally.

packages/framework/src/resources/workflow/workflow.test.ts
```diff
@@ -99,6 +99,36 @@ describe('workflow function', () => {
     });
   });
 
+  it('should include preferences', async () => {
+    const { definition } = workflow(
+      'setup-workflow',
+      async ({ step }) => {
+        // @ts-expect-error - email subject is missing from the output
+        await step.email('send-email', async () => ({
+          body: 'Test Body',
```
rifont: No need to keep this `ts-expect-error` directive, it was for a different test.

```suggestion
        await step.email('send-email', async () => ({
          subject: 'Test Subject',
          body: 'Test Body',
```

rifont (approved):
Awesome work!

PR #6330, opened 2024-08-15
feat(novui, web, framework): Step control autocomplete

ainouzgali:

_Based off of reverted PR that first introduced auto-complete_ #6097

- Setup autocomplete for Step Controls using `tiptap`
- Hide behavior behind [IS_CONTROLS_AUTOCOMPLETE_ENABLED](https://app.launchdarkly.com/projects/default/flags/IS_CONTROLS_AUTOCOMPLETE_ENABLED) flag

### Screenshots

https://www.loom.com/share/16c8772111a5408faeca9f992a564236?sid=0804ef41-9745-4775-a6a2-130f5ab14652


apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
```diff
@@ -583,7 +583,7 @@ contexts.forEach((context: Context) => {
               controlSchema: {
                 type: 'object',
                 properties: {
-                  name: { type: 'string', default: 'Hello {{name}}' },
+                  name: { type: 'string', default: 'Hello {{payload.name}}' },
```
rifont, line 586: Could we duplicate this test and add a test with the previous `{{name}}` syntax to verify backwards compatibility?

apps/web/src/components/layout/components/LocalStudioSidebar/LocalStudioSidebarOrganizationDisplay.tsx
```diff
@@ -24,6 +24,9 @@ export const LocalStudioSidebarOrganizationDisplay: FC<LocalStudioSidebarOrganiz
       offset={0}
       withinPortal
       title="Novu Local Studio"
+      classNames={{
+        dropdown: css({ bg: 'surface.popover !important', border: 'none !important', shadow: 'medium !important' }),
```
rifont, line 28: 👏 **praise:** Nice use of semantic tokens!‏

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
```diff
@@ -1,13 +1,20 @@
...
+import { Text } from '@novu/novui';
 import { hstack } from '@novu/novui/patterns';
+import { Center, Stack } from '@novu/novui/jsx';
+import { IconErrorOutline, IconExpandLess, IconExpandMore } from '@novu/novui/icons';
 
 export function ErrorPrettyRender({ error: unparsedError }) {
   const [isExpanded, { toggle }] = useDisclosure();
   const error = 'response' in unparsedError ? unparsedError?.response?.data : unparsedError;
+  // TODO: error framework types enum
+  const isInvalidControlSyntax = error?.code === 'CompilingStepControlInvalidError';
```
rifont: ☑ **chore:** Please add `@novu/framework` as a dependency of `@novu/web` and import the Error enum here.‏

apps/web/src/index.tsx
```diff
@@ -7,7 +7,6 @@ import reportWebVitals from './reportWebVitals';
 import { LAUNCH_DARKLY_CLIENT_SIDE_ID } from './config';
 
 import './index.css';
-import '@novu/novui/components.css';
```
rifont, line 10: ❓ **question:** Why did we remove this import‏

apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorControlsPanel.tsx
```diff
@@ -48,6 +49,16 @@ export const WorkflowStepEditorControlsPanel: FC<IWorkflowStepEditorControlsPane
     );
   }, [workflow?.payload?.schema, workflow?.options?.payloadSchema, workflow?.payloadSchema]);
 
+  const payloadProperties = useMemo(() => {
+    const payloadObject =
+      workflow?.payload?.schema?.properties ||
+      workflow?.options?.payloadSchema?.properties ||
+      workflow?.payloadSchema?.properties ||
```
rifont, line 49: Ouch. `workflow?.payload?.schema?.properties` is the correct attribute to access, we need to deprecate these other 2. I'll create a ticket.

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+const AUTOCOMPLETE_CLOSE_TAG = '}}';
+
+const VALID_VARIABLE_REGEX = new RegExp(`${AUTOCOMPLETE_OPEN_TAG}([^{}]*?)${AUTOCOMPLETE_CLOSE_TAG}`, 'g');
+
+// To get initial invalid syntax of {{..} or {{..<space>
+const INVALID_VARIABLE_REGEX = new RegExp(
+  `(${AUTOCOMPLETE_OPEN_TAG}[^{}|\\s]*[}|\\s](?!})|{{[^{}]*}(?!}})|{{[^{}]*$)`,
+  'g'
+);
+const AUTOCOMPLETE_REGEX_V2 = new RegExp(`${AUTOCOMPLETE_OPEN_TAG}(.*?(.*?))${AUTOCOMPLETE_CLOSE_TAG}`, 'm');
```
rifont: ✏ **suggestion:** I don't think the `_V2` suffix is necessary here, we should avoid such versioning bleed into the design system.‏
```suggestion
const AUTOCOMPLETE_REGEX = new RegExp(`${AUTOCOMPLETE_OPEN_TAG}(.*?(.*?))${AUTOCOMPLETE_CLOSE_TAG}`, 'm');
```

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+export function Mention({ editor, node, updateAttributes, ...props }: NodeViewProps) {
+  const [hasError, setHasError] = useState(node.attrs.error || false);
+
+  const updateVariableAttributes = (label: string, error: boolean) => {
+    updateAttributes({ label, error: `${error}` });
+    setHasError(error);
+  };
+
+  //In case of an invalid variable input, set content to be the label variable  which will include the whole invalid variable
```
rifont: ☑ **chore:** Missing space‏
```suggestion
  // In case of an invalid variable input, set content to be the label variable  which will include the whole invalid variable
```

libs/novui/src/tokens/semanticColors.tokens.ts
```diff
@@ -203,6 +203,20 @@ export const LEGACY_COLOR_SEMANTIC_TOKENS = defineSemanticTokens.colors({
       type: 'color',
     },
   },
+  variable: {
+    surface: {
+      value: { base: '{colors.legacy.BGLight}', _dark: '{colors.legacy.B20}' },
+      type: 'color',
+    },
+    text: {
+      value: { base: '#B743FF', _dark: '#AD74FF' },
```
rifont, line 212: ❓ **question:** Do we have a semantic color token we can use here?‏

packages/framework/src/client.ts
```diff
@@ -620,14 +621,20 @@ export class Client {
...
+        payload: event.payload || event.data,
+        subscriber: event.subscriber,
+        // Backwards compatibility, for allowing usage of variables without namespace (e.g. `{{name}}` instead of `{{payload.name}}`)
+        ...(event.payload || event.data),
+      });
 
-    return JSON.parse(compiledString);
+      return JSON.parse(compiledString);
+    } catch (error) {
+      throw new CompilingStepControlInvalidError(event.workflowId, event.stepId);
```
rifont: ⚠ **issue:** Even though we're not consuming it in the UI right now, we should include the `error` on the `data` attribute of the error to ensure that incorrect Liquid syntax will be reported to the Worker engine in the event of incorrect syntax finding it's way into template parsing.‏

packages/framework/src/client.ts
```diff
@@ -620,14 +621,20 @@ export class Client {
...
+      const templateString = this.templateEngine.parse(JSON.stringify(templateControls));
 
-    const compiledString = await this.templateEngine.render(templateString, {
-      ...(event.payload || event.data),
-      subscriber: event.subscriber,
-    });
+      const compiledString = await this.templateEngine.render(templateString, {
+        payload: event.payload || event.data,
+        subscriber: event.subscriber,
+        // Backwards compatibility, for allowing usage of variables without namespace (e.g. `{{name}}` instead of `{{payload.name}}`)
```
rifont, line 650: 👏 **praise:** Great comment!‏

libs/novui/package.json
```diff
@@ -131,13 +130,21 @@
...
+    "@rjsf/utils": "^5.20.0",
     "@rjsf/validator-ajv8": "^5.17.1",
     "@tanstack/react-table": "^8.17.3",
+    "@tiptap/extension-document": "^2.5.0",
+    "@tiptap/extension-mention": "^2.5.5",
+    "@tiptap/extension-paragraph": "^2.5.0",
+    "@tiptap/extension-text": "^2.5.0",
+    "@tiptap/pm": "^3.0.0",
+    "@tiptap/react": "^3.0.0",
+    "@tiptap/suggestion": "^2.5.0",
```
antonjoel82: ❓ **question:** Should all of these be set to `2.6.4` which is the latest release? I don't see a version 3.0.0, and it's generally good to have all related packages on the same version (as seen with Mantine)‏

libs/novui/src/recipes/input-editor-widget.recipe.ts
```diff
@@ -0,0 +1,68 @@
...
+        },
+
+        _error: {
+          color: 'input.border.error',
+          border: 'none',
+          background: 'none',
+          textDecoration: 'underline',
+          textDecorationStyle: 'wavy',
+          textDecorationColor: 'input.border.error',
+          textUnderlineOffset: '5px',
```
antonjoel82:
```suggestion
          textUnderlineOffset: '25',
```

libs/novui/src/recipes/variable-suggestion-list.recipe.ts
```diff
@@ -0,0 +1,34 @@
...
+      width: 'components.menu.width !important',
+      overflowY: 'auto',
+    },
+
+    options: {
+      overflowX: 'hidden !important',
+    },
+    option: {
+      ...SELECT_RECIPE.base.option,
+      // overflowX: 'hidden !important',
```
antonjoel82:
```suggestion
```

libs/novui/src/tokens/semanticColors.tokens.ts
antonjoel82, line 213:
```suggestion
      // TODO: Design owes us base tokens / palette values to use here instead
      value: { base: '#B743FF', _dark: '#AD74FF' },
```

libs/novui/src/tokens/semanticSizes.tokens.ts
```diff
@@ -14,6 +14,16 @@ export const SEMANTIC_SIZES_TOKENS = defineSemanticTokens.sizes({
         type: 'sizes',
       },
     },
+    menu: {
```
antonjoel82, line 17: 👏 **praise:** Thank you for adding these! ‏

packages/framework/src/client.test.ts
```diff
@@ -539,9 +539,9 @@ describe('Novu Client', () => {
               controlSchema: {
                 type: 'object',
                 properties: {
-                  name: { type: 'string', default: '{{name}}' },
+                  name: { type: 'string', default: '{{payload.name}}' },
                   lastName: { type: 'string', default: '{{subscriber.lastName}}' },
-                  role: { type: 'string', default: '{{role}}' },
+                  role: { type: 'string', default: '{{payload.role}}' },
```
antonjoel82, line 543: ❓ **question:** Do we need backwards compatibility tests here as Richard mentioned above?‏

packages/framework/src/errors/execution.errors.ts
```diff
@@ -64,6 +64,16 @@ export class ExecutionStateResultInvalidError extends BadRequestError {
   }
 }
 
+export class CompilingStepControlInvalidError extends BadRequestError {
+  code = ErrorCodeEnum.COMPILING_STEP_CONTROL_INVALID_ERROR;
+
+  constructor(workflowId: string, stepId: string) {
```
antonjoel82: 💭 **thought (non-blocking):** General suggestion (that would break / change current convention) -- use an object as constructor params. 

A few reasons:

1. Types can then be more easily reusable
2. Makes it harder to accidentally switch the order of params of the same type (i.e. user calls with `stepId`, then `workflowId`)
3. More extensible (for a large number of fields)
4. Not subject to changes in order, especially for optional params

CC: @rifont ‏

apps/web/src/studio/components/workflows/step-editor/WorkflowsStepEditorPage.tsx
antonjoel82, line 1: ☑ **chore:** Delete this file since it seems to be empty‏

apps/web/src/index.tsx
antonjoel82: @rifont We've consolidated the css to a single export from `novui`, so it's now covered by the `index.css` import above

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
antonjoel82: ❓ **question:** @rifont is there another way we can think to do this? It seems like a large overkill to add a whole new dependency to our primary app for a couple of string enum values‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  result = result?.replace(INVALID_VARIABLE_REGEX, (match) => {
+    return `<mention-component data-label="${match}" data-id="${match}" data-error="true"></mention-component>`;
+  });
+
+  return result;
+};
+
+export function Mention({ editor, node, updateAttributes, ...props }: NodeViewProps) {
+  const [hasError, setHasError] = useState(node.attrs.error || false);
```
antonjoel82:
```suggestion
  const [hasError, setHasError] = useState(node.attrs.error || false);
```
```suggestion
  const [hasError, setHasError] = useState<boolean>(node.attrs.error || false);
```

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const updateVariableAttributes = (label: string, error: boolean) => {
```
antonjoel82:
```suggestion
  const updateVariableAttributes = (label: string, hasError: boolean) => {
```

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+      : `${AUTOCOMPLETE_OPEN_TAG}${node.attrs.label ?? node.attrs.id}${AUTOCOMPLETE_CLOSE_TAG}`;
+  },
+  renderHTML({ HTMLAttributes, node }) {
+    const htmlOutput =
+      node.attrs.error === 'true'
+        ? node.attrs.label
+        : `${AUTOCOMPLETE_OPEN_TAG}${node.attrs.label ?? node.attrs.id}${AUTOCOMPLETE_CLOSE_TAG}`;
+
+    return ['mention-component', mergeAttributes(HTMLAttributes), htmlOutput];
+  },
```
antonjoel82: ✏ **suggestion:** Code comments would be helpful here to explain what these tiptap-specific functions are doing‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+  });
+
+  result = result?.replace(INVALID_VARIABLE_REGEX, (match) => {
+    return `<mention-component data-label="${match}" data-id="${match}" data-error="true"></mention-component>`;
+  });
+
+  return result;
+};
+
+export function Mention({ editor, node, updateAttributes, ...props }: NodeViewProps) {
```
antonjoel82: ☑ **chore:** Please extract to separate file as this is a standalone construct‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+          }
+        }}
+      >
+        {renderedNodeViewContent}
+      </NodeViewContent>
+    </NodeViewWrapper>
+  );
+}
+
+export const CustomMention = ExternalMention.extend({
```
antonjoel82: ☑ **chore:** Please extract to separate file as this is a standalone construct‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+      onChange(content, raiseError);
+    },
+  });
+
+  useEffect(() => {
+    if (editor) {
+      const output = getInitContentWithVariableNodeView(value?.toString());
+
+      // Set timeout is for a known tiptap error with setting content on initial render (ADD ISSUE URL)
```
antonjoel82: ☑ **chore:** Please add the ISSUE URL!‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,264 @@
...
+      setTimeout(() => {
+        editor.commands.setContent(output);
+      });
+    }
```
antonjoel82: ⚠ **issue:** Have to clear the timeout on unmount‏
```suggestion
    if (!editor) {
      return;
    }
    
    const output = getInitContentWithVariableNodeView(value?.toString());

    // Set timeout is for a known tiptap error with setting content on initial render (ADD ISSUE URL)
    const timeoutId = setTimeout(() => {
      editor.commands.setContent(output);
    });
    
    return () => {
      clearTimeout(timeoutId);
    }
  }
```

libs/novui/src/recipes/input-editor-widget.recipe.ts
antonjoel82:
```suggestion
          textUnderlineOffset: '25',
```

antonjoel82 (commented):
Leaving code feedback and will do functionality testing / feedback shortly!

antonjoel82:
Tested and works wonderfully! 

The one question that I still have is: what should happen when an invalid variable is entered? I.e. I type `{{asdf` and then add `}}` myself? Or `{{}}`?
  Screenshot 2024-08-15 at 2 25 32 PM

It feels to me like this should be marked invalid.

libs/novui/src/recipes/variable-suggestion-list.recipe.ts
ainouzgali: I wanted to ask you about this one. I need this `overflowX: 'hidden !important',` for a long variable name in the dropdown to have the 3 ellipsis .But then novui is not compiling with the error:
`Property 'overflowX' is incompatible with index signature.
            Type 'string' is not assignable to type 'Nested<(SystemProperties | GenericProperties) & CssVarProperties>'.`

 But I can not find a different place to have that overflow within the slot recipe

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
ainouzgali: oops, there used to be eight of those ;) thanks!

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
ainouzgali: Yes, thats why I wasn't sure if I should and left it as a todo. Maybe we could move some types to the shared lib? For this and for future things of this sort?

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
ainouzgali: yes, definitely. I will add some when I come back from vacation ❤️

ainouzgali:
Thanks @antonjoel82 :)
I am not sure what you mean here 

> what should happen when an invalid variable is entered? I.e. I type {{asdf and then add }} myself? Or {{}}?

Why do you feel it should be invalid? The result of that is a valid variable for liquid syntax

libs/novui/src/recipes/variable-suggestion-list.recipe.ts
antonjoel82: This may be due to the strict tokens setting. Can you please try `[hidden] !important`? (Using the escape hatch syntax)

antonjoel82:
> Thanks @antonjoel82 :) I am not sure what you mean here
> 
> > what should happen when an invalid variable is entered? I.e. I type {{asdf and then add }} myself? Or {{}}?
> 
> Why do you feel it should be invalid? The result of that is a valid variable for liquid syntax

Sorry! I mean to that while `{{asdf}}` and `{{}}` are _syntactically correct_ (meaning that they follow the RegEx), they do not contain valid variable references (which we have a list of to check against). Is there a use case for allowing arbitrary variable names here?

libs/novui/src/json-schema-components/utils.ts
```diff
@@ -2,6 +2,19 @@ import { JsonSchemaFormSectionVariant } from '../../styled-system/recipes';
...
+export const AUTOCOMPLETE_REGEX = new RegExp(`${AUTOCOMPLETE_OPEN_TAG}(.*?(.*?))${AUTOCOMPLETE_CLOSE_TAG}`, 'm');
+
+export const VALID_VARIABLE_REGEX = new RegExp(`${AUTOCOMPLETE_OPEN_TAG}([^{}]*?)${AUTOCOMPLETE_CLOSE_TAG}`, 'g');
+
+// To get initial invalid syntax of {{..} or {{..<space>
+export const INVALID_VARIABLE_REGEX = new RegExp(
+  `(${AUTOCOMPLETE_OPEN_TAG}[^{}|\\s]*[}|\\s](?!})|{{[^{}]*}(?!}})|{{[^{}]*$)`,
+  'g'
+);
+
```
antonjoel82: ✏ **suggestion:** I believe I made a mistake by putting `JSON_SCHEMA_FORM_ID_DELIMITER` here -- could we please move that and all these RegEx-related values to a separate `constants` file?‏

libs/novui/src/json-schema-components/widgets/InputEditorWidget.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+      onChange(content, raiseError);
+    },
+  });
+
+  useEffect(() => {
+    if (editor) {
+      const output = getInitContentWithVariableNodeView(value?.toString());
+
+      // Set timeout is for a known tiptap error with setting content on initial render https://github.com/ueberdosis/tiptap/issues/3764
```
antonjoel82, line 136: Just saw this https://github.com/ueberdosis/tiptap/issues/3580#issuecomment-2285576323

antonjoel82 (approved):
Nice job, Gali! Approving since my outstanding question about arbitrary variables is more of a product question and can be addressed separately if necessary

libs/novui/src/recipes/variable-suggestion-list.recipe.ts
ainouzgali: thats also not working 😔

ainouzgali:
> > Thanks @antonjoel82 :) I am not sure what you mean here
> > > what should happen when an invalid variable is entered? I.e. I type {{asdf and then add }} myself? Or {{}}?
> > 
> > 
> > Why do you feel it should be invalid? The result of that is a valid variable for liquid syntax
> 
> Sorry! I mean to that while `{{asdf}}` and `{{}}` are _syntactically correct_ (meaning that they follow the RegEx), they do not contain valid variable references (which we have a list of to check against). Is there a use case for allowing arbitrary variable names here?

It was decided to be like that in [this slack discussion](https://novu.slack.com/archives/C06UMR2M5U5/p1723117735768939)

apps/web/src/components/workflow/preview/ErrorPrettyRender.tsx
rifont: I was mistaken, we already have `@novu/framework` as a dependency of `@novu/web`.

We already have a number of places that benefits from use Framework typings, primarily the discovery types. We will continue to need such typings so it's not unreasonable to have this dependency in place. Framework defines the source of truth for such values, there's no need to extract to shared lib.

packages/framework/src/client.test.ts
```diff
@@ -519,6 +519,79 @@ describe('Novu Client', () => {
...
+      expect(emailExecutionResult).toBeDefined();
+      expect(emailExecutionResult.outputs).toBeDefined();
+      if (!emailExecutionResult.outputs) throw new Error('executionResult.outputs is undefined');
+      const subject = emailExecutionResult.outputs.subject;
+      expect(subject).toBe("body static prefix `default_name` Smith's product manager");
+      const body = emailExecutionResult.outputs.body;
+      expect(body).toContain('cat');
+      expect(body).toContain('dog');
+    });
+    it('should compile default control variable with backwards compatability for payload variables', async () => {
```
rifont:
```suggestion
    });

    it('should compile default control variable with backwards compatability for payload variables', async () => {
```

libs/novui/src/json-schema-components/widgets/Variable.tsx
```diff
@@ -0,0 +1,90 @@
...
+        }}
+        onBlur={(event: FocusEvent<HTMLInputElement>) => {
+          const text = event.target.innerText;
+          const variable = getValidatedVariable(text, variablesSet);
+
+          if (variable) {
+            updateVariableAttributes(variable);
+          } else {
+            updateVariableAttributes(text, VariableErrorCode.INVALID_NAME);
+          }
```
ainouzgali, line 65: ❓ **question:** Couldn't the else also be an INVALID_SYNTAX ?  For example if not having closing brackets `}}`

Screenshot 2024-08-25 at 13 58 28‏

libs/novui/src/json-schema-components/widgets/VariableSuggestionList.tsx
```diff
@@ -0,0 +1,162 @@
...
+        if (event.key === 'ArrowRight') {
+          customVariableLabel();
+
+          return true;
+        }
+        if (event.code === 'Space') {
+          // don't allow Space to close variable entry if the user hasn't typed the closing tag
+          if (!query.endsWith(AUTOCOMPLETE_CLOSE_TAG)) {
+            return false;
+          }
```
ainouzgali, line 98: 👏 **praise:** Nice!!‏

libs/novui/src/json-schema-components/widgets/Variable.tsx
```diff
@@ -0,0 +1,90 @@
...
+    return;
+  }
+
+  const variableName = variableContent.split(LIQUID_FILTER_CHAR)[0].trim();
+
+  if (!possibleVariables.has(variableName)) {
+    return;
+  }
+
+  return variableName;
```
ainouzgali:
```suggestion
  return variableContent;
```
Otherwise the liquid filter will not be applied to the rendered email

packages/framework/src/client.ts
```diff
@@ -640,14 +641,21 @@ export class Client {
...
-      ...(event.payload || event.data),
-      subscriber: event.subscriber,
-    });
+    try {
+      const templateString = this.templateEngine.parse(JSON.stringify(templateControls));
+
+      const compiledString = await this.templateEngine.render(templateString, {
+        payload: event.payload || event.data,
+        subscriber: event.subscriber,
+        controls: event.controls,
```
ainouzgali: ⚠ **issue:** If there is a use of a control variable that also uses a variable, the nested use of variables will not be compiled. 
Example:
`payload: {name: "Gali"}`
`var1: "Hello {{payload.name}}"` 
 and then `var2: "Hola {{controls.var1}}"` 
That var2 is rendered in the result as `"Hola Hello {{payload.name}}"`
instead of `"Hola Hello Gali"`
‏

ainouzgali (commented):
Thanks for picking this up @antonjoel82 , it looks great.

PR #6331, opened 2024-08-15
feat(api): add basic RBAC

ChmaraX:
* De-registered `RolesGuard` from being global guard, otherwise we wouldn't be able to access the `user` object inside the `RolesGuard` due to how [NestJS request lifecycle works](https://docs.nestjs.com/faq/request-lifecycle#summary):
```
1. Incoming request
2. Middleware
  2.1. Globally bound middleware
  2.2. Module bound middleware
3. Guards
  3.1 Global guards
  3.2 Controller guards
  3.3 Route guards
4. Interceptors (pre-controller)
  4.1 Global interceptors
  4.2 Controller interceptors
  4.3 Route interceptors
```

Since our `UserGuard` is Controller/Route guard and is also responsible for creating `User` object, it can't get executed later than RolesGuard, therefore RolesGuard need to be moved to Controller/Routes or alternatively interceptor, but thats probably not a good practice.  

* Moved `Roles` decorator and `RolesGuard` to `@novu/application-generic`, so it can be imported outside of `apps/api` context
* Added role decorator to specified endpoints which can be accessed only by admin
* ⚠️  Removed the ability to be invoked using API key from `/api-keys/regenerate` and `/billing/checkout` endpoints

Related enterprise PR: https://github.com/novuhq/packages-enterprise/pull/200


apps/api/src/metadata.ts
denis-kralj-novu, line 1: You should not have to commit this anymore, a different procedure is present for generating this file. Ensure your branch is rebased from the latest state of next and see how it changes

apps/api/src/app/workflows/workflow.controller.ts
```diff
@@ -132,6 +133,8 @@ export class WorkflowController {
   })
   @ExternalApiAccessible()
   deleteWorkflowById(@UserSession() user: UserSessionData, @Param('workflowId') workflowId: string): Promise<boolean> {
+    console.log('im i here');
```
denis-kralj-novu: remove this

apps/api/src/app/integrations/integrations.controller.ts
```diff
@@ -173,6 +175,7 @@ export class IntegrationsController {
   }
 
   @Put('/:integrationId')
+  @UseGuards(RolesGuard)
```
denis-kralj-novu: are we sure we want to add this here? I would actually remove the @Roles decorator instead, given that we didn't check roles up until now

apps/api/src/app/integrations/integrations.controller.ts
```diff
@@ -213,6 +216,7 @@ export class IntegrationsController {
   }
 
   @Post('/:integrationId/set-primary')
+  @UseGuards(RolesGuard)
```
denis-kralj-novu: similar consideration as above

apps/api/src/app/organization/ee.organization.controller.ts
```diff
@@ -75,6 +75,7 @@ export class EEOrganizationController {
 
   @Patch('/')
   @ExternalApiAccessible()
+  @UseGuards(RolesGuard)
```
denis-kralj-novu: is this even possible with clerk do do it via this endpoint? I thought clerk managed these changes

apps/api/src/app/organization/organization.controller.ts
```diff
@@ -114,6 +115,7 @@ export class OrganizationController {
   @SdkGroupName('Organizations.Members')
   @Delete('/members/:memberId')
   @ExternalApiAccessible()
+  @UseGuards(RolesGuard)
```
denis-kralj-novu: this file also seems to be only relevant for community edition

apps/api/src/app/workflow-overrides/workflow-overrides.controller.ts
```diff
@@ -180,7 +180,7 @@ export class WorkflowOverridesController {
   }
 
   @Delete('/:overrideId')
-  @UseGuards(RootEnvironmentGuard)
+  @UseGuards(RootEnvironmentGuard, RolesGuard)
```
denis-kralj-novu: same as before, would actually consider removing the `@Roles` altogether

apps/api/src/app/workflows/notification-template.controller.ts
denis-kralj-novu, line 1: again, none of these are in the definition of done for the RBAC rights

apps/api/src/app/workflows/workflow.controller.ts
```diff
@@ -121,7 +122,7 @@ export class WorkflowController {
   }
 
   @Delete('/:workflowId')
-  @UseGuards(RootEnvironmentGuard)
+  @UseGuards(RootEnvironmentGuard, RolesGuard)
```
denis-kralj-novu: not in DoD

denis-kralj-novu (changes requested):

apps/api/src/app/workflows/workflow.controller.ts
ChmaraX: 🤣

apps/api/src/app/integrations/integrations.controller.ts
ChmaraX: I agree, I wanted to keep the original implementation that was there before but then it would be more than defined in DoD - will remove

apps/api/src/app/organization/ee.organization.controller.ts
ChmaraX: Yes it calls update on Clerk - it was originally left here for backwards compatibility, when the intent was to keep all the endpoints as they were before.

We should probably inspect usage and consider removing this.

denis-kralj-novu (approved):

PR #6339, opened 2024-08-16
fix(api,application-generic): assign user to logger context

ChmaraX:
Currently, when we try to log something before the request completes, we don't have the user object and other properties present in the log because they're not yet available in the `customProps` context:

image

To assign custom properties and have them available for the entire request duration. we need to call `assign()` on a PinoLogger instance: https://github.com/iamolegga/nestjs-pino?tab=readme-ov-file#assign-extra-fields-for-future-calls  

image

Related EE PR: https://github.com/novuhq/packages-enterprise/pull/201

### DIFF
--- a/apps/api/src/app/rate-limiting/guards/throttler.guard.ts
+++ b/apps/api/src/app/rate-limiting/guards/throttler.guard.ts
@@ -177,7 +177,6 @@ export class ApiRateLimitInterceptor extends ThrottlerGuard implements NestInter
         apiServiceLevel
       )
     );
-    res.rateLimitPolicy = rateLimitPolicy;
 
     if (success) {
       return true;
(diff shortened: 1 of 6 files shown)


apps/api/src/app/rate-limiting/guards/throttler.guard.ts
```diff
@@ -174,15 +177,7 @@ export class ApiRateLimitInterceptor extends ThrottlerGuard implements NestInter
...
-    res.rateLimitPolicy = {
-      limit,
-      windowDuration,
-      burstLimit,
-      algorithm,
-      apiRateLimitCategory,
-      apiRateLimitCost,
-      apiServiceLevel,
-    };
+    res.rateLimitPolicy = rateLimitPolicy;
```
rifont: This line isn't required anymore, it was only used for the `customProps` downstream logging context. We should remove it.
```suggestion
```

rifont (approved):
💯

PR #6345, opened 2024-08-17
feat(web): Change UI state based on roles

denis-kralj-novu:
- Non admin in enterprise cannot regenerate apiKey
- Non Admin in enterprise cannot modify plan details

## Billing page for editor
image

## ApiKey page for editor
image

## Organization settings page for editor
image

## Organization team members page for editor
image

### DIFF
--- a/apps/web/src/components/providers/AuthProvider.tsx
+++ b/apps/web/src/components/providers/AuthProvider.tsx
@@ -39,6 +39,8 @@ export type AuthContextValue = UserState &
     redirectToSignUp: (params: { redirectURL?: string; origin?: string; anonymousId?: string }) => void;
     switchOrganization: (organizationId: string) => Promise<void>;
     reloadOrganization: () => Promise<{}>;
+    isCurrentUserAdministrator: boolean;
+    isCurrentUserEditor: boolean;
   };
 
 export const AuthProvider = ({ children }: { children: React.ReactNode }) => {
--- a/apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
+++ b/apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
@@ -9,6 +9,7 @@ import { useNavigate } from 'react-router-dom';
 import { useQueryClient } from '@tanstack/react-query';
 import { useSegment } from '../../../components/providers/SegmentProvider';
 import { ROUTES } from '../../../constants/routes';
+import { CLERK_ROLES } from './constants';
 
 const asyncNoop = async () => {};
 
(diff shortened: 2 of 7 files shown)


apps/web/src/ee/clerk/providers/EnterpriseAuthProvider.tsx
```diff
@@ -114,6 +115,9 @@ export const EnterpriseAuthProvider = ({ children }: { children: React.ReactNode
     [clerkOrganization]
   );
 
+  const isCurrentUserAdministrator = useMemo(() => (has ? has({ role: CLERK_ROLES.ADMINISTRATOR }) : false), [has]);
+  const isCurrentUserEditor = useMemo(() => (has ? has({ role: CLERK_ROLES.EDITOR }) : false), [has]);
```
antonjoel82, line 119:
```suggestion
  const isCurrentUserAdministrator = useMemo(() => (has?.({ role: CLERK_ROLES.ADMINISTRATOR }) ?? false), [has]);
  const isCurrentUserEditor = useMemo(() => (has?.({ role: CLERK_ROLES.EDITOR }) ?? false);
```

apps/web/src/components/providers/AuthProvider.tsx
```diff
@@ -39,6 +39,8 @@ export type AuthContextValue = UserState &
     redirectToSignUp: (params: { redirectURL?: string; origin?: string; anonymousId?: string }) => void;
     switchOrganization: (organizationId: string) => Promise<void>;
     reloadOrganization: () => Promise<{}>;
+    isCurrentUserAdministrator: boolean;
+    isCurrentUserEditor: boolean;
```
antonjoel82, line 43: I would recommend that this is where we should draw the distinction of Selectors in our code. Rather than bloating state with a number of flags (which will continue to come up), we can use functions that get the derived data based on the state. 

Then you would use some like `selectIsCurrentUserAdmin` instead.

antonjoel82 (approved):
Approving for now since this seems fine, but I do want to call-out that moving towards selectors for derived data is a cleaner, more extensible pattern. If we decide to do it in a follow-up ticket, can you please attach the Linear ticket?

apps/web/src/components/providers/AuthProvider.tsx
denis-kralj-novu: so if I understand correctly, the point is not to bloat the state to have this count of flags but instead supply just one function that can ask "is this user of this role"? That does seem like a good approach.

I fail to see how `selectIsCurrentUserAdmin` would help with that though, because then I am switching a flag for a function.

The way I understood the suggestion is that I can have a this function in the auth state:
```js
// signature
isCurrentUserInRole(role: string)

// call
const { isCurrentUserInRole } = useAuth();
...
isCurrentUserInRole(CLERK_ROLES.ADMINISTRATOR)

```
Am I understanding the suggestion correctly?

PR #6354, opened 2024-08-19
fix(api, shared, application-generic): Replace branded types with `string` type

rifont:

* Replace branded string types with `string` type
  * The previous attempt at branded typing actually had no effect due to using only `string` typings as the brand instead of applying a computed property with a type guard ([as documented here](https://egghead.io/blog/using-branded-types-in-typescript)). Therefore the brands were not actually having any effect
  * Branded types are not compatible with Nest.js SWC compiler due to SWC's lack of Typescript support. All types must be declared in ES modules that are reachable by Nest.js through `metadata.ts`, this wasn't the case for the branded types which lived in `@novu/shared`.

### Screenshots

_**BEFORE** - Topic DTOs showing incorrect `object` typings_
image

_**AFTER** - Topic DTOs now have expected `string` typings instead of `object`_
image

<details>

</details>


apps/api/src/app/inbox/dtos/update-all-notifications-request.dto.ts
```diff
@@ -1,8 +1,6 @@
 import { IsArray, IsOptional, IsString } from 'class-validator';
 
-import type { NotificationFilter } from '../utils/types';
-
-export class UpdateAllNotificationsRequestDto implements Pick<NotificationFilter, 'tags'> {
+export class UpdateAllNotificationsRequestDto {
```
rifont, line 3: `Pick`, `Omit` and other helper generic types are also incompatible with Nest.js' SWC metadata.

jainpawan21 (approved):

PR #6359, opened 2024-08-20
feat(api): expose organization usage; improve performance

ChmaraX:
Description in related PR: https://github.com/novuhq/packages-enterprise/pull/206

### DIFF
--- a/apps/api/src/app/testing/billing/quota-throttler.guard.e2e-ee.ts
+++ b/apps/api/src/app/testing/billing/quota-throttler.guard.e2e-ee.ts
@@ -1,8 +1,8 @@
 import { UserSession } from '@novu/testing';
 import { expect } from 'chai';
 import sinon from 'sinon';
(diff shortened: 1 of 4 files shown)


apps/api/src/app/testing/billing/quota-throttler.guard.e2e-ee.ts
```diff
@@ -1,8 +1,8 @@
 import { UserSession } from '@novu/testing';
 import { expect } from 'chai';
 import sinon from 'sinon';
-import { EvaluateEventResourceLimit } from '@novu/ee-billing';
 import { ApiServiceLevelEnum } from '@novu/shared';
+import { GetEventResourceLimit } from 'enterprise/packages/billing/dist';
```
rifont: ❓ **question:** Is this new absolute import path a mistake? We should prefer to use the NPM package name import.‏

rifont (approved):
🏎️

PR #6360, opened 2024-08-20
feat(js): Pixel perfect implementation

desiprisg:
### Screenshots

<details>

</details>


LetItRock (changes requested):
:( 

Screenshot 2024-08-22 at 18 43 04

desiprisg:
@LetItRock  I think this is the result of the issue we were talking about. There's an issue with the `build:watch` script. I keep getting the behaviour too. I think you'll see that some classes don't correspond to styles when inspecting. You'll need to build from scratch locally.

LetItRock:
> @LetItRock I think this is the result of the issue we were talking about. There's an issue with the `build:watch` script. I keep getting the behaviour too. I think you'll see that some classes don't correspond to styles when inspecting. You'll need to build from scratch locally.

ok, it's weird, I tested it the first time by rebuilding everything, but now I don't see the issue anymore.

LetItRock (approved):
awesome! 🙌

PR #6361, opened 2024-08-20
fix(web): Re-render for clerk component breaks error reporting

scopsy:

Remove the function definitions from outside of the component, since each rerender recreated the whole component structure, since the garbage collector deleted the function from memory. 

### Screenshots

CleanShot 2024-08-20 at 18 18 57@2x

<details>

</details>


denis-kralj-novu (approved):
tried to sign in with a non existing email, got the error message, works, 🚢 it!

PR #6362, opened 2024-08-20
fix(js,react): inbox custom bell unread count not updating

LetItRock:
Fixed the issue with the custom bell unread count value not being updated.

Additionally in this PR, I did a few updates to the playground app:
- updated the custom notification item example
- added the case of using the custom popover component
- added the case for the custom Novu theme

### Screenshots

https://github.com/user-attachments/assets/2bc71c70-264a-44fe-8f74-c91a3c629221

https://github.com/user-attachments/assets/083525b0-8ad1-4693-8509-078564eb5f6d


packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -200,9 +200,9 @@ export const DefaultNotification = (props: DefaultNotificationProps) => {
           </div>
         </div>
         <Show when={props.notification.subject}>
-          <p class={style('notificationSubject', 'nt-font-semibold')}>{props.notification.subject}</p>
+          <p class={style('notificationSubject', 'nt-text-start nt-font-semibold')}>{props.notification.subject}</p>
         </Show>
-        <p class={style('notificationBody')}>{props.notification.body}</p>
+        <p class={style('notificationBody', 'nt-text-start')}>{props.notification.body}</p>
```
LetItRock, line 207: the subject and content text of the notification should be aligned to "start" which supports RTL text

packages/js/src/ui/components/Notification/Notification.tsx
```diff
@@ -1,21 +1,30 @@
...
 export const Notification = (props: NotificationProps) => {
+  let ref: HTMLDivElement;
+
+  createEffect(() => {
+    const unmount = props.renderNotification?.(ref, props.notification);
+
+    onCleanup(() => {
+      unmount?.();
+    });
+  });
```
LetItRock: this is the fix for the same case in the custom notification item, but I'll explain the reasoning in the bell component

packages/js/src/ui/components/elements/Bell/Bell.tsx
```diff
@@ -1,19 +1,31 @@
...
+  let ref: HTMLDivElement;
   const { totalUnreadCount } = useTotalUnreadCount();
 
+  createEffect(() => {
+    const unmount = props.renderBell?.(ref, totalUnreadCount());
+
+    onCleanup(() => {
+      unmount?.();
+    });
+  });
```
LetItRock: The reasoning: before we were just mounting the bell component in the `onMount` hook, but we need also to rerender it when the `totalUnreadCount` changes. The `createEffect` is called on mount and also when signals are updated and it is exactly what we need to use instead.

I thought about splitting it into `mountBell` and `updateBell` but didn't decide to do that because:
- it causes multiple re-renders on the react side because `onMount` and `createEffect` called after each other
- on the React side we don't differentiate mounts/updates as there is just one callback `renderBell`, which renders the ReactNode and updates the state

I've removed the `ExternalElementMounter` from here because the "update" phase is specific to this component and I'm not sure if we can generalize it.

LetItRock: Also, I renamed the props so they reflect better now what that function is used for.

playground/nextjs/src/components/Header.tsx
```diff
@@ -13,6 +13,8 @@ const LINKS: LinkType[] = [
   { href: '/render-notification', label: 'Render Notification' },
   { href: '/preferences', label: 'Preferences' },
   { href: '/notifications', label: 'Notifications' },
+  { href: '/novu-theme', label: 'Novu Theme' },
+  { href: '/custom-popover', label: 'Custom Popover' },
```
LetItRock, line 17: two new examples in the playground

playground/nextjs/src/pages/render-notification/index.tsx
```diff
@@ -10,11 +10,26 @@ export default function Home() {
         {...novuConfig}
         renderNotification={(notification) => {
           return (
-            <div className="flex gap-2 flex-nowrap items-start self-stretch my-1 p-2 hover:bg-slate-200">
-              <div className="rounded-full w-8 h-8 overflow-hidden border border-cyan-200">Avatar</div>
+            <div
```
LetItRock, line 13: added this example to be sure that the `renderNotification` callback is called with the new value

packages/js/src/ui/components/Inbox.tsx
```diff
@@ -14,8 +14,8 @@ import { Button, Popover } from './primitives';
 
 export type InboxProps = {
   open?: boolean;
-  mountNotification?: NotificationMounter;
-  mountBell?: BellMounter;
+  renderNotification?: NotificationRenderer;
+  renderBell?: BellRenderer;
```
desiprisg, line 18: ⚠️ In the JS level, this is a mount method, not a render method. I think we should keep the names as they were. This also makes more sense to be explored in a new PR, or separate discussion.

packages/js/src/ui/components/elements/Bell/Bell.tsx
desiprisg: I understand what you're saying. Essentially, we want the mounter to be reactive.
I think what you've done here is the same as just having `createEffect` inside the mounter, right?

desiprisg (changes requested):
Let's please reserve this PR just for the fix, and I think we should be using the mounter. Let's discuss.

LetItRock:
> Let's please reserve this PR just for the fix, and I think we should be using the mounter. Let's discuss.

@desiprisg I'll address all the questions/concerns on the Slack with you ;)

packages/js/src/ui/components/ExternalElementRenderer.tsx
```diff
@@ -0,0 +1,26 @@
+import { onCleanup, createEffect, ParentProps } from 'solid-js';
+
+type ExternalElementMounterProps = ParentProps<{
+  render: (el: HTMLDivElement) => () => void;
+}>;
+
+export const ExternalElementRenderer = ({ render, ...rest }: ExternalElementMounterProps) => {
```
LetItRock, line 7: renamed the file and component accordingly to what we agreed

BiswaViraj (approved):

PR #6370, opened 2024-08-21
fix(js): Rerender custom components provided when data changes

desiprisg:

This sounds like it would cause a remount each time (cause of the name) but it actually doesn't.
Essentially the flow is:
* render function provided in react
* react package creates a `mountX` method based on that function
  * the `mountX` method essentially adds the element to the map of rendered elements in the react <Renderer />
* novu-js executes the `mountX` method on its element.
  * This element wraps the custom component and remains the same even if it is remounted.
* Since that element is the key for the map inside the `<Renderer />` , remounting acts as a recomputation of the rendered component, or a rerender.
  * Running mountX essentially executes the (data) => JSX in the `<Renderer />` in react. It's like doing `<>{renderX(data)}</>`
  
I've added an enter animation on the custom bell so it's visible that the (react) mount only happens once.
### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/ExternalElementMounter.tsx
+++ b/packages/js/src/ui/components/ExternalElementMounter.tsx
@@ -1,4 +1,4 @@
-import { onCleanup, onMount, ParentProps } from 'solid-js';
+import { createEffect, onCleanup, ParentProps } from 'solid-js';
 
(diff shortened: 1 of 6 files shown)


LetItRock:
As we agreed, I'm closing this one and will move the changes into my PR.

PR #6373, opened 2024-08-21
fix(framework): Stop requiring default properties to be specified in outputs

rifont:

* Stop requiring default properties to be specified in step outputs, provider outputs, and trigger payload
  * `json-schema-to-ts` supports `keepDefaultedPropertiesOptional` ([ref](https://github.com/ThomasAribart/json-schema-to-ts?tab=readme-ov-file#objects)) which is now set to `true` for unvalidated data (such as the return value of a `step.delay` step)
  * Add tests to verify the `undefined` properties can be specified for outputs with a `default` value set
* Fix schema validator types

### Screenshots

_BEFORE - properties with a `default` set are required to be specified in outputs_
image

_AFTER - properties with a `default` set are NOT required to be specified in outputs_
image

<details>

</details>

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -633,6 +633,8 @@
     "unpublish",
     "unsub",
     "untracked",
+    "unvalidated",
+    "Unvalidated",
     "upsert",
     "upserted",
     "upstash",
--- a/packages/framework/src/client.test.ts
+++ b/packages/framework/src/client.test.ts
@@ -8,9 +8,8 @@ import {
   WorkflowNotFoundError,
 } from './errors';
 import { workflow } from './resources';
-import { Event, Step, FromSchema } from './types';
-import { delayOutputSchema, channelStepSchemas } from './schemas';
-import { FRAMEWORK_VERSION, SDK_VERSION, ChannelStepEnum, PostActionEnum } from './constants';
+import { Event, Step } from './types';
+import { FRAMEWORK_VERSION, SDK_VERSION, PostActionEnum } from './constants';
 
 describe('Novu Client', () => {
   let client: Client;
@@ -444,11 +443,11 @@ describe('Novu Client', () => {
 
   describe('executeWorkflow method', () => {
     it('should execute workflow successfully when action is execute and payload is provided', async () => {
-      const delayConfiguration: FromSchema<typeof delayOutputSchema> = { type: 'regular', unit: 'seconds', amount: 1 };
-      const emailConfiguration: FromSchema<(typeof channelStepSchemas)[ChannelStepEnum.EMAIL]['output']> = {
+      const delayConfiguration = { unit: 'seconds', amount: 1 } as const;
+      const emailConfiguration = {
         body: 'Test Body',
         subject: 'Subject',
-      };
+      } as const;
       const newWorkflow = workflow('test-workflow', async ({ step }) => {
         await step.email('send-email', async () => emailConfiguration);
         await step.delay('delay', async () => delayConfiguration);
@@ -515,7 +514,7 @@ describe('Novu Client', () => {
       expect(amount).toBe(delayConfiguration.amount);
       expect(delayExecutionResult.providers).toEqual({});
       const type = delayExecutionResult.outputs.type;
-      expect(type).toBe(delayConfiguration.type);
+      expect(type).toBe('regular');
     });
 
     it('should compile default control variable', async () => {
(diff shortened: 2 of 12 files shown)


packages/framework/src/types/step.types.ts
```diff
@@ -173,47 +173,62 @@ export type ChannelStep<
 ) => StepOutput<T_Result>;
 
 export type EmailOutput = FromSchema<(typeof channelStepSchemas)[ChannelStepEnum.EMAIL]['output']>;
+export type EmailOutputUnvalidated = FromSchemaUnvalidated<
+  (typeof channelStepSchemas)[ChannelStepEnum.EMAIL]['output']
+>;
```
rifont, line 187: I am adding these `Unvalidated` types as new types because most of the time, type consumers of `@novu/framework` will need the validated types which don't have the `Unvalidated` suffix. We already have a number of places using the validated types, so I've opted not to update them both for the stated reason above regarding usage, and to avoid a large number of file changes.

packages/framework/src/client.test.ts
```diff
@@ -515,7 +514,7 @@ describe('Novu Client', () => {
       expect(amount).toBe(delayConfiguration.amount);
       expect(delayExecutionResult.providers).toEqual({});
       const type = delayExecutionResult.outputs.type;
-      expect(type).toBe(delayConfiguration.type);
+      expect(type).toBe('regular');
```
rifont, line 517: This file just has some tidy up and also tests the `delay` step without needing to provide the `type` property, which initially highlighted the bug.

.cspell.json
```diff
@@ -633,6 +633,8 @@
     "unpublish",
     "unsub",
     "untracked",
+    "unvalidated",
+    "Unvalidated",
```
antonjoel82, line 637: ❓ **question:** Does capitalization matter?‏

packages/framework/src/types/schema.types.ts
```diff
@@ -3,8 +3,37 @@ import zod from 'zod';
 
 export type JsonSchema = JSONSchema;
 
+/**
+ * A JSONSchema or a ZodSchema.
```
antonjoel82: 💭 **thought (non-blocking):** Not sure that this comment clarifies anything -- will we just add another "or ___" when we add a new schema type? Instead, perhaps we can explain what the purpose is of a `Schema`?‏

packages/framework/src/types/step.types.ts
```diff
@@ -173,47 +173,62 @@ export type ChannelStep<
...
 export type InAppResult = FromSchema<(typeof channelStepSchemas)[ChannelStepEnum.IN_APP]['result']>;
 
 export type DelayOutput = FromSchema<(typeof actionStepSchemas)[ActionStepEnum.DELAY]['output']>;
+export type DelayOutputUnvalidated = FromSchemaUnvalidated<(typeof actionStepSchemas)[ActionStepEnum.DELAY]['output']>;
 export type DelayResult = FromSchema<(typeof actionStepSchemas)[ActionStepEnum.DELAY]['result']>;
 
 export type digestRegularOutput = FromSchema<typeof digestRegularOutputSchema>;
+export type digestRegularOutputUnvalidated = FromSchemaUnvalidated<typeof digestRegularOutputSchema>;
 export type digestTimedOutput = FromSchema<typeof digestTimedOutputSchema>;
+export type digestTimedOutputUnvalidated = FromSchemaUnvalidated<typeof digestTimedOutputSchema>;
```
antonjoel82: ✏ **suggestion:** Perhaps we can deprecate the old ones and fix capitalization of the types here?‏

SokratisVidros (approved):

.cspell.json
rifont: Yes 😓

packages/framework/src/types/schema.types.ts
rifont: Great callout, I added a description and converted into a list view for easier extensibility.

packages/framework/src/types/step.types.ts
rifont: Thanks for highlighting this. I decided to update everything to have consistent capitlization. Deprecating is not needed as these uncapitlized types will almost certainly be used only by us.

In the rare case that any consumers are using the uncapitlized types, they will be greeted with a suggestion to use the new type:
image

PR #6376, opened 2024-08-22
fix(web): pixel perfect UI implementation of clerk components

denis-kralj-novu:
## Video click through
https://www.loom.com/share/9758365347d14664897fd2e20e88f88e

## Org dropdown
image

## user menu dropdown
image

### Profile
image

### Security
image

### Organization
image

### Members
image

### Billing
image


apps/web/src/ee/clerk/pages/ManageAccountPage.css
```diff
@@ -0,0 +1,83 @@
...
+  padding: var(--semantic-paddings-popover-vertical, 24px) var(--semantic-paddings-popover-horisontal, 24px);
+  flex-direction: column;
+  align-items: flex-start;
+  gap: var(--semantic-margins-layout-PageTitleBottom, 24px);
+  flex: 1 0 0;
+  align-self: stretch;
+  border-radius: 12px;
+  background: var(--nv-colors-surface-panel-subsection);
+  overflow-y: auto;
+  /* TODO: these 2 are wrong, need to calculate based on available parent size */
```
denis-kralj-novu: Help needed, I don't know how to calculate this values below to be based on available space in parent

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+    <Modal
+      opened
+      padding={8}
+      title={undefined}
+      onClose={() => {
+        navigate(ROUTES.WORKFLOWS);
+      }}
+      size="90%"
+    >
+      <>
```
SokratisVidros: This fragment can be removed.

apps/web/src/ee/clerk/pages/ManageAccountPage.css
```diff
@@ -0,0 +1,83 @@
+.mantine-Modal-modal {
+  border-radius: 12px;
```
SokratisVidros: @antonjoel82 do we have tokens or any other type of standards for border radius?

apps/web/src/ee/clerk/pages/ManageAccountPage.css
```diff
@@ -0,0 +1,83 @@
...
+  background-color: var(--nv-colors-surface-panel);
+  padding: 0px;
```
SokratisVidros:
```suggestion
  padding: 0;
```

apps/web/src/ee/clerk/pages/ManageAccountPage.css
```diff
@@ -0,0 +1,83 @@
...
+.mantine-Tabs-tabIcon>svg {
+  color: inherit;
+}
+
+.mantine-Tabs-tabLabel {
+  font-weight: inherit;
+  font-family: inherit;
+  font-size: inherit;
+  color: inherit;
+
```
SokratisVidros: Please remove the empty line.

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
```diff
@@ -0,0 +1,149 @@
+import './ManageAccountPage.css';
+import { OrganizationProfile, UserProfile } from '@clerk/clerk-react';
+import { useColorScheme } from '@novu/design-system';
+import { Modal, Tabs } from '@mantine/core';
```
SokratisVidros, line 2: This should be Mantine v5, right? Can we use Mantine 7?

apps/web/src/ee/clerk/pages/ManageAccountPage.css
SokratisVidros: If the parent is display: flex, does a height: 100% solve the issue? We can pair on this later today.

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
```diff
@@ -0,0 +1,149 @@
...
+      }}
+      size="90%"
+    >
+      <>
+        <Tabs value={tabValue} onTabChange={(value) => navigate(`/manage-account/${value}`)} orientation="vertical">
+          <Tabs.List>
+            <Tabs.Tab
+              disabled
+              value="title"
+              className={css({
```
SokratisVidros: I am confused with the addition of inline CSS. Mantine promotes css via React className but we also introduced a `ManageAccountPage.css` with BEM style classes. What was the reasoning behind this split?

SokratisVidros (changes requested):

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
denis-kralj-novu: yes, we use v5 within the dashboard project. We use v7 in novui, but I quickly came to realize that there would need to be some adjustments in this package to make it work the way we want and it took too much time off my plate. We can create a follow up task for this in the backlog.

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
denis-kralj-novu: This is a workaround to make one of the tabs look like the title. The tabs themselves have styles (the look of all other buttons) but I needed to override this particular one to be specific. I think I could give it a specific class name and then target that from the css file, wdyt?

SokratisVidros: That's an option. The addition of the .css file is a bit unexpected. If not mistaken in Mantine css in classname should be preferred.

apps/web/src/ee/clerk/pages/ManageAccountPage.css
denis-kralj-novu: Didn't work. I will ask Joel to help me with this

apps/web/src/ee/clerk/pages/ManageAccountPage.css
denis-kralj-novu: replaced with tokens

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
denis-kralj-novu: I've tried to move the definition to the .css file but it didn't work as expected. It will be another thing I will try to target with Joel

SokratisVidros (approved):

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
denis-kralj-novu, line 1: While not in line with the novui approach, this styling is done like this to quickly get clerk components to be aligned with our styles. 
I would fix this only in the future when we decide to invest in our own UI wrapper for the functionality exposed by these components. In that case, this would be thrown out anyway.

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
denis-kralj-novu, line 1: While not in line with the novui approach, this styling is done like this to quickly get clerk components to be aligned with our styles. 
I would fix this only in the future when we decide to invest in our own UI wrapper for the functionality exposed by these components. In that case, this would be thrown out anyway.

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
```diff
@@ -16,13 +16,166 @@ const CLERK_LOCALIZATION = {
       security: 'Access security',
     },
   },
-
+  organizationProfile: {
+    membersPage: {
+      requestsTab: { autoSuggestions: { headerTitle: '' } },
+      invitationsTab: { autoInvitations: { headerTitle: '' } },
+    },
```
denis-kralj-novu, line 23: This is done to hide links that break the flow of modals. The functionality is exposed in a dedicated space in a specific settings tab so it is just about hiding a small element on the UI that can cause the modal to break

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+    backgroundColor: 'surface.panel !important',
+    padding: '0 !important',
+    width: '[90%] !important',
+    height: 'full !important',
+  }),
+  body: css({ height: 'full !important' }),
+  header: css({ marginBottom: '0 !important' }),
+  close: css({
+    position: 'fixed !important',
+    right: '[12px] !important',
```
antonjoel82:
```suggestion
    right: '75 !important',
```

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+    top: '[10px] !important',
+    zIndex: '[2] !important',
+  }),
+};
+
+export const titleTab = css({
+  fontWeight: 'strong',
+  color: 'typography.text.main !important',
+  fontSize: 'x-large !important',
```
antonjoel82: ☑ **chore:** Use one of our numerical sizes or instead use a `Title` component to wrap the text‏

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+  letterSpacing: '0',
+  lineHeight: '175',
+});
+
+export const normalTabStyle = css({
+  fontWeight: 'strong',
+  fontSize: '88 !important',
+  color: 'typography.text.secondary !important',
+  padding: '[8px 12px] !important',
+  margin: '[4px]',
```
antonjoel82:
```suggestion
  py: '50 !important',
  px: '75 !important',
  margin: '25',
```
Our spacing / dimension values work based on 16px / 1 rem as 100%, so the token for that is '100'. If you need 4px which is 0.25rem (or 25% of 1rem), then the value is `'25'`

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+  border: 'unset !important',
+  borderRadius: '[4px] !important',
```
antonjoel82:
```suggestion
  borderRadius: '50 !important',
```
Radius is a little different and is based on 0.5rem as 100%. Check `radius.tokents.ts` for ref

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+    padding: '150',
+    borderRadius: 'l',
+    backgroundColor: 'surface.panelSubsection',
+    overflowY: 'auto',
+  }),
+  tabsList: css({
+    borderRight: 'unset !important',
+  }),
+  tabIcon: css({
+    marginRight: '[8px] !important',
```
antonjoel82:
```suggestion
    marginRight: '50 !important',
```

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const normalTabStyle = css({
+  fontWeight: 'strong',
+  fontSize: '88 !important',
+  color: 'typography.text.secondary !important',
+  padding: '[8px 12px] !important',
+  margin: '[4px]',
+  border: 'unset !important',
+  borderRadius: '[4px] !important',
+  minWidth: '[204px]',
```
antonjoel82: ❓ **question:** Is this specific size required? If so, please consider making it a token in the `web/panda.config.ts`‏

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
```diff
@@ -0,0 +1,125 @@
...
+export default function ManageAccountPage() {
+  const navigate = useNavigate();
+  const { tabValue } = useParams();
+  const isV2Enabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V2_ENABLED);
+
+  return (
+    <Modal
+      classNames={modalStyles}
+      opened
+      padding={8}
```
antonjoel82:
```suggestion
```
Please avoid using Mantine's built-in sizing so that we keep control of our scale

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
```diff
@@ -0,0 +1,125 @@
...
+      title={undefined}
+      onClose={() => {
+        navigate(ROUTES.WORKFLOWS);
+      }}
+    >
+      <Tabs
+        classNames={tabsStyles}
+        value={tabValue}
+        onTabChange={(value) => navigate(`/manage-account/${value}`)}
```
antonjoel82: ☑ **chore:** Please derive `/manage-account/` from a centralized Route value ‏

apps/web/src/ee/clerk/pages/ManageAccountPage.tsx
```diff
@@ -0,0 +1,125 @@
...
+        orientation="vertical"
+      >
+        <Tabs.List>
+          <Tabs.Tab disabled value="title" className={titleTab}>
+            Settings
+          </Tabs.Tab>
+          <Tabs.Tab
+            value="user-profile"
```
antonjoel82: ☑ **chore:** Please use enums / another centralized value for the tab values. This will allow us to better type the `tabValue` from `useParams` as well‏

antonjoel82 (commented):
Looks good overall, thank you for the changes!

My changes are not fully-blocking, but I do feel that we need to avoid using magic values for routing as much as possible, so I would appreciate these changes in a follow-up PR

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
denis-kralj-novu: Thank you, I was trying to conclude what the ratio is, this is very helpful :)

apps/web/src/constants/routes.ts
```diff
@@ -72,6 +72,23 @@ export enum ROUTES {
...
 
   LOCAL_STUDIO_AUTH = '/local-studio/auth',
+
+  /** Account settings routes, bookmarkable by tabs */
+  MANAGE_ACCOUNT_SEGMENT = '/manage-account/',
+  MANAGE_ACCOUNT_USER_PROFILE = '/manage-account/user-profile',
+  MANAGE_ACCOUNT_ACCESS_SECURITY = '/manage-account/access-security',
+  MANAGE_ACCOUNT_ORGANIZATION = '/manage-account/organization',
+  MANAGE_ACCOUNT_TEAM_MEMBERS = '/manage-account/team-members',
+  MANAGE_ACCOUNT_BILLING = '/manage-account/billing',
```
antonjoel82, line 82: ❓ **question:** Is there supposed to be  a `branding` route here? I see we have it as a route segment below?‏

apps/web/src/ee/clerk/pages/ManageAccountPage.styles.ts
```diff
@@ -0,0 +1,94 @@
...
+    width: '[90%] !important',
+    height: 'full !important',
+  }),
+  body: css({ height: 'full !important' }),
+  header: css({ marginBottom: '0 !important' }),
+  close: css({
+    position: 'fixed !important',
+    right: '75 !important',
+    top: '[10px] !important',
+    zIndex: '[2] !important',
```
antonjoel82: ☑ **chore:** Sorry, missed this one. Please use one of the values from the system. Perhaps `docked`?‏

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
```diff
@@ -16,13 +16,169 @@ const CLERK_LOCALIZATION = {
...
+  },
+  userButtonPopoverActionButton__signOutAll: {
+    gap: '12px',
+    '&:hover': {
+      color: 'var(--nv-colors-typography-text-main)',
+    },
+  },
+  userButtonPopoverActionButton__addAccount: {
+    gap: '12px',
```
antonjoel82:
```suggestion
    gap: 'var(--nv-spacing-75)',
```

apps/web/src/ee/clerk/providers/ClerkProvider.tsx
```diff
@@ -16,13 +16,169 @@ const CLERK_LOCALIZATION = {
...
+    fontWeight: 'var(--nv-font-weights-strong)',
+    gap: '12px',
+    padding: 'var(--nv-spacing-100) var(--nv-spacing-125)',
+    '&:hover': {
+      color: 'var(--nv-colors-typography-text-main) !important',
+    },
+  },
+  organizationSwitcherTrigger: {
+    '.cl-organizationPreview': {
+      gap: '12px',
```
antonjoel82:
```suggestion
      gap:  'var(--nv-spacing-75)',
```

antonjoel82 (approved):
🚀

PR #6377, opened 2024-08-22
feat(api): add api description and check for v2

davidsoderberg:


apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -190,4 +207,21 @@ export class TenantController {
       })
     );
   }
+
+  private async checkV2(user: UserSessionData) {
```
rifont: Suggestion for a more semantic naming of the method, to better encapsulate the v2 logic.

```suggestion
  private async verifyTenantsApiAvailability(user: UserSessionData) {
```

apps/api/src/app/tenant/tenant.controller.ts
```diff
@@ -51,6 +54,8 @@ import { UserAuthentication } from '../shared/framework/swagger/api.key.security
 
 import { SdkUsePagination } from '../shared/framework/swagger/sdk.decorators';
 
+const v2Description = ' Tenants is not supported in code first version of the API.';
```
rifont: As below, naming this semantically.
```suggestion
const v2TenantsApiDescription = ' Tenants is not supported in code first version of the API.';
```

rifont (approved):
💯

PR #6380, opened 2024-08-22
feat(js): Revise localization keys DX

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/elements/InboxStatus/InboxStatusDropdown.tsx
+++ b/packages/js/src/ui/components/elements/InboxStatus/InboxStatusDropdown.tsx
@@ -2,7 +2,7 @@ import { useInboxContext, useLocalization } from '../../../context';
 import { cn, useStyle } from '../../../helpers';
 import { ArrowDropDown } from '../../../icons';
 import { Button, buttonVariants, Dropdown } from '../../primitives';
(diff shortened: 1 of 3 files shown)


packages/js/src/ui/types.ts
```diff
@@ -55,7 +55,7 @@ export type NovuProviderProps = BaseNovuProviderProps & {
 };
 
 export enum NotificationStatus {
-  UNREAD_READ = 'unreadRead',
+  UNREAD_READ = 'all',
```
LetItRock: ☑ **chore:** that's not meaning all, because all is: "unread+read+archived+unarchived"

desiprisg: Do you think it matters that much for the key? I also did not like `unreadRead` as the name here. Should we name it `Inbox` or `home` or sth?

LetItRock: yes, I think it might confuse the users if we call it `all` as the localization key

desiprisg: Changed it!

LetItRock: you changed the name for the enum but not the localization key

LetItRock (approved):

PR #6384, opened 2024-08-23
fix(framework, node): Make the `payload` property optional during trigger

rifont:

* Make the `payload` property optional during trigger
  * `payload` is always optional for `@novu/node`
  * `payload` is optional for `@novu/framework` only when there are no required properties present on the `payloadSchema` for the workflow
* Add tests for all utility types (NB: Vitest requires the `.test-d.ts` extension to mark a test suite as being typechecked only, because it doesn't contain `expect` statements Vitest will ignore it by default)
* Add typechecking to Vitest package scripts for `@novu/framework`

### Screenshots

_`payloadSchema` containing required properties marks `payload` as required_
image

_`payloadSchema` containing only optional properties marks `payload` as optional_
image

_`payloadSchema` not specified marks `payload` as optional and `any`_
image

<details>

</details>


packages/framework/src/client.validation.test.ts
```diff
@@ -344,7 +345,7 @@ describe('validation', () => {
 
       try {
         await client.executeWorkflow({
-          action: 'execute',
+          action: PostActionEnum.EXECUTE,
```
rifont, line 348: After adding type checking to Vitest, this file needed updating.

packages/framework/src/client.test.ts
```diff
@@ -272,7 +272,7 @@ describe('Novu Client', () => {
           }),
           {
             providers: {
-              slack: async ({ controls }) => {
+              slack: async () => {
```
rifont, line 275: After adding type checking to Vitest, this test file complained about `controls` being implicit `any`. We weren't using it, so it's removed.

packages/framework/package.json
```diff
@@ -18,8 +18,8 @@
     "README.md"
   ],
   "scripts": {
-    "test": "vitest",
-    "test:watch": "vitest --watch",
+    "test": "vitest --typecheck",
+    "test:watch": "vitest --typecheck --watch",
```
rifont, line 22: This enables type checking as part of our testing process.

packages/framework/.gitignore
```diff
@@ -10,3 +10,6 @@ logs
 
 # Misc
 .DS_Store
+
+# Vitest
+tsconfig.vitest-temp.json
```
rifont, line 15: Vitest generates a temporary tsconfig after enabling typechecking, we can ignore it.

packages/framework/src/handler.ts
```diff
@@ -218,7 +218,6 @@ export class NovuRequestHandler<Input extends any[] = any[], Output = any> {
         transactionId: triggerEvent.transactionId,
         overrides: triggerEvent.overrides || {},
         ...(triggerEvent.actor && { actor: triggerEvent.actor }),
-        ...(triggerEvent.tenant && { tenant: triggerEvent.tenant }),
```
rifont, line 221: Removing `tenant` as they are not supported in V2.

packages/framework/src/resources/workflow.test.ts
```diff
@@ -171,6 +171,75 @@ describe('workflow function', () => {
...
+      );
+
+      // Capture in a test function to avoid throwing execution errors
+      const testFn = () =>
+        testWorkflow.trigger({
+          to: 'test@test.com',
+        });
+    });
+
+    it('should compile when the payload is not specified and the payloadSchema is not specified', async () => {
```
rifont, line 229: ❓ **question:** It's debatable whether payload should be omitted entirely if payloadSchema is not present. It's a move towards greater strictness + type safety in Framework, but I wonder if we should be this opinionated and force specification of payloadSchema for those that would just like to trigger a workflow quickly.‏

packages/framework/src/resources/workflow.ts
```diff
@@ -46,9 +46,10 @@ export function workflow<
       throw new MissingSecretKeyError();
     }
 
+    const unvalidatedData = (event.payload || {}) as T_PayloadUnvalidated;
```
rifont, line 49: `payload` falls back to an empty object when it's not specified in args during `workflow(...).trigger(<args>)`

packages/framework/src/types/event.types.ts
```diff
@@ -45,14 +43,6 @@ export type EventTriggerParams<T_Payload = EventPayload> = {
...
    */
   bridgeUrl?: string;
-  /**
-   * Payload to trigger the workflow with
-   */
-  payload: T_Payload;
-  /**
-   * Tenant to trigger the workflow with
-   */
-  tenant?: Tenant;
```
rifont, line 55: As above, removing `tenant` as it's not supported in V2.

packages/framework/src/types/event.types.ts
```diff
@@ -69,7 +59,15 @@ export type EventTriggerParams<T_Payload = EventPayload> = {
...
-};
+} & ConditionalPartial<
+  {
+    /**
+     * Payload to trigger the workflow with
+     */
+    payload: T_Payload;
+  },
+  PickRequiredKeys<T_Payload> extends never ? true : false
+>;
```
rifont, line 70: Using a new utility type which marks the first generic argument as partial when the second generic argument is `true`. Combined with `PickRequiredKeys` which returns `never` when there are no required keys, we can make `payload` optional when there are no required keys present in the `payload`.

packages/framework/src/types/util.types.test-d.ts
rifont, line 1: This file adds tests that were missing for existing utility types, plus tests for all the new utility types.

Note that the file extension is `.test-d.ts` to instruct Vitest to typecheck the file even when there are no `expect` statements present.

packages/node/src/lib/events/events.interface.ts
```diff
@@ -15,7 +15,7 @@ import {
 } from '@novu/shared';
 
 export interface IBroadcastPayloadOptions {
-  payload: ITriggerPayload;
+  payload?: ITriggerPayload;
```
rifont, line 18: In `@novu/node`, the `payload` is marked as optional in all cases, because the payload schema is not known.

The API already accepts `payload: undefined`, so there's no need for any further changes in `@novu/node`.

packages/framework/tsconfig.json
```diff
@@ -16,5 +16,5 @@
     "strict": true
   },
   "include": ["./src/**/*", "./package.json", "scripts/devtool.ts"],
-  "exclude": ["node_modules", "**/*.spec.ts"]
+  "exclude": ["node_modules"]
```
rifont, line 19: This had no effect, removing as part of housekeeping.

davidsoderberg (approved):

PR #6387, opened 2024-08-24
chore(application-generic): removed docker version & added healthchecks

twentyone24:
1. **Removed `version` as it's deprecated**:
   - The `version` field is no longer needed in Docker Compose files as it's deprecated, according to the [official Compose spec](https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-and-name-top-level-elements). By removing it, preventing obsolete warnings and ensuring future compatibility with the latest Docker Compose standards.
   
2. **Added healthchecks for containers and updated dependencies**:
   - Adding health checks improves the reliability of services by allowing Docker to ensure that services are healthy and running as expected. This change ensures that dependent services (like `api`, `worker`, and `ws`) only start after the health checks of their dependent services (like `mongodb` and `redis`) pass successfully. This avoids race conditions where services might try to connect to others before they're fully ready.

```bash

 ~  docker ps
CONTAINER ID   IMAGE                               COMMAND                  CREATED         STATUS                        PORTS                      NAMES
daa4ab35373a   ghcr.io/novuhq/novu/embed:0.24.0    "docker-entrypoint.s…"   2 minutes ago   Up 50 seconds (healthy)       0.0.0.0:4701->4701/tcp     embed
5aa0298cbc9f   ghcr.io/novuhq/novu/widget:0.24.0   "docker-entrypoint.s…"   2 minutes ago   Up About a minute (healthy)   0.0.0.0:4500->4500/tcp     widget
b506e29ce4cb   ghcr.io/novuhq/novu/web:0.24.0      "docker-entrypoint.s…"   2 minutes ago   Up About a minute (healthy)   0.0.0.0:4200->4200/tcp     web
9de5ebda5cf5   ghcr.io/novuhq/novu/worker:0.24.0   "docker-entrypoint.s…"   2 minutes ago   Up About a minute (healthy)                              worker
2bfc922f6300   ghcr.io/novuhq/novu/ws:0.24.0       "docker-entrypoint.s…"   2 minutes ago   Up About a minute (healthy)   0.0.0.0:3002->3002/tcp     ws
8f7f7d035da7   ghcr.io/novuhq/novu/api:0.24.0      "docker-entrypoint.s…"   2 minutes ago   Up About a minute (healthy)   0.0.0.0:3000->3000/tcp     api
9b88fe95bae7   mongo                               "docker-entrypoint.s…"   2 minutes ago   Up 2 minutes (healthy)        0.0.0.0:27017->27017/tcp   mongodb
a307275d4b8b   redis:alpine                        "docker-entrypoint.s…"   2 minutes ago   Up 2 minutes (healthy)        6379/tcp                   redis
```
<details><summary>Screenshot</summary>
<p>
image
</p>
</details> 

---
_Thanks for taking your time reviewing this :D_


docker/community/docker-compose.yml
```diff
@@ -66,11 +81,20 @@ services:
       IS_MULTI_TENANCY_ENABLED: ${IS_MULTI_TENANCY_ENABLED}
     ports:
       - '3000:3000'
+    healthcheck:
+      test: ["CMD-SHELL", "wget --spider --quiet http://localhost:3000/v1/health-check || exit 1"]
```
jainpawan21, line 85: @twentyone24 I see you have hardcoded value l`ocalhost:3000`. api service can run on different port, IP or custom domain in case of self hosting. Similar to all services health check

twentyone24: @jainpawan21 appreciate your comments, 
- regarding `HOST`, the `localhost`, `api` hosts should  still work even  with custom configuration or with reverse proxy, as these HEALTHCHECKs execute within the container under the same network. (also, `localhost` refers to the same container & the HOST `api` refers to the api container under the same docker network. 
- regarding `PORTS`, I've hardcoded them as the existing config has hardcoded ports setup `'3000:3000'` unless this is not changed that won't be affected. and better way to do this would be to decouple into an .env and pass them as arguments wherever required. _(if a lot of users always run on a different port)_

hope this resolves your concerns.

jainpawan21 (approved):

PR #6391, opened 2024-08-26
feat(framework,js): expose the data property on the in-app step and notification object

LetItRock:
Expose the `data` property on the framework `in-app` step output. The `data` value is stored on the `message` document and will be read by the Inbox API.

**This PR doesn't include any changes to the Studio UI.**

### Screenshots

Screenshot 2024-08-26 at 13 56 52

Screenshot 2024-08-26 at 13 57 04

### DIFF
--- a/apps/api/src/app/inbox/utils/notification-mapper.ts
+++ b/apps/api/src/app/inbox/utils/notification-mapper.ts
@@ -17,6 +17,7 @@ const mapSingleItem = ({
   avatar,
   cta,
   tags,
+  data,
 }: MessageEntity): InboxNotification => {
   const to: Subscriber = {
     id: subscriber?._id ?? '',
@@ -56,6 +57,7 @@ const mapSingleItem = ({
           url: cta.data.url,
         }
       : undefined,
+    data,
   };
 };
 
--- a/apps/api/src/app/inbox/utils/types.ts
+++ b/apps/api/src/app/inbox/utils/types.ts
@@ -31,6 +31,7 @@ export type InboxNotification = {
   redirect?: {
     url: string;
   };
+  data?: Record<string, unknown>;
 };
 
 export type NotificationFilter = {
--- a/apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
+++ b/apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
@@ -201,12 +201,15 @@ export class SendMessageInApp extends SendMessageBase {
     const bridgeOutputs = command.bridgeData?.outputs as InAppOutput;
     const inAppMessage = inAppMessageFromBridgeOutputs(bridgeOutputs);
 
-    const channelData: Partial<Pick<MessageEntity, 'content' | 'subject' | 'avatar' | 'payload' | 'cta' | 'tags'>> = {
+    const channelData: Partial<
+      Pick<MessageEntity, 'content' | 'subject' | 'avatar' | 'payload' | 'cta' | 'tags' | 'data'>
+    > = {
       content: (this.storeContent() ? inAppMessage.content || content : null) as string,
       cta: bridgeOutputs ? inAppMessage.cta : step.template.cta,
       subject: inAppMessage.subject,
       avatar: inAppMessage.avatar,
       payload: messagePayload,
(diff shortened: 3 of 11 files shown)


apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
```diff
@@ -201,12 +201,15 @@ export class SendMessageInApp extends SendMessageBase {
...
-    const channelData: Partial<Pick<MessageEntity, 'content' | 'subject' | 'avatar' | 'payload' | 'cta' | 'tags'>> = {
+    const channelData: Partial<
+      Pick<MessageEntity, 'content' | 'subject' | 'avatar' | 'payload' | 'cta' | 'tags' | 'data'>
+    > = {
       content: (this.storeContent() ? inAppMessage.content || content : null) as string,
       cta: bridgeOutputs ? inAppMessage.cta : step.template.cta,
       subject: inAppMessage.subject,
       avatar: inAppMessage.avatar,
       payload: messagePayload,
+      data: inAppMessage.data,
```
LetItRock, line 212: This is the part of the code that I'm mostly not sure about. The `channelData` is the object that will be saved on the `message` document, however, it already contains the `payload` field, and I've added `data` field.

The thing that I'm not sure about is whether we want to feature flag this part like when v2 only saves the `data` but not `payload`, or should we keep both?

apps/api/src/app/inbox/utils/notification-mapper.ts
```diff
@@ -56,6 +57,7 @@ const mapSingleItem = ({
           url: cta.data.url,
         }
       : undefined,
+    data,
```
LetItRock, line 60: This piece of the mapper code that converts the message doc to the inbox notification. It's part of the Inbox API.

apps/api/src/app/inbox/utils/types.ts
```diff
@@ -31,6 +31,7 @@ export type InboxNotification = {
   redirect?: {
     url: string;
   };
+  data?: Record<string, unknown>;
```
LetItRock, line 34: The data type that will be generated by `framework` too.

packages/framework/src/schemas/steps/channels/in-app.schema.ts
```diff
@@ -18,6 +18,7 @@ const inAppOutputSchema = {
     avatar: { type: 'string', format: 'uri' },
     primaryAction: actionSchema,
     secondaryAction: actionSchema,
+    data: { type: 'object', additionalProperties: true },
```
LetItRock, line 21: Looking also for a place where I can write the tests for this. I would appreciate it if someone could guide me.

EdenTurgeman:
Looks great @LetItRock just what i was looking for 👍🏿
I would also consider adding schema validation for the data object aswell since it'll probably come very handy for user-generated data.

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
LetItRock: Another question is whether we should allow for the `data` field for the other channels?

LetItRock:
> Looks great @LetItRock just what i was looking for 👍🏿 I would also consider adding schema validation for the data object aswell since it'll probably come very handy for user-generated data.

@EdenTurgeman thanks for the suggestion, I would need to check if that is currently possible 🙌

apps/worker/src/app/workflow/usecases/send-message/send-message-in-app.usecase.ts
rifont: 1. I think we can safely keep both `data` and `payload`, no need for any feature flagging on this behaviour
2. `data` field is not useful for other channels, we can keep it to only the In-App channel

packages/framework/src/schemas/steps/channels/in-app.schema.ts
rifont: Schema changes in Framework are implicitly handled by tests covering the Framework's schema capability. Since we are just adding a new field, we don't need to add explicit tests for this change.

rifont (approved):

PR #6394, opened 2024-08-27
feat(framework): cta support with target

BiswaViraj:

- added target support for redirect URLs
- added `redirect` URL for notification click
### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/inbox/utils/notification-mapper.ts
+++ b/apps/api/src/app/inbox/utils/notification-mapper.ts
@@ -45,16 +45,25 @@ const mapSingleItem = ({
     primaryAction: primaryCta && {
       label: primaryCta.content,
       isCompleted: actionType === ButtonTypeEnum.PRIMARY && actionStatus === MessageActionStatusEnum.DONE,
+      redirect: {
+        url: primaryCta.url,
+        target: primaryCta.target,
+      },
     },
     secondaryAction: secondaryCta && {
       label: secondaryCta.content,
       isCompleted: actionType === ButtonTypeEnum.SECONDARY && actionStatus === MessageActionStatusEnum.DONE,
+      redirect: {
+        url: secondaryCta.url,
+        target: secondaryCta.target,
+      },
     },
     channelType: channel,
     tags,
     redirect: cta.data?.url
       ? {
           url: cta.data.url,
+          target: cta.data.target,
         }
       : undefined,
     data,
--- a/apps/api/src/app/inbox/utils/notification-mapper.ts
+++ b/apps/api/src/app/inbox/utils/notification-mapper.ts
@@ -45,18 +45,22 @@ const mapSingleItem = ({
     primaryAction: primaryCta && {
       label: primaryCta.content,
       isCompleted: actionType === ButtonTypeEnum.PRIMARY && actionStatus === MessageActionStatusEnum.DONE,
-      redirect: {
-        url: primaryCta.url,
-        target: primaryCta.target,
-      },
+      redirect: primaryCta.url
+        ? {
+            url: primaryCta.url,
+            target: primaryCta.target,
+          }
+        : undefined,
     },
     secondaryAction: secondaryCta && {
       label: secondaryCta.content,
       isCompleted: actionType === ButtonTypeEnum.SECONDARY && actionStatus === MessageActionStatusEnum.DONE,
-      redirect: {
-        url: secondaryCta.url,
-        target: secondaryCta.target,
-      },
+      redirect: secondaryCta.url
+        ? {
+            url: secondaryCta.url,
+            target: secondaryCta.target,
+          }
+        : undefined,
     },
     channelType: channel,
     tags,
--- a/apps/api/src/app/inbox/utils/types.ts
+++ b/apps/api/src/app/inbox/utils/types.ts
@@ -1,4 +1,4 @@
-import type { ChannelTypeEnum, UrlTarget } from '@novu/shared';
+import type { ChannelTypeEnum, Redirect } from '@novu/shared';
 
 export type Subscriber = {
   id: string;
@@ -6,11 +6,6 @@ export type Subscriber = {
   lastName?: string;
   avatar?: string;
   subscriberId: string;
-};
-
-type Redirect = {
-  url?: string;
-  target?: UrlTarget;
 };
 
 type Action = {
--- a/apps/api/src/app/inbox/utils/types.ts
+++ b/apps/api/src/app/inbox/utils/types.ts
@@ -8,9 +8,15 @@ export type Subscriber = {
   subscriberId: string;
 };
 
+type Redirect = {
+  url?: string;
+  target?: '_self' | '_blank' | '_parent' | '_top' | '_unfencedTop';
+};
+
 type Action = {
   label: string;
   isCompleted: boolean;
+  redirect?: Redirect;
 };
 
 export type InboxNotification = {
@@ -28,10 +34,8 @@ export type InboxNotification = {
   secondaryAction?: Action;
   channelType: ChannelTypeEnum;
   tags?: string[];
-  redirect?: {
-    url: string;
-  };
   data?: Record<string, unknown>;
+  redirect?: Redirect;
 };
 
 export type NotificationFilter = {
--- a/apps/api/src/app/inbox/utils/types.ts
+++ b/apps/api/src/app/inbox/utils/types.ts
@@ -1,4 +1,4 @@
-import type { ChannelTypeEnum } from '@novu/shared';
+import type { ChannelTypeEnum, UrlTarget } from '@novu/shared';
 
 export type Subscriber = {
   id: string;
@@ -10,7 +10,7 @@ export type Subscriber = {
 
 type Redirect = {
   url?: string;
-  target?: '_self' | '_blank' | '_parent' | '_top' | '_unfencedTop';
+  target?: UrlTarget;
 };
 
 type Action = {
--- a/libs/shared/src/entities/messages/messages.interface.ts
+++ b/libs/shared/src/entities/messages/messages.interface.ts
@@ -29,10 +29,12 @@ export interface IMessage {
   subject?: string;
 }
 
+type UrlTarget = '_self' | '_blank' | '_parent' | '_top' | '_unfencedTop';
 export interface IMessageCTA {
   type: ChannelCTATypeEnum;
   data: {
     url?: string;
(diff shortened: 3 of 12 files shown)


packages/framework/src/schemas/steps/channels/in-app.schema.ts
```diff
@@ -1,10 +1,20 @@
 import { Schema } from '../../../types/schema.types';
 
+const redirectSchema = {
+  type: 'object',
+  properties: {
+    url: { type: 'string' },
+    target: { enum: ['_self', '_blank', '_parent', '_top'] },
```
LetItRock:
```suggestion
    target: { enum: ['_self', '_blank', '_parent', '_top', '_unfencedTop'] },
```

packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -45,7 +45,8 @@ export const DefaultNotification = (props: DefaultNotificationProps) => {
 
     props.onNotificationClick?.(props.notification);
     if (props.notification.redirect?.url) {
-      window.open(props.notification.redirect?.url, '_blank', 'noreferrer noopener');
+      const target = props.notification.redirect?.target || '_blank';
+      window.open(props.notification.redirect?.url, target, 'noreferrer noopener');
```
LetItRock: ✏ **suggestion:** please move the default target and default window features to a constants and please reuse them

apps/api/src/app/inbox/utils/notification-mapper.ts
```diff
@@ -44,16 +44,25 @@ const mapSingleItem = ({
     primaryAction: primaryCta && {
       label: primaryCta.content,
       isCompleted: actionType === ButtonTypeEnum.PRIMARY && actionStatus === MessageActionStatusEnum.DONE,
+      redirect: {
+        url: primaryCta.url,
+        target: primaryCta.target,
+      },
```
LetItRock: ☑ **chore:** please copy the same changes to this file: `packages/js/src/ws/socket.ts:27`, this is unfortunate that we have it duplicated but yeah :(

packages/js/src/ws/socket.ts
```diff
@@ -65,16 +65,19 @@ const mapToNotification = ({
     primaryAction: primaryCta && {
       label: primaryCta.content,
       isCompleted: actionType === ActionTypeEnum.PRIMARY && actionStatus === NotificationActionStatus.DONE,
+      target: primaryCta.target,
```
LetItRock:
```suggestion
      redirect: {
        url: primaryCta.url,
        target: primaryCta.target,
      },
```

packages/js/src/ws/socket.ts
```diff
@@ -65,16 +65,19 @@ const mapToNotification = ({
...
     },
     secondaryAction: secondaryCta && {
       label: secondaryCta.content,
       isCompleted: actionType === ActionTypeEnum.SECONDARY && actionStatus === NotificationActionStatus.DONE,
+      target: secondaryCta.target,
```
LetItRock:
```suggestion
      redirect: {
        url: secondaryCta.url,
        target: secondaryCta.target,
      },
```

packages/js/src/ws/socket.ts
BiswaViraj: 🤦🏼  my bad, good catch

libs/shared/src/entities/messages/messages.interface.ts
```diff
@@ -28,10 +28,12 @@ export interface IMessage {
   subject?: string;
 }
 
+type UrlTarget = '_self' | '_blank' | '_parent' | '_top' | '_unfencedTop';
```
rifont: Could we export this `UrlTarget` type and reuse it in other places within `@novu/shared` and `@novu/api`? We can reduce a lot of duplication

libs/shared/src/utils/bridge.utils.ts
```diff
@@ -19,14 +19,22 @@ type InAppOutput = {
...
+  redirect?: Redirect;
 };
 
+type Redirect = {
+  url: string;
+  target?: Target;
+};
+
+type Target = '_self' | '_blank' | '_parent' | '_top' | '_unfencedTop';
+
```
rifont: We can use a shared typing declared in `messages.interface.ts` for these, to avoid repetition and improve maintainability (see above comment).

packages/framework/src/schemas/steps/channels/in-app.schema.ts
```diff
@@ -1,10 +1,20 @@
 import { Schema } from '../../../types/schema.types';
 
+const redirectSchema = {
+  type: 'object',
+  properties: {
+    url: { type: 'string' },
+    target: { enum: ['_self', '_blank', '_parent', '_top', '_unfencedTop'] },
```
rifont: It's a good idea to explicitly declare enum properties with `type` field to ensure the enum values don't regress.

```suggestion
    target: { type: 'string', enum: ['_self', '_blank', '_parent', '_top', '_unfencedTop'] },
```

apps/api/src/app/inbox/utils/types.ts
```diff
@@ -8,9 +8,15 @@ export type Subscriber = {
   subscriberId: string;
 };
 
+type Redirect = {
+  url?: string;
+  target?: '_self' | '_blank' | '_parent' | '_top' | '_unfencedTop';
```
ainouzgali: 🤓 **nitpick (non-blocking):**  Maybe you could use the `type UrlTarget` you define in shared? Just so if we have a change there for some reason it will only happen in one place‏

packages/framework/src/schemas/steps/channels/in-app.schema.ts
ainouzgali: ❓ **question:** Do we want to have a default target?‏

BiswaViraj: 🤔 I guess yes, `default: '_blank'`  will be appropriate

libs/shared/src/utils/bridge.utils.ts
```diff
@@ -19,12 +19,18 @@ type InAppOutput = {
...
     label: string;
-    url?: string;
+    redirect?: Redirect;
   };
+  redirect?: Redirect;
+};
+
+type Redirect = {
+  url: string;
+  target?: UrlTarget;
```
ainouzgali: can we have this 'Redirect` also exported in shared?

ainouzgali (approved):

libs/shared/src/utils/bridge.utils.ts
BiswaViraj: Do you mean to just export this type from the `bridge.utils.ts` file?

ainouzgali: If possible, maybe both Redirect and Target types can be in types folder in shared and exported from there? 
Or exported from somewhere else that makes sense to you. 
And them imported in the api file and the other places you are using it, because there are multiple usage of Redirect as well, in several different packages

apps/api/src/app/inbox/utils/types.ts
```diff
@@ -8,9 +8,15 @@ export type Subscriber = {
   subscriberId: string;
 };
 
+type Redirect = {
+  url?: string;
+  target?: UrlTarget;
+};
+
```
rifont: I think we can use the `Redirect` typing declared in `@novu/shared` instead of redefining it here.

packages/framework/src/schemas/steps/channels/in-app.schema.ts
```diff
@@ -1,10 +1,20 @@
 import { Schema } from '../../../types/schema.types';
 
+const redirectSchema = {
+  type: 'object',
+  properties: {
+    url: { type: 'string' },
```
rifont: We should also add the `format` field wherever possible to narrow the schema down.

```suggestion
    url: { type: 'string', format: 'uri' },
```

packages/framework/src/schemas/steps/channels/in-app.schema.ts
```diff
@@ -1,10 +1,20 @@
...
+    target: { type: 'string', enum: ['_self', '_blank', '_parent', '_top', '_unfencedTop'], default: '_blank' },
```
rifont, line 7: Nice addition of `default`!

packages/js/src/ws/socket.ts
```diff
@@ -65,16 +65,25 @@ const mapToNotification = ({
     primaryAction: primaryCta && {
       label: primaryCta.content,
       isCompleted: actionType === ActionTypeEnum.PRIMARY && actionStatus === NotificationActionStatus.DONE,
+      redirect: {
+        target: primaryCta.target,
+        url: primaryCta?.url,
```
rifont: ❓ **question:** Is there a specific reason for using optional chaining for `url` but not for `target`?‏

packages/js/src/ws/socket.ts
```diff
@@ -65,16 +65,25 @@ const mapToNotification = ({
...
+      },
     },
     secondaryAction: secondaryCta && {
       label: secondaryCta.content,
       isCompleted: actionType === ActionTypeEnum.SECONDARY && actionStatus === NotificationActionStatus.DONE,
+      redirect: {
+        target: secondaryCta.target,
+        url: secondaryCta?.url,
```
rifont: Same as above re: optional chaining

packages/js/src/ws/socket.ts
BiswaViraj: No, missed it to update after the Redirect refactor

libs/shared/src/entities/messages/messages.interface.ts
```diff
@@ -33,6 +32,7 @@ export interface IMessageCTA {
   type: ChannelCTATypeEnum;
   data: {
     url?: string;
+    target?: UrlTarget;
   };
```
rifont, line 36: This should use the `Redirect` typing

BiswaViraj: In the Redirect we have the url as required, but in the message entity it is not required because it is the old V1 cta buttons.

BiswaViraj: We will use the Redirect type once we introduce the new fields that we discussed yesterday about aligning the DSL

rifont (approved):

PR #6396, opened 2024-08-28
feat(api): add preferences entity and schema

davidsoderberg:


libs/application-generic/src/usecases/write-preferences/write-preferences.usecase.ts
rifont, line 1: we can rename this to `upsert-preferences` to provide a semantic description of the use-case

libs/application-generic/src/usecases/get-preferences/get-preferences.command.ts
```diff
@@ -0,0 +1,7 @@
+import { EnvironmentCommand } from '../../commands';
+
+export class GetPreferencesCommand extends EnvironmentCommand {
+  readonly subscriberId: string;
```
rifont: I think this should be typed as an optional param.

apps/api/src/bootstrap.ts
```diff
@@ -89,7 +89,13 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
   app.use(helmet());
   app.enableCors(corsOptionsDelegate);
 
-  app.setGlobalPrefix(CONTEXT_PATH + 'v1');
+  app.setGlobalPrefix(CONTEXT_PATH);
```
davidsoderberg: This and the changes below allow us to in the future add new versions to our api

davidsoderberg:
Needs to add usage of preferences in sync usecase

apps/api/src/app/inbox/inbox.module.ts
```diff
@@ -1,15 +1,18 @@
...
+import { PreferencesRepository } from '@novu/dal';
 import { AuthModule } from '../auth/auth.module';
-import { SubscribersModule } from '../subscribers/subscribers.module';
 import { IntegrationModule } from '../integrations/integrations.module';
+import { SharedModule } from '../shared/shared.module';
+import { SubscribersModule } from '../subscribers/subscribers.module';
+import { InboxController } from './inbox.controller';
+import { USE_CASES } from './usecases';
+
+const EXTERNAL_USE_CASES = [GetPreferences, UpsertPreferences, PreferencesRepository];
```
rifont: I think we should specify `PreferencesModule` in the `imports` array instead to fully encapsulate the use-cases. This is the Nest.js recommended approach and ensures we don't need to come back to this file if we update the list of use-cases.

apps/api/src/app/preferences/dtos/preferences.dto.ts
```diff
@@ -0,0 +1,42 @@
+import { Type } from 'class-transformer';
+import { IsBoolean, ValidateNested } from 'class-validator';
+
+class Preference {
```
rifont:
```suggestion
export class Preference {
```
We need to export all nested DTOs to ensure that `@nestjs/swagger` can access all the nested typings when it produces the OpenAPI spec.

apps/api/src/app/preferences/dtos/preferences.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+  @IsBoolean()
+  defaultValue: boolean;
+
+  @IsBoolean()
+  readOnly: boolean;
+}
+
+class Channels {
```
rifont: As above

```suggestion
export class Channels {
```

apps/api/src/app/preferences/preferences.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+        organizationId: user.organizationId,
+        key: FeatureFlagsKeysEnum.IS_WORKFLOW_PREFERENCES_ENABLED,
+      })
+    );
+
+    if (!isEnabled) {
+      throw new NotFoundException();
+    }
+  }
+}
```
rifont, line 80: 👏 Great use of method encapsulation

apps/api/src/app/preferences/preferences.spec.ts
rifont, line 1: Very thorough tests, well done 👏

apps/api/src/app/subscribers/subscribers.module.ts
```diff
@@ -1,15 +1,19 @@
...
-import { SubscribersController } from './subscribers.controller';
+import { GetPreferences, UpsertPreferences } from '@novu/application-generic';
+import { PreferencesRepository } from '@novu/dal';
 import { AuthModule } from '../auth/auth.module';
+import { SharedModule } from '../shared/shared.module';
 import { WidgetsModule } from '../widgets/widgets.module';
+import { SubscribersController } from './subscribers.controller';
+import { USE_CASES } from './usecases';
+
+const EXTERNAL_USE_CASES = [GetPreferences, UpsertPreferences, PreferencesRepository];
```
rifont: Same goes here, we should add `PreferencesModule` to the `imports` array to encapsulate all the use-cases

apps/api/src/app/workflows/workflow.module.ts
```diff
@@ -1,17 +1,21 @@
...
+import { USE_CASES } from './usecases';
 import { WorkflowController } from './workflow.controller';
 
+const EXTERNAL_USE_CASES = [GetPreferences, UpsertPreferences, PreferencesRepository];
+
 @Module({
   imports: [SharedModule, MessageTemplateModule, ChangeModule, AuthModule, IntegrationModule],
   controllers: [NotificationTemplateController, WorkflowController],
-  providers: [...USE_CASES],
+  providers: [...USE_CASES, ...EXTERNAL_USE_CASES],
```
rifont: Same here, replace by adding `PreferencesModule` to `imports` array

apps/api/src/bootstrap.ts
```diff
@@ -3,7 +3,7 @@ import 'newrelic';
 import '@sentry/tracing';
 
 import helmet from 'helmet';
-import { INestApplication, Logger, ValidationPipe } from '@nestjs/common';
+import { INestApplication, Logger, ValidationPipe, VersioningType } from '@nestjs/common';
```
rifont: ☑ **chore:** Remove unused import‏
```suggestion
import { INestApplication, Logger, ValidationPipe } from '@nestjs/common';
```

apps/worker/src/app/workflow/workflow.module.ts
```diff
@@ -135,6 +136,7 @@ const USE_CASES = [
   InboundEmailParse,
   ExecutionLogRoute,
   ExecuteBridgeJob,
+  GetPreferences,
```
rifont, line 139: We can import the `PreferencesModule` here too.

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -0,0 +1,159 @@
...
+    command: GetPreferencesCommand,
+  ): Promise<WorkflowOptionsPreferences> {
+    const isEnabled = await this.getFeatureFlag.execute(
+      GetFeatureFlagCommand.create({
+        userId: 'system',
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        key: FeatureFlagsKeysEnum.IS_WORKFLOW_PREFERENCES_ENABLED,
+      }),
+    );
```
rifont, line 31: We had a feature flag check on the controller too. Given that we have this check on the use-case (which would be my preference, to co-locate the FF with the feature), I think we can remove the controller check, right?

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -0,0 +1,159 @@
...
+        in_app:
+          result.channels.in_app.defaultValue || result.workflow.defaultValue,
+        sms: result.channels.sms.defaultValue || result.workflow.defaultValue,
+        email:
+          result.channels.email.defaultValue || result.workflow.defaultValue,
+        push: result.channels.push.defaultValue || result.workflow.defaultValue,
+        chat: result.channels.chat.defaultValue || result.workflow.defaultValue,
+      };
+    } catch (e) {
+      return undefined;
```
rifont: ☑ **chore:** I think we should throw an error here instead. Why are we returning undefined if an error presents? If we do need it, let's add a code comment to highlight why.‏

libs/dal/src/repositories/preferences/preferences.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+type WorkflowOptionsPreferences = {
+  workflow: PreferenceOptions;
+  channels: Record<ChannelTypeEnum, PreferenceOptions>;
+};
+
+export enum PreferencesTypeEnum {
+  SUBSCRIBER_GLOBAL = 'SUBSCRIBER_GLOBAL',
+  SUBSCRIBER_WORKFLOW = 'SUBSCRIBER_WORKFLOW',
+  USER_WORKFLOW = 'USER_WORKFLOW',
+  CODE_FIRST_WORKFLOW = 'CODE_FIRST_WORKFLOW',
```
rifont: The naming is semantic, but I think we can do better to abstract away from the V2 experience. How about the following?

```suggestion
  WORKFLOW_RESOURCE = 'WORKFLOW_RESOURCE',
```

rifont (approved):
Amazing work David! 🚀

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -0,0 +1,159 @@
...
+      (item) => item.type === PreferencesTypeEnum.USER_WORKFLOW,
+    );
+  }
+
+  private getWorkflowPreferences(items: PreferencesEntity[]) {
+    return items.find(
+      (item) => item.type === PreferencesTypeEnum.CODE_FIRST_WORKFLOW,
+    );
+  }
+
```
antonjoel82: 💭 **thought (non-blocking):** Now that we have these enum values, we could just make a `getPreferenceByType` function that takes the array and the type.

Also, these methods could either be declared as external helper functions, or static methods‏

apps/api/src/app/preferences/dtos/preferences.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+  @Type(() => Preference)
+  sms: Preference;
+
+  @ValidateNested({ each: true })
+  @Type(() => Preference)
+  chat: Preference;
+
+  @ValidateNested({ each: true })
+  @Type(() => Preference)
+  push: Preference;
```
antonjoel82: ❓ **question:** Can these also be keyed by the enum?‏

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -171,4 +195,63 @@ export class UpdatePreferences {
...
+        push: {
+          defaultValue: item.channels.push || PREFERENCE_DEFAULT_VALUE,
+          readOnly: false,
+        },
+        chat: {
+          defaultValue: item.channels.chat || PREFERENCE_DEFAULT_VALUE,
+          readOnly: false,
+        },
+      },
+    };
```
antonjoel82, line 234: 👏 **praise:** Thank for the changes here!‏

antonjoel82 (approved):
Awesome! Good job on the updates -- I think this makes things easier to use and maintains encapsulation well

apps/worker/src/app/workflow/workflow.module.ts
davidsoderberg: Should we do it here as it is not in the api service?

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
davidsoderberg: No as we do not do this in upsert, but maybe we should do it there to?

apps/worker/src/app/workflow/workflow.module.ts
rifont: Ah fair point, let's keep Worker as is for now. But in the long term, we should have a way to share such modules across both API and Worker.

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
rifont: Yes, I think it makes sense to remain consistent.

PR #6400, opened 2024-08-28
feat(js,react): inbox preferences cache

LetItRock:
Inbox preferences cache. Storing in the cache and handling the optimistic updates will simplify the work required to implement the react hooks as we won't need to duplicate the storing/optimistic updates logic.


desiprisg (approved):

BiswaViraj (approved):
🚀

PR #6408, opened 2024-08-28
build(root): Upgrade eslint to v9.x and add eslint-plugin-deprecated

rifont:

* Upgrade eslint to v9.x
* add eslint-plugin-deprecated to catch deprecation warnings in CI
* set `reportUnusedDisableDirectives: 'error'` to ensure [unused disable directives](https://eslint.org/docs/latest/use/configure/rules#report-unused-eslint-disable-comments) are cleaned up

### Screenshots

_Error is now showing for properties marked with `@deprecated`_
image

_Unused eslint disable directives are reported as errors_
image

<details>

</details>


packages/providers/src/lib/push/fcm/fcm.provider.ts
```diff
@@ -68,7 +68,7 @@ export class FcmPushProvider extends BaseProvider implements IPushProvider {
     let res;
 
     if (type === 'data') {
-      res = await this.messaging.sendEachForMulticast(
+      res = await this.messaging.sendMulticast(
```
SokratisVidros: @rifont I guess we should revert this.

SokratisVidros (approved):

packages/providers/src/lib/push/fcm/fcm.provider.ts
rifont: Yes, it's here for testing presently.

PR #6419, opened 2024-08-30
feat(react): Introduce hooks

desiprisg:
### Screenshots

<details>

</details>


packages/react/src/components/NovuProvider.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const { children, ...options } = props;
+  const novu = useMemo(() => new Novu(options), []);
+
+  return <NovuContext.Provider value={novu}>{children}</NovuContext.Provider>;
+};
+
+export const useNovu = () => {
+  const context = useContext(NovuContext);
+  if (!context) {
+    throw new Error('useNovu must be used within a NovuProvider');
```
SokratisVidros:
```suggestion
    throw new Error('useNovu must be used within a <NovuProvider />');
```

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,123 @@
...
+export type UseNotificationsResult = {
+  readAll: () => Promise<void>;
+  archiveAll: () => Promise<void>;
+  archiveAllRead: () => Promise<void>;
+  data?: Array<Notification>;
+  error?: NovuError;
+  isLoading: boolean;
+  isFetching: boolean;
+  refetch: () => Promise<void>;
+  fetchMore: () => void;
```
SokratisVidros: Isn't this a promise?

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,123 @@
...
+  archived?: boolean;
+  limit?: number;
+  onSuccess?: (data: Notification[]) => void;
+  onError?: (error: NovuError) => void;
+};
+
+export const useNotifications = (props?: UseNotificationsProps): UseNotificationsResult => {
+  const { tags, read, archived, limit, onSuccess, onError } = props || {};
+  const { notifications, on } = useNovu();
+  const [data, setData] = useState<Array<Notification>>();
```
SokratisVidros, line 18: @desiprisg This solution works with the internal useState works. Did you consider using https://react.dev/reference/react/useSyncExternalStore instead and make it listen to the Novu events? 

I feel like it might help us with the `setData` handling as it's a hook tailored to be used with underlying event-driven state managers like the Novu class.

desiprisg: will try it out

desiprisg: Tbh I like the other approach more, since we have to call list() anyway in a useEffect. It also acts as a more universal example for other framework implementations.

desiprisg: Marking as resolved as discussed offline

packages/react/src/hooks/index.ts
```diff
@@ -1 +1,4 @@
 export * from './useDataRef';
```
LetItRock: ✏ **suggestion:** can we please move `useDataRef, useBrowserTabsChannel` to the `src/hooks/internal` folder? Let's not expose these.

packages/react/src/components/NovuProvider.tsx
```diff
@@ -0,0 +1,24 @@
+import { Novu, NovuOptions } from '@novu/js';
```
LetItRock: ✏ **suggestion:** Let's keep this file under the `hooks` folder for a better tree-shaking capability.

packages/react/src/components/NovuProvider.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { createContext, useContext, useMemo } from 'react';
+
+type NovuProviderProps = NovuOptions & {
```
LetItRock: 🗒 **note (non-blocking):** we should probably expose this type

packages/react/src/hooks/useWebsocketEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+  useEffect(() => {
+    const resolveLock = requestLock(`nv.${webSocketEvent}`, () => {
+      novu.on(webSocketEvent, updateReadCount);
+    });
+
+    return () => {
+      novu.off(webSocketEvent, updateReadCount);
+      resolveLock();
+    };
+  }, []);
```
LetItRock: ☑ **chore:** This hook is based on the Web APIs BroadcastChannel and Web Lock, we need to refactor it from these dependencies, like hiding under the `if` statement when these are not available or if not possible remove.

packages/react/src/components/NovuProvider.tsx
```diff
@@ -0,0 +1,24 @@
...
+  children: JSX.Element;
+};
+
+const NovuContext = createContext<Novu | undefined>(undefined);
+
+export const NovuProvider = (props: NovuProviderProps) => {
```
LetItRock: ❓ **question:** I don't see that this provider is used internally, do you plan to do this in another pr? if yes, please create the ticket.
The hardest part would be to share it between react/solid.

playground/nextjs/src/pages/_app.tsx
```diff
@@ -1,11 +1,15 @@
...
+import { NovuProvider } from '@novu/react';
 import type { AppProps } from 'next/app';
 import Layout from './layout';
 
 export default function App({ Component, pageProps }: AppProps) {
   return (
-    <Layout>
-      <Component {...pageProps} />
-    </Layout>
+    <NovuProvider {...novuConfig}>
```
LetItRock: ✏ **suggestion:** Let's wrap only the layout of the page (or route path) where we use hooks, because otherwise, we will see doubled session requests under the network tab when visiting other pages; and I'm not sure if this won't interfere currently;

playground/nextjs/src/pages/hooks/index.tsx
```diff
@@ -0,0 +1,41 @@
+import { useNotifications, usePreferences, useCounts } from '@novu/react';
+
+export default function HooksPage() {
```
LetItRock: ✏ **suggestion:** It would be nice if we could build the "real" UI with the hooks, something similar to what we have in the Inbox Playground app, just to dogfood this better and see if we cover all main functional cases.

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
+import { useState, useEffect, useCallback } from 'react';
```
LetItRock:
```suggestion
import { useState, useEffect } from 'react';
```

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+  };
+
+  useEffect(() => {
+    setAfter(undefined);
+    setHasMore(false);
+    fetchNotifications();
+
+    on('notifications.list.updated', sync);
+    on('notifications.list.pending', sync);
+    on('notifications.list.resolved', sync);
```
LetItRock: ✏ **suggestion:** let's clear all these too

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+    const response = await notifications.list({
+      tags,
+      read,
+      archived,
+      limit,
+      after: after,
+    });
+    if (response.error) {
+      setError(response.error as NovuError);
+      onError?.(response.error as NovuError);
```
LetItRock:
```suggestion
      setError(response.error);
      onError?.(response.error);
```

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+    setData(event.data.notifications);
+    const length = event.data.notifications.length;
+    setAfter(length ? event.data.notifications[length - 1].id : undefined);
+    setHasMore(event.data.hasMore);
+  };
+
+  useEffect(() => {
+    setAfter(undefined);
+    setHasMore(false);
```
LetItRock: ❓ **question:** call `resetState` ?

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+  const [hasMore, setHasMore] = useState(false);
+  const [after, setAfter] = useState<string>();
+
+  const sync = (event: { data?: ListNotificationsResponse }) => {
+    if (!event.data) {
+      return;
+    }
+    if (!isSameFilter(event.data.filter, { tags, read, archived })) {
+      return;
+    }
```
LetItRock:
```suggestion
    if (!event.data || !isSameFilter(event.data.filter, { tags, read, archived })) {
      return;
    }
```

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+  const fetchMore = async () => {
+    if (!hasMore || isFetching) return;
+    return fetchNotifications();
+  };
+
+  const readAll = async () => {
+    const res = await notifications.readAll({ tags });
+    if (res.error) {
+      setError(res.error);
+    }
```
LetItRock: ❓ **question:** I'm not sure if we should set the error in the state, because then the question is why we are not calling `onSuccess` and `onError` props, because they are for notifications. So maybe we should just return the result in this case. WDYT?

packages/react/src/hooks/usePreferences.ts
```diff
@@ -0,0 +1,71 @@
...
+    on('preferences.list.pending', sync);
+    on('preferences.list.resolved', sync);
+  }, []);
+
+  const fetchPreferences = async () => {
+    setIsFetching(true);
+    const response = await preferences.list();
+    if (response.error) {
+      setError(response.error as NovuError);
+      onError?.(response.error as NovuError);
```
LetItRock:
```suggestion
      setError(response.error);
      onError?.(response.error);
```

packages/react/src/hooks/useCounts.ts
```diff
@@ -0,0 +1,97 @@
...
+type UseCountsProps = {
+  filters: NotificationFilter[];
+  onSuccess?: (data: Notification[]) => void;
+  onError?: (error: NovuError) => void;
+};
+
+type CountsData = {
+  count: number;
+  filter: NotificationFilter;
+};
```
LetItRock:
```suggestion
```

packages/react/src/hooks/useCounts.ts
```diff
@@ -0,0 +1,97 @@
...
+
+type UseCountsResult = {
+  counts?: Count[];
+  error?: NovuError;
+  isLoading: boolean; // initial loading
+  isFetching: boolean; // the request is in flight
+  refetch: () => Promise<void>;
+};
+
+export const useCounts = (props: UseCountsProps): UseCountsResult => {
```
LetItRock, line 25: ❓ **question:** Should we take into account that the `filters` might change?

packages/react/src/hooks/useCounts.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+  });
+
+  useEffect(() => {
+    sync();
+  }, []);
+
+  const refetch = async () => {
+    await sync();
+    return;
```
LetItRock:
```suggestion
```

packages/react/src/hooks/useCounts.ts
```diff
@@ -0,0 +1,97 @@
...
+import { NotificationFilter, NovuError } from '@novu/js';
+import { useEffect, useState } from 'react';
+import { useNovu } from '../components/NovuProvider';
+import { useWebSocketEvent } from './useWebsocketEvent';
+import { Notification } from '@novu/js';
+
+type UseCountsProps = {
+  filters: NotificationFilter[];
+  onSuccess?: (data: Notification[]) => void;
+  onError?: (error: NovuError) => void;
```
LetItRock: ☑ **chore:** I don't see we call any of these.

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+  isFetching: boolean;
+  refetch: () => Promise<void>;
+  fetchMore: () => Promise<void>;
+  hasMore: boolean;
+};
+
+export type UseNotificationsProps = {
+  tags?: string[];
+  read?: boolean;
+  archived?: boolean;
```
LetItRock, line 8: ❓ **question:** I'm wondering whether by default `archived` should be set to `false` 🤔 wdyt?

LetItRock (commented):
1. The after prop is not cleared somehow

https://github.com/user-attachments/assets/b0d8d1c4-e925-4cb4-afe6-3203f5b5403a

2. After archiving notifications the `hasMore` stays as `true`.

https://github.com/user-attachments/assets/5ab31b4d-ca0b-401f-9827-19a97697bb00

3. After refetching notifications the `after` is not cleared.

https://github.com/user-attachments/assets/46611d09-2f53-4402-b977-bcf335dd5345

packages/react/src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,130 @@
...
+
+export const useNotifications = (props?: UseNotificationsProps): UseNotificationsResult => {
+  const { tags, read, archived, limit, onSuccess, onError } = props || {};
+  const { notifications, on } = useNovu();
+  const [data, setData] = useState<Array<Notification>>();
+  const [error, setError] = useState<NovuError>();
+  const [isLoading, setIsLoading] = useState(true);
+  const [isFetching, setIsFetching] = useState(false);
+  const [hasMore, setHasMore] = useState(false);
+  const [after, setAfter] = useState<string>();
```
LetItRock: ❓ **question:** Do we need to have it in the state? we can calculate it on the fly.

packages/react/src/components/NovuProvider.tsx
desiprisg: This is meant to be in react. If we decide to share this with the `<Inbox />`, we would pass the novu instance in the `mountComponent`  method.

packages/react/src/hooks/index.ts
desiprisg: They are not exposed right now. I explicitly export the hooks that we want to expose.

playground/nextjs/src/pages/hooks/index.tsx
desiprisg: Nice idea but I'd say we should do this in another PR, and maybe include in a starter repo as well.

packages/react/src/hooks/useNotifications.ts
desiprisg: I kinda find it easier to read this way, but I'll change it :P

packages/react/src/hooks/useNotifications.ts
desiprisg: Hmm, yeah you're right

packages/react/src/hooks/useNotifications.ts
desiprisg: It seems kind of misleading to me if it has a default value.

packages/react/src/hooks/useNotifications.ts
desiprisg: Nice suggestion!

packages/react/src/hooks/useWebsocketEvent.ts
desiprisg: Is this just for server env or do you think it applies elsewere as well ?
Also, what do we want the behaviour to be in that case?

LetItRock: Ideally, for the `web` env, we want to keep it, for the `mobile` or `server` I think we should hide (if possible under if) it as it won't be available.

packages/react/src/components/NovuProvider.tsx
LetItRock: Let's have a ticket to improve this, like: "Use the NovuProvide internally in the Inbox react" or smth

playground/nextjs/src/pages/hooks/index.tsx
LetItRock: ok, I'll create a ticket

packages/react/src/hooks/useNotifications.ts
LetItRock: I think that in most cases devs who will build using hooks will use it to fetch `read/unread` notifications, that's why I'm saying to make `archived: false` by default.

packages/react/src/hooks/useWebsocketEvent.ts
desiprisg: I added it by checking if the lock api exists in this case.

packages/react/src/components/NovuProvider.tsx
desiprisg: I think this is resolved since we will go with explicit or implicit novu.

packages/react/src/hooks/NovuProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+type NovuProviderProps = NovuOptions & {
+  children: JSX.Element;
+};
+
+const NovuContext = createContext<Novu | undefined>(undefined);
+
+export const NovuProvider = (props: NovuProviderProps) => {
+  const { children, ...options } = props;
+  const novu = useMemo(() => new Novu(options), []);
```
LetItRock: We should probably create a new Novu instance when the props change.

packages/react/src/hooks/index.ts
```diff
@@ -1 +1,3 @@
-export * from './useDataRef';
+export * from './useNotifications';
+export * from './usePreferences';
+export * from './useCounts';
```
LetItRock, line 3: ✏ **suggestion:** we should export here the provider as well and only `useNovu` hook (not exported rn)

packages/react/src/hooks/index.ts
```diff
@@ -1 +1,3 @@
-export * from './useDataRef';
+export * from './useNotifications';
```
LetItRock, line 1: ☑ **chore:** btw I'm missing the `package.json` export changes for the hooks, please check how we do this in the js package and do it similarly

packages/react/src/hooks/NovuProvider.tsx
```diff
@@ -0,0 +1,30 @@
+import { Novu, NovuOptions } from '@novu/js';
+import { createContext, useContext, useMemo } from 'react';
+
+type NovuProviderProps = NovuOptions & {
+  children: JSX.Element;
+};
+
+const NovuContext = createContext<Novu | undefined>(undefined);
+
+export const NovuProvider = (props: NovuProviderProps) => {
```
LetItRock: ❓ **question:** I'm not sure what we should do with the props on the Inbox in this case:
```
<NovuProvider subscriberId="subscriberId" applicationIdentifier="applicationIdentifier">
   <Inbox subscriberId="subscriberId" applicationIdentifier="applicationIdentifier" />
</NovuProvider>
```

packages/react/src/hooks/internal/useBrowserTabsChannel.ts
```diff
@@ -0,0 +1,29 @@
...
+  onMessage,
+}: {
+  channelName: string;
+  onMessage: (args: T) => void;
+}) => {
+  const [tabsChannel] = useState(new BroadcastChannel(channelName));
+
+  const postMessage = (data: T) => {
+    tabsChannel.postMessage(data);
+  };
```
LetItRock: ☑ **chore:** this function might still fail in RN, it's exposed and used when the ws event comes

packages/react/src/components/Inbox.tsx
```diff
@@ -35,8 +36,30 @@ const DefaultInbox = (props: DefaultInboxProps) => {
...
+    <NovuProvider
+      applicationIdentifier={applicationIdentifier}
+      subscriberId={subscriberId}
+      subscriberHash={subscriberHash}
+      backendUrl={backendUrl}
+      socketUrl={socketUrl}
+    >
+      <InboxChild {...props} />
+    </NovuProvider>
+  );
```
LetItRock, line 56: Support hooks in the Inbox children components.

packages/react/src/hooks/useNotifications.ts
LetItRock: we agreed on setting it to `archive: false` by default: https://novu.slack.com/archives/C06TPLSQGUT/p1725610768168889

LetItRock (approved):

BiswaViraj (changes requested):
```
Error: This JSX tag's 'children' prop expects a single child of type 'Element', but multiple children were provided.ts(2746)

<NovuProvider {...novuConfig}>
  <div>hello</div>
  <Notifications />
</NovuProvider>
      
```
It is not accepting multiple children

packages/react/src/hooks/NovuProvider.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const useNovu = () => {
+  const context = useContext(NovuContext);
+  if (!context) {
+    throw new Error('useNovu must be used within a <NovuProvider />');
+  }
+
+  return context;
+};
+
+export const useUnsafeNovu = () => {
```
SokratisVidros, line 36: 💯

packages/react/src/hooks/index.ts
SokratisVidros: @LetItRock is this comment still pending or can we resolve it?

packages/react/src/utils/requestLock.ts
```diff
@@ -0,0 +1,28 @@
+export function requestLock(id: string, cb: (id: string) => void) {
```
SokratisVidros, line 1: ❓ Did we add locks support now or did we have it already in @novu/js? Actually I am a bit surprised this is not in @novu/js

SokratisVidros (approved):
🪝

packages/react/src/hooks/index.ts
LetItRock: @SokratisVidros we can resolve it, the exports were added

packages/react/src/utils/requestLock.ts
LetItRock: @SokratisVidros We have it for the JS package. However, we needed to copy this code for the `useWebSocketEvent` hook implementation. But we can actually expose it from the JS package as part of the `@novu/js/internal` export or smth similar.

BiswaViraj (approved):

PR #6422, opened 2024-09-01
feat(novui,framework): controls validation in the dashboard2

ainouzgali:

 Currently, validation is only with the default JSON schema validation 

In this PR we ensure that the errors object provided by the validate action maps the errors back to the correct locations in the form (both errors defined in Zod  and from Json Schem)
### Screenshots

<details>

</details>


packages/framework/src/types/validator.types.ts
```diff
@@ -6,6 +6,7 @@ export type ValidateFunction<T = unknown> = AjvValidateFunction<T> | ((data: T)
 
 export type ValidationError = {
   path: string;
+  property: string;
```
ainouzgali, line 9: The mapping of the validation error to the schema object uses the property attribute with  `.`  and not `/`

packages/framework/src/validators/json-schema.validator.ts
```diff
@@ -22,6 +22,8 @@ export class JsonSchemaValidator implements Validator<JsonSchema> {
       useDefaults: true,
       // https://ajv.js.org/options.html#removeadditional
       removeAdditional: 'failing',
+      // https://ajv.js.org/options.html#allerrors
+      allErrors: true,
```
ainouzgali: To get all errors that failed, and not only the first one

packages/framework/src/validators/json-schema.validator.ts
```diff
@@ -60,10 +62,16 @@ export class JsonSchemaValidator implements Validator<JsonSchema> {
     } else {
       return {
         success: false,
-        errors: (validateFn.errors as ErrorObject<string, Record<string, unknown>, unknown>[]).map((err) => ({
-          path: err.instancePath,
-          message: err.message as string,
-        })),
+        errors: (validateFn.errors as ErrorObject<string, Record<string, unknown>, unknown>[]).map((err) => {
+          const nestedPath = err.instancePath.split('/').join('.');
+          const requiredFieldName = err.params?.missingProperty;
```
ainouzgali: For missing required fields, we have to manually had the field name. It is not part of the `instancePath`. Otherwise we won't show those required missing errors on the input

packages/framework/src/validators/json-schema.validator.ts
ainouzgali: I see now that adding this as true is messing up `anyOf` , `allOf` etc

ainouzgali: I will remove this

SokratisVidros:
@ainouzgali can we rebase this on top of latest next before proceeding with the review?

ainouzgali:
@SokratisVidros opened a [new PR](https://github.com/novuhq/novu/pull/6428) for this

PR #6428, opened 2024-09-02
feat(framework,novui): controls validation

ainouzgali:

Opened instead of this [closed PR](https://github.com/novuhq/novu/pull/6422)

Currently, validation is only with the default JSON schema validation

In this PR we ensure that the errors object provided by the validate action maps the errors back to the correct locations in the form (both errors defined in Zod and from Json Schema)

### Screenshots

<details>

</details>


packages/framework/src/validators/json-schema.validator.ts
```diff
@@ -60,10 +60,16 @@ export class JsonSchemaValidator implements Validator<JsonSchema> {
...
-          message: err.message as string,
-        })),
+        errors: (validateFn.errors as ErrorObject<string, Record<string, unknown>, unknown>[]).map((err) => {
+          const nestedPath = err.instancePath.split('/').join('.');
+          const requiredFieldName = err.params?.missingProperty as string;
+
+          return {
+            path: err.instancePath,
+            message: err.message as string,
+            property: `${nestedPath}${requiredFieldName ? `.${requiredFieldName}` : ''}`,
```
ainouzgali: For missing required fields, we have to manually had the field name. It is not part of the instancePath. Otherwise we won't show those "required" missing errors on the input

packages/framework/src/types/validator.types.ts
```diff
@@ -6,6 +6,7 @@ export type ValidateFunction<T = unknown> = AjvValidateFunction<T> | ((data: T)
 
 export type ValidationError = {
   path: string;
+  property: string;
```
ainouzgali: The mapping of the validation error to the schema object uses the property attribute with . and not /

I was wondering why we need the path @rifont ? should we maybe have only one of those (path/property) and where needed change one if those strings?

apps/web/src/components/workflow_v2/StepEditorComponent.tsx
```diff
@@ -27,6 +27,8 @@ export const WorkflowsStepEditor = ({
   onControlsSave?: () => void;
   isSavingControls?: boolean;
 }) => {
+  const errorData = error && 'response' in error ? error?.response?.data : error;
```
antonjoel82: ☑ **chore:** We should already have this sort of logic in place somewhere, and if not, we should extract it to its own centralized implementation (in a helper function or something)‏

libs/novui/src/json-schema-components/JsonSchemaForm.tsx
```diff
@@ -46,18 +47,48 @@ export type JsonSchemaFormProps<TFormData = any> = JsxStyleProps &
...
+    return {};
+  }
+  // adding a default error for the case of errors in an array not updating correctly
+  return { __errors: ['enableShowExtraErrors'], ...toErrorSchema(errors || []) };
+};
+
+// Always return an empty array to prevent the default error messages from showing
+function transformErrors(_: RJSFValidationError[]) {
+  return [];
+}
```
antonjoel82, line 77: 👏 **praise:** Helpful comments, thank you!‏

libs/novui/src/json-schema-components/JsonSchemaForm.tsx
```diff
@@ -46,18 +47,48 @@ export type JsonSchemaFormProps<TFormData = any> = JsxStyleProps &
...
+  const formRef = useRef<Form>(null);
+
+  function customValidate(formData: TFormData, formErrors: FormValidation<TFormData>) {
+    const extraErrors = getExtraErrors(errors);
+
+    if (Object.keys(extraErrors)?.length > 0) {
+      formErrors = { ...formErrors, ...extraErrors };
+    }
+
+    return formErrors;
```
antonjoel82: ❓ **question:** Is mutating the `formErrors` object (param) the recommended pattern / way of accomplishing this?‏

antonjoel82 (approved):
👍

libs/novui/src/json-schema-components/JsonSchemaForm.tsx
ainouzgali: Took this from an example in their documentation

libs/novui/src/json-schema-components/JsonSchemaForm.tsx
```diff
@@ -76,6 +107,11 @@ export function JsonSchemaForm<TFormData = any>(props: JsonSchemaFormProps<TForm
       uiSchema={UI_SCHEMA}
       widgets={isAutocompleteEnabled ? WIDGETS : LEGACY_WIDGETS}
       validator={validator}
+      // override default behavior don't print to console when there is a validation error
+      onError={() => {}}
```
rifont, line 128: 👏 **praise:** Nice catch 👏‏

packages/framework/src/types/validator.types.ts
rifont: You're absolutely right, we should use just one of these. My preference would be to modify the existing `path` property to use the `.`-separated syntax as I feel the name is better suited to describing a full path, where `property` is more singular.

Let's modify the validators to return `.`-separated syntax on the `path` property and remove the `property` entirely.

packages/framework/src/validators/validator.test.ts
```diff
@@ -289,18 +289,22 @@ describe('validators', () => {
...
                 path: '',
+                property: '.boolVal',
               },
               {
                 message: 'must match a schema in anyOf',
                 path: '',
+                property: '',
               },
             ],
           },
```
rifont: Can we add a new test for validating an array to demonstrate how a `path` is formed for a invalid property inside array data?

apps/web/src/studio/components/workflows/step-editor/WorkflowStepEditorControlsPanel.tsx
```diff
@@ -19,6 +19,7 @@ export type OnChangeType = 'step' | 'payload';
 interface IWorkflowStepEditorControlsPanelProps {
   step: any;
   workflow: any;
+  error?: any;
```
rifont: Please import the `ValidationError` type from `@novu/framework` to use here.

Also, I think this be a plural `errors` that this component receives as `<JsonSchemaForm/>` takes an `errors` prop.

PR #6429, opened 2024-09-02
fix(js): inbox notifications component gets remounting when render notification prop changes

LetItRock:
Fixed the issue described in this ticket: https://linear.app/novu/issue/COM-199/the-inbox-notifications-component-gets-remounted-when

### Screenshots

https://github.com/user-attachments/assets/316aba7f-48d8-45cd-99fc-17718ced88d8

### DIFF
--- a/packages/js/src/ui/components/Renderer.tsx
+++ b/packages/js/src/ui/components/Renderer.tsx
@@ -49,9 +49,7 @@ type RendererProps = {
 };
 
 export const Renderer = (props: RendererProps) => {
-  const nodes = createMemo(() => {
-    return Array.from(props.nodes.keys());
-  });
+  const nodes = () => [...props.nodes.keys()];
 
   onMount(() => {
     const id = 'novu-default-css';
@@ -81,7 +79,7 @@ export const Renderer = (props: RendererProps) => {
               <CountProvider>
                 <For each={nodes()}>
                   {(node) => {
-                    const novuComponent = createMemo(() => props.nodes.get(node)!);
+                    const novuComponent = () => props.nodes.get(node)!;
                     let portalDivElement: HTMLDivElement;
                     const Component = novuComponents[novuComponent().name];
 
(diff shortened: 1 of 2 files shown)


packages/js/src/ui/components/Renderer.tsx
```diff
@@ -75,13 +79,14 @@ export const Renderer = (props: RendererProps) => {
           <FocusManagerProvider>
             <InboxProvider tabs={props.tabs}>
               <CountProvider>
-                <For each={[...props.nodes]}>
-                  {([node, component]) => {
+                <For each={nodes()}>
```
LetItRock, line 80: The nodes aka DOM element refs don't change on re-render, so it's "safe" to use them in the `For` component `each` prop as the Solid is doing comparison based on refs when rendering DOM elements. This code will make sure that the corresponding Solid components won't remount on props change.

packages/js/src/ui/components/Renderer.tsx
```diff
@@ -75,13 +79,14 @@ export const Renderer = (props: RendererProps) => {
...
+                  {(node) => {
+                    const novuComponent = createMemo(() => props.nodes.get(node)!);
```
LetItRock: However, the Solid component props might change, which is why we should memoize it and update the signal when it has been changed.

playground/nextjs/src/pages/notifications/index.tsx
```diff
@@ -1,16 +1,53 @@
 import Title from '@/components/Title';
 import { novuConfig } from '@/utils/config';
 import { Inbox, Notifications } from '@novu/react';
+import { useState } from 'react';
 
 export default function Home() {
+  const [count, setCount] = useState(0);
```
LetItRock, line 7: This is the playground code to verify that the state changes are propagated but the `Notifications` component doesn't get remounted.

SokratisVidros (approved):
💯

desiprisg (approved):
Looks good! Let's replace the memo in JSX with just an arrow function!

PR #6432, opened 2024-09-03
feat(web): add hook for workflow channel preferences

davidsoderberg:


apps/api/src/app/inbox/inbox.module.ts
```diff
@@ -6,10 +6,12 @@ import { SharedModule } from '../shared/shared.module';
 import { AuthModule } from '../auth/auth.module';
 import { SubscribersModule } from '../subscribers/subscribers.module';
 import { IntegrationModule } from '../integrations/integrations.module';
+import { GetPreferences, UpsertPreferences } from '@novu/application-generic';
+import { PreferencesRepository } from '@novu/dal';
 
 @Module({
   imports: [SharedModule, SubscribersModule, AuthModule, IntegrationModule],
-  providers: [...USE_CASES],
+  providers: [...USE_CASES, GetPreferences, UpsertPreferences, PreferencesRepository],
```
antonjoel82, line 14:
```suggestion
  providers: [...USE_CASES],
```

✏ **suggestion:** ‏Add these to `USE_CASES` to follow conventions

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -171,4 +194,51 @@ export class UpdatePreferences {
...
+            email: {
+              defaultValue: item.channels.email || true,
+              readOnly: false,
+            },
+            push: {
+              defaultValue: item.channels.push || true,
+              readOnly: false,
+            },
+            chat: {
+              defaultValue: item.channels.chat || true,
```
antonjoel82, line 231:
```suggestion
            defaultValue: item.enabled || PREFERENCE_DEFAULT_VALUE,
            readOnly: false,
          },
          channels: {
            in_app: {
              defaultValue: item.channels.in_app || PREFERENCE_DEFAULT_VALUE,
              readOnly: false,
            },
            sms: {
              defaultValue: item.channels.sms || PREFERENCE_DEFAULT_VALUE,
              readOnly: false,
            },
            email: {
              defaultValue: item.channels.email || PREFERENCE_DEFAULT_VALUE,
              readOnly: false,
            },
            push: {
              defaultValue: item.channels.push || PREFERENCE_DEFAULT_VALUE,
              readOnly: false,
            },
            chat: {
              defaultValue: item.channels.chat || PREFERENCE_DEFAULT_VALUE,
```

🤓 **nitpick (non-blocking):** ‏Extract a default value as a constant for easier changes‏

apps/api/src/app/preferences/dtos/preferences.dto.ts
```diff
@@ -0,0 +1,42 @@
+import { Type } from 'class-transformer';
+import { IsBoolean, ValidateNested } from 'class-validator';
+
+class Preference {
```
antonjoel82, line 4: ☑ **chore:** Please rename to something better scoped. `Preference` is very generic, so we should scope it more to our use case‏

apps/api/src/app/preferences/preferences.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+        organizationId: user.organizationId,
+        key: FeatureFlagsKeysEnum.IS_WORKFLOW_PREFERENCES_ENABLED,
+      })
+    );
+
+    if (isEnabled) {
+      return;
+    }
+
+    throw new NotFoundException();
```
antonjoel82, line 84:
```suggestion
    if (!isEnabled) {
       throw new NotFoundException();
    }
```

apps/api/src/app/preferences/preferences.module.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const PROVIDERS = [PreferencesRepository, UpsertPreferences, GetPreferences];
+
+@Module({
+  imports: [SharedModule],
+  providers: [...PROVIDERS],
+  controllers: [PreferencesController],
+  exports: [],
+})
+export class PreferencesModule implements NestModule {
```
antonjoel82, line 16: ☑ **chore:** Sorry to sound like a broken record, but again this seems like too generic of a name for a top-level module, so I'd recommend a name more like `WorkflowPreferences`‏

apps/api/src/bootstrap.ts
```diff
@@ -89,7 +89,13 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
...
   app.enableCors(corsOptionsDelegate);
 
-  app.setGlobalPrefix(CONTEXT_PATH + 'v1');
+  app.setGlobalPrefix(CONTEXT_PATH);
+
+  app.enableVersioning({
+    type: VersioningType.URI,
+    defaultVersion: '1',
+    prefix: 'v',
+  });
```
antonjoel82, line 98: ❓ **question:** Should we make this change in this PR given that it affects everything in API? It seems like this may be better done in isolation to avoid "lump sum" changes having to be rolled back‏

libs/application-generic/src/usecases/get-preferences/get-preferences.command.ts
```diff
@@ -0,0 +1,7 @@
+import { EnvironmentCommand } from '../../commands';
+
+export class GetPreferencesCommand extends EnvironmentCommand {
+  subscriberId?: string;
+
```
antonjoel82, line 5:
```suggestion
```

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.command.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class UpsertPreferencesCommand extends EnvironmentCommand {
+  @IsDefined()
+  readonly preferences: DiscoverWorkflowOutputPreferences;
+
+  subscriberId?: string;
+
+  userId?: string;
+
+  templateId?: string;
```
antonjoel82, line 14: ❓ **question:** Would this not be required?‏

libs/dal/src/repositories/preferences/preferences.entity.ts
```diff
@@ -0,0 +1,48 @@
...
+    defaultValue: boolean;
+    readOnly: boolean;
+  };
+  channels: {
+    [key in (typeof ChannelTypeEnum)[keyof typeof ChannelTypeEnum]]?: {
+      defaultValue: boolean;
+      readOnly: boolean;
+    };
+  };
+};
```
antonjoel82, line 19:
```suggestion
type PreferenceOptions = {
    defaultValue: boolean;
    readOnly: boolean;
};

type WorkflowOptionsPreferences = {
  workflow: PreferenceOptions;
  channels: Record<ChannelTypeEnum, PreferenceOptions>;
};
```

or potentially it would need to be `Partial<Record<ChannelTypeEnum, PreferenceOptions>>;` if it doesn't have to be fully populated each time

libs/dal/src/repositories/preferences/preferences.schema.ts
```diff
@@ -0,0 +1,104 @@
...
+        defaultValue: {
+          type: Schema.Types.Boolean,
+          default: true,
+        },
+        readOnly: {
+          type: Schema.Types.Boolean,
+          default: false,
+        },
+      },
+      channels: {
```
antonjoel82, line 42: ❓ **question:** Should these all be keyed by `ChannelTypeEnum `?

I.e. `[ChannelTypeEnum.EMAIL]: { ...`‏

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -0,0 +1,125 @@
...
+        item.actor === PreferencesActorEnum.SUBSCRIBER &&
+        item._templateId == command.templateId
+    );
+
+    const preferences = [
+      workflowPreferences,
+      userPreferences,
+      subscriberGlobalPreferences,
+      subscriberWorkflowPreferences,
+    ]
```
antonjoel82, line 89: 💭 **thought (non-blocking):** It may be worthwhile to add a comment that order matters and why / how that ordering is chosen‏

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -0,0 +1,125 @@
...
+      (item) =>
+        item.actor === PreferencesActorEnum.SUBSCRIBER &&
+        item._templateId === undefined
+    );
+
+    const subscriberWorkflowPreferences = items.find(
+      (item) =>
+        item.actor === PreferencesActorEnum.SUBSCRIBER &&
+        item._templateId == command.templateId
+    );
```
antonjoel82, line 82: 💭 **thought (non-blocking):** There are a lot of things deriving data from one single data source. Please add comments and/or consider extracting to helper functions (selectors) that can be named semantically and reused.‏

apps/web/src/hooks/useWorkflowChannelPreferences.ts
```diff
@@ -0,0 +1,74 @@
...
+import { QueryKeys } from '../api/query.keys';
+import { useDiscover } from '../studio/hooks/useBridgeAPI';
+import { useStudioState } from '../studio/StudioStateProvider';
+import { errorMessage, successMessage } from '../utils/notifications';
+import { useNovuAPI } from './useNovuAPI';
+
+type ChannelPreference = {
+  defaultValue: boolean;
+  readOnly: boolean;
+};
```
antonjoel82, line 13: 🤓 **nitpick (non-blocking):** ‏ We're re-defining a type from our own back-end. Is there a way we can reuse it?‏

apps/web/src/hooks/useWorkflowChannelPreferences.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export type WorkflowChannelPreferences = {
+  workflow: ChannelPreference;
+  channels: {
+    [key in (typeof ChannelTypeEnum)[keyof typeof ChannelTypeEnum]]: ChannelPreference;
+  };
+};
```
antonjoel82, line 20: 💭 **thought (non-blocking):** Same thing here‏

apps/web/src/hooks/useWorkflowChannelPreferences.ts
```diff
@@ -0,0 +1,74 @@
...
+  }, [discoveryData, cloudData, isLocalStudio, workflowId]);
+
+  const { mutateAsync: createWorkflowChannelPreferences, isLoading: isCreating } = useMutation<
+    WorkflowChannelPreferences,
+    IResponseError,
+    WorkflowChannelPreferences
+  >((data) => api.upsertPreferences(workflowId, data), {
+    onSuccess: () => {
+      refetch();
+      successMessage('Workflow Channel Preferences created');
```
antonjoel82, line 61:
```suggestion
      successMessage('Workflow Channel Preferences updated');
```

apps/web/src/hooks/useWorkflowChannelPreferences.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    onError: () => {
+      errorMessage('Failed to create Workflow Channel Preferences');
```
antonjoel82, line 64:
```suggestion
      errorMessage('Failed to update Workflow Channel Preferences');
```

apps/web/src/hooks/useWorkflowChannelPreferences.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export type WorkflowChannelPreferences = {
+  workflow: ChannelPreference;
+  channels: {
+    [key in (typeof ChannelTypeEnum)[keyof typeof ChannelTypeEnum]]: ChannelPreference;
+  };
+};
+
+export const useWorkflowChannelPreferences = (
```
antonjoel82, line 22: 💭 **thought (non-blocking):** While I realize the ticket says "create _a‏_ hook", I don't think it's beneficial to put all of this behavior into a single hook. I think these could be three separate hooks, which can then be better used in their respective contexts: 

1. `useUpdateWorkflowChannelPreferences` - Handles only the upsert functionality. I do think we should use `update` because in reality, we already have default preferences available, so it is really "updating" from most perspectives
2. `useStudioWorkflowChannelPreferences` - Gets preferences only for local studio.
3. `useCloudWorkflowChannelPreferences` - Gets preferences only for the cloud.

I don't think it makes sense to merge the two "getters" (2 & 3) when they'll be used in different components accordingly. By separating the `update` out, we can compose the hooks however we want without bringing in unnecessary re-renders

apps/api/src/app/subscribers/subscribers.module.ts
```diff
@@ -5,11 +5,13 @@ import { USE_CASES } from './usecases';
...
 import { AuthModule } from '../auth/auth.module';
 import { WidgetsModule } from '../widgets/widgets.module';
+import { GetPreferences, UpsertPreferences } from '@novu/application-generic';
+import { PreferencesRepository } from '@novu/dal';
 
 @Module({
   imports: [SharedModule, AuthModule, TerminusModule, forwardRef(() => WidgetsModule)],
   controllers: [SubscribersController],
-  providers: [...USE_CASES],
+  providers: [...USE_CASES, GetPreferences, UpsertPreferences, PreferencesRepository],
```
antonjoel82, line 14: 🤓 **nitpick (non-blocking):** Add to `USE_CASES` instead? If not, please comment in code why not‏

apps/api/src/app/workflows/workflow.module.ts
```diff
@@ -7,11 +7,13 @@ import { ChangeModule } from '../change/change.module';
...
 import { IntegrationModule } from '../integrations/integrations.module';
 import { WorkflowController } from './workflow.controller';
+import { GetPreferences, UpsertPreferences } from '@novu/application-generic';
+import { PreferencesRepository } from '@novu/dal';
 
 @Module({
   imports: [SharedModule, MessageTemplateModule, ChangeModule, AuthModule, IntegrationModule],
   controllers: [NotificationTemplateController, WorkflowController],
-  providers: [...USE_CASES],
+  providers: [...USE_CASES, GetPreferences, UpsertPreferences, PreferencesRepository],
```
antonjoel82, line 16: 🤓 **nitpick (non-blocking):** Add to `USE_CASES` instead? If not, please comment in code why not‏

apps/api/src/app/preferences/preferences.spec.ts
```diff
@@ -0,0 +1,816 @@
...
+          environmentId: session.environment._id,
+          organizationId: session.organization._id,
+          templateId: workflowId,
+        })
+      );
+
+      expect(workflowPreferences.preferences.workflow.readOnly).to.be.true;
+    });
+
+    it('should validate preferences', async function () {
```
antonjoel82, line 304: 🤓 **nitpick (non-blocking):** Please consider separating into distinct tests; it is easier to tell what's failing when separate tests cover one piece of functionality‏

apps/api/src/app/preferences/preferences.spec.ts
antonjoel82, line 1: 👏 **praise:** Thanks for adding all these tests!
‏

antonjoel82 (commented):
Thanks for all the hard work on this David! Some high level feedback:

I mentioned it in some files, but there are too many instances to call it out each time: `Preferences` seems too generic and doesn't accurately or semantically capture the intent of these preferences. If we were to later allow the user to configure their date-time format or something else similar, those would also be Preferences, but wouldn't fit under this umbrella.

On a more technical note, I'm a bit concerned with opening ourselves up to the heterogenous nature of the stored Preferences -- having something that can be 4 different things, and putting all the logic in a single read usecase (`get-preferences`) makes it harder to work with the data (for reporting, debugging, etc). I would suggest that perhaps instead of computing the 4 different types on retrieval that instead we do so at upsert and store it as an enum string  (`preferenceType`) on the record? That way, it's more explicit what the data is and can be queried or reported on accordingly. Either way though, I think the logic for classifying a record should be extracted to static helpers or some other sort of pure function.

davidsoderberg:
As the rebase of the api PR made this super messy and I agree the hooks should be split to multiple, I close this one and create a new PR

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.command.ts
davidsoderberg: No as we like to allow subscriber to have global preferences and then it is not connected to a workflow and template id will not be defined.

apps/api/src/app/preferences/dtos/preferences.dto.ts
davidsoderberg: Me and @rifont see `Preference` as a top level domain and may not only be related to a workflow in the future that is why we like to keep it generic in its naming.

apps/api/src/app/preferences/preferences.module.ts
davidsoderberg: Me and @rifont see `Preference` as a top level domain and may not only be related to a workflow in the future that is why we like to keep it generic in its naming.

PR #6433, opened 2024-09-03
fix(api): mid and after request logging

ChmaraX:
This PR relates to previous logging improvements: https://github.com/novuhq/novu/pull/6339

We need both `this.logger.assign()` and `customProps` methods to cover props in mid-request logs and request completed/errored logs, one of them is not enough to cover both it seems

### DIFF
--- a/libs/application-generic/src/logging/index.ts
+++ b/libs/application-generic/src/logging/index.ts
@@ -41,9 +41,7 @@ export function getLogLevel() {
   if (loggingLevelArr.indexOf(logLevel) === -1) {
     // eslint-disable-next-line no-console
     console.log(
-      `${logLevel}is not a valid log level of ${
-        loggingLevelArr
-      }. Reverting to info.`,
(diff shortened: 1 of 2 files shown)


libs/application-generic/src/logging/index.ts
```diff
@@ -107,7 +105,7 @@ export function createNestLoggingModuleOptions(
   // eslint-disable-next-line no-console
   console.log(
     `Selected Log Transport ${!transport ? 'None' : 'pino-pretty'}`,
-    loggingLevelSet,
+    loggingLevelSet
```
rifont: 🗒 **note (non-blocking):** It looks like your Eslint configuration isn't working correctly. Can you update your IDE settings to disable the builtin IDE formatter and try again? I'm expecting to see a comma present at the end of this line (and lines above that have formatting changes)‏

rifont (approved):

PR #6436, opened 2024-09-03
fix(api,shared): added trigger route to extendedBodySizeRoutes and type string to IAttachmentOptions attachments

teawithfruit:
There are problems with the upload size of email attachments. To activate the extended size directly for the trigger route, I have added this PR.
I've added type string to file of IAttachmentOptions, to make the use of base64 encoded attachments possible.

https://discord.com/channels/895029566685462578/1238764906392850472


SokratisVidros (approved):

apps/api/src/bootstrap.ts
```diff
@@ -23,6 +23,7 @@ const compression = require('compression');
 
 const extendedBodySizeRoutes = [
   '/v1/events',
+  '/v1/events/trigger',
```
djabarovgeorge, line 26: This line won't fix the user's problem:
1. The line isn't necessary because Next.js automatically accepts the prefix as the path. You can check it in the `Path examples` section here: https://expressjs.com/en/4x/api.html#app.use.
2. The `extendedBodySizeRoutes` array actually limits routes to 20 MB, rather than increasing the limit. We should probably rename it to make that clearer.

djabarovgeorge (changes requested):
@teawithfruit Thanks for taking the time in order to solve the issue, although i am not sure it will solve the issue, we need to check why the isaiahdahl is getting this error. Did you had a chance to test this fix locally and see that it fixes the error?

apps/api/src/bootstrap.ts
teawithfruit: You are right, the URL should have worked before. 
Could it be that there are also some issues with the storage provider implementation?
When I test it locally with the localstack container, it seems to work. But when I try to run it with Azure Blob Storage, the error occurs.

But the second change regarding the attachment type should be approved from your side. A base64 file upload is currently not possible without this change.

PR #6437, opened 2024-09-03
chore(root): update Readme.md

paulwer:
Brevo has renamed a while ago, so this change will provide better user clarity, what is supported.

https://www.globenewswire.com/news-release/2023/05/04/2661442/0/en/Sendinblue-Becomes-Brevo-Reflecting-Company-Growth-and-Platform-Evolution.html


SokratisVidros (approved):

paulwer:
@SokratisVidros @jainpawan21 The naming is used in several other places, like the product-website and should be changed there as well.

SokratisVidros:
@paulwer Do you have specific points so that we can update them ASAP?

paulwer:
Here are 3 places, which I came along:

image
image
https://docs.novu.co/integrations/providers/email/sendinblue

Btw.: Nothing to stress here, just a recomendation :)

SokratisVidros:
Thank you. We will take care of it.

paulwer:
found another one, but in the application:
image

only the email provider. the sms provider was already changed to brevo

PR #6444, opened 2024-09-04
fix(js,react): inbox support custom navigate function for the relative redirect urls

LetItRock:
Inbox support a custom navigate function for SSR and SPA with the `routerPush` prop.

### Screenshots

https://github.com/user-attachments/assets/6ea4bee6-928f-40c1-a2d3-4804c2a43972

### DIFF
--- a/packages/js/src/ui/components/Notification/DefaultNotification.tsx
+++ b/packages/js/src/ui/components/Notification/DefaultNotification.tsx
@@ -1,7 +1,7 @@
 import { createEffect, createMemo, createSignal, JSX, Show } from 'solid-js';
 
 import type { Notification } from '../../../notifications';
-import { ActionTypeEnum } from '../../../types';
+import { ActionTypeEnum, Redirect } from '../../../types';
 import { useInboxContext, useLocalization } from '../../context';
 import { cn, DEFAULT_REFERRER, DEFAULT_TARGET, formatToRelativeTime, useStyle } from '../../helpers';
 import { Archive, ReadAll, Unarchive, Unread } from '../../icons';
@@ -20,7 +20,7 @@ type DefaultNotificationProps = {
 export const DefaultNotification = (props: DefaultNotificationProps) => {
   const style = useStyle();
   const { t, locale } = useLocalization();
-  const { status } = useInboxContext();
+  const { navigate, status } = useInboxContext();
   const [minutesPassed, setMinutesPassed] = createSignal(0);
   const date = createMemo(() => {
     minutesPassed(); // register as dep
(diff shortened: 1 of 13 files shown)


packages/js/src/ui/components/Notification/DefaultNotification.tsx
```diff
@@ -36,6 +36,15 @@ export const DefaultNotification = (props: DefaultNotificationProps) => {
...
   });
 
+  const handleUrlNavigation = (url?: string, target?: Redirect['target']) => {
+    const isRelativeUrl = url?.startsWith('/');
+    if (url && !isRelativeUrl) {
+      window.open(url, target, DEFAULT_REFERRER);
+    } else if (url && isRelativeUrl) {
+      navigate(url);
+    }
+  };
```
LetItRock: handle the redirect URL navigation:
- when the URL is absolute use the `window.open`
- otherwise use: `props.routerPush` or `window.history.pushState`

packages/js/src/ui/context/InboxContext.tsx
```diff
@@ -52,6 +54,18 @@ export const InboxProvider = (props: InboxProviderProps) => {
...
+  const navigate = (path: string) => {
+    if (props.routerPush) {
+      props.routerPush(path);
+
+      return;
+    }
+
+    const url = new URL(path, window.location.href);
+    const pushState = window.history.pushState.bind(window.history);
+    pushState({}, '', url);
```
LetItRock: `{}` is empty state object, `''` is a safe call for unused historical param

packages/js/src/ui/components/Notification/DefaultNotification.tsx
SokratisVidros: @LetItRock allow me to suggest the following:

Navigation logic should live in one place and handled by a single function. I suggest we merge this logic inside the default [navigate function](https://github.com/novuhq/novu/pull/6444/files#diff-f3efe00becd29459f65a5eed571053b33d0f06dc5fff9fab98a9d1cfa69917e6R57). 

Target should be an optional argument.

That way, the default function is smart enough to handle all cases.

SokratisVidros (changes requested):

packages/js/src/ui/components/Notification/DefaultNotification.tsx
LetItRock: yes, you are right, I noticed that after I opened PR, will update it

BiswaViraj (approved):

SokratisVidros (approved):

PR #6446, opened 2024-09-05
feat(api): make trial last 14 days

ChmaraX:
Verified in Stripe test mode

We use monthly `billingInterval` and the trial will now last 14 days but that shouldn't be a problem, after the rial ends, Stripe will automatically transition the subscription to the regular billing cycle, which in this case is monthly. The flow should be as following:

Trial Period:

1. The customer starts their 14-day trial.
2. They are not charged during this period.
3. The subscription status will be "trialing".

End of Trial:

1. After 14 days the trial ends.
2. The first charge occurs immediately after the trial ends.
3. This becomes the anchor date for future billing cycles.
4. Alternatively if the charge fails, the tier level returns to `free`

Subsequent Billing:

1. After the initial charge, the customer will be billed monthly from the date the trial ended.

Related EE PR: https://github.com/novuhq/packages-enterprise/pull/212
Related docs PR: https://github.com/novuhq/docs/pull/689


apps/web/src/ee/billing/utils/freeTrial.constants.ts
```diff
@@ -1,5 +1,5 @@
 export const COLOR_WARNING = '#FDE044';
-export const WARNING_LIMIT_DAYS = 10;
+export const WARNING_LIMIT_DAYS = 3;
```
ChmaraX: warning at roughly 20% remaining

apps/web/tests/billing.spec.ts
```diff
@@ -12,6 +13,7 @@ test.describe('Billing', () => {
 
   test.beforeEach(async ({ page }) => {
     await initializeSession(page, { noTemplates: true });
+    await setFeatureFlag(page, FeatureFlagsKeysEnum.IS_IMPROVED_BILLING_ENABLED, true);
```
ChmaraX: We don't need 2 types of tests here based on FF - the number of trial days constant doesn't really matter - if it works for number N it will work in same way for every other.

We also don't support LD in e2e tests therefore we can't make 2 test variations here - this test needs to be changed in the same commit that will remove the feature flag

SokratisVidros (approved):

PR #6447, opened 2024-09-05
feat(web): Add Workflow Preferences for Cloud & Studio

davidsoderberg:
### Context

- https://linear.app/novu/issue/NV-4205/implement-channelpreferences-component-in-cloud-studio
- https://linear.app/novu/issue/NV-4201/create-channelpreferences-component-in-web

### Changes
- Setup settings / preferences side panel for editor V2 and studio
- Handle change-tracking for the contents of the side panel

https://www.loom.com/share/4309d6859b754ce8829a6453e5476177?sid=5344b6f5-e0d0-4983-b5e7-686ece3c5fed


apps/api/src/app/preferences/preferences.module.ts
```diff
@@ -1,10 +1,10 @@
 import { MiddlewareConsumer, Module, NestModule } from '@nestjs/common';
-import { GetPreferences, UpsertPreferences } from '@novu/application-generic';
+import { getFeatureFlag, GetPreferences, UpsertPreferences } from '@novu/application-generic';
 import { PreferencesRepository } from '@novu/dal';
 import { SharedModule } from '../shared/shared.module';
 import { PreferencesController } from './preferences.controller';
 
-const PROVIDERS = [PreferencesRepository, UpsertPreferences, GetPreferences];
+const PROVIDERS = [PreferencesRepository, UpsertPreferences, GetPreferences, getFeatureFlag];
```
antonjoel82: ❓ **question:** Was this accidentally omitted @davidsoderberg? I saw this in other modules that had feature flags‏

apps/web/src/studio/components/workflows/preferences/StudioWorkflowSettingsSidePanel.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { setValue } = useFormContext<WorkflowDetailFormContext>();
+
+  const workflow = useMemo(() => {
+    return data?.workflows?.find((workflow) => workflow.workflowId === workflowId);
+  }, [data, workflowId]);
+
+  useEffect(() => {
+    if (workflow) {
+      setValue('general.workflowId', workflow.workflowId);
+      // TODO: how are we handling workflow name for a workflow from framework?
```
antonjoel82: 🗒 **note (non-blocking):** Unsure if we should just the `name` input for Studio since I don't believe we have this data for Framework at the moment‏

apps/web/src/studio/components/workflows/preferences/WorkflowDetailFormContextProvider.tsx
```diff
@@ -0,0 +1,41 @@
...
+    name: '',
+  },
+  preferences: DEFAULT_WORKFLOW_PREFERENCES,
+};
+
+export const WorkflowDetailFormContextProvider: FC<PropsWithChildren<IWorkflowDetailFormContextProviderProps>> = ({
+  children,
+}) => {
+  const formValues = useForm<WorkflowDetailFormContext>({
+    // resolver: zodResolver({} as any),
```
ainouzgali: leftover?

apps/web/src/hooks/workflowChannelPreferences/useUpdateWorkflowChannelPreferences.ts
```diff
@@ -17,6 +18,7 @@ export const useUpdateWorkflowChannelPreferences = (
     WorkflowChannelPreferences
   >((data) => api.upsertPreferences(workflowId, data), {
     onSuccess: () => {
+      onSuccess();
```
ainouzgali:
```suggestion
      onSuccess?.();
```
Could be done like that, instead of  setting empty default value ()=>{}

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
```diff
@@ -0,0 +1,62 @@
+import { INotificationTemplate } from '@novu/shared';
+import { useEffect, useMemo } from 'react';
+import { useFormContext, SubmitHandler } from 'react-hook-form';
+import { useUpdateTemplate } from '../../../api/hooks/index';
+import { useUpdateWorkflowChannelPreferences } from '../../../hooks/workflowChannelPreferences/useUpdateWorkflowChannelPreferences';
+import { WorkflowDetailFormContext } from '../../../studio/components/workflows/preferences/WorkflowDetailFormContextProvider';
+
+type UseWorkflowDetailPageFormProps = {
+  templateId: string;
```
ainouzgali, line 11: ✏ **suggestion:** maybe call it workflowId? as we are trying to deprecate template naming ‏

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
```diff
@@ -0,0 +1,62 @@
...
+  workflow: INotificationTemplate | undefined;
+};
+
+export const useWorkflowDetailPageForm = ({ templateId, workflow }: UseWorkflowDetailPageFormProps) => {
+  const { formState, setValue, resetField, reset, handleSubmit, getValues, watch } =
+    useFormContext<WorkflowDetailFormContext>();
+
+  const workflowName = watch('general.name');
```
ainouzgali: `useWatch` has better performance than `watch`

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
```diff
@@ -0,0 +1,48 @@
...
+      updateSettings({ ...settings, [fieldKey]: value });
+    };
+
+  return (
+    <Stack gap="150">
+      <Input label="Workflow name" value={settings.name} onChange={onChange('name')} disabled={areSettingsDisabled} />
+      <Input
+        label="Identifier"
+        description="Must be unique and all lowercase, using - only"
+        rightSection={<IconButton Icon={copied ? IconCheck : IconCopyAll} onClick={() => copy(settings.workflowId)} />}
```
ainouzgali: What if there is an error here? We are not passing it. Like an illegal identifier or an empty identifier (should be required)

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const { copied, copy } = useClipboard();
+  const onChange =
+    (fieldKey: keyof WorkflowGeneralSettings) => (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      const value = event.target.value;
+      updateSettings({ ...settings, [fieldKey]: value });
+    };
+
+  return (
+    <Stack gap="150">
+      <Input label="Workflow name" value={settings.name} onChange={onChange('name')} disabled={areSettingsDisabled} />
```
ainouzgali: Changing this value applies the changed name only after closing the sidebar, but it can be confusing because you see the title in the same screen 
Screenshot 2024-09-08 at 13 51 01

apps/web/src/hooks/workflowChannelPreferences/useUpdateWorkflowChannelPreferences.ts
```diff
@@ -17,6 +18,7 @@ export const useUpdateWorkflowChannelPreferences = (
...
       successMessage('Workflow Channel Preferences updated');
```
ainouzgali: ❓ **question:** why are we having a separate success message for preferences? When clicking save button it updates all changes - both of preferences and changes in the general tab. 
Follow up: when saving after changes only in the general tab there is no feedback to the user - no success message - only changing the save button to be disabled again ‏

apps/web/src/studio/components/workflows/preferences/StudioWorkflowSettingsSidePanel.tsx
ainouzgali: Maybe the name input should be hidden for now? As we are showing it as empty + disabled

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
```diff
@@ -0,0 +1,48 @@
...
+  settings,
+  updateSettings,
+  areSettingsDisabled,
+}) => {
+  const { copied, copy } = useClipboard();
+  const onChange =
+    (fieldKey: keyof WorkflowGeneralSettings) => (event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      const value = event.target.value;
+      updateSettings({ ...settings, [fieldKey]: value });
+    };
```
ainouzgali: ❓ **question:** why aren't we using a controller on each of these fields separately? What is the reason behind this single onChange?‏

apps/api/src/app/preferences/preferences.module.ts
davidsoderberg: It should not be needed as the SharedModule exports it and this module imports that module.

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
antonjoel82: It's a bit confusing since now we call the `kebab-case-name` "workflowId" in some places, but this actually refers to the Mongo ID.

apps/web/src/hooks/workflowChannelPreferences/useUpdateWorkflowChannelPreferences.ts
antonjoel82: I did not implement that part originally, but agree, I should have removed the message once I combined things into a single "save" button. I'll do it now!

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
antonjoel82: Thanks! I was following the example from `TemplateEditorFormProvider`

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
antonjoel82: Good call-out -- I had not gotten to validation / error handling, and have to add it here

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
antonjoel82: Good find -- thank you. I'm not sure why this is happening, but could perhaps be a config in `react-hook-form`? I'll have to investigate further

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
antonjoel82: Because I originally tried to write this component to be independent of `react-hook-form` and its management, and then be wrapped in a Controller that would inject this behavior. I will update this tomorrow when I'm online

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
antonjoel82: @ainouzgali Would you have any other suggestions for ways to name it? One that came to mind was `workflowFriendlyId` or `workflowHumanId`

ainouzgali (approved):
🌟

rifont (approved):
🚀

PR #6454, opened 2024-09-09
feat(js): Fix events sharing by replacing singleton with DI

BiswaViraj:

so when we render
```
<Inbox  1/>
<Inbox  2/>
```
During the initialisation of `Inbox 1` it creates the `singletons` of inboxService and eventEmitter
Then, during the `Inbox 2` init, it re-uses the previously created `singleton` instead of creating new instances. But it does update the authorization header of the fetch requests to be of `token` from the `Inbox 2`.
So when you make requests from `Inbox 1` it uses the token of Inbox 2  and throws an error if both Inbox have different `subscriberId`.

- In this PR I replaced the `singleton` by `DI` which enables multiple independent instances of Inbox
### Screenshots

https://github.com/user-attachments/assets/51dff6e2-0e2e-45df-b706-05caee3aafef

<details>

</details>


packages/js/src/event-emitter/novu-event-emitter.ts
```diff
@@ -1,22 +1,10 @@
...
 
-  static getInstance(options?: SingletonOptions): NovuEventEmitter {
-    if (options?.recreate) {
-      NovuEventEmitter.#instance = new NovuEventEmitter();
-    }
-
-    return NovuEventEmitter.#instance;
-  }
-
-  private constructor() {
```
BiswaViraj, line 19: removed the singleton

packages/js/src/utils/inbox-service-singleton.ts
BiswaViraj, line 1: Deleted the singleton file

packages/js/src/novu.ts
```diff
@@ -5,29 +5,46 @@ import { Session } from './session';
...
   constructor(options: NovuOptions) {
-    InboxServiceSingleton.getInstance({ backendUrl: options.backendUrl ?? PRODUCTION_BACKEND_URL });
-    this.#emitter = NovuEventEmitter.getInstance({ recreate: true });
-    this.#session = new Session({
-      applicationIdentifier: options.applicationIdentifier,
-      subscriberId: options.subscriberId,
-      subscriberHash: options.subscriberHash,
-    });
+    this.#inboxService = new InboxService({ backendUrl: options.backendUrl ?? PRODUCTION_BACKEND_URL });
+    this.#emitter = new NovuEventEmitter();
```
BiswaViraj, line 24: initialise the services and re-use the same instances of it for other classes

packages/js/src/notifications/helpers.ts
```diff
@@ -240,12 +274,16 @@ export const revertAction = async ({
 
 const getNotificationDetails = (
   args: ReadArgs | UnreadArgs | ArchivedArgs | UnarchivedArgs,
-  update: Partial<Notification>
+  update: Partial<Notification>,
+  instances: {
```
SokratisVidros: Maybe call this `dependencies`

packages/js/src/notifications/notification.ts
```diff
@@ -25,9 +24,9 @@ export class Notification implements Pick<NovuEventEmitter, 'on' | 'off'>, Inbox
   readonly redirect: InboxNotification['redirect'];
   readonly data?: InboxNotification['data'];
 
-  constructor(notification: InboxNotification) {
-    this.#emitter = NovuEventEmitter.getInstance();
-    this.#inboxService = InboxServiceSingleton.getInstance();
+  constructor(notification: InboxNotification, emitter: NovuEventEmitter, inboxService: InboxService) {
```
SokratisVidros, line 27: This is part of the public API of the SDKs as the Notification class is pass to callbacks. Should we maybe add protected specific setters for emitter and inbox service?

packages/js/src/novu.ts
```diff
@@ -1,42 +1,62 @@
...
+    this.#session = new Session(
+      {
+        applicationIdentifier: options.applicationIdentifier,
+        subscriberId: options.subscriberId,
+        subscriberHash: options.subscriberHash,
+      },
+      this.#inboxService,
```
SokratisVidros, line 31: Ditto (see Notification class)

packages/js/src/preferences/helpers.ts
```diff
@@ -19,13 +19,19 @@ export const updatePreference = async ({
...
+        ? new Preference(
+            {
+              ...args.preference,
+              channels: {
+                ...args.preference.channels,
+                ...channelPreferences,
+              },
             },
-          })
+            {
```
SokratisVidros, line 30: Ditto (see Notification) class. Also here they are passed as an object whereas in the other classes they are discrete arguments in the constructor.

SokratisVidros (changes requested):

packages/js/src/notifications/notification.ts
LetItRock: @SokratisVidros We export just the Notification type, not a class, and callbacks receive instances, so they won't bother with the constructor.
The protected setters only work when you extend from the class, but it is still TS syntactic sugar that is not removed in JS after transpilation.

SokratisVidros (approved):

PR #6456, opened 2024-09-09
fix(web): form issues for new settings sidebar

ainouzgali:

Fixes the following issues:
- Upon changing the Workflow name in the sidebar, the new name only appears in the "main page" as the title after closing the sidebar
- Not showing errors on the setting input fields
- Clicking save after changing only preferences, the settings input fields are empty 
- No success/error message on saving 

### Screenshots

<details>

</details>

### DIFF
--- a/apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
+++ b/apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
@@ -1,9 +1,10 @@
 import { INotificationTemplate } from '@novu/shared';
-import { useEffect, useMemo } from 'react';
-import { useFormContext, SubmitHandler } from 'react-hook-form';
-import { useUpdateTemplate } from '../../../api/hooks/index';
+import { useFormContext, SubmitHandler, useWatch } from 'react-hook-form';
+import { useUpdateTemplate } from '../../../api/hooks';
 import { useUpdateWorkflowChannelPreferences } from '../../../hooks/workflowChannelPreferences/useUpdateWorkflowChannelPreferences';
 import { WorkflowDetailFormContext } from '../../../studio/components/workflows/preferences/WorkflowDetailFormContextProvider';
+import { errorMessage, successMessage } from '../../../utils/notifications';
+import { useEffectOnce } from '../../../hooks';
 
 type UseWorkflowDetailPageFormProps = {
   templateId: string;
@@ -11,10 +12,15 @@ type UseWorkflowDetailPageFormProps = {
 };
 
 export const useWorkflowDetailPageForm = ({ templateId, workflow }: UseWorkflowDetailPageFormProps) => {
-  const { formState, setValue, resetField, reset, handleSubmit, getValues, watch } =
-    useFormContext<WorkflowDetailFormContext>();
+  const {
+    formState: { dirtyFields, isValid },
+    resetField,
+    reset,
+    handleSubmit,
+    getValues,
+  } = useFormContext<WorkflowDetailFormContext>();
 
-  const workflowName = watch('general.name');
+  const workflowName = useWatch({ name: 'general.name' });
 
   const { updateWorkflowChannelPreferences, isLoading: isUpdatingPreferences } = useUpdateWorkflowChannelPreferences(
     templateId,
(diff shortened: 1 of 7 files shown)


apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
```diff
@@ -30,34 +36,44 @@ export const useWorkflowDetailPageForm = ({ templateId, workflow }: UseWorkflowD
...
-    }
+      if (dirtyFields?.general) {
+        const { workflowId, ...templateValues } = getValues('general');
+        await updateTemplateMutation({ id: templateId, data: { ...templateValues, identifier: workflowId } });
+      }
 
-    reset();
+      successMessage('Workflow updated successfully');
+
+      reset(data);
```
ainouzgali, line 54: reset with the form data, otherwise it sets to empty strings(the default values)

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
```diff
@@ -30,34 +36,44 @@ export const useWorkflowDetailPageForm = ({ templateId, workflow }: UseWorkflowD
...
         workflowId: workflow.triggers?.[0]?.identifier ?? '',
         name: workflow.name,
         description: workflow.description,
-      });
-    }
-  }, [workflow]);
+      },
+    };
+    reset(form);
+  }, !!workflow);
```
ainouzgali, line 71: For the initial values

apps/web/src/studio/components/workflows/preferences/WorkflowDetailFormContextProvider.tsx
```diff
@@ -26,14 +26,6 @@ export const WorkflowDetailFormContextProvider: FC<PropsWithChildren<IWorkflowDe
...
     mode: 'onChange',
     defaultValues: DEFAULT_FORM_VALUES,
-    values: {
-      general: {
-        workflowId: '',
-        name: '',
-        description: undefined,
-      },
-      preferences: DEFAULT_WORKFLOW_PREFERENCES,
-    },
```
ainouzgali, line 36: There is no need to set the values here, we set from the default values and then set the values from the workflow

apps/web/src/studio/components/workflows/preferences/WorkflowGeneralSettingsForm.tsx
ainouzgali, line 1: @antonjoel82 I wasn't sure why you wanted to keep this component [independent](https://github.com/novuhq/novu/pull/6447#discussion_r1749612611) of `react-hook-form`, but as using the controller on each field is less prone to unexpected behavior and it handles things for us I went for it.  Let me know what you think ❤️

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
```diff
@@ -30,34 +36,44 @@ export const useWorkflowDetailPageForm = ({ templateId, workflow }: UseWorkflowD
...
+      }
 
-    if (formState.dirtyFields?.general) {
-      const { workflowId, ...templateValues } = getValues('general');
-      updateTemplateMutation({ id: templateId, data: { ...templateValues, identifier: workflowId } });
-    }
+      if (dirtyFields?.general) {
+        const { workflowId, ...templateValues } = getValues('general');
+        await updateTemplateMutation({ id: templateId, data: { ...templateValues, identifier: workflowId } });
+      }
```
antonjoel82, line 50: 💭 **thought (non-blocking):** One thing that made this more difficult / will require some extra work, is that I don't believe either of these functions throw errors since it's handled within `react-query`. So we don't actually know if they've succeeded at this point (or if one did and one failed). However, I can handle this in separate changes‏

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
antonjoel82: 💭 **thought (non-blocking):** The reason this was intended to run every time `workflow` changes is in the chance the back-end does some sort of transformation on the data, we'd want that to be the source of truth rather than letting the form hold the state.‏

antonjoel82 (approved):
Thank you! 🙌

apps/web/src/pages/templates/editor_v2/useWorkflowDetailPageForm.ts
ainouzgali: @antonjoel82 I don't know about the preference one, but the update template does throw errors. Try saving with a workflowId that already exists

PR #6477, opened 2024-09-10
feat(framework): Change framework capitalization: in_app -> inApp

antonjoel82:

https://linear.app/novu/issue/NV-4283/fix-interface-bug-change-in-app-to-inapp-in-workflow

- Use casing consistent with `step.inApp` for Preferences keys
- Use `shared` type instead of `framework`


packages/framework/src/constants/workflow-options.constants.ts
```diff
@@ -0,0 +1,8 @@
+export enum WorkflowOptionsChannelEnum {
```
davidsoderberg: Can we use `ChannelTypeEnum` for the values that are the same?

antonjoel82: Enums can't refer to other enum values (by using computed values), so it would either be this enum with a single value or the way it is here -- with all keys in the framework setting using this new enum

packages/framework/src/constants/workflow-options.constants.ts
```diff
@@ -0,0 +1,8 @@
...
+  EMAIL = 'email',
+  SMS = 'sms',
+  PUSH = 'push',
+  CHAT = 'chat',
+  /** Differs from ChannelTypeEnum in capitalization / snake_case */
+  IN_APP = 'inApp',
+}
```
rifont: Nice addition 👏

Whilst we're adding this human-friendly enum, it will be great to also update the Step types to use the same enum for the key properties [here](https://github.com/novuhq/novu/blob/next/packages/framework/src/types/step.types.ts#L236). I suggest we make this change in this PR to create a concrete point of reference for when to use this enum.

Lastly, to reflect that this enum is used across both steps and options, I suggest we rename the following:
* filename: `workflow-options.constants.ts` -> `workflow.constants.ts`
* enum: `WorkflowOptionsChannelEnum` -> `WorkflowChannelEnum`

packages/framework/src/types/workflow.types.ts
```diff
@@ -40,9 +40,7 @@ export type WorkflowOptionChannelPreference = {
 
 export type WorkflowOptionsPreferences = {
   workflow?: WorkflowOptionChannelPreference;
-  channels?: {
-    [key in (typeof ChannelTypeEnum)[keyof typeof ChannelTypeEnum]]?: WorkflowOptionChannelPreference;
-  };
+  channels?: Partial<Record<WorkflowOptionsChannelEnum, WorkflowOptionChannelPreference>>;
```
rifont: 🚀

rifont (approved):
Awesome work! 🚅

packages/framework/src/constants/workflow-options.constants.ts
rifont: I actually prefer to keep them separate as they are now to make the distinction between the two more clear - one is machine readable, the other is human readable.

PR #6475, opened 2024-09-10
fix(api): Use _subscriberId

BiswaViraj:
### What changed? Why was the change needed?

- Fixed the usecase to use `_subscriberId` instead of `subscriberId`
### Screenshots

[COM-224](https://linear.app/novu/issue/COM-224/use-subscriberid-instead-of-subscriberid-in-preferences)
<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #6478, opened 2024-09-11
fix(api): subscriber id being wrong id

davidsoderberg:
Our subscriber id is used to fetch preferences and not customers subscriber id.

### DIFF
--- a/apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
+++ b/apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
@@ -145,7 +145,7 @@ export class UpdatePreferences {
         channels: preference.channels,
         organizationId: command.organizationId,
         environmentId: command.environmentId,
-        subscriberId: command.subscriberId,
+        subscriberId: subscriber._id,
         templateId: workflow._id,
       });
 
@@ -176,7 +176,7 @@ export class UpdatePreferences {
       channels: preference.channels,
       organizationId: command.organizationId,
       environmentId: command.environmentId,
-      subscriberId: command.subscriberId,
+      subscriberId: subscriber._id,
     });
 
     return {
--- a/libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
+++ b/libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
@@ -67,6 +67,11 @@ export class GetPreferences {
       .filter((preference) => preference !== undefined)
       .map((item) => item.preferences);
 
+    // ensure we don't merge on an empty list
+    if (preferences.length === 0) {
+      throw new NotFoundException('We could not find any preferences');
+    }
+
     return deepMerge(preferences);
   }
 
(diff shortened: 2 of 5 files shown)


apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -145,7 +145,7 @@ export class UpdatePreferences {
         channels: preference.channels,
         organizationId: command.organizationId,
         environmentId: command.environmentId,
-        subscriberId: command.subscriberId,
+        subscriberId: subscriber._id,
```
antonjoel82, line 148: **concern 👀**: We're not using this same `subscriber._id` in the `getSubscriberTemplatePreferenceUsecase` call. Why doesn't the command's `subscriberId` work?

antonjoel82 (commented):
I will test now -- but confused why the `subscriber` ID is different than the the `command.subscriberId`

SokratisVidros (approved):

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
davidsoderberg: Not sure I follow your concern here is that we do not use command.subscriberId? The commands subscriber id is the customer specified id not our mongo object id that is needed for queries in our database.

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -67,6 +67,11 @@ export class GetPreferences {
       .filter((preference) => preference !== undefined)
       .map((item) => item.preferences);
 
+    // ensure we don't merge on an empty list
+    if (preferences.length === 0) {
+      throw new NotFoundException('We could not find any preferences');
+    }
```
antonjoel82, line 73: @davidsoderberg This prevents returning [] accidentally from this method

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -111,7 +116,7 @@ export class GetPreferences {
     return items.find(
       (item) =>
         item.type === PreferencesTypeEnum.SUBSCRIBER_WORKFLOW &&
-        item._templateId == templateId,
+        item._templateId === templateId,
```
antonjoel82, line 119: 🗒 note (non-blocking): Typo fix‏

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -176,7 +176,7 @@ export class UpdatePreferences {
       channels: preference.channels,
       organizationId: command.organizationId,
       environmentId: command.environmentId,
-      subscriberId: command.subscriberId,
+      subscriberId: subscriber._id,
```
BiswaViraj, line 179: I have made a similar change in my [PR here](https://github.com/novuhq/novu/pull/6475)
Can you please take a look at that

BiswaViraj (approved):

PR #6480, opened 2024-09-11
feat(api): store Stripe customer ids locally

ChmaraX:
Detailed description in related EE PR: https://github.com/novuhq/packages-enterprise/pull/216


libs/dal/src/repositories/organization/organization.entity.ts
```diff
@@ -28,6 +28,8 @@ export class OrganizationEntity implements IOrganizationEntity {
 
   externalId?: string;
 
+  stripeCustomerId?: string;
+
```
ChmaraX, line 32: This is not ideal - we should probably think of a way to split enterprise and community entities+schema so its clear which attributes are available using EE and community repositories, rather than marking them optional.

Will create an action item to address this.

SokratisVidros: Great call out! I agree 100%.

SokratisVidros (approved):

PR #6491, opened 2024-09-12
feat(api,application-generic): Handle readOnly for preferences

antonjoel82:

https://linear.app/novu/issue/NV-4295/implement-inbox-changes-for-readonly-property-handling

- Surface the `readOnly` properties of channels through backwards-compatible change using `critical` field.
  - **Note:** for now, a workflow with new preferences is considered 'critical' if its top-level `workflow.readOnly` is true, or if any of the channels are true. This is because we lack the granularity in our response to disable switches individually.
- When "merging" preferences, `defaultValue` (enablement) and `readOnly` need to be handled in different orders because different scenarios take precedence. I.e. when determining if a workflow should be enabled, the subscriber's selection has the topmost priority; however, for `readOnly`, the preference set in the Dashboard should be the highest priority.

### Screenshots

Screenshot 2024-09-12 at 12 05 43 AM

### DIFF
--- a/libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
+++ b/libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
@@ -1,36 +1,23 @@
 import { Injectable, NotFoundException } from '@nestjs/common';
-import {
-  PreferencesActorEnum,
-  PreferencesEntity,
-  PreferencesRepository,
-  PreferencesTypeEnum,
-} from '@novu/dal';
-import {
-  FeatureFlagsKeysEnum,
-  IPreferenceChannels,
-  WorkflowChannelPreferences,
-} from '@novu/shared';
+import { PreferencesActorEnum, PreferencesEntity, PreferencesRepository, PreferencesTypeEnum } from '@novu/dal';
+import { WorkflowOptionsPreferences } from '@novu/framework';
+import { FeatureFlagsKeysEnum, IPreferenceChannels, WorkflowChannelPreferences } from '@novu/shared';
 import { deepMerge } from '../../utils';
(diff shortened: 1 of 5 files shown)


libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -112,33 +76,108 @@ export class GetPreferences {
...
+      subscriberGlobalPreferences,
+      subscriberWorkflowPreferences,
+    ]
+      .filter((preference) => preference !== undefined)
+      .map((item) => item.preferences);
+
+    // ensure we don't merge on an empty list
+    if (preferences.length === 0) {
+      return;
+    }
```
antonjoel82: 🗒 **note (non-blocking):** This is existing functionality just moved‏

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -112,33 +76,108 @@ export class GetPreferences {
...
+        in_app: { readOnly: channels['in_app'].readOnly },
+        email: { readOnly: channels['email'].readOnly },
+        sms: { readOnly: channels['sms'].readOnly },
+        chat: { readOnly: channels['chat'].readOnly },
+        push: { readOnly: channels['push'].readOnly },
+      },
+    })) as WorkflowChannelPreferences[];
+
+    // by merging only the read-only values after the full objects, we ensure that only the readOnly field is affected.
+    return deepMerge([...preferences, ...readOnlyPreferences]);
```
antonjoel82: 🗒 **note (non-blocking):** This is new functionality to merge `readOnly` preferences in proper order‏

rifont (approved):

PR #6493, opened 2024-09-12
fix(api): remove environment id from get integration

jainpawan21:

When user is in development environment and go to production sendgrid integration, webhook URL is not showing because at API level, _environmentId is still development one and integration is not found in development env. Above case can be reproducesd for development integrations in the production environment.
### Screenshots

<details>

</details>


SokratisVidros (approved):

SokratisVidros (changes requested):
Does the environment come from the Novu-Environment-Id http header? Since we have separate integrations per environment, shouldn't the environment be set regardless but not from the header?

jainpawan21:
> Does the environment come from the Novu-Environment-Id http header? Since we have separate integrations per environment, shouldn't the environment be set regardless but not from the header?

environmentId comes with user session. Production integrations can be accessed from the development environment.

https://github.com/user-attachments/assets/9420674a-639d-48d8-a498-b766fa25bb4d

jainpawan21:
@SokratisVidros 

Could you please review this PR again?

jainpawan21:
@SokratisVidros 

bumping this 👆🏻

scopsy (approved):

PR #6494, opened 2024-09-12
feat(root): Add `@novu/themes` package

scopsy:

- Add a new @novu/themes package
- Notion Theme
- Default Theme
- Added a themes to the existing playground
### Screenshots

<details>

</details>


packages/themes/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "react-dom": "^18.3.1",
+    "tsup": "^8.2.1",
+    "typescript": "4.9.5"
+  },
+  "peerDependencies": {
+    "react": ">=17",
+    "react-dom": ">=17"
+  },
+  "dependencies": {
+    "@novu/js": "workspace:*"
```
rifont: Should `@novu/js` be a peer dependency instead? I expect you would be using this package in conjunction with `@novu/js` and it's downstream packages like `@novu/react`

rifont: On second thought, making `@novu/js` a peer dependency only would necessitate that the consumer also installs `@novu/js`, which isn't necessary if you're using `@novu/react`.

I think instead we should make `@novu/js` a `peerDependencies` and specify [`peerDependenciesMeta.<package>.optional: true`](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#peerdependenciesmeta) to ensure that those peer dependencies aren't automatically installed by package managers.

scopsy: I like this approach! Will add it

SokratisVidros (approved):

PR #6498, opened 2024-09-12
feat(api): add option to remove Novu branding in the inbox

ChmaraX:
Added option to remove Novu branding from Inbox component for paid customers. 

- Only non-free tiers are able to update this setting
- After customer is set back to `free`, this setting is set back to `false`
- Value will be always `false` unless explicitly set to `true` on integration entity
- This setting is exposed for Inbox component inside session
- This feature is under a feature flag `IS_IMPROVED_BILLING_ENABLED`

**Related PRs:**
https://github.com/novuhq/novu/pull/6500
https://github.com/novuhq/packages-enterprise/pull/218

### DIFF
--- a/apps/api/src/app/inbox/usecases/session/session.usecase.ts
+++ b/apps/api/src/app/inbox/usecases/session/session.usecase.ts
@@ -86,9 +86,12 @@ export class Session {
 
     const token = await this.authService.getSubscriberWidgetToken(subscriber);
 
+    const removeNovuBranding = inAppIntegration.removeNovuBranding || false;
+
     return {
       token,
       totalUnreadCount,
+      removeNovuBranding,
     };
   }
 }
--- a/apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
+++ b/apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
@@ -97,16 +97,39 @@ export class UpdateIntegration {
     return result;
   }
 
-  private async setRemoveNovuBranding(command: UpdateIntegrationCommand): Promise<boolean> {
-    const organization = await this.communityOrganizationRepository.findOne({
-      _id: command.organizationId,
-    });
+  private async shouldUpdateRemoveNovuBranding(
+    command: UpdateIntegrationCommand,
+    existingIntegration: IntegrationEntity
+  ): Promise<boolean> {
+    const [isImprovedBillingEnabled, organization] = await Promise.all([
+      this.getFeatureFlag.execute(
+        GetFeatureFlagCommand.create({
+          userId: 'system',
+          environmentId: 'system',
+          organizationId: command.organizationId,
+          key: FeatureFlagsKeysEnum.IS_IMPROVED_BILLING_ENABLED,
+        })
+      ),
+      this.communityOrganizationRepository.findOne({ _id: command.organizationId }),
+    ]);
 
-    if (organization && organization.apiServiceLevel !== ApiServiceLevelEnum.FREE) {
-      return command.removeNovuBranding ?? false;
+    if (!isImprovedBillingEnabled) {
+      return false;
     }
 
-    return false;
+    const isRemoveNovuBrandingDefined = typeof command.removeNovuBranding !== 'undefined';
+    const isRemoveNovuBrandingChanged =
+      isRemoveNovuBrandingDefined && existingIntegration.removeNovuBranding !== command.removeNovuBranding;
+
+    if (!isRemoveNovuBrandingChanged) {
+      return false;
+    }
+
+    if (!organization || organization.apiServiceLevel === ApiServiceLevelEnum.FREE) {
+      return false;
+    }
+
+    return true;
   }
 
   async execute(command: UpdateIntegrationCommand): Promise<IntegrationEntity> {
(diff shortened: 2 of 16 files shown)


libs/dal/src/repositories/organization/community.organization.repository.ts
```diff
@@ -63,6 +65,10 @@ export class CommunityOrganizationRepository
   }
 
   async updateServiceLevel(organizationId: string, apiServiceLevel: ApiServiceLevelEnum) {
+    if (apiServiceLevel === ApiServiceLevelEnum.FREE) {
+      await this.integrationRepository.setRemoveNovuBranding(organizationId, false);
+    }
+
```
ChmaraX, line 71: This needs to be tied closely to `updateServiceLevel` since it needs to be always changed together

apps/api/src/app/inbox/usecases/session/session.usecase.ts
```diff
@@ -86,9 +86,12 @@ export class Session {
 
     const token = await this.authService.getSubscriberWidgetToken(subscriber);
 
+    const removeNovuBranding = inAppIntegration.removeNovuBranding || false;
```
ChmaraX, line 89: It will be always `false` unless explicitely specified as `true` on integration entity in db - no migration required

SokratisVidros: Lovely!

SokratisVidros (approved):

apps/api/src/app/inbox/usecases/session/session.usecase.ts
LetItRock: ❓ **question:** shouldn't this be set on the db level? I mean in the schema you can provide the default value, like:
```
removeNovuBranding: {
  type: Schema.Types.Boolean,
  default: false,
},
```
Otherwise, if we use it in different places, we would have to remember to do `inAppIntegration.removeNovuBranding || false`

LetItRock: But yeah, this will require migration with setting to false for everyone.

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -185,6 +198,25 @@ export class UpdateIntegration {
...
+          environmentId: 'system',
+          organizationId: command.organizationId,
+          key: FeatureFlagsKeysEnum.IS_IMPROVED_BILLING_ENABLED,
+        })
+      );
+
+      if (isImprovedBillingEnabled) {
+        updatePayload.removeNovuBranding = await this.setRemoveNovuBranding(command);
+      }
+    }
```
LetItRock: ✏ **suggestion:** can you please cover this code with e2e tests?

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
```diff
@@ -96,6 +97,18 @@ export class UpdateIntegration {
...
   }
 
+  private async setRemoveNovuBranding(command: UpdateIntegrationCommand): Promise<boolean> {
+    const organization = await this.communityOrganizationRepository.findOne({
+      _id: command.organizationId,
+    });
+
+    if (organization && organization.apiServiceLevel !== ApiServiceLevelEnum.FREE) {
+      return command.removeNovuBranding ?? false;
+    }
```
LetItRock: ❓ **question:** I'm wondering whether we shouldn't throw an error if the org is on free trier and somebody tries to update the value to `true` 🤔 Because otherwise, the operation will be successful.

apps/api/src/app/inbox/usecases/session/session.usecase.ts
ChmaraX: @LetItRock That's what I initially did, but then I thought this way it might be a little bit better:

Given the `default: false` value on schema, we would end up in this state:
- `removeNovuBranding` would be on every newly created integration entity (not just on `type: in-app` ones) (the default is automatically set upon creation if im not mistaken)

Without the default value, we are I guess more following the no-schema MongoDB approach, compared to the previous case which is more schema-like:
- we have the `removeNovuBranding` set **only** on `in-app` integration entities, where user interacted with this specific setting, so we are not storing something that not necessary for the other ones

 > Otherwise, if we use it in different places, we would have to remember to do inAppIntegration.removeNovuBranding || false

I think in both cases this is implicit without remembering this check because if you do:
```typescript
if (integration.removeNovuBranding) {
...
}
```
And the property doesn't exist on the entity, it would result in `false`, which follows the idea here that its `true` only if its explicitly set as `true`. 

This line `inAppIntegration.removeNovuBranding || false` is not required, I put it there as a safeguard in case we would somehow manually change the type in database to "string", since Inbox component is quite important part of the app I wanted to make sure its always `false` unless `true`

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
ChmaraX: Good point, now this method would return `false` and proceed to set it as `false` in db, ideally we don't want to set this property at all. I will try to implement some check if update is needed.

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
ChmaraX: Done ✅

apps/api/src/app/integrations/usecases/update-integration/update-integration.usecase.ts
ChmaraX: Changed:
- encapsulated the entire related logic in the method
- now the value is updated in DB only if necessary

apps/api/src/app/inbox/usecases/session/session.usecase.ts
LetItRock: @ChmaraX, thanks for the detailed explanation, I see what you are saying. FYI this flag might be used in the future for the other channels as well, for example, there were plans to introduce some branding to the emails sent with Novu integration. But I think it's ok to proceed with this approach for now.

LetItRock (approved):

PR #6501, opened 2024-09-13
chore(root): Framework v2.0.3

SokratisVidros:
Latest framework release plus a script to help us restore the workspace:* protocol


scripts/restore-workspace-protocol.mjs
```diff
@@ -0,0 +1,38 @@
+import fs from 'fs-extra';
+import path from 'node:path';
+import { glob } from 'glob';
+import {execa} from 'execa';
```
rifont: Small whitespace fix

```suggestion
import { execa } from 'execa';
```

scripts/restore-workspace-protocol.mjs
```diff
@@ -0,0 +1,38 @@
...
+
+  updateNovuDependencies(packageJson.dependencies);
+  updateNovuDependencies(packageJson.devDependencies);
+
+  fs.writeJsonSync(filePath, packageJson, { spaces: 2 });
+  console.log(`Restored workspace protocol at ${filePath}`);
+}
+
+// Find all package.json files in the repo
+const files = await glob('packages/**/package.json', { ignore: 'packages/**/node_modules/**' });
```
rifont, line 37: Is it intentional to target `packages` directory only?

scripts/restore-workspace-protocol.mjs
```diff
@@ -0,0 +1,38 @@
...
+      dependencies[key] = 'workspace:*';
+    }
+  }
+}
+
+function processPackageJson(filePath) {
+  const packageJson = fs.readJsonSync(filePath);
+
+  updateNovuDependencies(packageJson.dependencies);
+  updateNovuDependencies(packageJson.devDependencies);
```
rifont, line 28: we need to cover `peerDependencies` and `optionalDependencies` too, as some packages use them ([ref](https://github.com/novuhq/novu/blob/next/enterprise/packages/billing/package.json#L52))

scripts/restore-workspace-protocol.mjs
```diff
@@ -0,0 +1,38 @@
+import fs from 'fs-extra';
+import path from 'node:path';
+import { glob } from 'glob';
+import {execa} from 'execa';
+
+// The following node packages don't belong to the monorepo and must always be references with specific semver.
+const IGNORE_LIST = ["@novu/ntfr-client", "@novu/floating-vue"];
```
antonjoel82:
```suggestion
const IGNORE_LIST = new Set(["@novu/ntfr-client", "@novu/floating-vue"]);
```

scripts/restore-workspace-protocol.mjs
```diff
@@ -0,0 +1,38 @@
...
+async function getPackageNames() {
+  const { stdout } = await execa`pnpm nx show projects`;
+  return stdout;
+}
+
+
+// Function to update versions of @novu dependencies
+function updateNovuDependencies(dependencies) {
+  for (const [key] of Object.entries(dependencies || {})) {
+    if (key.startsWith('@novu/') && IGNORE_LIST.indexOf(key) === -1) {
```
antonjoel82:
```suggestion
    if (key.startsWith('@novu/') && IGNORE_LIST.has(key)) {
```

scripts/restore-workspace-protocol.mjs
SokratisVidros: For now yes, as they are the only ones that we change the workspace during publishing.

scripts/restore-workspace-protocol.mjs
SokratisVidros: Fixed

scripts/restore-workspace-protocol.mjs
SokratisVidros: Good call.

scripts/restore-workspace-protocol.mjs
SokratisVidros: Fixed

LetItRock (approved):

PR #6503, opened 2024-09-13
feat(shared, web, application-generic): Create util for building preferences

antonjoel82:

https://linear.app/novu/issue/NV-4297/create-shared-utility-for-creating-default-preferences-and-use

- Implement and test shared utility for building preferences from any incomplete preference object
- Use util to centralize defaulting logic


packages/framework/src/resources/workflow/index.ts
```diff
@@ -81,6 +82,9 @@ export function workflow<
     };
   };
 
+  // TODO: delegate preference building to API later on. Must keep the mapping layer here though
+  const preferences = buildWorkflowChannelPreferences(mapPreferences(options.preferences));
```
antonjoel82: 🗒 **note (non-blocking):** As discussed with @rifont, we will later delegate any / all building of preferences to the API by removing this `buildWorkflowChannelPreferences`. This will require type changes in `framework` that will subsequently change types in `api` as well, so this will come at a later time‏

apps/api/src/app/preferences/preferences.controller.ts
```diff
@@ -57,7 +57,7 @@ export class PreferencesController {
         environmentId: user.environmentId,
         organizationId: user.organizationId,
         userId: user._id,
-        preferences: data.preferences,
+        preferences: buildWorkflowChannelPreferences(data.preferences),
```
antonjoel82: ❓ **question:** Is this an okay place to put this?

The issue is that this flows into a usecase in `application-generic` (which doesn't have `shared` as a dependency), so we can't use the centralized utility for building preferences after this point...‏

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.spec.ts
```diff
@@ -64,14 +66,16 @@ describe('UpdatePreferences', () => {
...
 
     updatePreferences = new UpdatePreferences(
       subscriberPreferenceRepositoryMock as any,
       notificationTemplateRepositoryMock as any,
       subscriberRepositoryMock as any,
       analyticsServiceMock as any,
       getSubscriberGlobalPreferenceMock as any,
-      getSubscriberTemplatePreferenceUsecase as any
+      getSubscriberTemplatePreferenceUsecase as any,
+      upsertPreferenceMock as any
```
antonjoel82: 🗒 **note (non-blocking):** Unrelated to my changes but my editor was (rightfully) complaining about an incorrect number of dependencies here‏

apps/api/src/app/preferences/preferences.controller.ts
rifont: `@novu/shared` is in fact a dependency of `@novu/application-generic`, we can move the builder into the `upsert-preferences` use-case. I'll update it.

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -215,7 +215,7 @@ export class Sync {
           })
         );
 
-        if (isWorkflowPreferencesEnabled && workflow.preferences) {
+        if (isWorkflowPreferencesEnabled) {
```
rifont, line 218: Undefined preferences are handled by the use-case now.

apps/api/src/app/events/e2e/bridge-sync.e2e.ts
```diff
@@ -325,12 +325,12 @@ describe('Bridge Sync - /bridge/sync (POST)', async () => {
       {
         preferences: {
           workflow: {
-            defaultValue: false,
+            enabled: false,
```
rifont, line 328: `defaultValue` was uncomfortable, especially when it comes to the Subscriber making changes to the `defaultValue` - it's better renamed to `enabled` to dictate the actual outcome, rather than a relation to the resources modifying it.

rifont (approved):
Awesome work!

PR #6514, opened 2024-09-17
fix(api): add clean v2 version

djabarovgeorge:
Instead of using the prefix as the version, i refactored the [nest js native versioning by URI feature](https://docs.nestjs.com/techniques/versioning#uri-versioning-type) 

### Screenshots

<details>

</details>


apps/api/src/bootstrap.ts
```diff
@@ -72,6 +72,12 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
     app = await NestFactory.create(AppModule, { bufferLogs: true, ...nestOptions });
   }
 
+  app.setGlobalPrefix(`${CONTEXT_PATH}`);
```
djabarovgeorge: I don’t have any information on the CONTEXT_PATH use case, and I’m wondering if this change could potentially be a breaking one. Based on my understanding of the code, previously, users would have had to manually add an extra forward slash if needed. Now, it’s being added automatically after the CONTEXT_PATH and before the version. 

For example:
- Previously if `prefix = ${CONTEXT_PATH}v1` and `CONTEXT_PATH=backend/`, the result would be `backend/v1/api/<controller>`.  
- In the new implementation, the result will be `backend//v1/<controller>`.

Another example:
- Previously if `prefix = ${CONTEXT_PATH}v1` and `CONTEXT_PATH=backend-`, the result would be `backend-v1/api/<controller>`.
- In the new implementation, the result will be `backend-/v1/api/<controller>`. 

Unless I am missing something this could cause issues in certain cases.

djabarovgeorge: solved this issue with prefix option in enableVersioning 👏

djabarovgeorge:
this work will be merged with `nv-4282-post-v2-workflows` branch

PR #6519, opened 2024-09-18
feat(js,react): inbox allow filtering preferences by tags

LetItRock:
Inbox - allow filtering preferences view by provided tags through the props.
Updated:
- Inbox props to allow passing `preferencesFilter` prop
- Deprecated Inbox prop `tabs.value` suggested to use `tabs.filter` instead
- Updated `usePreferences` hook to allow filtering
- Updated headless JS SDK, InboxService, cache

### Screenshots
Screenshot 2024-09-18 at 09 14 46

https://github.com/user-attachments/assets/3ad2c432-a108-4439-b009-bc05278dfbdd


packages/js/src/api/inbox-service.ts
```diff
@@ -141,8 +141,15 @@ export class InboxService {
...
-    return this.#httpClient.get(`${INBOX_ROUTE}/preferences`);
+  fetchPreferences(tags?: string[]): Promise<PreferencesResponse[]> {
+    const queryParams = new URLSearchParams();
+    if (tags) {
+      tags.forEach((tag) => queryParams.append('tags[]', tag));
+    }
+
+    const query = queryParams.size ? `?${queryParams.toString()}` : '';
+
+    return this.#httpClient.get(`${INBOX_ROUTE}/preferences${query}`);
```
LetItRock, line 152: Updated the InboxService to send the `tags` to the `GET /inbox/preferences` endpoint.

packages/js/src/cache/preferences-cache.ts
```diff
@@ -62,12 +74,18 @@ export class PreferencesCache {
...
-  getAll(): Preference[] | undefined {
-    return this.#cache.get(DEFAULT_KEY);
+  set(args: ListPreferencesArgs, data: Preference[]): void {
+    this.#cache.set(getCacheKey(args), data);
+  }
+
+  getAll(args: ListPreferencesArgs): Preference[] | undefined {
+    if (this.has(args)) {
+      return this.#cache.get(getCacheKey(args));
+    }
```
LetItRock, line 90: The preferences cache now stores key pairs like: `tags <--> preferences`

packages/js/src/preferences/preferences.ts
```diff
@@ -31,14 +31,14 @@ export class Preferences extends BaseModule {
     this.#useCache = useCache;
   }
 
-  async list(): Result<Preference[]> {
+  async list(args: ListPreferencesArgs = {}): Result<Preference[]> {
```
LetItRock, line 34: Updated the headless JS SDK function to get preferences

packages/js/src/ui/components/InboxTabs/InboxTab.tsx
```diff
@@ -26,7 +26,7 @@ export const InboxTabUnreadNotificationsCount = (props: { count: number }) => {
 export const InboxTab = (props: Tab & { class?: ClassName }) => {
   const { status } = useInboxContext();
   const style = useStyle();
-  const unreadCount = useUnreadCount({ filter: { tags: props.value } });
+  const unreadCount = useUnreadCount({ filter: { tags: getTagsFromTab(props) } });
```
LetItRock, line 29: Because we have deprecated the `value` property on the tab object and introduced the `filter` property instead, I created this helper function that takes the tags value from the new prop or fallbacks to the old. This function is used around the code.

packages/react/src/hooks/usePreferences.ts
```diff
@@ -3,6 +3,7 @@ import { useEffect, useState } from 'react';
 import { useNovu } from './NovuProvider';
 
 type UsePreferencesProps = {
+  filter?: { tags?: string[] };
```
LetItRock, line 6: `usePreferences` hook takes a filter prop.

packages/js/src/cache/preferences-cache.ts
```diff
@@ -3,11 +3,23 @@ import type { Cache } from './types';
...
 import { Preference } from '../preferences/preference';
+import { ListPreferencesArgs } from '../preferences/types';
 
 // these events should update the preferences in the cache
 const updateEvents: PreferenceEvents[] = ['preference.update.pending', 'preference.update.resolved'];
 
-const DEFAULT_KEY = 'default';
+const excludeEmpty = ({ tags }: ListPreferencesArgs) =>
+  Object.entries({ tags })
+    .filter(([_, value]) => value !== null && value !== undefined && !(Array.isArray(value) && value.length === 0))
```
SokratisVidros: 🤓 We can combine filter in the reduce function and save some performance from one less iteration.

packages/js/src/ui/components/InboxTabs/InboxTab.tsx
SokratisVidros: Please add this as a comment in the code as in a future major or minor version we can remove the code and remove the deprecation.

SokratisVidros (approved):
Very nicely done!

BiswaViraj (approved):

packages/js/src/cache/preferences-cache.ts
LetItRock: the object has only "one entry" - `tags`, so in this case, it will be always two iterations, one for the filter, and one for the reduce 🤓 but we can reduce it to one 😂

packages/react/src/hooks/usePreferences.ts
```diff
@@ -46,7 +47,7 @@ export const usePreferences = (props?: UsePreferencesProps): UsePreferencesResul
 
   const fetchPreferences = async () => {
     setIsFetching(true);
-    const response = await preferences.list();
+    const response = await preferences.list(props?.filter);
```
desiprisg, line 50: Does this need a default? Will the result be the same with `undefined` here?

PR #6520, opened 2024-09-18
chore(root): Upgrade to latest typescript

SokratisVidros:
Latest TS, cause why not?


apps/web/package.json
```diff
@@ -135,7 +135,7 @@
     "rimraf": "^3.0.2",
     "slugify": "^1.4.6",
     "storybook-dark-mode": "^3.0.1",
-    "typescript": "4.9.5",
+    "typescript": "5.6.2",
```
LetItRock: duplicate

package.json
```diff
@@ -176,7 +176,7 @@
     "tar": "^6.1.11",
     "ts-jest": "27.1.5",
     "ts-node": "~10.9.1",
-    "typescript": "4.9.5",
+    "typescript": "5.6.2",
```
LetItRock, line 179: Should we consider removing it from here? This is a guard that forces using the same TS version across monorepo.

LetItRock (approved):
banger! 💥

tsconfig.base.json
rifont, line 1: 👏 **praise:** Yes! 🧹‏

packages/js/src/cache/notifications-cache.ts
```diff
@@ -9,6 +9,7 @@ const excludeEmpty = ({ tags, read, archived, limit, offset, after }: ListNotifi
   Object.entries({ tags, read, archived, limit, offset, after })
     .filter(([_, value]) => value !== null && value !== undefined && !(Array.isArray(value) && value.length === 0))
     .reduce((acc, [key, value]) => {
+      // @ts-expect-error
```
rifont, line 12: ☑ **chore:** Please add a reason for the expected error‏

packages/framework/src/servers/nuxt.ts
```diff
@@ -12,7 +12,6 @@ export const serve = (options: ServeHandlerOptions) => {
     /*
      * TODO: Fix this
      */
-    // @ts-expect-error - Request type is not correct
```
rifont, line 15: 👏 **praise:** Nice!‏

desiprisg (approved):

package.json
rifont: I think we need it here because without it, the top-level `tsconfig.json` won't know what configuration options are avaialable. Is that right @SokratisVidros ?

libs/novui/src/json-schema-components/JsonSchemaForm.tsx
```diff
@@ -73,18 +73,22 @@ export function JsonSchemaForm<TFormData = any>(props: JsonSchemaFormProps<TForm
...
       autoComplete={'false'}
       formContext={{ variables }}
       idSeparator={JSON_SCHEMA_FORM_ID_DELIMITER}
       templates={{
+        // @ts-expect-error
         ArrayFieldTitleTemplate,
         ArrayFieldTemplate,
         ArrayFieldItemTemplate,
         ObjectFieldTemplate,
+        // @ts-expect-error
```
rifont, line 91: ☑ **chore:** Please add a reason for the expected errors‏

package.json
LetItRock: @rifont I tried that in the other PR and it was fine 🤷‍♂️

libs/design-system/src/ThemeProvider.tsx
```diff
@@ -56,10 +56,6 @@ const notificationStyles = (theme: MantineTheme) => {
   };
 };
 
-declare module '@mantine/core' {
-  export type MantineColor = MantineColor | 'gradient';
-}
-
```
rifont, line 62: ❓ **question:** Why did we need to remove this? I think this reduces type safety when using the `color`/`c` prop on Mantine components‏

package.json
SokratisVidros: @rifont you are correct. If we remove it from here VS code gets a bit confused. The other thing I tried is to use `typescript: "*"` and then just set a fixed TS at the top level but I read that it's not advisable for the publishable packages as they end up in NPM. TBH I don't get it as it's a dev dependency but I didn't delve deeper.

libs/design-system/src/ThemeProvider.tsx
SokratisVidros: TS was complaining as it's a recursive type definition. MantineColor wasn't imported anywhere in this file.

PR #6522, opened 2024-09-18
fix(web): Inject latest Clerk JWT via Axios interceptor

SokratisVidros:

Fetching the Clerk JWT should always be done via the Clerk SDK, not the __session cookie that acts as a cache and creates the risk of getting an outdated JWT.

### Screenshots
image (1)


apps/web/src/api/api.client.ts
```diff
@@ -78,46 +59,42 @@ function buildUrl(url: string, absoluteUrl: boolean) {
...
-        'Novu-Environment-Id': lastEnvironmentId || '',
-      }
-    : {};
-}
-
 // WIP: The static API client needs to be replaced by a dynamic API client where api keys are injected.
 export function buildApiHttpClient({
   baseURL = API_ROOT || 'https://api.novu.co',
   secretKey,
   jwt,
```
ChmaraX: is this `jwt` param needed here? I don't see it being used in the function

ChmaraX (approved):

apps/web/src/api/api.client.ts
SokratisVidros: Good catch. I removed it.

apps/web/src/components/providers/AuthProvider.tsx
```diff
@@ -60,14 +58,10 @@ export const useAuth = () => {
...
-  const token = IS_EE_AUTH_ENABLED ? eeAuthTokenCookie.get() : getCommunityAuthToken();
-
-  return token || '';
-}
-
-export function getTokenClaims(): IJwtClaims | null {
-  const token = getToken();
+export async function getToken() {
+  if (IS_EE_AUTH_ENABLED) {
+    return (await window?.Clerk?.session?.getToken()) || '';
```
LetItRock, line 63: Somehow, for me, in my PR `window. Clerk` was a little buggy with probably race-conditons. But when used without the `window` object, it was working fine.

SokratisVidros: Weird, I guess it has to do with React renderings.

PR #6527, opened 2024-09-18
feat(api, worker, application-generic, dal): Resolve preferences for stateless workflow executions

rifont:

* Resolve preferences for stateless workflow executions
  * This ensures a consistent experience for workflow execution in both Studio and Dashboard, with the exception of Subscriber Workflow preferences which are reserved for use on stateful executions only.

### Screenshots

<details>

</details>


apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -110,7 +111,7 @@ export class RunJob {
       if (job.step.shouldStopOnFail || this.shouldBackoff(error)) {
         shouldQueueNextJob = false;
       }
-      throw new PlatformException(error.message);
+      throw error;
```
rifont, line 114: This change ensures the full error stacktrace can bubble up without obfuscating the thrower.

apps/worker/package.json
```diff
@@ -13,7 +13,7 @@
     "docker:build:depot": "pnpm --silent --workspace-root pnpm-context -- apps/worker/Dockerfile | depot build --build-arg PACKAGE_PATH=apps/worker - -t novu-worker --load",
     "start": "pnpm start:dev",
     "start:dev": "nest start --watch",
-    "start:test": "cross-env NODE_ENV=test nest start",
+    "start:test": "cross-env IS_WORKFLOW_PREFERENCES_ENABLED=true NODE_ENV=test nest start",
```
rifont: This is necessary because Worker doesn't currently listen to LaunchDarkly for Feature Flags with `NODE_ENV=test`. It will be removed in a subsequent feature-flag tidy-up PR.

apps/worker/src/app/workflow/usecases/send-message/send-message.command.ts
```diff
@@ -50,4 +51,7 @@ export class SendMessageCommand extends EnvironmentWithUserCommand {
 
   @IsDefined()
   tags: string[];
+
+  @IsOptional()
+  statelessPreferences?: WorkflowChannelPreferences;
```
rifont, line 56: I have opted to take a low-touch approach to adding support for stateless preferences, we will need to refactor the send-message flow at a later point to inject the NotificationTemplate into the SendMessage use-case and downstream use-cases.

This naming reflects this low-touch approach to highlight that stateless Workflows are dealt with separately in the flow.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -96,7 +97,7 @@ export class SendMessage {
     const stepType = command.step?.template?.type;
 
     let bridgeResponse: ExecuteOutput | null = null;
-    if (![StepTypeEnum.DIGEST, StepTypeEnum.DELAY, StepTypeEnum.TRIGGER].includes(stepType as any)) {
+    if (![StepTypeEnum.DIGEST, StepTypeEnum.DELAY, StepTypeEnum.TRIGGER].includes(stepType as StepTypeEnum)) {
```
rifont, line 100: Type fix.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -206,34 +207,19 @@ export class SendMessage {
     command: SendMessageCommand,
     variables: IFilterVariables
   ): Promise<{ filterResult: IConditionsFilterResponse | null; channelPreferenceResult: boolean | null }> {
-    const skipHalt = this.shouldSkipHalt(bridgeSkip, command.job?.step?.bridgeUrl);
-    if (skipHalt) {
+    if (bridgeSkip === true) {
```
rifont, line 202: We now only skip filter resolution when the Bridge returns `skip: true`, allowing for Global Subscriber and Workflow level Preferences to be evaluated for both stateful _and_ stateless workflows.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -206,34 +207,19 @@ export class SendMessage {
...
       return { filterResult: { passed: true, conditions: [], variables: {} }, channelPreferenceResult: true };
     }
 
     const [filterResult, channelPreferenceResult] = await Promise.all([
-      this.filter(command, variables),
-      this.filterPreferredChannels(command.job),
+      this.filterConditions(command, variables),
```
rifont: Renamed `this.filter` -> `this.filterConditions` to create a distinction between the two types of filters evaluated. One is Step conditions, the other is Delivery conditions.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -206,34 +207,19 @@ export class SendMessage {
...
-   * - If `bridgeSkip` is true (highest priority), skips all checks.
-   * - If `bridgeUrl` is provided, skips all checks (use `skip` option in workflow definition instead).
-   *
-   * @param bridgeSkip Whether to skip bridge checks (optional).
-   * @param bridgeUrl URL of the bridge (optional).
-   * @return True if bridge skip is happening, false otherwise.
-   */
-  private shouldSkipHalt(bridgeSkip: boolean | undefined, bridgeUrl: string | undefined): boolean {
-    return bridgeSkip === true || !!bridgeUrl;
-  }
```
rifont, line 234: This method is no longer needed.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -301,16 +287,15 @@ export class SendMessage {
...
     const template = await this.getNotificationTemplate({
       _id: job._templateId,
       environmentId: job._environmentId,
     });
-    if (!template) {
-      throw new PlatformException(`Notification template ${job._templateId} is not found`);
-    }
 
-    if (template.critical || this.isActionStep(job)) {
+    if (template?.critical || this.isActionStep(job)) {
```
rifont: Template can be undefined at this point of the flow now.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -346,18 +331,40 @@ export class SendMessage {
...
+       *
+       * TODO: Refactor the send-message flow to better handle stateless workflows
+       */
+      const workflowPreference = GetPreferences.mapWorkflowChannelPreferencesToChannelPreferences(
+        command.statelessPreferences
+      );
+      subscriberPreference = {
+        enabled: true,
+        channels: workflowPreference,
+      };
```
rifont, line 344: In keeping with a low-touch approach, we have this path specifically for stateless workflow preference evaluation.

libs/application-generic/src/usecases/create-notification-jobs/create-notification-jobs.usecase.ts
```diff
@@ -64,6 +64,7 @@ export class CreateNotificationJobs {
       channels,
       controls: command.controls,
       tags: command.template.tags,
+      preferences: command.template.preferences,
```
rifont: Taking the same approach as we had with `tags`, by injecting the `preferences` onto the mapped `template` that was derived from the discovered bridge workflow.

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -94,11 +98,7 @@ export class GetPreferences {
   /** Transform WorkflowChannelPreferences into IPreferenceChannels */
   public static mapWorkflowChannelPreferencesToChannelPreferences(
     workflowPreferences: WorkflowChannelPreferences,
-  ): IPreferenceChannels | undefined {
-    if (!workflowPreferences) {
-      return undefined;
-    }
-
+  ): IPreferenceChannels {
```
rifont, line 101: The changes in this file ensure the `mapWorkflowChannelPreferencesToChannelPreferences` will only be used for defined `workflowPreferences`, so we need to handle the `undefined` results outside of this mapper now.

libs/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -79,10 +86,11 @@ export class GetSubscriberTemplatePreference {
...
 
-    const subscriberPreferenceChannels =
-      GetPreferences.mapWorkflowChannelPreferencesToChannelPreferences(
-        subscriberWorkflowPreferences,
-      ) || subscriberPreference?.channels;
+    const subscriberPreferenceChannels = subscriberWorkflowPreferences
+      ? GetPreferences.mapWorkflowChannelPreferencesToChannelPreferences(
+          subscriberWorkflowPreferences,
+        )
+      : subscriberPreference?.channels;
```
rifont, line 93: Conditionally populating `subscriberPreferenceChannels` for both the V1 & V2 preferences, preferring the V2 preference if it exists.

libs/dal/src/repositories/notification/notification.entity.ts
```diff
@@ -38,6 +38,7 @@ export class NotificationEntity {
   updatedAt?: string;
   expireAt?: string;
   tags?: string[];
+  preferences?: WorkflowChannelPreferences;
```
rifont: This is probably the most uncomfortable part - `preferences` is never persisted to `NotificationTemplateEntity` for both _stateful and stateless_ executions, this change here is made to keep in line with mapping of `tags` for stateless workflow executions to the `NotificationTemplateEntity` which the SubscriberJobBound use-case deals with via injection.

I'm open to suggestions for other ways to handle this which don't involve changes to the `NotificationTemplateEntity` to deal with `preferences`.

apps/worker/src/app/workflow/usecases/send-message/send-message.command.ts
SokratisVidros: Good idea. I agree 100% with the approach. This should go hand-in-hand with the NotificationTemplate to Workflow remodelling.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
SokratisVidros: Is the first filter about skip conditions? If yes I'd say we can do `this.filterSkipConditions` to be more precise.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -346,18 +331,40 @@ export class SendMessage {
...
-        environmentId: job._environmentId,
-        template,
-        subscriber,
-        tenant: job.tenant,
-      })
-    );
+    let subscriberPreference: { enabled: boolean; channels: IPreferenceChannels };
+    if (command.statelessPreferences) {
+      /*
+       * Stateless Workflow executions do not have their definitions stored in the database.
```
SokratisVidros, line 333: @rifont Would it be more helpful to apply the same strategy when it comes to stored workflows by adapting the event trigger usecase and fetching their preferences from the database, before triggering the `send-message.usecase`? 

In other words, always inject preferences to `send-message` and all the downstream use-cases.

libs/dal/src/repositories/notification/notification.entity.ts
SokratisVidros: Does this change mean that we will start persisting them in Mongo on the Notification entity? If not mistaken, we should do that as it's the part that we actually copy the workflow & preference configuration from its blueprint. 

Am I missing something?

SokratisVidros (changes requested):

apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
djabarovgeorge, line 1: tests! 🔥🥳

apps/worker/package.json
djabarovgeorge: I wonder if we could find a better place for this env variable because this script could get very long quickly.
does the line in the test `process.env.IS_WORKFLOW_PREFERENCES_ENABLED = 'true';` did not work as expected?

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
djabarovgeorge: We're confident that the tests cover the use of the filter and filterPreferredChannels methods below? ensuring that no exceptions will be thrown due to unsynced workflows in these methods

```
    const [filterResult, channelPreferenceResult] = await Promise.all([
      this.filter(command, variables),
      this.filterPreferredChannels(command.job),
    ]);
```

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
djabarovgeorge: 🧹

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -301,16 +287,15 @@ export class SendMessage {
   }
 
   @Instrument()
-  private async filterPreferredChannels(job: JobEntity): Promise<boolean> {
+  private async filterPreferredChannels(command: SendMessageCommand): Promise<boolean> {
+    const { job } = command;
+
     const template = await this.getNotificationTemplate({
```
djabarovgeorge: If you dont mind could you rename this one if we are already here 🫠
I really hate those `template` variables in the code. :( 

```suggestion
    const workflow = await this.getWorkflow({
```

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
djabarovgeorge: We don't care about the "critical" workflow here because it will be handled in the new get preference with the command.preference right?

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
rifont: Yes, it is. I renamed again to be more precise, we now have:
```typescript
    const [stepCondition, channelPreference] = await Promise.all([
      this.evaluateStepCondition(command, variables),
      this.evaluateChannelPreference(command),
    ]);
```

libs/dal/src/repositories/notification/notification.entity.ts
rifont: No, it doesn't mean we will start persisting them. This is purely to make Typescript happy and avoid a sweeping change to the way NotificationTemplate settings are injected in the send-message flow.

Workflow Preferences are stored on the `Preferences` entity instead of the Workflow with this release, to provide a single source of truth for all preference lookups. The downside of this approach is that without making a sweeping change to the send-message flow, this property sits as an artifact on the NotificaitonTemplateEntity until we do a refactor.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
djabarovgeorge: I don’t see a big difference in your suggestion, as this logic will still need to be handled somewhere. As Richard mentioned, I agree that we would benefit greatly from refactoring the send message use case into a different structure. This way, data retrieval tasks like extracting `workflows` and `preferences`, as well as actions like `filtering`, can be managed by dedicated handlers and act as a middleware for the send action. This would make the code much cleaner and easier to maintain.

rifont: Yes, it would be more helpful. There's a question of whether we want to make such a change now, or whether we can live with intentional naming that highlights this shortcoming for the interim.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
```diff
@@ -346,18 +331,40 @@ export class SendMessage {
...
+        command.statelessPreferences
+      );
+      subscriberPreference = {
+        enabled: true,
+        channels: workflowPreference,
+      };
+    } else {
+      if (!template) {
+        throw new PlatformException(`Notification template ${job._templateId} is not found`);
+      }
```
djabarovgeorge: What do you think about adding this check above so we will have more flattened code?

```suggestion
      if (!template && !command.statelessPreferences) {
        throw new PlatformException(`Notification template ${job._templateId} is not found`);
      }
```

libs/dal/src/repositories/notification/notification.entity.ts
rifont: One approach we can take to address this is to start persisting the Workflow Preference on the NotificationTemplateEntity and modify the `GetPreference` use-case to retrieve the Workflow Preference from the NotificationTemplateEntity during preference evaluation, rather than persist/retrieve the Workflow Preference from the Preference entity.

This feels like the correct change to enable a single entity copy use-case suggested above, and resolve this uncomfortable artifact.

djabarovgeorge: Thanks for bringing this up. Another solution, as I mentioned in our sync, is to reuse the same logic we have for controls. 
Currently, [we store the controls](https://github.com/novuhq/novu/blob/nv-4290-get-v1-controls-step-type-schema/apps/worker/src/app/workflow/usecases/subscriber-job-bound/subscriber-job-bound.usecase.ts#L178) (received from the trigger payload) in the job entity. Since the unsynced flow is not production-related, the storage cost should be negligible. This way, we can use the stored data as needed without modifying the Notification entity for local debugging. 
When we have a better architecture in place, we can create a cleaner solution. 
Let me know your thoughts on this.

djabarovgeorge (commented):
Awesome work, i left couple of comments :)

libs/dal/src/repositories/notification/notification.entity.ts
SokratisVidros: George is right, after checking the code it needs to start rom the job entity, and then possible evaluate of we have to propagate them to the notifications entity. We need to apply the "copy from blueprint" on even trigger logic for all the workflow definition parts. 

Controls and preferences are part of the workflow definition and we need to treat them as such in the code. Having said that, we can for now make them follow the same logic and in future work, teach the code to consider them one think, that is:

Workflow definition: Steps (the JSON from the TS code that is now the NotificationTemplate), Controls, Preferences, Tags.

rifont: I'm inclined to go in this direction of storing the `preferences` on the job, that way we don't need to modify the NotificationTemplate & Notification entity and the `preferences` are passed purely as an implementation detail. We won't be locked in to any DAL changes, and the refactoring later will be easier.

Great suggestion!

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
rifont: That's correct for V2 Preferences, however we still need this check in place to cater to V1 Workflows. I will add a code comment to make that clear.

apps/worker/package.json
rifont: I could drop this into `.env.test` in Worker instead, I agree it could get unwieldy. I'll do that instead.

The line suggested doesn't work because we are running the tests that depend on this env-var being set from API, rather than Worker.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
rifont: Good catch, done!

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
rifont: I prefer the `if/else` approach because it creates a clear delineation between both stateless/stateful workflow execution modes. The suggestion above leaks "stateless" mode into the stateful mode with the additional check. Summarising, I think the code is more clear as it stands currently.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
rifont: As agreed, we will address these changes in future refactoring.

apps/worker/src/app/workflow/usecases/send-message/send-message.usecase.ts
rifont: Yes, we are confident - the `bridge-trigger` tests evaluate both stateful and stateless Workflow execution modes, ensuring that these paths are fully tested.

apps/worker/src/app/workflow/usecases/run-job/run-job.usecase.ts
```diff
@@ -99,6 +99,7 @@ export class RunJob {
           events: job.digest?.events,
           job,
           tags: notification.tags || [],
+          statelessPreferences: job.preferences,
```
SokratisVidros, line 102: Is it necessary to call these `statelessPreferences`? We are storing them regardless of the type of the workflow, right?

rifont: No, only stateless workflow executions have their preferences stored on the `JobEntity`. Stateful Workflows have their preferences stored on the `PreferenceEntity` during sync, [here](https://github.com/novuhq/novu/blob/next/apps/api/src/app/bridge/usecases/sync/sync.usecase.ts#L219).

SokratisVidros: Yes. Let me rephrase: would it be easier for the use case and our reasoning about it to store all preferences in the job when triggered? That way, we ensure that all types of workflows have the same preference resolution in case preferences change as the workflow executes.

rifont: From a User's point of view, it's arguable that preference resolution should use be in real-time to use the most up-to-date Preference, ensuring that Workflows with long delays use the most recently modified Preference, rather than the Preference set at trigger time. If we store preferences on the job for stateful Workflow executions, we risk not respecting the most up to date Preference, and Subscriber's could receive notifications when the User didn't intend them to.

One inverted approach that @djabarovgeorge and I discussed to add consistent Preference resolution (and by extension, execution) for both stateless + stateful Workflow executions involves storing a "transient" (could be coined: "Transient Sync") `NotificationTemplateEntity` & `PreferenceEntity` for Stateless executions during Trigger. Each of these entities can have a `stateful: boolean` flag added to indicate they are transient, to support API client lookups that only care about stateful entities. 
This approach will enable the Worker to be refactored to not care about stateless vs stateful entirely. There is considerable refactoring required to take this approach, however.

SokratisVidros: I see. Let's part it for now although we need to get back to this.

SokratisVidros (approved):

PR #6528, opened 2024-09-18
feat(api): add priority to queue based from 0.24.x (ref: PR 6489)

isaiahdahl:
see: https://github.com/novuhq/novu/pull/6489

### DIFF
--- a/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
+++ b/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
@@ -8,6 +8,7 @@ import {
 } from '@novu/shared';
 
 import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
+import { ApiHideProperty } from '@nestjs/swagger';
 
 export class ParseEventRequestBaseCommand extends EnvironmentWithUserCommand {
   @IsDefined()
@@ -38,6 +39,7 @@ export class ParseEventRequestBaseCommand extends EnvironmentWithUserCommand {
   @IsEnum(TriggerRequestCategoryEnum)
   requestCategory?: TriggerRequestCategoryEnum;
 
+  @ApiHideProperty()
   @IsOptional()
   @IsNumber()
   priority?: number;
--- a/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
+++ b/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
@@ -1,4 +1,4 @@
-import { IsDefined, IsString, IsOptional, ValidateNested, ValidateIf, IsEnum } from 'class-validator';
+import { IsDefined, IsString, IsOptional, ValidateNested, ValidateIf, IsEnum, IsNumber } from 'class-validator';
 import {
   AddressingTypeEnum,
   TriggerRecipients,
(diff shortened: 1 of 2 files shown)


SokratisVidros (approved):

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
```diff
@@ -37,6 +37,10 @@ export class ParseEventRequestBaseCommand extends EnvironmentWithUserCommand {
   @IsOptional()
   @IsEnum(TriggerRequestCategoryEnum)
   requestCategory?: TriggerRequestCategoryEnum;
+
+  @IsOptional()
+  @IsNumber()
+  priority?: number;
```
rifont, line 43: This change will result in the `priority` property becoming available on the OpenAPI spec at https://api.novu.co/openapi.yaml, which in turn, is used to generate documentation at https://docs.novu.co/api-reference/overview. 

Given that this property is only usable in self-hosted via `IS_DOCKER_HOSTED === 'true'` in the implementation, we also need to conditionally apply [@ApiHideProperty](https://docs.nestjs.com/openapi/cli-plugin#:~:text=ApiProperty%20unless%20%40-,ApiHideProperty,-is%20used) when the same condition is met, to ensure the property doesn't become available in public documentation, which pertains only to usage of Novu Cloud at present.

rifont, line 45: This change will result in the `priority` property becoming available on the OpenAPI spec at https://api.novu.co/openapi.yaml, which in turn, is used to generate documentation at https://docs.novu.co/api-reference/overview. 

Given that this property is only usable in self-hosted via `IS_DOCKER_HOSTED === 'true'` in the implementation, we also need to conditionally apply [@ApiHideProperty](https://docs.nestjs.com/openapi/cli-plugin#:~:text=ApiProperty%20unless%20%40-,ApiHideProperty,-is%20used) when the same condition is met, to ensure the property doesn't become available in public documentation, which pertains only to usage of Novu Cloud at present.

rifont (changes requested):
Thank you for your contribution 🙏

Please apply the ApiHideProperty flag conditionally to ensure the `priority` flag appears on the OpenAPI spec only for self-hosted instances.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
isaiahdahl: Done

SokratisVidros (approved):

rifont (approved):
We will accept the PR in the current state, but note that the Community Edition won't have the new `priority` property available in the OpenAPI specification until we conditionally expose the property via a custom decorator that operates on the `IS_DOCKER_HOSTED` environment variable flag.

We will raise a ticket to address this conditionality for the Community Edition in a separate follow-up ticket.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.command.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from '@novu/shared';
 
 import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
+import { ApiHideProperty } from '@nestjs/swagger';
```
rifont: Needed to resolve the import order linting problem [here](https://cloud.nx.app/runs/x6j60dkQbF).

```suggestion
} from '@novu/shared';
import { ApiHideProperty } from '@nestjs/swagger';

import { EnvironmentWithUserCommand } from '../../../shared/commands/project.command';
```

isaiahdahl: done

isaiahdahl:
This was at one point passing and ready to merge, but looks like now the CI has changed.

PR #6531, opened 2024-09-18
feat(js,react): Export InboxContent component

desiprisg:

This PR:
* Exports the `<InboxContent />` component which is what is used under the hood to render `<Notifications />` and/or `<Preferences />`. It's usage will be primarily with custom popovers (See example in playground).
* Fixes the custom popover issue of not opening on click, because of the react tree hierarchy (caused by our rendering higher up). This is done via one renderer per component we export in `@novu/react`.
* Fixes the issue where the custom popover closes because we render something in a portal (thus the popover loses focus). This is done via rendering all dropdowns and tooltips without a portal.
  * Also fixed an issue that appeared because of this, with `autoUpdate` where the layout was affected in a weird way. Passed options ```elementResize: false,
        ancestorScroll: false,
        animationFrame: false,
        layoutShift: false``` to it to fix.
### Screenshots

<details>

</details>


packages/react/src/components/NovuUI.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { ComponentType, PropsWithChildren, ReactNode, RefObject, useCallback, useEffect, useState } from 'react';
+import ReactDOM from 'react-dom';
```
LetItRock:
```suggestion
import React, { useCallback, useEffect, useState } from 'react';
```

packages/react/src/components/Notifications.tsx
```diff
@@ -35,3 +38,5 @@ export const Notifications = React.memo((props: NotificationProps) => {
 
   return <Mounter mount={mount} />;
 });
+
+export const Notifications = withRenderer(_Notifications);
```
LetItRock, line 42: 💭 **thought (non-blocking):** It seems like Renderer + Mounter can become one thing as they would always be used in the pair. And all this could be simplified to something like (name tbd):
```
const Notifications = () => {
   ...
   const mount = (el, mountElement) => {
      return novuUI.mountComponent({
        name: 'Notifications',
        element,
        props: {
          renderNotification: renderNotification
            ? (el, notification) => mountElement(el, renderNotification(notification))
            : undefined,
          onNotificationClick,
          onPrimaryActionClick,
          onSecondaryActionClick,
        },
   }
   
   return <RendererPlusMounter mount={mount} />
}

const RendererPlusMounter = () => {
  ...
  // combination of Mounter and Renderer code
  ...

  useEffect(() => {
    ...
    if (element && mount) {
      const possibleUnmount = mount(element, mountElement);
      ...
    }
   ...
  }, [ref, mount]);
  
  return <RendererProvider value={{ mountElement }}>
      {[...mountedElements].map(([element, mountedElement]) => {
        return createPortal(mountedElement, element);
      })}
      <div ref={ref} />
    </RendererProvider>
}
```
It's just a thought, not a suggestion. WDYT?

LetItRock (commented):
Overall works well, great job! 🙌 
Just noticed that the custom popover looks a bit odd.

Screenshot 2024-09-19 at 18 13 12

packages/react/src/components/Notifications.tsx
desiprisg: Not all components we use will need a renderer, that's why I've separated them!

desiprisg:
> Overall works well, great job! 🙌 Just noticed that the custom popover looks a bit odd.
> 
> Screenshot 2024-09-19 at 18 13 12

Yeah I never applied a max height I think, that's why!

playground/nextjs/src/pages/custom-popover/index.tsx
```diff
@@ -1,35 +1,26 @@
...
+          <Bell
+            renderBell={(unreadCount) => (
+              <div>
+                <span>{unreadCount}</span>
+                <BellIcon />
+              </div>
+            )}
+          />
+        </PopoverTrigger>
+        <PopoverContent className="max-h-[500px] w-[400px] overflow-auto p-0">
```
LetItRock:
```suggestion
        <PopoverContent className="h-[500px] w-[400px] overflow-auto p-0">
```

packages/js/src/ui/components/primitives/Tooltip/TooltipContent.tsx
```diff
@@ -28,7 +28,7 @@ const TooltipContentBody = (props: TooltipContentProps) => {
     <div
       ref={setFloating}
       class={local.class ? local.class : style(local.appearanceKey || 'tooltipContent', tooltipContentVariants())}
-      style={floatingStyles()}
+      style={{ ...floatingStyles(), 'z-index': 99999 }}
```
LetItRock, line 31: `10` value works here, but not sure why it's `10` 😅

LetItRock (approved):
works well, I didn't notice anything! awesome work! 🙌

packages/js/src/ui/components/primitives/Tooltip/TooltipContent.tsx
desiprisg: I don't think `10` would work in a custom popover scenario. You can't know what the host app has rendered and with what `z-index` and since the tooltip is portaled, it might show under something else. Does this make sense?
Example with `z-index: 10` in our playground with a custom popover from radix: 
image

PR #6537, opened 2024-09-19
fix(web): identify from app to pass traits

scopsy:

This helps with passing the proper account details to destinations
### Screenshots

<details>

</details>


djabarovgeorge (approved):
i wonder how this fixes the issue don't we identify by Id?

scopsy:
@djabarovgeorge It will just help with updating the contact credentials during the identify method, we also send the information when upserting users. But not all segment destinations support this. This particular one is for Hubspot

PR #6543, opened 2024-09-20
feat(api): improve typescript compilation to avoid reindexing in webs…

tatarco:
due to multiple freezing issues in webstorm I went to understrand why .js files are being created in place by my ide, the research results ended up with typescript as the culprit, I believe our project is not configured correctly, there is more work to be done, but the project typescript definitions are not aligned between modules, this alone made a huge change and for 48 hours i havent had anymore webstorm issues
### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/tsconfig.json
+++ b/apps/api/tsconfig.json
@@ -1,15 +1,15 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "outDir": "../../dist/libs/api",
+    "rootDir": ".",
     "allowJs": false,
-    "allowSyntheticDefaultImports": true,
-    "emitDecoratorMetadata": true,
-    "esModuleInterop": true,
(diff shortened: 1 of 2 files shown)


apps/api/tsconfig.json
```diff
@@ -1,15 +1,15 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "outDir": "../../dist/libs/api",
```
SokratisVidros, line 4: Did you try Webstorm without this changes, only with the addition `"include": ["src"],`

SokratisVidros: Can you please double-check that this is the intended behavior? This will export JS files in the top `/dist/libs/api` folder. It breaks the package boundaries of the monorepo and I am pretty sure our Docker setup won't be happy about it.

PR #6547, opened 2024-09-22
feat(web): Hub spot identify

scopsy:

- On signup this properly ties the Hubspot entity with the new account.
- 
### Screenshots

<details>

</details>


apps/api/src/app/analytics/analytics.controller.ts
```diff
@@ -26,4 +31,41 @@ export class AnalyticsController {
...
     };
   }
+
+  @Post('/identify')
+  @ExternalApiAccessible()
+  @UserAuthentication()
+  async identifyUser(@Body() userData: any, @UserSession() user: UserSessionData): Promise<any> {
+    try {
+      const hubspotPortalId = '44416662';
+      const hubspotFormId = 'fc39aa98-4285-4322-9514-52da978baae8';
```
scopsy: Left them here hardcoded, as those are public values.

apps/web/src/api/telemtry.ts
djabarovgeorge, line 1: can we rename the file to `telemetry` instead?

apps/web/src/api/api.client.ts
```diff
@@ -45,7 +45,7 @@ export const api = {
   },
   post(url: string, payload, params?: CustomDataType) {
     return axios
-      .post(`${API_ROOT}${url}`, payload, { params, headers: getHeaders() })
+      .post(`${API_ROOT}${url}`, payload, { params, headers: { ...getHeaders() } })
```
djabarovgeorge: what is the difference here between `getHeaders()` and `{ ...getHeaders() }`?

apps/api/src/app/analytics/analytics.controller.ts
djabarovgeorge: Do those identifiers tend to change? if there are, should we consider storing them in the environment variables?

apps/api/src/app/analytics/analytics.controller.ts
djabarovgeorge, line 1: All of this logic could be a use case :), for your consideration of the refactoring is needed

djabarovgeorge (approved):
go go analytics!!! awesome work I added some comments to your review

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
```diff
@@ -58,6 +59,20 @@ export function QuestionnaireForm() {
     };
     await updateOrganizationMutation(updateClerkOrgDto);
 
+    const hubspotContext = document.cookie
```
SokratisVidros: You can use [js-cookies](https://github.com/novuhq/novu/blob/5cd0e5a6a1c9aaa0fd042b62959d78ae02c58994/apps/web/package.json#L99) to get the hubspot cookie value

SokratisVidros (commented):
Does this change ensure that Hubspot context is the same across subdomains, for example between nocu.co and dashboard.novu.co?

apps/web/src/api/api.client.ts
scopsy: Good catch, I just had to pass some headers here previously. Not needed anymore. Reverted.

scopsy:
Yes 🙏 Did a few tests and so far so good. Will also monitor this in the wild

PR #6562, opened 2024-09-24
feat(dashboard): vite multi page app poc

LetItRock:
This PR includes the base configuration and folder structure to run and deploy the new Dashboard Vite Multipage App.
Check out the comments to understand how it works.

### Screenshots

https://github.com/user-attachments/assets/e4fc08d6-617b-4539-9863-20c72df178b1


apps/dashboard/netlify.toml
```diff
@@ -0,0 +1,31 @@
+[build.environment]
+  DISABLE_ESLINT_PLUGIN="true"
+
+[build]
+  # default build command
+  command = "pnpm run build:dashboard --skip-nx-cache"
```
LetItRock, line 6: the new dashboard build command in the root `package.json`

apps/dashboard/netlify.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[context.deploy-preview]
+  command = "pnpm run build:dashboard --skip-nx-cache"
+
+[[redirects]]
+  from = "/legacy/*"
+  to = "/legacy/index.html"
+  status = 200
```
LetItRock, line 14: Redirect the `text/html` requests with the `/legacy/*` to the `/legacy/index.html` file.
For example, the case, when the user lands on the `/legacy/integrations` URL, the browser sends the request to get the default `index.html` file by that URL, but if there is no such resource under that URL and the HTTP server will return the default `index.html` file for the `/` URL, and in our case that would be the new Dashboard `index.html`, but this is not what we want, as we want to serve `legacy/index.html` file. That's why we need to tell the server how to deal with requests with `/legacy/*` URLs.
This solves the issue of refreshing the page under the `/legacy/integrations` URL, and also browser navigation with back and forward buttons.

apps/dashboard/package.json
```diff
@@ -4,9 +4,13 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "start": "vite",
+    "start": "pnpm legacy:clean && pnpm legacy:copy && vite",
```
LetItRock, line 7: For the local development, clean and copy the web app bundle files to the `/legacy` folder. This command assumes that the Web app was already built "for the dashboard." I'll explain later in the comments what this means.

apps/dashboard/package.json
```diff
@@ -4,9 +4,13 @@
...
     "dev": "pnpm start",
-    "build": "tsc -b && vite build",
+    "build": "pnpm legacy:prepare && tsc -b && vite build",
```
LetItRock, line 9: However, when doing the `build` we need to build the Web app too.

apps/dashboard/vite.config.ts
```diff
@@ -1,7 +1,44 @@
...
+  plugins: [
+    react(),
+    viteStaticCopy({
+      targets: [
+        {
+          src: path.resolve(__dirname, './legacy') + '/[!.]*',
+          dest: './legacy',
+        },
+      ],
+    }),
```
LetItRock, line 17: Copy the `legacy` folder to the `dist` bundle folder.

apps/dashboard/vite.config.ts
```diff
@@ -1,7 +1,44 @@
...
+              return;
+            } else {
+              res.statusCode = 404;
+              res.end('Legacy index.html not found');
+              return;
+            }
+          }
+          next();
+        });
+      },
```
LetItRock: This is the redirect configuration for the local development to serve the `/legacy/index.html` file. Check the explanation above in the `netlify.toml` file.

apps/web/.env.dashboard
```diff
@@ -0,0 +1,21 @@
...
+REACT_APP_SENTRY_DSN=
+REACT_APP_BLUEPRINTS_API_URL=
+REACT_APP_MAIL_SERVER_DOMAIN=
+REACT_APP_LAUNCH_DARKLY_CLIENT_SIDE_ID=
+IS_TEMPLATE_STORE_ENABLED=
+IS_MULTI_PROVIDER_CONFIGURATION_ENABLED=
+IS_MULTI_TENANCY_ENABLED=
+REACT_APP_CLERK_PUBLISHABLE_KEY=
+REACT_APP_IS_EE_AUTH_ENABLED=
+FRONT_BASE_CONTEXT_PATH=/legacy/
```
LetItRock, line 21: The last line in this file is important as it sets the "base" path for all static resources and route requests from the Web application. This is the "runtime" variable, meaning it should be defined in the `env-config.js` file that is served with the `index.html` file.

LetItRock: This file is used to specifically build the `env-config.js` for the Web app bundle when used in the new Dashboard app.

apps/web/env.sh
```diff
@@ -5,9 +5,11 @@ rm -rf ./env-config.js
 
 touch ./env-config.js
 
+envFile=${1:-.env.sample}
```
LetItRock, line 8: allow passing the `.env` file name, otherwise use default `.env.sample`

apps/web/package.json
```diff
@@ -6,10 +6,12 @@
     "start": "pnpm panda --watch & cross-env NODE_OPTIONS=--max_old_space_size=8192 DISABLE_ESLINT_PLUGIN=true PORT=4200 react-app-rewired start",
     "prebuild": "rimraf build",
     "build": "pnpm panda && cross-env NODE_OPTIONS=--max_old_space_size=4096 GENERATE_SOURCEMAP=false DISABLE_ESLINT_PLUGIN=true react-app-rewired --max_old_space_size=4096 build",
+    "build:web-for-dashboard": "pnpm panda && cross-env NODE_OPTIONS=--max_old_space_size=4096 GENERATE_SOURCEMAP=false DISABLE_ESLINT_PLUGIN=true PUBLIC_URL=/legacy/ react-app-rewired --max_old_space_size=4096 build && pnpm envsetup:dashboard",
```
LetItRock, line 9: The difference from the normal build is that we use `PUBLIC_URL=/legacy/` variable which will build the `index.html` file with all the resources URLs starting with the `/legacy` prefix.

LetItRock: Then, after the build, it also runs the `envsetup: dashboard,` which creates the `env-config.js` file.

nx.json
```diff
@@ -53,6 +53,11 @@
       "dependsOn": [
         "^build"
       ]
+    },
+    "build:web-for-dashboard": {
+      "dependsOn": [
+        "^build"
+      ]
```
LetItRock, line 60: When `nx build:web-for-dashboard @novu/web` command is running --> build all the web app dependencies first.

apps/dashboard/package.json
LetItRock: So basically in the `apps/web` folder you need to run the `pnpm run build:web-for-dashboard` command.

desiprisg (approved):

apps/dashboard/package.json
SokratisVidros: What is `tsc -b` necessary?  Won't Vite do the heavy lifting?

apps/dashboard/package.json
```diff
@@ -4,9 +4,13 @@
...
+    "legacy:prepare": "pnpm legacy:clean && pnpm legacy:build && pnpm legacy:copy",
+    "legacy:clean": "rimraf legacy/*",
+    "legacy:build": "cd ../.. && pnpm build:web-for-dashboard",
```
SokratisVidros, line 12: Since these tasks affect the web app, shall we add them to the package.json of `web`?

apps/dashboard/vite.config.ts
```diff
@@ -1,7 +1,44 @@
...
+            const legacyPath = path.resolve(__dirname, 'legacy', 'index.html');
+            const fileExists = fs.existsSync(legacyPath);
+
+            if (fileExists) {
+              // Serve the legacy index.html file
+              res.setHeader('Content-Type', 'text/html');
+              res.statusCode = 200;
+              fs.createReadStream(legacyPath).pipe(res);
+              return;
+            } else {
```
SokratisVidros:
```suggestion
            }
```

SokratisVidros (approved):

apps/dashboard/package.json
LetItRock: this part `tsc -b && vite build` was generated by Vite cli

apps/dashboard/package.json
LetItRock: This one `build:web-for-dashboard` was added to the `web`, the others, are the logic of cleaning and copying that I think should stay in the `dashboard` project.

PR #6571, opened 2024-09-25
feat(dashboard): Add clerk and segment

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/config/index.ts
+++ b/apps/dashboard/src/config/index.ts
@@ -1,0 +1,16 @@
+export const SENTRY_DSN = import.meta.env.REACT_APP_SENTRY_DSN;
+
+export const LAUNCH_DARKLY_CLIENT_SIDE_ID = import.meta.env
+  .REACT_APP_LAUNCH_DARKLY_CLIENT_SIDE_ID;
+
+export const HUBSPOT_PORTAL_ID = import.meta.env.REACT_APP_HUBSPOT_EMBED;
+
+export const IS_EE_AUTH_ENABLED =
+  import.meta.env.REACT_APP_IS_EE_AUTH_ENABLED === "true";
+
+export const CLERK_PUBLISHABLE_KEY =
+  import.meta.env.REACT_APP_CLERK_PUBLISHABLE_KEY || "";
+
+if (IS_EE_AUTH_ENABLED && !CLERK_PUBLISHABLE_KEY) {
+  throw new Error("Missing Clerk Publishable Key");
+}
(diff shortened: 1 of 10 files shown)


apps/dashboard/src/routes/root.tsx
```diff
@@ -1,26 +1,32 @@
...
-              target="_self"
-              className="text-blue-600 visited:text-purple-600 hover:border-current hover:border-b"
-            >
-              Go to the Web App
-            </a>
-          </li>
-        </ul>
+    <QueryClientProvider client={queryClient}>
+      <ClerkProvider>
+        <SegmentProvider>
```
LetItRock, line 12: ✏ **suggestion:** can we maybe move this to the `Providers` component?

apps/dashboard/src/utils/segment.ts
```diff
@@ -0,0 +1,135 @@
+import { AnalyticsBrowser } from "@segment/analytics-next";
+import { IUserEntity } from "@novu/shared";
```
LetItRock, line 2: 💭 **thought (non-blocking):** we need to discuss whether we want to use the `@novu/shared` package, because it's a kind of bloated package

LetItRock (approved):

apps/dashboard/src/utils/segment.ts
LetItRock: https://novu.slack.com/archives/C07HVLG7G2U/p1727277810634729

apps/dashboard/src/config/index.ts
```diff
@@ -0,0 +1,16 @@
+export const SENTRY_DSN = import.meta.env.REACT_APP_SENTRY_DSN;
```
SokratisVidros, line 1: Shouldn't this be `VITE_` prefixed?

apps/dashboard/src/routes/root.tsx
SokratisVidros: I am not a big fan of this split, as we did it in the `web` project. It's better to see all of the context in the root file. If it gets too big, we will revise.

SokratisVidros (approved):

apps/dashboard/src/utils/segment.ts
LetItRock: we decided to use `import type ... from '@novu/shared';`

SokratisVidros: Can we add a quick eslint rule for this?

apps/dashboard/src/config/index.ts
desiprisg: yes, fixed it in another PR

PR #6574, opened 2024-09-25
feat(api, js, react): Com 244 hide critical workflow preferences from inbox

BiswaViraj:

- filter preferences using readOnly flag
### Screenshots

<details>

</details>


libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -33,6 +33,7 @@ export class GetSubscriberPreference {
       {
         organizationId: command.organizationId,
         environmentId: command.environmentId,
+        critical: command.readOnly,
```
rifont: The `critical` flag is always `false` for code-first Workflows created via the Bridge syncing method, so I believe this implementation will never return code-first defined Workflows when `command.readOnly=true`. Instead, Bridge defined Workflows only use the `PreferenceEntity.preferences.all.readOnly` flag to mark a Workflow as "critical".

Can you please test this and confirm?

If this is the case, we will need to fetch all Workflows (i.e. remove this additional filter on the DB layer) and perform custom filtering using JS to handle both of:
1. V1 Workflows: Workflows who have `critical: true`
2. V2 Workflows: Workflows who have `type: 'BRIDGE'`, combined with a `PreferenceEntity.preferences.all.readOnly: true`

rifont (changes requested):
Looks great, just a question regarding Bridge Workflows filtering that we need to address.

libs/application-generic/src/usecases/get-preferences/get-preferences.usecase.ts
rifont, line 1: Regarding the requested changes, if we are doing the filtering in JS, I think we should revert all changes to `get-preferences` use-case otherwise the filtering logic will become quite complex.

Futhermore, the `get-preferences` command handles logic to determine when a Preference "set" (i.e. the merged Code-First & User-defined) is `readOnly: true`. The consumer of this use-case should use that merged `readOnly` value rather than perform DB filtering operations which can prevent merging behaviour from working as intended.

SokratisVidros: Good catch. In this case, let's make unify the behavior without any client-specific query string option. The API

libs/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.command.ts
```diff
@@ -1,7 +1,6 @@
 import { NotificationTemplateEntity, SubscriberEntity } from '@novu/dal';
-import { IsDefined, IsNotEmpty, IsOptional } from 'class-validator';
-
 import { ITenantDefine } from '@novu/shared';
+import { IsDefined, IsNotEmpty, IsOptional } from 'class-validator';
```
rifont: Can we revert the whitespace changes here and on the below file to keep the PR tidy?

packages/js/src/ui/components/elements/Preferences/Preferences.tsx
```diff
@@ -164,27 +159,9 @@ const PreferencesRow = (props: {
...
-                      >
-                        <Lock />
-                      </span>
-                    )}
-                  />
-                  <Tooltip.Content data-localization="preferences.workflow.disabled.tooltip">
-                    {t('preferences.workflow.disabled.tooltip')}
-                  </Tooltip.Content>
-                </Tooltip.Root>
-              </Show>
```
BiswaViraj, line 187: It no more fetches the critical workflows show we don't need this UI part to indicate a workflow is critical

SokratisVidros (approved):
💯

packages/js/src/ui/components/elements/Preferences/ChannelRow.tsx
BiswaViraj, line 1: removed UI for critical workflows

packages/js/src/ui/components/elements/Preferences/Switch.tsx
BiswaViraj, line 1: removed data-attributes and styles related to critical workflows

apps/api/src/app/inbox/usecases/get-preferences/get-preferences.usecase.ts
```diff
@@ -90,6 +91,8 @@ export class GetPreferences {
       })
     );
 
-    return [updatedGlobalPreference, ...workflowPreferences];
+    const nonCriticalWorkflows = workflowPreferences.filter((preference) => preference.workflow.critical === false);
```
BiswaViraj, line 94: only send non-critical workflows

rifont (approved):
Awesome work! 🚅

PR #6575, opened 2024-09-25
refactor(api, application-generic, web, shared, dal): Remove `readOnly` from Preferences `channels`

rifont:

* Remove `readOnly` from Preferences `channels`
* Add explicit "Override Preferences" toggle in UI to control specification of User Workflow preferences

### Screenshots

_Local Studio_
image

_Dashboard - **without** overrides_
image

_Dashboard - **with** overrides_
image

<details>

</details>


apps/web/src/hooks/workflowPreferences/useUpdateWorkflowPreferences.ts
```diff
@@ -4,20 +4,30 @@ import { useNovuAPI } from '../useNovuAPI';
...
-    WorkflowPreferences
-  >((data) => api.upsertPreferences(workflowId, data), {
-    ...options,
-  });
+    WorkflowPreferences | null
+  >(
+    (data) => {
+      if (data === null) {
+        return api.deletePreferences(workflowId);
+      } else {
```
SokratisVidros, line 25:
```suggestion
      }
```

SokratisVidros (approved):

apps/web/src/hooks/workflowPreferences/useUpdateWorkflowPreferences.ts
rifont: I generally prefer to keep an `else` statement when using a `return` in the `if` closure to ensure that removal of the `return` keyword doesn't modify the `else` path.

PR #6580, opened 2024-09-26
fix(worker): Build stateless preferences for bridge-sourced definition

rifont:

* Ensure that old bridge-sourced Workflow definitions without `preferences` can execute statelessly
  * Rather than check for existence of `preferences` before populating the `statelessPreferences`, we instead check for presence of a `bridge.workflow` to indicate whether the execution is stateless, ensuring backward compatiblity for old Framework versions

### Screenshots

_BEFORE - failing test case when `preferences` doesn't exist_
image

_AFTER - passing test case when `preferences` doesn't exist_
image

<details>

</details>

### DIFF
--- a/apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
+++ b/apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
@@ -766,6 +766,38 @@ contexts.forEach((context: Context) => {
       expect(messageBodies).to.include('Hello there 2');
     });
 
+    it(`should deliver message if the Workflow Definition doesn't contain preferences [${context.name}]`, async () => {
+      process.env.IS_WORKFLOW_PREFERENCES_ENABLED = 'true';
+      const workflowId = `without-preferences-workflow-${`${context.name}-${uuidv4()}`}`;
+      const newWorkflow = workflow(workflowId, async ({ step }) => {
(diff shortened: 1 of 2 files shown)


apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
```diff
@@ -766,6 +766,38 @@ contexts.forEach((context: Context) => {
...
+      const workflowId = `without-preferences-workflow-${`${context.name}-${uuidv4()}`}`;
+      const newWorkflow = workflow(workflowId, async ({ step }) => {
+        await step.inApp('send-in-app', () => ({ body: 'Hello there 1' }));
+      });
+
+      /*
+       * Delete `preferences` from the Workflow Definition to simulate an old
+       * Workflow Definition (i.e. from old Framework version) that doesn't have the `preferences` property.
+       */
+      delete newWorkflow.definition.preferences;
```
djabarovgeorge, line 780: we need in because the `workflow` is defined with preferences by default, right?

rifont: Yes, that's correct. E2E suite always uses the latest Framework version which produces a `preferences` property, so we need to delete it manually to test old Framework version that don't produce a `preferences` proeprty.

djabarovgeorge (approved):
amazing work! left a small question regarding the test :)

PR #6583, opened 2024-09-27
feat(web): add usage widget; simplify subscription provider

ChmaraX:
- adds new usage widget / plan info component according to design
- simplifies subscription flow - now `useSubscription` uses just 1 endpoint to get all required data
- fixes billing page bugs, shows correct CTAs based on presence of active subscription rather than presence of payment method 

Fixes PLAT-183, PLAT-180, PLAT-178, PLAT-84

TODO:
- [x] tests

### Screenshots
image
image
image
image


apps/web/src/ee/billing/components/PlanHeader.tsx
```diff
@@ -34,7 +37,7 @@ export const PlanHeader = () => {
   const [apiServiceLevel, setApiServiceLevel] = useState(
     isLoadingSubscriptionData ? subscriptionApiServiceLevel : ApiServiceLevelEnum.FREE
   );
-  const [billingInterval, setBillingInterval] = useState<'month' | 'year'>('month');
+  const [billingInterval, setBillingInterval] = useState<'month' | 'year'>(subscriptionBillingInterval || 'month');
```
ChmaraX, line 42: We need intial `billingInterval` from an active subscription in order to not show "monthly" sub for existing "annual" customers

apps/web/src/ee/billing/components/PlanHeader.tsx
```diff
@@ -136,7 +139,7 @@ export const PlanHeader = () => {
                   </Text>
                 </When>
               </Group>
-              <When truthy={!hasPaymentMethod}>
+              <When truthy={!isActive}>
```
ChmaraX: These checks need to be based on active subscription, otherwise customers which previously bought subscription won't be able to buy again because interval switcher and upgrade button were based on `hasPaymentMethod` which would be true for them.

apps/web/src/ee/billing/components/SubscriptionProvider.tsx
```diff
@@ -1,18 +1,26 @@
...
-const SubscriptionContext = React.createContext<{
-  daysTotal: number;
-  daysLeft: number;
-  isLoading: boolean;
-  isFreeTrialActive: boolean;
-  hasPaymentMethod: boolean;
-}>({
-  daysTotal: 0,
-  daysLeft: 0,
+const SubscriptionContext = React.createContext<UseSubscriptionType>({
```
ChmaraX, line 5: Typed based on shared `/subscription` DTO between API and Web

apps/web/src/ee/billing/hooks/useSubscription.ts
```diff
@@ -3,55 +3,55 @@ import { useMemo } from 'react';
...
-      ? 0
-      : differenceInDays(new Date(subscription.trialEnd), today);
-  }, [subscription.trialEnd]);
+    if (!subscription?.trial.end) return 0;
 
-  const isFreeTrialActive = subscription.status === 'trialing';
+    return isSameDay(new Date(subscription.trial.end), today)
+      ? 0
+      : differenceInDays(new Date(subscription.trial.end), today);
+  }, [subscription.trial.end]);
```
ChmaraX, line 48: This can't be moved to server cause "today" needs to be in clients timezone

apps/web/src/ee/billing/components/SubscriptionProvider.tsx
```diff
@@ -1,18 +1,26 @@
...
   isLoading: false,
-  isFreeTrialActive: false,
-  hasPaymentMethod: false,
+  apiServiceLevel: ApiServiceLevelEnum.FREE,
```
SokratisVidros, line 7: Do we need these default values, or is `null` just simpler to avoid duplicating the API schema in the FE code?

SokratisVidros (approved):

apps/web/src/ee/billing/components/SubscriptionProvider.tsx
ChmaraX: It mostly prevents consumers from having to constantly check if the context value is null or undefined + it can be used also as a fallback

PR #6584, opened 2024-09-27
feat(dashboard): Add inbox

desiprisg:
### Screenshots

<details>

</details>


.prettierrc
```diff
@@ -6,5 +6,6 @@
   "tabWidth": 2,
   "quoteProps": "as-needed",
   "jsxSingleQuote": false,
-  "arrowParens": "always"
+  "arrowParens": "always",
+  "plugins": ["prettier-plugin-tailwindcss"]
```
desiprisg, line 10: Added this here as we were having issues with a new config in the dashboard app

apps/dashboard/src/components/inbox-button.tsx
```diff
@@ -0,0 +1,27 @@
+import { Popover, PopoverContent, PopoverTrigger } from '@/components/primitives/popover';
+import { APP_ID } from '@/config';
+import { Bell, Inbox, InboxContent } from '@novu/react';
+import { PopoverPortal } from '@radix-ui/react-popover';
+
+export const InboxButton = () => {
+  //   const { user } = useUser();
```
desiprisg: Not sure how we will handle dev env vars yet so clerk instance wasn't there.

apps/dashboard/src/components/inbox-button.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  //   if (!user) {
+  //     return null;
+  //   }
+
+  return (
+    <Inbox subscriberId={'1234'} applicationIdentifier={APP_ID}>
```
desiprisg: We will swap this out with the id

apps/dashboard/src/config/index.ts
desiprisg, line 1: Added the `VITE` prefix instead of `REACT_APP` as needed

apps/dashboard/src/components/inbox-button.tsx
SokratisVidros: We will use .env as Vite says. Does this help?

apps/dashboard/src/components/inbox-button.tsx
SokratisVidros: Please convert this to an environment variables and use their presence as a feature flag to render <Inbox/> or not.

SokratisVidros (approved):

LetItRock (approved):

PR #6585, opened 2024-09-27
chore(root): Upgrade to latest NX

SokratisVidros:
Upgrading to latest NX should help us resolve the latest NX issues we have during building all monorepo after introducing the new Dashboard project.


LetItRock (approved):
awesome! 🙌

PR #6590, opened 2024-09-30
fix(root): Build only public packages during preview deployments

SokratisVidros:

Prevent building all Novu-affected packages, including apps. Reduce the possibility of an error and increase the speed of the workflow.


rifont (approved):
🚀

PR #6591, opened 2024-09-30
fix(web): Use initial Workflow Preferences state to determine override state

rifont:

* Use initial Workflow Preferences state to determine override state
* Modify form state population to set default only if it's not set already
  * This is necessary due to the Preferences resource being fetched over API from a child component not present in the main view, we need to ensure the form state is only updated on the Preferences panel component initial render

### Screenshots

https://www.loom.com/share/8b5aa3e3f8a749f48ff6b0d4425bb48d?sid=237293f4-746d-4cda-a46c-0350b62456db

<details>

</details>


apps/web/src/pages/templates/editor_v2/CloudWorkflowSettingsSidePanel.tsx
```diff
@@ -16,10 +16,10 @@ export const CloudWorkflowSettingsSidePanel: FC<CloudWorkflowSettingsSidePanelPr
   const { templateId: workflowId = '' } = useParams<{ templateId: string }>();
   const [searchParams] = useSearchParams();
   const { isLoading, workflowUserPreferences, workflowResourcePreferences } = useCloudWorkflowPreferences(workflowId);
-  const { setValue } = useFormContext<WorkflowDetailFormContext>();
+  const { setValue, getValues } = useFormContext<WorkflowDetailFormContext>();
 
   useEffect(() => {
-    if (workflowUserPreferences !== undefined) {
+    if (workflowUserPreferences !== undefined && getValues('preferences') === undefined) {
```
rifont, line 22: Using the form state to retain context of the form values, only setting the form values if it wasn't set already. This prevents the Preferences component state from being reset to `null` when the API fetch occurs on component render.

apps/web/src/studio/components/workflows/preferences/WorkflowSubscriptionPreferences.tsx
```diff
@@ -47,7 +47,7 @@ export const WorkflowSubscriptionPreferences: FC<WorkflowSubscriptionPreferences
   updateWorkflowPreferences,
   arePreferencesDisabled,
 }) => {
-  const [isOverridingPreferences, setIsOverridingPreferences] = useState(false);
+  const [isOverridingPreferences, setIsOverridingPreferences] = useState(workflowUserPreferences !== null);
```
rifont, line 50: Fix to ensure the initial state for preference override toggle is correct to ensure the useEffect condition that resets the form value isn't fired on initial render.

SokratisVidros (approved):

PR #6595, opened 2024-09-30
feat(dashboard): clerk auth, public and protected routes

LetItRock:

Implemented the Dashboard application auth flow with Clerk.

### Screenshots

https://github.com/user-attachments/assets/fd39e1e1-fd17-4657-90f8-aac0f6cba008


apps/dashboard/package.json
```diff
@@ -22,12 +22,15 @@
     "@radix-ui/react-slot": "^1.1.0",
     "@segment/analytics-next": "^1.73.0",
     "@tanstack/react-query": "^4.20.4",
+    "axios": "^1.6.8",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.1",
+    "js-cookie": "^3.0.5",
```
LetItRock: used to get the session cookie

apps/dashboard/package.json
```diff
@@ -22,12 +22,15 @@
...
     "lucide-react": "^0.439.0",
     "mixpanel-browser": "^2.52.0",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
+    "react-helmet-async": "^1.3.0",
```
LetItRock, line 39: used to set page metadata

apps/dashboard/src/api/api.client.ts
```diff
@@ -0,0 +1,79 @@
+import axios from 'axios';
+import { API_ROOT } from '../config';
+
+export function buildApiHttpClient({
```
LetItRock: copied the HTTP client from the Web app; it allows to dynamically build the api client for the JWT or ApiKey

apps/dashboard/src/components/protected-routes.tsx
```diff
@@ -0,0 +1,22 @@
+import { Navigate, useLocation } from 'react-router-dom';
+import { SignedIn, SignedOut } from '@clerk/clerk-react';
+import { EnvironmentProvider } from '@/context';
+import { ROUTES } from '@/utils/routes';
+
+export const ProtectedRoutes = ({ children }: { children: React.ReactNode }) => {
```
LetItRock: The protected routes component wraps all the private routes. It uses the Clerk components to render a particular page when the user is authenticated or redirects to `/auth/sign-in` page.

apps/dashboard/src/components/protected-routes.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const { pathname } = useLocation();
+  if (pathname === '/') {
+    return <Navigate to={ROUTES.WORKFLOWS} replace />;
+  }
```
LetItRock: on `/` redirect to the `/workflows` page

apps/dashboard/package.json
```diff
@@ -22,12 +22,15 @@
     "@radix-ui/react-slot": "^1.1.0",
     "@segment/analytics-next": "^1.73.0",
     "@tanstack/react-query": "^4.20.4",
+    "axios": "^1.6.8",
```
SokratisVidros: Are we using Axios? Let's just use `fetch` without any dependency.

apps/dashboard/src/components/protected-routes.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+  return (
+    <>
+      <SignedIn>
+        <EnvironmentProvider>{children}</EnvironmentProvider>
```
LetItRock: we only need to fetch environments information when the user is authenticated

apps/dashboard/src/context/auth/auth-provider.tsx
```diff
@@ -0,0 +1,143 @@
...
+import type { UserResource } from '@clerk/types';
+import { ROUTES } from '@/utils/routes';
+import type { AuthContextValue } from './types';
+import { toOrganizationEntity, toUserEntity } from './mappers';
+import { AuthContext } from './auth-context';
+
+const asyncNoop = async () => {};
+const noop = () => {};
+
+export const AuthProvider = ({ children }: { children: ReactNode }) => {
```
LetItRock, line 9: This file was copied from the Web app, but improved slightly with `useUser, useOrganization` hooks and removed the `logout` as it is handled by Clerk.

LetItRock: Some of the other functionality like switching org might be useful later, so I left it as is.

apps/dashboard/src/context/auth/mappers.ts
```diff
@@ -0,0 +1,57 @@
+import type { IOrganizationEntity, IUserEntity } from '@novu/shared';
+import { OrganizationResource, UserResource } from '@clerk/types';
+
+export const toUserEntity = (clerkUser: UserResource): IUserEntity => {
```
LetItRock, line 4: extracted from AuthProvider

apps/dashboard/src/context/auth/auth-provider.tsx
```diff
@@ -0,0 +1,143 @@
+import { ReactNode, useCallback, useEffect, useMemo } from 'react';
```
LetItRock, line 1: Moved all providers under the `/src/context`, tbd.

apps/dashboard/src/context/environment/hooks.ts
```diff
@@ -0,0 +1,66 @@
...
+  const { readOnly, ...rest } = useEnvironmentContext();
+
+  return {
+    ...rest,
+    readOnly: readOnly || bridge || false,
+    bridge: bridge || false,
+  };
+}
+
+export const useFetchEnvironments = ({
```
LetItRock, line 16: the API hooks could be moved under the `/src/hooks/api` or smth

apps/dashboard/src/layout/auth-layout.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  return (
+    <div className="grid h-screen grid-cols-2 gap-8">
+      <div className="grow">Auth Layout</div>
+      <div className="flex items-center justify-center">
+        <Outlet />
+      </div>
+    </div>
+  );
+};
```
LetItRock: This is the public and shared layout for the`/auth/sign-in/*` and `/auth/sign-up/*` pages.

apps/dashboard/src/layout/dashboard-layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+          href="/legacy/integrations"
+          target="_self"
+          className="text-blue-600 visited:text-purple-600 hover:border-b hover:border-current"
+        >
+          Integrations
+        </a>
+        <div className="ml-auto">
+          <UserProfile />
+        </div>
+      </div>
```
LetItRock: this will become the header in the future

apps/dashboard/src/layout/dashboard-layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+      <div className="pt-16">
+        <Outlet />
```
LetItRock: this will be particular dashboard page

apps/dashboard/src/main.tsx
```diff
@@ -1,27 +1,55 @@
...
+          {
+            path: 'sign-up/*',
+            element: <SignUp />,
+          },
+          {
+            path: 'sign-up/organization-list',
+            element: <OrganizationList />,
+          },
+        ],
+      },
```
LetItRock: the public routes: `/auth/sign-in/*` and `/auth/sign-up/*`

apps/dashboard/src/main.tsx
```diff
@@ -1,27 +1,55 @@
...
+          <ProtectedRoutes>
+            <DashboardLayout />
+          </ProtectedRoutes>
+        ),
+        children: [
+          {
+            path: 'workflows',
+            element: <Workflows />,
+          },
+        ],
```
LetItRock: The protected dashboard route `/workflows`.

The important note: because of the way the react-router works and picks the routes for rendering, the protected routes `/*` should be defined at the end of `Root.children` array. 
First, it checks and matches with the public routes. Otherwise, it checks against the protected routes.

apps/dashboard/src/pages/organization-list.tsx
```diff
@@ -0,0 +1,24 @@
...
+          elements: {
+            organizationAvatarUploaderContainer: {
+              display: 'none',
+            },
+          },
+        }}
+        hidePersonal
+        skipInvitationScreen
+        afterSelectOrganizationUrl={ROUTES.WORKFLOWS}
+        afterCreateOrganizationUrl={ROUTES.WORKFLOWS}
```
LetItRock, line 20: Redirect to `/workflows` after org is created/picked, we might change that in the future when will be having the new onboarding flow.

apps/web/package.json
```diff
@@ -11,7 +11,7 @@
     "precommit": "lint-staged",
     "docker:build": "docker buildx build --load -f ./Dockerfile -t novu-web ./../.. $DOCKER_BUILD_ARGUMENTS",
     "docker:build:depot": "depot build -f ./Dockerfile -t novu-web ./../.. --load",
-    "envsetup:dashboard": "chmod +x ./env.sh && ./env.sh .env.dashboard && mv ./env-config.js ./build/env-config.js && rm .env",
+    "envsetup:dashboard": "chmod +x ./env.sh && ./env.sh .env.dashboard && mv ./env-config.js ../../apps/dashboard/legacy/env-config.js && rm .env",
```
LetItRock, line 14: the `env-config.js` file after the web app is built should be moved to the dashboard `/legacy` folder

apps/web/src/ApplicationReadyGuard.tsx
```diff
@@ -64,7 +64,7 @@ export function ApplicationReadyGuard({ children }: PropsWithChildren<{}>) {
     return <Navigate to={ROUTES.AUTH_LOGIN} replace />;
   }
 
-  if (!isOnboardingComplete() && location.pathname !== ROUTES.AUTH_APPLICATION) {
+  if (IS_UNDER_DASHBOARD !== 'true' && !isOnboardingComplete() && location.pathname !== ROUTES.AUTH_APPLICATION) {
```
LetItRock, line 67: We don't want to show the onboarding page for the accounts created with the new Dashboard. cc @SokratisVidros

apps/web/.env.dashboard
desiprisg, line 1: Do we want these committed?

apps/dashboard/src/api/api.client.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+};
+
+export const get = <T>(endpoint: string) => request<T>(endpoint, 'GET');
+
+export const post = <T>(endpoint: string, data: unknown) => request<T>(endpoint, 'POST', data);
+
+export const put = <T>(endpoint: string, data: unknown) => request<T>(endpoint, 'PUT', data);
+
+export const del = <T>(endpoint: string) => request<T>(endpoint, 'DELETE');
```
LetItRock, line 70: the base API functions

apps/dashboard/src/api/environments.ts
```diff
@@ -0,0 +1,14 @@
+import type { IEnvironment } from '@novu/shared';
```
LetItRock, line 1: api helpers

apps/dashboard/src/api/api.client.ts
```diff
@@ -0,0 +1,70 @@
...
+    if (!response.ok) {
+      const errorData = await response.json();
+      throw new NovuApiError(`Novu API error`, errorData, response.status);
+    }
+
+    if (response.status === 204) {
+      return {} as T;
+    }
+
+    return await response.json();
```
SokratisVidros, line 52: If not mistaken we need another `type assertion here`

apps/dashboard/src/api/environments.ts
```diff
@@ -0,0 +1,14 @@
...
+import { get } from './api.client';
+
+export async function getEnvironments() {
+  const { data } = await get<{ data: IEnvironment[] }>('/environments');
+
+  return data;
+}
+
+export async function getCurrentEnvironment() {
```
SokratisVidros: This endpoint will be deprecated from the backend so please remove this function. Environment selection should be a pure FE concern.

apps/dashboard/src/components/protected-routes.tsx
```diff
@@ -0,0 +1,22 @@
...
+      </SignedIn>
+      <SignedOut>
+        <Navigate to={ROUTES.AUTH_SIGN_IN} replace />
```
SokratisVidros, line 18: Does this navigate to the old web sign-in for now?

apps/dashboard/src/context/environment/environment-context.tsx
```diff
@@ -0,0 +1,14 @@
+import type { IEnvironment } from '@novu/shared';
```
SokratisVidros, line 1: I'd keep context and provider in the same files to remove the need for another nesting level in the folder structure. That is `/context/environmentContext`, `/context/authContext`, etc...

apps/dashboard/src/context/environment/hooks.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  return {
+    environments,
+    areEnvironmentsInitialLoading,
+    refetchEnvironments,
+  };
+};
+
+export const useFetchCurrentEnvironment = ({
```
SokratisVidros: Ditto

apps/dashboard/src/layout/root.tsx
```diff
@@ -0,0 +1,22 @@
+import { Outlet } from 'react-router-dom';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { AuthProvider, ClerkProvider, SegmentProvider } from '@/context';
+import { HelmetProvider } from 'react-helmet-async';
+
+const queryClient = new QueryClient();
+
+export const Root = () => {
+  return (
+    <QueryClientProvider client={queryClient}>
```
SokratisVidros: Please move these to the root file of the app to have everything in one place. This is not the layout after all; it's all the top-level providers. It's easier to have them in one place in a top level file.

apps/dashboard/src/utils/context.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Nice little utility function from https://github.com/clerk/javascript/blob/main/packages/shared/src/react/hooks/createContextAndHook.ts
+ *
+ * Creates and returns a Context and two hooks that return the context value.
+ * The Context type is derived from the type passed in by the user.
+ * The first hook returned guarantees that the context exists so the returned value is always CtxValue
+ * The second hook makes no guarantees, so the returned value can be CtxValue | undefined
+ *
+ */
+export const createContextAndHook = <CtxVal>(
```
SokratisVidros, line 23: ❤️

apps/dashboard/src/components/protected-routes.tsx
SokratisVidros: You can also use https://clerk.com/docs/components/control/redirect-to-signin#redirect-to-sign-in

apps/dashboard/src/config/index.ts
```diff
@@ -10,6 +10,8 @@ export const CLERK_PUBLISHABLE_KEY = import.meta.env.VITE_CLERK_PUBLISHABLE_KEY
 
 export const APP_ID = import.meta.env.VITE_NOVU_APP_ID || '';
 
-if (IS_EE_AUTH_ENABLED && !CLERK_PUBLISHABLE_KEY) {
+if (!CLERK_PUBLISHABLE_KEY) {
   throw new Error('Missing Clerk Publishable Key');
 }
+
+export const API_ROOT = import.meta.env.VITE_APP_API_URL;
```
SokratisVidros: To be precise:

```suggestion
export const API_HOSTNAME = import.meta.env.VITE_APP_API_HOSTNAME;
```

apps/dashboard/.env.example
```diff
@@ -0,0 +1,6 @@
+VITE_SENTRY_DSN=
+VITE_LAUNCH_DARKLY_CLIENT_SIDE_ID=
+VITE_HUBSPOT_EMBED=
+VITE_APP_API_URL=http://localhost:3000
```
SokratisVidros:
```suggestion
VITE_APP_API_HOSTNAME=http://localhost:3000
```

apps/dashboard/src/context/environment/hooks.ts
```diff
@@ -0,0 +1,69 @@
+import { useQuery } from '@tanstack/react-query';
+import type { IEnvironment } from '@novu/shared';
+import { QueryKeys } from '@/utils/query-keys';
+import { useEnvironmentContext } from './environment-context';
+import { getCurrentEnvironment, getEnvironments } from '@/api/environments';
+
+export function useEnvironment({ bridge }: { bridge?: boolean } = {}) {
```
SokratisVidros: Remove bridge boolean for now until we need it.

apps/dashboard/src/utils/local-storage.ts
```diff
@@ -0,0 +1,9 @@
+const LOCAL_STORAGE_LAST_ENVIRONMENT_ID = 'nv_last_environment_id';
```
SokratisVidros: The filename should be utils/environment.ts

apps/dashboard/src/utils/query-keys.ts
```diff
@@ -0,0 +1,11 @@
+interface IQueryKeys {
+  myEnvironments: string;
+  currentEnvironment: string;
+  defaultLocale: string;
+}
+
+export const QueryKeys: IQueryKeys = Object.freeze({
+  myEnvironments: 'myEnvironments',
+  currentEnvironment: 'currentEnvironment',
+  defaultLocale: 'defaultLocale',
```
SokratisVidros: Please remove until we need this.

apps/dashboard/src/utils/query-keys.ts
```diff
@@ -0,0 +1,11 @@
+interface IQueryKeys {
```
SokratisVidros: The interface is not necessary. If we need the types we will get the `typeof` QueryKeys using TS inference.

apps/dashboard/src/utils/routes.ts
```diff
@@ -0,0 +1,6 @@
+export const ROUTES = {
+  AUTH_SIGN_IN: '/auth/sign-in',
```
SokratisVidros:
```suggestion
  AUTH_SIGN_IN: '/sign-in',
```

apps/dashboard/src/utils/routes.ts
```diff
@@ -0,0 +1,6 @@
...
+  AUTH_SIGN_UP: '/auth/sign-up',
```
SokratisVidros:
```suggestion
  AUTH_SIGN_UP: '/sign-up',
```

apps/dashboard/src/utils/routes.ts
```diff
@@ -0,0 +1,6 @@
...
+  AUTH_SIGNUP_ORGANIZATION_LIST: '/auth/sign-up/organization-list',
```
SokratisVidros:
```suggestion
  AUTH_SIGNUP_ORGANIZATION_LIST: '/sign-up/organization-list',
```

apps/web/src/ApplicationReadyGuard.tsx
SokratisVidros: Correct. Good catch, we can leave it for now although we will get back to polishing the sign-in/sign-up flow between the two apps before the first request.

SokratisVidros (approved):

apps/web/.env.dashboard
LetItRock: yes, as this file is picked by the script when we are building web app for the dashboard

apps/dashboard/src/api/api.client.ts
LetItRock: not needed because `response.json()` type is `any` which is assigned to `T`

apps/dashboard/src/context/environment/environment-context.tsx
LetItRock: Yes, I did the same initially but got the eslint error saying: `Fast refresh only works when a file only exports components. Use a new file to share constants or functions between components.`

That's why I moved it to a separate "context" file, but I'm still seeing the same error because of this `{ EnvironmentContext, useEnvironmentContext };` but we don't care much about fast refresh in the "context" file as the code will be constant.

apps/dashboard/src/api/environments.ts
LetItRock: ahh I saw somewhere your comment saying to use the `/environments/me` endpoint instead of these calculations, but I can change that.

SokratisVidros: Please send me the link. I might be missing something.

apps/dashboard/src/layout/root.tsx
LetItRock: This is what they suggest doing in the latest react-router, basically to have the `main.tsx` entry file and define the Root component for your shared code and outlet, [ref](https://reactrouter.com/en/main/start/tutorial).
The thing is that the new `RouterProvider` doesn't take children as a prop, so we can't go with the suggestion of having App and providers defined there as some of them are dependent on the router.

SokratisVidros: You can ignore my comment. In the suggestion they use the name `routes` that I think it's better than layout that speaks about pure UI concerns.

apps/dashboard/src/components/protected-routes.tsx
LetItRock: ok, I'll take a look, probably we will need to teach the ClerkProvider about these URLs

apps/dashboard/src/api/environments.ts
LetItRock: https://github.com/novuhq/novu/blob/7b6e0e276c7b734ab19b09a2a63e8e5106e5ea13/apps/web/src/components/providers/EnvironmentProvider.tsx#L45

apps/dashboard/src/layout/root.tsx
LetItRock: The Root component, yes, it doesn't have anything layout-related, but the `AuthLayout, DashBoardLayout` they do. But I think we can use routes and separately define layout components.

PR #6596, opened 2024-09-30
chore(root): Prevent multi-level imports for `@novu/*` packages

rifont:

* Prevent multi-level imports for `@novu/*` packages
  * Fix some rules and ignore EE issues and CSS imports

### Screenshots

image

<details>

</details>

### DIFF
--- a/apps/api/src/app/testing/billing/create-usage-records.e2e-ee.ts
+++ b/apps/api/src/app/testing/billing/create-usage-records.e2e-ee.ts
@@ -3,6 +3,7 @@ import { Logger } from '@nestjs/common';
 import sinon from 'sinon';
 import { expect } from 'chai';
 import { ApiServiceLevelEnum } from '@novu/shared';
+// eslint-disable-next-line no-restricted-imports
 import { StripeBillingIntervalEnum, StripeUsageTypeEnum } from '@novu/ee-billing/src/stripe/types';
 
 const mockMonthlyBusinessSubscription = {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -43,6 +43,7 @@ const noRestrictedImportsMultiLevelNovuPattern = {
   group: [
     '@novu/*/**/*',
     // These packages have legitimate exports 1 path part below the root level
+    // This flatmap logic ignores the path 1 below the root level and prevents deeper imports.
     ...['framework', 'js', 'novui'].flatMap((pkg) => [`!@novu/${pkg}/**/*`, `@novu/${pkg}/*/**/*`]),
   ],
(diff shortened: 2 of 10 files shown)


apps/api/src/app/testing/billing/create-usage-records.e2e-ee.ts
```diff
@@ -3,6 +3,7 @@ import { Logger } from '@nestjs/common';
 import sinon from 'sinon';
 import { expect } from 'chai';
 import { ApiServiceLevelEnum } from '@novu/shared';
+// eslint-disable-next-line no-restricted-imports
```
rifont, line 6: I didn't want to raise an EE PR for these, so they are ignored for now.

eslint.config.mjs
```diff
@@ -29,6 +29,26 @@ import { FlatCompat } from '@eslint/eslintrc';
...
 
+/**
+ * REUSED RULE CONFIGURATIONS
+ *
+ * This is necessary because Eslint doesn't merge rule configurations
+ * when they are targeting different paths.
+ */
+
+/**
+ * This rule ensures that "multi-level" imports are not used for `@novu/*` packages.
```
djabarovgeorge, line 40: I'm wondering if we should approach this from a different direction, specifically by not allowing the `@novu` domain libraries to have the `/src` suffix.  `@novu/**/src` For example: `@novu/framework/src`, `@novu/dal/src`, `@novu/ee/3rd-party/something/src`, `@novu/novui/src`. I believe that, for now, this approach will be more beneficial and won't limit us in potential future implementations.
let me know what you think :)

SokratisVidros (approved):

eslint.config.mjs
```diff
@@ -29,6 +29,26 @@ import { FlatCompat } from '@eslint/eslintrc';
...
+ */
+const noRestrictedImportsMultiLevelNovuPattern = {
+  group: [
+    '@novu/*/**/*',
+    // These packages have legitimate exports 1 path part below the root level
+    // This flatmap logic ignores the path 1 below the root level and prevents deeper imports.
```
rifont:
```suggestion
    // This flatMap logic ignores the path 1 below the root level and prevents deeper imports.
```

eslint.config.mjs
rifont: For packages under our control, I'm of the opinion that we should only allow imports from known `exports` values in the `package.json`. This benefits us by:
* keeping module imports clean
* creating a clear module boundary
* ensuring that public package consumers can import the same things that we do, allowing us to dogfood

This PR introduces a pattern for exceptions of the rule, namely for multiple entry-point packages. That same pattern can be extended to handle other exceptions we might like to add in the future. I think the benefits of a "by exception" approach outweigh the cons of allowing any multi-level imports right now.

SokratisVidros (approved):

PR #6600, opened 2024-10-01
feat(js): Com 229 update the in app preview component in the web app to

BiswaViraj:

- created `novu/js/internal` package which exports the markdown parsing function
- imported the markdown parsing function from novu/js/internal into the dashboard to support bold in-app previews
### Screenshots

image

<details>

</details>


apps/web/src/components/workflow/preview/in-app/v2/InboxPreviewContent.tsx
```diff
@@ -22,6 +23,31 @@ export const INBOX_TOKENS = {
   'Inbox/margin/message/txt/buttons': '1rem',
 } as const;
 
+const parseText = (text?: string) => {
```
SokratisVidros: Nit-picky comment, we are rendering the markdown text, not parsing it ;)

```suggestion
const renderText = (text?: string) => {
```

packages/js/src/ui/internal/parseMarkdown.tsx
```diff
@@ -0,0 +1,36 @@
+export interface Token {
```
SokratisVidros, line 1: Is Token a standard terminology in the Markdown realm?

SokratisVidros (approved):

packages/js/src/ui/internal/parseMarkdown.tsx
BiswaViraj: Yes, it seems so. [ref](https://github.com/markedjs/marked/blob/master/src/Tokens.ts).

PR #6601, opened 2024-10-01
feat(api): Add api exception mapping

tatarco:
We are having issues caused by a very basic error handling in our nest.js app, causing 500 error to not be logged in a good way, the solution is described in the loom here 
https://www.loom.com/share/e6d13666ce94441c9316e9f13bd94472?sid=17c07829-447c-40ac-a9bd-9e01279729ad

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -24,6 +24,7 @@ const TEST_TAGS = ['test'];
 let session: UserSession;
 
 const SCHEMA_WITH_TEXT: JsonSchema = {
+  $schema: 'http://json-schema.org/draft-07/schema#',
```
tatarco: `  
`

apps/api/src/bootstrap.ts
```diff
@@ -121,7 +122,8 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
   await setupSwagger(app);
 
   Logger.log('BOOTSTRAPPED SUCCESSFULLY');
-
+  app.useGlobalFilters(new AllExceptionsFilter(app.get(PinoLogger)));
+  Logger.log('Filter Applied');
```
SokratisVidros: Can we move this comment into the constructor of the exception filter? I guess it's instantiated once on bootstrapping. That way we will have everything contained.

apps/api/src/bootstrap.ts
```diff
@@ -121,7 +122,8 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
   await setupSwagger(app);
 
   Logger.log('BOOTSTRAPPED SUCCESSFULLY');
-
+  app.useGlobalFilters(new AllExceptionsFilter(app.get(PinoLogger)));
```
SokratisVidros, line 103: @tatarco The PR looks great! I expected to add this but remove the current handling of exceptions and erroneous response bodies. Where are we generating response 5xx and 4xx responses today? 

Do we need to shortcircuit that or remove the code alltogether?

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    const { status, message } = this.getResponseMetadata(exception);
+    const responseBOdy = this.buildResponseBody(status, request, message, exception);
+
+    response.status(status).json(responseBOdy);
+  }
+
+  private buildResponseBody(status: number, request: Request, message: string | object | Object, exception: unknown) {
+    let responseBody = this.buildBaseResponseBody(status, request, message);
+    if (status === HttpStatus.INTERNAL_SERVER_ERROR) {
```
SokratisVidros: The response body should be constructed for all 5xx errors. 

```suggestion
    if (status >= HttpStatus.INTERNAL_SERVER_ERROR) {
```

SokratisVidros (approved):

apps/api/src/bootstrap.ts
```diff
@@ -121,7 +122,8 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
...
+  Logger.log('Filter Applied');
   if (expressApp) {
```
rifont: The successful bootstrap log should be the last one we see to indicate the app has fully started up - if filter application fails, we could be mislead that the application started when it failed right at the end.

✔️ chore: Move these 2 newly added lines above the bootstrap message

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,87 @@
...
+    const response = ctx.getResponse<Response>();
+    const request = ctx.getRequest<Request>();
+
+    const { status, message } = this.getResponseMetadata(exception);
+    const responseBOdy = this.buildResponseBody(status, request, message, exception);
+
+    response.status(status).json(responseBOdy);
+  }
+
+  private buildResponseBody(status: number, request: Request, message: string | object | Object, exception: unknown) {
```
rifont: ☑ **chore:** Please add a response type to all methods‏

apps/api/src/bootstrap.ts
```diff
@@ -121,7 +122,8 @@ export async function bootstrap(expressApp?): Promise<INestApplication> {
   await setupSwagger(app);
```
rifont, line 101: ❓ **question:** Please see [line 93](https://github.com/novuhq/novu/pull/6601/files#diff-11933634457bf1519b686a8747a356d1c21a23aa19c22984b3da6c875919c6c3R93) above, we already have Sentry instrumentation to handle exceptions. We should make sure that we're handling exceptions consistently with a single entry point.

Related: it seems that the latest version of `@sentry/node` removed the `Handlers` export, [see here for more details on migrating](https://github.com/getsentry/sentry-javascript/issues/12007). We should explore this upgrade, it's possible that the newer versions of Sentry have built-in APIs for exception mapping.‏

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,87 @@
...
+    };
+  }
+}
+function generateUUID(): string {
+  // Generate a random 4-byte hex string
+  const randomHex = () => randomBytes(2).toString('hex');
+
+  // Construct the UUID using the random hex values
+  return `${randomHex()}${randomHex()}-${randomHex()}-${randomHex()}-${randomHex()}-${randomHex()}${randomHex()}${randomHex()}`;
+}
```
rifont: ❓ **question:** `node:crypto` has a [`randomUUID`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)‏ function as of Node v19, is there a specific reason for rolling our own UUID generator? If there is, please add a code comment explaining why. Otherwise we can replace with `randomUUID`.

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,87 @@
+import { ArgumentsHost, Catch, ExceptionFilter, HttpException, HttpStatus } from '@nestjs/common';
+import { Response } from 'express';
+import { randomBytes } from 'crypto';
```
rifont: ☑ **chore:** Please use the `node:` import directive for built-in modules to avoid the risk of colliding with modules in `node_modules`. It also helps to draw attention to the fact that a built-in module is being used. You can read more about this directive [here](https://2ality.com/2021/12/node-protocol-imports.html).‏
```suggestion
import { randomBytes } from 'node:crypto';
```

rifont (commented):
The direction is a valid one, let's verify firstly if Sentry can provide exception mappers that provide the necessary functionality and proceed from there.

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,87 @@
...
+    response.status(status).json(responseBOdy);
+  }
+
+  private buildResponseBody(status: number, request: Request, message: string | object | Object, exception: unknown) {
+    let responseBody = this.buildBaseResponseBody(status, request, message);
+    if (status === HttpStatus.INTERNAL_SERVER_ERROR) {
+      const uuid = generateUUID();
+      this.pinoLogger.error(`[${uuid}] Service thrown an unexpected exception: `, formatError(exception));
+
+      return { ...responseBody, errorId: uuid };
```
rifont: As discussed:

We verified that Sentry reports all errors when `SENTRY_DSN` environment variable is supplied.

See if we can use the Sentry ID if it exists (see reference error in [Sentry here](https://novu-r9.sentry.io/issues/5945392524/events/5e1de89d6eeb4250b092781b8beadb2e/)) [Cloud Edition], otherwise generate a UUID to support Community Edition.

apps/api/src/bootstrap.ts
tatarco: removed

apps/api/src/bootstrap.ts
tatarco: Richard and I did some research, this does not damage the existing functionality, I will attempt to use the sentry ID if possible, currently we are using the default simplistic node.js error mapping which is just mapping 500 for unexpected exceptions and map other apiError's with their constructor values.

apps/api/src/bootstrap.ts
tatarco: for context: we have synced and tested the change with sentry, the decision is to attempt to match the sentry error id with the uuid if exist (for cloud) besides that this functionality works great and does not damage the sentry functionality

apps/api/src/exception-filter.ts
tatarco: there was some git secutiry issue with it, i'll change

apps/api/src/app/testing/product-feature.e2e.ts
```diff
@@ -27,6 +27,6 @@ describe('Product feature Test @skip-in-ee', async () => {
 
   it('should return a 402 response when required api service level does not exists on organization for feature', async () => {
     const { body } = await session.testAgent.get(path).set('authorization', `ApiKey ${session.apiKey}`).expect(402);
-    expect(body).to.deep.equal({ statusCode: 402, message: 'Payment Required' });
+    expect(body.statusCode).to.equal(402);
```
djabarovgeorge, line 30: Can we bring back the `message` expect just in case it will break?

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export class AllExceptionsFilter implements ExceptionFilter {
+  constructor(private readonly pinoLogger: PinoLogger) {}
+  catch(exception: unknown, host: ArgumentsHost) {
+    const ctx = host.switchToHttp();
+    const response = ctx.getResponse<Response>();
+    const request = ctx.getRequest<Request>();
+
+    const { status, message } = this.getResponseMetadata(exception);
+    const responseBOdy = this.buildResponseBody(status, request, message, exception);
```
djabarovgeorge:
```suggestion
    const responseBody = this.buildResponseBody(status, request, message, exception);
```

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,93 @@
...
+  private buildResponseBody(
+    status: number,
+    request: Request,
+    message: string | object | Object,
+    exception: unknown
+  ): ErrorDto {
+    let responseBody = this.buildBaseResponseBody(status, request, message);
+    if (status === HttpStatus.INTERNAL_SERVER_ERROR) {
+      const uuid = getUuid(exception);
+      this.pinoLogger.error(`[${uuid}] Service thrown an unexpected exception: `, formatError(exception));
```
djabarovgeorge: Mentioned below that maybe it will be beneficial to change the error type and params order.

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,93 @@
...
+function formatError(error: unknown): string {
+  if (error instanceof Error) {
+    return `
+      Error: ${error.message}
+      Stack: ${error.stack || 'No stack trace available'}
+    `;
+  } else {
+    return `
+      Unknown error: ${JSON.stringify(error, null, 2)}
+    `;
```
djabarovgeorge: If we create an object and pass it as the first param in the logger error function `Logger.error(errorObject, message, context)` it will be parsed as separate params in New Relic see picture below. what do you think?

image


```suggestion
    return {
      Error: ${error.message}
      Stack: ${error.stack || 'No stack trace available'}
    };
  } else {
    return {
    Error:
      Unknown error: ${JSON.stringify(error, null, 2)}
    };
```

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function formatError(error: unknown): string {
+  if (error instanceof Error) {
+    return `
+      Error: ${error.message}
+      Stack: ${error.stack || 'No stack trace available'}
+    `;
+  } else {
+    return `
+      Unknown error: ${JSON.stringify(error, null, 2)}
```
djabarovgeorge: Should we guard ourselves here from 3rd party potential errors with circular reference?

apps/api/src/exception-filter.ts
```diff
@@ -0,0 +1,93 @@
...
+    `;
+  }
+}
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export interface ErrorDto {
+  statusCode: number;
+  timestamp: string;
+  errorId?: string;
```
djabarovgeorge, line 97: Let's add a comment explaining what the `errorId` is, so future maintainers can understand its purpose and why it's optional.

apps/ws/src/app.module.ts
```diff
@@ -1,22 +1,20 @@
...
-import { HealthModule } from './health/health.module';
-import { SocketModule } from './socket/socket.module';
+import { AppController } from "./app.controller";
+import { AppService } from "./app.service";
+import { SharedModule } from "./shared/shared.module";
+import { HealthModule } from "./health/health.module";
+import { SocketModule } from "./socket/socket.module";
 
-import packageJson from '../package.json';
+import packageJson from "../package.json";
```
djabarovgeorge: small question, was the single quotation to double quotation intentional?

djabarovgeorge (commented):
amazing work!! i left couple of comments for your review.

apps/ws/src/app.module.ts
tatarco: it's this missmatch betweeen preetier to eslint, I'll revert

tatarco: its preetier?

apps/api/src/exception-filter.ts
tatarco: done

apps/api/src/exception-filter.ts
tatarco: i'll add, could be usefull

apps/api/src/exception-filter.ts
tatarco: i'm trying to make sure it's working locally good, if so i'll commit it

apps/api/src/exception-filter.ts
tatarco: i don't understand what you want here

SokratisVidros (approved):

PR #6608, opened 2024-10-03
feat(dashboard): side navigation

LetItRock:
The Dashboard Side Navigation component:
- clerk organization dropdown
- environment dropdown - without the logic of changing env, and url
- navigation links that point to the legacy web app

### Screenshots

https://github.com/user-attachments/assets/6ec4b6c0-f068-4685-9eb7-19e767384552


apps/dashboard/src/components/primitives/select.tsx
```diff
@@ -0,0 +1,145 @@
+import * as React from 'react';
```
LetItRock, line 1: autogenerated from shadcn, haven't updated the designs

apps/dashboard/src/components/side-navigation/constants.ts
```diff
@@ -0,0 +1,84 @@
...
+  RiPaintBrushLine,
+  RiRouteFill,
+  RiSettings4Line,
+  RiStore3Line,
+  RiUserAddLine,
+} from 'react-icons/ri';
+import { NavItemsGroup } from './types';
+import { LEGACY_ROUTES, ROUTES } from '@/utils/routes';
+
+export const navitationItems: NavItemsGroup[] = [
```
LetItRock: the navigation groups - sections with links

apps/dashboard/src/components/side-navigation/organization-dropdown.tsx
```diff
@@ -0,0 +1,25 @@
...
+    },
+    organizationPreviewMainIdentifier: 'text-foreground text-base',
+    organizationPreviewAvatarContainer: 'size-6 rounded-full',
+    organizationPreviewAvatarBox: 'rounded-full size-6',
+    organizationPreview: 'py-1.5 px-3',
+    organizationSwitcherPopoverActionButton: 'py-3 px-3',
+    organizationSwitcherPopoverCard: 'w-64',
+    organizationSwitcherPreviewButton: 'p-0 [&>svg]:mr-2',
+  },
+};
```
LetItRock: just updated the dropdown styles to something that looks a little better, but we need to have better designs maybe

apps/dashboard/src/components/show.tsx
```diff
@@ -0,0 +1,5 @@
+import { ReactNode } from 'react';
+
+export const Show = ({ when, fallback, children }: { when: boolean; fallback?: ReactNode; children: ReactNode }) => {
```
SokratisVidros: Nit picky comment: I always prefer ternary JSX instead of this component. There is always an extra import without a significant DX benefit.

apps/dashboard/src/components/side-navigation/constants.ts
```diff
@@ -0,0 +1,84 @@
...
+        label: 'Branding',
+        icon: RiPaintBrushLine,
+        to: LEGACY_ROUTES.BRANDING,
+        isExternal: true,
+      },
+      {
+        label: 'Settings',
+        icon: RiSettings4Line,
+        to: LEGACY_ROUTES.SETTINGS,
+        isExternal: true,
```
SokratisVidros, line 72: Lovely modelling. 

Since we call the constants `LegacyRoute` should we maybe call this flag isLegacy as well? Ignore this comment if `external` is used for cross origin redirections.

apps/dashboard/src/components/side-navigation/environment-dropdown.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+});
+
+type EnvironmentDropdownProps = {
+  value?: string;
+  data?: string[];
+  onChange?: (value: string) => void;
+};
+
+export const EnvironmentDropdown = ({ value, data, onChange }: EnvironmentDropdownProps) => {
```
SokratisVidros, line 23: 😍

SokratisVidros (approved):

apps/dashboard/tailwind.config.js
```diff
@@ -84,6 +84,9 @@ export default {
       stable: {
         DEFAULT: 'hsl(var(--stable))',
       },
+      stroke: {
```
desiprisg: Can we put this in `border` ?

apps/dashboard/src/components/show.tsx
LetItRock: yeah, we agreed to proceed with ternary as it's better also working with TS

apps/dashboard/src/components/side-navigation/constants.ts
LetItRock: it will be for the docs

desiprisg (approved):

PR #6610, opened 2024-10-03
fix(web): subscription with no events included

ChmaraX:
This handles a case where user doesn't have any metered subscription active and `includedEvents` is `null`.

### DIFF
--- a/apps/web/src/ee/billing/components/billingV2/ActivePlanBanner.tsx
+++ b/apps/web/src/ee/billing/components/billingV2/ActivePlanBanner.tsx
@@ -60,7 +60,7 @@ const PlanInfo = ({ apiServiceLevel, currentEvents, maxEvents }) => {
       <div className={styles.eventsUsage}>
         <div className={styles.eventsCount}>
           <Text className={styles.eventsNumber} style={{ color }}>
-            {currentEvents.toLocaleString()}
+            {currentEvents?.toLocaleString()}
           </Text>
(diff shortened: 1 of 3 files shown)


apps/web/src/ee/billing/hooks/useSubscription.ts
```diff
@@ -36,6 +36,16 @@ export const useSubscription = () => {
           daysTotal: 0,
         },
       },
+      select: (data) => {
+        return {
+          ...data,
+          events: {
+            ...data.events,
+            // if included is null, customer doesn't have a valid metered subcription, default to 0
+            included: data.events.included ?? 0,
```
ChmaraX: When implement steps to ensure that all users have an active subscription, then we can remove this.

apps/web/src/ee/billing/components/billingV2/ActivePlanBanner.tsx
```diff
@@ -60,7 +60,7 @@ const PlanInfo = ({ apiServiceLevel, currentEvents, maxEvents }) => {
       <div className={styles.eventsUsage}>
         <div className={styles.eventsCount}>
           <Text className={styles.eventsNumber} style={{ color }}>
-            {currentEvents.toLocaleString()}
+            {currentEvents?.toLocaleString()}
```
SokratisVidros, line 63: We need a fall back text value here. Such as `n/a`.

SokratisVidros (approved):

PR #6625, opened 2024-10-04
feat(root): This introduces telemetry in the OS Build

merrcury:
This fixes INF-437. This pushes telemetry to a mixpanel proxy. 

### Screenshots

This is how it will look on Mixpanel
Screenshot 2024-10-04 at 5 01 33 PM

<details>

</details>


apps/worker/src/app/telemetry/usecases/machineInfoService.usecase.ts
```diff
@@ -0,0 +1,39 @@
...
+    await sendDataToMixpanel(this.httpService, eventName, telemetryData);
+  }
+
+  async onApplicationBootstrap() {
+    if (process.env.IS_SELF_HOSTED === 'true') {
+      this.machineId = loadOrCreateMachineId();
+      await this.sendMachineTelemetry('Initial Setup - [OS Telemetry]');
+    } else {
+      Logger.log('Skipping this');
+      Logger;
```
SokratisVidros: Is this necessary? If not please remove.

SokratisVidros (approved):

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
```diff
@@ -0,0 +1,22 @@
+import { HttpService } from '@nestjs/axios';
+import { Logger } from '@nestjs/common';
+import { firstValueFrom } from 'rxjs';
+
+export async function sendDataToMixpanel(httpService: HttpService, event: string, properties: any) {
```
SokratisVidros:
```suggestion
export async function sendDataToNovuTrace(httpService: HttpService, event: string, properties: any) {
```

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    const dataToSend = {
+      event,
+      properties: {
+        ...properties,
+        timestamp: new Date().toISOString(),
+      },
+    };
+
+    const res = await firstValueFrom(httpService.post(process.env.OS_TELEMETRY_URL as string, dataToSend));
+    Logger.debug(`Response from Mixpanel: ${JSON.stringify(res.data)}`);
```
SokratisVidros:
```suggestion
    Logger.debug(`Response from trace.novu.sh: ${JSON.stringify(res.data)}`);
```

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    Logger.debug(`'${event}' event sent to Mixpanel successfully.`);
```
SokratisVidros:
```suggestion
    Logger.debug(`'${event}' event sent to trace.novu.sh successfully.`);
```

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  } catch (error) {
+    Logger.error(`Error sending '${event}' event to Mixpanel:`, error);
```
SokratisVidros:
```suggestion
    Logger.error(`Error sending '${event}' event to trace.novu.sh:`, error);
```

apps/worker/src/app/telemetry/usecases/machineInfoService.usecase.ts
merrcury: I will remove this

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
merrcury: I did not wanted to give away the domain name. So people dont start spamming thats was the whole reason I used it from env variable instead of the hardcoding. I'll remove this log altogether

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
merrcury: I did not wanted to give away the domain name. So people dont start spamming thats was the whole reason I used it from env variable instead of the hardcoding. I'll remove this log altogether

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
merrcury: I did not wanted to give away the domain name. So people dont start spamming thats was the whole reason I used it from env variable instead of the hardcoding. I'll remove this log altogether

apps/worker/src/app/telemetry/utils/sendDataToMixpanel.utils.ts
SokratisVidros: You can use Novu Trace or remove it.

merrcury: I'll thats better

merrcury:
@SokratisVidros I've made the changes

SokratisVidros (approved):

libs/dal/src/repositories/user/user.repository.ts
```diff
@@ -88,4 +88,8 @@ export class UserRepository implements IUserRepository {
   bulkWrite(bulkOperations: any, ordered: boolean): Promise<any> {
     return this.userRepository.bulkWrite(bulkOperations, ordered);
   }
+
+  estimatedDocumentCount(): Promise<number> {
+    return this.userRepository.estimatedDocumentCount();
+  }
```
djabarovgeorge, line 94: We already have the BaseRepository.estimatedDocumentCount(), i think we could use it instead of overriding it on child classes. what do you think about it?

apps/worker/src/app/telemetry/usecases/userInfoService.usecase.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  @Cron(CronExpression.EVERY_DAY_AT_MIDNIGHT)
+  async sendDailyUserTelemetry() {
+    const isSelfHosted = process.env.IS_SELF_HOSTED === 'true';
+    const telemetryEnabled = process.env.NOVU_TELEMETRY !== 'false';
+
+    if (isSelfHosted && telemetryEnabled) {
+      const userData = await this.getUserData();
+      await sendDataToNovuTrace(this.httpService, 'User Info - [OS Telemetry]', userData);
+    }
```
djabarovgeorge, line 62: I know that this PR is a telemetry metric for the OS, well I'm not exactly sure how we currently send our Novu Hosted metrics to Mixpanel. I wonder, though, why not send Novu Hosted and OS metrics the same way? 

That way, we could have all the metrics centralized in one place. 
Let me know what you think about it.

djabarovgeorge (commented):
Amazing work!
Sending love to the new mertics 🥳

libs/dal/src/repositories/user/user.repository.ts
merrcury: Yes, that actually make sense. I added this as place holder if I want to make change in future also to bring parity. Surely using directly from BaseRepository is a good. Let me change that in next iteration.

apps/worker/src/app/telemetry/usecases/userInfoService.usecase.ts
merrcury: Currently, Managed Novu Metrics goes through both segment and to mixpanel directly using the analytics service in our code. 

Why did we choose this route?

1. We did not wanted to expose our segment or mixpanel token in open source docker to get exploited.
2. Currently this will go through a proxy built on top of cloudflare. We will place a WAF on top of it which will stop its exploitation as well and this will let change the provider seamlessly. Currently, Its set to mixpanel. Let say if in future we plan to switch to PostHog we would not have to think about lost metrics with open source installation.

PR #6626, opened 2024-10-04
feat(js, react, shared): user agents

BiswaViraj:

- added novu-user-agent header
### Screenshots

image
image

<details>

</details>

### DIFF
--- a/packages/js/src/types.ts
+++ b/packages/js/src/types.ts
@@ -179,4 +179,8 @@ export type NovuOptions = {
   backendUrl?: string;
   socketUrl?: string;
   useCache?: boolean;
+  /**
+   * @internal Should be used internally
+   */
+  __userAgent?: string;
 };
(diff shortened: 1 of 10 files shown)


packages/js/src/types.ts
```diff
@@ -179,4 +179,8 @@ export type NovuOptions = {
   backendUrl?: string;
   socketUrl?: string;
   useCache?: boolean;
+  /**
+   * @internal Should be used internally
+   */
+  __userAgent?: string;
```
BiswaViraj, line 185: couldn't find a better way to make it not expose to the public
Let me know if there's a better approach.
This `__userAgent` will appear in the autocomplete list if someone tries to initialise the Novu class from the @novu/js package:

```js
 const novu = new Novu()
 ```

SokratisVidros: The only other alternative is not to add a property at all and add a ts ignore when you use it ;)

SokratisVidros (approved):

PR #6632, opened 2024-10-04
feat(dashboard): env change logic and routes

LetItRock:
Switch environments logic and env-based URL routing.

### Screenshots

https://github.com/user-attachments/assets/4d3a22be-52fc-4df9-a392-e58a572cc0cb

https://github.com/user-attachments/assets/55115bca-2cc4-4272-94d5-448676581933

https://github.com/user-attachments/assets/2721a6bf-2d1a-48f1-97fb-162468888a9c

https://github.com/user-attachments/assets/44ba5f15-68e5-415c-8fda-41c3341b08c1

https://github.com/user-attachments/assets/832c2f40-76b1-4b70-a964-be80c14e88b7


apps/dashboard/src/components/side-navigation/constants.ts
```diff
@@ -8,79 +8,86 @@ import {
...
-      {
-        label: 'Invite teammates',
-        icon: RiUserAddLine,
-        to: LEGACY_ROUTES.INVITE_TEAM_MEMBERS,
-        isExternal: true,
-      },
-    ],
-  },
-];
+export const buildNavigationItems = ({
```
LetItRock, line 15: we need to dynamically build nav items because the workflows link depends on the current environment

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +37,49 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
...
+      ) {
+        navigate(buildRoute(ROUTES.WORKFLOWS, { environmentId: selectedEnvironment._id }));
+        return;
+      } else if (pathname.includes(ROUTES.ENV) && selectedEnvironment) {
+        const newPath = pathname.replace(/\/env\/[^/]+(\/|$)/, `${ROUTES.ENV}/${selectedEnvironment._id}$1`);
+        navigate(newPath);
+        return;
+      }
+
+      navigate(pathname, { replace: true });
```
LetItRock: the navigation logic based on the current pathname handles the cases when the route is:
- `/`
- `/env`
- `/env/`
- `/env/:environmentId`
- `/env/:environmentId/`
- `/env/:environmentId/workflows`
- `/env/:environmentId/anything`

For the cases when the `environmentId` is present in the URL, it will replace it with the new environment id

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +37,49 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
...
-    // TODO: in the next PR
-  }, []);
+  useLayoutEffect(() => {
+    if (!environments) {
+      return;
+    }
+
+    const selectedEnvironment = selectEnvironment(environments, getEnvironmentId());
+    switchEnvironmentInternal(environments, selectedEnvironment?._id);
+  }, [environments, switchEnvironmentInternal]);
```
LetItRock: when the environment is loaded, or when the component is mounted and we have environments in the query cache --> set the current environment and redirect the user

apps/dashboard/src/utils/routes.ts
```diff
@@ -3,7 +3,14 @@ export const ROUTES = {
...
   ROOT: '/',
-  WORKFLOWS: '/workflows',
+  ENV: '/env',
+  WORKFLOWS: '/env/:environmentId/workflows',
+};
+
+export const buildRoute = (route: string, params: Record<string, string>) => {
+  return Object.entries(params).reduce((acc, [key, value]) => {
+    return acc.replace(`:${key}`, value);
+  }, route);
```
LetItRock, line 13: the helper to build the params based urls like: `/env/:environmentId/workflows`

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +37,49 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
 
 export function EnvironmentProvider({ children }: { children: React.ReactNode }) {
   const { currentOrganization } = useAuth();
+  const navigate = useNavigate();
+  const { pathname } = useLocation();
```
LetItRock, line 36: TODO: the env in the URL should drive the selection of the current environment

apps/dashboard/src/context/environment/environment-provider.tsx
LetItRock: TODO: probably the last `navigate` is not needed

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +32,51 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
...
+  const { environmentId: paramsEnvironmentId } = useParams<{ environmentId?: string }>();
   const [currentEnvironment, setCurrentEnvironment] = useState<IEnvironment | null>(null);
 
+  const switchEnvironmentInternal = useCallback(
+    (allEnvironments: IEnvironment[], environmentId?: string | null) => {
+      const selectedEnvironment = selectEnvironment(allEnvironments, environmentId);
+      setCurrentEnvironment(selectedEnvironment);
```
LetItRock, line 43: select the env based on the URL:
- if there is no env ID in the URL it will default to the development environment
- otherwise, it will select related env

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +32,51 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
...
+      const newEnvironmentId = selectedEnvironment._id;
+      const isNewEnvironmentDifferent = paramsEnvironmentId !== selectedEnvironment._id;
+
+      if (pathname === ROUTES.ROOT || pathname === ROUTES.ENV || pathname === `${ROUTES.ENV}/`) {
+        navigate(buildRoute(ROUTES.WORKFLOWS, { environmentId: newEnvironmentId }));
```
LetItRock, line 48: if URL is: `/` or `/env` or `/env/` then redirect to the `/env/:environmentId/workflows`

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +32,51 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
...
+      } else if (pathname.includes(ROUTES.ENV) && isNewEnvironmentDifferent) {
+        const newPath = pathname.replace(/\/env\/[^/]+(\/|$)/, `${ROUTES.ENV}/${newEnvironmentId}$1`);
+        navigate(newPath);
```
LetItRock, line 51: if URL includes `/env` and env id in the URL is different from the "selected" (from all envs), then update the URL

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +32,51 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
 
 export function EnvironmentProvider({ children }: { children: React.ReactNode }) {
   const { currentOrganization } = useAuth();
+  const navigate = useNavigate();
+  const { pathname } = useLocation();
+  const { environmentId: paramsEnvironmentId } = useParams<{ environmentId?: string }>();
```
LetItRock, line 37: Info: we will update the `environmentId` later to `slug-hash` when it will be available

apps/dashboard/src/components/side-navigation/constants.ts
```diff
@@ -8,79 +8,86 @@ import {
...
+  currentEnvironment,
+}: {
+  currentEnvironment?: IEnvironment | null;
```
SokratisVidros: Null shouldn't be allowed as a type. This method must be called only when the env is set. This will also simplify all the checks in the method.

SokratisVidros: To ensure that, we can do exactly what we did in the current `web` app, introduce a Guard component that doesn't allow rendering of the layout until Clerk + Env are resolved.

apps/dashboard/src/context/environment/environment-provider.tsx
```diff
@@ -35,18 +32,51 @@ function selectEnvironment(environments?: IEnvironment[], selectedEnvironmentId?
...
   const [currentEnvironment, setCurrentEnvironment] = useState<IEnvironment | null>(null);
 
+  const switchEnvironmentInternal = useCallback(
+    (allEnvironments: IEnvironment[], environmentId?: string | null) => {
+      const selectedEnvironment = selectEnvironment(allEnvironments, environmentId);
+      setCurrentEnvironment(selectedEnvironment);
+      const newEnvironmentId = selectedEnvironment._id;
+      const isNewEnvironmentDifferent = paramsEnvironmentId !== selectedEnvironment._id;
```
SokratisVidros, line 45: Wouldn't the environment be always different on switching?

apps/dashboard/src/context/environment/environment-provider.tsx
SokratisVidros: I am not sure if this will always work. If the URL contains more ids, for example, the workflow slug or the step slug, switching across environments will return a 404 errors as the workflow or step id won't match the current environment. 

My suggestion is to always go to `/env/:environmentId/workflows` after switching environment.

apps/dashboard/src/utils/routes.ts
```diff
@@ -3,7 +3,14 @@ export const ROUTES = {
...
   SIGNUP_ORGANIZATION_LIST: '/sign-up/organization-list',
   ROOT: '/',
-  WORKFLOWS: '/workflows',
+  ENV: '/env',
+  WORKFLOWS: '/env/:environmentId/workflows',
+};
+
+export const buildRoute = (route: string, params: Record<string, string>) => {
+  return Object.entries(params).reduce((acc, [key, value]) => {
+    return acc.replace(`:${key}`, value);
```
SokratisVidros, line 12: You can just use a map in a single iteration.

SokratisVidros (changes requested):

apps/dashboard/src/components/side-navigation/constants.ts
LetItRock: @SokratisVidros, IMO, it creates a bad UX, especially since it will wait for two requests to complete. It would be better to render some loading indicator/skeleton or smth; I mentioned this here:  https://novu.slack.com/archives/C07HVLG7G2U/p1728060743365279

apps/dashboard/src/context/environment/environment-provider.tsx
LetItRock: @SokratisVidros no, this is for case when someone will type the wrong `environmentId` or just typed `asdf`

apps/dashboard/src/context/environment/environment-provider.tsx
LetItRock: @SokratisVidros it works with all these cases:

Screenshot 2024-10-07 at 18 12 12

---

IMO it's important to keep the user on the exact same page where he was, otherwise, it will be a frustrating experience, especially if the user is somewhere two levels deep. And this logic is intended to do that.

If we want to cover the case when the workflow or step doesn't exist, I think that we should have the redirect logic in the corresponding places where we make requests, and when it fails redirect one level down, until we go to the root route. 

For example, if the user received from someone the wrong URL (or switched the env) `/env/:environmentId/workflows/:workflowId/steps/:stepId`
We can have the logic that:
- load `:stepId` when it fails redirect to `/env/:environmentId/workflows/:workflowId`
- load `:workflowId` when it fails redirect to `/env/:environmentId/workflows`
- `:environmentId` doesn't exist - switch to develop

This is like a waterfall effect, but IMO, it will provide better UX to cover similar cases as the application "will try to do best" to render the closest page to the context.

SokratisVidros: The `workflowId` and `stepId` will always be different between environments. Let's take it offline tomorrow.

apps/dashboard/src/components/side-navigation/constants.ts
SokratisVidros: I agree 100% with the need for a splash/loading screen. That's exactly what the guard does in `web`. We "server-side" render the loader via the public.html and then it is removed only when env is resolved. If not mistaken we are saying the same thing. See [here](https://github.com/novuhq/novu/blob/4408daa2c02c938efa4ca71a6e5e9fa6ede69f03/apps/web/src/ApplicationReadyGuard.tsx#L31)

LetItRock: > If not mistaken we are saying the same thing

Not really, what I'm saying is to render the application shell while environments are loading: header, side nav, and in the main content, show the loading indicator. We can think about how it should look design-wise. I want to avoid empty screens, loading, blinking, etc., we can do better.

SokratisVidros: I am aligned with that. The key is not to make the app interactive until Clerk + Env is loaded.

SokratisVidros: The UX about the environment is different from allowing the `null` type. I still suggest to remove the `null` bit and prevent /env/undefined route generation.

SokratisVidros (approved):

apps/dashboard/src/components/side-navigation/constants.ts
LetItRock: @SokratisVidros, we will solve it with UI state for the side nav while the env's are loading. Otherwise, I have to do the `if` statement and return nothing, so there will be a blinking experience

PR #6634, opened 2024-10-04
fix(framework): Default to health action

SokratisVidros:
Visiting `/api/novu` should return the health action response instead of an error.

### Screenshots
Screenshot 2024-09-02 at 15 06 36


djabarovgeorge (approved):
🚀

jainpawan21 (approved):

PR #6635, opened 2024-10-05
fix(client): ensuring that the response code is correct for proper value in hasLoadedSuccessfully key

12ya:
ensuring that the response code is correct for proper value in hasLoadedSuccessfully key


scopsy:
@12ya please submit an issue before a pull request 🙏

PR #6643, opened 2024-10-07
fix(api): Log bridge url validation errors

SokratisVidros:
Debug Bridge endpoint connectivity issues.

### DIFF
--- a/apps/api/src/app/bridge/usecases/get-bridge-status/get-bridge-status.usecase.ts
+++ b/apps/api/src/app/bridge/usecases/get-bridge-status/get-bridge-status.usecase.ts
@@ -1,9 +1,11 @@
-import { BadRequestException, Injectable } from '@nestjs/common';
+import { BadRequestException, Logger, Injectable } from '@nestjs/common';
 import axios from 'axios';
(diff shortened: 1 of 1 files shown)


scopsy (approved):

apps/api/src/app/bridge/usecases/get-bridge-status/get-bridge-status.usecase.ts
```diff
@@ -18,6 +20,11 @@ export class GetBridgeStatus {
 
       return response.data;
     } catch (err: any) {
+      Logger.error(
+        `Failed to verify Bridge endpoint: ${(err as Error).message || err}`,
+        (err as Error).stack,
+        LOG_CONTEXT
+      );
```
rifont: I suggest we include the Bridge URL in the error message to assist in isolating & debugging error messages.

```suggestion
      Logger.error(
        `Failed to verify Bridge endpoint ${command.bridgeUrl} with error: ${(err as Error).message || err}`,
        (err as Error).stack,
        LOG_CONTEXT
      );
```

rifont (approved):

PR #6645, opened 2024-10-07
feat(dashboard): Workflow list polish

desiprisg:

* Loading state skeletons
* Fill up remaining screen height
* Pagination corresponding to correct API data
* show tooltip on name hover when there is not enough space
* differentiate code-first workflows with the corresponding icon
* Add status badge from API
### Screenshots

image

<details>

</details>


SokratisVidros (approved):

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -21,29 +22,54 @@ import {
...
+  const setLimit = useCallback(
+    (limit: number) =>
+      setSearchParams((searchParams) => {
+        searchParams.set('limit', limit.toString());
+        return searchParams;
+      }),
+    [setSearchParams]
+  );
+  const offset = useMemo(() => parseInt(searchParams.get('offset') || '0'), [searchParams]);
+  const limit = useMemo(() => parseInt(searchParams.get('limit') || '12'), [searchParams]);
```
LetItRock: we probably don't need to memoize these

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -55,25 +81,59 @@ export const WorkflowList = () => {
...
           ) : (
             <>
               {workflowsQuery.data.workflows.map((workflow) => (
-                <TableRow>
+                <TableRow key={workflow._id}>
                   <TableCell className="font-medium">
-                    <p>{workflow.name}</p>
-                    <p className="text-foreground-400 font-code text-xs">{workflow._id}</p>
+                    <div className="flex items-center gap-1">
+                      {workflow.origin === WorkflowOriginEnum.EXTERNAL && (
```
LetItRock, line 99: I have checked my framework workflows and we should base this logic on the `type: 'BRIDGE'` instead:
Screenshot 2024-10-08 at 23 20 32

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -55,25 +81,59 @@ export const WorkflowList = () => {
...
+            </>
           ) : (
             <>
               {workflowsQuery.data.workflows.map((workflow) => (
-                <TableRow>
+                <TableRow key={workflow._id}>
                   <TableCell className="font-medium">
-                    <p>{workflow.name}</p>
-                    <p className="text-foreground-400 font-code text-xs">{workflow._id}</p>
+                    <div className="flex items-center gap-1">
```
LetItRock, line 98: nice to have: the cols allow to overflow, we can try to avoid it if easy to do

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -83,46 +143,117 @@ export const WorkflowList = () => {
...
+                    Page {currentPage} of {Math.ceil(workflowsQuery.data.totalCount / limit)}
+                  </span>
+                ) : (
+                  <Skeleton className="h-5 w-[20ch]" />
+                )}
+                {workflowsQuery.data ? (
+                  <Pagination>
+                    <PaginationContent>
+                      <PaginationItem>
+                        <PaginationStart href="#" onClick={() => setOffset(0)} />
```
LetItRock: is there a reason why we need "#"?

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -83,46 +143,117 @@ export const WorkflowList = () => {
...
-                <Select>
+                {workflowsQuery.data ? (
+                  <span className="text-foreground-600 block text-sm font-normal">
+                    Page {currentPage} of {Math.ceil(workflowsQuery.data.totalCount / limit)}
+                  </span>
+                ) : (
+                  <Skeleton className="h-5 w-[20ch]" />
+                )}
+                {workflowsQuery.data ? (
+                  <Pagination>
```
LetItRock: I feel like this part is screaming for a separate component

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -83,46 +143,117 @@ export const WorkflowList = () => {
...
+                      <PaginationItem>
+                        <PaginationPrevious href="#" onClick={() => setOffset(Math.max(0, offset - limit))} />
+                      </PaginationItem>
+                      {(() => {
+                        const currentPage = Math.floor(offset / limit) + 1;
+                        const totalPages = Math.ceil(workflowsQuery.data.totalCount / limit);
+                        const startPage = Math.max(1, currentPage - 2);
+                        const endPage = Math.min(totalPages, currentPage + 2);
+
+                        const pageItems = [];
```
LetItRock: When I have just only one page the pagination buttons are not disabled:

https://github.com/user-attachments/assets/900f595b-3184-4451-b4f1-dd25c1d6c644

apps/dashboard/src/components/workflow-list.tsx
LetItRock: let's not change this for a moment, there is a be bug that we need to solve first and then we will see if this requires the update or not

apps/dashboard/src/components/workflow-list.tsx
desiprisg: I missed this. I plan on making these links instead of using the on click handlers.

LetItRock (approved):

PR #6650, opened 2024-10-08
feat(js): Com 208 improve the dx of the novu on function to return the cleanup

BiswaViraj:

- return cleanup function instead of separate off method
### Screenshots

<details>

</details>


SokratisVidros (approved):
@BiswaViraj Reminder to also update the docs

desiprisg (dismissed):
We need to keep the existing functionality (`off`) as well, since removing it is a breaking change.

SokratisVidros (approved):

PR #6651, opened 2024-10-08
feat(api): refactor get step schema

djabarovgeorge:
### Screenshots

<details>

</details>


apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge, line 1: self note: refactor the workflow cast to any

apps/api/src/app/step-schemas/step-schemas.controller.ts
```diff
@@ -30,16 +34,56 @@ export class StepSchemasController {
...
+    @Query('stepType') stepType: StepType,
+    @Query('workflowId') workflowId: string,
+    @Query('stepId') stepId: string
+  ): Promise<StepSchemaDto> {
+    return await this.getStepDefaultSchemaUsecase.execute(this.createCommand(user, stepType, workflowId, stepId));
+  }
 
-    return { schema };
+  private createCommand(user: UserSessionData, stepType: StepType, workflowId: string, stepId: string) {
+    return stepType
```
djabarovgeorge: self note: change priority it should be first by identifiers (workflowId and stepId) and then fallback to stepType

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
```diff
@@ -0,0 +1,283 @@
...
+    // @ts-ignore
+    process.env.IS_WORKFLOW_PREFERENCES_ENABLED = 'false';
+  });
+
+  describe('Get Control Schema with stepType', () => {
+    it('should get step schema for in app step type', async function () {
+      const { data } = (await session.testAgent.get(`/v1/step-schemas?stepType=${StepTypeEnum.IN_APP}`)).body;
+
+      // in app output schema
+      expect(data.controls.type).to.equal('object');
```
SokratisVidros, line 21: 😓 I guess Chai doesn't have match snapshot.

apps/api/src/app/step-schemas/step-schemas.controller.ts
```diff
@@ -1,45 +1,48 @@
...
+  private createCommand(user: UserSessionData, stepType: StepType, workflowId: string, stepId: string) {
+    return workflowId && stepId
+      ? GetExistingStepSchemaCommand.create({
+          organizationId: user.organizationId,
+          environmentId: user.environmentId,
+          userId: user._id,
+          workflowId,
+          stepId,
+        })
+      : GetStepTypeSchemaCommand.create({
```
SokratisVidros:
```suggestion
      : GetDefaultStepTypeSchemaCommand.create({
```

apps/api/src/app/step-schemas/types/index.ts
```diff
@@ -11,4 +11,22 @@ export const customActionStepSchema = {
   additionalProperties: false,
 } as const satisfies Schema;
 
-export type SchemaOutput = (typeof schemasList)[number] | typeof customActionStepSchema;
+export const mapStepTypeToOutput = {
```
SokratisVidros: 💯

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+    });
+
+    if (!workflow) {
+      throw new BadRequestException(`No workflow found with the given id ${command.workflowId}`);
+    }
+
+    const currentStep = workflow.steps.find((stepItem) => stepItem._id === command.stepId);
+
+    if (!currentStep) {
+      throw new BadRequestException(`No step found with the given id ${command.stepId}`);
```
SokratisVidros:
```suggestion
      throw new BadRequestException(`No step found with the given id ${command.stepId} for workflow ${command.workflowId}`);
```

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+    additionalProperties: false,
+  } as const satisfies Schema;
 };
+
+function buildPreviousStepsSchema(previousSteps: NotificationStepEntity[] | undefined) {
+  type StepUUID = string;
+  let previousStepsProperties: Record<StepUUID, JsonSchema> = {};
+
+  if (previousSteps) {
+    previousStepsProperties = previousSteps.reduce(
```
SokratisVidros: You can remove the if completely and the initialization in line 125.
```suggestion
    previousStepsProperties = (previousSteps || []).reduce(
```

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+    return { currentStep, previousSteps };
   }
 }
 
-export const getStepDefaultOutput = (stepType: StepType): SchemaOutput => {
-  return StepTypeToOutputMap[stepType];
+const isGetByStepType = (command: GetStepSchemaCommand): command is GetStepTypeSchemaCommand =>
+  (command as GetStepTypeSchemaCommand).stepType !== undefined;
+
+const isGetByExistingStep = (command: GetStepSchemaCommand): command is GetExistingStepSchemaCommand =>
```
SokratisVidros:
```suggestion
const isGetByStepId = (command: GetStepSchemaCommand): command is GetExistingStepSchemaCommand =>
```

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.command.ts
```diff
@@ -1,10 +1,23 @@
...
 const StepTypeValue = { ...ChannelStepEnum, ...ActionStepEnum };
 
-export class GetStepSchemaCommand extends EnvironmentWithUserCommand {
+export class GetStepTypeSchemaCommand extends EnvironmentWithUserCommand {
   @IsEnum(StepTypeValue)
+  @IsNotEmpty()
   stepType: StepType;
 }
+
+export class GetExistingStepSchemaCommand extends EnvironmentWithUserCommand {
```
SokratisVidros, line 14: Nitpicky comment: I'd use one command to make all fields optional to avoid an extra [if here](https://github.com/novuhq/novu/pull/6651/files#diff-cbaa2172c5a107d6e34e4ef3c67d70200da6697594a666ce8ff9b2cfb924adffR34). You can safely ignore this comment,

SokratisVidros (approved):

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -0,0 +1,6 @@
+import { Schema } from '@novu/framework';
+
+export type StepSchemaDto = {
+  controls: Schema;
```
tatarco: Maybe we could import JsonSchema directly instead of using this wrapper. Our framework doesn't seem to provide any unique functionality here, so there's no need for this extra dependency. Also, I think the DTO should be placed in the shared directory to make it accessible for the front-end client.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
```diff
@@ -0,0 +1,283 @@
...
+      expect(data.controls.properties).to.have.property('subject');
+      expect(data.controls.properties.subject.type).to.equal('string');
+      expect(data.controls.properties).to.have.property('body');
+      expect(data.controls.properties.body.type).to.equal('string');
```
tatarco, line 25: ￼This seems quite specific. Perhaps we could consider building an object and implementing a deep match instead? It might help improve maintainability in the long run.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
```diff
@@ -0,0 +1,283 @@
...
+        )
+      ).body;
+
+      // in app output schema
+      expect(data.controls.type).to.equal('object');
+      expect(data.controls.properties).to.have.property('subject');
+      expect(data.controls.properties.subject.type).to.equal('string');
+      expect(data.controls.properties).to.have.property('body');
+      expect(data.controls.properties.body.type).to.equal('string');
+      expect(data.controls.properties).to.have.property('avatar');
```
tatarco: same here

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
```diff
@@ -0,0 +1,283 @@
...
+  });
+
+  afterEach(async () => {
+    // @ts-ignore
+    process.env.IS_WORKFLOW_PREFERENCES_ENABLED = 'false';
+  });
+
+  describe('Get Control Schema with stepType', () => {
+    it('should get step schema for in app step type', async function () {
+      const { data } = (await session.testAgent.get(`/v1/step-schemas?stepType=${StepTypeEnum.IN_APP}`)).body;
```
tatarco: Could we extract the actual REST call and centralize it? The repeated URI could complicate maintenance in the future.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
```diff
@@ -0,0 +1,283 @@
...
+      expect(data.controls.required).to.deep.equal(['body']);
+      expect(data.controls.additionalProperties).to.be.false;
+      expect(data.controls.description).to.equal(
+        'Output of the step, including any controls defined in the Bridge App'
+      );
+
+      expect(data.variables.type).to.equal('object');
+      expect(data.variables.description).to.equal(
+        // eslint-disable-next-line max-len
+        'Variables that can be used with Liquid JS Template syntax. Includes subscriber attributes, payload variables, and supports liquid filters for formatting.'
```
tatarco: This seems quite specific, and while it may not break any functionality, it could lead to potential issues down the line.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
```diff
@@ -0,0 +1,283 @@
...
+    });
+
+    it('should get error for invalid workflow id', async function () {
+      const invalidWorkflowId = createdWorkflow.steps[0].stepUuid;
+
+      const res = await session.testAgent.get(
+        `/v1/step-schemas?workflowId=${invalidWorkflowId}&stepId=${createdWorkflow.steps[0].stepUuid}`
+      );
+      expect(res.status).to.equal(400);
+      expect(res.body.message).to.equal(`No workflow found with the given id ${invalidWorkflowId}`);
```
tatarco: We previously agreed that error messages should not be parameterized, and this implementation seems to not meet that standard.

apps/api/src/app/step-schemas/step-schemas.controller.ts
```diff
@@ -1,45 +1,48 @@
...
+    return await this.getStepDefaultSchemaUsecase.execute(this.createCommand(user, stepType, workflowId, stepId));
+  }
 
-    return { schema };
+  private createCommand(user: UserSessionData, stepType: StepType, workflowId: string, stepId: string) {
+    return workflowId && stepId
+      ? GetExistingStepSchemaCommand.create({
+          organizationId: user.organizationId,
+          environmentId: user.environmentId,
+          userId: user._id,
```
tatarco: Why not use the new EnvironmentWithUserObjectCommand? It would reduce the code while maintaining the same functionality.

apps/api/src/app/step-schemas/step-schemas.controller.ts
```diff
@@ -1,45 +1,48 @@
...
+    @Query('workflowId') workflowId: string,
+    @Query('stepId') stepId: string
+  ): Promise<StepSchemaDto> {
+    return await this.getStepDefaultSchemaUsecase.execute(this.createCommand(user, stepType, workflowId, stepId));
+  }
 
-    return { schema };
+  private createCommand(user: UserSessionData, stepType: StepType, workflowId: string, stepId: string) {
+    return workflowId && stepId
+      ? GetExistingStepSchemaCommand.create({
```
tatarco: Maybe we could use a regular `if` statement or encapsulate the command generation? The ternary operator makes the code harder to read.

apps/api/src/app/step-schemas/step-schemas.controller.ts
```diff
@@ -1,45 +1,48 @@
...
-    );
+    @Query('stepType') stepType: StepType,
+    @Query('workflowId') workflowId: string,
+    @Query('stepId') stepId: string
+  ): Promise<StepSchemaDto> {
+    return await this.getStepDefaultSchemaUsecase.execute(this.createCommand(user, stepType, workflowId, stepId));
+  }
 
-    return { schema };
+  private createCommand(user: UserSessionData, stepType: StepType, workflowId: string, stepId: string) {
```
tatarco: Would it make sense to move this logic into the use case class? Keeping the controller simple and free of business logic could enhance readability and maintainability.

apps/api/src/app/step-schemas/types/index.ts
```diff
@@ -11,4 +11,22 @@ export const customActionStepSchema = {
...
+  [ChannelStepEnum.SMS]: channelStepSchemas[ChannelStepEnum.SMS].output,
+  [ChannelStepEnum.EMAIL]: channelStepSchemas[ChannelStepEnum.EMAIL].output,
+  [ChannelStepEnum.PUSH]: channelStepSchemas[ChannelStepEnum.PUSH].output,
```
tatarco: Could we consider having a type file instead of the index, while keeping the index for exports? I find this approach less confusing.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.command.ts
```diff
@@ -1,10 +1,23 @@
...
+  @IsString()
+  @IsNotEmpty()
+  workflowId: string;
+
+  @IsString()
+  @IsNotEmpty()
+  stepId: string;
+}
+
+export type GetStepSchemaCommand = GetStepTypeSchemaCommand | GetExistingStepSchemaCommand;
```
tatarco, line 24: Could we simplify this command to support multiple use cases and have the internal use case handle the matching and business logic?

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+    if (isGetByExistingStep(command)) {
+      const { currentStep, previousSteps } = await this.findSteps(command);
+
+      return {
+        controls: buildControlsSchema(currentStep?.template?.type as StepType),
+        variables: buildVariablesSchema(previousSteps),
+      };
+    }
+
+    throw new BadRequestException('Invalid command');
```
tatarco, line 33: Could we create a new error type that includes all relevant information and issues? The current error message doesn’t provide enough detail for the API user to effectively resolve their problems.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+  }
+
+  private async findSteps(command: GetExistingStepSchemaCommand) {
+    const workflow = await this.notificationTemplateRepository.findByIdQuery({
+      id: command.workflowId,
+      environmentId: command.environmentId,
+    });
+
+    if (!workflow) {
+      throw new BadRequestException(`No workflow found with the given id ${command.workflowId}`);
```
tatarco: Could we create a new error type to eliminate string literals in the code? This would provide a clearer separation of concerns and improve searchability of issues, as it would avoid parameterization in the error messages.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
tatarco: Could we create a new error type that includes all relevant context, such as the step we searched for and the actual steps received? This would provide more clarity for troubleshooting.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+const buildPayloadSchema = () =>
+  ({
     type: 'object',
     properties: {},
     required: [],
     additionalProperties: false,
-  } as const satisfies Schema,
+    description:
+      'Payload Schema - For Developers. Passed during the novu.trigger method, and controlled by the developer.',
+  }) as const satisfies Schema;
```
tatarco: Didn’t we agree that this is irrelevant and not currently used? Shouldn’t we consider removing it completely?

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+    type: 'object',
+    description:
+      // eslint-disable-next-line max-len
+      'Variables that can be used with Liquid JS Template syntax. Includes subscriber attributes, payload variables, and supports liquid filters for formatting.',
+    properties: {
+      subscriber: buildSubscriberSchema(),
+      payload: buildPayloadSchema(),
+      steps: buildPreviousStepsSchema(previousSteps),
+    },
+    required: ['subscriber', 'payload'],
```
tatarco: What’s the purpose of the required field in the schema here? What are we trying to convey to the front end? Since it's a full object, not just a single variable, I’m curious about its necessity.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+
+function buildPreviousStepsSchema(previousSteps: NotificationStepEntity[] | undefined) {
+  type StepUUID = string;
+  let previousStepsProperties: Record<StepUUID, JsonSchema> = {};
+
+  if (previousSteps) {
+    previousStepsProperties = previousSteps.reduce(
+      (acc, step) => {
+        if (step.template?._id) {
+          acc[step.template._id] = mapStepTypeToResult[step.template.type as StepType];
```
tatarco: cool implementation 👯

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -1,39 +1,149 @@
...
+  async execute(command: GetStepSchemaCommand): Promise<StepSchemaDto> {
+    if (isGetByStepType(command)) {
+      return { controls: buildControlsSchema(command.stepType), variables: buildVariablesSchema() };
+    }
+
+    if (isGetByExistingStep(command)) {
+      const { currentStep, previousSteps } = await this.findSteps(command);
+
+      return {
+        controls: buildControlsSchema(currentStep?.template?.type as StepType),
```
tatarco: Didn’t we decide in our original discussion that the controls are not updated without a migration? If we’re dealing with an existing step, shouldn’t we serve the schema that is already persisted in it?

tatarco (approved):
Approved with comments

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge: do you mean deep check? if so it has but I prefer more explicit tests.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: flat it all!  🚀

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: thanks, i had to many renaming there on the params to make it more clearer 😅

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.command.ts
djabarovgeorge: I won't lie, i was not sure about the solution. I structured it this way for two reasons: first, I wanted to pass the validation to `class-validator`, and second, I wanted to make sure each command had a clear interface to keep the code cleaner (in a way).

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
djabarovgeorge: totally agree i was my mismatch, thanks! 🙏

djabarovgeorge: i wanted to make it compatible with framework but we definitely can not over-engineer it at the moment.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge: I do like more specific tests, for example, let's look at the `origin` bug we had today.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge: same here will resolve for now.

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge: done

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge: fair point i updated the test

apps/api/src/app/step-schemas/e2e/get-step-schema.e2e.ts
djabarovgeorge: You are right!!!!! it was a bad habit!

apps/api/src/app/step-schemas/step-schemas.controller.ts
djabarovgeorge: sorry, but i still do not see to much reason for passing all of the user session data.

apps/api/src/app/step-schemas/step-schemas.controller.ts
djabarovgeorge: Let me check if it will be more clear

djabarovgeorge: not sure if not clearer but i do not mind to change

apps/api/src/app/step-schemas/step-schemas.controller.ts
djabarovgeorge: Love the idea of the clean controller. i will move it to the command file as it is its responsibility

apps/api/src/app/step-schemas/types/index.ts
djabarovgeorge: usually i refactor once it becomes messy

djabarovgeorge: done

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.command.ts
djabarovgeorge: yes but i think it will be overkill for the logic we have today

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: i refactored it to a simple message with params.
regarding the new error type, lets make a decision in the team on what solution we want to go with and then do it all together, right now i will keep it as is because i am afraid we will end up with a code base that looks like 1K people maintaining.

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: yeah i added it before we agreed on this, i think it wont be harmful at the moment. nevertheless i will remove it so we will add it once it is needed 👍

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: Totally agree

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: we are providing the step id we searched for, what else can we pass here? there are no steps received in this situation

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
djabarovgeorge: Can you provide more details? Do you think it will affect in a bad way the frontend?

PR #6657, opened 2024-10-09
feat(api): add workflow trigger identifier parity

djabarovgeorge:
Add trigger identifier parity with the v2 Workflows API CRUD.

#### Why? (Context)

Currently, we're missing some functionality that was available in v1, the `identifier` was missing in the DTO meaning the frontend was not able to provide it in the client.
image

This ticket aims to address that gap.

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
djabarovgeorge, line 1: The test "should not allow creating two workflows for the same user with the same name" will fail because we do allow workflows with the same name. 
We need to clarify the terms: `name` is not unique and can be duplicated, while `workflowIdentifier` (based on the workflow name) cannot be duplicated.

Please note, that I’ve seen discussions across the product regarding this topic, but the current logic in the product is as follows:

- On the first workflow creation, say "Password Reset," if the user hasn't provided a custom trigger identifier, we generate a slug from the workflow name.
- On the second creation of a workflow with the same name ("Password Reset"), if the user hasn't provided a custom trigger identifier, we gracefully allow the same name but append a short ID to the slug.

When updating a workflow, we will throw an error if the user provides a trigger identifier that already exists in the system. This is because having two workflows with the same trigger identifier is invalid. We enforce this rule both in the application and the database layer, where MongoDB will throw an exception if we don’t validate it.

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -243,6 +244,10 @@ function buildCreateWorkflowDto(nameSuffix: string): CreateWorkflowDto {
   return {
     __source: WorkflowCreationSourceEnum.EDITOR,
     name: TEST_WORKFLOW_NAME + nameSuffix,
+    triggerIdentifier: `${slugify(TEST_WORKFLOW_NAME + nameSuffix, {
+      lower: true,
+      strict: true,
+    })}`,
```
djabarovgeorge: self note: extract to app-gen and reuse.

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -64,6 +64,10 @@ export class WorkflowCommonsFields {
   @IsDefined()
   name: string;
 
+  @IsString()
+  @IsDefined()
+  triggerIdentifier: string;
```
djabarovgeorge: IMO we can refactor the legacy term and make it simplifier with `identifier` what do you think?

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -184,28 +185,13 @@ export class UpsertWorkflowUseCase {
...
-
-  private async workflowExistByExternalId(upsertCommand: UpsertWorkflowCommand) {
-    const { environmentId } = upsertCommand.user;
-    const workflowByDbId = await this.notificationTemplateRepository.findByTriggerIdentifier(
-      environmentId,
-      upsertCommand.workflowDto.name
-    );
-
-    return !!workflowByDbId;
-  }
```
djabarovgeorge, line 211: we don't need to validate because this logic is handled in the Create/Update workflow application-generic logic. so this call is duplicated.
(small note we enforce it on the database layer as well)

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
```diff
@@ -135,6 +119,32 @@ export class CreateWorkflow {
     return storedWorkflow;
   }
 
+  private generateTriggerIdentifier(command: CreateWorkflowCommand) {
+    if (command.triggerIdentifier) {
+      return command.triggerIdentifier;
+    }
```
djabarovgeorge, line 126: If provided, we will skip the internal generation. This is somewhat of a risk area, which is why I added a comment on the command parameter.

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
```diff
@@ -135,6 +119,32 @@ export class CreateWorkflow {
...
+    else {
+      /**
+       * For non-bridge workflows, we use a slugified version of the workflow name
+       * as the trigger identifier to provide a better trigger DX.
+       */
+      triggerIdentifier = `${slugify(command.name, {
+        lower: true,
+        strict: true,
+      })}`;
+    }
```
rifont, line 141: If a workflow becomes ejected and obtains a `type === WorkflowTypeEnum.BRIDGE`, how do you anticipate the Workflow identifier changing through the ejection process?

As stated above, Bridge workflows assume the identifier is equivalent to the name, if we need to align on this pattern for ejected workflows it could break existing trigger integrations.

packages/shared/src/dto/workflows/workflow-commons-fields.ts
rifont: I agree!

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
djabarovgeorge: synced and validated that there is no issue at the current architecture, solving this one.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -188,28 +189,13 @@ export class UpsertWorkflowUseCase {
       description: workflowDto.description || '',
       tags: workflowDto.tags || [],
       critical: false,
+      triggerIdentifier: `${slugify(workflowDto.name, {
```
SokratisVidros: This introduces a new term called `triggerIdentifier`. Isn't this the same as the `workflowId`? If yes, please rename accordingly.

SokratisVidros: Moreover we need to use the same help `slugifyIdentifier` helper

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -229,6 +215,7 @@ export class UpsertWorkflowUseCase {
       description: workflowDto.description,
       tags: workflowDto.tags,
       active: workflowDto.active ?? true,
+      identifier: workflowDto.identifier,
```
SokratisVidros: `workflowId` or `identifier`?

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
```diff
@@ -136,6 +120,29 @@ export class CreateWorkflow {
...
 
+  private generateTriggerIdentifier(command: CreateWorkflowCommand) {
+    if (command.triggerIdentifier) {
+      return command.triggerIdentifier;
+    }
+
+    let triggerIdentifier: string;
+    if (command.type === WorkflowTypeEnum.BRIDGE)
+      /*
+       * Bridge workflows need to have the identifier preserved to ensure that
```
SokratisVidros, line 131: In code-first, we allow human-friendly workflow identifiers. This contradicts the latest definition of a workflow identifier as defined in the Dashboard as workflowIds defined in code are not necessarily slugs. They can contain spaces, tabs, and uppercase letters. 

My suggestion is that with need to be stricter in @novu/framework when it comes to workflowIds in TS and apply the same rules that will be applied to workflowIds degined in Dashboard.

libs/application-generic/src/utils/slugify-identifier.ts
```diff
@@ -0,0 +1,8 @@
+import slugify from 'slugify';
+
+export const slugifyIdentifier = (identifier: string): string => {
```
SokratisVidros:
```suggestion
export const slugify = (identifier: string): string => {
```

or `slufigyName`.

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -64,6 +64,10 @@ export class WorkflowCommonsFields {
   @IsDefined()
   name: string;
 
+  @IsString()
+  @IsDefined()
+  identifier: string;
```
SokratisVidros: Ditto about introducing yet another identifier field on the workflow collection.

SokratisVidros (changes requested):

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: > This introduces a new term called triggerIdentifier.

this is not true, under the hood CreateWorkflow usecase is responsible for creation of 
workflow.triggers.identifier
where workflow is NotificationTemplateEntity)
and triggers is NotificationTriggerEntity[]

In CreateWorkflow usecase context, i think we should provide it with more explicit terms if possible, until/if we decide to refactor our workflow entity

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: taking into to consideration the new naming it is workflowId, i will update it.

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
djabarovgeorge: If i am not wrong the code here will stay the same, i will create a [ticket](https://linear.app/novu/issue/NV-4495/parity-of-the-creation-of-workflowid-in-framework) about this topic in order to update framework.

libs/application-generic/src/utils/slugify-identifier.ts
djabarovgeorge: Will do slufigyName in order to be more specific and so it won't collide with the lib function

packages/shared/src/dto/workflows/workflow-commons-fields.ts
djabarovgeorge: following the new Workflow Name DX Guide i already updated to workflowId

SokratisVidros (approved):

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -243,6 +245,7 @@ function buildCreateWorkflowDto(nameSuffix: string): CreateWorkflowDto {
   return {
     __source: WorkflowCreationSourceEnum.EDITOR,
     name: TEST_WORKFLOW_NAME + nameSuffix,
+    workflowId: `${slugifyName(TEST_WORKFLOW_NAME + nameSuffix)}`,
```
rifont, line 248: I think it's an antipattern to supply a slugified `workflowId` during workflow creation, that concern should be handled in the API. Am I missing something?

djabarovgeorge: Good point i made it like this because of the following design where the client has validation.

image

PR #6660, opened 2024-10-09
chore(api): Remove Sentry profiling from all apps

SokratisVidros:
The hypothesis is that using Sentry profiler with a rate of 1.0, sends everything to Sentry, stressing the containers and the AWS tasks

### DIFF
--- a/apps/api/src/instrument.ts
+++ b/apps/api/src/instrument.ts
@@ -1,24 +1,11 @@
-import { nodeProfilingIntegration } from '@sentry/profiling-node';
-import { init } from '@sentry/nestjs';
+import { init } from '@sentry/node';
+import { version } from '../package.json';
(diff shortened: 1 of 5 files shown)


ChmaraX (approved):

apps/api/src/instrument.ts
```diff
@@ -1,24 +1,11 @@
-import { nodeProfilingIntegration } from '@sentry/profiling-node';
-import { init } from '@sentry/nestjs';
+import { init } from '@sentry/node';
```
rifont: Why did we change from the `nestjs` import to `node`?

.source
rifont, line 1: Is this submodule change intentional?

rifont (approved):
💯

PR #6662, opened 2024-10-09
fix(worker): Reduce distraction errors

SokratisVidros:

1. Report queue metrics to Newrelic every 30 seconds, not every second.
2. Remove Better Stack heartbeat as it wasn't properly configured on the Better Stack. We will add it again as part of future work.

Screenshot 2024-10-09 at 13 50 49


apps/worker/src/app/workflow/services/active-jobs-metric.service.ts
```diff
@@ -66,7 +65,7 @@ export class ActiveJobsMetricService {
               repeatJobKey: METRIC_JOB_ID,
               repeat: {
                 immediately: true,
-                pattern: '* * * * * *',
+                pattern: '*/30 * * * * *', // Every 30 seconds
```
rifont: We have a `CronExpressionEnum.EVERY_30_SECONDS` defined [here](https://github.com/novuhq/novu/blob/next/packages/shared/src/types/cron/cron.types.ts#L9) in `@novu/shared` that we might like to use.

SokratisVidros: #til. Updated :)

rifont (approved):
🚇

PR #6673, opened 2024-10-10
refactor(api,worker): untagle modules and deps

ChmaraX:
See rnterprise counterpart: https://github.com/novuhq/packages-enterprise/pull/231


apps/api/src/app/shared/shared.module.ts
```diff
@@ -85,7 +74,6 @@ const DAL_MODELS = [
   WorkflowOverrideRepository,
   ControlValuesRepository,
   PreferencesRepository,
-  ...getDynamicAuthProviders(),
```
ChmaraX, line 88: we don't need to import these into `SharedModule` because its already imported in `AuthModule` (which is part of `AppModule` same as `SharedModule`)

apps/worker/package.json
```diff
@@ -89,7 +89,6 @@
     "typescript": "5.6.2"
   },
   "optionalDependencies": {
-    "@novu/ee-auth": "workspace:*",
```
ChmaraX, line 92: Worker doesn't need `@novu/ee-auth` as a dependency anymore (see changes in https://github.com/novuhq/packages-enterprise/pull/231)

apps/worker/src/app/telemetry/usecases/userInfoService.usecase.ts
```diff
@@ -2,22 +2,22 @@ import { HttpService } from '@nestjs/axios';
...
-  UserRepository,
 } from '@novu/dal';
 import { loadOrCreateMachineId } from '../utils/machine.utils';
 import { sendDataToNovuTrace } from '../utils/sendDataToNovuTrace.utils';
 
 @Injectable()
 export class UserInfoService {
   constructor(
-    private readonly userRepository: UserRepository,
-    private readonly organizationRepository: OrganizationRepository,
```
ChmaraX, line 20: This service is used for telemetry, which counts data in local mongodb, therefore we can use community repository versions directly.

`UserRepository` and `OrganizationRepository` are being resolved conditionally based on if the app is in enterprise version or not. In this case we always know we want to use non-enterprise repositories.

enterprise/packages/billing/package.json
```diff
@@ -47,8 +48,5 @@
     "@nestjs/swagger": "7.4.0",
     "@nestjs/throttler": "6.2.1",
     "@novu/dal": "workspace:*"
-  },
-  "optionalDependencies": {
-    "@novu/ee-auth": "workspace:*"
```
ChmaraX, line 52: Billing is always used as an enterprise package therefore it always need `@novu/ee-auth`

enterprise/packages/translation/package.json
```diff
@@ -39,12 +39,8 @@
   },
   "peerDependencies": {
     "@nestjs/common": "10.4.1",
-    "@nestjs/jwt": "10.2.0",
     "@nestjs/swagger": "7.4.0",
     "@nestjs/platform-express": "10.4.1",
     "@novu/dal": "workspace:*"
-  },
-  "optionalDependencies": {
-    "@novu/ee-auth": "workspace:*"
```
ChmaraX, line 48: Translation package doesn't need these dependencies anymore (see https://github.com/novuhq/packages-enterprise/pull/231)

apps/api/src/app/shared/shared.module.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable global-require */
```
ChmaraX: We should not be using `shared.module` at all and rather import what each modules requires one by one.
This creates a situation where we put everything inside shared module and then we just import the shared module everywhere (also in tests) without caring if we really need everything thats in shared causing various issues.

SokratisVidros (approved):

apps/api/src/app/shared/shared.module.ts
ChmaraX: I reverted this part - reason in the comment in code

SokratisVidros (approved):

PR #6676, opened 2024-10-10
feat(web): Request company size during sign-up

SokratisVidros:
https://linear.app/novu/issue/PLAT-196/signup-add-company-size-dropdown-to-organization-creation

### Screenshots

Screenshot 2024-10-11 at 00 52 22

### DIFF
--- a/apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
+++ b/apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
@@ -29,6 +29,8 @@ function updateClerkOrgMetadata(data: UpdateExternalOrganizationDto) {
   return api.post('/v1/clerk/organization', data);
 }
 
+const companySizeOptions = ['<20', '20-50', '51-100', '101-200', '200+'];
+
 export function QuestionnaireForm() {
   const { isSupported } = useWebContainerSupported();
   const isPlaygroundOnboardingEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_PLAYGROUND_ONBOARDING_ENABLED);
@@ -45,10 +47,6 @@ export function QuestionnaireForm() {
   const location = useLocation();
   const [_, setParams] = useSearchParams();
 
-  const { mutateAsync: updateOrganizationMutation } = useMutation<{ _id: string }, IResponseError, any>(
-    (data: UpdateExternalOrganizationDto) => updateClerkOrgMetadata(data)
-  );
-
   async function updateOrganization(data: UpdateExternalOrganizationDto) {
     const selectedLanguages = Object.keys(data.language || {}).filter((key) => data.language && data.language[key]);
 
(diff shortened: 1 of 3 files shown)


apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
```diff
@@ -45,19 +47,16 @@ export function QuestionnaireForm() {
   const location = useLocation();
   const [_, setParams] = useSearchParams();
 
-  const { mutateAsync: updateOrganizationMutation } = useMutation<{ _id: string }, IResponseError, any>(
```
SokratisVidros, line 48: @ChmaraX the `useMutation` hook felt unnecessary at this point. Could you please confirm that's correct?

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
```diff
@@ -144,16 +145,34 @@ export function QuestionnaireForm() {
     }
   `;
 
-  const trackRedirectionToOnboarding = () => {
```
SokratisVidros, line 147: This was unused, so it was removed.

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
```diff
@@ -200,7 +220,7 @@ export function QuestionnaireForm() {
               label="Choose your back-end stack"
               styles={inputStyles}
               error={fieldState.error?.message}
-              mt={32}
+              mt={16}
```
SokratisVidros, line 223: Some styling fixes.

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
ChmaraX: Yes I think you are correct - using mutation doesn't do anything here.

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
```diff
@@ -45,19 +47,16 @@ export function QuestionnaireForm() {
...
     const selectedLanguages = Object.keys(data.language || {}).filter((key) => data.language && data.language[key]);
 
     const updateClerkOrgDto: UpdateExternalOrganizationDto = {
       jobTitle: data.jobTitle,
       domain: data.domain,
       language: selectedLanguages,
+      companySize: data.companySize,
     };
-    await updateOrganizationMutation(updateClerkOrgDto);
+    await updateClerkOrgMetadata(updateClerkOrgDto);
```
ChmaraX, line 59: We need to also update these 2 things:
- backend endpoint logic and DTOs, ref: https://github.com/novuhq/packages-enterprise/blob/next/auth/src/controllers/clerk.metadata.controller.ts#L27
- Clerk public metadata type so that its typed, ref: https://github.com/novuhq/novu/blob/next/packages/shared/src/types/clerk/types.ts
  - btw here you can also remove `defaultLocale` and `apiServiceLevel` from the type since we dont store that in Clerk

ChmaraX (changes requested):

apps/web/src/ee/clerk/components/QuestionnaireForm.tsx
SokratisVidros: Done. PTAL

ChmaraX (approved):

PR #6678, opened 2024-10-11
ci(root): remove redundant API unit test

ChmaraX:
On PR we are running this workflow: https://github.com/novuhq/novu/blob/next/.github/workflows/on-pr.yml
which already has a step called `test_unit_services` which runs API unit tests.

We don't need another 2x runs of API unit tests inside e2e test workflow (2x runs because the e2e runs twice for enterprise and regular API)/

### DIFF
--- a/.github/workflows/on-pr.yml
+++ b/.github/workflows/on-pr.yml
@@ -244,6 +244,7 @@ jobs:
       test-e2e-affected: ${{ contains(fromJson(needs.get-affected.outputs.test-e2e), '@novu/api') || contains(fromJson(needs.get-affected.outputs.test-e2e), '@novu/worker') }}
       test-e2e-ee-affected: ${{ contains(fromJson(needs.get-affected.outputs.test-e2e-ee), '@novu/api') || contains(fromJson(needs.get-affected.outputs.test-e2e-ee), '@novu/worker') }}
       job-name: ${{ matrix.name }}
+      test-unit: false
     secrets: inherit    
 
   test_e2e_web:
--- a/.github/workflows/reusable-api-e2e.yml
+++ b/.github/workflows/reusable-api-e2e.yml
@@ -116,10 +116,6 @@ jobs:
       - name: Kill port for worker 1342 for unit tests
         run: sudo kill -9 $(sudo lsof -t -i:1342)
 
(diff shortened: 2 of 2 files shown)


rifont (commented):
This is in the right direction.

However, the [on-pr workflow](https://github.com/novuhq/novu/blob/next/.github/workflows/on-pr.yml#L7) only runs on PR, whilst [reusable-api-e2e](https://github.com/novuhq/novu/pull/6678/files#diff-8241a3fda8f4b433572021646c90464b447515d09d65c8ee2e429afbed83a92e) runs on merge comits in `next` too. This change will save time on PR checks, but it has the effect of removing API unit tests from being run on merges to `next`.

Given that we don't require branches to be up to date with `next` before merging, our current strategy is to run all the unit tests on `next` to ensure that changes introduced from PRs not up to date with `next` are compatible with what is in `next`.

In summary, we need to ensure that both Unit and E2E tests run on all commits to `next`. One approach would be to extract the unit tests from `on-pr` and run them as a `reusable-api-unit-tests` PR.

rifont (approved):
💯

SokratisVidros:
> This is in the right direction.
> 
> However, the [on-pr workflow](https://github.com/novuhq/novu/blob/next/.github/workflows/on-pr.yml#L7) only runs on PR, whilst [reusable-api-e2e](https://github.com/novuhq/novu/pull/6678/files#diff-8241a3fda8f4b433572021646c90464b447515d09d65c8ee2e429afbed83a92e) runs on merge comits in `next` too. This change will save time on PR checks, but it has the effect of removing API unit tests from being run on merges to `next`.
> 
> Given that we don't require branches to be up to date with `next` before merging, our current strategy is to run all the unit tests on `next` to ensure that changes introduced from PRs not up to date with `next` are compatible with what is in `next`.
> 
> In summary, we need to ensure that both Unit and E2E tests run on all commits to `next`. One approach would be to extract the unit tests from `on-pr` and run them as a `reusable-api-unit-tests` PR.

Good call out. The reusable-* actions we have must be more fine-grained and act as discrete building blocks so that we can mix and match for each use case. I've added this overhaul to the infra team roadmap.

PR #6460, opened 2024-09-09
feat(api): wip create post v2 workflow end point

djabarovgeorge:
### What changed? Why was the change needed?
POST /v2/workflows

Creates/update a new V1 Workflow DSL from the supplied V2 Workflow DSL

Adds `origin: 'novu' | 'external'` to Notification Template entity (which decides if during Preview & Execution, I need to call a Novu vs Customer hosted Bridge endpoint)
### Screenshots

<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #6679, opened 2024-10-11
fix(api): Use the MongoId for stepId when fetching step controls

rifont:

* Use the MongoId for stepId when fetching step controls
  * Previously the user-provided `stepId` was used when fetching step controls. A recent [PR](https://github.com/novuhq/novu/pull/6460/files#diff-b09fec7e783cfcbe7c2a8edaf102bf9a95b1c8e58aa36f53c31a0e8d61aeb8ccL96) removed persistence of `stepId`, this caused `null` to always be returned when fetching step controls from the Dashboard, meaning that the default controls were always shown
* Remove redundant `controlSchema` property from `UpsertControlValues` use-case
  * Correspondingly, remove the `json-schema-defaults` package from API which is redundant - `@novu/framework` already generates the defaults
* Remove redundant `workflowId` and `stepId` properties from `ControlValuesEntity`

### Screenshots

https://github.com/user-attachments/assets/4916bc6e-a047-4ea0-9fb9-83f340a270a3

<details>

</details>

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -71,7 +71,6 @@
     "helmet": "^6.0.1",
     "i18next": "^23.7.6",
     "ioredis": "5.3.2",
-    "json-schema-defaults": "^0.4.0",
     "jsonwebtoken": "9.0.0",
     "lodash": "^4.17.15",
     "nanoid": "^3.1.20",
--- a/apps/api/src/app/bridge/bridge.controller.ts
+++ b/apps/api/src/app/bridge/bridge.controller.ts
@@ -174,12 +174,17 @@ export class BridgeController {
     if (!workflowExist) {
       throw new NotFoundException('Workflow not found');
     }
+    const step = workflowExist?.steps.find((item) => item.stepId === stepId);
+
+    if (!step || !step._id) {
+      throw new NotFoundException('Step not found');
+    }
 
     const result = await this.controlValuesRepository.findOne({
       _environmentId: user.environmentId,
       _organizationId: user.organizationId,
       _workflowId: workflowExist._id,
-      stepId,
+      _stepId: step._id,
       level: ControlVariablesLevelEnum.STEP_CONTROLS,
     });
 
--- a/apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
+++ b/apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
@@ -1,10 +1,7 @@
-import { Injectable } from '@nestjs/common';
-import defaults from 'json-schema-defaults';
-
+import { Injectable, NotFoundException } from '@nestjs/common';
 import { NotificationTemplateRepository } from '@novu/dal';
-import { JsonSchema } from '@novu/framework';
 
-import { ApiException, UpsertControlValuesCommand, UpsertControlValuesUseCase } from '@novu/application-generic';
+import { UpsertControlValuesCommand, UpsertControlValuesUseCase } from '@novu/application-generic';
 import { StoreControlVariablesCommand } from './store-control-variables.command';
 
 @Injectable()
@@ -21,19 +18,14 @@ export class StoreControlVariablesUseCase {
     );
 
     if (!workflowExist) {
-      throw new ApiException('Workflow not found');
+      throw new NotFoundException('Workflow not found');
     }
 
     const step = workflowExist?.steps.find((item) => item.stepId === command.stepId);
 
     if (!step || !step._id) {
-      throw new ApiException('Step not found');
+      throw new NotFoundException('Step not found');
     }
-
-    const stepDefaultControls = defaults(
-      (step.template as any)?.controls?.schema || (step.template as any)?.inputs?.schema,
-      {}
-    ) as JsonSchema;
 
     return await this.upsertControlValuesUseCase.execute(
       UpsertControlValuesCommand.create({
@@ -42,7 +34,6 @@ export class StoreControlVariablesUseCase {
         notificationStepEntity: step,
         workflowId: workflowExist._id,
         newControlValues: command.variables,
-        controlSchemas: { schema: stepDefaultControls },
       })
     );
   }
--- a/libs/application-generic/src/usecases/upsert-control-values/index.ts
+++ b/libs/application-generic/src/usecases/upsert-control-values/index.ts
@@ -1,2 +1,2 @@
-export * from './upsert-control-values-command';
-export * from './upsert-control-values-use-case';
+export * from './upsert-control-values.command';
+export * from './upsert-control-values.usecase';
--- a/libs/application-generic/src/usecases/upsert-control-values/upsert-control-values.command.ts
+++ b/libs/application-generic/src/usecases/upsert-control-values/upsert-control-values.command.ts
@@ -1,9 +1,8 @@
 import { IsNotEmpty, IsObject, IsOptional, IsString } from 'class-validator';
-import { JsonSchema } from '@novu/framework';
 import { NotificationStepEntity } from '@novu/dal';
-import { OrganizationLevelCommand } from '../../commands';
+import { EnvironmentCommand } from '../../commands';
 
-export class UpsertControlValuesCommand extends OrganizationLevelCommand {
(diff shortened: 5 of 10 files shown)


libs/dal/src/repositories/control-variables/controlValuesEntity.ts
```diff
@@ -12,6 +12,4 @@ export class ControlValuesEntity {
   controls: Record<string, unknown>;
   _workflowId: string;
   _stepId: string;
-  workflowId: string;
-  stepId: string;
```
rifont, line 16: These are no longer used.

libs/application-generic/src/usecases/upsert-control-values/upsert-control-values.command.ts
```diff
@@ -14,7 +13,4 @@ export class UpsertControlValuesCommand extends OrganizationLevelCommand {
   @IsObject()
   @IsOptional()
   newControlValues?: Record<string, unknown>;
-
-  @IsObject()
-  controlSchemas: { schema: JsonSchema };
```
rifont, line 19: This was never used. Cleaning up.

libs/application-generic/src/usecases/upsert-control-values/upsert-control-values.command.ts
```diff
@@ -1,9 +1,8 @@
 import { IsNotEmpty, IsObject, IsOptional, IsString } from 'class-validator';
-import { JsonSchema } from '@novu/framework';
 import { NotificationStepEntity } from '@novu/dal';
-import { OrganizationLevelCommand } from '../../commands';
+import { EnvironmentCommand } from '../../commands';
 
-export class UpsertControlValuesCommand extends OrganizationLevelCommand {
+export class UpsertControlValuesCommand extends EnvironmentCommand {
```
rifont, line 5: The incorrect base command was used here, which resulted in `environmentId` being an optional type when it should have been required.

libs/application-generic/src/usecases/upsert-control-values/index.ts
```diff
@@ -1,2 +1,2 @@
-export * from './upsert-control-values-command';
-export * from './upsert-control-values-use-case';
+export * from './upsert-control-values.command';
+export * from './upsert-control-values.usecase';
```
rifont, line 2: Fixing to follow naming conventions.

apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
```diff
@@ -21,28 +18,22 @@ export class StoreControlVariablesUseCase {
...
     if (!step || !step._id) {
-      throw new ApiException('Step not found');
+      throw new NotFoundException('Step not found');
     }
 
-    const stepDefaultControls = defaults(
-      (step.template as any)?.controls?.schema || (step.template as any)?.inputs?.schema,
-      {}
-    ) as JsonSchema;
-
```
rifont, line 37: Defaults are generated on `@novu/framework`, there is no need to generate defaults on the API.

apps/api/src/app/bridge/bridge.controller.ts
```diff
@@ -174,12 +174,17 @@ export class BridgeController {
...
+    if (!step || !step._id) {
+      throw new NotFoundException('Step not found');
+    }
 
     const result = await this.controlValuesRepository.findOne({
       _environmentId: user.environmentId,
       _organizationId: user.organizationId,
       _workflowId: workflowExist._id,
-      stepId,
+      _stepId: step._id,
```
rifont, line 187: This change fixes the fetching of persisted step controls.

apps/api/package.json
```diff
@@ -71,7 +71,6 @@
     "helmet": "^6.0.1",
     "i18next": "^23.7.6",
     "ioredis": "5.3.2",
-    "json-schema-defaults": "^0.4.0",
```
rifont, line 74: Not needed.

apps/api/src/app/bridge/bridge.controller.ts
```diff
@@ -174,12 +174,17 @@ export class BridgeController {
     if (!workflowExist) {
       throw new NotFoundException('Workflow not found');
     }
+    const step = workflowExist?.steps.find((item) => item.stepId === stepId);
+
+    if (!step || !step._id) {
+      throw new NotFoundException('Step not found');
+    }
```
djabarovgeorge, line 181: Question: can we do something like this?
I started to use this pattern in [this PR](https://github.com/novuhq/novu/pull/6651/files#diff-b92f8bced0d347f2344f3113183234889caa44ab84492116e7c4edc4e6cc8ef8) this way we have the clean message to search by with the additional metadata (step id) 

```suggestion
    if (!step || !step._id) {
      throw new NotFoundException({
        message: 'Step not found',
        stepId: stepId,
      });
    }
```

apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
```diff
@@ -21,28 +18,22 @@ export class StoreControlVariablesUseCase {
...
     if (!workflowExist) {
-      throw new ApiException('Workflow not found');
+      throw new NotFoundException('Workflow not found');
     }
 
     const step = workflowExist?.steps.find((item) => item.stepId === command.stepId);
 
     if (!step || !step._id) {
-      throw new ApiException('Step not found');
+      throw new NotFoundException('Step not found');
```
djabarovgeorge, line 27: same here.

apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
djabarovgeorge: question just to clarify, for the dashboard, we need to make sure it gets the default values for the code-first solution, so it can display them. If we have a code-first preview, we first get the default values from the bridge app and then override them with control values if they exist (persisted in novu), correct?

apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
```diff
@@ -21,28 +18,22 @@ export class StoreControlVariablesUseCase {
...
-    ) as JsonSchema;
-
     return await this.upsertControlValuesUseCase.execute(
       UpsertControlValuesCommand.create({
         organizationId: command.organizationId,
         environmentId: command.environmentId,
         notificationStepEntity: step,
         workflowId: workflowExist._id,
         newControlValues: command.variables,
-        controlSchemas: { schema: stepDefaultControls },
```
djabarovgeorge, line 45: 🧹 👏

libs/application-generic/src/usecases/upsert-control-values/upsert-control-values.usecase.ts
```diff
@@ -2,15 +2,15 @@ import { Injectable } from '@nestjs/common';
 
 import { ControlValuesEntity, ControlValuesRepository } from '@novu/dal';
 import { ControlVariablesLevelEnum } from '@novu/shared';
-import { UpsertControlValuesCommand } from './upsert-control-values-command';
```
djabarovgeorge, line 5: i am a bit confused should not it be a dot here?

djabarovgeorge (approved):
awesome work, left couple of comments for your consideration

apps/api/src/app/bridge/bridge.controller.ts
rifont: As discussed, we will address this holistically across the API using shared classes.

apps/api/src/app/bridge/usecases/store-control-variables/store-control-variables.usecase.ts
rifont: The JSON schema from `@novu/framework` is passed to the frontend to render the default values, so we don't need to create defaults on the API.

libs/application-generic/src/usecases/upsert-control-values/upsert-control-values.usecase.ts
rifont: The new code shows a dot.

PR #6451, opened 2024-09-06
feat(api): Add Novu-managed Bridge endpoint per environment

rifont:
### What changed? Why was the change needed?

* Add custom Novu-managed Bridge-endpoint endpoint at `/v1/environments/:environmentId/bridge`
  * dynamically constructs a Framework `workflow` derived from the persisted Workflow definition
* Refactor all use-cases making requests to Novu Bridge endpoints to use `ExecuteBridgeRequest`. This provides a central location for:
  * HTTP concerns
  * error handling
  * fetching environment secret key
  * hmac signature creation for bridge request authenticity

### Screenshots

<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #6681, opened 2024-10-11
feat(dashboard): workflow editor route and basic layout

LetItRock:
Workflow Editor route and basic layout

### Screenshots

Screenshot 2024-10-11 at 16 59 07

https://github.com/user-attachments/assets/5c7dc555-656d-44f6-86b7-c55ebad895d4


apps/dashboard/eslint.config.js
```diff
@@ -1,29 +1,36 @@
...
+      '@typescript-eslint/no-explicit-any': 'warn',
+      'react-refresh/only-export-components': ['warn', { allowConstantExport: true }],
+      '@typescript-eslint/no-unused-vars': ['off'],
+      '@typescript-eslint/naming-convention': [
+        'error',
+        {
+          filter: '_',
+          selector: 'variableLike',
+          leadingUnderscore: 'allow',
+          format: ['PascalCase', 'camelCase', 'UPPER_CASE'],
```
LetItRock, line 31: allow using `_` for the unused variables

apps/dashboard/src/components/edit-workflow-layout.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <IntercomProvider appId={INTERCOM_APP_ID}>
+      <div className="relative flex w-full">
+        <div className="flex flex-1 flex-col overflow-y-auto overflow-x-hidden">
+          <HeaderNavigation startItems={headerStartItems} />
+
+          <div className="flex min-h-dvh flex-col overflow-y-auto overflow-x-hidden">{children}</div>
+        </div>
+      </div>
+    </IntercomProvider>
```
LetItRock, line 24: similar layout to the dashboard layout, but we might change here the header navigation and we don't need to show the side navigation

apps/dashboard/src/components/primitives/tabs.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
```
LetItRock, line 1: the tabs component updated according to our designs

apps/dashboard/src/components/workflow-editor.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const { workflow } = useFetchWorkflow({
+    workflowId,
+    onSuccess: (data) => {
+      reset(data);
+    },
+    onError: () => {
+      navigate(buildRoute(ROUTES.WORKFLOWS, { environmentId: currentEnvironment?._id ?? '' }));
+    },
+  });
```
LetItRock, line 45: fetch the workflow and set it in the form state

LetItRock: when BE returns an error, navigate the user to the workflow list page

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -158,7 +160,18 @@ export const WorkflowList = () => {
...
+                          text={workflow.name}
+                          onClick={() => {
+                            navigate(
+                              buildRoute(ROUTES.EDIT_WORKFLOW, {
+                                environmentId: currentEnvironment?._id ?? '',
+                                workflowId: workflow._id,
+                              })
+                            );
+                          }}
+                        />
```
LetItRock, line 174: I wasn't sure if this is what it should be because, from the designs, it's not clear. But decided to do it like when clicking on the workflow name, redirect the user to the workflow editor.

apps/dashboard/src/api/bridge.ts
```diff
@@ -0,0 +1,14 @@
+import { BridgeStatus } from '@/utils/types';
+import { get, post } from './api.client';
+
+export const getBridgeHealthCheck = async () => {
+  const { data } = await get<{ data: BridgeStatus }>('/bridge/status');
```
SokratisVidros: We need to return errors as well. Then in a separate PR we need to wire them to a global error handler.

apps/dashboard/src/components/header-navigation/edit-bridge-url-button.tsx
```diff
@@ -0,0 +1,124 @@
+import { useLayoutEffect, useState } from 'react';
+import { RiLinkM, RiPencilFill } from 'react-icons/ri';
+import { PopoverPortal } from '@radix-ui/react-popover';
```
SokratisVidros: Is this the recommended way to use a PopoverPortal for ShadCN? I'd expect this to come from primitives folder leveraging the ShadCN code downloaded via the CLI.

apps/dashboard/src/components/header-navigation/edit-bridge-url-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+    reset({ bridgeUrl: envBridgeUrl });
+  }, [reset, envBridgeUrl]);
+
+  const onSubmit = async ({ bridgeUrl }: z.infer<typeof formSchema>) => {
+    const { isValid } = await validateBridgeUrl(bridgeUrl);
+    if (isValid) {
+      await updateBridgeUrl({ url: bridgeUrl, environmentId: currentEnvironment?._id ?? '' });
+      setBridgeUrl(bridgeUrl);
+    } else {
+      setError('bridgeUrl', { message: 'The provided URL is not the Novu Endpoint URL' });
```
SokratisVidros: The new API, the team, prepared has discrete error types. So we should use the server response instead of hardcoding a frontend error message. 

See [here](https://github.com/novuhq/novu/pull/6451/files#diff-51c461c9bbd4f2626eeaa09cc7258dfe2f8e5747889cf4bb893c2a31c36a91e6R155)

apps/dashboard/src/components/header-navigation/edit-bridge-url-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+                  render={({ field }) => (
+                    <FormItem>
+                      <FormLabel>Bridge Endpoint URL</FormLabel>
+                      <FormControl>
+                        <InputField variant="xs" state={errors.bridgeUrl?.message ? 'error' : 'default'}>
+                          <RiLinkM className="size-5 min-w-5" />
+                          <Input id="bridgeUrl" {...field} />
+                        </InputField>
+                      </FormControl>
+                      <FormMessage>Full path URL (e.g., https://your.api.com/api/novu)</FormMessage>
```
SokratisVidros: `Full path URL` doesn't make a lot of sense as an expression. Use `URL` or `Set URL`.

```suggestion
                      <FormMessage>URL (e.g., https://your.api.com/api/novu)</FormMessage>
```

apps/dashboard/src/components/header-navigation/edit-bridge-url-button.tsx
```diff
@@ -0,0 +1,124 @@
...
+                />
+              </div>
+              <div className="flex items-center justify-between border-t border-neutral-200 px-5 py-3">
+                <a
+                  href="https://docs.novu.co/concepts/endpoint#bridge-endpoint"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  className="text-xs"
+                >
+                  How it works?
```
SokratisVidros:
```suggestion
                  Learn more
```

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -158,7 +160,18 @@ export const WorkflowList = () => {
...
                           </Badge>
                         )}
-                        <TruncatedText text={workflow.name} />
+                        <TruncatedText
+                          className="cursor-pointer"
+                          text={workflow.name}
+                          onClick={() => {
+                            navigate(
+                              buildRoute(ROUTES.EDIT_WORKFLOW, {
+                                environmentId: currentEnvironment?._id ?? '',
```
SokratisVidros, line 169: This is pure inner DX but also a way to prevent invalid states in the front end, such as a workflow route, without environmentId. 

Can we please update the hook times so that they always return a currentEnvironment when used to avoid this ternary on each current environment?

There are two ways to go about this.

1. The hook with throw an error if used and the environment is not set (fail fast), but the environment return type will always be `IEnvironment`.

2. We can teach the hook to return the `{isLoaded, currentEnvironment}` tuple and use discriminated unions so that if `isLoaded` is true currentEnvironment is always an IEnvironment. Then, in all the components that use `useEnvironment` we can do `if (!isLoaded) { return null }`

apps/dashboard/src/hooks/use-fetch-workflow.ts
```diff
@@ -0,0 +1,35 @@
...
+  workflowId,
+  onSuccess,
+  onError,
+}: {
+  workflowId?: string;
+  onSuccess?: (data: WorkflowResponseDto) => void;
+  onError?: (error: unknown) => void;
+}) => {
+  const { data, isPending, error } = useQuery<WorkflowResponseDto>({
+    queryKey: [QueryKeys.bridgeHealthCheck, workflowId],
```
SokratisVidros: Please add the environment as well in the keys.

apps/dashboard/src/hooks/use-update-bridge-url.ts
```diff
@@ -0,0 +1,16 @@
+import { updateBridgeUrl } from '@/api/environments';
+import { useMutation } from '@tanstack/react-query';
+
+export const useUpdateBridgeUrl = () => {
```
SokratisVidros: Kudos on the naming convention on CRUD hooks

SokratisVidros (approved):

apps/dashboard/src/components/header-navigation/edit-bridge-url-button.tsx
LetItRock: Unfortunately, the cli doesn't generate the portal 🤷‍♂️  but we can re-export it, I'll do it

apps/dashboard/src/components/header-navigation/edit-bridge-url-button.tsx
LetItRock: @SokratisVidros, the `updateBridgeUrl` is calling the `/environments/:environmentId` endpoint. Do I miss something?

apps/dashboard/src/components/workflow-list.tsx
LetItRock: @SokratisVidros, it's not that simple:
1. Throwing an error won't work, because on the initial render the `currentEnvironment` is always undefined, so the `useEnvironment` hook will always throw.
2. That's a partial solution to this problem because all code above the `if (!isLoaded)` including hooks will still need to check whether the current env is present.

SokratisVidros: We will take this offline. It's not a good enough blocker for this PR.

SokratisVidros (approved):

PR #6683, opened 2024-10-13
fix(api): created at identify using ISO date instead of epoch

scopsy:
Mixpanel and segment expect an ISO date to properly function.

### Screenshots

<details>

</details>


djabarovgeorge (approved):

djabarovgeorge (commented):
maybe it will be a good idea to type the `traits` a bit better so the future maintainers will get it in compilation time.

PR #6684, opened 2024-10-13
fix(api): analytics module

scopsy:

Mis import of the analytics module, caused FE events to be dismissed.

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/analytics/analytics.module.ts
+++ b/apps/api/src/app/analytics/analytics.module.ts
@@ -1,12 +1,12 @@
 import { Module } from '@nestjs/common';
-import { AnalyticsService } from '@novu/application-generic';
 import { HttpModule } from '@nestjs/axios';
 import { AnalyticsController } from './analytics.controller';
 import { HubspotIdentifyFormUsecase } from './usecases/hubspot-identify-form/hubspot-identify-form.usecase';
+import { SharedModule } from '../shared/shared.module';
 
 @Module({
-  imports: [HttpModule],
+  imports: [SharedModule, HttpModule],
   controllers: [AnalyticsController],
-  providers: [AnalyticsService, HubspotIdentifyFormUsecase],
+  providers: [HubspotIdentifyFormUsecase],
 })
 export class AnalyticsModule {}


apps/api/src/app/analytics/analytics.module.ts
```diff
@@ -1,12 +1,12 @@
...
 import { AnalyticsController } from './analytics.controller';
 import { HubspotIdentifyFormUsecase } from './usecases/hubspot-identify-form/hubspot-identify-form.usecase';
+import { SharedModule } from '../shared/shared.module';
 
 @Module({
-  imports: [HttpModule],
+  imports: [SharedModule, HttpModule],
   controllers: [AnalyticsController],
-  providers: [AnalyticsService, HubspotIdentifyFormUsecase],
+  providers: [HubspotIdentifyFormUsecase],
```
djabarovgeorge, line 10:
```suggestion
  providers: [analyticsService, HubspotIdentifyFormUsecase],
```

apps/api/src/app/analytics/analytics.module.ts
```diff
@@ -1,12 +1,12 @@
 import { Module } from '@nestjs/common';
-import { AnalyticsService } from '@novu/application-generic';
 import { HttpModule } from '@nestjs/axios';
 import { AnalyticsController } from './analytics.controller';
 import { HubspotIdentifyFormUsecase } from './usecases/hubspot-identify-form/hubspot-identify-form.usecase';
+import { SharedModule } from '../shared/shared.module';
 
 @Module({
-  imports: [HttpModule],
+  imports: [SharedModule, HttpModule],
```
djabarovgeorge, line 8:
```suggestion
  imports: [HttpModule],
```

djabarovgeorge (commented):
importing a custom `analyticsService` provider won't be enough?

apps/api/src/app/analytics/analytics.module.ts
djabarovgeorge: feels like you were missing the process.env.SEGMENT_TOKEN injection? was there something else?

scopsy: It's already imported under the SharedModule, so I think I would love to have a single instance of it

scopsy: It's confusing, but that's how all of our service are exported from app generic

CleanShot 2024-10-13 at 17 41 30@2x

djabarovgeorge (approved):

apps/api/src/app/analytics/analytics.module.ts
djabarovgeorge: got it, i was not sure if it would create 2 instances.

PR #6685, opened 2024-10-14
fix(api): Log exception traits

rifont:

* The current exception logging garbles the error due to the `Error.toString()` method not working as expected. This change explicitly logs out the error traits

### Screenshots

_Before_
image

_After_
image

<details>

</details>


tatarco (approved):
approving to make this go faster, the change was done toghether in sync, should be err: exceprtion

PR #6687, opened 2024-10-14
fix(api): fix getWorkflowBug

tatarco:
https://novu.slack.com/archives/C07HVLG7G2U/p1728652077678029
fix for the bug raised in the following thread 

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
@@ -8,7 +8,7 @@ import {
 } from '@novu/dal';
 import { ControlValuesLevelEnum, WorkflowResponseDto } from '@novu/shared';
 import { GetPreferences, GetPreferencesCommand } from '@novu/application-generic';
-
+import { Error } from 'mongoose';
 import { GetWorkflowCommand } from './get-workflow.command';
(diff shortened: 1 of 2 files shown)


apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
```diff
@@ -40,6 +38,20 @@ export class GetWorkflowUseCase {
...
+    try {
+      return await this.notificationTemplateRepository.findByIdQuery({
+        id: command._workflowId,
+        environmentId: command.user.environmentId,
+      });
+    } catch (err) {
+      if (err instanceof Error.CastError) {
+        throw new WorkflowNotFoundException(command._workflowId);
+      }
+      throw err;
```
LetItRock, line 51: I'm not sure if, in this layer, we should use Mogo specific errors, we should catch that error in the repository and throw the `DalException` instead and then throw  `WorkflowNotFoundException`

SokratisVidros: I agree 100% with Pawel. I'd suggest that at this level, we return null if not found. Then we throw the exception in the controller that uses the use case.

SokratisVidros (changes requested):

PR #6689, opened 2024-10-14
fix(api): update previous step identifier to step id instead of inter…

djabarovgeorge:
**Please do not review this PR, i still need to merge a fix to the workflow creation, that will make sure we store `stepId`.**

Why?
Currently, we're using the internal ID "655b3427b8a8fe17cb62fe48" as the reference for the previous step. This means if a user wants to access variables from that step, they have to type something like {{steps.655b3427b8a8fe17cb62fe48.seen}}. This PR aims to improve the UX by allowing them to use a more readable format, like {{steps.in_app.seen}} (where "in_app" is the slugified step name).

### Screenshots

<details>

</details>


SokratisVidros (approved):
❤️

PR #6690, opened 2024-10-14
fix(js): Bypass cache during novu.notifications.list()

SokratisVidros:
Introduce `novu.notifications.list({ useCache: true | false })` option to always fetch the latest notifications and bypass the cache. This should give the ability to build an Inbox UI that auto-refreshes when a new notification is available.

The `useCache` option of the list function defaults to true.

### DIFF
--- a/packages/js/src/notifications/notifications.ts
+++ b/packages/js/src/notifications/notifications.ts
@@ -63,7 +63,8 @@ export class Notifications extends BaseModule {
     return this.callWithSession(async () => {
       const args = { limit, ...restOptions };
       try {
-        let data: ListNotificationsResponse | undefined = this.#useCache ? this.cache.getAll(args) : undefined;
+        const shouldUseCache = 'useCache' in args ? args.useCache : this.#useCache;
+        let data: ListNotificationsResponse | undefined = shouldUseCache ? this.cache.getAll(args) : undefined;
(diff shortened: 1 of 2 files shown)


packages/js/src/notifications/notifications.ts
```diff
@@ -59,11 +59,16 @@ export class Notifications extends BaseModule {
...
+    limit = 10,
+    useCache = true,
+    ...restOptions
+  }: ListNotificationsArgs = {}): Result<ListNotificationsResponse> {
     return this.callWithSession(async () => {
       const args = { limit, ...restOptions };
       try {
-        let data: ListNotificationsResponse | undefined = this.#useCache ? this.cache.getAll(args) : undefined;
+        let data: ListNotificationsResponse | undefined =
+          this.#useCache || useCache ? this.cache.getAll(args) : undefined;
```
LetItRock: @SokratisVidros, if I have set `useCache: true` in the options to the `Novu` class, then this function option `useCache` won't help with the override because the first part of `or` will be satisfied. Only when the options on the `Novu` class are `use cache: false` it will control it.

SokratisVidros: 🤦 good catch.It should be the other way around.

LetItRock (approved):

PR #6695, opened 2024-10-15
fix(js): Stabilize JS build process

SokratisVidros:
It seems that runAfterLast sometimes runs before the dist/cjs or dist/esm folder is created resulting in failure sin CI.


LetItRock (approved):
nice, thank you! 🤝

PR #6699, opened 2024-10-16
feat(dashboard): Create workflow sheet

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/primitives/form/form.tsx
+++ b/apps/dashboard/src/components/primitives/form/form.tsx
@@ -6,8 +6,9 @@ import { Controller, ControllerProps, FieldPath, FieldValues, FormProvider } fro
 import { cn } from '@/utils/ui';
 import { Label } from '@/components/primitives/label';
 import { cva } from 'class-variance-authority';
-import { RiInformationFill } from 'react-icons/ri';
 import { FormFieldContext, FormItemContext, useFormField } from './form-context';
+import { RiErrorWarningFill, RiInformationFill } from 'react-icons/ri';
+import { BsFillInfoCircleFill } from 'react-icons/bs';
 
 const Form = FormProvider;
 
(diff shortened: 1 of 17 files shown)


SokratisVidros (approved):

apps/dashboard/src/components/primitives/form/form.tsx
```diff
@@ -6,8 +6,9 @@ import { Controller, ControllerProps, FieldPath, FieldValues, FormProvider } fro
 import { cn } from '@/utils/ui';
 import { Label } from '@/components/primitives/label';
 import { cva } from 'class-variance-authority';
-import { RiInformationFill } from 'react-icons/ri';
 import { FormFieldContext, FormItemContext, useFormField } from './form-context';
+import { RiErrorWarningFill, RiInformationFill } from 'react-icons/ri';
+import { BsFillInfoCircleFill } from 'react-icons/bs';
```
BiswaViraj, line 11: ❓ **question:** What is our approach for icons for the dashboard?
I see we are also creating react svg files from figma.‏

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -154,86 +153,71 @@ export const WorkflowList = () => {
...
-                          className="cursor-pointer"
-                          text={workflow.name}
-                          onClick={() => {
-                            navigate(
-                              buildRoute(ROUTES.EDIT_WORKFLOW, {
-                                environmentId: currentEnvironment?._id ?? '',
-                                workflowId: workflow._id,
-                              })
-                            );
-                          }}
```
BiswaViraj, line 178: i think this is removed by mistake, without this we wont be able to navigate to the edit workflow screen

apps/dashboard/src/components/primitives/form/form.tsx
desiprisg: This was the closest I could find. I generally thought we were using remix icons, but it seems that it's not the case.

apps/dashboard/src/components/workflow-list.tsx
desiprisg: Maybe I did this on the rebase by accident. I'll fix it.

apps/dashboard/src/components/primitives/form/form.tsx
BiswaViraj: we are also using icons like [this](https://github.com/novuhq/novu/blob/next/apps/dashboard/src/components/icons/notification-5-fill.tsx) by copying from the figma directly

desiprisg: Yeah but some match icons in `react-icons`, so there's no need to do that. Do you think I should copy the svg in this case?

BiswaViraj: No i think it's fine, was just curious about how we are using the icons in the new dashboard

BiswaViraj (approved):

apps/dashboard/src/components/primitives/form/form.tsx
LetItRock: only these are the custom icons: https://www.figma.com/design/iOS9MnuJjLR7avdSIHemFy/novu-v3.0-Design-System-%E2%9A%A1%EF%B8%8F?node-id=125-263&node-type=text&m=dev
the rest are remix icons

LetItRock (approved):

PR #6700, opened 2024-10-16
feat(dashboard): Nv 4476 workflow editor header navigation

BiswaViraj:
### Screenshots

https://github.com/user-attachments/assets/74ea5490-834d-4921-8442-2d5cab670b65

<details>

</details>


apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -170,7 +170,10 @@ export const WorkflowList = () => {
                               buildRoute(ROUTES.EDIT_WORKFLOW, {
                                 environmentId: currentEnvironment?._id ?? '',
                                 workflowId: workflow._id,
-                              })
+                              }),
+                              {
+                                state: { workflowName: workflow.name },
+                              }
```
LetItRock: we will allow later updating of the workflow name in the drawer, so having this code won't work because the name won't be updated

apps/dashboard/src/pages/edit-workflow.tsx
```diff
@@ -1,10 +1,59 @@
...
+      label: 'Workflows',
+      href: buildRoute(ROUTES.WORKFLOWS, { environmentId: currentEnvironment?._id ?? '' }),
+    },
+  ];
+
+  return (
+    <>
+      <Breadcrumb>
+        <BreadcrumbList>
+          {breadcrumbs.map(({ label, href }) => (
```
LetItRock, line 55: missing the arrow left icon

apps/dashboard/src/pages/edit-workflow.tsx
```diff
@@ -1,10 +1,59 @@
...
+import { WorkflowEditor } from '@/components/workflow-editor';
+import { useEnvironment } from '@/context/environment/hooks';
+import { buildRoute, ROUTES } from '@/utils/routes';
+import React from 'react';
+import { useLocation } from 'react-router-dom';
 
 export const EditWorkflowPage = () => {
   return (
-    <EditWorkflowLayout headerStartItems={<h1 className="text-foreground-950">Edit Workflow</h1>}>
+    <EditWorkflowLayout headerStartItems={<StartItems />}>
```
LetItRock, line 21: we also need to hide the local studio btn

apps/dashboard/src/components/primitives/breadcrumb.tsx
```diff
@@ -0,0 +1,96 @@
+import { cn } from '@/utils/ui';
```
SokratisVidros, line 1: Nit picky comment: Is breadcrumb a primitive? It's a specific component displayed in one place in the app. Should we remove it from this folder?

apps/dashboard/src/components/primitives/breadcrumb.tsx
```diff
@@ -0,0 +1,96 @@
...
+  HTMLElement,
+  React.ComponentPropsWithoutRef<'nav'> & {
+    separator?: React.ReactNode;
+  }
+>(({ ...props }, ref) => <nav ref={ref} aria-label="breadcrumb" {...props} />);
+Breadcrumb.displayName = 'Breadcrumb';
+
+const BreadcrumbList = React.forwardRef<HTMLOListElement, React.ComponentPropsWithoutRef<'ol'>>(
+  ({ className, ...props }, ref) => (
+    <ol
```
SokratisVidros, line 17: 💯

SokratisVidros (approved):

apps/dashboard/src/components/primitives/breadcrumb.tsx
BiswaViraj: I guess we will use it in other areas too, like Activity feed, Subscribers list page
`< Development / Subscribers / Biswajeet Das`

apps/dashboard/src/components/workflow-list.tsx
BiswaViraj: Lifting up the useFetchworkflow to the this component will also require to lift up other form states and all,
may we can have a global state or context of currentActivePage 🤔 wdyt?

LetItRock: I'm not sure you need to lift up the form state, just the query call... the queries are created by the key, so it is fine to call `useQuery` multiple times, and the request will be triggered only once...  then we can update the cache and decide when to do it... because I'm not sure if it should be after every response from save (fyi we will have autosave)

apps/dashboard/src/pages/edit-workflow.tsx
```diff
@@ -1,10 +1,73 @@
...
+  const workflowsRoute = buildRoute(ROUTES.WORKFLOWS, { environmentId: currentEnvironment?._id ?? '' });
+  const { workflow } = useFetchWorkflow({
+    workflowId,
+  });
+
+  const breadcrumbs = [
+    { label: currentEnvironment?.name, href: '/' },
+    {
+      label: 'Workflows',
+      href: workflowsRoute,
```
LetItRock: these two will be the same route

LetItRock (approved):

PR #6705, opened 2024-10-16
feat(api): add slug parser in the api requests

djabarovgeorge:

**The PR is not ready please do not review**

### Screenshots

<details>

### Related enterprise PR
[EE-PR](https://github.com/novuhq/packages-enterprise/pull/235)

</details>


apps/api/src/app/workflows-v2/pipes/parse-slug-Id.pipe.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (validId) {
+    return validId;
+  }
+
+  const encodedValue = value.slice(-16);
+  let decodedValue: string;
+  try {
+    decodedValue = decodeBase62(encodedValue);
+  } catch (error) {
+    return value;
```
djabarovgeorge, line 44: if not base 62

apps/api/package.json
```diff
@@ -57,6 +57,7 @@
     "@types/newrelic": "^9.14.0",
     "@upstash/ratelimit": "^0.4.4",
     "axios": "^1.6.8",
+    "base-x": "^5.0.0",
```
rifont: ⚠ **issue:** Can we write our own base62 encoder from scratch? The pipe logic is effectively using unvalidated data. I'm concerned that any vulnerabilities (intentionally or maliciously) added to the `base-x` package could expose us‏ to an attack vector

apps/api/src/app/shared/helpers/base62.ts
```diff
@@ -0,0 +1,22 @@
...
+  const uint8Array = decode(encoded);
+
+  return Buffer.from(uint8Array).toString('hex');
+}
+
+export function isBase62(input: string): boolean {
+  const base62Regex = /^[0-9A-Za-z]+$/;
+
+  return base62Regex.test(input);
+}
```
rifont: ⚠ **issue:** Beware, [`base-x` has some logic](https://github.com/cryptocoinjs/base-x/blob/64e196c84f1d0b51eb7650d3881868568a0f525c/ts_src/index.ts#L39) to "skip" leading zeros when encoding/decoding. We should add tests for this case.‏

apps/api/package.json
djabarovgeorge: What do you think if will have fixed version 5.0.0? Will it solve the concern for us? Then, we could upgrade the version if needed and only to a stable and secure version.

SokratisVidros: Is there a know security issue with the package?

SokratisVidros: The truth is that a custom implementation with two functions, encodeBase62, decodeBase62 would be much much simpler, no need for pipes, no need to worry about potential issues.

SokratisVidros (changes requested):
I am requesting changes only because of base-x

apps/api/package.json
djabarovgeorge: > ⚠ issue: Can we write our own base62 encoder from scratch? The pipe logic is effectively using unvalidated data. I'm concerned that any vulnerabilities (intentionally or maliciously) added to the base-x package could expose us‏ to an attack vector

i had a hard time implementing the code from scratch, i added up comping the code from base-x so we won't get any vulnerabilities by mistake.

> Is there a know security issue with the package?

not at the moment.


> The truth is that a custom implementation with two functions, encodeBase62, decodeBase62 would be much much simpler, no need for pipes, no need to worry about potential issues.

i implemented encodeBase62, decodeBase62 based on base-x, pipes are actually a nice touch of nest js making the boundaries of controller and use case responsibility cleaner

apps/api/src/app/shared/helpers/base62.ts
djabarovgeorge: mongo object is the first byte of the timestamp component, and the timestamp component is always a positive integer. however in order to make sure i added tests, from the tests i found out that the encoding supports 2 leading zero.

SokratisVidros (approved):

PR #6713, opened 2024-10-17
feat(api): add get environment tags endpoint

ChmaraX:
* added new environments V2 module and controller with the new get tags endpoint
* renamed legacy environments to environments V1 following existing workflows module naming
* added e2e tests


apps/api/src/app.module.ts
```diff
@@ -76,7 +77,7 @@ const baseModules: Array<Type | DynamicModule | Promise<DynamicModule> | Forward
   InboundParseModule,
   SharedModule,
   HealthModule,
-  EnvironmentsModule,
+  EnvironmentsModuleV1,
```
ChmaraX, line 80: I assumed we want to isolate logic for new dashboard in separate module, e.g. as with workflows, but it could as well be in the same just under different controller, if we want to.

SokratisVidros (approved):

PR #6715, opened 2024-10-17
feat(api, worker, application-generic): Add exhaustive error handling for bridge requests

rifont:

Add exhaustive error handling during Bridge requests for all of:
* `got` package error codes (it does a good job of handling these with explicit classes per error type
* All Node.js Network error code ([ref](https://nodejs.org/dist/latest-v20.x/docs/api/errors.html#nodejs-error-codes))

### Screenshots

<details>

</details>


apps/worker/src/app/workflow/usecases/execute-bridge-job/execute-bridge-job.usecase.ts
```diff
@@ -249,12 +249,23 @@ export class ExecuteBridgeJob {
...
+          };
+        }
+
         raw = {
           url: statelessBridgeUrl,
           statusCode: error.response?.statusCode,
           message: error.response?.statusMessage,
           ...(error.response?.retryCount ? { retryCount: error.response?.retryCount } : {}),
-          ...(error?.response?.body?.length > 0 ? { raw: JSON.parse(error?.response?.body) } : {}),
+          ...(error?.response?.body?.length > 0 ? { raw: rawMessage } : {}),
```
rifont, line 268: I believe this lack of error handling during body parsing was contributing to execution detail logs not surfacing due to an inability to parse the error body. This was observed with errors such as "BSON circular value cannot be converted..."

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -256,8 +214,19 @@ export class ExecuteBridgeRequest {
...
+      case WorkflowOriginEnum.EXTERNAL: {
+        if (!environmentBridgeUrl) {
+          throw new BadRequestException({
+            code: BRIDGE_EXECUTION_ERROR.INVALID_BRIDGE_URL.code,
+            message:
+              BRIDGE_EXECUTION_ERROR.INVALID_BRIDGE_URL.message(
+                environmentBridgeUrl,
+              ),
+          });
+        }
```
rifont, line 226: Throw an exception here explicitly if trying to request to an external Bridge URL when it doesn't exist on the environment.

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -56,8 +56,12 @@ export class Sync {
         retriesLimit: 1,
         workflowOrigin: WorkflowOriginEnum.EXTERNAL,
       })) as DiscoverOutput;
-    } catch (error: any) {
-      throw new BadRequestException(`Bridge URL is not valid. ${error.message}`);
+    } catch (error) {
+      if (error instanceof HttpException) {
+        throw new BadRequestException(error.message);
+      }
```
rifont, line 62: Simplifying the error thrown here.

Let non-HTTP exceptions bubble up to the API exception handler.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -76,11 +88,6 @@ export class ExecuteBridgeRequest {
       );
     }
 
-    const secretKey = await this.getDecryptedSecretKey.execute(
-      GetDecryptedSecretKeyCommand.create({
-        environmentId: command.environmentId,
-      }),
-    );
```
rifont, line 83: Refactoring many of these to discrete methods to make the `execute` method easier to read.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -152,75 +147,38 @@ export class ExecuteBridgeRequest {
         headers,
       }).json();
     } catch (error) {
-      if (error instanceof RequestError) {
```
rifont, line 155: Moved to a discrete handler for errors.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -272,4 +241,135 @@ export class ExecuteBridgeRequest {
 
     return apiUrl;
   }
+
+  private handleResponseError(error: unknown, url: string) {
```
rifont, line 245: Exhaustive error handling added for all of:
* `got` package error codes (it does a good job of handling these with explicit classes per error type
* All Node.js error code ([ref](https://nodejs.org/dist/latest-v20.x/docs/api/errors.html#nodejs-error-codes))

libs/application-generic/src/utils/bridge.ts
rifont, line 1: All thrown error messages now include the URL to support client debugging.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -4,8 +4,19 @@ import {
   NotFoundException,
   BadRequestException,
   HttpException,
+  GatewayTimeoutException,
 } from '@nestjs/common';
-import got, { OptionsOfTextResponseBody, RequestError } from 'got';
+import got, {
```
SokratisVidros, line 9: In the era of fetch, do we need to rely on `got`? Got by default gives us retries which I am not very sure if it's a great idea.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -272,4 +241,135 @@ export class ExecuteBridgeRequest {
...
+        body = {};
+      }
+
+      if (
+        error instanceof HTTPError &&
+        Object.values(ErrorCodeEnum).includes(body.code as ErrorCodeEnum)
+      ) {
+        // Handle known Bridge errors. Propagate the error code and message.
+        throw new HttpException(body, error.response.statusCode);
+      } else if (body.code === TUNNEL_ERROR_CODE) {
```
SokratisVidros: Readability tip, since you are throwing, you don't need the `else if`. You can put an if in a new line.

SokratisVidros (approved):

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
SokratisVidros: We will tackle this offline.

rifont: Elaborating a little, is the thinking that we should handle retries elsewhere? At the queue layer, for example?

I tend to agree, the value we're getting from retries in practice is minimal, especially considering the total retry duration after incorporating backoff is less than 1 minute. It's unlikely an external bridge endpoint would recover from a total network fault within that timespan.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
rifont: Great point! I will update.

PR #6717, opened 2024-10-17
fix(application-generic): Allow unauthorized certs for bridge url

SokratisVidros:

Reject self-signed and invalid certificates in Production environments but allow them in Development as it's common for developers to use self-signed certificates in local environments.

### Screenshots
Screenshot 2024-10-18 at 00 06 21


apps/web/tests/integrations-list-modal.spec.ts
```diff
@@ -466,10 +466,6 @@ test('should update the mailjet integration', async ({ page }) => {
   await updateButton.click();
   await expect(updateButton).toBeDisabled();
 
-  // Test that the update button is enabled again when the form value changes
-  await senderName.fill('Novu edited');
-  await expect(updateButton).toBeEnabled();
-
```
rifont: Is this change intended? It seems unrelated to the PR title

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -130,6 +130,11 @@ export class ExecuteBridgeRequest {
         afterResponse:
           command.afterResponse !== undefined ? [command.afterResponse] : [],
       },
+      /*
+       * Reject self-signed and invalid certificates in Production environments but allow them in Development
+       * as it's common for developers to use self-signed certificates in local environments.
+       */
+      rejectUnauthorized: environment.name.toLowerCase() === 'production',
```
rifont: For later: we need to add and consume an constant value for all of these types Environment checks. We have many of them scattered through the codebase.

rifont (approved):
Looks good!

My only question is: what does the API response look like when attempting to use a self-signed certificate in Prod? We should make sure it's not a 5xx error.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
SokratisVidros: Totally, and it's not ideal that we rely on the name and not on an environment type.

SokratisVidros:
> Looks good!
> 
> My only question is: what does the API response look like when attempting to use a self-signed certificate in Prod? We should make sure it's not a 5xx error.

Screenshot 2024-10-19 at 16 18 38

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -130,6 +131,11 @@ export class ExecuteBridgeRequest {
         afterResponse:
           command.afterResponse !== undefined ? [command.afterResponse] : [],
       },
+      /*
+       * Reject self-signed and invalid certificates in Production environments but allow them in Development
+       * as it's common for developers to use self-signed certificates in local environments.
+       */
+      rejectUnauthorized: true,
```
rifont, line 138: @SokratisVidros the new error response looks great! One thing though - you forgot to revert the change here to check Environement.name after testing

SokratisVidros: 😮‍💨 Fixed on next. Thanks for noticing. That's what happens if you are committing code in between Duplo gaming sessions.

PR #6718, opened 2024-10-17
feat(dashboard): workflow editor autosave

LetItRock:

Workflow Editor autosave functionality and sonner toast for successfully saved action.

Because there are a few issues with the BE implementation on the video you can see that I only added one step.

### Screenshots

https://github.com/user-attachments/assets/724ed95b-e4dc-47a0-9fd0-0f540f520add


apps/dashboard/package.json
```diff
@@ -41,17 +41,21 @@
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.1",
     "date-fns": "^4.1.0",
+    "debounce": "^2.2.0",
```
LetItRock: the debounce library used to debounce save workflow requests

apps/dashboard/package.json
```diff
@@ -41,17 +41,21 @@
...
     "react-dom": "^18.3.1",
     "react-helmet-async": "^1.3.0",
     "react-hook-form": "7.43.9",
     "react-icons": "^5.3.0",
     "react-router-dom": "6.26.2",
     "react-use-intercom": "^2.0.0",
+    "sonner": "^1.5.0",
     "tailwind-merge": "^2.4.0",
     "tailwindcss-animate": "^1.0.7",
+    "use-deep-compare-effect": "^1.8.1",
```
LetItRock, line 59: used in the autosave hook, more details below

apps/dashboard/src/components/primitives/sonner.tsx
```diff
@@ -0,0 +1,42 @@
...
+        className
+      )}
+      {...props}
+    >
+      {children}
+    </div>
+  );
+};
+
+const Toaster = ({ ...props }: ToasterProps) => {
```
LetItRock, line 21: autogenerated

apps/dashboard/src/components/primitives/sonner.tsx
```diff
@@ -0,0 +1,42 @@
+import { cn } from '@/utils/ui';
+import { useTheme } from 'next-themes';
+import { Toaster as Sonner } from 'sonner';
+
+type ToasterProps = React.ComponentProps<typeof Sonner>;
+
+const SmallToast = ({ children, className, ...props }: React.HTMLAttributes<HTMLDivElement>) => {
```
LetItRock, line 7: small sonner implementation shown in the workflow editor I will explain to you why later in the comments

apps/dashboard/src/components/workflow-editor/schema.ts
```diff
@@ -3,17 +3,17 @@ import { StepTypeEnum } from '@/utils/enums';
...
-      type: z.nativeEnum(StepTypeEnum),
-      controls: z.object({ schema: z.object({}) }),
-      controlValues: z.object({}),
-    })
+    z
+      .object({
+        name: z.string(),
+        type: z.nativeEnum(StepTypeEnum),
+      })
+      .passthrough()
```
LetItRock, line 17: `passthrough` is passing all the other fields from the persisted step apart from what is defined in the schema, for example `stepUuid, controls, controlValues`, etc. Otherwise during the save we will be missing these and the BE will treat it as a new step.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -27,36 +30,72 @@ const STEP_NAME_BY_TYPE: Record<StepTypeEnum, string> = {
...
     control: form.control,
     name: 'steps',
   });
 
-  const { workflow: _workflow } = useFetchWorkflow({
+  const { workflow } = useFetchWorkflow({
     workflowId,
     onSuccess: (data) => {
-      reset(data);
+      reset({ ...data, steps: data.steps.map((step) => ({ ...step })) });
```
LetItRock: to satisfy TS until the BE types are fixed

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -27,36 +30,72 @@ const STEP_NAME_BY_TYPE: Record<StepTypeEnum, string> = {
...
+  });
+
+  useFormAutoSave({
+    form,
+    onSubmit: async (data: z.infer<typeof formSchema>) => {
+      if (!workflow) {
+        return;
+      }
+
+      updateWorkflow({ id: workflow._id, workflow: { ...workflow, ...data } as any });
```
LetItRock, line 100: to satisfy TS until the BE types are fixed

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -27,36 +30,72 @@ const STEP_NAME_BY_TYPE: Record<StepTypeEnum, string> = {
...
+          position: 'bottom-left',
+          unstyled: true,
+          classNames: {
+            toast: 'ml-10',
+          },
+          onAutoClose: () => {
+            changesSavedToastIdRef.current = undefined;
+          },
+        }
+      );
```
LetItRock: This is how the sonner toast is used. Opened for the discussion.
Basically, in the Workflow Editor, we might want to show different types of toasts, like a small version when the workflow is saved or a bigger one when the step is updated. And I haven't seen a way of saying which variant you want to show, this can only be defined when `Toaster` is used. That's why the only solution I saw was to use `unstyled` property and provide the component to the `toast` function call.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -27,36 +30,72 @@ const STEP_NAME_BY_TYPE: Record<StepTypeEnum, string> = {
...
+        return;
+      }
+
+      const id = toast(
+        <SmallToast>
+          <RiProgress1Line className="size-6" />
+          <span className="text-sm">Saved</span>
+        </SmallToast>,
+        {
+          duration: 5000,
```
LetItRock: tbd

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -27,36 +30,72 @@ const STEP_NAME_BY_TYPE: Record<StepTypeEnum, string> = {
...
+          onAutoClose: () => {
+            changesSavedToastIdRef.current = undefined;
+          },
+        }
+      );
+      changesSavedToastIdRef.current = id;
+    },
+  });
+
+  useFormAutoSave({
```
LetItRock, line 93: form autosave hook

apps/dashboard/package.json
SokratisVidros: Nit-picky one: Prefer lodash so that we have a single point of reference for such utilities.

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -0,0 +1,33 @@
...
+      handleSubmit(onSubmitRef.current)();
+    }, 1000),
+    [handleSubmit]
+  );
+
+  useDeepCompareEffect(() => {
+    if (formState.isDirty) {
+      debouncedSave();
+    }
+  }, [watchedData]);
```
LetItRock: Basically, we watch all the form data changes and compare it deeply with the previous form-watched data between re-renders. If the form data was changed and the form is dirty, we call debounced save, which calls the react hook form `handleSubmit` that triggers validation. If the data is valid, then the `onSubmit` will be called.

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -0,0 +1,33 @@
...
+  const { formState, control, handleSubmit } = form;
+
+  const watchedData = useWatch<T>({
+    control,
+  });
+
+  const debouncedSave = useCallback<() => void>(
+    debounce(() => {
+      handleSubmit(onSubmitRef.current)();
+    }, 1000),
```
SokratisVidros: Pass this as an option to that hook that defaults to 1000

apps/dashboard/src/hooks/use-form-autosave.ts
LetItRock: Maybe additionally we can show the sonner when there are validation errors?

SokratisVidros (approved):

apps/dashboard/src/components/workflow-editor/schema.ts
desiprisg: Oh so this just doesn't strip out the unspecified fields.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
desiprisg: Have you checked this out? https://github.com/shadcn-ui/ui/issues/2254#issuecomment-1872982460
We can probably create our own functions with predefined css here. Or just a function that is just `toast()` with a custom component and use that across the app. Unless we are not planning to reuse this specific one somewhere else.

apps/dashboard/src/hooks/use-form-autosave.ts
desiprisg: A `useDebouncedCallback` hook would be nice.

desiprisg (approved):
Nice work! Just some minor comments/suggestions.

apps/dashboard/src/components/workflow-editor/schema.ts
LetItRock: yes

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
LetItRock: Ohh nice, I didn't see that... seems to be a better approach, I'll update it then 🙌

LetItRock: ahh I thought you could create your custom function...

LetItRock: There are two options:
- create a variant on the `Toaster` 
- use `unstyled` prop and provide a custom component <--- this is what I did

I didn't go with the first approach as we might want to show different toast types on the same page.

LetItRock: I'll move the code into the custom function so that it will be more reusable

PR #6728, opened 2024-10-21
feat(dashboard): Tag input suggestions

desiprisg:

* Removed barrel files
* Added the suggestions for tag inputs
### Screenshots

https://github.com/user-attachments/assets/aed7024e-cdaa-4af4-b5ec-2709d7dfd8fa

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/create-workflow-button.tsx
+++ b/apps/dashboard/src/components/create-workflow-button.tsx
@@ -1,6 +1,6 @@
 import { createWorkflow } from '@/api/workflows';
 import { Button } from '@/components/primitives/button';
-import { Form, FormControl, FormField, FormItem, FormLabel, FormMessage } from '@/components/primitives/form';
+import { FormField, FormItem, FormLabel, FormControl, FormMessage, Form } from '@/components/primitives/form/form';
 import { Input, InputField } from '@/components/primitives/input';
 import { Separator } from '@/components/primitives/separator';
 import {
@@ -16,6 +16,7 @@ import {
 import { TagInput } from '@/components/primitives/tag-input';
 import { Textarea } from '@/components/primitives/textarea';
 import { useEnvironment } from '@/context/environment/hooks';
+import { useTagsQuery } from '@/hooks/use-tags-query';
 import { QueryKeys } from '@/utils/query-keys';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { type CreateWorkflowDto, WorkflowCreationSourceEnum } from '@novu/shared';
@@ -54,6 +55,8 @@ export const CreateWorkflowButton = (props: CreateWorkflowButtonProps) => {
       setIsOpen(false);
     },
   });
+  const tagsQuery = useTagsQuery();
+
   const form = useForm<z.infer<typeof formSchema>>({
     resolver: zodResolver(formSchema),
     defaultValues: { description: '', identifier: '', name: '', tags: [] },
@@ -147,7 +150,7 @@ export const CreateWorkflowButton = (props: CreateWorkflowButtonProps) => {
                       <FormLabel hint="(max. 8)">Add tags</FormLabel>
                     </div>
                     <FormControl>
-                      <TagInput {...field} />
+                      <TagInput suggestions={tagsQuery.data?.data.map((tag) => tag.name) || []} {...field} />
                     </FormControl>
                     <FormMessage />
                   </FormItem>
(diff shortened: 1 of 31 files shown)


apps/dashboard/src/components/create-workflow-button.tsx
```diff
@@ -54,6 +55,12 @@ export const CreateWorkflowButton = (props: CreateWorkflowButtonProps) => {
       setIsOpen(false);
     },
   });
+  const { data: tagData } = useQuery<{ data: { name: string }[] }>({
```
SokratisVidros: Let's factor this out to a dedicated `useTags` hook. We will need it in other screens as well.

apps/dashboard/src/components/primitives/tag-input.tsx
```diff
@@ -39,47 +45,74 @@ const TagInput = forwardRef<HTMLInputElement, TagInputProps>((props, ref) => {
...
-              onClick={() => removeTag(tag)}
-              className="ml-1 rounded-full outline-none focus:outline-none"
-            >
-              <RiCloseFill className="size-3" />
-              <span className="sr-only">Remove tag</span>
-            </button>
-          </Badge>
-        ))}
-      </div>
+    <div>
```
SokratisVidros: This extra div can be removed

apps/dashboard/src/components/primitives/tag-input.tsx
```diff
@@ -39,47 +45,74 @@ const TagInput = forwardRef<HTMLInputElement, TagInputProps>((props, ref) => {
...
+            portal={false}
+            onOpenAutoFocus={(e) => {
+              e.preventDefault();
+            }}
+          >
+            <CommandList>
+              <CommandGroup>
+                {inputValue !== '' && (
+                  <CommandItem
+                    value={`input-${inputValue}`} // if this exists in the list below there are bugs introduces to the key up/down behaviour
```
SokratisVidros: @desiprisg This comment doesn't make a lot of sense. Can you make it more descriptive?

SokratisVidros (approved):

apps/dashboard/src/components/primitives/tag-input.tsx
desiprisg: Done

apps/dashboard/src/components/create-workflow-button.tsx
desiprisg: Done. I added a `useTagsQuery` here and I think that pattern will suit us better than filtering out what we pass / return. @LetItRock  Let's discuss this and see what pattern will fit us best.

apps/dashboard/src/components/primitives/tag-input.tsx
desiprisg: Done

apps/dashboard/src/hooks/use-tags-query.ts
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from '@tanstack/react-query';
+import { QueryKeys } from '@/utils/query-keys';
+import { useEnvironment } from '@/context/environment/hooks';
+import { getV2 } from '@/api/api.client';
+
+export const useTagsQuery = () => {
+  const { currentEnvironment } = useEnvironment();
+  const query = useQuery<{ data: { name: string }[] }>({
+    queryKey: [QueryKeys.fetchWorkflow, currentEnvironment?._id],
+    queryFn: async () => await getV2(`/environments/${currentEnvironment!._id}/tags`),
```
LetItRock, line 10: please move it to the API helper function, then we won't need to do `data.data` when using the query hook

LetItRock (approved):

apps/dashboard/src/components/create-workflow-button.tsx
LetItRock: @desiprisg I'm totally fine with following `useXQuery, useXMutation` pattern ;) let's just quickly update the hooks in the separate ticket

SokratisVidros: Folks, the one we have now is OK as well. So please align them separately.

desiprisg: I also meant not passing `onSuccess` and `onError` funcs, and just passing through what `@tanstack/react-query` accepts.

apps/dashboard/src/hooks/use-tags-query.ts
desiprisg: I don't think this is correct. If it's a `useTagsQuery` hook, then you should get back whatever was in the body, right? If `data` doesn't make sense, then that's a BE concern I guess.

desiprisg: merged in order to get the import changes in. If you feel strongly about this, let's discuss privately.

PR #6733, opened 2024-10-21
fix(application-generic): handle lock acquisition conflicts

rifont:

* Throw a `409 Conflict Error` when a cache lock cannot be acquired. `409` is commonly understood to be a retryable 4xx error, so this change is semantic.

### Screenshots

<details>

</details>

### DIFF
--- a/libs/application-generic/src/services/cache/interceptors/cached-entity.interceptor.ts
+++ b/libs/application-generic/src/services/cache/interceptors/cached-entity.interceptor.ts
@@ -1,4 +1,4 @@
-import { Inject, Logger } from '@nestjs/common';
+import { ConflictException, Inject, Logger } from '@nestjs/common';
 import { DistributedLockService } from '../../distributed-lock';
 import { CacheService, CachingConfig } from '../cache.service';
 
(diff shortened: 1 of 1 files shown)


libs/application-generic/src/services/cache/interceptors/cached-entity.interceptor.ts
```diff
@@ -89,7 +89,9 @@ export function CachedEntity<T_Output = any>({
             `Failed to acquire lock for key: ${cacheKey} in "method: ${methodName}"`,
             LOG_CONTEXT,
           );
-          throw new Error(`Failed to acquire lock for key: ${cacheKey}`);
+          throw new ConflictException(
+            `Failed to acquire cache lock. Please try again later.`,
+          );
```
rifont, line 94: It was a mistake to include `cacheKey` in the log as it can expose server implementation details. We just return a simple error message. Note, it's still not ideal that cache issues should throw a client error.

ChmaraX (approved):

SokratisVidros (approved):
Lovely!

PR #6743, opened 2024-10-22
chore(shared): Eliminate circular dependencies

SokratisVidros:

Cause, why not ❓ 

@novu/shared doesn't have strict conventions regarding its folder structure. As a result, types import entities, dtos import types or entities, and so on. This resulted in a very large number of TS circular imports, causing unexpected issues and TS headaches.

This PR addresses that with the help of [madge](https://www.npmjs.com/package/marge). 🙌 

Consider it, as a first step towards a more structured, treeshakable `@novu/shared`, that contains meaningful share utility functions and types.


packages/shared/src/dto/workflows/workflow-response-dto.ts
```diff
@@ -1,4 +1,4 @@
-import { PreferencesResponseDto, StepResponseDto, WorkflowCommonsFields } from './workflow-commons-fields';
+import { PreferencesResponseDto, StepResponseDto, WorkflowCommonsFields } from './workflow-response.dto';
```
SokratisVidros: @tatarco I think we made the exact parallel change in two PRs. Let's chat tomorrow to see how can we merge, the `workflow-response.dto`or `workflow-response-dto`.

packages/framework/package.json
```diff
@@ -26,11 +26,12 @@
     "format:fix": "prettier --write --ignore-path .gitignore .",
     "build": "tsup && pnpm check:circulars",
```
rifont: We should remove the addition invocation of `check:circulars` in `build` now that we have `postbuild`
```suggestion
    "build": "tsup",
```

rifont (approved):
🧹💅🚀 awesome!

PR #6754, opened 2024-10-23
refactor(root): introducing graceful shutdown and optimise docker images

merrcury:
This PR linked to below issues
- Fixes INF-447
- Fixes  INF-443
- Fixes  INF-445

### Screenshots
New Images are now 46% Slimmer 
Screenshot 2024-10-23 at 3 09 50 PM


libs/application-generic/src/index.ts
```diff
@@ -5,14 +5,15 @@ export * from './decorators';
 export * from './dtos';
 export * from './encryption/index';
 export * from './factories/index';
+export * from './graceful-shutdown/index';
```
rifont: ☑ **chore:** Extra path not needed‏

```suggestion
export * from './graceful-shutdown';
```

libs/application-generic/src/graceful-shutdown/shutdown.service.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable } from '@nestjs/common';
+
+@Injectable()
+export class ShutdownService {
+  public readonly timeout =
+    parseInt(process.env.GRACEFUL_SHUTDOWN_TIMEOUT, 10) || 5000;
```
rifont: ☑ **chore:** Please add the new environment variable to all the [env.validators](https://github.com/novuhq/novu/blob/next/apps/api/src/config/env.validators.ts) files in services that need it. This ensures that the application won't startup without the env-var being present. You will also be able to specify the `5000` default there to fully encapsulate the necessary fallback logic.‏

packages/framework/package.json
```diff
@@ -211,7 +211,6 @@
   "devDependencies": {
     "@apidevtools/json-schema-ref-parser": "11.6.4",
     "@arethetypeswrong/cli": "^0.16.4",
-    "@nestjs/common": "10.4.1",
```
rifont, line 214: ⚠ **issue:** We need to keep this so that the test suites have the needed types. Please revert the removal.‏

SokratisVidros (approved):
😍

apps/api/Dockerfile
```diff
@@ -1,72 +1,54 @@
...
-ENV NEW_RELIC_NO_CONFIG_FILE=true
+# Install production dependencies
+RUN --mount=type=cache,id=pnpm-store-api,target=/root/.pnpm-store \
+    --mount=type=secret,id=BULL_MQ_PRO_NPM_TOKEN,uid=1000 \
+    pnpm install --filter "{${PACKAGE_PATH}}..." --frozen-lockfile --unsafe-perm --prod
 
+# Set the working directory to the API app and start the application using pm2-runtime
 WORKDIR /usr/src/app/apps/api
-CMD [ "pm2-runtime","start", "dist/main.js" ]
+CMD [ "pm2-runtime", "start", "dist/main.js" ]
```
SokratisVidros, line 54: Didn't we remove `pm2`?

merrcury: Next Iteration, with ecs. We are ready to remove it but maybe in next two days

libs/application-generic/src/index.ts
merrcury: Okay 👍🏼

packages/framework/package.json
merrcury: Hey @rifont, If I am adding this into dev dependency and then running `--prod` its not getting installed and API build up is failing. any workaround/idea?

rifont: That's strange, as `@novu/api` declares `@nestjs/common` as a `dependency` [here](https://github.com/novuhq/novu/pull/6754/files#diff-2c40985d6d91eed8ae85ec1c8e754a85984ee32e156a600d2b7a467423d7e338R36) which means it should be installed when using the `--prod` flag.

Could you please share the error log for further debugging?

I believe reverting the removal of this dependency is a blocker for Framework tests to pass.

merrcury: Hey @rifont, Please find error below:
Screenshot 2024-10-29 at 11 49 50 AM

libs/application-generic/src/index.ts
merrcury: Done

libs/application-generic/src/graceful-shutdown/shutdown.service.ts
merrcury: Done

packages/framework/package.json
SokratisVidros: This is now solved. See https://github.com/novuhq/novu/pull/6754/commits/4a3e98d6be620f52a25aec47a057fdd4c2033e58. 

The trick is that @novu/framework requires a range for @nestjs/common peer dependency but by pinning the @nestjs/common version via PNPM overrides we effectively teach PNPM to add the correct dependencies when installing only production dependencies.

merrcury: I tested locally as well. It worked fabulous for me. Sok, we need an article for all of this 😄 .

SokratisVidros: We will! 😆

PR #6758, opened 2024-10-23
fix(node): add @novu/nest deprecated warning and add novu/node self host info

jainpawan21:

1. Add `@novu/nest` deprecation warning
2. Add `@novu/node` self host config info 
### Screenshots

<details>

</details>

### DIFF
--- a/packages/node/README.md
+++ b/packages/node/README.md
@@ -62,9 +62,9 @@ yarn add @novu/node
 ```ts
 import { Novu } from '@novu/node';
 
-const novu = new Novu(process.env.NOVU_SECRET_KEY);
+const novu = new Novu('<NOVU_SECRET_KEY>');
 
-await novu.trigger('<REPLACE_WITH_EVENT_NAME_FROM_ADMIN_PANEL>', {
+await novu.trigger('workflowIdentifier', {
   to: {
     subscriberId: '<USER_IDENTIFIER>',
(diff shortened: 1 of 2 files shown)


SokratisVidros (approved):

packages/node/README.md
```diff
@@ -62,9 +62,9 @@ yarn add @novu/node
 ```ts
 
-const novu = new Novu(process.env.NOVU_SECRET_KEY);
+const novu = new Novu('<NOVU_SECRET_KEY>');
```
rifont, line 65: I think most folks will make use the previous of `process.env.NOVU_SECRET_KEY` pattern, this is the same pattern we follow elsewhere in our documentation. Was there a specific reason for this change?

rifont (approved):
🔥

packages/node/README.md
jainpawan21: Sorry I merge this PR before reading this comment. in this whole readme file, we were using `<NOVU_SECRET_KEY>`. So I replace process.env.NOVU_SECRET_KEY with <NOVU_SECRET_KEY>

rifont: That makes sense for this PR, thanks for clarifying. We should align on a pattern everywhere in future docs updates. I prefer `new Novu(process.env.NOVU_SECRET_KEY);` as it's workable out of the box.

PR #6762, opened 2024-10-23
fix(framework): Experiement with importing json-schema-faker

SokratisVidros:

This dependency is causing issues when @novu/framework is used in Next.js while webpack is trying to load it via ESM.

See [here](https://github.com/json-schema-faker/json-schema-faker/issues/796#issuecomment-2433335751) and [here](https://github.com/json-schema-faker/json-schema-faker/blob/master/docs/DEPRECATED.md#deprecation-notice).


rifont (approved):
Awesome! Great detective work. Let's add some code comments referencing the outstanding GH issues, mentioning both:
* not to use the `default` export with `json-schema-faker`
* not to use `import` keyword with `json-schema-faker`

PR #6742, opened 2024-10-22
feat(api): remove controls from response

djabarovgeorge:
### What changed? Why was the change needed?
remove redundant controls from step schema controller response, the controls moved to workflows response

### Screenshots

<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #6764, opened 2024-10-24
feat(api): add ui scehma

tatarco:
removed control scehma from step schema endpoint.
added UI schema concept.
added first implementatino of ui schema for inApp with control group

https://linear.app/novu/issue/NV-4547/create-ui-schema

### Screenshots

<details>

</details>


apps/api/src/app/step-schemas/shared/mappers/map-step-type-to-output.mapper.ts
```diff
@@ -1,12 +1,23 @@
...
-  [ActionStepEnum.DELAY]: actionStepSchemas[ActionStepEnum.DELAY].output,
-  [ActionStepEnum.DIGEST]: actionStepSchemas[ActionStepEnum.DIGEST].output,
+export const PERMISSIVE_EMPTY_SCHEMA = {
+  type: 'object',
+  properties: {},
+  required: [],
+  additionalProperties: true,
+} as const;
+
+export const mapStepTypeToControlScema: Record<ChannelStepEnum | ActionStepEnum, ControlsSchema> = {
```
SokratisVidros:
```suggestion
export const mapStepTypeToControlSchema: Record<ChannelStepEnum | ActionStepEnum, ControlsSchema> = {
```

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,5 @@
 import { JSONSchema } from 'json-schema-to-ts';
 
 export type StepSchemaDto = {
-  controls: JSONSchema;
```
djabarovgeorge, line 4: As discussed, I've already removed the controls from the DTO in PR #6742.

apps/api/src/app/step-schemas/shared/mappers/map-step-type-to-output.mapper.ts
```diff
@@ -1,12 +1,45 @@
...
-  [ActionStepEnum.DELAY]: actionStepSchemas[ActionStepEnum.DELAY].output,
-  [ActionStepEnum.DIGEST]: actionStepSchemas[ActionStepEnum.DIGEST].output,
+export const PERMISSIVE_EMPTY_SCHEMA = {
+  type: 'object',
+  properties: {},
+  required: [],
+  additionalProperties: true,
+} as const;
+
+export const mapStepTypeToControlScema: Record<ChannelStepEnum | ActionStepEnum, ControlsDto> = {
```
SokratisVidros:
```suggestion
export const mapStepTypeToControlSchema: Record<ChannelStepEnum | ActionStepEnum, ControlsDto> = {
```

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
...
+export const mapStepTypeToUiSchema: Record<ChannelStepEnum | ActionStepEnum, UiSchema> = {
+  [ChannelStepEnum.SMS]: { type: 'sms', properties: smsProperties },
+  [ChannelStepEnum.EMAIL]: { type: 'email', properties: emailProperties },
+  [ChannelStepEnum.PUSH]: { type: 'push', properties: pushProperties },
+  [ChannelStepEnum.CHAT]: { type: 'chat', properties: chatProperties },
+  [ChannelStepEnum.IN_APP]: { type: 'in_app', properties: inAppProperties },
+  [ActionStepEnum.DELAY]: { type: 'delay', properties: delayProperties },
+  [ActionStepEnum.DIGEST]: { type: 'digest', properties: digestProperties },
+  [ActionStepEnum.CUSTOM]: { type: 'custom', properties: customProperties },
+};
```
rifont: We can get some stronger downstream type-safety by using the `satisfies` operator with `const`:
```suggestion
export const mapStepTypeToUiSchema = {
  [ChannelStepEnum.SMS]: { type: 'sms', properties: smsProperties },
  [ChannelStepEnum.EMAIL]: { type: 'email', properties: emailProperties },
  [ChannelStepEnum.PUSH]: { type: 'push', properties: pushProperties },
  [ChannelStepEnum.CHAT]: { type: 'chat', properties: chatProperties },
  [ChannelStepEnum.IN_APP]: { type: 'in_app', properties: inAppProperties },
  [ActionStepEnum.DELAY]: { type: 'delay', properties: delayProperties },
  [ActionStepEnum.DIGEST]: { type: 'digest', properties: digestProperties },
  [ActionStepEnum.CUSTOM]: { type: 'custom', properties: customProperties },
} as const satisfies Record<ChannelStepEnum | ActionStepEnum, UiSchema>;
```

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
 import { JSONSchema } from 'json-schema-to-ts';
+import { StepType } from '@novu/shared';
+
+import { ActionStepEnum, ChannelStepEnum } from '@novu/framework/internal';
+
+const chatProperties = {
+  body: { type: 'chatBody' },
```
rifont, line 7: I like the simple direction you've taken with prefixing the step type to the property name. It's opinionated and consistent and makes it easier for us to iterate on later.

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
...
+  | typeof inAppProperties
+  | typeof pushProperties
+  | typeof smsProperties
+  | typeof delayProperties
+  | typeof digestProperties
+  | typeof customProperties;
+
+export type UiSchema = {
+  type: `${StepType}`;
+  properties?: Properties;
```
rifont: Should properties always be supplied?
```suggestion
  properties: Properties;
```

rifont: Furthermore, I think it would be preferable to type out the `Properties` interface separately rather than collect all the different ui schema properties into the type. That way the `Properties` interface can evolve independently from the actual UI schema definitions.

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
...
+const emailProperties = {
+  subject: { type: 'emailSubject' },
+  body: { type: 'emailBody' },
+};
+
+const inAppProperties = {
+  subject: { type: 'inAppSubject' },
+  body: { type: 'inAppBody' },
+  avatar: { type: 'inAppAvatar' },
+  primaryAction: { type: 'inAppPrimaryAction' },
```
tatarco, line 19: I wanted to clarify the reasoning behind our approach. The goal was not to create a one-to-one mapping but rather to specify particular overrides for the component type. Having a complete mapping may not be necessary, as it requires the front-end to duplicate specific elements for each type of step, which can lead to unnecessary complexity.
Could we revisit this to ensure it aligns with our original objectives?

apps/api/src/app/step-schemas/usecases/get-step-schema/get-step-schema.usecase.ts
```diff
@@ -9,24 +9,37 @@ import {
...
+      }
+
+      if (!currentStep.template?.controls?.schema) {
+        throw new BadRequestException('No controls schema found');
+      }
+
       return {
-        controls: buildControlsSchema({ controlsSchema: currentStep.template?.controls?.schema }),
+        controls: buildControlsSchema({
+          stepType: currentStep.template?.type as StepType,
```
rifont: Why was this type-coercion needed? Can it be fixed upstream in the types for `currentStep`?

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -5,9 +5,23 @@ import { WorkflowResponseDto } from './workflow-response-dto';
...
+  controlGroup?: UiControlGroupEnum.INBOX;
+  elements?: UiElement[];
+}
+class UiElement {
+  elementTypeOverride: UiElementTypeEnum;
+}
 
+enum UiElementTypeEnum {
+  EMAIL_EDITOR = 'EMAIL_EDITOR',
+}
```
rifont, line 21: Should all of these be deleted with the new approach?

rifont: Should all of these be deleted with the new approach?
```suggestion
```

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
...
+  secondaryAction: { type: 'inAppSecondaryAction' },
+  data: { type: 'inAppData' },
+  redirect: { type: 'inAppRedirect' },
+};
+
+const pushProperties = {
+  subject: { type: 'pushSubject' },
+  body: { type: 'pushBody' },
```
tatarco, line 27: why not just text? it's going to make the FE work super problematic

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
...
+};
+
+const smsProperties = {
```
tatarco, line 30: also it's not correct, because in code first they can have as many controls as they want, they shouldn't be limited

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
djabarovgeorge: thanks i implemented the dx guide :)

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
djabarovgeorge: i agree. I was not sure what approach to take here: simpler or more generic. you mean to convert the type to a generic type instead. did I get you right?

packages/shared/src/dto/workflows/workflow-commons-fields.ts
djabarovgeorge: i will, i thought to delete it separately while removing the controls from the /workflows response

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
djabarovgeorge: i think we still need support for code first, we will need to store uiSchema on sync in the database and then extract it here.

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
djabarovgeorge: as discussed in the meeting, in this solution, the `client knows something` rather than `nothing` the second solution suggested in the DX guide. Here, the client needs to know how to render `pushBody`. In the future, we can expand this approach to provide even more data to the client if needed.

apps/api/src/app/step-schemas/dtos/step-schema.dto.ts
```diff
@@ -1,6 +1,83 @@
...
+  | typeof chatProperties
+  | typeof emailProperties
+  | typeof inAppProperties
+  | typeof pushProperties
+  | typeof smsProperties
+  | typeof delayProperties
+  | typeof digestProperties
+  | typeof customProperties;
+
+export type UiSchema = {
```
SokratisVidros: We can use discriminated unions and drastically simplify the type definition. See [here](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions).

That is we create a union for each step type and we don't need the map anymore.

SokratisVidros (approved):

PR #6766, opened 2024-10-24
fix(framework): Resolve CJS issues this time with json-schema-faker

### LINKED ISSUE #6332: 🐛 Bug Report: Cannot find module '@novu/framework/next' or its corresponding type declarations.ts
### 📜 Description

Cannot find module '@novu/framework/next' or its corresponding type declarations.ts

### 👟 Reproduction steps
import { serve } from "@novu/framework/next";

"@novu/framework": "2.0.0-canary.5"

### 👍 Expected behavior

Typescript support

Unable to import from '@novu/framework/next'

### Novu version

2.0.0-canary.5

### npm version

10.5.0

### node version

v18.20.1

_No response_

### LINKED ISSUE #6701: 🚀 Feature: @novu/framework support for commonjs modules
### 🔖 Feature description

Add .cjs modules into the build of the packages.

Currently the packages are not supporting commonjs, which is still widly used.

f.ex. https://stackoverflow.com/questions/74937600/how-to-support-es-modules-and-commonjs-modules-at-the-same-time

### 🔄️ Additional Information

maybe this would also be helpfull for other packages.

None

SokratisVidros:

Apply more configuration changes for better ESM/CJS support. Also try to remove the Node `https` dependency from `json-schema-faker` 

Fixes [#6701](https://github.com/novuhq/novu/issues/6701)
Fixes https://github.com/novuhq/novu/issues/6332


packages/framework/express/package.json
```diff
@@ -1,5 +1,3 @@
 {
-  "main": "../dist/cjs/servers/express.cjs",
```
SokratisVidros, line 2: After careful testing, these are not required as ESM sub-exports will always be fetched from the top-level package.json

packages/framework/package.json
```diff
@@ -239,7 +239,6 @@
     "better-ajv-errors": "^1.2.0",
     "chalk": "^4.1.2",
     "cross-fetch": "^4.0.0",
-    "json-schema-faker": "^0.5.6",
```
SokratisVidros, line 242: This package is causing a lot of ESM/CJS trouble. It was decided to hardcode its UMD version in the framework.

packages/framework/package.json
```diff
@@ -37,7 +37,7 @@
     "format:fix": "prettier --write --ignore-path .gitignore .",
     "build": "NODE_ENV=production tsup",
     "build:watch": "tsup --watch",
-    "postbuild": "./scripts/copy-package-json.sh && pnpm run check:exports && pnpm check:circulars",
```
SokratisVidros, line 40: Copy-package.json is not anymore assuming package exports work correctly. Attw confirms the removal.

packages/framework/tsup.config.ts
```diff
@@ -1,18 +1,22 @@
...
+import { version } from './package.json';
+import { type SupportedFrameworkName } from './src/internal';
 
 const frameworks: SupportedFrameworkName[] = ['h3', 'express', 'next', 'nuxt', 'sveltekit', 'remix', 'lambda', 'nest'];
 
 const baseConfig: Options = {
   entry: ['src/index.ts', 'src/internal/index.ts', ...frameworks.map((framework) => `src/servers/${framework}.ts`)],
   sourcemap: false,
   clean: true,
-  treeshake: true,
```
SokratisVidros, line 10: Libraries need to be treeshaken by the host application bundler.

packages/framework/tsup.config.ts
```diff
@@ -1,18 +1,22 @@
...
   dts: true,
   minify: true,
```
SokratisVidros, line 12: Libraries shouldn't be minified. The host application bundler should handle that. 

However, setting `bundle: false` and `minify: false` in `tsup` config, breaks Nest.js loading @novu/framework. 

The hypothesis is that building through ESBuild alone is not enough. Minify triggers a rollup processing that, for some reason, satisfies Nest.js CommonJS requirements. 

We will get back to this in a separate PR, as this one is already big enough.

packages/framework/tsup.config.ts
```diff
@@ -1,18 +1,22 @@
...
   minifyWhitespace: true,
   minifyIdentifiers: true,
   minifySyntax: true,
+  define: {
+    SDK_VERSION: `"${version}"`,
```
SokratisVidros, line 17: Static code replacements allow us to eliminate importing `package.json` on runtime that can cause some CJS/ESM issues.

packages/framework/tsup.config.ts
rifont: Thanks for the detailed explanation. Sounds like a good plan 👍

packages/framework/tsup.config.ts
rifont: Awesome! 👍

rifont (approved):

PR #6767, opened 2024-10-24
feat(api): Delete subscriber channel preference when updating global channel

rifont:

* API
  * Delete subscriber channel preference when updating global channel
    * This ensures that the global preference can take priority again after updating the global preference
  * Return only the active & non-critical channels for Global Subscriber preferences
    * Previously all channels were shown regardless of the active or critical state of the Workflow. This change ensures the Subscriber only sees channels relevant to them, as all the other channels are not relevant and may cause confusion
  * Persist only the provided values for all preference sets with the exception of Workflow Resource preferences, to ensure that `PATCH` operations for preferences don't create preferences for non-relevant channels
  * Refactoring
    * Rename `GetPreferences` for Inbox to `GetInboxPreferences` to eliminate naming duplication with the generic `GetPreferences` use-case
    * Refactor preference use-cases to eliminate duplication and ease transition to V2 Preferences
  * **_TODO: should we throw a `400 Bad Request` response when attempting to modify global preference channels that don't have active & non-critical workflows that would be affected? Right now, the `PATCH` responses for global preferences return `undefined` for channels that are irrelevant and it doesn't make much sense. We have failing tests that need to be addressed with the outcome of this decision._**
* `<Inbox />`
  * Optimistically update Subscriber Workflow preferences in `<Inbox />` when modifying the Subscriber Global preferences, to ensure that the unset Workflow preferences display immediately.

### Screenshots

_Toggling the Subscriber Global Channel Preferences in `<Inbox />` modifies the corresponding Subscriber Workflow Channel Preference fields._
**Important Note:** This PR only has changes to reflect local state changes optimistically, it does not refetch all Preferences from the API after the optimistic update.

https://github.com/user-attachments/assets/27dbd5b4-c3b7-4460-a898-48fa84fc3e19

_`<Inbox />` displays only Global Preference channels for active & non-critical workflow channels_
image

<details>

</details>


libs/dal/src/repositories/preferences/preferences.schema.ts
rifont, line 1: Removing defaults to ensure that only the provided mutation values are persisted, enabling `PATCH` operations to work as expected.

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -323,6 +329,10 @@ export class Sync {
     return notificationGroupId;
   }
 
+  private getWorkflowPreferences(workflow: DiscoverWorkflowOutput): WorkflowPreferencesPartial {
+    return workflow.preferences || {};
```
rifont, line 333: A fallback empty object is now necessary during sync after strengthening validation of the `UpsertWorkflow` use-case to disallow `undefined` preferences.

apps/api/src/app/inbox/inbox.controller.ts
rifont, line 1: Here and in the following files, we'll see renaming from `GetPreferences` to `GetInboxPreferences` to eliminate the duplicated usage of `GetPreferences` use-case naming.

apps/api/src/app/inbox/usecases/get-inbox-preferences/get-inbox-preferences.usecase.ts
rifont, line 1: This is a slightly bad diff, as the file is not entirely new. This use-case is however now considerably simpler, owing to better encapsulation of both:
* `GetSubscriberGlobalPreference` filtering out channels in the response to include only active & non-critical channels
* `GetSubscriberPreference` filtering out preference sets to include only active & non-critical workflows
* Full encapsulation of V1 & V2 preferences in both of the above use-cases

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -230,18 +193,17 @@ export class UpdatePreferences {
...
     _subscriberId: string;
     environmentId: string;
     templateId?: string;
-  }) {
+  }): Promise<void> {
     const preferences: WorkflowPreferencesPartial = {
-      all: {
-        enabled: PREFERENCE_DEFAULT_VALUE,
-        readOnly: false,
-      },
```
rifont, line 244: There is no "all" channel preferences capability in Inbox today, so we won't persist anything.

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
rifont, line 1: This file just has some cleanup to ensure a channel preference object is built without containing `undefined` values. This is necessary after adding stricter validation to the `UpsertPreferences` use-case.

apps/api/src/app/shared/commands/base.command.ts
rifont, line 1: The changes here (and the exact same change for `BaseCommand` in `@novu/application-generic`) are made to ensure that the `ValidateIf` usage in the `UpsertPreferencesCommand` works as expected.

Without this change, when an  empty object is used for the `UpsertPreferencesCommand.preferences` value,  `Object.values(item.constraints)` encounters a type error due to an `undefined` value being used.

libs/application-generic/src/commands/base.command.ts
```diff
@@ -1,30 +1,35 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
```
rifont, line 1: 🧹, same as the `BaseCommand` in API. Also, we should remove the duplication of these `BaseCommand`'s at some point.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
rifont, line 1: I'm 99% sure that this change will step on @ChmaraX 's toes, it's also needed after improving the `UpsertPreferences` use-case validation to disallow `undefined` preferences.

We will need to come back to this as we don't want to create preferences for the user when it wasn't intended, ensuring that the Preference "override" UI will function as expected on initial Workflow load.

libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.command.ts
```diff
@@ -2,6 +2,4 @@ import { SubscriberEntity } from '@novu/dal';
 
 import { EnvironmentWithSubscriber } from '../../commands';
 
-export class GetSubscriberGlobalPreferenceCommand extends EnvironmentWithSubscriber {
-  subscriber?: SubscriberEntity;
```
rifont, line 6: This was never used.

libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
```diff
@@ -6,78 +6,121 @@ import {
...
+      await this.getPreferences.getPreferenceChannels({
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        subscriberId,
+      });
+
+    const subscriberGlobalPreference =
+      subscriberGlobalPreferenceV2 ??
+      subscriberGlobalPreferenceV1?.channels ??
+      {};
```
rifont: There is a case where no global preferences have been stored for the Subscriber, so we need to fallback to an empty preference set.

libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
```diff
@@ -6,78 +6,121 @@ import {
...
-      subscriberChannelPreference ?? {},
+    const channelsWithDefaults = this.buildDefaultPreferences(
+      subscriberGlobalPreference,
     );
 
+    const channels = filteredPreference(channelsWithDefaults, activeChannels);
+
     return {
       preference: {
-        enabled: subscriberPreference?.enabled ?? true,
```
rifont, line 54: TODO: I will fix this in this PR

libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
```diff
@@ -6,78 +6,121 @@ import {
...
+
+    const subscriberGlobalPreference =
+      subscriberGlobalPreferenceV2 ??
+      subscriberGlobalPreferenceV1?.channels ??
+      {};
+
+    return subscriberGlobalPreference;
+  }
+
+  private async getActiveChannels(
```
rifont, line 89: This is the change in Subscriber Global Preference fetching that is responsible for ensuring that only active channels are shown in the Inbox/NotificationCenter preferences.

libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
```diff
@@ -6,78 +6,121 @@ import {
...
+
+    if (!subscriber) {
+      throw new ApiException(`Subscriber ${command.subscriberId} not found`);
+    }
+
+    return subscriber;
   }
   // adds default state for missing channels
-  private updatePreferenceStateWithDefault(preference: IPreferenceChannels) {
+  private buildDefaultPreferences(preference: IPreferenceChannels) {
```
rifont, line 135: Better naming.

libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
rifont, line 1: Here we see updates to better encapsulate the fallback pattern for V1 & V2 preferences, alongside a change to return only active global preference channels.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -58,5 +63,11 @@ export class GetSubscriberPreference {
         ),
       ),
     );
+
+    const nonCriticalWorkflowPreferences = subscriberWorkflowPreferences.filter(
+      (preference) => preference.template.critical === false,
+    );
```
rifont, line 69: There are no use-cases where it's necessary for preferences belonging to critical workflows to be returned, so we're filtering them out here now. If a use-case arises for returning critical preferences, we should add a command property to control that.

libs/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
rifont, line 1: Similar to subscriber global preference fetching, the changes here improve readability and better encapsulate the logic for the fallback pattern of V1 & V2 Preferences.

libs/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -108,26 +71,71 @@ export class GetSubscriberTemplatePreference {
...
+      ...(subscriberWorkflowChannelPreferences.critical !== undefined && {
+        critical: subscriberWorkflowChannelPreferences.critical,
       }),
     });
 
     return {
       template,
       preference: {
-        enabled: subscriberPreference?.enabled ?? true,
+        enabled: true,
```
rifont: TODO: will fix in this PR

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.command.ts
rifont, line 1: The changes here are purely to strengthen validation for the `UpsertPreferences` use-case, which is used widely across the API for Workflow syncing, User preference updates (via Dashboard), and both Subscriber Global & Workflow updates via Inbox.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -35,6 +43,32 @@ export class UpsertPreferences {
     });
   }
 
+  private async deleteSubscriberChannelPreferences(
```
rifont: This method is responsible for "forgetting" the subscriber workflow preferences for relevant channels when the subscriber modifies a global preference for the same channels.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -70,18 +104,11 @@ export class UpsertPreferences {
       return this.deletePreferences(command, foundId);
     }
 
-    const builtPreferences = buildWorkflowPreferences(command.preferences);
-
-    const builtCommand = {
-      ...command,
-      preferences: builtPreferences,
-    };
-
```
rifont: We no longer build a full preference set for all Preference types, ensuring that `PATCH` operations mutate only the minimal preference channels.

libs/dal/src/repositories/preferences/preferences.schema.ts
rifont, line 1: As above in `UpsertPreferences`, we only store the minimal provided preferences now to support `PATCH` operations. Therefore it's necessary to remove all defaults from the DAL.

libs/application-generic/src/usecases/upsert-preferences/upsert-subscriber-global-preferences.command.ts
rifont, line 1: More stronger validation for `UpsertPreferences` in the next 4 files.

Additionally, the Preference type specific commands now extend the `UpsertPreferencesBaseCommand` enforces contains strong validation for the `preferences` property.

libs/application-generic/src/usecases/upsert-preferences/upsert-subscriber-global-preferences.command.ts
```diff
@@ -1,12 +1,8 @@
...
-  @IsDefined()
-  readonly preferences: WorkflowPreferencesPartial;
+import { IsMongoId, IsNotEmpty } from 'class-validator';
+import { UpsertPreferencesBaseCommand } from './upsert-preferences.command';
 
+export class UpsertSubscriberGlobalPreferencesCommand extends UpsertPreferencesBaseCommand {
   @IsNotEmpty()
   @IsMongoId()
-  _subscriberId: string;
+  readonly _subscriberId: string;
```
rifont, line 7: `readonly` ensures the command can't be mutated after validation.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.command.ts
```diff
@@ -1,17 +1,108 @@
-import { IsDefined, IsEnum } from 'class-validator';
-import { PreferencesTypeEnum, WorkflowPreferencesPartial } from '@novu/shared';
+import {
```
rifont, line 1: The changes here are purely to strengthen validation for the `UpsertPreferences` use-case, which is used widely across the API for Workflow syncing, User preference updates (via Dashboard), and both Subscriber Global & Workflow updates via Inbox.

libs/dal/src/repositories/preferences/preferences.schema.ts
```diff
@@ -32,42 +32,35 @@ const preferencesSchema = new Schema<PreferencesDBModel>(
       all: {
```
rifont, line 32: As above in `UpsertPreferences`, we only store the minimal provided preferences now to support `PATCH` operations. Therefore it's necessary to remove all defaults from the DAL.

apps/api/src/app/inbox/e2e/get-preferences.e2e.ts
```diff
@@ -9,22 +10,28 @@ describe('Get all preferences - /inbox/preferences (GET)', function () {
...
-  it('should get both global and workflow preferences if workflow is present', async function () {
+  it('should get both global preferences for active channels and workflow preferences if workflow is present', async function () {
     await session.createTemplate({
       noFeedId: true,
+      steps: [
+        {
+          type: StepTypeEnum.EMAIL,
+          content: 'Test notification content',
+        },
+      ],
```
rifont, line 34: Explicitly specifying that only email is present, to assert that `in_app` should not appear.

.github/actions/setup-project/action.yml
```diff
@@ -39,9 +39,9 @@ runs:
 
     - name: 📚 Start MongoDB
       if: ${{ inputs.slim == 'false' }}
-      uses: supercharge/mongodb-github-action@v1.9.0
+      uses: supercharge/mongodb-github-action@v1.11.0
       with:
-        mongodb-version: 4.2.8
+        mongodb-version: 5.0.29
```
rifont, line 44: This MongoDB version bump to >5 was needed to solve [this runtime error](https://github.com/novuhq/novu/actions/runs/11514455656/job/32053213154?pr=6767#step:9:1644) that disallows setting empty objects:
> `Invalid $addFields :: caused by :: an empty object is not a valid value. Found empty object at path preferences.all`

Version 5 supports empty object setting per [this thread and referenced documentation](https://www.mongodb.com/community/forums/t/mongoservererror-invalid-set-caused-by-an-empty-object-is-not-a-valid-value/148344/4). 

`5.0.29` is the latest LTS 5.x release per the MongoDB [changelog](https://www.mongodb.com/docs/manual/release-notes/5.0/#5.0.29---sep-30--2024). MongoDB 5.0 reaches end of life this month, [October 2024](https://www.mongodb.com/legal/support-policy/lifecycles#:~:text=July%202021-,October,-2024).

**Note, this means that we are effectively dropping support of MongoDB v4 for Novu Self-hosted. MongoDB v4 reached [end of life in February 2024](https://www.mongodb.com/legal/support-policy/lifecycles#:~:text=July%202020-,February%202024,-MongoDB%205.0).**

cc @merrcury

merrcury: Noted, Thanks 👍🏼. We will post a note a to community/

apps/api/src/app/inbox/usecases/get-inbox-preferences/get-inbox-preferences.spec.ts
rifont, line 1: This was a bad diff - it has an extra test spec which caused it to become bad after renaming from `get-preferences` -> `get-inbox-preferences`

.github/actions/setup-project/action.yml
SokratisVidros: Good call out!

libs/dal/src/repositories/preferences/preferences.schema.ts
SokratisVidros: 👍

SokratisVidros (approved):

packages/js/src/preferences/preferences.ts
```diff
@@ -62,10 +64,4 @@ export class Preferences extends BaseModule {
       }
     });
   }
-
-  async update(args: UpdatePreferencesArgs): Result<Preference> {
-    return this.callWithSession(async () =>
-      updatePreference({ emitter: this._emitter, apiService: this._inboxService, args })
-    );
-  }
```
rifont, line 70: This `update` method on the full `Preferences` resource was never used, so I'm deleting it.

PR #6771, opened 2024-10-25
feat(api): add promote workflow endpoint

ChmaraX:
* added SyncToEnvironment usecase
* added promote workflow endpoint
* added basic e2e tests to cover the endpoint
* fixed few bugs in other places linked to workflow CRUD


apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
```diff
@@ -34,6 +34,7 @@ export class GetWorkflowUseCase {
       GetPreferencesCommand.create({
         environmentId: command.user.environmentId,
         organizationId: command.user.organizationId,
+        templateId: workflowEntity._id,
```
ChmaraX, line 36: We should provide `templateId` when retrieving specific workflow, otherwise it would return any preference from the org and environment I guess.
(existing e2e test was failing due to this)

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -0,0 +1,160 @@
...
+    const upsertedWorkflow = await this.upsertWorkflowUseCase.execute(
+      UpsertWorkflowCommand.create({
+        user: { ...command.user, environmentId: command.targetEnvironmentId },
+        identifierOrInternalId: existingWorkflow?._id,
+        workflowDto,
+      })
+    );
+
+    return upsertedWorkflow;
+  }
```
ChmaraX: With some modification, this usecase could in theory be reused in Bridge sync, since the logic is very similar. 
This usecase would need to accept array of workflows instead of workflowId.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -123,15 +123,13 @@ export class UpsertWorkflowUseCase {
     command: UpsertWorkflowCommand,
     workflow: NotificationTemplateEntity
   ): Promise<GetPreferencesResponseDto | undefined> {
-    if (!command.workflowDto.preferences?.user) {
```
ChmaraX, line 128: Removed this because we want to always execute `upsertPreferences`, but in case it gets empty preferences, it will insert default ones.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -123,15 +123,13 @@ export class UpsertWorkflowUseCase {
...
-      return undefined;
-    }
-    await this.upsertPreferences(workflow, command);
+    await this.upsertUserWorkflowPreferences(workflow, command);
+    await this.upsertWorkflowPreferences(workflow, command);
```
ChmaraX, line 130: Added this to always insert both user and workflow preferences, in case they're not provided, we create default ones.

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -193,6 +193,116 @@ describe('Workflow Controller E2E API Testing', () => {
       expect(idsDeduplicated.size).to.be.equal(10);
     });
   });
+
+  describe('Promote Workflow Permutations', () => {
```
ChmaraX, line 228: Given the fact that worfklow creation and promotion in general is a production critical part, I think it would make sense to cover `upsert-workflow.usecase.ts` with usecase specific tests to check if all entities are being created and updated correctly, by explicitly checking repositories.

These e2e tests serves more like a smoke tests, but they can't verify the functionality properly here.

I will create another task for this and add this in separate PR.

apps/api/src/app/workflows-v2/workflow.controller.ts
```diff
@@ -63,6 +67,23 @@ export class WorkflowController {
...
+  @UseGuards(UserAuthGuard)
+  async promote(
+    @UserSession() user: UserSessionData,
+    @Param('workflowId') workflowId: string,
+    @Body() promoteWorkflowDto: PromoteWorkflowDto
+  ): Promise<WorkflowResponseDto> {
+    return this.syncToEnvironmentUseCase.execute(
+      SyncToEnvironmentCommand.create({
+        identifierOrInternalId: workflowId,
+        targetEnvironmentId: promoteWorkflowDto.targetEnvironmentId,
```
ChmaraX, line 87: I could wrap this usecase into `PromoteWorkflowUsecase` where I would retrieve environment with name "Production" to avoid the need for user to provide `targetEnvironmentId`. But I think this is a bit more flexible and we don't need additional usecase.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -61,12 +61,12 @@ export class UpsertPreferences {
...
-  private async upsert(
+  public async upsert(
     command: UpsertPreferencesCommand,
   ): Promise<PreferencesEntity> {
     const foundId = await this.getPreferencesId(command);
 
-    if (command.preferences === null) {
+    if (command.preferences === null && foundId) {
       return this.deletePreferences(command, foundId);
     }
```
ChmaraX: `this.deletePreferences` needs `foundId` for deletion. If the `foundId` is undefined, it throws and error.

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -80,4 +80,5 @@ export type PreferencesResponseDto = {
 
 export type PreferencesRequestDto = {
   user: WorkflowPreferences | null;
+  workflow?: WorkflowPreferences | null;
```
ChmaraX, line 81: In case of promotion, we want to clone also workflow preferences from dev -> prod.

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -0,0 +1,160 @@
...
+    private getWorkflowByIdsUseCase: GetWorkflowByIdsUseCase,
+    private preferencesRepository: PreferencesRepository,
+    private upsertWorkflowUseCase: UpsertWorkflowUseCase
+  ) {}
+
+  async execute(command: SyncToEnvironmentCommand): Promise<WorkflowResponseDto> {
+    const workflowToClone = await this.getWorkflowToClone(command);
+    const preferencesToClone = await this.getWorkflowPreferences(workflowToClone._id, workflowToClone._environmentId);
+    const externalId = workflowToClone.triggers[0].identifier;
+    const existingWorkflow = await this.findExistingWorkflow(command, externalId);
```
LetItRock:
```suggestion
    const existingWorkflow = await this.findWorkflowInTargetEnvironment(command, externalId);
```

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -193,6 +193,116 @@ describe('Workflow Controller E2E API Testing', () => {
...
+      expect(prodWorkflow.steps).to.have.lengthOf(devWorkflow.steps.length);
+      prodWorkflow.steps.forEach((prodStep, index) => {
+        const devStep = devWorkflow.steps[index];
+        expect(prodStep.stepUuid).to.not.equal(devStep.stepUuid, 'Step UUID should be different');
+        expect(prodStep.controlValues).to.deep.equal(devStep.controlValues, 'Step controlValues should match');
+        expect(prodStep.name).to.equal(devStep.name, 'Step name should match');
+        expect(prodStep.type).to.equal(devStep.type, 'Step type should match');
+      });
+    });
+
```
LetItRock, line 270: I guess we should not allow promotion in the same environment, right? if yes, let's add the test and validation

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -0,0 +1,163 @@
...
+      /*
+       * If we are updating an existing workflow, we need to map the updated steps to the existing steps
+       * (!) 'existingWorkflowSteps' are from a different environment than 'steps' - the only thing that doesn't change
+       *  in steps across environments is the stepId (TODO)
+       */
+      ...(existingWorkflowSteps && {
+        _id:
+          existingWorkflowSteps.find((existingStep) => existingStep.stepId === step.stepId)?._templateId ??
+          step._templateId,
+      }),
```
ChmaraX, line 139: This doesn't matter yet, because `stepId` is not kept the same between the same workflow in different environments, therefore steps are always deleted & recreated.

Once we make `stepId` (or other step property) immutable, this will work as expected

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -207,7 +223,7 @@ export class UpsertWorkflowUseCase {
 
     return {
       id: existingWorkflow._id,
-      environmentId: user.environmentId,
+      environmentId: existingWorkflow._environmentId,
```
ChmaraX, line 235: This needs to be the environment where the workflow that we are updating is stored, not the user current environment.

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
rifont: Good catch!

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -225,6 +225,130 @@ describe('Workflow Controller E2E API Testing', () => {
...
+    it('should throw an error if the workflow to promote is not found', async () => {
+      const res = await session.testAgent.put(`${v2Prefix}/workflows/123/promote`).send({ targetEnvironmentId: '123' });
+
+      expect(res.status).to.equal(404);
+      expect(res.body.message).to.equal('Workflow cannot be found');
+      expect(res.body.workflowId).to.equal('123');
+    });
+  });
+
```
rifont, line 351: Nice tests ✨

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
rifont: You're spot on with the `foundId` part, I made the same change in my PR.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -61,12 +61,12 @@ export class UpsertPreferences {
     });
   }
 
-  private async upsert(
+  public async upsert(
```
rifont: ⚠ **issue:** Only the `execute` method should be `public` in use-cases. We should revert this change and use `execute` as needed by passing in the necessary command properties to control the use-case behaviour.‏
```suggestion
  private async upsert(
```

rifont: I see that other methods are `public` in the `UpsertPreferences` - I will address those in a follow-up PR, but we should avoid increasing the problem for now.

rifont (approved):
Very clean! Nice work 🎉

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
djabarovgeorge, line 1: How would we feel about renaming `sync` to something else? I think using `sync` here might be confusing since we already use it to sync data from the bridge app (like workflows, preferences, etc.). 

This is similar but has its own differences, so I wonder if it would be better to avoid reusing the same term.

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -0,0 +1,167 @@
...
+      workflow: preferences.find((pref) => pref.type === PreferencesTypeEnum.WORKFLOW_RESOURCE)?.preferences ?? null,
+    };
+  }
+
+  private async getWorkflowPreferences(workflowId: string, environmentId: string): Promise<PreferencesEntity[]> {
+    const workflowPreferences = await this.preferencesRepository.find({
+      _templateId: workflowId,
+      _environmentId: environmentId,
+      type: {
+        $in: [PreferencesTypeEnum.WORKFLOW_RESOURCE, PreferencesTypeEnum.USER_WORKFLOW],
```
djabarovgeorge, line 164: question because i am missing the context, do we allow to `:workflowId/promote` code-first workflows as well?

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: thought: agree on `user` preference default upsert but, i wonder if we need to create a default `workflow` preference for users that do not use code-first, feels a bit redundant and misleading data.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: good point, besides the tests i would add a const with a comment to prevent future maintainers break it while refactoring the code. 
not a deal breaker just as a thought for your consideration.

apps/api/src/app/workflows-v2/workflow.controller.ts
djabarovgeorge: i think we will have only Prod to promote to in the near-far future but as long as the API client knows the prod env i think this approach is good enough and as you said more flexible let's keep it for now.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
djabarovgeorge: great mind think alike 😉

djabarovgeorge (commented):
amazing work, super clean!

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
ChmaraX: I think for now we agreed to **not** promote code-first workflows using dashboard UI, because they were some open questions and issues with that.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
ChmaraX: Oh, I was using upsert directly at some point then I removed that implementation and forgot to put the method back to private.

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
ChmaraX: I agree its not 100% clear what it does, however I was thinking maybe with some small adjustments we could use it also in in the place of bridge sync, then the naming would be alright I guess.

I'm was thinking about `promote.usecase` but thats less flexible or maybe something with `upsert` or `clone`.

PR #6772, opened 2024-10-25
feat(api): Complete email preview logic

tatarco:
Completed the work around maily hydration and expansion, including modifying validation and preparation for payload schema restructuring 

### Screenshots

<details>

</details>


.source
djabarovgeorge, line 1: if we do not have ee changes in this pr, we should not have this hash changes

SokratisVidros: Correct, this needs to be reverted.☝️

apps/api/src/app/environments-v1/usecases/output-renderers/email-schema-expander.usecase.ts
```diff
@@ -2,76 +2,146 @@
...
   }
 
-  private processItemNode(node: TipTapNode, item: any): TipTapNode {
-    if (node.type === 'text' && typeof node.text === 'string') {
-      const regex = /{#item\.(\w+)#}/g;
-      node.text = node.text.replace(regex, (_, key: string) => {
-        const propertyName = key;
-
-        return item[propertyName] !== undefined ? item[propertyName] : _;
+  private traverseAndAugment(node: TipTapNode, parentNode?: TipTapNode) {
```
SokratisVidros: Nitpicky comment to improve readability and self-documentation

```suggestion
  private processShowAndForControls(node: TipTapNode, parentNode?: TipTapNode) {
```

apps/api/src/app/environments-v1/usecases/output-renderers/email-schema-expander.usecase.ts
```diff
@@ -2,76 +2,146 @@
...
+    } else if (this.hasEach(node)) {
+      const newContent = this.expendedForEach(node);
+      node.content = newContent;
+      this.removeForNodeAndUpgradeContent(node, newContent, parentNode);
+    }
+  }
 
-    if (node.content) {
-      node.content = node.content.map((innerNode) => this.processItemNode(innerNode, item));
+  private removeForNodeAndUpgradeContent(node: TipTapNode, expandedContents: TipTapNode[], parentNode?: TipTapNode) {
```
SokratisVidros: Another one: I would move this code into `processShowAndForControls` and keep all that processing in a single function as a single unit of work.

apps/api/src/app/environments-v1/usecases/output-renderers/email-schema-expander.usecase.ts
```diff
@@ -2,76 +2,146 @@
...
+      const hydratedContent = this.replacePlaceholders(templateContent, value);
+      expandedContent.push(...hydratedContent);
     }
 
-    if (this.hasEachAttr(node)) {
-      return { type: 'section', content: this.handleFor(node) };
+    return expandedContent;
+  }
+
+  private isOrderedList(templateContent: TipTapNode[]) {
```
SokratisVidros: Please add some comments about the extra handling we do here for bullets and ordered list and why it's necessary.

apps/api/src/app/environments-v1/usecases/output-renderers/email-output-renderer.usecase.ts
```diff
@@ -1,44 +1,49 @@
 import { EmailRenderOutput, TipTapNode } from '@novu/shared';
```
SokratisVidros: Our use cases are named with verbs. So this one should be `render-email-output.usecase.ts`. I just saw that the names are also off for other steps. We can fix them in a follow up PR, it's not urgent.

apps/api/src/exception-filter.ts
```diff
@@ -133,10 +147,34 @@ function handleZod(exception: ZodError<any>) {
 
   return { status, message };
 }
+function handleMongo(exception: MongoServerError) {
```
SokratisVidros: Is this change relevant to the email preview logic? If no it should have been a separate PR.

apps/api/src/app/workflows-v2/usecases/placeholder-enrichment/payload-preview-value-generator.usecase.ts
```diff
@@ -22,20 +18,29 @@ export class CreateMockPayloadUseCase {
...
-  private buildPayloadForEmailEditor(controlValue: unknown): Record<string, unknown> {
-    const collectPlaceholderMappings = this.collectPlaceholdersFromTipTapSchemaUsecase.execute({
-      node: controlValue as TipTapNode,
-    });
-    const transformPlaceholderMap = this.transformPlaceholderMapUseCase.execute({ input: collectPlaceholderMappings });
+  private safeAttemptToParseEmailSchema(controlValue: string) {
+    try {
+      const { nestedPayload } = this.hydrateEmailSchemaUseCase.execute({
+        emailEditor: controlValue,
+        masterPayload: {
```
SokratisVidros: This concept of `masterPayload` is a bit confusing as it makes the reader believe there are different types of payloads in the app. How about passing payload, subscriber, and steps directly as arguments?

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -21,6 +21,12 @@ import {
   SmsOutputRendererUsecase,
 } from '../output-renderers';
 
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export interface MasterPayload {
```
SokratisVidros: See my comment later about introducing this new term of MasterPayload.

apps/api/src/app/environments-v1/usecases/output-renderers/expend-email-editor-schema-command.ts
```diff
@@ -3,7 +3,6 @@
 import { BaseCommand } from '@novu/application-generic';
```
SokratisVidros: Please rename to `expand-email-editor-schema.command.ts`. There is a typo in the current filename.

apps/api/src/app/environments-v1/usecases/output-renderers/email-schema-expander.usecase.ts
```diff
@@ -2,76 +2,146 @@
 import { TipTapNode } from '@novu/shared';
```
SokratisVidros: Ditto about renaming the use case to `expand-email-schema.usecase.ts`

apps/api/src/app/environments-v1/usecases/output-renderers/hydrate-email-schema.usecase.ts
```diff
@@ -0,0 +1,214 @@
+/* eslint-disable no-param-reassign */
+import { Injectable } from '@nestjs/common';
+import { TipTapNode } from '@novu/shared';
+import { z } from 'zod';
+import { MasterPayload } from '../construct-framework-workflow';
+import { HydrateEmailSchemaCommand } from './hydrate-email-schema.command';
+
+@Injectable()
+export class HydrateEmailSchemaUseCase {
```
SokratisVidros, line 8: One thing that sounds a bit confusing is why do we have hydrate and expand email use cases separately. I'd expect a single use case that does both.

SokratisVidros (changes requested):

apps/api/src/app/workflows-v2/usecases/generate-preview/construct-payload-from-placeholders-with-defaults-use-case.service.ts
djabarovgeorge, line 1: can we find a better location for this file? The route folder should be kept clean, containing only module file, controller, and folders.

packages/shared/src/dto/step-schemas/generate-preview-response.dto.ts
```diff
@@ -59,8 +59,14 @@ export class ControlPreviewIssue {
   variableName?: string;
   message: string;
 }
+export class MasterPayload {
```
djabarovgeorge: what do you think about changing the name here to something else?

```suggestion
export class MainPayload {
```

apps/api/src/app/environments-v1/usecases/output-renderers/email-schema-expander.usecase.ts
SokratisVidros: Reminder for this.

packages/shared/src/dto/step-schemas/generate-preview-response.dto.ts
SokratisVidros: I agree. The name is too generic and introduces a master payload concept that doesn't exist in the product. As said earlier, we should eliminate it https://github.com/novuhq/novu/pull/6772/files#r1827360737

apps/api/src/app/workflows-v2/usecases/placeholder-enrichment/payload-preview-value-generator.usecase.ts
tatarco: just because there will be more, we havn't done actor yet,. that's the reason I added this concept, because its a higher level payload that includes all not only the customer provided payload, I want to minimize the changes when we add actor and others to the framework, create a structure that minimizes the maintenance, 
eventually the placeholders are not reffering to payload, they reffer to such a master payload.

packages/shared/src/dto/step-schemas/generate-preview-response.dto.ts
tatarco: is main payload exists?

SokratisVidros: This is just a grouping for which we don't have a strong argument yet. I suggest we use separate parameters for payload, subscriber, and steps, as they all have a very specific meaning in Novu land.

If we need a new umbrella term, we will add it in the future in a more structured way.

apps/api/src/exception-filter.ts
tatarco: i'll revert it

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
tatarco: lets talk about it, I would like to find another solution but having to maintain all seperatly passing them in and out of functions, we should have a term for the all inclusive payload including us and the customers

apps/api/src/app/environments-v1/usecases/output-renderers/email-schema-expander.usecase.ts
tatarco: all good, i havn't seen the comments before now, all being handled

apps/api/src/app/environments-v1/usecases/output-renderers/hydrate-email-schema.usecase.ts
tatarco: I'm trying to create clear maintainable logic. there are two distinct behaviors here, we hydrate (from pointers to data) and than perform logic on data, those are hard to follow algorithms so I don't want to create a job security algorithem but one which is easy to learn, I'll add a comment explaining it

apps/api/src/app/workflows-v2/usecases/generate-preview/construct-payload-from-placeholders-with-defaults.usecase.ts
```diff
@@ -0,0 +1,157 @@
+/* eslint-disable no-param-reassign */
```
djabarovgeorge: why do we need this?

tatarco: there is a lint not allowing reassignment, but it's used in this logic

SokratisVidros (approved):

PR #6774, opened 2024-10-25
fix(dashboard): Create workflow drawer fixes

desiprisg:

Related enterprise package changes [PR](https://github.com/novuhq/packages-enterprise/pull/236)

### Screenshots

<details>

</details>


SokratisVidros (approved):

apps/dashboard/src/utils/slugify.ts
```diff
@@ -0,0 +1,9 @@
+export function slugify(text: string) {
```
SokratisVidros: This can co into shared, to be reused between BE and FE

rifont: @djabarovgeorge , could you please coordinate with @desiprisg to reuse the `slugify` utility you've added in a separate PR?

SokratisVidros: ☝️

djabarovgeorge: sure, if we want to use signify function without dependencies so we can share it with the client we need to move this one to Shared package and delete [slugifyName](https://github.com/novuhq/novu/blob/add-env-switch-by-url/libs/application-generic/src/utils/slugify-name.ts#L3) from app-gen so everyone will use this one.

djabarovgeorge: At the same time, we need to make sure we remove "slugify" from the dependencies across the monorepo so no one will use it.

LetItRock: I'll update the PR with this suggestion

PR #6775, opened 2024-10-25
chore(root): Upgrade NX to latest packages

SokratisVidros:

Upgrades, upgrades, upgrades!


libs/automation/nx.json
```diff
@@ -1,15 +1,5 @@
 {
   "$schema": "./node_modules/nx/schemas/nx-schema.json",
-  "tasksRunnerOptions": {
-    "default": {
-      "runner": "@nrwl/nx-cloud",
-      "options": {
-        "cacheableOperations": ["build", "test", "lint", "package", "prepare"],
-        "canTrackAnalytics": false,
-        "accessToken": "ZTQ3Yzc3MjEtM2M2ZS00MzRlLWI0OWItZjhmOWVhNWQ1MTM0fHJlYWQ="
```
LetItRock: how we will reuse the Nx cloud cache?

SokratisVidros: According to the latest NX docs, this setting is not necessary anymore. Let me double-check that.

PR #6780, opened 2024-10-27
feat(api): treat workflow name as editable, non-unique values

djabarovgeorge:

- The workflow name must be unique within each environment.
  - If a duplicate name is chosen, an error should be raised.
  - A workflow name can be provided when creating a workflow.

- A workflow name can also be provided during an update. However, if the provided ID matches an existing name, an error should be raised.
  - The workflow ID must be unique within each environment (as it currently is).
  - If a duplicate workflow ID is chosen, an error should be raised.

The workflow ID can be provided during creation but is read-only on the client side (it cannot be updated).
### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -194,7 +276,7 @@ export class UpsertWorkflowUseCase {
       description: workflowDto.description || '',
       tags: workflowDto.tags || [],
       critical: false,
-      triggerIdentifier: slugifyName(workflowDto.name),
+      triggerIdentifier: command.workflowDto.workflowId,
```
djabarovgeorge: bug fix, the client supports passing workflow id but the backend is not processing it.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -168,6 +178,78 @@ export class UpsertWorkflowUseCase {
...
+  private async validateWorkflowUniqueness(
+    existingWorkflow: NotificationTemplateEntity | null,
+    command: UpsertWorkflowCommand
+  ): Promise<void> {
+    const { name } = command.workflowDto;
+    const { environmentId, organizationId } = command.user;
+
+    await this.validateWorkflowIdUniqueness(existingWorkflow, command, name);
+
+    await this.validateWorkflowNameUniqueness(name, environmentId, organizationId, existingWorkflow, command);
```
SokratisVidros: George, according to this https://linear.app/novu/issue/NV-4521/treat-workflowid-and-stepid-as-read-only-values we need to treat only workflow Id and step ID as unique. 

Why did we add name uniqueness?

SokratisVidros (changes requested):
@djabarovgeorge Are there workflows and steps from V1 or current V2 in the database that might violate this new uniqueness enforcement of workflowId and stepId? I'd like us to avoid block updates on these workflows due to a failing uniqueness requirement.

If this is the case we should:

1. Do a data migration if we can
2. Adjust the code to enforce uniqueness as these workflows are getting updated.

If the DB is aligned with uniqueness requirements, no action is required.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: Good point! I think we should clarify this in the ticket. In our [discussion](https://novu.slack.com/archives/C07HVLG7G2U/p1729510091455469?thread_ts=1729509114.021689&cid=C07HVLG7G2U), we mentioned that the name should be unique as well to improve the user experience in the future. We could help the user by automatically adding a suffix to the name in such cases. For example, if "Welcome Notification" is created twice, the second one would be named "Welcome Notification (1)." or something similiar

djabarovgeorge:
> @djabarovgeorge Are there workflows and steps from V1 or current V2 in the database that might violate this new uniqueness enforcement of workflowId and stepId? I'd like us to avoid block updates on these workflows due to a failing uniqueness requirement.
> 
> If this is the case we should:
> 
> 1. Do a data migration if we can
> 2. Adjust the code to enforce uniqueness as these workflows are getting updated.
> 
> If the DB is aligned with uniqueness requirements, no action is required.

This PR is only about workflow, please note that this is only for requests for /v2/workflows on API level Create and Update requests, as far as i know, v1 users won't be part of this dashboard at the moment. once we migrate them we will need to create migration anyways.

djabarovgeorge:
converting to draft for now, first, we need to decide on the design of the DX Guide on this matter.

libs/application-generic/src/utils/generate-id.ts
```diff
@@ -0,0 +1,8 @@
+import { customAlphabet } from 'nanoid';
+
+export const ALPHABET = '0123456789abcdefghijklmnopqrstuvwxyz';
```
djabarovgeorge, line 3: The alphabet is up for debate. I created this one as this is what I understood the requirements to be. The probability of hitting the same workflow ID is 1 to 1,679,616. I created a graceful way to retry twice to create a unique ID. If needed, we can increase the number of retries.

packages/shared/src/dto/workflows/update-workflow-dto.ts
```diff
@@ -1,8 +1,6 @@
 import { PreferencesRequestDto, StepCreateDto, StepUpdateDto, WorkflowCommonsFields } from './workflow-commons-fields';
 
 export type UpdateWorkflowDto = WorkflowCommonsFields & {
-  updatedAt: string;
```
djabarovgeorge, line 4: anyone have an idea why we have it here?

packages/shared/src/dto/workflows/create-workflow-dto.ts
```diff
@@ -1,7 +1,10 @@
+import { IsDefined, IsNotEmpty, IsString } from 'class-validator';
 import { PreferencesRequestDto, StepCreateDto, WorkflowCommonsFields } from './workflow-commons-fields';
 import { WorkflowCreationSourceEnum } from '../../types';
 
 export type CreateWorkflowDto = WorkflowCommonsFields & {
+  workflowId: string;
```
djabarovgeorge, line 6: we allow to pass workflow id only on create.

SokratisVidros (approved):

Issue #6777, opened 2024-10-26
🐛 Bug Report: @novu/frameworks throws no reasonable error when having invalid configuration


paulwer:
### 📜 Description

We had an error, which was releated, to an invalid configuration of our env variables: NOVU_API_URL and NOVU_SECRET_KEY.
We had used the @novu/node lib before and therefore had env in place which met the naming from this lib. (NOVU_BACKEND_URL/NOVU_API_KEY)

This resulted in a misconfiguration (our fault, we could have been reading better :) ) with a strange/wrong debugging experience.
We had to take a deep dive why the following error was present and we were not able to pin down the exact origin for a few hours.

### 👟 Reproduction steps

1. Deploy self-hosted version
2. Deploy bridge server without proper env configuration

### 👍 Expected behavior

The system tried to reach the cloud url of novu instead of the self-hosted on, which is correct, but the error should represent, that the endpoint could not be reached with this configuration. like a 401 error or similar. The errors in the console of @novu/framework should also be more detailed.

Other suggestions: 
- we would also suggest to use a similar naming across libs for the same thing :)
- manualy configurable keys, without need to just use the env vars.

### 👎 Actual Behavior with Screenshots
The error present at the ui was:
Workflow with id: order-recieved has an invalid state. Step with id: in-app has invalid result. Please provide the correct step result.
![Unbekannt](https://github.com/user-attachments/assets/d7656893-907d-4bb1-aa36-0b831c4c3e2c)
The console just shows, that the step could not be hydrated.
![image](https://github.com/user-attachments/assets/dbafd520-ca63-4bb0-a771-776fd9bf9e2b)
The ui shows this weird error. We think this is due to the fact, that the step before could not be handled because of the misconfiguration.
![Unbekannt](https://github.com/user-attachments/assets/eb8a2980-8850-4db7-bd2a-beab874d6637)

### Novu version

Self Hosted

### npm version

k.A.

### node version

20

### 📃 Provide any additional context for the Bug.

selfhosted@2.0.0
@novu/framework@2.3.0 also tested on @novu/framework@2.0.0

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

None

### Our workflow
```
workflow(
        'order-recieved',
[...]

PR #6757, opened 2024-10-23
fix(root): add IS_WORKFLOW_PREFERENCES_ENABLED variable

jainpawan21:
### What changed? Why was the change needed?

- Add `IS_WORKFLOW_PREFERENCES_ENABLED` variable in worker image to resolve the feature flag error. That error will be fixed by [this PR](https://github.com/novuhq/novu/pull/6636)

### Screenshots

<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #6784, opened 2024-10-28
fix(worker): Update docker-compose.yml - #6777

paulwer:
According to https://github.com/novuhq/novu/issues/6777 the missing env resulted in an unwanted behavior and an strange error message.
This will fix this issue.

https://github.com/novuhq/novu/issues/6777


paulwer:
@SokratisVidros @merrcury the error message seen in https://github.com/novuhq/novu/issues/6777 should be looked into seperatly.

merrcury:
We have a duplicate one open [here](https://github.com/novuhq/novu/pull/6757). We should be be close this with discussion there. Thanks

PR #6785, opened 2024-10-28
chore(api, web, application-generic, shared, dal, framework): Remove deprecated `inputs`, `data` and `options` Framework properties

rifont:

* `@novu/framework`
  * Remove deprecated `inputs` property from Framework Workflow & Step options
  * Remove deprecated `data` property from Framework Workflow options
  * Remove deprecated `options` from Framework workflow discovery output
  * Remove backwards compatible LiquidJS template variable usage without `payload` namespace
* Remove usage of the deprecated Framework properties across Platform (API, Worker, Web) 

### Screenshots

<details>

</details>


apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
```diff
@@ -626,62 +626,6 @@ contexts.forEach((context: Context) => {
       expect(sentMessage[1].subject).to.include('prefix Hello default_name');
       expect(sentMessage[0].subject).to.include('prefix Hello payload_name');
     });
-    it(`should trigger the bridge workflow with control default and payload data [${context.name}] - with backwards compatability for payload variable`, async () => {
```
rifont, line 629: Backwards compatibility is now dropped for usage of `payload` variables in LiquidJS without using the `payload` namespace.

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -215,24 +215,17 @@ export class Sync {
...
-          /** @deprecated */
-          (workflow.options?.payloadSchema as Record<string, unknown>),
+        payloadSchema: workflow.payload?.schema,
         active: isWorkflowActive,
         description: this.getWorkflowDescription(workflow),
-        data: this.castToAnyNotSupportedParam(workflow).options?.data,
+        data: this.castToAnyNotSupportedParam(workflow)?.data,
         tags: this.getWorkflowTags(workflow),
-        critical: this.castToAnyNotSupportedParam(workflow.options)?.critical ?? false,
-        preferenceSettings: this.castToAnyNotSupportedParam(workflow.options)?.preferenceSettings,
```
djabarovgeorge, line 235: i think the WorkflowDefine still has `option` param, any reason for this change?

djabarovgeorge (commented):
Less is more 🤩

Left one comment about what i was not sure about.

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
rifont: `DiscoverWorkflowOutput.options` was also removed [here](https://github.com/novuhq/novu/pull/6785/files#diff-31d89519a2321267dc7302f2f855767fdf775d3a59bca0ae7e8834dffd5f8fb9L56-L57), it was deprecated for some time alongside the other deprecated properties.

djabarovgeorge: ok but DiscoverWorkflowOutput and WorkflowDefine are "different" objects, if what you say is right we need to:
1. remove options from WorkflowDefine.
2. remove all the mapping in createWorkflow and updateWorkflow.

SokratisVidros (approved):
🧹

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
rifont: Good catch! I have removed those redundant properties from the `SyncCommand`

PR #6792, opened 2024-10-29
feat(root): move docker images from github packages to aws ecr

merrcury:

This PR linked to below issues
- Fixes INF-444
- Fixes INF-446

### Screenshots
Please refer to AWS console or cli on dev region

PR #6754 slims these images down. So lets merge that first.


.github/workflows/cloud-deploy/cloud-deploy-webhook.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       health_check: true
```
SokratisVidros: You can safely delete all cloud-deploy ymls. They are not needed anymore.

SokratisVidros (approved):

.github/actions/docker/build-api/action.yml
```diff
@@ -105,45 +80,43 @@ runs:
       id: build-image
       shell: bash
       env:
-        REGISTRY_OWNER: novuhq
-        DOCKER_NAME: ${{ inputs.docker_name }}
+        REGISTRY: ${{ steps.login-ecr.outputs.registry }}
```
SokratisVidros, line 83: Is this action reused during OS builds?

.github/workflows/cloud-deploy/cloud-deploy-webhook.yml
merrcury: Sure 👍🏼

.github/actions/docker/build-api/action.yml
merrcury: No, We have separate workflow built on top of matrix. Its entirely different from this 😄.

SokratisVidros: Do we maybe need to rename this shared YML part to make it clearer it's for the Novu Cloud?

merrcury: Sadly its only for dev 😢

PR #6794, opened 2024-10-29
feat(providers): Add onesignal push provider new api

suplere:
This provider is by new Onesignal recommendation - new user model API. Here is the doc from [Onesignal web](https://documentation.onesignal.com/docs/users).

This provider not replace existing player based Onesignal provider, but add new which use last Onesignal user model api. 
### Screenshots

<details>
User si subscribed and set External ID from application database (at most cases). When is External Id set on Onesignal providers - via this methods:
`
include_aliases: {
    external_id: options.target,
}
`

Here is link for example GitHub repository - you can see how you can manage OneSignal external UserID, add this to NOVU subscribers, and send some test Push message.

[novu-onesignal-nextjs](https://github.com/suplere/novu-onesignal-nextjs)

This provider not replace existing player based Onesignal provider, but add new which use last Onesignal user model api. 

</details>


suplere:
Could you someone have a look on this new implementation Onesignal API?

L-U-C-K-Y:
@scopsy any chance you could take a look at this? Would also fix our problem that prevents us from going live after our big refactor. 

Thanks!

L-U-C-K-Y:
@suplere why did you close the MR?

scopsy:
@L-U-C-K-Y there is another MR open by @suplere

PR #6804, opened 2024-10-30
feat(dashboard): workflow promotion

ChmaraX:
- added hook to encapsulate workflow promotion logic and side-effects
- promote workflow button with multiple tooltips depending on the workflow properties
- confirmation modal in case of overwriting of already existing workflow
- custom toasts
- modified API endpoint to optionally accept `environmentId` in order to be able to retrieve workflows from different then user session environment
- added `workflowId` to `GET /workflows` response

### Screenshots

Arbitrary 1s delay was added to the API request to showcase the loading toast.

Promote new:

https://github.com/user-attachments/assets/2a063d28-10b2-46de-895c-21ecb8bf8f80

Promote existing:

https://github.com/user-attachments/assets/ed080c6c-8428-4974-a571-c0bfef640a87

Promote tooltips:

https://github.com/user-attachments/assets/42f57d79-8575-4dd1-a7d9-aa2d17b26f6f

Promote error:

https://github.com/user-attachments/assets/54195035-9435-4057-9b26-29d9de9da67f


apps/api/src/app/workflows-v2/workflow.controller.ts
```diff
@@ -114,9 +114,18 @@ export class WorkflowController {
   @UseGuards(UserAuthGuard)
   async getWorkflow(
     @UserSession(ParseSlugEnvironmentIdPipe) user: UserSessionData,
-    @Param('workflowId', ParseSlugIdPipe) workflowId: IdentifierOrInternalId
+    @Param('workflowId', ParseSlugIdPipe) workflowId: IdentifierOrInternalId,
+    @Query('environmentId') environmentId?: string
```
ChmaraX, line 121: I added this as an optional param to be able to retrieve workflow from different than current user session environment.

We use this when deciding if we need to show confirmation modal informing that workflow already exist in production - we need to find dev `workflowId` in prod environment.

apps/dashboard/src/api/api.client.ts
```diff
@@ -2,7 +2,7 @@ import { getToken } from '@/utils/auth';
 import { API_HOSTNAME } from '../config';
 import { getEnvironmentId } from '@/utils/environment';
 
-class NovuApiError extends Error {
```
ChmaraX, line 5: I need to react to a 404 error from `GET workflow`

 ```typescript
 try {
      if (await isWorkflowInProduction()) {
        return { needsConfirmation: true };
      }
    } catch (error) {
      if (error instanceof NovuApiError && error.status === 404) {
        return { needsConfirmation: false };
      }
 ```

apps/dashboard/src/components/primitives/sonner.tsx
```diff
@@ -91,8 +123,8 @@ const Toaster = ({ ...props }: ToasterProps) => {
       toastOptions={{
         classNames: {
           toast:
-            'group toast group-[.toaster]:bg-background group-[.toaster]:text-foreground group-[.toaster]:border-border group-[.toaster]:shadow-lg',
-          description: 'group-[.toast]:text-muted-foreground',
+            'group toast group-[.toaster]:bg-background group-[.toaster]:text-foreground group-[.toaster]:border-border group-[.toaster]:shadow-lg text-foreground-950',
+          description: 'group-[.toast]:text-foreground-600',
```
ChmaraX, line 95: before it was defaulting to white text on white background

apps/dashboard/src/components/workflow-list.tsx
```diff
@@ -156,91 +132,9 @@ export const WorkflowList = () => {
...
-                          </DropdownMenuGroup>
-                        </DropdownMenuContent>
-                      </DropdownMenu>
-                    </TableCell>
-                  </TableRow>
-                );
-              })}
+              {workflowsQuery.data.workflows.map((workflow) => (
+                <WorkflowRow workflow={workflow} />
+              ))}
```
ChmaraX, line 136: Just moved row logic to standalone file/component to separate concerns and make the `workflow-list` more readable + to be able to use hook inside row component

apps/dashboard/src/routes/dashboard.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { useMonitoring } from '@/hooks/use-monitoring';
+import { Toaster } from '@/components/primitives/sonner';
 
 export const DashboardRoute = () => {
   useMonitoring();
 
   return (
     <ProtectedRoute>
       <Outlet />
+      <Toaster />
```
ChmaraX, line 12: I guess it should be one Toaster placeholder on the entire app.

apps/dashboard/src/components/primitives/sonner.tsx
```diff
@@ -81,6 +82,37 @@ const ToastClose = ({ className, ...props }: React.HTMLAttributes<HTMLButtonElem
   );
 };
 
+const PromoteToast = ({
```
LetItRock: please do not create the specific promote toast, all this code should be in the place where you want to show it, meaning in the `const onPromoteSuccess = () => {` functions

apps/dashboard/src/components/primitives/sonner-helpers.tsx
```diff
@@ -17,3 +17,20 @@ export const showToast = ({
     ...options,
   });
 };
+
+export const promoteToast = ({
```
LetItRock: please use the `showToast` function instead of creating a new function; check the examples throughout the code or contact me for a help ;)

apps/dashboard/src/components/primitives/sonner.tsx
```diff
@@ -81,6 +82,37 @@ const ToastClose = ({ className, ...props }: React.HTMLAttributes<HTMLButtonElem
...
+}: {
+  t: number | string;
+  title: string;
+  description: string;
+  action: { label: string; onClick: () => void };
+}) => {
+  return (
+    <Toast variant="lg" className="gap-3">
+      <ToastIcon variant="default" />
+      <div className="flex flex-[1_0_0] flex-col items-start gap-2.5">
```
LetItRock:
```suggestion
      <div className="flex flex-1 flex-col items-start gap-2.5">
```

apps/dashboard/src/components/primitives/sonner-helpers.tsx
LetItRock: btw please pull the latest changes ;)

apps/dashboard/src/components/workflow-row.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+  const handlePromote = async () => {
+    const { needsConfirmation } = await safePromote();
+
+    if (needsConfirmation) {
+      setShowPromoteModal(true);
+    } else {
+      await promote();
+    }
+  };
```
LetItRock: All this is only relevant to the `PromoteWorkflowMenuItem` and could be encapsulated there, including the modal. Because there are cases when the modal is not needed at all, like the exact same cases when the menu item is `null`.

apps/dashboard/src/components/workflow-row.tsx
```diff
@@ -0,0 +1,239 @@
...
+    await promoteWorkflow();
+  }
+
+  return (
+    <Dialog modal open={open}>
+      <DialogPortal>
+        <DialogOverlay />
+        <DialogContent className="sm:max-w-[440px]">
+          <div className="flex items-start gap-4 self-stretch">
+            <div className="flex items-center justify-center gap-2 rounded-[10px] bg-[#FF84471A] p-2">
```
LetItRock: please do not use the hardcoded values for colors, `bg-warning/10` should work in this case

apps/dashboard/src/components/workflow-row.tsx
```diff
@@ -0,0 +1,239 @@
...
+              <RiAlertFill className="text-warning size-6" />
+            </div>
+            <div className="flex flex-[1_0_0] flex-col items-start gap-1">
```
LetItRock:
```suggestion
            <div className="flex flex-1 flex-col items-start gap-1">
```

apps/dashboard/src/components/workflow-row.tsx
```diff
@@ -0,0 +1,239 @@
...
+              <DialogTitle className="text-md font-medium">Promote workflow to Production</DialogTitle>
+              <DialogDescription className="text-foreground-600">
+                Workflow already exists in Production. Proceeding will overwrite the existing workflow.
+              </DialogDescription>
+            </div>
+          </div>
+          <DialogFooter className="[&~button]:hidden">
```
LetItRock: is this `[&~button]:hidden` needed?

apps/dashboard/src/hooks/use-promote-workflow.ts
```diff
@@ -0,0 +1,114 @@
...
+        description: error instanceof Error ? error.message : 'There was an error promoting the workflow.',
+      });
+
+      return { needsConfirmation: false };
+    }
+
+    await promoteWorkflowMutation();
+
+    return { needsConfirmation: false };
+  };
```
LetItRock: I would suggest you to split this hook to two: `useIsWorkflowInProduction` and `usePromoteWorkflow`.

In the first one you can use `useMutation` and pass `onSuccess` and `onError` callbacks where you can handle all the logic to show/not the confirmation modal, and call the `promoteWorkflowMutation`.

LetItRock (commented):
hey @ChmaraX 👋 I left you a few suggestions, but overall you did a very good job 🤝

apps/dashboard/src/components/workflow-row.tsx
ChmaraX: I moved everything including the toast and modal to the `use-promote-workflow` hook and changed it to `.tsx`.

Although `usePromoteWorkflow` still needs to be used in `WorkflowRow` and not `PromoteWorkflowMenuItem` because the item is part of the dropdown menu and the menu including item disappears after clicking on item, so it would the state + modal etc.

apps/dashboard/src/hooks/use-promote-workflow.ts
ChmaraX: I think splitting this to 2 hooks would cause more confusion, since these 2 hooks will always be used together. 

If `useIsWorkflowInProduction` would only check if workflow is in production then I thinks its too small usecase for a hook, since its just 1 API call and as of now it won't be reused. If it would also include the display of confirmation modal, then it needs to be always used directly with `usePromoteWorkflow` only - its a part of promotion logic not a standalone hook.

I think the simplest option is what we have now - just one not that complicated (< 200 lines) hook, which encapsulates everything the user needs to promote workflow in any place/context

apps/dashboard/src/components/workflow-row.tsx
LetItRock: You can move that hook to be used only inside the `PromoteWorkflowMenuItem` with the modal included. As I mentioned before, you don't always need to render the modal.

apps/dashboard/src/hooks/use-promote-workflow.ts
LetItRock: I'm fine with that, but please just use the `useMutation` for the `isWorkflowInProduction` check, as we don't use the API directly anywhere but always with the `react-query` hooks ;)

SokratisVidros (approved):

apps/dashboard/src/hooks/use-sync-workflow.tsx
```diff
@@ -0,0 +1,213 @@
...
+      </Dialog>
+    );
+  };
+
+  return {
+    safeSync,
+    sync: syncWorkflowMutation,
+    isSyncable,
+    isLoading,
+    tooltipContent: getTooltipContent(),
```
SokratisVidros, line 108: This hook is overloaded with headless and UI logic. TooltipContent and ConfirmationModal need to be transformed to components and be removed from the hook.

SokratisVidros (approved):

apps/dashboard/src/components/workflow-row.tsx
ChmaraX: Please check the last comment - hook inside `PromoteWorkflowMenuItem` would lose state because when you chose/click some item from the menu, the menu automatically closes incl. MenuItems, therefore the state incl. modal needs to be at the row level not menu - menu closes on click. We can sync on that and I can showcase it directly.

apps/dashboard/src/hooks/use-sync-workflow.tsx
ChmaraX: I can understand moving `ConfirmationModal` to separate component although I would argue that the confirmation modal is closely coupled just with the promotion logic and will be used only with this hook and nowhere else, so it should be either in one file or in one folder e.g. `/hooks/sync/*`

But why `TooltipContent`? Its not component its just a message, did you mean Toast?

ChmaraX: Moved the UI logic to separate components and simplified the hook by extracting environment logic to `useEnvironment` hook

PR #6807, opened 2024-10-30
feat(api): treat step name as editable, non-unique values

djabarovgeorge:
1. Step Ids
* Autogenerated based on step name during step creation
* If the same step_id is generated based on workflow a suffix of 4 random characters is added top the step_id from the * server upon workflow creationStep id cannot be edited later.
* Step ids cannot be edited later.

2. Step Name

* Editable
* Not unique
* Used in the step path slug for in URLs

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -73,16 +73,6 @@ describe('Workflow Controller E2E API Testing', () => {
...
-
-    it('should throw error when creating workflow with duplicate step ids', async () => {
-      const nameSuffix = `Test Workflow${new Date().toString()}`;
-      const createWorkflowDto: CreateWorkflowDto = buildCreateWorkflowDto(nameSuffix, {
-        steps: [buildEmailStep(), buildEmailStep(), buildInAppStep(), buildInAppStep()],
-      });
-      const res = await session.testAgent.post(`${v2Prefix}/workflows`).send(createWorkflowDto);
-      expect(res.status).to.be.equal(400);
-      expect(res.body.message).to.be.equal('Duplicate stepIds are not allowed: email-test-step, in-app-test-step');
-    });
```
djabarovgeorge, line 85: Now we create the IDs gracefully using the step name.

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -319,8 +310,7 @@ describe('Workflow Controller E2E API Testing', () => {
       // Verify updated steps
       expect(prodWorkflowUpdated.steps).to.have.lengthOf(2);
       expect(prodWorkflowUpdated.steps[0].name).to.equal('Updated Email Step');
-      // TODO: verify that the stepId (or some) is the same across env for the same prod/dev step
-      expect(prodWorkflowUpdated.steps[0]._id).to.not.equal(prodWorkflowCreated.steps[0]._id);
+      expect(prodWorkflowUpdated.steps[0].stepId).to.equal(prodWorkflowCreated.steps[0].stepId);
```
djabarovgeorge, line 312: @ChmaraX, I was not sure why this test started to fail on this PR. Should the stepId and _id be the same in both cases?

SokratisVidros (approved):

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
djabarovgeorge: Do you mean to add _id is equal? if this is the case, i agree i forgot to add it. will do now

djabarovgeorge: done

PR #6809, opened 2024-10-30
feat(dashboard): in-app editor form fields based on codemirror

LetItRock:
In-App Editor - updated the form fields to be based on the "editor" primitive, which is the CodeMirror component.

**TODO:** 
- styling
- autosuggestions

### Screenshots

Screenshot 2024-10-30 at 23 56 17


apps/dashboard/package.json
```diff
@@ -43,10 +43,14 @@
     "@segment/analytics-next": "^1.73.0",
     "@sentry/react": "^8.35.0",
     "@tanstack/react-query": "^5.59.6",
+    "@uiw/codemirror-theme-white": "^4.23.6",
```
LetItRock, line 46: use the CodeMirror white theme as default; we will adjust it to our needs

apps/dashboard/package.json
```diff
@@ -43,10 +43,14 @@
...
+    "@uiw/codemirror-themes": "^4.23.6",
+    "@uiw/react-codemirror": "^4.23.6",
     "@xyflow/react": "^12.3.2",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.1.1",
     "cmdk": "1.0.0",
+    "codemirror-lang-liquid": "^1.0.0",
```
LetItRock, line 53: liquid syntax and autocomplete

apps/dashboard/src/components/primitives/editor.tsx
```diff
@@ -0,0 +1,64 @@
...
+  variants: {
+    size: {
+      default: 'text-xs [&_.cm-editor]:py-1',
+      md: 'text-sm [&_.cm-editor]:py-2',
+    },
+  },
+  defaultVariants: {
+    size: 'default',
+  },
+});
```
LetItRock, line 17: these styles adjust the editor to look good when placed inside the `InputField` component; basically to have the correct padding from the top

apps/dashboard/src/components/primitives/editor.tsx
```diff
@@ -0,0 +1,64 @@
...
+    value,
+    onChange,
+    theme,
+  });
+
+  useEffect(() => {
+    if (editor.current) {
+      setContainer(editor.current);
+    }
+  }, [setContainer]);
```
LetItRock, line 61: this is the recommended way from the docs

apps/dashboard/src/components/primitives/editor.tsx
```diff
@@ -0,0 +1,64 @@
...
+  placeholder?: string;
+  className?: string;
+  height?: string;
+  onChange: (val: string) => void;
+} & VariantProps<typeof editorVariants>;
+
+export const Editor = ({ value, placeholder, className, height, size, onChange }: EditorProps) => {
+  const editor = useRef<HTMLDivElement>(null);
+  const { setContainer } = useCodeMirror({
+    extensions: [LiquidHTML({}), EditorView.lineWrapping],
```
LetItRock, line 39: liquid and line wrapping extensions

apps/dashboard/src/components/primitives/url-input.tsx
```diff
@@ -14,23 +13,35 @@ type URLInputProps = Omit<InputProps, 'value' | 'onChange' | 'size'> & {
...
-            onChange={(e) => onChange({ ...value, url: e.target.value })}
-            {...rest}
-          />
+          {asEditor ? (
+            <Editor
+              size={size}
+              placeholder={placeholder}
+              value={value.url}
+              onChange={(val) => onChange({ ...value, url: val })}
+              height={size === 'md' ? '38px' : '30px'}
```
LetItRock, line 32: I'm not sure how to do this better without hardcoding these values, basically, our inputs heights are `md 40px` and `default 32px`. These values exclude border. 
And the height should be set if we want to have "single line" input, otherwise the editor will be growing with more content.

apps/dashboard/src/components/workflow-editor/action-picker.tsx
```diff
@@ -178,7 +179,7 @@ const ConfigureActionPopover = (
       }}
     >
       <PopoverTrigger {...rest} />
-      <PopoverContent>
+      <PopoverContent className="max-w-72">
```
LetItRock, line 181: the editor fields expand the popover content, so I had to set the max

apps/dashboard/src/components/workflow-editor/steps/in-app-editor.tsx
```diff
@@ -1,28 +1,32 @@
...
+import { Editor } from '@/components/primitives/editor';
 
 const tabsContentClassName = 'h-full w-full px-3 py-3.5';
 
 export const InAppEditor = () => {
   const navigate = useNavigate();
   const { formState } = useFormContext<z.infer<typeof workflowSchema>>();
 
+  const [subject, setSubject] = useState('');
+  const [body, setBody] = useState('');
```
LetItRock, line 28: just temporarily

apps/dashboard/src/components/primitives/url-input.tsx
LetItRock: Maybe I'll move to constants or something.

SokratisVidros (approved):
Code mirror!!! Yay!!!

apps/dashboard/src/components/primitives/url-input.tsx
SokratisVidros: Are we using code mirror in online inputs to reuse the autocomplete functionality for variables? It's a smart approach to keep it dry, although it feels like an overkill for simple text or email inputs. I wonder if it's more future proof to have the autocomplete CodeMirror offers for code text areas and then implement it again with a popover for regular inputs.

SokratisVidros (approved):
I approve this one and we will chat about using CodeMirror editor across for all form fields separately.

apps/dashboard/src/components/primitives/url-input.tsx
LetItRock: Yes, we use it in all form fields, subject, redirect etc. 
The "regular" single-line fields are limited in the number of characters that you can type and also the way it works currently with CodeMirror is not optimal as you can move the text to the next line and in this case the previous line won't be visible. We need to think about the pattern here, the potential solution might be to have "auto" height on the fields, but then you can't build all types of form UI with this approach as you can easily break the UI.

PR #6810, opened 2024-10-30
feat(api): move step-schema to step 

tatarco:
move logic regarding step schema to step, remove all the transient logic as it doesn't exist and make the endpooint return all the metadata required. for a given workflow.
Add UI schema concept and a clear definition for InApp main UI 

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/workflow.controller.ts
```diff
@@ -161,6 +164,23 @@ export class WorkflowController {
     );
   }
 
+  @Get('/:workflowId/step/:stepUuid/metadata')
```
djabarovgeorge, line 167: Small alignment with the naming.

```suggestion
  @Get('/:workflowId/steps/:stepId/metadata')
```

djabarovgeorge: Small alignment with the naming.

```suggestion
  @Get('/:workflowId/steps/:stepId/metadata')
```

djabarovgeorge: question: i wonder if we can do a clean route so in the future we will use the same endpoint to get all the step data. wdyt?


```suggestion
  @Get('/:workflowId/step/:stepUuid')
```

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    });
+
+    if (!workflow) {
+      throw new BadRequestException({
+        message: 'No workflow found',
+        workflowId: command.workflowId,
+      });
+    }
+
+    const currentStep = workflow.steps.find((stepItem) => stepItem._id === command.stepId);
```
djabarovgeorge, line 71: On the first step creation we could have an empty array, right?

```suggestion
    const currentStep = workflow.steps?.find((stepItem) => stepItem._id === command.stepId);
```

apps/api/src/app/workflows-v2/workflow.controller.ts
djabarovgeorge: we do not support fetching by type now? the client won't be missing the data to create a new step?

apps/api/src/app/workflows-v2/workflow.controller.ts
```diff
@@ -161,6 +164,23 @@ export class WorkflowController {
...
+    @UserSession(ParseSlugEnvironmentIdPipe) user: UserSessionData,
+    @Param('workflowId', ParseSlugIdPipe) workflowId: IdentifierOrInternalId,
+    @Param('stepId', ParseSlugIdPipe) stepId: string
+  ): Promise<StepMetadataDto> {
+    return await this.getStepMetadata.execute(
+      GetStepSchemaCommand.create({
+        organizationId: user.organizationId,
+        environmentId: user.environmentId,
+        userId: user._id,
+        workflowId,
```
LetItRock: can you please adjust it to `identifierOrInternalId` ?

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+import { ControlValuesLevelEnum, StepMetadataDto, StepType, StepTypeEnum } from '@novu/shared';
+import { JSONSchema } from 'json-schema-to-ts';
+import { ControlValuesRepository, NotificationStepEntity, NotificationTemplateRepository } from '@novu/dal';
+import { GetStepSchemaCommand } from './get-step-schema.command';
+import { mapStepTypeToResult } from '../../shared';
+
+@Injectable()
+export class GetStepMetadataUseCase {
+  constructor(
+    private notificationTemplateRepository: NotificationTemplateRepository,
```
LetItRock: please use `private getWorkflowByIdsUseCase: GetWorkflowByIdsUseCase`

apps/api/src/app/workflows-v2/workflow.controller.ts
tatarco: ### it's not a usecase anymore

LetItRock:
```suggestion
  @Get('/:workflowId/steps/:stepId/metadata')
```

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  private assertStepValidity(currentStep: NotificationStepEntity) {
+    if (!currentStep.template?.type) {
+      throw new BadRequestException('No step type found');
+    }
+
+    if (!currentStep.template?.controls?.schema) {
+      throw new BadRequestException('No controls schema found');
+    }
```
LetItRock: I got this error because the dashboard created steps don't have `template.controls.schema`. 
As I got it right there will be only available `uiSchema`, but I don't see anywhere any logic for that.

Screenshot 2024-10-31 at 10 38 40

tatarco: talked sync

apps/api/src/app/workflows-v2/workflow.controller.ts
tatarco: if it will be in produciton anyhow you won't be able to do it, also the variables will not be in a natural place in a get step endpont, I rather keep it for now and handle it in the future if it will happen

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
djabarovgeorge: what was the issue?

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
```diff
@@ -13,21 +13,21 @@ export class InAppOutputRendererUsecase {
       subject: inApp.subject,
       body: inApp.body,
       avatar: inApp.avatar,
-      primaryAction: inApp.primaryAction
+      primaryAction: inApp.buttonSettings?.primaryAction
```
SokratisVidros: In `@novu/framework`, we already have an inApp output schema. We should be consistent and reuse it. PTAL [here](https://github.com/novuhq/novu/blob/4522dd303246b176c95d378549f3c383e1a6f8d2/packages/framework/src/schemas/steps/channels/in-app.schema.ts#L54).

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
```diff
@@ -41,30 +41,34 @@ export class InAppOutputRendererUsecase {
...
 }
-const RedirectTargetEnumSchema = z.enum(['_self', '_blank', '_parent', '_top', '_unfencedTop']);
+const RedirectTargetEnumSchema = z.enum(['_self', '_blank']); // Example enum, replace with your actual enum
 
 const InAppRenderOutputSchema = z.object({
   subject: z.string().optional(),
   body: z.string(),
   avatar: z.string().optional(),
-  primaryAction: z
+  buttonSettings: z
```
SokratisVidros: Ditto about sticking to the inAppOutputSchema as defined in @novu/framework

apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
```diff
@@ -0,0 +1,71 @@
...
+  additionalProperties: false,
+} as const satisfies JSONSchema;
+export const InAppUiSchema: UiSchema = {
+  group: UiSchemaGroupEnum.IN_APP,
+  properties: {
+    body: {
+      component: UiComponentEnum.IN_APP_BODY,
+    },
+    avatar: {
+      component: UiComponentEnum.IN_APP_AVATER,
```
SokratisVidros:
```suggestion
      component: UiComponentEnum.IN_APP_AVATAR,
```

apps/api/src/app/workflows-v2/usecases/generate-preview/step-id-missing.exception.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export class StepIdMissingException extends InternalServerErrorException {
+  constructor(workflowId: string) {
+    super({ message: `StepId is missing for the workflowId: ${workflowId}`, workflowId });
+  }
+}
+
+export class OriginMissingException extends InternalServerErrorException {
+  constructor(stepId: string) {
+    super({ message: `Origin is missing for the stepId: ${stepId}`, stepId });
```
SokratisVidros, line 11: Do steps have origin? What is this error about?

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+        dataSchema: currentStep.controls?.schema,
+        uiSchema: currentStep.controls?.uiSchema,
+        values: await this.getValues(command, currentStep),
+      },
+      variables: buildVariablesSchema(previousSteps),
+    };
+  }
+
+  private async getValues(command: GetStepSchemaCommand, currentStep: NotificationStepEntity) {
+    const controlValuesEntity = await this.controlValuesRepository.findFirst({
```
SokratisVidros: We have to be stricter with the database and assume that there will only be a single entry in the controls collection.  That is, instead of findFirst we should use findOne.

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+      _organizationId: command.organizationId,
+      _workflowId: command.workflowId,
+      _stepId: currentStep._templateId,
+      level: ControlValuesLevelEnum.STEP_CONTROLS,
+    });
+
+    return controlValuesEntity?.controls || {};
+  }
+
+  private assertStepValidity(currentStep: NotificationStepEntity) {
```
SokratisVidros: Maybe we can assert that we have one set of control values only.

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
+import { BadRequestException, Injectable } from '@nestjs/common';
```
SokratisVidros: Let's rename this to just get-step-schema use case to align with the controller. Metadata is a very broad term and creates confusion. In fact, the returned data are the step definition, not metadata about a step.

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+  private assertStepValidity(currentStep: NotificationStepEntity) {
+    if (!currentStep.template?.type) {
+      throw new BadRequestException('No step type found');
+    }
+
+    if (!currentStep.template?.controls?.schema) {
+      throw new BadRequestException('No controls schema found');
+    }
+
+    if (!isStepType(currentStep.template?.type)) {
```
SokratisVidros:
```suggestion
    if (!isValidStepType(currentStep.template?.type)) {
```

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+      _workflowId: command.workflowId,
+      _stepId: currentStep._templateId,
+      level: ControlValuesLevelEnum.STEP_CONTROLS,
+    });
+
+    return controlValuesEntity?.controls || {};
+  }
+
+  private assertStepValidity(currentStep: NotificationStepEntity) {
+    if (!currentStep.template?.type) {
```
SokratisVidros: Please move this to isValidStepType function. A step without type is invalid so one function should check that.

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    const previousSteps = workflow.steps.slice(
+      0,
+      workflow.steps.findIndex((stepItem) => stepItem._id === command.stepId)
+    );
+
+    return { currentStep, previousSteps };
+  }
+}
+
+const buildSubscriberSchema = () =>
```
SokratisVidros: Won't we need this also in the test payload endpoint? It feels that this belongs to a shared folder inside the API.

apps/api/src/app/workflows-v2/workflow.controller.ts
SokratisVidros: Folks, I am sorry to be the party pooper but please remove the metadata please. This is the Get Step endpoint.

If not mistaken instead of `stepUuid`, we must use `stepIdOrStepUuid` to make this more versatile for the FE.

apps/api/src/app/workflows-v2/shared/schemas/workflow-zod-schemas.ts
```diff
@@ -0,0 +1,63 @@
+import { z } from 'zod';
```
SokratisVidros: If I am not mistaken, these DTOs are requests and responses for workflow and step endpoints. I would stick to the strategy we already have in the repo to define them so that, if I am not mistaken, it's not using Zod.

Although the DX that zod offers is much better, consistency and reusability is key in order for us to move fast. So this work should be part of a separate refactoring project we will do in the future..

SokratisVidros (changes requested):

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
tatarco: we are using that output schema, you are commenting on the renderrer which is using the control schema, 
control schema <> output schema
There is no way to implement the UI schema we talked about where a component == 1 control in that way,
specifically the button dropdown, it's a single UI item responsible for 8 diffrent elements and 2 level one elemnts, it must be nested to have a consistent contract

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
tatarco: Thats a bed naming, the output schema stayed the same, thats referring to the control schema

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
tatarco: no, that's that is purely for defining the variables, it's not reused

apps/api/src/app/workflows-v2/workflow.controller.ts
tatarco: it is not a get step endpoint, it's not returning the step resource, it's returning metadata required to display a step editor.
the variables constructed there does not belog on the step

apps/api/src/app/workflows-v2/shared/schemas/workflow-zod-schemas.ts
tatarco: I'll remove it, it was there to help with the e2e tests, unsuccessfully

apps/api/src/app/workflows-v2/usecases/generate-preview/step-id-missing.exception.ts
tatarco: once we are using the notification entity there are lots of optional fields that are causing compilation issues although they are not expected to be empty, in order to typeguard it i'm throwing 500 exceptions as they are some kind of inconsistant bug

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
SokratisVidros: For posterity, this was resolved offline.

apps/api/src/app/workflows-v2/workflow.controller.ts
SokratisVidros: For posterity, this was resolved offline.

apps/api/src/app/workflows-v2/usecases/get-step-schema/get-step-metadata-use-case.service.ts
tatarco: changing to get step data ok? it's not only schmas, also the controler moved to the workflow controler

apps/api/src/app/workflows-v2/exceptions/step-has-issues.exception.ts
```diff
@@ -0,0 +1,8 @@
+import { InternalServerErrorException } from '@nestjs/common';
+import { NotificationStepEntity } from '@novu/dal';
+
+export class StepHasIssuesException extends InternalServerErrorException {
```
SokratisVidros: A step with issues should raise a 4XX error, not a 5XX as it's an expected error state based on client provided data.

```suggestion
export class StepHasIssuesException extends InternalServerErrorException {
```

apps/api/src/app/workflows-v2/exceptions/step-has-issues.exception.ts
```diff
@@ -0,0 +1,8 @@
...
+  constructor(currentStep: NotificationStepEntity) {
+    super({ message: 'Step has internal issues, missing id or name', step: currentStep });
```
SokratisVidros: This error message is confusing. Is it about a missing id or name or is it about the issues each step can have for its controls?

apps/api/src/app/workflows-v2/exceptions/step-not-found-exception.ts
```diff
@@ -1,7 +1,12 @@
-import { NotFoundException } from '@nestjs/common';
+import { InternalServerErrorException, NotFoundException } from '@nestjs/common';
 
 export class StepNotFoundException extends NotFoundException {
   constructor(stepUuid: string) {
     super({ message: 'Step cannot be found using the UUID Supplied', stepUuid });
   }
 }
+export class StepMissingControlsException extends InternalServerErrorException {
```
SokratisVidros, line 8: Ditto, should be a 4xx.

apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
```diff
@@ -0,0 +1,68 @@
...
+const actionSchema = {
+  type: 'object',
+  properties: {
+    label: { type: 'string' },
+    redirect: redirectSchema,
+  },
+  required: ['label'],
+  additionalProperties: false,
+} as const satisfies JSONSchema;
+export const inAppControlSchema = {
```
SokratisVidros, line 33: Please add new lines between these declarations.

SokratisVidros (approved):

PR #6817, opened 2024-10-31
fix(worker): Add simple Subscriber Global Preference use-case for SendMessage usage

rifont:

* Add simple Subscriber Global Preference use-case for SendMessage usage
  * The reused `GetSubscriberGlobalPreference` recently introduced new logic to filter the channels returned to include only relevant channels for the Subscriber. This logic involved fetching all Workflow-level preferences for the subscriber, which placed significant load on the Worker runtime and database due to a `1:N` lookup against the Preference collection occurring when fetching Global Preferences for the Subscriber, where `N` is the number of Workflows the Subscriber has access to. The use-case now only makes a single lookup against the database.
  * This new use-case will be deleted after we migrate to V2 Only Preferences
* Add instrumentation for all preference use-cases

### Screenshots

<details>

</details>


libs/application-generic/src/usecases/get-subscriber-global-preference-v1/get-subscriber-global-preference-v1.usecase.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const subscriberGlobalChannels =
+      subscriberGlobalPreferenceV1?.channels ?? {};
+    const enabled = subscriberGlobalPreferenceV1?.enabled ?? true;
+
+    return {
+      channels: subscriberGlobalChannels,
+      enabled,
+    };
+  }
```
rifont, line 74: This new use-case only performs a single lookup against the V1 `SubscriberPreferenceRepository` to fetch the Global Preference, rather than iterating over all Workflow-level Preferences.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -50,6 +50,7 @@ export class GetSubscriberPreference {
       },
     );
 
+    // TODO: replace this runtime mapping with a single query to the database
```
rifont, line 55: This was the problematic piece of code that is now avoided in the `SendMessage` use-case.

libs/application-generic/src/usecases/get-subscriber-global-preference-v1/get-subscriber-global-preference-v1.usecase.ts
```diff
@@ -0,0 +1,101 @@
+import { Injectable } from '@nestjs/common';
```
rifont, line 1: Compare this new, simplified use-case to the [previously consumed use-case](https://github.com/novuhq/novu/blob/add-simple-global-preference-use-case/libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts) - that previously consumed use-case that is now only used for fetching Inbox preferences.

libs/application-generic/src/usecases/get-subscriber-global-preference-v1/get-subscriber-global-preference-v1.usecase.ts
```diff
@@ -0,0 +1,101 @@
...
+  SubscriberRepository,
+} from '@novu/dal';
+
+import { IPreferenceChannels, ChannelTypeEnum } from '@novu/shared';
+import { GetSubscriberGlobalPreferenceCommandV1 } from './get-subscriber-global-preference-v1.command';
+import { buildSubscriberKey, CachedEntity } from '../../services/cache';
+import { ApiException } from '../../utils/exceptions';
+import { InstrumentUsecase } from '../../instrumentation';
+
+/** @deprecated - use `GetPreferences` instead */
```
SokratisVidros: Please elaborate on this until we migrate. We should GetPreferences apart from the places that the deprecated usecase is already used, right?

SokratisVidros (approved):

libs/application-generic/src/usecases/get-subscriber-global-preference-v1/get-subscriber-global-preference-v1.usecase.ts
rifont: Yes, exacty.

I added some more clarity [here](https://github.com/novuhq/novu/pull/6817/files#diff-b85899d0434586dd3aeeb1910c51c1742571347dc8ff8f396ea9396670dfdca4R15-R23).

Let me know if that clears up the approach.

PR #6819, opened 2024-10-31
fix(framework): Polish secretKey and apiUrl resolution

SokratisVidros:
1. Centralize secretKey and apiUrl resolution to ensure that provided values are always prioritized over NOVU env variables.
2. Allow users to pass dynamic secretKey and apiUrl parameters in workflow.trigger() function.


packages/framework/src/utils/http.utils.ts
```diff
@@ -1,12 +1,14 @@
 import { checkIsResponseError } from '@novu/shared';
-import { BridgeError, PlatformError } from '../errors';
+import { BridgeError, MissingSecretKeyError, PlatformError } from '../errors';
 
-export const initApiClient = (apiKey: string, baseURL = 'https://api.novu.co') => {
-  const apiUrl = process.env.NOVU_API_URL || baseURL;
+export const initApiClient = (apiKey: string, apiUrl?: string) => {
+  if (!apiKey) {
+    throw new MissingSecretKeyError();
+  }
```
rifont:
```suggestion
export const initApiClient = (secretKey: string, apiUrl?: string) => {
  if (!secretKey) {
    throw new MissingSecretKeyError();
  }
```

packages/framework/src/utils/options.utils.ts
```diff
@@ -0,0 +1,7 @@
+export function resolveApiUrl(providedApiUrl?: string) {
+  return providedApiUrl || process.env.NOVU_API_URL || 'https://api.novu.co';
+}
+
+export function resolveSecretKey(providedSecretKey?: string) {
+  return providedSecretKey || process.env.NOVU_SECRET_KEY || process.env.NOVU_API_KEY || '';
+}
```
rifont: Adding explicit return types to enforce the function boundary
```suggestion
export function resolveApiUrl(providedApiUrl?: string): string {
  return providedApiUrl || process.env.NOVU_API_URL || 'https://api.novu.co';
}

export function resolveSecretKey(providedSecretKey?: string): string {
  return providedSecretKey || process.env.NOVU_SECRET_KEY || process.env.NOVU_API_KEY || '';
}
```

packages/framework/src/utils/http.utils.ts
```diff
@@ -1,12 +1,14 @@
 import { checkIsResponseError } from '@novu/shared';
-import { BridgeError, PlatformError } from '../errors';
+import { BridgeError, MissingSecretKeyError, PlatformError } from '../errors';
 
-export const initApiClient = (apiKey: string, baseURL = 'https://api.novu.co') => {
-  const apiUrl = process.env.NOVU_API_URL || baseURL;
+export const initApiClient = (secretKey: string, apiUrl?: string) => {
```
rifont: It looks like `apiUrl` is always provided as the `resolveApiUrl` function handles it, so I think we can remove the fallback.
```suggestion
export const initApiClient = (secretKey: string, apiUrl: string) => {
```

packages/framework/src/utils/http.utils.ts
```diff
@@ -1,12 +1,14 @@
...
+  if (!secretKey) {
+    throw new MissingSecretKeyError();
+  }
 
   return {
     post: async <T = unknown>(route: string, data: Record<string, unknown>): Promise<T> => {
-      const response = await fetch(`${apiUrl}/v1${route}`, {
+      const response = await fetch(`${apiUrl || 'https://api.novu.co'}/v1${route}`, {
```
rifont: As above
```suggestion
      const response = await fetch(`${apiUrl}/v1${route}`, {
```

rifont (approved):
💯 Approved with a few tidy up comments

packages/framework/src/utils/http.utils.ts
SokratisVidros: If we do that, we will have to throw an error if the `apiUrl` is missing. Then I thought, do we need the extra bytes for that error?

packages/framework/src/utils/options.utils.ts
SokratisVidros: Done

packages/framework/src/utils/http.utils.ts
rifont: The `resolveApiUrl` provides the same `https://api.novu.co` fallback here [though](https://github.com/novuhq/novu/pull/6819/files#diff-5e3656182382c7a6da8b4d2e5c7427ffd51d9a0b53a8093310e125547d37dcbaR2), and `initApiClient` always receives an `apiUrl` resolved from that function.

So, an `apiUrl` would always be present and we don't need to handle the undefined `apiUrl` case.

SokratisVidros: You are 100% right. I will update it.

PR #6824, opened 2024-11-01
fix(api): Preserve control values across syncs

SokratisVidros:

Syncing the same workflow multiple times overrides control values set in the Dashboard. This PR fixes it.

### DIFF
--- a/apps/api/src/app/events/e2e/bridge-sync.e2e.ts
+++ b/apps/api/src/app/events/e2e/bridge-sync.e2e.ts
@@ -1,7 +1,12 @@
 import { UserSession } from '@novu/testing';
 import { expect } from 'chai';
-import { EnvironmentRepository, NotificationTemplateRepository, MessageTemplateRepository } from '@novu/dal';
-import { FeatureFlagsKeysEnum, WorkflowTypeEnum } from '@novu/shared';
+import {
+  EnvironmentRepository,
+  NotificationTemplateRepository,
+  MessageTemplateRepository,
+  ControlValuesRepository,
+} from '@novu/dal';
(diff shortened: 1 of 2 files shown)


djabarovgeorge (approved):
🚀

apps/api/src/app/events/e2e/bridge-sync.e2e.ts
```diff
@@ -553,4 +559,89 @@ describe('Bridge Sync - /bridge/sync (POST)', async () => {
...
+    expect(firstWorkflowResponse.steps[0]._templateId).to.exist;
+
+    await session.testAgent.put(`/v1/bridge/controls/${workflowId}/send-email`).send({
+      variables: { subject: 'Hello World again' },
+    });
+
+    const firstControlValueResponse = await controlValuesRepository.find({
+      _environmentId: session.environment._id,
+      _workflowId: firstWorkflowResponse._id,
+    });
```
rifont, line 607: Should we perhaps access these controls via API rather than directly to the DB to ensure any business logic between the API endpoint and controls query is tested too?

rifont (approved):
💯

apps/api/src/app/events/e2e/bridge-sync.e2e.ts
djabarovgeorge: due to our current use case e2e tests architecture, IMO the main test here is to `set` control values through API (sync), i would create a separate test for `get` logic where we would definitely need to use the API endpoint.

SokratisVidros: I will do both checks. Good point.

PR #6827, opened 2024-11-01
feat(application-generic, api, worker): Add instrumentation suffix builder

rifont:

* Add instrumentation suffix builder
  * Provides a means for using method arguments to suffix the span identifier, which can be useful for generic methods where it can be difficult to understand the purpose of the call
* Some minor refactoring to move the instrumentation code out of the `index.ts` file into a better-named `instrumentation.decorator.ts` file

### Screenshots

_FeatureFlag `get` method now shows the feature flag `key` requested_
image

<details>

</details>


libs/application-generic/src/instrumentation/instrumentation.decorator.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    if (nr) {
+      const isAsync = method.constructor.name === 'AsyncFunction';
+
+      if (!isAsync) {
+        // eslint-disable-next-line no-param-reassign
+        descriptor.value = function (...args) {
+          const customSuffix = buildTransactionIdSuffix
+            ? `:${buildTransactionIdSuffix(...args)}`
```
rifont, line 63: What do we think of the colon `:` delimiter? We could alternately use the same slash `/` delimiter as used on `transactionIdentifierBase`, but I chose a different delimiter to differentiate the suffixed part.

libs/application-generic/src/instrumentation/instrumentation.decorator.ts
```diff
@@ -0,0 +1,104 @@
...
+    return descriptor;
+  };
+}
+
+const buildTransactionId = (
+  transactionIdentifierBase: string,
+  buildSuffix: InstrumentationOptions['buildTransactionIdSuffix'],
+  args: unknown[],
+): string => {
+  const suffix = buildSuffix ? `:${buildSuffix(...args)}` : '';
```
rifont, line 101: What do we think of the colon `:` delimiter? We could alternately use the same slash `/` delimiter as used on `transactionIdentifierBase`, but I chose a different delimiter to differentiate the suffixed part.

SokratisVidros: Sounds good.

SokratisVidros (approved):

PR #6831, opened 2024-11-02
chore(dashboard): InApp /docs redirect fix

twentyone24:
currently clicking on the in-app step takes to docs homepage, instead of inApp step page, this PR fixes it.


SokratisVidros (approved):
😍

PR #6833, opened 2024-11-03
feat(api): persist payload variables

djabarovgeorge:
### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -375,3 +400,19 @@ function isWorkflowUpdateDto(
 const isUniqueStepId = (stepIdToValidate: string, otherStepsIds: string[]) => {
   return !otherStepsIds.some((stepId) => stepId === stepIdToValidate);
 };
+
+function extractPlaceholders(text: string): string[] {
```
djabarovgeorge, line 399: self note: reuse already existing function in the preview use case.

djabarovgeorge:
not relevant anymore as we do not want to store this payload

PR #6840, opened 2024-11-04
feat(framework): add class-validator support

### LINKED ISSUE #6682: 🚀 Feature: @novu/framework Integration with class-validator
### 🔖 Feature description

Hi there,
we are using novu with a nestjs service, and have multiple classes already defined with class-validator for validation and dont use zod right now. Would it be possible to add support for class-validator for schema definition instead of zod?

NestJs and multiple other services are using this lib already for validation and it would help quite alot to just reuse existing classes within the workflow definitions.

We dont know for certain if there are any technical limitations, but if not it would be great if both libraries could be used as input.

### 🔄️ Additional Information

class-validator to json schema: https://www.npmjs.com/package/class-validator-jsonschema

None

paulwer:
Class Validator support for @novu/framework

### Screenhots
_Class validator peer dependency import failure message_
```bash
 ⨯ Error: Tried to use a class-validator schema in @novu/framework without class-validator-jsonschema installed. Please install it by running `npm install class-validator-jsonschema`.
    at se (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:77618)
    at async si.canHandle (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:79403)
    at async sp (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:81104)
    at async Object.discover (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:10531)
    at async sl.addWorkflows (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:501:8314)
    at async (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:3508)
    at async te.do (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:17826)
    at async te.handle (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:22492)
    at async doRender (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/server/base-server.js:1455:42) {
  statusCode: 500,
  code: 'MissingDependencyError',
  data: [Array]
}
```

<details>

</details>

closes #6682


paulwer:
@rifont i marked you to may review this PR, as you have created the original validator implementation.
If there is anything I can provide, please reach out :)

SokratisVidros:
@paulwer Thanks for submitting this PR. It coincides with an internal conversation about allowing contributors to bring their own validations, such as Zod, ClassValidator, etc...

Our dynamic require("x") pattern, along with peer dependencies, served us well so far, but as we moved our package to its ESM/CJS state, we noticed it caused some issues in some bundlers, such as Turbopack beta.

We expect to conclude our strategy by the end of this week, and the decision might also affect this PR. We would love to hear your opinions or suggestions on the matter.

paulwer:
Hi @SokratisVidros, thanks for the feedback. If the descision would be to only allow json-schema, i would like to encourage you to provide examples or mapper functions to be installed additionaly.

If I can support this in any way, please let me know :)

SokratisVidros:
@paulwer White smoke! We are doubling down on ESM, making schema validation asynchronous so that we can use async imports. We expect this work to be merged by tomorrow. Then your PR will need some adjustments before merging.

We would love your feedback, PTAL https://github.com/novuhq/novu/pull/6879

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,139 @@
...
+    T_Unvalidated = FromSchemaUnvalidated<T_Schema>,
+    T_Validated = FromSchema<T_Schema>,
+  >(data: T_Unvalidated, schema: T_Schema): Promise<ValidateResult<T_Validated>> {
+    try {
+      // eslint-disable-next-line global-require
+      require('reflect-metadata') as typeof import('reflect-metadata');
+      // eslint-disable-next-line global-require
+      const { plainToInstance, instanceToPlain } = require('class-transformer') as typeof import('class-transformer');
+      // eslint-disable-next-line global-require
+      const { validate } = require('class-validator') as typeof import('class-validator');
```
rifont: We can convert all of these to use `import` now 🎉

```suggestion
      await import('reflect-metadata');
      const { plainToInstance, instanceToPlain } = await import('class-transformer');
      const { validate } = await import('class-validator');
```

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  transformToJsonSchema(schema: ClassType): JsonSchema {
+    try {
+      // eslint-disable-next-line global-require
+      const { defaultMetadataStorage } = require('class-transformer/cjs/storage');
+      // eslint-disable-next-line global-require
+      const { getMetadataStorage } = require('class-validator') as typeof import('class-validator');
+      // eslint-disable-next-line global-require, prettier/prettier
+      const { targetConstructorToSchema, validationMetadatasToSchemas } =
+        require('class-validator-jsonschema') as typeof import('class-validator-jsonschema');
```
rifont: Same applies here, we can use `import`.

Also, we need to replace the CJS import for class-transformer with an ESM entry point.

packages/framework/src/types/schema.types.ts
```diff
@@ -3,14 +3,17 @@ import zod from 'zod';
 
 export type JsonSchema = JSONSchema;
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ClassType<T = any> = new (...args: any[]) => T;
```
rifont: Could this be converted to `unknown[]` rather than `any[]`?

packages/framework/src/types/step.types.ts
```diff
@@ -64,7 +64,8 @@ export type ActionStep<
   /**
    * The controls for the step.
    */
-  T_Controls extends Record<string, unknown> = FromSchema<T_ControlSchema>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  T_Controls extends Record<string, any> = FromSchema<T_ControlSchema>,
```
rifont: Could you please clarify why these changes were needed?

We are preferring an approach of type strictness where assertions need to be made on the control value types when schemas are not declared. If I'm not mistaken, this change will weaken that type strictness.

rifont (commented):
Awesome contribution @paulwer ! This will provide a strong basis for NestJS usage of the Framework.

I added some comments for your review, we will get this merged in and released ASAP.

packages/framework/src/validators/class-validator.validator.ts
paulwer: does this mean cjs is still supported in the future or is it dropped entirely?

packages/framework/src/types/schema.types.ts
paulwer: my ide was not able to resolve the typings in this case :(
also several other places have no working typings, but i guess this is not a problem introduced with class-validator.
Take look at the test suites, here you see what Else is failing.

packages/framework/src/types/step.types.ts
paulwer: as explained in the comment above. 
I can revert this change if you like, but this has to be looked on seperatly.

paulwer:
@rifont any considderations for me regarding: https://github.com/novuhq/novu/pull/6428?

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,133 @@
...
+        );
+      }
+      throw error;
+    }
+  }
+
+  async transformToJsonSchema(schema: ClassType): Promise<JsonSchema> {
+    try {
+      // eslint-disable-next-line global-require
+      const { defaultMetadataStorage } = require('class-transformer/cjs/storage');
```
paulwer: @rifont 
I am not able to replace this line with the apropiate import.
This will result in a failing test for get json-schema for nested properties. Would you mind to take a look?

rifont: The package internals should use ESM for full compatibility, we still transpile the ESM internals down to both a CJS + ESM bundle. We have no plans to drop CJS support.

There's some discussion [here](https://github.com/typestack/class-transformer/issues/563#issuecomment-803262394) about importing from the `class-transformer` internals via the `class-transformer/esm5/storage`. Could we try that?

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,133 @@
...
+        path: currentPath,
+        message: Object.values(err.constraints || {}).join(', '),
+      };
+    }
+  });
+}
+
+export class ClassValidatorValidator implements Validator<ClassType> {
+  canHandle(schema: Schema): schema is ClassType {
+    return typeof (schema as ClassType) === 'function';
```
rifont: We need to be a bit more specific on the `canHandle` operator to ensure we won't collide with other schemas in the future. Currently, the schema `canHandle` checking order matters because a Zod Schema is also a typeof 'function'.

Could we add an assertion to narrow the check?

Here is a refined guard suggestion
```suggestion
    return typeof schema === 'function' && schema.prototype !== undefined && schema.prototype.constructor === schema;
```

packages/framework/src/validators/class-validator.validator.ts
paulwer: I've added a workaround, to determine which of those should be imported. But idk if this is the way to go.
Thanks for the hint to the discussion btw.
We may should check if they can resolve this shortly, because I think a native import would be much better.

The tests will only succeed currently, when cjs is imported.

rifont: Ok thanks for confirming. Let's go with a CJS import for now and we can address full ESM compatibility here later.

Could you please add a link to the native import discussion with a `TODO` directive to use the native import when it's available?

rifont:
> @rifont any considderations for me regarding: #6428?

No, we can address support for class-validators in that PR later. Thanks for checking!

packages/framework/src/validators/class-validator.validator.ts
paulwer: please take a look at the current implementation, do you see any pitfalls here? should we go with just cjs to begin with?

paulwer:
@rifont there is also a patch for the lib created, which we can try to use if there are any issues while building

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,151 @@
...
+      }
+      throw error;
+    }
+  }
+
+  async transformToJsonSchema(schema: ClassType): Promise<JsonSchema> {
+    try {
+      /*
+       * TODO: replace with direct import, when metadataStorage is exported by default: https://github.com/typestack/class-transformer/issues/563#issuecomment-803262394
+       * @ts-ignore
```
rifont: It's preferable to use `ts-expect-error` directives over `ts-ignore` to become aware of when a downstream change fixes an issue (`ts-expect-error` without an error it's blocking will cause TS compilation to fail).

```suggestion
       * @ts-expect-error - class-transformer doesn't export `defaultMetadataStorage` from the root module
```

packages/framework/src/types/schema.types.ts
rifont: It seems the `FromSchema` inference was causing the troubles. I've separated into a `ClassType` and `InferenceClassProperties` generic and we are golden! Fix coming here soon.

packages/framework/src/types/step.types.ts
rifont: I have reverted this after resolving the type inference issues.

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,141 @@
...
+    );
+  }
+
+  async validate<
+    T_Schema extends ClassType = ClassType,
+    T_Unvalidated = FromSchemaUnvalidated<T_Schema>,
+    T_Validated = FromSchema<T_Schema>,
+  >(data: T_Unvalidated, schema: T_Schema): Promise<ValidateResult<T_Validated>> {
+    try {
+      await import('reflect-metadata');
```
paulwer: @rifont I am currently not on my machine. please try to remove this, as of this is only required in the global scope and would be a unnecessary prod dependency.

You can check the readme of the class validator lib for more informations.

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,141 @@
...
+            classTransformerMetadataStorage: defaultMetadataStorage,
+          }),
+          schemas
+        )
+      );
+    } catch (error) {
+      if ((error as Error)?.message?.includes('Cannot find module')) {
+        // eslint-disable-next-line no-console
+        console.error(
+          'Tried to use a class-validator schema in @novu/framework without `class-validator-jsonschema` installed. ' +
```
paulwer: Should also be changed to represent errors of missing class-validator and class-transformer imports.

Please also check if class-validator could be used as standalone. I am not sure about that.

You should also add a distinct section within the docs later to explain the necessary use of (at)Type(()=> ClassName) in the class when having nesten props. and also (at)ValidateNested()

The transformer will not reconize the class without it.

packages/framework/src/validators/fixures/class-validator.fixtures.ts
```diff
@@ -0,0 +1,30 @@
+import 'reflect-metadata';
+import { IsNumber, IsOptional, IsString, ValidateNested } from 'class-validator';
+import { Type } from 'class-transformer';
+
+export class SimpleStringSchema {
+  @IsString()
+  @IsOptional()
+  name?: string;
```
paulwer: check if this should realy be optional.

rifont: Good catch, it shouldn't be optional. I will make it required.

packages/framework/src/validators/class-validator.validator.ts
rifont: You're absolutely right, `reflect-metadata` should be imported once as a singleton. I will look at removing.

packages/framework/src/validators/class-validator.validator.ts
rifont: I have introduced a [`checkDependencies` util](https://github.com/novuhq/novu/pull/6912/files#diff-1cc9be706dd3a6a268390467632ceb7b69b3585cafa05418dd9d7bbe67ecbfcd) in a schema refactor for Framework. It turns out there's a lot of nuance with detecting dependency availability across different runtime environments, but I'm confident this new utility will capture all import issues robustly.

The new util dynamically detects the missing imports and adjusts the message accordingly.

rifont: Regarding the docs usage for nested values, it's a great callout. We'll be sure to mention it ✅

packages/framework/src/types/schema.types/zod.schema.types.ts
```diff
@@ -1,3 +1,4 @@
+
```
rifont:
```suggestion
```

packages/framework/src/client.test.ts
```diff
@@ -313,6 +313,8 @@ describe('Novu Client', () => {
       if (stepChat === undefined) throw new Error('stepEmail is undefined');
       expect(stepChat.type).toBe('chat');
       expect(stepChat.code).toContain(`body: "Test Body"`);
+      console.log(stepChat.providers);
+      // await new Promise((resolve) => setTimeout(resolve, 1000));
```
rifont:
```suggestion
```

packages/framework/package.json
```diff
@@ -244,6 +264,8 @@
     "sanitize-html": "^2.13.0"
   },
   "nx": {
-    "tags": ["package:public"]
+    "tags": [
+      "package:public"
+    ]
```
rifont:
```suggestion
    "tags": ["package:public"]
```

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,163 @@
...
+      name: 'reflect-metadata',
+      import: import('reflect-metadata'),
+      exports: [],
+    },
+    {
+      name: 'class-validator-jsonschema',
+      import: import('class-validator-jsonschema'),
+      exports: ['validationMetadatasToSchemas', 'targetConstructorToSchema'],
+    },
+  ];
```
rifont, line 94: All dependencies are now explicitly declared for availability checks.

packages/framework/src/validators/class-validator.validator.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  async canHandle(schema: Schema): Promise<boolean> {
+    const canHandle =
+      typeof (schema as ClassValidatorSchema) === 'function' &&
+      (schema as ClassValidatorSchema).prototype !== undefined &&
+      (schema as ClassValidatorSchema).prototype.constructor === schema;
+
+    if (canHandle) {
+      await checkDependencies(this.requiredImports, 'class-validator schema');
```
rifont:
```suggestion
      await checkDependencies(this.requiredImports, 'Class Validator schema');
```

packages/framework/src/validators/class-validator.validator.ts
rifont: In the end, we'll go with CJS! I have tested on both CJS + ESM successfully.

packages/framework/src/validators/class-validator.validator.ts
rifont: I have removed this import. We now have an explicit check for the `reflect-metadata` package via the `checkDependencies` util.

packages/framework/src/validators/validator.test.ts
```diff
@@ -1,16 +1,19 @@
 import { describe, it, expect } from 'vitest';
 import { z } from 'zod';
 import { validateData, transformSchema } from './base.validator';
-import { Schema, ZodSchema, JsonSchema } from '../types/schema.types';
+import type { Schema, ZodSchema, JsonSchema } from '../types/schema.types';
+import { SimpleStringSchema, NestedSchema, SimpleStringAndNumberSchema } from './fixures/class-validator.fixtures';
+import { ClassValidatorSchema } from '../types/schema.types/class.schema.types';
```
rifont:
```suggestion
import type { Schema, ZodSchema, JsonSchema, ClassValidatorSchema } from '../types/schema.types';
import { SimpleStringSchema, NestedSchema, SimpleStringAndNumberSchema } from './fixures/class-validator.fixtures';
```

paulwer:
solid changes :)
I had reviewed it again and have nothing to complain <3

paulwer:
Not true, i have some minor suggestions :) 
1. add tests for enum support of class-validator/-transformer.

2. You may also need to recheck anyOf/allOf etc.
f.ex. https://stackoverflow.com/questions/75782271/how-to-validate-nested-array-of-multiple-types-using-class-validator

3. add tests for validating array of objects should be added as a test (ValidateNested({ each }))

rifont:
Ahh, so evidently Github will automatically close a fork PR when deleting an upstream branch (which in the case of `novuhq/novu`, happens automatically after merging a PR).

@paulwer , do you have the capability to reopen the PR on your end? If not we can create a new PR with all the existing changes.

paulwer:
no worries, i created another pr with the latest changes :)

please check

PR #6848, opened 2024-11-05
feat(api): add plain support service

jainpawan21:

- Add plain support service, use case, command and dto
- Add SupportModal to submit a question
- Add dropdown on click of header nav help icon 
### Screenshots

<details>

</details>


apps/api/src/app/support/usecases/create-thread.command.ts
```diff
@@ -0,0 +1,20 @@
...
+  @IsString()
+  email: string;
+
+  @IsDefined()
+  @IsString()
+  firstName: string;
+
+  @IsDefined()
+  @IsString()
+  lastName: string;
```
scopsy: This might be empty in some cases, suggest doing it optional

apps/api/src/app/support/usecases/create-thread.usecase.ts
```diff
@@ -0,0 +1,29 @@
...
+@Injectable()
+export class CreateSupportThreadUsecase {
+  constructor(private supportService: SupportService) {}
+
+  async execute(command: CreateSupportThreadCommand) {
+    const firstName = capitalize(command.firstName ?? '');
+    const lastName = capitalize(command.lastName ?? '');
+    const plainCustomer = await this.supportService.upsertCustomer({
+      emailAddress: command.email,
+      fullName: `${firstName} ${lastName}`,
```
scopsy, line 15: Let's also pass the externalId here to point to our internaluser id

apps/api/src/app/shared/shared.module.ts
```diff
@@ -116,6 +117,7 @@ const PROVIDERS = [
   ExecuteBridgeRequest,
   getFeatureFlag,
   GetDecryptedSecretKey,
+  supportService,
```
scopsy: Should we maybe initialize this only on the Support module?

apps/api/src/app.module.ts
```diff
@@ -107,6 +108,7 @@ const baseModules: Array<Type | DynamicModule | Promise<DynamicModule> | Forward
   PreferencesModule,
   WorkflowModule,
   EnvironmentsModule,
+  SupportModule,
```
scopsy: Let's only inject this if the ENTERPRISE env flag is on

libs/application-generic/src/services/support.service.ts
```diff
@@ -0,0 +1,84 @@
...
+      this.plainClient = new PlainClient({
+        apiKey: this.plainKey,
+      });
+    } else {
+      Logger.error(
+        'Missing Plain API key: Unable to perform support operations. Please ensure the API key is set in the environment variables.',
+        LOG_CONTEXT,
+      );
+    }
+  }
```
scopsy:
```suggestion
  constructor(private plainKey?: string | null) {
    if (this.plainKey) {
      this.plainClient = new PlainClient({
        apiKey: this.plainKey,
      });
    } else {
      Logger.error(
        'Missing Plain API key: Unable to perform support operations. Please ensure the API key is set in the environment variables.',
        LOG_CONTEXT,
      );
    }
  }
```

A small one, but I think we can get rid of the initialize method here as nothing here needs to happen async. Wdyt?

scopsy (approved):

PR #6853, opened 2024-11-05
perf(api, worker): Add indexes to preferences schema for lookup

rifont:

* Add indexes to preferences schema for lookup
* Only lookup V1 Preferences for Subscriber if V2 Preferences don't exist

### Screenshots

<details>

</details>

### DIFF
--- a/libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
+++ b/libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
@@ -58,13 +58,6 @@ export class GetSubscriberGlobalPreference {
   }> {
     let subscriberGlobalChannels: IPreferenceChannels;
     let enabled: boolean;
-    /** @deprecated */
-    const subscriberGlobalPreferenceV1 =
-      await this.subscriberPreferenceRepository.findOne({
-        _environmentId: command.environmentId,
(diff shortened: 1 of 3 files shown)


libs/application-generic/src/usecases/get-subscriber-global-preference/get-subscriber-global-preference.usecase.ts
```diff
@@ -78,6 +71,15 @@ export class GetSubscriberGlobalPreference {
       subscriberGlobalChannels = subscriberGlobalPreferenceV2;
       enabled = true;
     } else {
+      // Lookup V1 preferences only if V2 is not available
```
rifont, line 74: We save an extra DB call in some cases now (but all cases after migration) when V2 Preferences exist.

libs/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -132,6 +120,19 @@ export class GetSubscriberTemplatePreference {
       critical = subscriberWorkflowPreferenceV2.preferences?.all?.readOnly;
       enabled = true;
     } else {
+      // Lookup V1 preferences only if V2 is not available
```
rifont, line 123: Same here as above, but for Subscriber Workflow preferences

SokratisVidros (approved):

PR #6860, opened 2024-11-05
feat(dashboard): delete step editor functionality

ChmaraX:
- adds ability to delete a step + confirmation modal

https://github.com/user-attachments/assets/ca3cf889-ce36-480a-ada9-336374e7dd63


apps/dashboard/src/components/workflow-editor/steps/configure-step.tsx
```diff
@@ -11,13 +11,25 @@ import { motion } from 'framer-motion';
...
+    stepSlug: string;
   }>();
-  const { isReadOnly } = useWorkflowEditorContext();
+  const { isReadOnly, deleteStep } = useWorkflowEditorContext();
+
+  const [isDeleteModalOpen, setIsDeleteModalOpen] = useState(false);
+
+  const onDeleteStep = () => {
+    deleteStep(stepSlug);
+    navigate(buildRoute(ROUTES.EDIT_WORKFLOW, { environmentSlug: currentEnvironment?.slug ?? '', workflowSlug }));
```
ChmaraX, line 31: Focus back to workflow after opened step is deleted, otherwise the sidebar with deleted step will remain opened with undefined data.

apps/dashboard/src/components/workflow-editor/steps/delete-step-modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { RiAlertFill } from 'react-icons/ri';
+
+type DeleteStepModalProps = {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  onConfirm: () => void;
+  stepName: string;
+};
+
+export const DeleteStepModal = ({ open, onOpenChange, onConfirm, stepName }: DeleteStepModalProps) => {
```
ChmaraX, line 22: I'll create a generic `confirmation-modal` in separate PR, since we use this same modal also for promotion and probably it will be used in other places.

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -14,19 +15,42 @@ export const useFormAutoSave = <T extends FieldValues>({
...
+  const debouncedSave = useDebounce(save, 500);
+
+  const checkStepsDeleted = useCallback(() => {
+    const currentStepsLength = watchedData.steps?.length ?? 0;
+    const wasDeleted = previousStepsLength.current !== null && currentStepsLength < previousStepsLength.current;
+
+    previousStepsLength.current = currentStepsLength;
+
+    return wasDeleted;
+  }, [watchedData]);
```
ChmaraX: Logic to determine if step was deleted in order to not use debounced save.

If we used debounced save for deletion, user could potentially refresh in the 500ms time window and the step wouldn't be deleted, although I'm not sure if its worth the additional logic.

apps/dashboard/src/components/workflow-editor/steps/delete-step-modal.tsx
ChmaraX: https://github.com/novuhq/novu/pull/6861

SokratisVidros (approved):

PR #6861, opened 2024-11-05
feat(dashboard): generic confirmation modal

ChmaraX:
We are reusing the same confirmation modal in multiple places.


apps/dashboard/src/components/promote-workflow/promote-success-toast.tsx
```diff
@@ -14,7 +14,7 @@ export function PromoteSuccessToast({ workflow, onClose }: PromoteSuccessToastPr
 
   function onSwitchEnvironmentClick() {
     onClose();
-    switchEnvironment(oppositeEnvironment?._id || '');
+    switchEnvironment(oppositeEnvironment?.slug || '');
```
ChmaraX, line 17: This is a bug fix, we are using slugs now.

SokratisVidros (approved):

PR #6864, opened 2024-11-05
fix(api): Remove double logging of Bridge errors

SokratisVidros:
We log once in ExecuteBridgeRequest, so there is no need to log the error again in `get-bridge-status.usecase.ts`

### Screenshots
The fix should remove the log entry in the red-dotted box
Screenshot 2024-11-05 at 23 34 27

### DIFF
--- a/apps/api/src/app/bridge/bridge.controller.ts
+++ b/apps/api/src/app/bridge/bridge.controller.ts
@@ -50,13 +50,11 @@ export class BridgeController {
   @Get('/status')
   @UseGuards(UserAuthGuard)
   async health(@UserSession() user: UserSessionData) {
-    const result = await this.getBridgeStatus.execute(
+    return this.getBridgeStatus.execute(
       GetBridgeStatusCommand.create({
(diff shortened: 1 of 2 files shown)


apps/api/src/app/bridge/bridge.controller.ts
```diff
@@ -50,13 +50,11 @@ export class BridgeController {
   @Get('/status')
   @UseGuards(UserAuthGuard)
   async health(@UserSession() user: UserSessionData) {
-    const result = await this.getBridgeStatus.execute(
+    return this.getBridgeStatus.execute(
```
SokratisVidros, line 53: Note that the [HealthCheck](https://github.com/novuhq/novu/blob/452c5b8411dbc23231266cc62152f06e74250f8c/packages/framework/src/types/health-check.types.ts#L2) type has a status `error` and is not currently used. 

As part of future work, we will discuss whether Bridge errors must return a HealthCheck payload with status set to 'error'.

rifont (approved):
💯

PR #6868, opened 2024-11-06
chore(root): Remove @novu/automation library

SokratisVidros:
It seems to be unused.


SokratisVidros:
False interpretation, this package is used to generate a new provider.

PR #6870, opened 2024-11-06
fix(api): refactor SupportService to use environment variable for API…

merrcury:
This fixes the SupportService created on PlainClient

### DIFF
--- a/libs/application-generic/src/custom-providers/index.ts
+++ b/libs/application-generic/src/custom-providers/index.ts
@@ -91,7 +91,7 @@ export const distributedLockService = {
 export const supportService = {
   provide: SupportService,
   useFactory: async () => {
-    const service = new SupportService(process.env.PLAIN_SUPPORT_KEY);
+    const service = new SupportService();
 
     return service;
   },
--- a/libs/application-generic/src/custom-providers/index.ts
+++ b/libs/application-generic/src/custom-providers/index.ts
@@ -78,7 +78,7 @@ export const distributedLockService = {
       cacheInMemoryProviderService.useFactory();
 
     const service = new DistributedLockService(
-      factoryCacheInMemoryProviderService
+      factoryCacheInMemoryProviderService,
     );
 
     await service.initialize();
--- a/libs/application-generic/src/custom-providers/index.ts
+++ b/libs/application-generic/src/custom-providers/index.ts
@@ -21,7 +21,7 @@ export const featureFlagsService = {
 export const getFeatureFlag = {
   provide: GetFeatureFlag,
   useFactory: async (
-    featureFlagsServiceItem: FeatureFlagsService
+    featureFlagsServiceItem: FeatureFlagsService,
   ): Promise<GetFeatureFlag> => {
     const useCase = new GetFeatureFlag(featureFlagsServiceItem);
 
--- a/libs/application-generic/src/custom-providers/index.ts
+++ b/libs/application-generic/src/custom-providers/index.ts
@@ -5,7 +5,6 @@ import {
   CacheService,
   DistributedLockService,
   FeatureFlagsService,
-  SupportService,
 } from '../services';
 import { GetFeatureFlag } from '../usecases';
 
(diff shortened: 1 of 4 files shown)


jainpawan21 (approved):

libs/application-generic/src/custom-providers/index.ts
```diff
@@ -91,7 +91,7 @@ export const distributedLockService = {
 export const supportService = {
   provide: SupportService,
   useFactory: async () => {
-    const service = new SupportService(process.env.PLAIN_SUPPORT_KEY);
+    const service = new SupportService();
 
     return service;
   },
```
rifont: We can remove this custom provider now, it's not doing anything.

merrcury: Thats what I reckon.

libs/application-generic/src/services/support.service.ts
```diff
@@ -5,12 +5,11 @@ const LOG_CONTEXT = 'SupportService';
...
-  constructor(private plainKey?: string | null) {
-    if (this.plainKey) {
-      this.plainClient = new PlainClient({
-        apiKey: this.plainKey,
-      });
-    }
+  private readonly plainKey: string;
+  constructor() {
+    this.plainKey = process.env.PLAIN_SUPPORT_KEY;
+    this.plainClient = new PlainClient({ apiKey: this.plainKey });
```
rifont: How does the constructor behave when `PLAIN_SUPPORT_KEY` is _not_ present? Does this throw an error?

If it errors, the community edition will face issues. I would suggest only instantiating the client when `plainKey` is present in the environment. Then, conditionally access the upsert method like `this.plainClient?.upsertCustomer` in the upsert method

jainpawan21: @rifont SupportService is being used in SupportModule and SupportModule is imported only when in enterprise environment

merrcury: Made the Changes. One more thing with respect to invocation
For Community User, Its governed in main web only
```
export const IS_NOVU_PROD_STAGING = !IS_SELF_HOSTED && !API_ROOT.includes('localhost');
```

libs/application-generic/src/custom-providers/index.ts
```diff
@@ -22,7 +21,7 @@ export const featureFlagsService = {
 export const getFeatureFlag = {
   provide: GetFeatureFlag,
   useFactory: async (
-    featureFlagsServiceItem: FeatureFlagsService,
+    featureFlagsServiceItem: FeatureFlagsService
```
rifont:
```suggestion
    featureFlagsServiceItem: FeatureFlagsService,
```

libs/application-generic/src/custom-providers/index.ts
```diff
@@ -79,20 +78,11 @@ export const distributedLockService = {
       cacheInMemoryProviderService.useFactory();
 
     const service = new DistributedLockService(
-      factoryCacheInMemoryProviderService,
+      factoryCacheInMemoryProviderService
```
rifont: Removing formatting change to keep the diff clean (side note, it's strange that a Linting error isn't thrown here, it does in my local. We should fix that separately.
```suggestion
      factoryCacheInMemoryProviderService,
```

libs/application-generic/src/services/support.service.ts
```diff
@@ -41,7 +44,7 @@ export class SupportService {
       Logger.error(
         { emailAddress, fullName, error: res.error },
         res.error.message,
-        LOG_CONTEXT,
+        LOG_CONTEXT
```
rifont:
```suggestion
        LOG_CONTEXT,
```

libs/application-generic/src/services/support.service.ts
```diff
@@ -67,7 +70,7 @@ export class SupportService {
       Logger.error(
         { plainCustomerId, threadText, error: res.error },
         res.error.message,
-        LOG_CONTEXT,
+        LOG_CONTEXT
```
rifont:
```suggestion
        LOG_CONTEXT,
```

rifont (approved):
💯

libs/application-generic/src/services/support.service.ts
rifont: Thanks for clarifying, team. That makes sense.

PR #6873, opened 2024-11-06
feat(web): v3 dashboard opt-in widget

ChmaraX:
- adds v3 dashboard opt-in widget
- saves the opt-in state to Clerk user unsafe metadata

Related EE repo PR: https://github.com/novuhq/packages-enterprise/pull/238

image

image


apps/web/src/components/layout/components/v2/V3DashboardOptInWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+    });
+  };
+
+  function handleOptIn() {
+    const v3DashboardUrl = process.env.V3_DASHBOARD_URL;
+    if (!v3DashboardUrl || !user) return;
+
+    updateUserOptInStatus(V3DashboardOptInStatusEnum.OPTED_IN);
+    window.location.href = v3DashboardUrl;
+  }
```
ChmaraX: I'll implement mixpanel tracking in separate PR

apps/web/src/components/layout/components/v2/V3DashboardOptInWidget.tsx
```diff
@@ -0,0 +1,112 @@
+import { Card } from '@mantine/core';
+import { css } from '@novu/novui/css';
+import { Text, Title, Button, IconButton } from '@novu/novui';
+import { IconOutlineClose } from '@novu/novui/icons';
+import { useUser } from '@clerk/clerk-react';
+import { FeatureFlagsKeysEnum, V3DashboardOptInStatusEnum } from '@novu/shared';
+import { IS_SELF_HOSTED } from '../../../../config';
+import { useFeatureFlag } from '../../../../hooks';
+
+export function V3DashboardOptInWidget() {
```
SokratisVidros:
```suggestion
export function V2DashboardOptInWidget() {
```

We are not in V3 yet :p

apps/web/src/components/layout/components/v2/V3DashboardOptInWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const { user } = useUser();
+  const isV3DashboardEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V3_DASHBOARD_ENABLED);
```
SokratisVidros:
```suggestion
  const isV2DashboardEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V2_DASHBOARD_ENABLED);
```

SokratisVidros (approved):

packages/shared/src/types/clerk/types.ts
```diff
@@ -19,3 +19,17 @@ export type UserPublicMetadata = {
...
 };
+
+/**
+ * Unsafe metadata can be updated from the frontend directly
+ */
+export type UserUnsafeMetadata = {
+  v3DashboardOptInStatus?: V3DashboardOptInStatusEnum;
+};
+
+export enum V3DashboardOptInStatusEnum {
```
SokratisVidros: 💡 I'd keep all this state into a single variable in clerk metadata:

`dashboard_version: 'v1' | 'v2' | null | undefined.

V1 is opt-out, V2 is opt-in, null is dismissed or never opted-in.

apps/web/src/components/layout/components/v2/V3DashboardOptInWidget.tsx
ChmaraX: I kept it as V3 because all the figma designs are called "v3 dashboard" and we already use "V2" for something else. Imo this way its less confusing in code and communication.

packages/shared/src/types/clerk/types.ts
ChmaraX: This is inconclusive because user always uses one of the dashboards, either `v1` or `v2`.

The naming you proposed suggests that I can use that variable to show something deterministically based on dashboard version, which is not true because by default if user didn't interacted with the opt-in widget he would have no `dashboard_version`, therefore `undefined`. Thats why I think its better to tie the variable name directly to the opt-in operation not the dashboard version.

apps/web/src/components/layout/components/v2/NewDashboardOptInWidget.tsx
```diff
@@ -0,0 +1,112 @@
...
+        newDashboardOptInStatus: status,
+      },
+    });
+  };
+
+  function handleOptIn() {
+    const newDashboardUrl = process.env.NEW_DASHBOARD_URL;
+    if (!newDashboardUrl || !user) return;
+
+    updateUserOptInStatus(NewDashboardOptInStatusEnum.OPTED_IN);
```
LetItRock, line 35: ❓ **question:** is it possible to await for this function to complete? because I'm wondering whether the browser won't cancel the request when it does redirect to a new page

packages/shared/src/types/clerk/types.ts
```diff
@@ -19,3 +19,17 @@ export type UserPublicMetadata = {
...
+ * Unsafe metadata can be updated from the frontend directly
+ */
+export type UserUnsafeMetadata = {
+  newDashboardOptInStatus?: NewDashboardOptInStatusEnum;
+};
+
+export enum NewDashboardOptInStatusEnum {
+  OPTED_IN = 'opted_in', // user switched to the new dashboard
+  DISMISSED = 'dismissed', // user dismissed the opt-in widget
+  OPTED_OUT = 'opted_out', // user switched back to the old dashboard
```
LetItRock, line 33: ❓ **question:** aren't these two the same? like if I dismissed, I actually opted out

ChmaraX: How I thought about this is that `opted_out` is that you specifically decided to not use new dashboard after you switched to it and actually seen/tried it.

`dismissed` is when you click the `x` icon on the opt-in widget, that means you didnt visit the new dashboard, you just don't want to see the widget.

I think these two cases have different meaning in regards to tracking and user feedback. Although I would personally not give the option to dismiss the widget, but rather make it smaller and still visible.

apps/web/src/components/layout/components/v2/NewDashboardOptInWidget.tsx
ChmaraX: Good point, I'll try that when I finalize all the places where the redirect will happen and adjust according to that.

PR #6875, opened 2024-11-06
refactor(worker): Remove redundant Bridge error handling

rifont:

* Fix the `calculatedDelay` computation to respect `got` needs to review the `computedValue`, ensuring that the retriable status codes and error codes are respected
* Remove redundant bridge failure execution detail creation, as all failures are handled by the `afterResponse` hook

### Screenshots

_No retries for `400` error, execution logs now shows only 1 entry_
image

_No retries for `408` error, execution logs now shows only 1 entry_
image

_No retries for `502` error, execution logs now shows only 1 entry_
image

<details>

</details>


apps/worker/src/app/workflow/usecases/execute-bridge-job/execute-bridge-job.usecase.ts
```diff
@@ -193,99 +193,48 @@ export class ExecuteBridgeJob {
...
-        },
-        workflowOrigin,
-        environmentId,
-      }) as Promise<ExecuteOutput>;
-    } catch (error: any) {
-      Logger.error(error, 'Error sending Bridge request:', LOG_CONTEXT);
-
-      let raw: { retryCount?: number; statusCode?: number; message: string; url?: string };
-
-      if (error.response) {
```
rifont, line 244: This code path was never executed with recent changes to the Bridge error handling, where the `got` error is now captured and a NestJS exception is thrown. It's now redundant (similar to the removal of the extra health check log)

libs/application-generic/src/usecases/create-execution-details/types/index.ts
```diff
@@ -14,7 +14,6 @@ export enum DetailEnum {
   MESSAGE_CONTENT_SYNTAX_ERROR = 'Message content could not be generated due to syntax error in email editor',
   MESSAGE_CREATED = 'Message created',
   SUCCESSFUL_BRIDGE_RESPONSE_RECEIVED = 'Successful Bridge response received',
-  FAILED_BRIDGE_RETRY = 'Failed to get response from Bridge, will be retried',
```
rifont, line 17: Not used anymore.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -121,11 +121,16 @@ export class ExecuteBridgeRequest {
       timeout: DEFAULT_TIMEOUT,
       json: command.event,
       retry: {
-        calculateDelay: ({
-          attemptCount,
-        }: {
-          attemptCount: number;
-        }): number => {
+        calculateDelay: ({ attemptCount, computedValue }) => {
+          if (computedValue === 0) {
```
djabarovgeorge, line 145: This shouldn't happen in our case because we don't reduce the delay or use the 'limit' option, but it's a nice safeguard for the future.

libs/application-generic/src/usecases/create-execution-details/types/index.ts
djabarovgeorge: what will happen if, for example, Novu hits 408 Request Timeout on the bridge app?

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
rifont: It does happen, because we supply the `statusCodes` and `errorCodes` options. The lack of this check was causing the `statusCodes` and `errorCodes` configuration to be ignored, resulting in 3 retries always occurring. There's no need to retry a `400 Bad Request` response, for example. This configuration ensures that such responses are not retried anymore.

I will add a code comment to clarify this.

libs/application-generic/src/usecases/create-execution-details/types/index.ts
rifont: With this change, they will now only see a `Bridge execution failed` error x3. I will add a screenshot to demonstrate.

The current injection of the `afterResponse` hook doesn't expose an interface for reading that a retry occurred due to the retry condition being met, so some rework is necessary to continue using this error code.

I'm not convinced that much value is being added by this `"will be retried"` sentence as the retries occur immediately and will log immediately, so I decided to remove it.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -121,11 +121,16 @@ export class ExecuteBridgeRequest {
...
+            /*
+             * If the computed value is 0, the retry conditions were not met and we don't want to retry.
+             *
```
rifont:
```suggestion
             * If the computed value is 0, the retry conditions were not met and we don't want to retry.
             * The retry condition is only met when the response has a `statusCodes` or `errorCodes`
             * that matches the supplied retry configuration values.
```

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
rifont: I added a comment below, let me know if that provides more clarity.

libs/application-generic/src/usecases/create-execution-details/types/index.ts
rifont: Ok, thanks for calling this out.

The extra execution detail logging is in fact needed for reporting the case of `408 Request Timeout`. Without it, the step fails to execute without logging any execution details (see below screen).

image


I'm looking in to why.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
SokratisVidros: It ends up being overwhelming as we often need to opt out of everything that comes out of the box. We need more control and an opt-in approach. It's not actionable for now, I just wanted to share my point of view.

SokratisVidros (approved):

libs/application-generic/src/usecases/create-execution-details/types/index.ts
rifont: Resolved, see PR description.

apps/api/src/app/events/e2e/bridge-trigger.e2e.ts
```diff
@@ -329,11 +329,11 @@ contexts.forEach((context: Context) => {
       expect(messagesAfter.length).to.be.eq(0);
       const executionDetailsRequired = await executionDetailsRepository.find({
         _environmentId: session.environment._id,
-        status: ExecutionDetailsStatusEnum.WARNING,
+        status: ExecutionDetailsStatusEnum.FAILED,
       });
 
       let raw = JSON.parse(executionDetailsRequired[0]?.raw ?? '');
-      let error = raw.raw.data[0].message;
+      let error = raw.data[0].message;
```
rifont, line 336: This `raw.raw` stutter was unnecessary and made the UI more difficult to read.

libs/application-generic/src/utils/bridge.ts
```diff
@@ -38,64 +38,68 @@ export const isRegularDigestOutput = (
 
 export const BRIDGE_EXECUTION_ERROR = {
   INVALID_BRIDGE_URL: {
-    code: 'INVALID_BRIDGE_URL',
+    code: 'InvalidBridgeUrl',
```
rifont, line 41: It's easier to read a pascal-cased code on the UI, plus Framework uses PascalCase for all the error codes already and we should be consistent.

libs/application-generic/src/utils/bridge.ts
```diff
@@ -38,64 +38,68 @@ export const isRegularDigestOutput = (
...
   },
+  UNKNOWN_BRIDGE_REQUEST_ERROR: {
+    code: 'UnknownBridgeRequestError',
+    message: (url: string) => `Unknown bridge request error calling \`${url}\``,
+  },
+  UNKNOWN_BRIDGE_NON_REQUEST_ERROR: {
+    code: 'UnknownBridgeNonRequestError',
+    message: (url: string) =>
+      `Unknown bridge non-request error calling \`${url}\``,
+  },
```
rifont, line 104: Making sure that unknown cases also have an error code to refer to.

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -135,10 +161,6 @@ export class ExecuteBridgeRequest {
         statusCodes: RETRYABLE_HTTP_CODES,
         errorCodes: RETRYABLE_ERROR_CODES,
       },
-      hooks: {
-        afterResponse:
-          command.afterResponse !== undefined ? [command.afterResponse] : [],
-      },
```
rifont, line 141: A step to move away from depending on the `got` interface for errors.

djabarovgeorge (approved):
🚀

PR #6876, opened 2024-11-06
fix(framework): Stop validating controls for non previewed step

rifont:

* Due to some code misconfiguration for `skip` step logical execution, non-previewed steps were having controls validated. This was a bug as no controls are ever supplied to the non-executed step. This change removes controls validation for non-previewed steps

### Screenshots

<details>

</details>

### DIFF
--- a/packages/framework/src/client.test.ts
+++ b/packages/framework/src/client.test.ts
@@ -1350,7 +1350,15 @@ describe('Novu Client', () => {
         workflowId: 'test-workflow',
         stepId: 'active-step-id',
         subscriber: {},
-        state: [],
+        state: [
+          {
+            stepId: 'skipped-step-id',
+            outputs: {},
+            state: {
+              status: 'success',
+            },
+          },
+        ],
         payload: {},
         controls: {},
       };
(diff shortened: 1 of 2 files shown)


packages/framework/src/client.test.ts
```diff
@@ -1350,7 +1350,15 @@ describe('Novu Client', () => {
...
-        state: [],
+        state: [
+          {
+            stepId: 'skipped-step-id',
+            outputs: {},
+            state: {
+              status: 'success',
+            },
+          },
+        ],
```
rifont, line 1361: An empty `outputs` should have been supplied for this test, ensuring that a hydration can still occur for skipped steps. This fixes the assertion.

packages/framework/src/client.ts
```diff
@@ -31,6 +31,7 @@ import type {
   HealthCheck,
   Schema,
   Skip,
+  State,
```
rifont:
```suggestion
```

packages/framework/src/client.ts
```diff
@@ -277,25 +277,27 @@ export class Client {
...
       const step = this.getStep(event.workflowId, stepId);
-      const controls = await this.createStepControls(step, event);
       const isPreview = event.action === PostActionEnum.PREVIEW;
 
-      if (!isPreview && (await this.shouldSkip(options?.skip as typeof step.options.skip, controls))) {
-        if (stepId === event.stepId) {
-          // Only set the result when the step is the current step.
+      // Only evaluate a skip condition when the step is the current step and not in preview mode.
+      if (!isPreview && stepId === event.stepId) {
+        const controls = await this.createStepControls(step, event);
```
rifont, line 284: With this change, we ensure step controls are only validated during execution where the controls are guaranteed to be available for the current step being executed.

packages/framework/src/client.ts
```diff
@@ -277,25 +277,27 @@ export class Client {
...
 
-        /*
-         * Return an empty object for results when a step is skipped.
-         * TODO: fix typings when `skip` is specified to return `Partial<T_Result>`
-         */
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        return {} as any;
+          /*
+           * Return an empty object for results when a step is skipped.
+           * TODO: fix typings when `skip` is specified to return `Partial<T_Result>`
```
tatarco, line 296: just to make sure it doesn't bypass the second PR and results will be supplied if i provide them

tatarco (approved):

SokratisVidros (approved):

packages/framework/src/client.ts
rifont: Yes, this code branch is only for the non-preview case.

PR #6877, opened 2024-11-06
feat(dashboard): in-app editor form driven by BE schema

LetItRock:
Dashboard - In-App Editor - BE schema-driven form generation and validation.

In this PR:
- renamed a bunch of components 
- added the default values to the `dataSchema` returned from BE
- created the utils that parse the JsonSchema to Zod schema and parse JsonSchema to default values for the `react-hook-form`
- react-hook-form with dynamic schema and validation (hardcore)
- render the In-App UI based on `uiSchema` and component types
- update workflow on In-App Editor form submission

### Screenshots

https://github.com/user-attachments/assets/343fb836-4b29-459e-9fc2-9284efc4ac6a


apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
```diff
@@ -9,33 +9,36 @@ const redirectSchema = {
     url: {
       type: 'string',
       pattern: ABSOLUTE_AND_RELATIVE_URL_REGEX,
+      default: '',
```
LetItRock: Added the default values to the `dataSchema`, these are used to generate the default values for the form. The default values are required for all visible text fields by the `react-hook-form`, [ref](https://react-hook-form.com/docs/useform#defaultValues).

apps/dashboard/package.json
```diff
@@ -75,6 +75,7 @@
   "devDependencies": {
     "@clerk/types": "^4.6.1",
     "@eslint/js": "^9.9.0",
+    "@hookform/devtools": "^4.3.0",
```
LetItRock, line 79: A really useful tool to debug the `react-hook-form` state

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
```diff
@@ -40,14 +40,17 @@ export const AvatarPicker = forwardRef<HTMLInputElement, AvatarPickerProps>(({ i
...
-              <Avatar>
+              <Avatar className="p-[1px]">
                 <AvatarImage src={props.value as string} />
               </Avatar>
             ) : (
               <RiImageEditFill className="size-5" />
             )}
+            {error && (
+              <RiErrorWarningFill className="text-destructive outline-destructive absolute right-0 top-0 size-3 -translate-y-1/2 translate-x-1/2 rounded-full outline outline-1 outline-offset-1" />
+            )}
```
LetItRock, line 53: it was missing the error icon

apps/dashboard/src/components/primitives/form/form.tsx
```diff
@@ -101,28 +101,35 @@ const formMessageVariants = cva('flex items-center gap-1', {
   },
 });
 
-const FormMessage = React.forwardRef<HTMLParagraphElement, React.HTMLAttributes<HTMLParagraphElement>>(
-  ({ className, children, ...props }, ref) => {
-    const { error, formMessageId } = useFormField();
-    const body = error ? String(error?.message) : children;
+const FormMessagePure = React.forwardRef<
```
LetItRock, line 104: The `FormMessagePure` is a label + icon that is not dependent on the `FormField`, it's used in a few cases.

apps/dashboard/src/components/workflow-editor/steps/configure-step.tsx
```diff
@@ -78,7 +78,7 @@ const Step = () => {
   const { stepType: channel } = useStep();
   switch (channel) {
     case StepTypeEnum.IN_APP:
-      return <InApp />;
+      return <ConfigureInApp />;
```
LetItRock, line 105: renamed

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-action.tsx
```diff
@@ -0,0 +1,198 @@
...
+import { Separator } from '@/components/primitives/separator';
+import { URLInput } from '@/components/workflow-editor/url-input';
+import { cn } from '@/utils/ui';
+import { urlTargetTypes } from '@/utils/url';
+import { Editor } from '@/components/primitives/editor';
+
+const primaryActionKey = 'primaryAction';
+const secondaryActionKey = 'secondaryAction';
+
+export const InAppAction = () => {
```
LetItRock, line 33: This file and component were renamed from `ActionPicker`.
The changes that were added are: the form fields and errors; handling onChange logic;

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { getComponentByType } from '@/components/workflow-editor/steps/component-utils';
+
+const avatarKey = 'avatar';
+const subjectKey = 'subject';
+const bodyKey = 'body';
+const redirectKey = 'redirect';
+const primaryActionKey = 'primaryAction';
+const secondaryActionKey = 'secondaryAction';
+
+export const InAppEditor = ({ uiSchema }: { uiSchema?: UiSchema }) => {
```
LetItRock, line 13: The component that renders the In-App Editor UI form dynamically based on the `uiSchema`.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { InAppEditorPreview } from '@/components/workflow-editor/steps/in-app/in-app-editor-preview';
+import { useUpdateWorkflow } from '@/hooks/use-update-workflow';
+import { buildDynamicZodSchema, buildDefaultValues } from '@/utils/schema';
+import { InAppEditor } from '@/components/workflow-editor/steps/in-app/in-app-editor';
+import { showToast } from '@/components/primitives/sonner-helpers';
+import { ToastIcon } from '@/components/primitives/sonner';
+
+const tabsContentClassName = 'h-full w-full px-3 py-3.5';
+
+export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; step: StepDataDto }) => {
```
LetItRock, line 22: This is the main component that is rendered in the In-App drawer, it consists of the editor and preview tabs, react-hook-form and submission logic.

apps/dashboard/src/components/workflow-editor/url-input.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+type URLInputProps = Omit<InputProps, 'value' | 'onChange' | 'size'> & {
+  options: string[];
+  asEditor?: boolean;
+  urlKey: string;
+  targetKey: string;
+  withHint?: boolean;
+} & Pick<InputFieldProps, 'size'>;
+
+export const URLInput = ({
```
LetItRock, line 21: I've moved this component from the primitives as it's based on the form fields, and I couldn't make it work without the form fields.

apps/dashboard/src/utils/schema.ts
```diff
@@ -0,0 +1,167 @@
...
+  // remove empty strings
+  return stringValue.transform((val) => (val === '' ? undefined : val));
+};
+
+/**
+ * Transform JSONSchema to Zod schema.
+ * The function will recursively build the schema based on the JSONSchema object.
+ * It removes empty strings and objects with empty required fields during the transformation phase after parsing.
+ */
+export const buildDynamicZodSchema = (obj: JSONSchema): z.AnyZodObject => {
```
LetItRock, line 99: This is the util used to parse the `dataSchema` (which is JSONSchema type) to `Zod.` The converted schema is used in the form for the validation and then transformation after the parsing (success validation) phase. The transformed value is returned in the `onSubmit` handler.

apps/dashboard/src/utils/schema.ts
```diff
@@ -0,0 +1,167 @@
...
+    return { ...acc, [key]: zodValue };
+  }, {});
+
+  return z.object({ ...keys });
+};
+
+/**
+ * Build default values based on the JSONSchema object.
+ */
+export const buildDefaultValues = (obj: JSONSchema): object => {
```
LetItRock: This util takes the `dataSchema: JSONSchema` and retrieves the default values from it.

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
```diff
@@ -40,14 +40,17 @@ export const AvatarPicker = forwardRef<HTMLInputElement, AvatarPickerProps>(({ i
     <div className="space-y-2">
       <Popover modal={true} open={isOpen} onOpenChange={setIsOpen}>
         <PopoverTrigger asChild>
-          <Button variant="outline" size="icon" className="text-foreground-600 size-10">
+          <Button variant="outline" size="icon" className="text-foreground-600 relative size-10">
             {props.value ? (
-              <Avatar>
+              <Avatar className="p-[1px]">
```
desiprisg: You can do `p-px` here :D

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
```diff
@@ -40,14 +40,17 @@ export const AvatarPicker = forwardRef<HTMLInputElement, AvatarPickerProps>(({ i
...
                 <AvatarImage src={props.value as string} />
               </Avatar>
             ) : (
               <RiImageEditFill className="size-5" />
             )}
+            {error && (
+              <RiErrorWarningFill className="text-destructive outline-destructive absolute right-0 top-0 size-3 -translate-y-1/2 translate-x-1/2 rounded-full outline outline-1 outline-offset-1" />
```
desiprisg, line 52: You can do `m-auto inset-0` to center this a bit cleaner i think.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-redirect.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <div className="flex flex-col gap-1">
+      <FormLabel hint="">Redirect URL</FormLabel>
+      <URLInput
+        options={urlTargetTypes}
+        placeholder="/tasks/{{taskid}}"
+        size="md"
+        asEditor
+        urlKey="redirect.url"
+        targetKey="redirect.target"
```
desiprisg: Maybe name these `urlFieldName` and `targetFieldName`, or a `fieldNames` prop that has both? Could be a nice convention for components with multiple inputs.

apps/dashboard/src/components/workflow-editor/url-input.tsx
desiprisg: Makes sense. Should we create an `elements` directory for components like this? If it makes sense outside of the workflow editor of course.

apps/dashboard/src/utils/schema.ts
desiprisg, line 1: Wow. Is there like no easier way to do this ?

desiprisg (approved):

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <span>In-app Template</span>
+      </div>
+      <div className="flex flex-col gap-1 rounded-xl border border-neutral-100 p-1">
+        <div className="flex gap-1">
+          {avatar && getComponentByType({ component: avatar.component })}
+          {subject && getComponentByType({ component: subject.component })}
+        </div>
+        {body && getComponentByType({ component: body.component })}
+        {(primaryAction || secondaryAction) &&
+          getComponentByType({
```
tatarco, line 42: doesn't it means that if the secondary key is missing it will still show both options in the UI ? same for primary

apps/dashboard/src/utils/schema.ts
tatarco: if you need zod why do we serve jsonSchema? @SokratisVidros

apps/dashboard/src/utils/schema.ts
tatarco: @SokratisVidros the work done here shows how unnecessary was using the json schema if we need to manually parse it in the FE, we should have made the zod available via shared library and convert the zod to jsonschema for the library usage

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <div className="flex items-center gap-2.5">
+        <RiPencilRuler2Line className="text-foreground-950 size-5 p-0.5 text-sm font-medium" />
+        <span>In-app Template</span>
+      </div>
+      <div className="flex flex-col gap-1 rounded-xl border border-neutral-100 p-1">
+        <div className="flex gap-1">
+          {avatar && getComponentByType({ component: avatar.component })}
+          {subject && getComponentByType({ component: subject.component })}
+        </div>
+        {body && getComponentByType({ component: body.component })}
```
BiswaViraj: 🗒 **note (non-blocking):** We might need to handle the case where the `Subject` will be missing and move the body next to avatar in that case‏

SokratisVidros: Correct. If both avatar and subject are missing their wrapping div should also be removed as well.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
SokratisVidros: Hmmm, also when there is no action set yet, the widget should be displayed.

apps/dashboard/src/utils/schema.ts
SokratisVidros: I guess that the FE prefers to work with Zod for client-side validations so they transform the JSON schema to zod.

apps/dashboard/src/utils/schema.ts
SokratisVidros: I am not sure I understand the comment. If we hardcode the values per step how can we support dynamic configurations?

SokratisVidros (approved):

apps/dashboard/src/utils/schema.ts
LetItRock: @tatarco, the Zod schema is not serializable; please check [this discussion](https://github.com/colinhacks/zod/discussions/2030). There are tools that can convert Zod <> JSON schema and vice versa, but they are not "perfect" and require some runtime evaluation.

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
LetItRock: still learning :D

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
LetItRock: I'm not sure I get it, we want to show the error icon in the top right corner of the avatar. Can you please clarify?

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
LetItRock: > We might need to handle the case where the Subject will be missing and move the body next to avatar in that case

@BiswaViraj I don't think that moving the body up in that case is correct, because it will create a "gap" on the left side of the form. I would rather keep the `avatar` and `body` in separate rows.

However, we can support what @SokratisVidros is saying and remove unnecessary div.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
LetItRock: > doesn't it means that if the secondary key is missing it will still show both options in the UI ? same for primary

@tatarco No, this code is just responsible for rendering or not the "whole" component, but it has the logic inside that determines which buttons to render or not.

Screenshot 2024-11-07 at 11 04 14

LetItRock: > Hmmm, also when there is no action set yet, the widget should be displayed.

@SokratisVidros `primaryAction || secondaryAction` variables are taken from the `controls.uiSchema`, which controls what components to render. But the `controls.dataSchema` default values or `controls.values` do control the "state" of the component, meaning that when there are no `control.values`, the component will still be rendered with a "no action" state.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
BiswaViraj: I am unsure how it will create a gap on the left side?
Also we are doing the same on the Inbox as well, if the subject is missing we are putting the body in that place which aligns the UI to have Avatar and Body on the same line

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
LetItRock: you can see that in the attached video ;)

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-redirect.tsx
LetItRock: Nice, I like it, I will change that.

apps/dashboard/src/components/workflow-editor/url-input.tsx
LetItRock: Let's see if there will be a need to reuse this one.

apps/dashboard/src/utils/schema.ts
LetItRock: There is a [json-schema-to-zod](https://www.npmjs.com/package/json-schema-to-zod) lib, but it generates the string JS module that you have to evaluate at runtime, and there are some limitations.
Also this custom code has additional "transformations" that are applied on the form object after the "succesfull validation" and will be returned in the "onSuccess" handler of the form.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
LetItRock: Screenshot 2024-11-07 at 11 27 45

@BiswaViraj check the screenshot, also to cover that case it will bloat the code a bit

BiswaViraj: Ok got it, my bad confused it with the Preview 🤦🏼 
For the editor all the fields will be available always.

BiswaViraj (approved):

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
desiprisg: ah i thought you were centering. I think that you can do `inset: 0 0 auto auto;`

apps/dashboard/src/utils/schema.ts
rifont: @tatarco is right. We should use the [JSON Schema binding](https://github.com/vtex/react-hook-form-jsonschema) for React-hook form rather than transform to a Zod schema, else we risk loss of validation strictness.

PR #6879, opened 2024-11-07
refactor(framework): Make discovery completely asynchronous

rifont:

* Make discovery completely asynchronous
  * Asynchronous discovery is necessary to allow for use of dynamic awaitable imports (`await import(...)`) of peer dependencies to provide an optimal DX, where no manual dependency injection is required.
* Only log out discovered workflows that are served (move the `prettyPrintDiscovery` to the `client`. This also lays the groundwork for making the pretty discovery use a custom logger

### Screenshots

`next dev --turbopack` compatibility is a step closer now, erroring out with `module not found` rather than the previous `require` polyfill error.Some further investigation is to identify the root cause for the new problem. Judging by the docs at https://nextjs.org/docs/messages/module-not-found, it's probably:
`The module you're trying to import uses Node.js specific modules, for example dns, outside of getStaticProps / getStaticPaths / getServerSideProps`. [Here](https://github.com/search?q=repo%3Anovuhq%2Fnovu%20path%3A%2F%5Epackages%5C%2Fframework%5C%2Fsrc%5C%2F%2F%20node%3A&type=code) is the current usage of `node` built-in modules (`node:crypto` is used at runtime for HMAC header validation, `node:util/types` is used for an error type guard).

```bash
> next dev --turbopack --port 4000

   ▲ Next.js 15.0.2 (Turbopack)
   - Local:        http://localhost:4000
   - Environments: .env.local

 ✓ Starting...
 ✓ Ready in 748ms
 ○ Compiling /api/novu ...
 ✓ Compiled /api/novu in 555ms
 ⨯ ./src/app/api/novu/route.ts:2:1
Module not found: Can't resolve '@novu/framework'
  1 | import { serve } from '@novu/framework/next';
> 2 | import { workflow } from '@novu/framework';
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  3 |
  4 | const myWorkflow = workflow('welcome-onboarding-2', async ({ step }) => {
  5 |   await step.inApp('send-in-app', async () => ({ body: 'Welcome to our platform!', }));



https://nextjs.org/docs/messages/module-not-found
```

<details>

</details>


packages/framework/src/validators/zod.validator.ts
```diff
@@ -27,10 +27,9 @@ export class ZodValidator implements Validator<ZodSchema> {
     }
   }
 
-  transformToJsonSchema(schema: ZodSchema): JsonSchema {
+  async transformToJsonSchema(schema: ZodSchema): Promise<JsonSchema> {
     try {
-      // eslint-disable-next-line global-require
-      const { zodToJsonSchema } = require('zod-to-json-schema') as typeof import('zod-to-json-schema');
+      const { zodToJsonSchema } = await import('zod-to-json-schema');
```
rifont, line 32: Begone, dynamic `require`!

SokratisVidros (approved):

PR #6883, opened 2024-11-07
feat(api): Billing alerts on usage emails

scopsy:

- Adds a new @novu/notifications package to manage workflows
- Added a novu endpoint handler connected to our ORG in prod
- Added controllerDecorators for the framework SDK to support custom decorators for nest applications (solving the nestjs swagger issue)
- Added React email boilerplate. 
### Screenshots

CleanShot 2024-11-07 at 12 28 19@2x

https://github.com/novuhq/packages-enterprise/pull/239 

<details>

</details>


.vscode/settings.json
```diff
@@ -30,5 +30,6 @@
   },
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "vsicons.presets.nestjs": true
```
scopsy, line 34: Just pretty icons :)

apps/api/.swcrc
```diff
@@ -1,7 +1,13 @@
...
   "$schema": "https://swc.rs/schema.json",
   "jsc": {
-    "target": "es5"
+    "target": "es5",
+    "parser": {
+      "syntax": "typescript",
+      "tsx": true,
+      "decorators": true,
+      "dynamicImport": true
+    }
```
scopsy, line 10: Was required for React Email to work

apps/api/nest-cli.json
```diff
@@ -8,6 +8,7 @@
     "builder": {
       "type": "swc",
       "options": {
+        "extensions": [".js", ".ts", ".jsx", ".tsx"],
```
scopsy, line 11: Supporting tsx compilation

apps/api/src/app.module.ts
```diff
@@ -149,6 +154,19 @@ if (process.env.NODE_ENV === 'test') {
...
 }
 
+modules.push(
+  NovuModule.register({
+    apiPath: '/api/novu',
+    client: new Client({
+      secretKey: process.env.NOVU_INTERNAL_SECRET_KEY,
+      strictAuthentication:
+        process.env.NODE_ENV === 'production' || process.env.NOVU_STRICT_AUTHENTICATION_ENABLED === 'true',
+    }),
```
scopsy, line 164: Client is instantiated manually here

libs/notifications/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@novu/notifications",
```
scopsy, line 2: New package for all notification workflows to allow reusing them for triggering across the monorepo

packages/framework/src/servers/nest/nest.interface.ts
```diff
@@ -2,4 +2,5 @@ import type { ServeHandlerOptions } from '../../handler';
 
 export type NovuModuleOptions = ServeHandlerOptions & {
   apiPath: string;
+  controllerDecorators?: ClassDecorator[];
```
scopsy, line 5: This allows customers to add custom decorators to the controller and fixed the swagger conflict.

packages/framework/src/servers/nest/nest.utils.ts
rifont, line 1: Did we explore using the builtin `@nestjs/common` [`applyDecorators`](https://github.com/nestjs/nest/blob/master/packages/common/decorators/core/apply-decorators.ts) function?

scopsy: Yes, unfortunately it was not working and messing up the controller class construct when used in a dynamic way.

libs/notifications/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "build:module": "tsc -p tsconfig.module.json",
+    "fix": "run-s fix:*",
+    "fix:prettier": "prettier \"src/**/*.ts\" --write",
+    "fix:lint": "eslint src --ext .ts --fix",
+    "test": "jest --passWithNoTests",
+    "lint": "eslint src",
+    "test:watch": "jest --watch",
+    "watch:build": "tsc -p tsconfig.json -w",
+    "reset-hard": "git clean -dfx && git reset --hard && pnpm install",
+    "start:studio": "npx novu@latest dev --port 3000 --route /v1/api/novu"
```
rifont: Could we point to the local monorepo `novu` CLI package command instead to fully dogfood the dev process?

libs/notifications/src/workflows/usage-limits/email.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+      <Text className="text-[14px] leading-[24px] text-black">
+        To ensure uninterrupted service and access to additional features, we recommend upgrading your plan before
+        reaching the limit.
+      </Text>
+
+      <Section className="mb-[32px] mt-[32px] text-center">
+        <Button
+          className="rounded bg-[#000000] px-5 py-3 text-center text-[12px] font-semibold text-white no-underline"
+          href={'https://dashboard.novu.co/settings/billing'}
```
rifont, line 32: Should we inject the dashboard URL via payload instead to support EU?

libs/notifications/src/workflows/usage-limits/email.tsx
```diff
@@ -0,0 +1,57 @@
...
+        >
+          Upgrade Your Plan
+        </Button>
+      </Section>
+
+      <Text className="text-[12px] leading-[20px] text-gray-500">
+        Note: Once you reach 100% utilization, notifications will be blocked until you upgrade or the next billing cycle
```
rifont: Improving wording
```suggestion
        Note: Once you consume 100% of your monthly usage limit, notifications will be blocked until you upgrade or the next billing cycle begins.
```

libs/notifications/src/workflows/usage-limits/usage-limits.workflow.ts
```diff
@@ -0,0 +1,31 @@
+import { workflow } from '@novu/framework';
+import { z } from 'zod';
+import { renderUsageLimitsEmail } from './email';
+
+export const usageLimitsWorkflow = workflow(
+  'usage-limits',
+  async ({ step, payload }) => {
```
rifont, line 7: Should we also send a simple in-app notification to the customer?

packages/framework/src/servers/nest/nest.interface.ts
rifont: Awesome!

packages/framework/src/servers/nest/nest.utils.ts
```diff
@@ -0,0 +1,9 @@
+import { Type } from '@nestjs/common';
+
+export function applyDecorators<T>(baseClass: Type<T>, decorators: Array<ClassDecorator> = []): Type<T> {
+  return decorators.reduce((decoratedController, decorator) => {
+    const result = decorator(decoratedController);
```
rifont: I think it can be more generic than referencing `controller`

```suggestion
  return decorators.reduce((decoratedClass, decorator) => {
    const result = decorator(decoratedClass);
```

libs/notifications/package.json
scopsy: Sure!

libs/notifications/src/workflows/usage-limits/email.tsx
scopsy: Yes, was thinking about this. How does Eu works today with stripe webhooks?

rifont: I think Stripe webhooks are irrelevant - we can just read the `FRONT_BASE_URL` environment variable ([ref](https://github.com/novuhq/novu/blob/next/apps/api/src/.env.test#L46)) in the trigger logic to dynamically set the Dashboard URL

libs/notifications/src/workflows/usage-limits/usage-limits.workflow.ts
scopsy: Yes!

.github/workflows/dev-deploy-api.yml
```diff
@@ -133,6 +133,26 @@ jobs:
...
+          # The secret key used to authenticate with Novu Cloud
+          # To get the secret key, go to https://web.novu.co/api-keys.
+          # Required.
+          secret-key: ${{ secrets.NOVU_INTERNAL_SECRET_KEY }}
+
+          # The publicly available endpoint hosting the bridge application
+          # where notification entities (eg. workflows, topics) are defined.
+          # Required.
+          bridge-url: https://api.novu-staging.co/v1/bridge/novu
+
```
rifont, line 155: 😍

apps/api/src/.example.env
```diff
@@ -84,3 +84,4 @@ CLERK_LONG_LIVED_TOKEN=
 
 TUNNEL_BASE_ADDRESS=
 PLAIN_SUPPORT_KEY='PLAIN_SUPPORT_KEY'
+NOVU_SECRET_KEY=
```
rifont: Typo, incorrect env var key

```suggestion
NOVU_INTERNAL_SECRET_KEY =
```

libs/application-generic/package.json
```diff
@@ -83,6 +83,8 @@
     "launchdarkly-node-server-sdk": "^7.0.1",
     "lodash": "^4.17.15",
     "mixpanel": "^0.17.0",
+    "nanoid": "^3.1.20",
+    "nestjs-graceful-shutdown": "^1.0.3",
```
rifont: ❓ **question:** `nestjs-graceful-shutdown` appears to be a newly added dependency. Was this intentional?‏

libs/notifications/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "scripts": {
+    "build": "run-p build:*",
+    "build:main": "tsc -p tsconfig.json",
+    "build:module": "tsc -p tsconfig.module.json",
+    "fix": "run-s fix:*",
+    "fix:prettier": "prettier \"src/**/*.ts\" --write",
+    "fix:lint": "eslint src --ext .ts --fix",
+    "test": "jest --passWithNoTests",
+    "lint": "eslint src",
+    "test:watch": "jest --watch",
```
rifont: Preparing for better test tooling 😉

```suggestion
    "test": "vitest --passWithNoTests",
    "lint": "eslint src",
    "test:watch": "vitest --watch",
```

libs/notifications/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@novu/framework": "workspace:*",
+    "@novu/shared": "workspace:*",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "zod": "^3.23.8"
+  },
+  "devDependencies": {
+    "@types/react": "^18.2.0",
+    "@types/react-dom": "^18.2.0",
+    "@types/jest": "^29.5.2",
```
rifont: As above, let's use `vitest` here if necessary.
```suggestion
```

libs/notifications/src/workflows/usage-limits/usage-limits.workflow.ts
```diff
@@ -0,0 +1,47 @@
...
+          title: controls.title,
+          body: controls.body,
+        };
+      },
+      {
+        controlSchema: z.object({
+          title: z.string().default('You are approaching your usage limits'),
+          body: z.string().default('You have used {{payload.percentage}}% of your monthly events'),
+        }),
+      }
```
rifont: I would expect to see a type error here, it should be `subject`
```suggestion
      async (controls) => {
        return {
          subject: controls.subject,
          body: controls.body,
        };
      },
      {
        controlSchema: z.object({
          subject: z.string().default('You are approaching your usage limits'),
          body: z.string().default('You have used {{payload.percentage}}% of your monthly events'),
        }),
      }
```

packages/framework/src/servers/nest/nest.module.ts
```diff
@@ -43,7 +44,7 @@ export class NovuModule extends NovuBaseModule {
   static register(options: typeof OPTIONS_TYPE, customProviders?: Provider[]) {
     const superModule = super.register(options);
 
-    superModule.controllers = [NovuController];
+    superModule.controllers = [applyDecorators(NovuController, options.controllerDecorators || [])];
     superModule.providers?.push(registerApiPath, NovuClient, NovuHandler, ...(customProviders || []));
     superModule.exports = [NovuClient, NovuHandler];
```
rifont, line 50: ⚠ **issue:** I can't add a comment on it because there's no diff, but we also need to support the injected `controllerDecorators` on the `registerAsync` method too to ensure consistency on both APIs.‏

libs/application-generic/package.json
scopsy: I think it was removed, and a bad merge conflict. Thanks!

libs/notifications/package.json
scopsy: 💪

libs/notifications/src/workflows/usage-limits/usage-limits.workflow.ts
scopsy: Hmm strange, none happened :(

packages/framework/src/servers/nest/nest.module.ts
scopsy: Good point! I couldn't follow the options logic or where they are defined. Should it come just from the useFactory?

rifont: I had some trouble here too when I implemented the NestJS Framework module. I think it will require something like calling the `super.registerAsync` method and accessing the defined options on the super class, but I'm not certain.

rifont: I will check this on my local soon.

rifont: Check out the NestJS [`HttpModule`](https://github.com/nestjs/axios/blob/457584e0a1673f6fae44d7be8e246a50525f7f6f/lib/http.module.ts#L60), this covers the same pattern we need (StackOverflow [ref](https://stackoverflow.com/a/66164772)). They've implemented their `registerAsync` a little more flexibly without using the NestJS built-in `ConfigurableModuleBuilder`, which grants access to async options more easily.

This is the approach we need to take for the `customProviders` and `controllerDecorators` properties. We can then deprecate the second method parameter which supplies only `customProviders` in favour of a single configuration object for both `register` + `registerAsync`.

scopsy: Couldn't make it work :| I'm going to merge this PR as I need to test the Stripe behaviour on staging, and will raise a separate PR for the registerAsync behaviour later today.

rifont: Ok sounds good, I'll publish Framework from 1 commit this branch's merge commit to `next` to avoid releasing this change in that case. We should publish the feature in totality alongside the docs.

PR #6887, opened 2024-11-07
feat(api): invalidate stale workflows

djabarovgeorge:
### Screenshots

<details>

</details>


apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -90,6 +96,47 @@ export class Sync {
...
+    const discoverWorkflowIdentifiers = discover.workflows.map((workflow) => workflow.workflowId);
+    const staleWorkflows = existingWorkflows.filter(
+      (_worklfow) => !discoverWorkflowIdentifiers.includes(_worklfow.triggers[0].identifier)
+    );
+    const staleWorkflowIdentifiers = staleWorkflows.map((workflow) => workflow.triggers[0].identifier);
+    const staleWorkflowIds = staleWorkflows.map((workflow) => workflow._id);
+
+    for (const identifier of staleWorkflowIdentifiers) {
+      await this.invalidateCacheService.invalidateByKey({
+        key: buildNotificationTemplateIdentifierKey({
```
SokratisVidros: The update-workflow.command that is handling existing workflows already invalidates the cache for both keys - See [here](https://github.com/novuhq/novu/blob/8e35dad2c5675301d67b4cbfb7ff9af3eb001137/libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts#L212).

I suggest we apply a more surgical solution to avoid fetching working multiple times from the database that adds extra DB requests.

SokratisVidros (changes requested):

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
djabarovgeorge: You make a good point. The data that gets validated on updates should indeed be part of what we check. 

To illustrate today's issue, here’s the scenario we need to address:

1. A trigger is sent to a non-existent workflow with ID "best-workflow." On the trigger, we try to fetch the workflow, but since it doesn’t exist, we store `null` in the cache.
2. We then sync "best-workflow."
3. When we trigger "best-workflow" again, we pull the `null` value from the cache.

To resolve this, [we could prevent `null` from being cached on creation](https://github.com/novuhq/novu/pull/6887/commits/85a1385d044795f3eaedf9432d762dcb6220b698).

While reviewing the issue, I thought we had another problem, but I realized it's not a bug—just an instance of stale memory that we could release. Now, I’m considering [removing the extra cache invalidation step](https://github.com/novuhq/novu/pull/6887/commits/c0d66c3a3df679ab6938b49ffd923d9c157aa5a6) and relying on the TTL to clear it, which would simplify the logic and reduce the load on Redis. 

What do you think?

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -111,21 +115,40 @@ export class Sync {
...
+      this.deleteWorkflow.execute(
+        DeleteWorkflowCommand.create({
+          environmentId: command.environmentId,
+          organizationId: command.organizationId,
+          userId: command.userId,
+          workflowId: workflow._id,
+        })
+      )
+    );
+    const invalidateCachesByIdentifiersPromises = workflowsToDelete.map((workflow) =>
```
SokratisVidros: This invalidation should be added in the [delete-workflow usecase](https://github.com/novuhq/novu/pull/6887/files#diff-b97ae3759b8dec4d0c5057b0e10ca4798437b9c7896dacf049bf8da4de7f8000R130). The cache clearance should applies every time a workflow is deleted, not only via sync.

SokratisVidros: Moreover, the delete-workflow use case needs to delete from the preferences collection all entries of type `WORKFLOW_RESOURCE` by _environmentId and workflow internal id. This is irrelevant to this bug, but as I reviewed the code, it's an omission we can easily tackle.

SokratisVidros (changes requested):

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
djabarovgeorge: > This invalidation should be added in the delete-workflow usecase. The cache clearance should applies every time a workflow is deleted, not only via sync.

First of all please note the last part i mentioned in this comment.
https://github.com/novuhq/novu/pull/6887#discussion_r1833005669

We can't invalidate these IDs because, in the delete workflow use case, we don't know which specific IDs need to be deleted. We're missing the current list of workflows that should be synced with the full snapshot list. Even if we had this data, we would still invalidate unnecessary records since we already know the exact IDs that need invalidation.

djabarovgeorge: > Moreover, the delete-workflow use case needs to delete from the preferences collection all entries of type WORKFLOW_RESOURCE by _environmentId and workflow internal id. This is irrelevant to this bug, but as I reviewed the code, it's an omission we can easily tackle.

Agree i will take a look at it.

djabarovgeorge: Done.
I deleted the old bridge.deleteWorkflow so we could reuse one reusable delete workflow from app-gen.

libs/application-generic/src/usecases/workflow/delete-workflow/delete-workflow.usecase.ts
```diff
@@ -0,0 +1,136 @@
...
+      GetWorkflowByIdsCommand.create({
+        ...command,
+        identifierOrInternalId: command.identifierOrInternalId,
+      }),
+    );
+
+    await this.deleteRelatedEntities(command, workflowEntity);
+  }
+
+  private async deleteRelatedEntities(
```
SokratisVidros, line 45: @djabarovgeorge we need a transaction here in the same fashion we did it [here](https://github.com/novuhq/novu/blob/5a0efd7f02cb4c579c65439751daefa029cc551e/apps/api/src/app/subscribers/usecases/remove-subscriber/remove-subscriber.usecase.ts#L58). to ensure that all of these are deleted atomically.

libs/application-generic/src/usecases/workflow/delete-workflow/delete-workflow.usecase.ts
```diff
@@ -0,0 +1,136 @@
...
+    command: DeleteWorkflowCommand,
+    workflow: NotificationTemplateEntity,
+  ) {
+    await this.deleteControlValues(command, workflow);
+    await this.deleteMessageTemplates(workflow, command);
+    await this.deletePreferences(command, workflow);
+    await this.invalidateCacheForWorkflow(workflow, command);
```
SokratisVidros: Please invalidate the cache first as its an isolated task that should be done regardless of what happens in Mongo. Note that cache invalidation can't be part of the transaction as it targets a different DB.

SokratisVidros (approved):
Great clean up @djabarovgeorge 👏 👏 👏

libs/application-generic/src/usecases/workflow/delete-workflow/delete-workflow.usecase.ts
djabarovgeorge: done

libs/application-generic/src/usecases/workflow/delete-workflow/delete-workflow.usecase.ts
djabarovgeorge: done

PR #6889, opened 2024-11-07
refactor(api): Use `UpdatePreference` use-case for all Subscriber Preference updates

rifont:

* Use `UpdatePreference` use-case for all Subscriber Preference updates
  * The use-cases previously used by Subscriber and Widget API were not updating V2 Preferences. This resulted in cases of Subscriber Preference reads to incorrectly return default preferences
* Skip all tests that asserted on the top-level `enabled` flag
  * The `enabled` flag was introduced at a time before `critical` workflows were available. Setting the `enabled` flag to false on a preference set previously had the effect of preventing that Subscriber's preference settings from being read during notification delivery preference resolution. The Workflow `critical` flag handles this use-case in a more read-optimal way by requiring only a single Workflow resource to be mutated, rather than all Subscriber resources. I have left these tests in the codebase for now, we can debate further on complete deletion of the `enabled` flag as part of [`NV-4597`](https://linear.app/novu/issue/NV-4597/remove-all-usage-of-v1-preferences-in-codebase)
* Add instrumentation for the `UpdatePreference` use-case

### Screenshots

<details>

</details>


apps/api/src/app/subscribers/e2e/update-global-preference.e2e.ts
```diff
@@ -77,7 +77,14 @@ describe('Update Subscribers global preferences - /subscribers/:subscriberId/pre
     expect(response.data.data.preference.channels).to.eql({});
   });
 
-  it('should update user global preference and disable the flag for the future channels update', async function () {
+  it('should unset all preferences when the preferences object is empty', async function () {
+    const response = await updateGlobalPreferences({}, session);
+
+    expect(response.data.data.preference.channels).to.eql({});
+  });
```
rifont, line 84: This test case asserts that empty objects can be supplied when updating global preferences in relation to [this `UpsertPreferences` change](https://github.com/novuhq/novu/pull/6889/files#diff-bccd184dcf09c9cf0b60d2edcdcfac879d8f678d9b8655c8e471ff6fc3f3db0bR56)

apps/api/src/app/subscribers/subscribers.controller.ts
rifont, line 1: As you will see further down, it's not ideal that we have some mapping logic in the `controller` level. I decided that's a suitable approach for now as we'll be performing a Subscriber API overhaul soon.

apps/api/src/app/subscribers/usecases/update-preference/update-preference.command.ts
rifont, line 1: A bunch of deletions for the old use-cases that only updated V1 Preferences 🧹

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -67,6 +66,7 @@ export class UpdatePreferences {
     return await this.findPreference(command, subscriber);
   }
 
+  @Instrument()
```
rifont, line 63: Adding some more instrumentation to the update preference flow.

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
rifont, line 1: In a follow-up PR (to keep this diff clean), I will rename this usecase from `UpdatePreferences` to `UpsertSubscriberPreferences` as it functions as such now.

apps/api/src/app/inbox/usecases/update-preferences/update-preferences.usecase.ts
```diff
@@ -55,46 +55,16 @@ export class UpdatePreferences {
...
 
-    const userPreference: SubscriberPreferenceEntity | null = await this.subscriberPreferenceRepository.findOne(
-      this.commonQuery(command, subscriber)
-    );
-    if (!userPreference) {
-      await this.createUserPreference(command, subscriber);
-    } else {
-      await this.updateUserPreference(command, subscriber);
-    }
+    await this.updateSubscriberPreference(command, subscriber);
```
rifont, line 58: As said above, we treat this as an upsert now. `GetSubscriberPreferences` always returns a valid preference set now (falling back to default when nothing has been persisted), so there's no need to perform an extra lookup now before persisting the data.

SokratisVidros: Much cleaner!

apps/api/src/app/subscribers/subscribers.controller.ts
```diff
@@ -490,16 +509,29 @@ export class SubscribersController {
...
-    const command = UpdateSubscriberGlobalPreferencesCommand.create({
-      organizationId: user.organizationId,
-      subscriberId,
-      environmentId: user.environmentId,
-      enabled: body.enabled,
-      preferences: body.preferences,
-    });
+  ): Promise<Omit<UpdateSubscriberPreferenceResponseDto, 'template'>> {
+    const channels = body.preferences?.reduce((acc, curr) => {
+      acc[curr.type] = curr.enabled;
```
SokratisVidros, line 514: Is this enabled flag the same as the one we talked about in skipped tests?

SokratisVidros (approved):

apps/api/src/app/subscribers/subscribers.controller.ts
rifont: No, this is the channel-level flag. There are 2 `enabled` flags:
* `enabled` at top level of V1 Preferences, which controls whether to use the subscriber preference during preference merging
* `enabled` flag for each channel, which controls delivery for the channel

PR #6901, opened 2024-11-08
fix(dal): Reduce default Mongo connections

SokratisVidros:
### Screenshots

<details>

</details>


libs/dal/src/dal.service.ts
```diff
@@ -1,16 +1,16 @@
 import mongoose, { Connection, ConnectOptions } from 'mongoose';
 
+export const baseConfig: ConnectOptions = {
+  autoIndex: process.env.MONGO_AUTO_CREATE_INDEXES === 'true',
```
djabarovgeorge: This is a breaking change for the self-hosted users do we want to use AUTO_CREATE_INDEXES as well for couple of version and then deprecated it?

```suggestion
  autoIndex: process.env.AUTO_CREATE_INDEXES === 'true' || process.env.MONGO_AUTO_CREATE_INDEXES === 'true',
```

libs/dal/src/dal.service.ts
```diff
@@ -1,16 +1,16 @@
...
+  maxIdleTimeMS: process.env.MONGO_MAX_IDLE_TIME_IN_MS ? Number(process.env.MONGO_MAX_IDLE_TIME_IN_MS) : 1000 * 60,
+  maxPoolSize: process.env.MONGO_MAX_POOL_SIZE ? Number(process.env.MONGO_MAX_POOL_SIZE) : 50,
```
djabarovgeorge, line 7: The same for the maxPoolSize default and maxIdleTimeMS, it will introduce a breaking change for the self-hosted users, i wonder if we should just change the MONGO_MAX_POOL_SIZE to 50 for us.

SokratisVidros: Great callout! Fixed!

djabarovgeorge (approved):
🚀

libs/dal/src/dal.service.ts
```diff
@@ -1,16 +1,17 @@
 import mongoose, { Connection, ConnectOptions } from 'mongoose';
 
+export const baseConfig: ConnectOptions = {
+  // AUTO_CREATE_INDEXES is deprecated, use MONGO_AUTO_CREATE_INDEXES
+  autoIndex: process.env.AUTO_CREATE_INDEXES === 'true' || process.env.MONGO_AUTO_CREATE_INDEXES === 'true',
+  maxIdleTimeMS: process.env.MONGO_MAX_IDLE_TIME_IN_MS ? Number(process.env.MONGO_MAX_IDLE_TIME_IN_MS) : 1000 * 60,
+  maxPoolSize: process.env.MONGO_MAX_POOL_SIZE ? Number(process.env.MONGO_MAX_POOL_SIZE) : 50,
+  minPoolSize: process.env.MONGO_MIN_POOL_SIZE ? Number(process.env.MONGO_MIN_POOL_SIZE) : 10,
```
rifont: We need to the `MONGO_AUTO_CREATE_INDEXES` and `MONGO_MAX_IDLE_TIME_IN_MS` env vars to the `env.validators.ts` file in API ([example for API](https://github.com/novuhq/novu/blob/next/apps/api/src/config/env.validators.ts)), Worker, Websocket, and Webhook to ensure they are valid values during application startup.

Then we can populate all the defaults in there, avoiding env var defaults in the code.

djabarovgeorge: Good point! :)

SokratisVidros: #til about these validators. Excellent callout! Fixed.

apps/api/src/config/env.validators.ts
```diff
@@ -28,6 +28,7 @@ export const envValidators = {
   MONGO_URL: str(),
   MONGO_MIN_POOL_SIZE: num({ default: 10 }),
   MONGO_MAX_POOL_SIZE: num({ default: 500 }),
+  MONGO_MAX_IDLE_TIME_IN_MS: num({ default: 1000 * 30 }),
```
rifont, line 31:
```suggestion
  AUTO_CREATE_INDEXES: bool({ default: true }),
  MONGO_AUTO_CREATE_INDEXES: bool({ default: true }),
```

apps/worker/src/config/env.validators.ts
```diff
@@ -54,6 +54,7 @@ export const envValidators = {
   MONGO_URL: str(),
   MONGO_MIN_POOL_SIZE: num({ default: 10 }),
   MONGO_MAX_POOL_SIZE: num({ default: 500 }),
+  MONGO_MAX_IDLE_TIME_IN_MS: num({ default: 1000 * 30 }),
```
rifont, line 57: Adding the missing env vars for auto index creation

```suggestion
  MONGO_MAX_IDLE_TIME_IN_MS: num({ default: 1000 * 30 }),
  /** @deprecated - use `MONGO_AUTO_CREATE_INDEXES` instead */
  AUTO_CREATE_INDEXES: bool({ default: true }),
  MONGO_AUTO_CREATE_INDEXES: bool({ default: true }),
```

apps/worker/src/config/env.validators.ts
```diff
@@ -54,6 +54,7 @@ export const envValidators = {
   MONGO_URL: str(),
   MONGO_MIN_POOL_SIZE: num({ default: 10 }),
   MONGO_MAX_POOL_SIZE: num({ default: 500 }),
```
rifont: Replacing with our new lower default

```suggestion
  MONGO_MAX_POOL_SIZE: num({ default: 50 }),
```

apps/api/src/config/env.validators.ts
```diff
@@ -28,6 +28,7 @@ export const envValidators = {
   MONGO_URL: str(),
   MONGO_MIN_POOL_SIZE: num({ default: 10 }),
   MONGO_MAX_POOL_SIZE: num({ default: 500 }),
```
rifont: Replacing with our new lower default

```suggestion
  MONGO_MAX_POOL_SIZE: num({ default: 50 }),
```

libs/dal/src/dal.service.ts
```diff
@@ -1,16 +1,17 @@
 import mongoose, { Connection, ConnectOptions } from 'mongoose';
 
+export const baseConfig: ConnectOptions = {
+  // AUTO_CREATE_INDEXES is deprecated, use MONGO_AUTO_CREATE_INDEXES
+  autoIndex: process.env.AUTO_CREATE_INDEXES === 'true' || process.env.MONGO_AUTO_CREATE_INDEXES === 'true',
+  maxIdleTimeMS: process.env.MONGO_MAX_IDLE_TIME_IN_MS ? Number(process.env.MONGO_MAX_IDLE_TIME_IN_MS) : 1000 * 30,
+  maxPoolSize: process.env.MONGO_MAX_POOL_SIZE ? Number(process.env.MONGO_MAX_POOL_SIZE) : 50,
+  minPoolSize: process.env.MONGO_MIN_POOL_SIZE ? Number(process.env.MONGO_MIN_POOL_SIZE) : 10,
```
rifont, line 8: No need for any defaults here now
```suggestion
  autoIndex: Boolean(process.env.AUTO_CREATE_INDEXES) === true || Boolean(process.env.MONGO_AUTO_CREATE_INDEXES) === true,
  maxIdleTimeMS: Number(process.env.MONGO_MAX_IDLE_TIME_IN_MS),
  maxPoolSize: Number(process.env.MONGO_MAX_POOL_SIZE),
  minPoolSize: Number(process.env.MONGO_MIN_POOL_SIZE),
```

rifont: We need to add all of these env vars to WS + Webhook `env.validators.ts` to ensure the defaults populate

SokratisVidros: I'd keep it as is for now. The [env validator comes after imports](https://github.com/novuhq/novu/blob/fcc12b04b3eb478586f32cfdb88ddbc46e791d53/apps/worker/src/bootstrap.ts#L19). It's not the first thing that runs in the code so we might have some unwanted side-effects.

The truth is that we need only one place to set defaults but I want to avoid touching the env validator as part of this PR.

rifont: Sounds good 👍

rifont (approved):
💯

PR #6904, opened 2024-11-08
fix(api):  bug bash preview issues resolved

tatarco:
During the bug bash the preview endpoint appeared to be very fragile, attempt to make it more robust was difficult due to the structure of the use cases, multiple linear bugs resolved + a massive cleanup and alignment to consistent company standards 
what was done? 

1. Pipe {{ttt | upper}} not converting to payload correctly.
2.. {{}} failing preview 
3.. Variable validation agains Available Variable schema (validate {{name}} illegal.
4.. Control values sanitation on preview ( illegal placeholders will be removed for preview but persisted along with errors in the DB to keep the UX consistent 
5.. computed Payload schema is now  persisted (as string ) on upsert and served from the DB on get test data 
 

### Screenshots

<details>

</details>


.source
djabarovgeorge, line 1: If we don't have changes in ee in this PR we should not have change here.

.cspell.json
```diff
@@ -789,6 +789,7 @@
     "packages/shared/src/types/timezones/timezones.types.ts",
     "*.riv",
     "websockets",
+    "apps/api/src/app/workflows-v2/usecases/validate-content/validate-placeholders/validate-placeholder.usecase.ts",
```
djabarovgeorge, line 792: can you add `/* cspell:disable */` on this file, i would not put here files that can be moved in the source code in the future.

apps/api/src/app/environments-v1/usecases/output-renderers/expand-email-editor-schema.usecase.ts
```diff
@@ -9,16 +9,19 @@ export class ExpandEmailEditorSchemaUsecase {
   constructor(private hydrateEmailSchemaUseCase: HydrateEmailSchemaUseCase) {}
 
   execute(command: ExpandEmailEditorSchemaCommand): TipTapNode {
+    console.log('PRE:hydratedEmailSchema\n\n', JSON.stringify(command.emailEditorJsonAsString));
```
djabarovgeorge: Did we decide to start console log actions?

```suggestion
```

apps/api/src/app/environments-v1/usecases/output-renderers/expand-email-editor-schema.usecase.ts
```diff
@@ -9,16 +9,19 @@ export class ExpandEmailEditorSchemaUsecase {
...
 
     return emailSchemaHydrated;
   }
   private hydrate(command: ExpandEmailEditorSchemaCommand) {
     const { hydratedEmailSchema } = this.hydrateEmailSchemaUseCase.execute({
-      emailEditor: command.body,
+      emailEditor: command.emailEditorJsonAsString,
       fullPayloadForRender: command.fullPayloadForRender,
     });
+    console.log('hydratedEmailSchema\n\n', JSON.stringify(hydratedEmailSchema));
```
djabarovgeorge: Same here.

```suggestion
```

apps/api/src/app/environments-v1/usecases/output-renderers/expand-email-editor-schema-command.ts
```diff
@@ -4,6 +4,6 @@ import { BaseCommand } from '@novu/application-generic';
 import { FullPayloadForRender } from './render-command';
 
 export class ExpandEmailEditorSchemaCommand extends BaseCommand {
-  body: string;
+  emailEditorJsonAsString: string;
```
djabarovgeorge: i think the type is enough to tell us it is a string, no? 

```suggestion
  emailEditorJson: string;
```

apps/api/src/app/environments-v1/usecases/output-renderers/expand-email-editor-schema.usecase.ts
tatarco: no no, left from the debugging

SokratisVidros (approved):

PR #6912, opened 2024-11-10
fix(framework): Ensure missing schemas return unknown record type

rifont:

* Ensure missing schemas return unknown record type
  * Previously due to not having a minimal Zod schema, lack of being able to access the `zod.ZodType` import meant that the inferred Zod type became `any`, and in turn the inferred `FromSchema` type became any. A minimal Zod schema included in the Framework types now ensures that Zod can successfully be skipped during type inference when the import is unavailable
* Refactor schema types
  * move type tests into the Workflow resource type tests with better unit test encapsulation
    * simplifies addition of new validators (see [Class Validators PR](https://github.com/novuhq/novu/pull/6840/files) for example)
  * schema integration tests for validation now focus only on testing the `validateData` and `transformData` functions
* Fix provider discovery to be completely asynchronous
  * This surfaced in an existing test for providers code detection

### Screenshots
_Missing `controlSchema` results in `Record<string, unknown>` type for controls, payload, and results_
image

_Missing Zod dependencies reported at runtime_
```bash
> next start "--port" "4000"

   ▲ Next.js 15.0.3-canary.9
   - Local:        http://localhost:4000

 ✓ Starting...
 ✓ Ready in 208ms
 ⨯ Error: Tried to use a zod schema in @novu/framework without zod-to-json-schema installed. Please install it by running `npm install zod-to-json-schema`.
    at st (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:77627)
    at async si.canHandle (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:78497)
    at async sd (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:81223)
    at async Object.email (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:7404)
    at async workflows.payloadSchema (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:11395)
    at async Object.discover (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:10746)
    at async su.addWorkflows (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:501:8314)
    at async (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:3508)
    at async te.do (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:17826) {
  data: [Object],
  statusCode: 500,
  code: 'MissingDependencyError'
}
```

Missing dependencies are determined dynamically (preview with `class-validators`, missing only `class-validator-jsonschema` )
```bash
> next start "--port" "4000"

   ▲ Next.js 15.0.3-canary.9
   - Local:        http://localhost:4000

 ✓ Starting...
 ✓ Ready in 211ms
 ⨯ Error: Tried to use a class-validator schema in @novu/framework without class-validator-jsonschema installed. Please install it by running `npm install class-validator-jsonschema`.
    at se (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:77618)
    at async si.canHandle (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:79403)
    at async sp (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:411:81104)
    at async Object.discover (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:10531)
    at async sl.addWorkflows (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:501:8314)
    at async (/Users/rifont/git/framework-next-turbo/.next/server/app/api/novu/route.js:505:3508)
    at async te.do (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:17826)
    at async te.handle (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/compiled/next-server/app-route.runtime.prod.js:18:22492)
    at async doRender (/Users/rifont/git/framework-next-turbo/node_modules/.pnpm/next@15.0.3-canary.9_react-dom@19.0.0-rc-02c0e824-20241028_react@19.0.0-rc-02c0e824-20241028__utrvdij6okqvk6cvj3mqcl5cuy/node_modules/next/dist/server/base-server.js:1455:42) {
  statusCode: 500,
  code: 'MissingDependencyError',
  data: [Array]
}
```

<details>

</details>


packages/framework/src/client.validation.test.ts
rifont, line 1: This test suite was destined to grow significantly larger as more validators were added. It's now significantly reduced in size to focus only on Json Schema as the de factor integration testing schema. The Zod schema runtime assertions are now captured by integration tests for `validateData` and `transformSchema`, and the type assertions are now captured with `{json,zod}.schema.type.test-d.ts` [Vitest type tests](https://vitest.dev/api/expect-typeof) (which provide an awesome type testing API! 🎉)

packages/framework/src/client.ts
```diff
@@ -195,7 +195,7 @@ export class Client {
           throw new Error(`Invalid component: '${component}'`);
       }
     } else {
-      return result.data;
+      return result.data as T;
```
rifont, line 225: `result.data` was incorrectly inferred as `any` due to lack of schema context, this is fixed as we now need to coerce to the validated type within a generic schema context.

packages/framework/src/constants/error.constants.ts
```diff
@@ -11,6 +11,7 @@ export enum ErrorCodeEnum {
   EXECUTION_STATE_RESULT_INVALID_ERROR = 'ExecutionStateResultInvalidError',
   INVALID_ACTION_ERROR = 'InvalidActionError',
   METHOD_NOT_ALLOWED_ERROR = 'MethodNotAllowedError',
+  MISSING_DEPENDENCY_ERROR = 'MissingDependencyError',
```
rifont, line 14: New error code to report missing dependencies at runtime. We can refer to this code in the docs.

packages/framework/src/errors/import.errors.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  constructor(usageReason: string, missingDependencies: string[]) {
+    const pronoun = missingDependencies.length === 1 ? 'it' : 'them';
+    super(
+      `Tried to use a ${usageReason} in @novu/framework without ${missingDependencies.join(
+        ', '
+      )} installed. Please install ${pronoun} by running \`npm install ${missingDependencies.join(' ')}\`.`
+    );
+  }
+}
```
rifont, line 16: We now have a generic error for all missing dependency errors. This covers only validators right now in the implementation, but can be extended to other plugin domains.

packages/framework/src/resources/workflow/discover-action-step-factory.ts
```diff
@@ -32,7 +32,7 @@ export async function discoverActionStepFactory(
       },
       resolve: resolve as (controls: Record<string, unknown>) => Awaitable<Record<string, unknown>>,
       code: resolve.toString(),
-      options,
+      options: options as StepOptions<FromSchema<Schema>, FromSchema<Schema>>,
```
rifont: Similar to the above validation coercion, we need a similar approach here for a generic schema context in the discovery methods.

packages/framework/src/resources/workflow/discover-channel-step-factory.ts
```diff
@@ -32,14 +40,14 @@ export async function discoverChannelStepFactory(
...
-      options,
+      options: options as StepOptions<FromSchema<Schema>, FromSchema<Schema>>,
       providers: [],
     };
 
     await discoverStep(targetWorkflow, stepId, step);
 
     if (Object.keys(options.providers || {}).length > 0) {
-      discoverProviders(step, type as ChannelStepEnum, options.providers || {});
+      await discoverProviders(step, type as ChannelStepEnum, options.providers || {});
```
rifont, line 50: This `await` for provider discovery was missing before and could result in providers not being discovered inside the async event loop.

packages/framework/src/resources/workflow/discover-providers.ts
```diff
@@ -20,19 +20,22 @@ export async function discoverProviders(
...
       resolve,
       outputs: {
         schema: await transformSchema(schemas.output),
         unknownSchema: schemas.output,
       },
-    });
+    };
   });
+
+  step.providers.push(...(await Promise.all(providerPromises)));
```
rifont, line 40: Using a list of promises is preferable than `for ... await ... of` syntax as we can parallelize the operations with better performance, whereas the latter is sequential.

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
rifont, line 1: Bringing all the schema type assertions into a separate file, using the brilliant [Vitest type test API](https://vitest.dev/api/expect-typeof)!

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
```diff
@@ -0,0 +1,281 @@
...
+          email: unknown;
+          sms: unknown;
+          push: unknown;
+          chat: unknown;
+          inApp: unknown;
+          digest: unknown;
+          delay: unknown;
+          custom: unknown;
+        }>();
+      });
```
rifont, line 25: Adding some new basic tests for the workflow execution parameters.

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
```diff
@@ -0,0 +1,281 @@
...
+        await step.email('without-schema', async (controls) => {
+          expectTypeOf(controls).toEqualTypeOf<Record<string, unknown>>();
+
+          return {
+            subject: 'Test subject',
+            body: 'Test body',
+          };
+        });
+      });
+    });
```
rifont: New test that asserts a missing control schema should result in an unknown record controls data type. Similar tests follow for other parts of the `workflow` API with schema usage.

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
```diff
@@ -0,0 +1,281 @@
...
+            foo: 'bar',
+          };
+        });
+
+        expectTypeOf(result).toMatchTypeOf<Record<string, unknown>>();
+      });
+    });
+  });
+
+  describe('json-schema', () => {
```
rifont, line 101: Json schema is used as the baseline schema for asserting on inferred schema usage within the `workflow` API. This provides an adequate level of integration testing as other and future schemas are tested only with the `validateData`, `transformSchema`, `FromSchema`, and `FromSchemaUnvalidated` APIs, which as an integration point for schema validation and inference, respectively.

packages/framework/src/types/import.types.ts
```diff
@@ -0,0 +1,34 @@
+export type ImportRequirement = {
```
rifont, line 1: Used by the `checkDependencies` util and in places that need to list their peer dependencies for availability checks, such as the validators.

packages/framework/src/types/schema.types.ts
rifont, line 1: Moved to the new `schema.types` directory for better organization.

packages/framework/src/types/schema.types/base.schema.types.test-d.ts
rifont, line 1: Integration tests for `FromSchema` and `FromSchemaUnvalidated` type inference APIs

packages/framework/src/types/schema.types/base.schema.types.ts
```diff
@@ -0,0 +1,53 @@
...
+  | InferJsonSchema<T, Options>
+  | InferZodSchema<T, Options>
+  | never extends infer U
+  ? // If all inferred types are `never`, return an unknown record
+    [U] extends [never]
+    ? Record<string, unknown>
+    : U extends Record<string, unknown>
+      ? U
+      : Record<string, unknown>
+  : Record<string, unknown>;
```
rifont: Schema inference is simplified to a single entry point with a generic option of `{ validated: boolean }` - this new Option is passed to specific schema library inference APIs who can provide determine how to handle the options independently.

This effectively makes type inference more encapsulated per schema library to improve maintainability, whilst reducing schema inference complexity between validated/unvalidated types in the entry point to simplify the addition of new schema libraries.

packages/framework/src/types/schema.types/json.schema.types.test-d.ts
rifont, line 1: Each schema library now implements a much simpler set of unit tests for their types, leaning on Vitest type testing APIs for a lot of heavy lifting.

packages/framework/src/types/schema.types/zod.schema.types.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * A minimal ZodSchema type.
+ *
+ * It is necessary to define a minimal ZodSchema type to enable correct inference
+ * when Zod is not available, as Typescript doesn't support detection of module
+ * availability via `typeof import('zod')`.
+ */
+export type ZodSchemaMinimal = {
+  readonly safeParseAsync: unknown;
+};
```
rifont, line 17: This is the new minimal zod schema for use when Zod is not available in `node_modules`. It ensures we can defer to a `Record<string, unknown>` type after exhausting all schema possibilities.

packages/framework/src/types/util.types.ts
```diff
@@ -1,3 +1,8 @@
+/*
+ * THIS FILE SHOULD NOT DEPEND ON ANY OTHER FILES.
+ * IT SHOULD ONLY CONTAIN UTILITY TYPES.
+ */
+
```
rifont, line 5: Just a simple warning message to ensure we don't create any circular deps here!  😀

packages/framework/src/types/validator.types.ts
```diff
@@ -19,15 +21,15 @@ export type ValidateResult<T> =
...
+export type Validator<T_Schema extends Schema = Schema> = {
   validate: <
     T_Unvalidated extends Record<string, unknown> = FromSchemaUnvalidated<T_Schema>,
     T_Validated extends Record<string, unknown> = FromSchema<T_Schema>,
   >(
     data: T_Unvalidated,
     schema: T_Schema
   ) => Promise<ValidateResult<T_Validated>>;
-  canHandle: (schema: Schema) => schema is T_Schema;
+  canHandle: (schema: Schema) => Promise<boolean>;
```
rifont, line 32: We make it a promise so that the `canHandle` method provides a single entry point for checking dependency availability for both `validateData` and `transformSchema`, which needs a Promise type because of `await import()` asynchronosity.

We have to forgo the type guard here unfortunately, because Typescript doesn't support async type guards yet :/

packages/framework/src/types/validator.types.ts
```diff
@@ -19,15 +21,15 @@ export type ValidateResult<T> =
...
   transformToJsonSchema: (schema: T_Schema) => Promise<JsonSchema>;
-}
+  requiredImports: readonly ImportRequirement[];
```
rifont, line 34: Each validator now declares a list of required imports that can be checked concretely at runtime.

I've found this to be the most reliable approach as different build + runtiem environments (CJS/ESM + `next dev`/`next dev --turbo` can handle import results differently - sometimes they throw during `await import(...)`, other times they resolve with empty exports. We handle both failure cases with `checkDependencies` now and we can dynamically construct the list of missing imports, providing an excellent `npm install {x, y, z}` DX.

packages/framework/src/validators/base.validator.ts
```diff
@@ -14,20 +21,31 @@ export const validateData = async <
...
-  if (zodValidator.canHandle(schema)) {
-    return zodValidator.validate(data, schema);
-  } else if (jsonSchemaValidator.canHandle(schema)) {
-    return jsonSchemaValidator.validate(data, schema);
+  /**
+   * TODO: Replace type coercion with async type guard when available.
+   *
+   * @see https://github.com/microsoft/typescript/issues/37681
+   */
+  if (await zodValidator.canHandle(schema)) {
```
rifont, line 29: `canValidated` now returns a `Promise` to enable the `checkDependencies` usage of `await import(...)`

packages/shared/package.json
```diff
@@ -38,9 +38,6 @@
     }
   },
   "devDependencies": {
-    "@types/bluebird": "^3.5.24",
-    "@types/jest": "29.5.2",
-    "json-schema-to-ts": "^3.0.0",
```
rifont: Not used 🧹

apps/dashboard/package.json
```diff
@@ -7,6 +7,8 @@
     "start": "vite",
     "start:static:build": "pm2 start proxy-server.js",
     "dev": "pnpm start",
+    "check:types": "tsc --noEmit",
+    "build:new": "tsc -b && vite build",
```
rifont:
```suggestion
```

packages/framework/src/errors/base.errors.ts
```diff
@@ -67,7 +67,7 @@ export abstract class ServerError extends FrameworkError {
         stack: cause.stack ?? message,
       };
     } else {
-      super(`${message}: ${JSON.stringify(cause, null, 2)}`);
+      super(`${message}${cause ? `: ${JSON.stringify(cause, null, 2)}` : ''}`);
```
SokratisVidros, line 70: I'd suggest we replace this dynamic message with a new key on the data attribute to easy error grouping in host applications.

packages/framework/src/errors/import.errors.ts
SokratisVidros: Perfect. We need to add this to the docs mentioning what to do in this error and how to use Zod or Class Validator.

packages/framework/src/resources/workflow/discover-providers.ts
SokratisVidros: Of course!

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
```diff
@@ -0,0 +1,281 @@
...
+
+          return {
+            subject: 'Test subject',
+            body: 'Test body',
+          };
+        });
+      });
+    });
+
+    it('should infer an unknown record type in the skip function step controls', async () => {
```
SokratisVidros: Nit-picky comment: I'd consider merging this test about skip and the following about providers with the previous one. They all belong under the control type inference umbrella as an argument in different functions of the framewok.

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
```diff
@@ -0,0 +1,281 @@
...
+              body: 'Test body',
+            };
+          },
+          {
+            controlSchema: jsonSchema,
+          }
+        );
+      });
+
+      it('should infer correct types in the skip function step controls', async () => {
```
SokratisVidros: Nit-picky: Same approach about merging tests.

packages/framework/src/types/import.types.ts
SokratisVidros: Nit-picky: If I remember correctly, `checkDependencies` is an internal API. I am under the impression that src/types contain public-facing type exports. So we can move the `ImportRequirement` type declaration next to the `checkDependencies` function declaration.

packages/framework/src/types/schema.types/base.schema.types.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param T - The Schema to infer the type of.
+ * @param Options - Configuration options for the type inference. The `validated` flag determines whether the schema has been validated. If `validated` is true, all properties are required unless specified otherwise. If false, properties with default values are optional.
+ */
+type InferSchema<T extends Schema, Options extends { validated: boolean }> =
+  | InferJsonSchema<T, Options>
+  | InferZodSchema<T, Options>
+  | never extends infer U
+  ? // If all inferred types are `never`, return an unknown record
+    [U] extends [never]
```
SokratisVidros: TIL about this syntax to check all types!

packages/framework/src/types/schema.types/base.schema.types.ts
```diff
@@ -0,0 +1,53 @@
...
+    ? Record<string, unknown>
+    : U extends Record<string, unknown>
```
SokratisVidros: I am curious if simplifying a bit the last part of the ternary has the same effect in generating the right Record<string unknown/>

```? U extends Record<string, unknown> ? U : Record<string, unknown>```

**Review by SokratisVidros — APPROVED** [review]

File: packages/framework/src/validators/zod.validator.ts [hunk 24]
```diff
@@ -28,20 +54,9 @@ export class ZodValidator implements Validator<ZodSchema> {
   }
 
   async transformToJsonSchema(schema: ZodSchema): Promise<JsonSchema> {
-    try {
-      const { zodToJsonSchema } = await import('zod-to-json-schema');
+    const { zodToJsonSchema } = await import('zod-to-json-schema');
```
SokratisVidros, line 57: Much cleaner!

packages/framework/src/validators/zod.validator.ts
```diff
@@ -1,19 +1,45 @@
...
+import { checkDependencies } from '../utils/import.utils';
+import { ImportRequirement } from '../types/import.types';
 
 export class ZodValidator implements Validator<ZodSchema> {
-  canHandle(schema: Schema): schema is ZodSchema {
-    return (schema as ZodSchema).safeParseAsync !== undefined;
+  readonly requiredImports: readonly ImportRequirement[] = [
+    {
+      name: 'zod',
+      import: import('zod'),
```
SokratisVidros, line 17: Is it better to pass the name in this list and let all the async import handling happen in `checkDependencies`? I see two marginal benefits with that approach:

1. Simpler `ImportRequirement` interface.
2. Handle asynchronicity in one place.

packages/framework/src/errors/import.errors.ts
rifont: 100%. I will address documentation for this case when writing docs for Class Validators.

packages/framework/src/validators/zod.validator.ts
rifont: Unfortunately this approach doesn't work, see [this comment](https://github.com/novuhq/novu/blob/7e314123d5faaf819f5762387cae793f7b129b18/packages/framework/src/types/import.types.ts#L15:L16) in the code. An explicit `import(...)` statement is needed to ensure the bundler makes the dependency available for import.

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
rifont: Fair point, we had it this way earlier but I changed it to have a single expect statement per test. It's less readable/maintainable this way though, so I'm going with your suggestion.

packages/framework/src/resources/workflow/workflow.resource.test-d.ts
rifont: I have merged the tests.

packages/framework/src/types/import.types.ts
rifont: We [declare all the public-facing exports explicitly](https://github.com/novuhq/novu/blob/next/packages/framework/src/index.ts), so this shouldn't be a problem.

Here is a screen showing that the `ImportRequirement` type isn't available publicly:
image

rifont: As a side note for completeness, it's preferable to keep all the types located in the `types` directory and use explicit exports to make specific APIs available to keep the folder structure well organized.

SokratisVidros: We are aligned!

packages/framework/src/types/schema.types/base.schema.types.ts
rifont: You're absolutely right, primitive type schemas were currently permitted. After restricting to record only schemas, this we can significantly simplify the base schema entry point.

packages/framework/src/errors/base.errors.ts
rifont: It's a fair point. I think this part needs a bit of rework, especially with regards to logging nice, actionable error messages in the console. For now, the `cause.stack` is added as an error attribute and shown in the Local Studio.

I suggest we debate this approach further outside the scope of this PR as part of better logging.

PR #6913, opened 2024-11-11
fix(framework): Prevent adding duplicate workflows

rifont:
* Add locking mechanism to prevent adding duplicate workflows in `Client.addWorkflows`
* Convert `discoveredWorkflows` internal `Client` attribute to a `Map` for more performant lookups of existing workflows

### Screenshots
_Before - duplicate workflows could be added if more than 1 Local Studio tab is open in Browser during hot reload_
image

_After - duplicate workflows are not added_
image

<details>

</details>

### DIFF
--- a/packages/framework/src/client.ts
+++ b/packages/framework/src/client.ts
@@ -54,8 +54,7 @@ function isRuntimeInDevelopment() {
 
 export class Client {
   private discoveredWorkflows = new Map<string, DiscoverWorkflowOutput>();
-  private addingWorkflows = new Set<string>();
-  private workflowPromises = new Map<string, Promise<void>>();
+  private discoverWorkflowPromises = new Map<string, Promise<void>>();
 
   private templateEngine = new Liquid({
     outputEscape: (output) => {
@@ -110,16 +109,15 @@ export class Client {
         continue;
       }
 
-      if (this.addingWorkflows.has(workflow.id)) {
+      const existingPromise = this.discoverWorkflowPromises.get(workflow.id);
+      if (existingPromise) {
         // Wait for the existing promise to resolve if the workflow is already being added
-        await this.workflowPromises.get(workflow.id);
+        await existingPromise;
         continue;
       }
 
-      this.addingWorkflows.add(workflow.id);
-
       const workflowPromise = this.addWorkflow(workflow);
-      this.workflowPromises.set(workflow.id, workflowPromise);
+      this.discoverWorkflowPromises.set(workflow.id, workflowPromise);
 
       await workflowPromise;
     }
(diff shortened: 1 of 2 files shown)


packages/framework/src/client.ts
```diff
@@ -95,21 +97,49 @@ export class Client {
...
-        return;
-      } else {
-        const definition = await workflow.discover();
-        prettyPrintDiscovery(definition);
-        this.discoveredWorkflows.push(definition);
+      if (this.discoveredWorkflows.has(workflow.id)) {
+        continue;
+      }
+
+      if (this.addingWorkflows.has(workflow.id)) {
```
SokratisVidros: Since the workflow promises Set is using the workflowId, do we need the addingWorkflows set? Can we use the map of promises as an indication of a workflow discovery already being in progress?

packages/framework/src/client.ts
```diff
@@ -95,21 +97,49 @@ export class Client {
...
+      const workflowPromise = this.addWorkflow(workflow);
+      this.workflowPromises.set(workflow.id, workflowPromise);
+
+      await workflowPromise;
+    }
+  }
+
+  private async addWorkflow(workflow: Workflow): Promise<void> {
+    try {
+      const definition = await workflow.discover();
```
SokratisVidros, line 128: Good catch! With an async discover and transform this could become more prominent.

packages/framework/src/client.ts
```diff
@@ -95,21 +97,49 @@ export class Client {
...
+      this.workflowPromises.delete(workflow.id);
     }
   }
 
   public healthCheck(): HealthCheck {
-    const workflowCount = this.discoveredWorkflows.length;
-    const stepCount = this.discoveredWorkflows.reduce((acc, workflow) => acc + workflow.steps.length, 0);
+    const discoveredWorkflows = this.getRegisteredWorkflows();
+    const workflowCount = discoveredWorkflows.length;
+    const stepCount = discoveredWorkflows.reduce((acc, workflow) => acc + workflow.steps.length, 0);
```
SokratisVidros, line 139: I wonder why we report the total stepCount across workflows. It doesn't feel very useful.

SokratisVidros (approved):

packages/framework/src/client.ts
rifont: The main intent at the time was to support simple calculations such as averages of steps per workflow in Local Studio telemetry.

packages/framework/src/client.ts
rifont: Yes, you are right! I have removed the simple string map and we lean on the promises only now.

packages/framework/src/client.ts
SokratisVidros: You can leave it for now. I was just curious.

PR #6933, opened 2024-11-11
fix(dashboard): promote toast env name

ChmaraX:

Ties env name in sync toast to the actual workflow that was synced rather than the current environment.

This fixes an issue where:
1. wf was synced to e.g. prod
2. toast showed "synced to prod"
3. user switched to dev
4. toast changed labels to "dev"

### DIFF
--- a/apps/dashboard/src/components/promote-workflow/promote-success-toast.tsx
+++ b/apps/dashboard/src/components/promote-workflow/promote-success-toast.tsx
@@ -1,20 +1,21 @@
 import { Button } from '@/components/primitives/button';
 import { ToastClose, ToastIcon } from '@/components/primitives/sonner';
 import { useEnvironment } from '@/context/environment/hooks';
-import { WorkflowListResponseDto } from '@novu/shared';
+import { IEnvironment, WorkflowResponseDto } from '@novu/shared';
 import { RiArrowRightSLine } from 'react-icons/ri';
 
 interface PromoteSuccessToastProps {
-  workflow: WorkflowListResponseDto;
+  workflow: WorkflowResponseDto;
+  environment?: IEnvironment;
   onClose: () => void;
 }
 
-export function PromoteSuccessToast({ workflow, onClose }: PromoteSuccessToastProps) {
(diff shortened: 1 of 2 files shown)


apps/dashboard/src/hooks/use-sync-workflow.tsx
```diff
@@ -4,7 +4,13 @@ import { ConfirmationModal } from '@/components/confirmation-modal';
 import { showToast } from '@/components/primitives/sonner-helpers';
 import { PromoteSuccessToast } from '@/components/promote-workflow/promote-success-toast';
 import { useEnvironment } from '@/context/environment/hooks';
-import { WorkflowListResponseDto, WorkflowOriginEnum, WorkflowResponseDto, WorkflowStatusEnum } from '@novu/shared';
+import {
```
SokratisVidros:
```suggestion
import type {
```

apps/dashboard/src/components/promote-workflow/promote-success-toast.tsx
```diff
@@ -1,35 +1,36 @@
 import { Button } from '@/components/primitives/button';
 import { ToastClose, ToastIcon } from '@/components/primitives/sonner';
 import { useEnvironment } from '@/context/environment/hooks';
-import { WorkflowListResponseDto } from '@novu/shared';
+import { IEnvironment, WorkflowResponseDto } from '@novu/shared';
```
SokratisVidros, line 4:
```suggestion
import type { IEnvironment, WorkflowResponseDto } from '@novu/shared';
```

SokratisVidros (approved):
Please import types from shared to avoid overloading the dashboard bundle. We haven't been strict about that in the Dashboard code but we should moving forward. 

Adding a custom ES lint rule at some point should help.

apps/dashboard/src/hooks/use-sync-workflow.tsx
ChmaraX: Will do that in bulk in separate PR + will try to add the rule. Now its throwing some errors when I do that.

PR #6967, opened 2024-11-12
feat(dashboard): styling for read-only input fields

ChmaraX:

Before:

image

After:

image

### DIFF
--- a/apps/dashboard/src/components/primitives/input.tsx
+++ b/apps/dashboard/src/components/primitives/input.tsx
@@ -5,7 +5,37 @@ import { cva, VariantProps } from 'class-variance-authority';
 import { inputVariants } from '@/components/primitives/variants';
 
 const inputFieldVariants = cva(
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/components/primitives/input.tsx
```diff
@@ -5,7 +5,37 @@ import { cva, VariantProps } from 'class-variance-authority';
...
+
+    // Empty value state
+    'has-[input[value=""]]:text-foreground-400',
+
+    // Read-only state
+    'has-[input:read-only]:text-foreground-700',
+    'has-[input:read-only]:bg-neutral-alpha-100',
+    'has-[input:read-only]:opacity-70',
+    'has-[input:read-only]:border-neutral-alpha-200',
+  ].join(' '),
```
ChmaraX, line 38: This format is more readable, not sure if it can have some unintended consequences by not being static but using join

LetItRock: are these styles agreed with Naveen? because I don't see the design system updated

ChmaraX: Just [confirmed with Naveen](https://linear.app/novu/issue/NV-4637/apply-styling-fixes-to-workflow-identifier-field#comment-9de33e29), should be good.

LetItRock (approved):

PR #6969, opened 2024-11-12
refactor(api): remove $or query for hash on env

scopsy:

- Removes the legacy $or query to environments when resolving API keys
- Only use the hash key on the environment for matching

This should help with some of the ee API key transaction performance

### Screenshots

https://github.com/novuhq/packages-enterprise/pull/243
<details>

### Related enterprise PR
https://github.com/novuhq/packages-enterprise/pull/243

</details>

### DIFF
--- a/libs/application-generic/src/services/auth/community.auth.service.ts
+++ b/libs/application-generic/src/services/auth/community.auth.service.ts
@@ -364,7 +364,6 @@ export class CommunityAuthService implements IAuthService {
     const hashedApiKey = createHash('sha256').update(apiKey).digest('hex');
 
     const environment = await this.environmentRepository.findByApiKey({
-      key: apiKey,
       hash: hashedApiKey,
     });
 
@@ -373,16 +372,7 @@ export class CommunityAuthService implements IAuthService {
       return { error: 'API Key not found' };
     }
 
-    let key = environment.apiKeys.find((i) => i.hash === hashedApiKey);
-
-    if (!key) {
-      /*
-       * backward compatibility - delete after encrypt-api-keys-migration execution
-       * find by decrypted key if key not found, because of backward compatibility
-       * use-case: findByApiKey found by decrypted key, so we need to validate by decrypted key
-       */
-      key = environment.apiKeys.find((i) => i.key === apiKey);
-    }
+    const key = environment.apiKeys.find((i) => i.hash === hashedApiKey);
 
     if (!key) {
       return { error: 'API Key not found' };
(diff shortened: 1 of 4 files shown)


SokratisVidros (changes requested):
Perfect. 

Can you please also remove the caching from the following places?

With the right index, we don't need caching for this query. We must always go to Mongo.

- https://github.com/novuhq/novu/blob/next/enterprise/packages/auth/src/usecases/ee.auth.service.ts#L154
- https://github.com/novuhq/novu/blob/a305ec690cbdd7b89e636379cbfd36657d55a662/libs/application-generic/src/services/auth/community.auth.service.ts#L359

SokratisVidros:
@scopsy We need to add the index to the Mongoose environment schema.

scopsy:
@SokratisVidros why removing the caching for this? This will create a lot of pressure on MongoDB for information that rarely changes.

I'm very much concerned with the negative effect of removing the caching here at scale without properly monitoring it with a Feature flag

SokratisVidros:
> @SokratisVidros why removing the caching for this? This will create a lot of pressure on MongoDB for information that rarely changes.
> 
> I'm very much concerned with the negative effect of removing the caching here at scale without properly monitoring it with a Feature flag

It won't stress Mongo. Environments don't grow extremely fast; the index is super small in size, and it will always be in Mongo memory, making it work exactly like the Elastic Cache. 

It also removes the need to invalidate this cache in other places of the codebase.

scopsy:
@SokratisVidros i'm ok for trying this, but we must look on the metrics post deployment to look for any negative effects and roll back immediately.

SokratisVidros:
> @SokratisVidros i'm ok for trying this, but we must look on the metrics post deployment to look for any negative effects and roll back immediately.

Of course.

SokratisVidros (approved):
Reminder for https://github.com/novuhq/novu/blob/a305ec690cbdd7b89e636379cbfd36657d55a662/libs/application-generic/src/services/auth/community.auth.service.ts#L353

Other than that it looks great!

libs/application-generic/src/services/auth/community.auth.service.ts
```diff
@@ -350,12 +350,6 @@ export class CommunityAuthService implements IAuthService {
     );
   }
 
-  @CachedEntity({
-    builder: ({ apiKey }: { apiKey: string }) =>
-      buildAuthServiceKey({
```
rifont, line 355: ☑️ **chore:** There's a few more occurrences of `buildAuthServiceKey` (including the cache invalidation for the key) that we need to remove from the code - [ref](https://github.com/search?q=repo%3Anovuhq%2Fnovu%20buildAuthServiceKey&type=code)‏

rifont: hey @scopsy , I see the PR has been merged, but the above `buildAuthServiceKey` references are still present in `next`. Is there a plan to remove those? They are currently dead code and we are making a cache invalidation call that does nothing.

PR #6970, opened 2024-11-12
fix(api): upsert preference if exist

djabarovgeorge:

We are currently throwing an exception if the old preferences do not exist. This is a bug, as we should not enforce the use of preferences.
image

image

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
@@ -165,7 +165,7 @@ export class UpsertWorkflowUseCase {
 
   private async upsertUserWorkflowPreferences(workflow: NotificationTemplateEntity, command: UpsertWorkflowCommand) {
     let preferences: WorkflowPreferences | null;
-    if (command.workflowDto.preferences?.user !== undefined) {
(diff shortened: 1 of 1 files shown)


apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -183,12 +183,16 @@ export class UpsertWorkflowUseCase {
   }
 
   private async upsertWorkflowPreferences(workflow: NotificationTemplateEntity, command: UpsertWorkflowCommand) {
+    if (!command.workflowDto.preferences?.workflow) {
```
SokratisVidros, line 186: This DTO looks wider than it should. The comment is not actionable as part of this PR, but we must make it wider rather than deeper. Let's treat it as a code smell.

Reading this code, the hypothesis is that this check ensures that the workflow has preferences, right?

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: correct, we ensure that we have workflow preferences to upsert.

PR #6972, opened 2024-11-12
fix(dashboard): Workflow saving status on toast and multiple toasts

desiprisg:
### Screenshots

https://github.com/user-attachments/assets/3c207bb2-d988-46f3-8501-fa95762dcd89

<details>

</details>


SokratisVidros (approved):

SokratisVidros (approved):

apps/dashboard/src/hooks/use-update-workflow.ts
```diff
@@ -1,22 +1,17 @@
...
-  onSuccess,
-  onError,
-}: { onSuccess?: (data: WorkflowResponseDto) => void; onError?: (error: unknown) => void } = {}) => {
-  const { mutateAsync, isPending, error, data } = useMutation({
-    mutationFn: async ({ id, workflow }: { id: string; workflow: UpdateWorkflowDto }) =>
-      updateWorkflow({ id, workflow }),
-    onSuccess,
-    onError,
+export const useUpdateWorkflow = (
+  options?: UseMutationOptions<WorkflowResponseDto, unknown, Parameters<typeof updateWorkflow>[0]>
```
LetItRock, line 6: IMO we should not expose the react-query options as we will be tied by it's interface, but rather expose ours with the needed functionality.

LetItRock (approved):

apps/dashboard/src/hooks/use-update-workflow.ts
desiprisg: Ideally, I wouldn't create a `useUpdateWorkflow` hook in general. If we create one hook per mutation/query, we are already tied by its interface. I don't see any difference.

LetItRock: the difference is that we will be free to change the implementation of the hook to whatever we want if we decide

PR #6973, opened 2024-11-12
feat(api): add origin backward compatibility mapping

djabarovgeorge:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
@@ -9,10 +9,12 @@ import {
   StepContentIssueEnum,
   StepTypeEnum,
   WorkflowOriginEnum,
+  WorkflowTypeEnum,
 } from '@novu/shared';
 import { merge } from 'lodash/fp';
 import _ = require('lodash');
 import { GetWorkflowByIdsUseCase } from '@novu/application-generic';
+import { NotificationTemplateEntity } from '@novu/dal';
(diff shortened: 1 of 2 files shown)


apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -141,15 +143,35 @@ export class GeneratePreviewUsecase {
...
+   * We need to do so for backward compatibility reasons.
+   */
+  private buildOrigin(persistedWorkflow: NotificationTemplateEntity): WorkflowOriginEnum {
+    if (persistedWorkflow.origin) {
+      return persistedWorkflow.origin;
+    }
+
+    if (persistedWorkflow.type === WorkflowTypeEnum.ECHO || persistedWorkflow.type === WorkflowTypeEnum.BRIDGE) {
+      return WorkflowOriginEnum.EXTERNAL;
+    } else if (persistedWorkflow.type === WorkflowTypeEnum.REGULAR) {
```
SokratisVidros: There is no need for else when using return ;)
```suggestion
    }
    if (persistedWorkflow.type === WorkflowTypeEnum.REGULAR) {
```

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: you right, the `returns` were added after a refactor 😅

PR #6976, opened 2024-11-12
feat(providers): Add possibility to Onesignal choose version api

suplere:
### What changed? Why was the change needed?

This PR add possibility to Onesignal provider choose between:
a) player model
b) user model

[User model documentation](https://documentation.onesignal.com/docs/users) 

<details>
<summary><strong>Expand for optional sections</strong></summary>

### Special notes for your reviewer

</details>

PR #6979, opened 2024-11-13
feat(providers): provide the ability to send to include_external_user…

L-U-C-K-Y:

Implements: https://github.com/novuhq/novu/issues/3479#issue-1721795062, https://discord.com/channels/895029566685462578/1185702635219714219/1185702635219714219

This pull request enhances the functionality of the OneSignal push notification provider by adding support for targeting users via external user IDs. It includes changes to the provider's main class and its test suite to accommodate this new feature.

* **Provider Class Changes:**
  * Added `oneSignalOptions` to the overrides in the `OneSignalPushProvider` class to support targeting via `include_external_user_ids` or `include_player_ids` based on the `useExternalUserIds` flag.

* **Interface Changes:**
  * Updated the `IPushOptions` interface to include the new `oneSignalOptions` field with an optional `useExternalUserIds` boolean property.

* **Test Suite Additions:**
  * Added new tests to verify that `include_external_user_ids` is used when `useExternalUserIds` is true.
  * Added tests to ensure `include_player_ids` is used when `useExternalUserIds` is false.
  * Added tests to confirm that `include_player_ids` is the default when `oneSignalOptions` is not provided.…_ids instead of include_player_ids

### Screenshots

<details>

</details>

### DIFF
--- a/packages/providers/src/lib/push/one-signal/one-signal.provider.spec.ts
+++ b/packages/providers/src/lib/push/one-signal/one-signal.provider.spec.ts
@@ -126,4 +126,100 @@ describe('test onesignal notification api', () => {
     });
     expect(res.id).toEqual(response.data.id);
   });
+
+  test('should use include_external_user_ids when useExternalUserIds is true', async () => {
+    const provider = new OneSignalPushProvider({
+      appId: 'test-app-id',
+      apiKey: 'test-key',
+    });
+
+    const response = {
+      data: {
+        id: 'result',
+      },
+    };
+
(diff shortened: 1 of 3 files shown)


suplere:
Hello, this solution still use old API. The old API may be deprecated or even disabled over time.
You use include_external_user_ids - but with old API.

L-U-C-K-Y:
> Hello, this solution still use old API. The old API may be deprecated or even disabled over time. You use include_external_user_ids - but with old API.

I agree, we should use this one, it's the more proper way: https://github.com/novuhq/novu/pull/6976

PR #6980, opened 2024-11-13
feat(node): add deleteByTransactionId method

jainpawan21:
- Add `deleteByTransactionId` method

### Screenshots

<details>

</details>

### DIFF
--- a/packages/node/src/lib/messages/messages.ts
+++ b/packages/node/src/lib/messages/messages.ts
@@ -19,7 +19,21 @@ export class Messages extends WithHttp implements IMessages {
     });
   }
 
+  /**
+   * Deletes a single message notification by messageId.
+   *
+   * @param {string} messageId - The MongoDB ID of the message to delete.
+   */
   async deleteById(messageId: string) {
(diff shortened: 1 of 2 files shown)


packages/node/src/lib/messages/messages.ts
```diff
@@ -19,7 +19,21 @@ export class Messages extends WithHttp implements IMessages {
...
   async deleteById(messageId: string) {
     return await this.http.delete(`${BASE_PATH}/${messageId}`);
   }
+
+  /**
+   * Deletes multiple message notifications by transactionId.
+   *
+   * @param {string} transactionId - The unique identifier of the event whose notifications will be deleted.
+   */
+  async deleteByTransactionId(transactionId: string) {
```
jainpawan21, line 36: Shall I rename this method to `deleteMessagesBytTransactionId` to highlight that all messages related to this transactionId will be deleted?

scopsy (approved):

packages/node/src/lib/messages/messages.ts
scopsy: No, it's already on the messaged domain. So I think it's totally fine

PR #6983, opened 2024-11-13
feat(framework): Expose `Workflow` resource type in public API

### LINKED ISSUE #6954: 🐛 Bug Report: @novu/framework: 2.4.0 Breaking imports
### 📜 Description

version 2.4.0 no longer exports the type Workflow, causing imports to break in typescript

### 👟 Reproduction steps

install latest @novu/framework
Try importinng the Workflow type

### 👍 Expected behavior

Type should be exported and should be accessible in client code, where type inference may not be straight forward, for example in a custom monorepo

image

### Novu version

Novu SaaS

### npm version

9.5.1

### node version

18.16.0

_No response_

None

rifont:
* Expose `Workflow` resource type in public API
  * This typing is useful for library and plugin authors to create generic Novu Framework integrations with type inference

closes https://github.com/novuhq/novu/issues/6954

### Screenshots
_Intelliense of exposed `Workflow` type_
image

<details>

</details>

### DIFF
--- a/packages/framework/src/types/discover.types.ts
+++ b/packages/framework/src/types/discover.types.ts
@@ -65,10 +65,16 @@ export type DiscoverWorkflowOutput = {
   description?: string;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Workflow<T_Payload = any> = {
+/**
+ * A workflow resource.
+ *
+ * @property `id` - The unique identifier for the workflow.
+ * @property `trigger` - The function to trigger the workflow.
+ * @property `discover` - The function to discover the workflow definition.
+ */
+export type Workflow<T_Payload = Record<string, unknown>> = {
   /**
-   * The unique identifier of the workflow.
+   * The unique identifier for the workflow.
    */
   id: string;
   /**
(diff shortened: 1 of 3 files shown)


packages/framework/src/types/discover.types.ts
```diff
@@ -65,10 +65,16 @@ export type DiscoverWorkflowOutput = {
...
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Workflow<T_Payload = any> = {
+/**
+ * A workflow resource.
+ *
+ * @property `id` - The unique identifier for the workflow.
+ * @property `trigger` - The function to trigger the workflow.
+ * @property `discover` - The function to discover the workflow definition.
+ */
+export type Workflow<T_Payload = Record<string, unknown>> = {
```
rifont, line 75: We should never have a case where `T_Payload` is not inferred, but removing the `any` here makes it completely strict and prevents regression.

packages/framework/src/types/discover.types.ts
```diff
@@ -65,10 +65,16 @@ export type DiscoverWorkflowOutput = {
...
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export type Workflow<T_Payload = any> = {
+/**
+ * A workflow resource.
+ *
+ * @property `id` - The unique identifier for the workflow.
+ * @property `trigger` - The function to trigger the workflow.
+ * @property `discover` - The function to discover the workflow definition.
+ */
+export type Workflow<T_Payload = never> = {
```
rifont, line 75: There is never a case where `T_Payload` is inferrable without access to the `payloadSchema`, so we make it as such. This prevents using an inferred Workflow payload type without a corresponding `payloadSchema`.

SokratisVidros (approved):

PR #6992, opened 2024-11-13
feat(api): allow pass empty step control value

djabarovgeorge:
### Screenshots

<details>

</details>


SokratisVidros (approved):

libs/application-generic/src/usecases/message-template/delete-message-template/delete-message-template.usecase.ts
```diff
@@ -16,40 +17,48 @@ export class DeleteMessageTemplate {
...
-      const changeId = await this.changeRepository.getChangeId(
-        command.environmentId,
-        ChangeEntityTypeEnum.MESSAGE_TEMPLATE,
-        command.messageTemplateId,
-      );
+      await this.messageTemplateRepository.withTransaction(async () => {
+        await this.controlValuesRepository.delete({
+          _environmentId: command.environmentId,
+          _stepId: command.messageTemplateId,
+        });
```
djabarovgeorge, line 29: till now on every message removal, we had a stale preference entity kept in the db.
in the last week, we added a removal on workflow delete, but in case the workflow is not deleted we would still have those controls.

SokratisVidros:
@djabarovgeorge please double check if we still need this PR and if yes, rebase it.

tatarco:
this is available via the patch, no need for this PR anymore

PR #7002, opened 2024-11-14
feat(dashboard,web): opt-in app switching and redirects

ChmaraX:

- added providers to check if user is opted-in/out, in order to redirect:
  - if user is opted-in and visits legacy dashboard path which exists in new dashboard - redirect there
  - if user is opted-out and visits new dashboard - redirect back to old
- automatically set theme to light when user opts in for new dashboard

### DIFF
--- a/apps/dashboard/src/context/opt-in-provider.tsx
+++ b/apps/dashboard/src/context/opt-in-provider.tsx
@@ -1,18 +1,16 @@
 import { useNewDashboardOptIn } from '@/hooks/use-new-dashboard-opt-in';
 import { NewDashboardOptInStatusEnum } from '@novu/shared';
 import { PropsWithChildren, useEffect } from 'react';
-import { useNavigate } from 'react-router-dom';
 
 export const OptInProvider = (props: PropsWithChildren) => {
-  const navigate = useNavigate();
   const { children } = props;
-  const { status, isLoaded } = useNewDashboardOptIn();
+  const { status, isLoaded, redirectToLegacyDashboard } = useNewDashboardOptIn();
 
   useEffect(() => {
     if (isLoaded && status !== NewDashboardOptInStatusEnum.OPTED_IN) {
-      window.location.href = '/legacy/workflows';
+      redirectToLegacyDashboard();
     }
-  }, [status, navigate, isLoaded]);
+  }, [status, redirectToLegacyDashboard, isLoaded]);
 
   return <>{children}</>;
 };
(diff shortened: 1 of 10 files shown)


apps/dashboard/src/context/opt-in-provider.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { useNavigate } from 'react-router-dom';
+
+export const OptInProvider = (props: PropsWithChildren) => {
+  const navigate = useNavigate();
+  const { children } = props;
+  const { status, isLoaded } = useNewDashboardOptIn();
+
+  useEffect(() => {
+    if (isLoaded && status !== NewDashboardOptInStatusEnum.OPTED_IN) {
+      window.location.href = '/legacy/workflows';
```
ChmaraX: using `navigate()` here won't work since its external url

apps/web/src/hooks/useNewDashboardOptIn.ts
```diff
@@ -28,9 +29,11 @@ export function useNewDashboardOptIn() {
     window.location.href = NEW_DASHBOARD_URL || window.location.origin;
   };
 
-  const optIn = () => {
+  const optIn = async () => {
     segment.track('New dashboard opt-in');
-    updateUserOptInStatus(NewDashboardOptInStatusEnum.OPTED_IN);
+    await updateUserOptInStatus(NewDashboardOptInStatusEnum.OPTED_IN);
+    localStorage.setItem('mantine-theme', 'light');
```
ChmaraX, line 35: this needs to be set directly in local storage otherwise when using hook the operation seems to be async and can't be awaited

apps/web/src/components/providers/OptInProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+       * - /legacy/workflows/edit/123 -> /workflows
+       */
+      if (NEW_DASHBOARD_ROUTES.some((route) => currentRoute.includes(route))) {
+        /**
+         * TODO: in order to redirect to the same route, we need to translate the
+         * "dev_env_<id>" or wf/step slugs to legacy environment id and vice-versa
+         *
+         * note: /legacy is part of public URL, so we can't navigate() outside of that
+         */
+        window.location.href = window.location.origin;
```
ChmaraX, line 43: for now we redirect to the default page which is the workflows page - otherwise we would need to parse old env id to new env slugs to redirect to `env/<env_slug>/workflows` - will do in separate PR if there is a hard requirement to keep env id

SokratisVidros (approved):

PR #7005, opened 2024-11-14
refactor(api, application-generic): Encapsulate preference mutations in workflow use-cases

rifont:
* Encapsulate preference mutations in workflow use-cases
  * Move all preference handling to `CreateWorkflow`, `UpdateWorkflow`, and `DeleteWorkflow` use-cases
  * Remove preference handling from `UpsertPreferences` use-case

### Screenshots

<details>

</details>


apps/api/src/app/preferences/dtos/preferences.dto.ts
```diff
@@ -40,7 +40,7 @@ export class Channels {
 export class PreferencesDto {
   @ValidateNested({ each: true })
   @Type(() => WorkflowPreference)
-  workflow: WorkflowPreference;
+  all: WorkflowPreference;
```
rifont, line 43: this was just completely wrong before.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
```diff
@@ -299,7 +333,14 @@ export class UpdateWorkflow {
...
 
-    return notificationTemplateWithStepTemplate;
+    return this.getWorkflowByIdsUseCase.execute(
+      GetWorkflowByIdsCommand.create({
+        userId: command.userId,
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        identifierOrInternalId: command.id,
+      }),
+    );
```
rifont: Always use the `GetWorkflow` command at the end. Not the most performant, but the endpoint is infrequently used so it's a suitable compromise to avoid an even bigger refactor.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
```diff
@@ -201,6 +179,72 @@ export class UpdateWorkflow {
...
+
+    await this.notificationTemplateRepository.update(
+      {
+        _id: command.id,
+        _environmentId: command.environmentId,
+      },
+      {
+        $set: updatePayload,
+      },
+    );
```
rifont: The database mutation should happen before invalidating.

apps/api/src/app/workflows-v1/notification-template.controller.ts
```diff
@@ -107,8 +113,10 @@ export class NotificationTemplateController {
         tags: body.tags,
         description: body.description,
         workflowId: body.identifier,
-        critical: body.critical,
-        preferenceSettings: body.preferenceSettings,
+        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
+        userPreferences:
+          body.preferenceSettings &&
+          buildWorkflowPreferencesFromPreferenceChannels(body.critical ?? false, body.preferenceSettings),
```
rifont: Internally, all Workflow DTOs now carry a `defaultPreferences` and `userPreferences` object now.

`preferenceSettings` and `critical` are removed from all internal DTOs in preparation for their full removal.

apps/api/src/app/workflows-v1/workflow-v1.controller.ts
```diff
@@ -112,8 +118,11 @@ export class WorkflowControllerV1 {
         tags: body.tags,
         description: body.description,
         workflowId: body.identifier,
-        critical: body.critical,
-        preferenceSettings: body.preferenceSettings,
+        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
+
+        userPreferences:
+          body.preferenceSettings &&
+          buildWorkflowPreferencesFromPreferenceChannels(body.critical ?? false, body.preferenceSettings),
```
rifont:
```suggestion
        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
        userPreferences:
          body.preferenceSettings &&
          buildWorkflowPreferencesFromPreferenceChannels(body.critical ?? false, body.preferenceSettings),
```

apps/api/src/app/workflows-v1/workflow-v1.controller.ts
```diff
@@ -209,8 +217,11 @@ export class WorkflowControllerV1 {
         notificationGroup: body.notificationGroup,
         active: body.active ?? false,
         draft: !body.active,
-        critical: body.critical ?? false,
-        preferenceSettings: body.preferenceSettings,
+        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
+
+        userPreferences:
```
rifont:
```suggestion
        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
        userPreferences:
```

apps/api/src/app/workflows-v1/notification-template.controller.ts
```diff
@@ -107,8 +113,10 @@ export class NotificationTemplateController {
         tags: body.tags,
         description: body.description,
         workflowId: body.identifier,
-        critical: body.critical,
-        preferenceSettings: body.preferenceSettings,
+        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
```
rifont, line 117: All V1 Workflows now write a `defaultPreferences` to provide a default for the preference merging behaviour.

apps/api/src/app/workflows-v1/usecases/get-notification-template/get-notification-template.usecase.ts
```diff
@@ -9,29 +10,18 @@ import { GetNotificationTemplateCommand } from './get-notification-template.comm
...
   async execute(command: GetNotificationTemplateCommand): Promise<NotificationTemplateEntity> {
-    const isInternalId = NotificationTemplateRepository.isInternalId(command.workflowIdOrIdentifier);
+    const workflow = await this.getWorkflowByIdsUseCase.execute(
+      GetWorkflowByIdsCommand.create({
+        identifierOrInternalId: command.workflowIdOrIdentifier,
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        userId: command.userId,
+      })
+    );
```
rifont, line 23: Use the updated `GetWorkflowByIdsUseCase` which handles all preference fetching when retrieving V1 workflows.

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
```diff
@@ -14,34 +12,31 @@ import {
...
-export function toResponseWorkflowDto(
-  template: NotificationTemplateEntity,
-  preferences: GetPreferencesResponseDto | undefined
-): WorkflowResponseDto {
+export function toResponseWorkflowDto(workflow: GetWorkflowByIdsResponseDto): WorkflowResponseDto {
   const preferencesDto: PreferencesResponseDto = {
-    user: preferences?.source[PreferencesTypeEnum.USER_WORKFLOW] || null,
-    default: preferences?.source[PreferencesTypeEnum.WORKFLOW_RESOURCE] || DEFAULT_WORKFLOW_PREFERENCES,
+    user: workflow.userPreferences,
+    default: workflow.defaultPreferences,
```
rifont, line 21: Mostly renaming, this is the only functional change to start reading from the internal Workflow DTO.

apps/api/src/app/workflows-v2/usecases/build-step-data/build-step-data.usecase.ts
```diff
@@ -44,21 +44,12 @@ export class BuildStepDataUsecase {
...
     });
-
-    if (!workflow) {
-      throw new BadRequestException({
-        message: 'No workflow found',
-        workflowId: command.identifierOrInternalId,
-      });
-    }
-
-    return workflow;
```
rifont, line 66: There are many changes like this to centralize Workflow fetching, delegating the preference fetching to the Workflow use-case layer.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
rifont, line 1: `UpsertWorkflow` stops dealing with preferences now, leaving it to the downstream `CreateWorkflow` and `UpdateWorkflow` use-cases.

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
```diff
@@ -289,7 +305,7 @@ export class CreateWorkflow {
     templateSteps: INotificationTemplateStep[],
     trigger: INotificationTrigger,
     triggerIdentifier: string,
-  ) {
+  ): Promise<GetWorkflowByIdsResponseDto> {
```
rifont: Could possibly rename `GetWorkflowByIdsResponseDto` to something more generic, but I think that's for a follow-up PR.

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
```diff
@@ -338,7 +380,14 @@ export class CreateWorkflow {
...
 
-    return item;
+    return this.getWorkflowByIdsUseCase.execute(
+      GetWorkflowByIdsCommand.create({
+        userId: command.userId,
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        identifierOrInternalId: savedWorkflow._id,
+      }),
+    );
```
rifont, line 403: Just fetch again using the `GetWorkflowByIds` use-case to let that use-case handle data collection merging in a single place. Not the most performant, but this is an infrequently called use-case so it's acceptable.

We see the same thing happen in `UpdateWorkflow` later.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.command.ts
rifont, line 1: Typing & class validator tidy ups to make sure that:
* Workflow prefs have fully formed preference sets
* Subscriber prefs allow partial preference sets

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -1,37 +1,49 @@
...
+  ): Promise<WorkflowPreferencesFull> {
     return this.upsert({
       templateId: command.templateId,
       environmentId: command.environmentId,
       organizationId: command.organizationId,
-      preferences: builtPreferences,
+      preferences: command.preferences,
       type: PreferencesTypeEnum.WORKFLOW_RESOURCE,
-    });
+    }) as Promise<WorkflowPreferencesFull>;
```
rifont, line 46: Not the most elegant, but it does the job for now without introducing a lot more generics for response types.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -112,7 +124,8 @@ export class UpsertPreferences {
 
     if (command.preferences === null) {
       if (!foundPreference) {
-        throw new BadRequestException('Preference not found');
+        // Gracefully return if the preference is not found
+        return;
```
rifont: Needed to allow for safe upserts when provided with `null` in the case of a User-Workflow preference set, ensuring the operation can be idempotent.

libs/application-generic/src/usecases/upsert-preferences/upsert-preferences.usecase.ts
```diff
@@ -112,7 +124,8 @@ export class UpsertPreferences {
 
     if (command.preferences === null) {
       if (!foundPreference) {
-        throw new BadRequestException('Preference not found');
+        // Gracefully return if the preference is not found
```
rifont:
```suggestion
        // Gracefully return if the preference is not found,
        // ensuring the operation is idempotent.
```

packages/shared/src/utils/buildWorkflowPreferences.ts
```diff
@@ -42,3 +43,26 @@ export const buildWorkflowPreferences = (
...
+      (output, [channel, value]) => ({
+        ...output,
+        [channel as ChannelTypeEnum]: {
+          enabled: value,
+        },
+      }),
+      {} as WorkflowPreferences['channels']
+    ),
+  });
+};
```
rifont, line 68: Helper function for V1 workflow preferences.

apps/api/src/app/workflows-v1/notification-template.controller.ts
```diff
@@ -108,7 +114,10 @@ export class NotificationTemplateController {
         description: body.description,
         workflowId: body.identifier,
         critical: body.critical,
-        preferenceSettings: body.preferenceSettings,
+        defaultPreferences: DEFAULT_WORKFLOW_PREFERENCES,
+        userPreferences:
+          body.preferenceSettings &&
+          buildWorkflowPreferencesFromPreferenceChannels(body.critical, body.preferenceSettings),
```
SokratisVidros, line 120: Should this code be identical to [this line](https://github.com/novuhq/novu/pull/7005/files#diff-c995a3ecb137586ca535a06504687da459250bbc1a11ecf9a28c11bb0b25bc5dR205)?

libs/application-generic/src/usecases/workflow/get-workflow-by-ids/get-workflow-by-ids.dto.ts
```diff
@@ -0,0 +1,8 @@
+import { NotificationTemplateEntity } from '@novu/dal';
+import { WorkflowPreferences } from '@novu/shared';
+
+export class GetWorkflowByIdsResponseDto extends NotificationTemplateEntity {
```
SokratisVidros:
```suggestion
export class WorkflowResponseDto extends NotificationTemplateEntity {
```

packages/shared/src/utils/buildWorkflowPreferences.ts
SokratisVidros: Will this be removed after the v1->v2 preference data migration?

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
SokratisVidros: It's okay for now. One thing I noticed, don't we also fetch at [this point](https://github.com/novuhq/novu/pull/7005/files#diff-215a725ccdc52f5207a653ad21479e039d533723daef8101e1897e14b8da5d30R302)? If not mistaken, no updates take place since that fetching.

SokratisVidros: We can also use https://mongoosejs.com/docs/tutorials/findoneandupdate.html and do everything in a single query.

apps/api/src/app/workflows-v1/notification-template.controller.ts
rifont: Yes it should, good catch.

rifont: Updated.

packages/shared/src/utils/buildWorkflowPreferences.ts
rifont: Yes, exactly. We will need to add a long-lived branched that folks can checkout to perform the migration with the legacy repositories accessible, whilst our main branch will have it removed.

libs/application-generic/src/usecases/workflow/get-workflow-by-ids/get-workflow-by-ids.dto.ts
rifont: I renamed to `WorkflowInternalResponseDto` to avoid a conflict with the V2 Workflow `WorkflowResponseDto`. Let me know if you think that's acceptable.

libs/application-generic/src/usecases/create-workflow/create-workflow.usecase.ts
rifont: Done - renamed to `WorkflowInternalResponseDto`.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
rifont: Good catch. I removed the duplicate lookup.

We absolutely should seek to use `findOneAndUpdate` everywhere.

libs/application-generic/src/usecases/get-subscriber-template-preference/get-subscriber-template-preference.usecase.ts
```diff
@@ -117,7 +117,7 @@ export class GetSubscriberTemplatePreference {
           subscriberWorkflowPreferenceV2.preferences,
         );
       subscriberPreferenceType = subscriberWorkflowPreferenceV2.type;
-      critical = subscriberWorkflowPreferenceV2.preferences?.all?.readOnly;
+      critical = subscriberWorkflowPreferenceV2.preferences.all.readOnly;
```
rifont:
```suggestion
      critical = subscriberWorkflowPreferenceV2.preferences?.all?.readOnly;
```

libs/application-generic/src/usecases/workflow/delete-workflow/delete-workflow.usecase.ts
```diff
@@ -78,6 +84,31 @@ export class DeleteWorkflowUseCase {
         });
       }
 
+      const upsertUserWorkflowPreferencesPromise =
+        this.upsertPreferences.upsertUserWorkflowPreferences(
```
SokratisVidros: This is effectively a preference deletion. Do we need to use the upsert workflow preference use case? The naming is off, as upsert implies create or update. Passing `null` to indicate removal is a dangerous pattern.

I also suggest we add a transaction. See [here](https://github.com/novuhq/novu/blob/5a0efd7f02cb4c579c65439751daefa029cc551e/apps/api/src/app/subscribers/usecases/remove-subscriber/remove-subscriber.usecase.ts#L58).

libs/application-generic/src/usecases/workflow/get-workflow-by-ids/get-workflow-by-ids.usecase.ts
```diff
@@ -41,6 +50,26 @@ export class GetWorkflowByIdsUseCase {
...
-    return workflowEntity;
+    const workflowPreferences = await this.getPreferences.safeExecute(
+      GetPreferencesCommand.create({
+        environmentId: command.environmentId,
+        organizationId: command.organizationId,
+        templateId: workflowEntity._id,
+      }),
+    );
+    /**
+     * @deprecated - use `userPreferences` and `defaultPreferences` instead
```
SokratisVidros, line 61: I guess this needs to be removed after data migration. If that's the case, let's mark all the places in the code that can be removed after the data migration so that we can spot them easily.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
```diff
@@ -197,10 +184,106 @@ export class UpdateWorkflow {
...
+      const defaultUserPreferences =
+        command.userPreferences ?? existingTemplate.userPreferences;
+      const defaultCritical =
+        command.userPreferences?.all?.readOnly ??
+        command.critical ??
+        existingTemplate.userPreferences?.all?.readOnly ??
+        existingTemplate.critical;
+
+      const userPreferences =
+        command.userPreferences === null
```
SokratisVidros: Does this mean that user preferences are deleted?

SokratisVidros (changes requested):

libs/application-generic/src/usecases/workflow/get-workflow-by-ids/get-workflow-by-ids.usecase.ts
rifont: Yes, it does. Everywhere that needs removal is effectively via the presence of the `preferenceSettings` data property, that implicitly acts as the marker.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
rifont: Yes. I added an explicit `DeletePreference` use-case now and we don't have this ternary anymore.

libs/application-generic/src/usecases/workflow/delete-workflow/delete-workflow.usecase.ts
rifont: I agree, I have added a new `DeletePreference` use-case to explicitly handle deletion now.

Regarding the transaction, this would necessitate passing a transaction reference across use-cases, which isn't something that we have a pattern for. Do we think adding a transaction now is a requirement, or should we address it separately?

rifont: Scratch the above, I'm adding a transaction - it's easy as there is no need to explicitly pass a context, it's implicit available.

rifont: Transactions are now in place for all the workflow mutation use-case - `CreateWorkflow`, `UpdateWorkflow`, `DeleteWorkflow`.

SokratisVidros (approved):

PR #7006, opened 2024-11-14
feat(dashboard): in-app editor loading state

LetItRock:

1. In-App Editor loading skeleton.
2. Refactored the editor code a bit and introduced `StepEditorContext` which is the central place where we fetch the step information. More about in the comments.
3. Added the `variables` to missing In-App form fields.

### Screenshots
Screenshot 2024-11-14 at 14 33 44

Screenshot 2024-11-14 at 14 44 34

https://github.com/user-attachments/assets/9bfd5d05-3697-4de9-b022-8a3826f28e9f

https://github.com/user-attachments/assets/d9b778aa-f29a-4c3d-9ce1-51eb7c5d0754

### DIFF
--- a/apps/dashboard/src/components/primitives/form/avatar-picker.tsx
+++ b/apps/dashboard/src/components/primitives/form/avatar-picker.tsx
@@ -1,14 +1,20 @@
+import { useState, forwardRef, useMemo } from 'react';
+import { liquid } from '@codemirror/lang-liquid';
+import { EditorView } from '@uiw/react-codemirror';
+import { RiEdit2Line, RiErrorWarningFill, RiImageEditFill } from 'react-icons/ri';
+
 import { Avatar, AvatarImage } from '@/components/primitives/avatar';
 import { Button } from '@/components/primitives/button';
 import { FormMessage } from '@/components/primitives/form/form';
-import { Input, InputField } from '@/components/primitives/input';
+import { InputField } from '@/components/primitives/input';
 import { Label } from '@/components/primitives/label';
 import { Popover, PopoverContent, PopoverTrigger } from '@/components/primitives/popover';
 import { Separator } from '@/components/primitives/separator';
 import TextSeparator from '@/components/primitives/text-separator';
-import { useState, forwardRef } from 'react';
-import { RiEdit2Line, RiErrorWarningFill, RiImageEditFill } from 'react-icons/ri';
 import { useFormField } from './form-context';
+import { Editor } from '../editor';
+import { useStepEditorContext } from '@/components/workflow-editor/steps/hooks';
+import { parseStepVariablesToLiquidVariables } from '@/utils/parseStepVariablesToLiquidVariables';
 
 const predefinedAvatars = [
   `${window.location.origin}/images/avatar.svg`,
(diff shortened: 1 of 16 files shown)


apps/dashboard/src/components/primitives/form/avatar-picker.tsx
```diff
@@ -25,25 +31,31 @@ const predefinedAvatars = [
...
+type AvatarPickerProps = {
+  name: string;
+  value: string;
+  onChange?: (value: string) => void;
+};
 
-export const AvatarPicker = forwardRef<HTMLInputElement, AvatarPickerProps>(({ id, ...props }, ref) => {
+export const AvatarPicker = forwardRef<HTMLInputElement, AvatarPickerProps>(({ name, value, onChange }, ref) => {
+  const { step } = useStepEditorContext();
+  const variables = useMemo(() => (step ? parseStepVariablesToLiquidVariables(step.variables) : []), [step]);
```
LetItRock, line 42: the component was missing `variables` and the CodeMirror editor field for the avatar URL

apps/dashboard/src/components/workflow-editor/steps/edit-step-sidebar.tsx
```diff
@@ -10,45 +9,25 @@ import {
...
-
-  const { workflow } = useFetchWorkflow({
-    workflowSlug,
-  });
-
-  const { step } = useFetchStep({ workflowSlug, stepSlug });
-  const stepType = useMemo(
-    () => workflow?.steps.find((el) => getStepBase62Id(el.slug) === getStepBase62Id(stepSlug))?.type,
-    [stepSlug, workflow]
-  );
```
LetItRock, line 35: Moved to the `StepEditorProvider` component which shares the step across the components.
I decided to do that because, in a few places, we were fetching the step with the `useFetchStep` hook, which on mount fetches the data from API, triggering the loading skeleton to be shown, and it was infinite because the skeleton was in the parent.

apps/dashboard/src/components/workflow-editor/steps/step-skeleton.tsx
```diff
@@ -0,0 +1,118 @@
...
+  },
+  [StepTypeEnum.SMS]: SingleLineSkeleton,
+  [StepTypeEnum.PUSH]: SingleLineSkeleton,
+  [StepTypeEnum.DIGEST]: () => null,
+  [StepTypeEnum.DELAY]: () => null,
+  [StepTypeEnum.TRIGGER]: () => null,
+  [StepTypeEnum.CUSTOM]: () => null,
+};
+
+export const StepSkeleton = ({
```
LetItRock, line 57: the step skeleton which simulates the step editor shell for the sidebar with: header, body, footer

apps/dashboard/src/components/workflow-editor/steps/step-skeleton.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <span className="sr-only">Close</span>
+        </Button>
+      </header>
+      <Separator />
+      <div className="flex h-full w-full flex-col gap-3 px-3 py-3.5">
+        {workflowOrigin && workflowOrigin !== WorkflowOriginEnum.EXTERNAL ? (
+          <SkeletonContent />
+        ) : (
+          <SingleLineSkeleton />
+        )}
```
LetItRock, line 108: There is a problem that I didn't know how we can solve. Basically, when the page is refreshed, we don't have the `stepType` or `workflow.origin` fields. This creates the problem that we don't know what skeleton to render as the content, because ideally it should be:
- different for code created steps - because the editor will have only custom control fields
- different for the dashboard created steps, even different depending on the step type

So what the code does here is:
- on page refresh it will show `SingleLineSkeleton`, while the `workflow` is loading
- on app navigation, the `workflow` is loaded and we can show the loader for a particular step type

Maybe I unnecessarily complicated it, but it doesn't look bad 😅

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
```diff
@@ -25,25 +31,31 @@ const predefinedAvatars = [
...
-    <div className="space-y-2">
+    <div className="size-10 space-y-2">
       <Popover modal={true} open={isOpen} onOpenChange={setIsOpen}>
         <PopoverTrigger asChild>
           <Button variant="outline" size="icon" className="text-foreground-600 relative size-10">
-            {props.value ? (
+            {value ? (
               <Avatar className="p-px">
-                <AvatarImage src={props.value as string} />
+                <AvatarImage src={value as string} />
```
BiswaViraj, line 58:
```suggestion
                <AvatarImage src={value} />
```
It's a string already so this isn't needed

BiswaViraj (commented):
image
image

The url inputs are not accepting the payload variables

LetItRock:
> The url inputs are not accepting the payload variables

thanks @BiswaViraj, we need to remove the URL validation, I guess. Otherwise, we can't allow for the variables

BiswaViraj:
> we need to remove the URL validation, I guess. Otherwise, we can't allow for the variables


Yes, I think we can remove it.currently it will also fail for relative URLs as well "/posts/new"

LetItRock:
the validation will be removed on the BE side in separate PR

BiswaViraj (approved):

PR #7015, opened 2024-11-15
fix(api): add patch step api and consolidate post update processing

tatarco:

1. Added a Patch Step supporting name + control values.
2. Add e2e tests to validate the behavior stays consistent.
3. Deprecate (but not delete) existing way of updating to not break FE.
4. migrate internals into a joint logic for upsert and patch in order to remove it easily next week.
5. consolidate postprocess logic (issues + payload schema builder) to work after patch & upsert to keep a consistent behavior

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/patch-step-data/patch-step-data.usecase.ts
```diff
@@ -0,0 +1,116 @@
...
+      organizationId: command.user.organizationId,
+      userId: command.user._id,
+    });
+
+    if (!workflow) {
+      throw new BadRequestException({
+        message: 'No workflow found',
+        workflowId: command.identifierOrInternalId,
+      });
+    }
```
rifont: `getWorkflowByIdsUseCase` never returns falsy values, it handles the null case internally. So we can remove this redundant check.

```suggestion
```

packages/shared/src/clients/workflows-client.ts
```diff
@@ -98,5 +107,6 @@ export const createWorkflowClient = (baseUrl: string, headers: HeadersInit = {})
     searchWorkflows,
     getWorkflowTestData,
     getWorkflowStepData,
+    patchWorkflowStepData, // Include the new method in the returned object
```
rifont: ❓ **question:** Seems like a redundant/superfluous comment, was this intentional?‏
```suggestion
    patchWorkflowStepData,
```

rifont (approved):

packages/shared/src/clients/workflows-client.ts
tatarco: chatgpt... forgot to tell him no comments

PR #7019, opened 2024-11-15
fix(api): Add a Patch Workflow endpoint 

tatarco:

1. Added Workflow Patch API.
2. Modified Rest to work without a list of fields to update, both for step and workflow.
3. removed URI validation from avatar to enable sending variable inside.

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
+++ b/apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
@@ -1,4 +1,4 @@
-import { JSONSchemaDto, UiComponentEnum, UiSchema, UiSchemaGroupEnum, UiSchemaProperty } from '@novu/shared';
+import { JSONSchemaDto, UiComponentEnum, UiSchema, UiSchemaGroupEnum } from '@novu/shared';
 
 const ABSOLUTE_AND_RELATIVE_URL_REGEX = '^(?!mailto:)(?:(https?):\\/\\/[^\\s/$.?#].[^\\s]*)|^(\\/[^\\s]*)$';
 
@@ -34,7 +34,7 @@ export const inAppControlSchema = {
   properties: {
     subject: { type: 'string' },
     body: { type: 'string' },
-    avatar: { type: 'string', format: 'uri' },
+    avatar: { type: 'string' },
     primaryAction: actionSchema, // Nested primaryAction
     secondaryAction: actionSchema, // Nested secondaryAction
     data: { type: 'object', additionalProperties: true },
--- a/apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
+++ b/apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
@@ -1,13 +1,10 @@
 import { JSONSchemaDto, UiComponentEnum, UiSchema, UiSchemaGroupEnum } from '@novu/shared';
 
-const ABSOLUTE_AND_RELATIVE_URL_REGEX = '^(?!mailto:)(?:(https?):\\/\\/[^\\s/$.?#].[^\\s]*)|^(\\/[^\\s]*)$';
-
 const redirectSchema = {
   type: 'object',
   properties: {
     url: {
       type: 'string',
-      pattern: ABSOLUTE_AND_RELATIVE_URL_REGEX,
     },
     target: {
       type: 'string',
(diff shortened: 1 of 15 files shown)


apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
```diff
@@ -34,7 +34,7 @@ export const inAppControlSchema = {
   properties: {
     subject: { type: 'string' },
     body: { type: 'string' },
-    avatar: { type: 'string', format: 'uri' },
+    avatar: { type: 'string' },
```
SokratisVidros, line 34: 👍

apps/api/src/app/workflows-v2/usecases/patch-step-data/patch-step-data.command.ts
```diff
@@ -10,8 +10,6 @@ export class PatchStepDataCommand extends EnvironmentWithUserObjectCommand {
   @IsString()
   @IsNotEmpty()
   stepId: IdentifierOrInternalId;
-  @IsArray()
-  fieldsToUpdate: PatchStepFieldEnum[];
```
SokratisVidros, line 14: 💯

apps/api/src/app/workflows-v2/usecases/patch-step-data/patch-step-data.usecase.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable no-param-reassign */
 import { BadRequestException, Injectable } from '@nestjs/common';
```
SokratisVidros, line 2: We can now rename the usecase to `patch-step.usecase.ts`

apps/api/src/app/workflows-v2/usecases/patch-workflow/patch-workflow.usecase.ts
```diff
@@ -0,0 +1,76 @@
...
+  private async fetchWorkflow(command: PatchWorkflowCommand) {
+    const workflow = await this.getWorkflowByIdsUseCase.execute({
+      identifierOrInternalId: command.identifierOrInternalId,
+      environmentId: command.user.environmentId,
+      organizationId: command.user.organizationId,
+      userId: command.user._id,
+    });
+
+    if (!workflow) {
+      throw new BadRequestException({
```
SokratisVidros:
```suggestion
      throw new NotFoundException({
```

apps/api/src/app/workflows-v2/usecases/patch-workflow/patch-workflow.usecase.ts
```diff
@@ -0,0 +1,76 @@
...
+    await this.persistWorkflow(transientWorkflow, command.user);
+
+    return await this.getWorkflowUseCase.execute({
+      identifierOrInternalId: command.identifierOrInternalId,
+      user: command.user,
+    });
+  }
+
+  private patchWorkflowFields(persistedWorkflow, command: PatchWorkflowCommand) {
+    const transientWorkflow = { ...persistedWorkflow };
```
SokratisVidros, line 35: What about the other workflow fields, such as name, description, and tags? Since we got to the point and of all them belong to the same form, we should be able to patch all of them at once and ease the FE.

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/patch-workflow/patch-workflow.usecase.ts
tatarco: true, but they are also accessible from upsert, this one was a blocker for @BiswaViraj , the structure is there, we can add what ever we want with relatively small effort, was not trying to solve all the problems, only the one in front of me in an extendable way

apps/api/src/app/workflows-v2/shared/schemas/in-app-control.schema.ts
```diff
@@ -1,4 +1,4 @@
-import { JSONSchemaDto, UiComponentEnum, UiSchema, UiSchemaGroupEnum, UiSchemaProperty } from '@novu/shared';
+import { JSONSchemaDto, UiComponentEnum, UiSchema, UiSchemaGroupEnum } from '@novu/shared';
 
 const ABSOLUTE_AND_RELATIVE_URL_REGEX = '^(?!mailto:)(?:(https?):\\/\\/[^\\s/$.?#].[^\\s]*)|^(\\/[^\\s]*)$';
```
LetItRock: please remove this one as well

PR #7017, opened 2024-11-15
fix(dashboard): opt-in redirect logic fix

LetItRock:
### What changed? Why was the change needed?

- fixed the dashboard opt-in logic that was running to early
- improved the login page layout

### Screenshots

https://github.com/user-attachments/assets/5984d74d-b25c-46ef-9566-eabe66ea61ee

PR #7022, opened 2024-11-15
fix(web): when user is opted-in redirect to the new dashboard on route change

LetItRock:

When the user navigates in the web app to the route that starts with `/workflows` and is opted-in, he will be redirected to the new Dashboard app. The same logic applies when the Web app is embedded in the new Dashboard.

### Screenshots

https://github.com/user-attachments/assets/cf854cd8-6002-417c-93bc-65304892c132


apps/dashboard/src/context/opt-in-provider.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { NewDashboardOptInStatusEnum } from '@novu/shared';
-import { PropsWithChildren, useEffect } from 'react';
+import { useNewDashboardOptIn } from '@/hooks/use-new-dashboard-opt-in';
 
 export const OptInProvider = (props: PropsWithChildren) => {
   const { children } = props;
   const { status, isLoaded, redirectToLegacyDashboard } = useNewDashboardOptIn();
 
-  useEffect(() => {
+  useLayoutEffect(() => {
```
LetItRock, line 9: redirect before dom is updated

apps/web/src/Providers.tsx
```diff
@@ -37,15 +36,13 @@ const Providers: React.FC<PropsWithChildren<{}>> = ({ children }) => {
           <SegmentProvider>
             <QueryClientProvider client={queryClient}>
               <AuthProvider>
-                <OptInProvider>
```
LetItRock, line 40: the `OptInProvider` was only called when the application started, but we want to check opt-in when the route changes during app navigation

apps/web/src/components/layout/components/PrivatePageLayout.tsx
```diff
@@ -62,6 +63,8 @@ export function PrivatePageLayout() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  useOptInRedirect();
```
LetItRock: The `PrivatePageLayout` is a top-level component that also has access to the router location, that's why the opt-in hook was placed here.

apps/web/src/hooks/useOptInRedirect.ts
```diff
@@ -1,30 +1,24 @@
...
-  const { children } = props;
   const { status, isLoaded, redirectToNewDashboard } = useNewDashboardOptIn();
+  const isUserOptedIn = isLoaded && status && status === NewDashboardOptInStatusEnum.OPTED_IN;
+  const shouldHandleOptInRedirect = IS_EE_AUTH_ENABLED && isNewDashboardEnabled && isUserOptedIn;
 
-  useEffect(() => {
-    if (isLoaded && status === NewDashboardOptInStatusEnum.OPTED_IN) {
-      const currentRoute = window.location.pathname.replace('/legacy', '');
+  useLayoutEffect(() => {
+    if (shouldHandleOptInRedirect) {
```
LetItRock, line 20: in short, when the new dashboard ff is on, and the user is opted-in, we want to perform the check for the current `pathname` on app mount, and when the route changes

LetItRock: if the route starts with `/workflows` then the user will be redirected to the new dashboard app

SokratisVidros (approved):

SokratisVidros (approved):

ChmaraX (approved):

apps/web/src/Providers.tsx
ChmaraX: I was wondering why this worked for me before but I guess it was because of this `navigate()` that was present before and triggered the useEffect:
https://github.com/novuhq/novu/pull/7017/files#diff-cc17c9f31a51ec7181c4bc7af9932388e5a3a796df090c2c17517d0f72401e97L23

apps/web/src/components/layout/components/PrivatePageLayout.tsx
ChmaraX: The issue here is that `PrivatePageLayout` is also used in community version, the `useOptInRedirect` calls clerk `useUser()`

PR #7025, opened 2024-11-15
feat(web, dashboard, api, shared): Add enhanced `slugify` to handle multilingual, special, and emoji characters

rifont:
* Add enhanced `slugify` to handle multilingual, special, and emoji characters

### Screenshots
_Demonstration of multilingual, special, and emoji character transformation_

image

_Dashboard demo_

https://github.com/user-attachments/assets/5c4bd4e5-533c-4048-9156-5818805ced85

<details>

</details>


packages/shared/src/utils/slugify/builtinReplacements.ts
```diff
@@ -0,0 +1,2065 @@
+/* cspell:disable */
+
+export const builtinReplacements = [
```
SokratisVidros, line 3: How much does this add to the bundle size of @novu/shared? Since a lot of our libs still depend on it, is it maybe safer to add the slugify code to application-generic?

SokratisVidros (approved):

packages/shared/src/utils/slugify/builtinReplacements.ts
rifont: `slugify` is needed by the frontend, so without a `@novu/utils` distinct from a `@novu/types`, this is the only way at the moment. It's almost identical to [`slugify`](https://www.npmjs.com/package/slugify) from NPM, so should be about ~20kB unpacked.

rifont: @SokratisVidros I'm going to merge this to get the fix in ahead of tomorrow's bug bashing session. I won't release `shared` yet in case we need to take a different direction for home of `slugify`.

PR #7035, opened 2024-11-18
feat(api): add deferred action step restriction issues 

djabarovgeorge:

https://linear.app/novu/issue/NV-4739/digest-add-api-restriction-on-the-tier-plan
https://linear.app/novu/issue/NV-4736/delay-add-api-restriction-on-the-tier-plan

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/post-process-workflow-update/post-process-workflow-update.usecase.ts
```diff
@@ -82,18 +110,185 @@ export class PostProcessWorkflowUpdate {
...
+          body: this.addStepBodyIssues(step),
+          controls: await this.buildControlIssues(step, _workflowId, user),
+        };
+      })
+    );
 
     return stepIdToIssues;
   }
 
+  private async buildControlIssues(step: NotificationStepEntity, _workflowId: string, user: UserSessionData) {
```
djabarovgeorge: Currently, there's no reliable way to determine whether a control value should be validated or not.

For example, we might have unit/amount control values in the code-first value. Should we validate them the same way or not?

djabarovgeorge:
will be merged as part https://github.com/novuhq/novu/pull/7088

PR #7048, opened 2024-11-18
fix(api): payload schema parsing

ChmaraX:

handles different types of payload schema


apps/api/src/app/workflows-v2/usecases/build-test-data/build-workflow-test-data.usecase.ts
```diff
@@ -12,7 +12,7 @@ export class BuildWorkflowTestDataUseCase {
   async execute(command: WorkflowTestDataCommand): Promise<WorkflowTestDataResponseDto> {
     const _workflowEntity: NotificationTemplateEntity = await this.fetchWorkflow(command);
     const toSchema = buildToFieldSchema({ user: command.user, steps: _workflowEntity.steps });
-    const payloadSchema = JSON.parse(_workflowEntity.payloadSchema);
+    const payloadSchema = parsePayloadSchema(_workflowEntity.payloadSchema);
```
ChmaraX, line 15: since `payloadSchema` on workflow entity has `any` type, we need to make sure its handled and parsed properly

SokratisVidros (approved):

PR #7050, opened 2024-11-18
fix(dashboard): add control values to step editor in-app preview

ChmaraX:

The preview endpoint needs control values on payload. The control values are available in `StepEditorProvider`.


apps/dashboard/src/components/workflow-editor/steps/configure-step.tsx
```diff
@@ -99,4 +100,12 @@ export function ConfigureStep() {
...
-}
+};
+
+export const ConfigureStep = () => {
+  return (
+    <StepEditorProvider>
+      <ConfigureStepInternal />
+    </StepEditorProvider>
+  );
+};
```
ChmaraX, line 111: Following similar logic as with EditStepSidebar. They're mounted on different URL paths therefore we need to wrap both of them to this provider. 

```jsx
export const EditStepSidebar = () => {
  return (
    <StepEditorProvider>
      <EditStepSidebarInternal />
    </StepEditorProvider>
  );
};

```

SokratisVidros (approved):

PR #7053, opened 2024-11-18
fix(dashboard): propagate saved step changes to in-app step editor

ChmaraX:

The changes in this PR fixes the In-App Editor form dirty state which is also used to show the "unsaved changes" modal.

The strategy that we currently have:
- merge the default values from the `uiSchema` with the `values` sent from the API and set it as the initial form state
- when the form is saved, send the `values` combined with only dirty form values
- the form state is sent as it is, meaning if `subject: ""` then it will be sent like that to the API, similar applies to the other fields as well

### Screenshots

https://github.com/user-attachments/assets/57cf56d6-4a6f-46ac-9fa3-3a5888b87c6d

https://github.com/user-attachments/assets/6e1eee20-dd0c-4630-82e4-0088af1f47c7


apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -24,22 +24,26 @@ import { useWorkflowEditorContext } from '../../hooks';
...
 export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; step: StepDataDto }) => {
+  const navigate = useNavigate();
   const { stepSlug = '', workflowSlug = '' } = useParams<{ workflowSlug: string; stepSlug: string }>();
   const { resetWorkflowForm } = useWorkflowEditorContext();
+  const { refetch: refetchStep } = useStepEditorContext();
   const { step: workflowStep } = useStep();
-  const { dataSchema, uiSchema } = step.controls;
-  const navigate = useNavigate();
+
+  const { dataSchema, uiSchema, values } = step.controls;
```
ChmaraX, line 39: form values are coming from `step` prop, which is coming from `useStepEditorContext` up the chain, which calls `GET /workflows/${workflowSlug}/steps/${stepSlug}` endpoint

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -59,6 +63,7 @@ export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; s
   const { isPending, updateWorkflow } = useUpdateWorkflow({
     onSuccess: (data) => {
       resetWorkflowForm(data);
+      refetchStep();
```
ChmaraX, line 66: When the new form values are saved, we execute `PUT` request to save the workflow (incl. new control values), but the response (updated workflow) doesn't contain control values in DTO - we need to call refetch on step to get the newly updated control values, so the form updates.

I'm not sure if I like this approach, maybe we should include the control values in workflow DTO and this logic wouldn't be needed.

apps/dashboard/package.json
```diff
@@ -65,6 +65,7 @@
     "flat": "^6.0.1",
     "framer-motion": "^11.3.19",
     "lodash.debounce": "^4.0.8",
+    "lodash.merge": "^4.6.2",
```
LetItRock, line 68: used to merge the default values with the control values for the form initial state

apps/dashboard/src/components/workflow-editor/steps/edit-step-sidebar.tsx
```diff
@@ -22,12 +22,12 @@ const transitionSetting = { ease: [0.29, 0.83, 0.57, 0.99], duration: 0.4 };
...
   const navigate = useNavigate();
   const { workflow, isPendingWorkflow } = useWorkflowEditorContext();
-  const { step, stepType, isPendingStep, isRefetchingStep } = useStepEditorContext();
+  const { step, stepType, isPendingStep } = useStepEditorContext();
   const handleCloseSidebar = () => {
     navigate('..', { relative: 'path' });
   };
 
-  const isPending = isPendingWorkflow || isPendingStep || isRefetchingStep;
+  const isPending = isPendingWorkflow || isPendingStep;
```
LetItRock, line 30: no need to show the loader when the step is refetched

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-action.tsx
```diff
@@ -93,15 +94,15 @@ export const InAppAction = () => {
           </DropdownMenuItem>
           <DropdownMenuItem
             onClick={() => {
-              setValue(
-                primaryActionKey,
+              const primaryActionValue = merge(
```
LetItRock, line 98: when the actions are picked from the dropdown, preserve the previous state, for example: the form is initialized with the `primaryAction` then the user has picked `primaryAction + secondaryAction` we need to preserve the state of the `primaryAction` with what was before

LetItRock: there is a video attached that shows this case

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-action.tsx
```diff
@@ -154,7 +155,7 @@ const ConfigureActionPopover = (props: ComponentProps<typeof PopoverTrigger> & {
   return (
     <Popover modal={true}>
       <PopoverTrigger {...rest} />
-      <PopoverContent className="max-w-72">
+      <PopoverContent className="max-w-72" side="bottom" align="end">
```
LetItRock, line 159: the actions popover form is aligned with the end of the action button. Otherwise, when text is typed, it will jump

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -24,33 +25,36 @@ import { useWorkflowEditorContext } from '../../hooks';
...
   const { resetWorkflowForm } = useWorkflowEditorContext();
+  const { refetch: refetchStep } = useStepEditorContext();
   const { step: workflowStep } = useStep();
-  const { dataSchema, uiSchema } = step.controls;
-  const navigate = useNavigate();
-  const schema = buildDynamicZodSchema(dataSchema ?? {});
+
+  const { dataSchema, uiSchema, values } = step.controls;
+  const schema = useMemo(() => buildDynamicZodSchema(dataSchema ?? {}), [dataSchema]);
+  const newFormValues = useMemo(() => merge(buildDefaultValues(uiSchema ?? {}), values), [uiSchema, values]);
```
LetItRock, line 41: merge the default values with the control values to set the form values initial state

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
LetItRock: This is fine for now, we will later use the `PATCH /workflows/:id/steps/:id` endpoint that should solve this problem.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -93,16 +98,29 @@ export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; s
...
     await updateWorkflow({
       id: workflow._id,
       workflow: {
         ...workflow,
         steps: workflow.steps.map((step) =>
-          step.slug === stepSlug ? ({ ...step, controlValues: { ...data }, issues: undefined } as StepUpdateDto) : step
+          step.slug === stepSlug
+            ? ({
+                ...step,
+                controlValues: { ...values, ...updatedValues },
```
LetItRock, line 117: merge the control values with only the form dirty fields and save it

apps/dashboard/src/utils/schema.ts
```diff
@@ -86,8 +87,7 @@ const handleStringType = ({
     stringValue = stringValue.default(defaultValue as string);
   }
 
-  // remove empty strings
-  return stringValue.transform((val) => (val === '' ? undefined : val));
```
LetItRock, line 90: no need to strip the form empty string values

apps/dashboard/src/utils/schema.ts
```diff
@@ -153,10 +154,14 @@ export const buildDefaultValues = (uiSchema: UiSchema): object => {
...
 
     const { placeholder: defaultValue } = property;
-    if (defaultValue === null || typeof defaultValue === 'undefined') {
+    if (typeof defaultValue === 'undefined') {
       return acc;
     }
 
+    if (defaultValue === null) {
+      return { ...acc, [key]: defaultValue };
+    }
```
LetItRock, line 163: when `null` is sent as the default value, just set it on the form (it supports null but not undefined)

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -93,16 +98,29 @@ export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; s
   });
 
   const onSubmit = async (data: any) => {
+    const updatedValues = Object.keys(formState.dirtyFields).reduce(
```
SokratisVidros, line 101: Nice

SokratisVidros (approved):

PR #7054, opened 2024-11-18
fix(dashboard): Fix autosuggest UI issues

desiprisg:
Use the `autocomplete` extension and only offer suggestion after `{{`. No HTML and no liquid operator suggestions.

### Screenshots

https://github.com/user-attachments/assets/84d78cbf-bb43-4e08-9dc1-72d2683413c0

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/url-input.tsx
+++ b/apps/dashboard/src/components/workflow-editor/url-input.tsx
@@ -1,10 +1,12 @@
-import { liquid } from '@codemirror/lang-liquid';
 import { useFormContext } from 'react-hook-form';
 
 import { Editor } from '@/components/primitives/editor';
 import { FormControl, FormField, FormItem, FormMessagePure } from '@/components/primitives/form/form';
 import { Input, InputField, InputFieldProps, InputProps } from '@/components/primitives/input';
 import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/primitives/select';
+import { completions } from '@/utils/liquid-autocomplete';
+import { LiquidVariable } from '@/utils/parseStepVariablesToLiquidVariables';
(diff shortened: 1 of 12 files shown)


apps/dashboard/src/utils/liquid-autocomplete.ts
```diff
@@ -0,0 +1,39 @@
...
+        afterCursor.includes('}}') &&
+        beforeCursor.lastIndexOf('{{') > beforeCursor.lastIndexOf('}}'));
+
+    if (!isInsideOrAfterBraces) {
+      return null;
+    }
+
+    const word = context.matchBefore(/[\w.]+/);
+    if (!word?.text && beforeCursor.endsWith('{{')) {
+      console.log('show');
```
SokratisVidros: Please remove

apps/dashboard/src/components/workflow-editor/url-input.tsx
```diff
@@ -46,11 +48,7 @@ export const URLInput = ({
                       <Editor
                         fontFamily="inherit"
                         placeholder={placeholder}
-                        extensions={[
-                          liquid({
-                            variables,
-                          }),
-                        ]}
+                        extensions={[autocompletion({ override: [completions(variables)] })]}
```
SokratisVidros, line 51: Since this is the URL input, we shouldn't autocomplete any Liquid filters as they don't make sense in the context of a url-input. Autocomplete for this control should only include, payload, subscriber and previous step values only if there is a previous step (this should be handled by the server).

SokratisVidros (changes requested):

SokratisVidros (approved):

LetItRock (commented):
you updated the `.source` file

PR #7056, opened 2024-11-18
fix(api): make zod optional

tatarco:
zod is failing preview calls

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
+++ b/apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
@@ -61,18 +61,18 @@ const InAppRenderOutputSchema = z.object({
   avatar: z.string().optional(),
   primaryAction: z
     .object({
-      label: z.string(),
+      label: z.string().optional(),
       redirect: z.object({
-        url: z.string(),
+        url: z.string().optional(),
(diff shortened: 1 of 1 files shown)


SokratisVidros (approved):

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
```diff
@@ -61,26 +61,26 @@ const InAppRenderOutputSchema = z.object({
   avatar: z.string().optional(),
   primaryAction: z
     .object({
-      label: z.string(),
+      label: z.string().optional(),
```
rifont, line 64: I don't think this is the right approach, it's invalid to have a primary action without a label, etc. The correct fix IMO is to not send an empty object for `primaryAction` when the primary action isn't needed.

SokratisVidros: Are you talking about not sending an empty object from the FE to the API or preprocess the control values [here](https://github.com/novuhq/novu/blob/22122b7be478187f9c4b4edf8a8644c1ea174288/apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts#L10)?

PR #7063, opened 2024-11-19
fix(api): Nv 4798 unify default step controls definition using only zod

tatarco:

1. Change Control schemas to use Zod as source of truth.
2. Use zod to infer the typescript type within renderers
3. Add digest ZOD & data schema.
4. Add Digest renderer.
5. Add Dealy renderer.
6. Add Delate ZOD
7. Add Delay data schema

### Screenshots

<details>

</details>


apps/api/src/app/environments-v1/usecases/output-renderers/digest-output-renderer.usecase.ts
```diff
@@ -0,0 +1,46 @@
...
+        },
+      };
+    }
+    if (isDigestTimedControl(parse) && parse.cron) {
+      return {
+        cron: parse.cron,
+        digestKey: parse.digestKey,
+      };
+    }
+    throw new DigestRenderProblem(renderCommand.controlValues);
```
SokratisVidros:
```suggestion
    throw new ≈(renderCommand.controlValues);
```

Or just reuse one of the existing exceptions. I don't think we need a specific one only for digest.

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
```diff
@@ -1,87 +1,57 @@
...
-  redirect: z
-    .object({
-      url: z.string().optional(),
-      target: RedirectTargetEnumSchema.optional(), // Optional target
-    })
-    .optional(),
-});
+  private buildRedirect(redirect?: InAppRedirectType) {
+    if (!(redirect && redirect.url)) {
+      return undefined;
```
SokratisVidros, line 35: Is the default undefined or null? I think the FE expected null but I might be wrong.

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
```diff
@@ -1,87 +1,57 @@
...
+
+    return {
+      url: redirect.url,
+      target: redirect.target as RedirectTargetEnum,
+    };
+  }
+
+  private buildActionIfAllPartsAvailable(primaryAction?: InAppActionType) {
+    if (!(primaryAction && primaryAction.label && primaryAction.redirect && primaryAction.redirect.url)) {
+      return undefined;
```
SokratisVidros, line 46: Ditto about null vs undefined

SokratisVidros (approved):

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
tatarco: undefined, it's not a default, it's the outcome of the render.

apps/api/src/app/environments-v1/usecases/output-renderers/in-app-output-renderer.usecase.ts
tatarco: same here, it's not the default, it's the render outcome

PR #7064, opened 2024-11-19
fix(dashboard): workflow editor name change updates the workflow slug in the url

LetItRock:

This PR fixes the case when the workflow name is updated in the Workflow Editor we generate a new slug for the workflow on the BE, which should also update the slug used in the URL to make it a source of truth. 

All the `useFetchWorkflow, useFetchStep, useUpdateWorkflow` hook should use the `workflow id` from the `slug` as the key for the react-query state.

### Screenshots

https://github.com/user-attachments/assets/c696d5d6-9c05-4547-8c82-d6b727869ca5

https://github.com/user-attachments/assets/ff80cb4a-fbd0-48b1-8f9d-37d0cf3492ca


apps/dashboard/src/components/workflow-editor/configure-workflow.tsx
```diff
@@ -100,7 +129,13 @@ export function ConfigureWorkflow() {
                 <FormLabel>Workflow Name</FormLabel>
                 <FormControl>
                   <InputField>
-                    <Input placeholder="Untitled" {...field} disabled={isReadOnly} />
+                    <Input
+                      placeholder="Untitled"
```
SokratisVidros:
```suggestion
                      placeholder="New workflow"
```

apps/dashboard/src/utils/step.ts
```diff
@@ -1,8 +1,9 @@
-import { ShortIsPrefixEnum, StepResponseDto } from '@novu/shared';
+import { ShortIsPrefixEnum } from '@novu/shared';
 
-const divider = `_${ShortIsPrefixEnum.STEP}`;
+export const WORKFLOW_DIVIDER = `_${ShortIsPrefixEnum.WORKFLOW}`;
+export const STEP_DIVIDER = `_${ShortIsPrefixEnum.STEP}`;
 
-export const getStepBase62Id = (slug: StepResponseDto['slug'] | string = divider) => {
+export const getBase62Id = ({ slug, divider }: { slug: string; divider: string }) => {
```
SokratisVidros: We will soon update the base62 to a shorter version. So let's keep FE agnostic of that change.

```suggestion
export const getEncodedId = ({ slug, divider }: { slug: string; divider: string }) => {
```

SokratisVidros (approved):

PR #7066, opened 2024-11-19
feat(dashboard): make unsupported steps read-only for all wfs

ChmaraX:

https://github.com/user-attachments/assets/cc67ce49-ffca-466c-aa81-c7de2cfca82e

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/steps/common-fields.tsx
+++ b/apps/dashboard/src/components/workflow-editor/steps/common-fields.tsx
@@ -6,13 +6,16 @@ import { FormControl, FormField, FormItem, FormLabel, FormMessage } from '../../
 import { Input, InputField } from '../../primitives/input';
 import { useStep } from './use-step';
 import { buildRoute, ROUTES } from '@/utils/routes';
+import { EXCLUDED_EDITOR_TYPES } from '@/utils/constants';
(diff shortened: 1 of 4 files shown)


apps/dashboard/src/components/workflow-editor/steps/common-fields.tsx
```diff
@@ -6,14 +6,17 @@ import { FormControl, FormField, FormItem, FormLabel, FormMessage } from '../../
...
 
 export function CommonFields() {
   const { stepIndex, control, step } = useStep();
   const navigate = useNavigate();
   const { stepSlug } = useParams<{ stepSlug: string }>();
-  const { isReadOnly } = useWorkflowEditorContext();
+  const { isReadOnly: isWorkflowReadOnly } = useWorkflowEditorContext();
   const [isBlurred, setIsBlurred] = useState(false);
 
+  const isReadOnly = isWorkflowReadOnly || EXCLUDED_EDITOR_TYPES.includes(step?.type ?? '');
```
ChmaraX, line 18: I think its more UX friendly to show the step read-only info in the sidebar (name, identifier) for the dashboard first workflows, rather than make it not clickable at all.

SokratisVidros (approved):

PR #7068, opened 2024-11-19
fix(dashboard): Make preferences.default a nullable field

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/schema.ts
+++ b/apps/dashboard/src/components/workflow-editor/schema.ts
@@ -1,7 +1,7 @@
-import * as z from 'zod';
-import type { JSONSchemaDefinition } from '@novu/shared';
 import { StepTypeEnum } from '@/utils/enums';
 import { capitalize } from '@/utils/string';
+import type { JSONSchemaDefinition } from '@novu/shared';
+import * as z from 'zod';
 
 const enabledSchema = z.object({
   enabled: z.boolean(),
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/components/workflow-editor/schema.ts
```diff
@@ -61,13 +61,15 @@ export const workflowSchema = workflowMinimalSchema.extend({
...
-    }),
+    default: z
+      .object({
+        all: z.object({
+          enabled: z.boolean(),
+          readOnly: z.boolean(),
+        }),
+        channels: channelsSchema,
+      })
+      .nullable(),
```
rifont, line 72: The `default` preferences isn't nullable, it should always be present. The `default` preferences for Dashboard-sourced workflows will always be the `DEFAULT_WORKFLOW_PREFERENCES`, whereas the `default` preferences for Framework-sourced workflows are that defined in code.

What issue is this solving?

rifont (changes requested):

apps/dashboard/src/components/workflow-editor/schema.ts
rifont: If no `default` preferences are supplied, the `UpsertWorkflow` use-case falls back to `DEFAULT_WORKFLOW_PREFERENCES` [here](https://github.com/novuhq/novu/blob/next/apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts#L142). So instead, we should not send `default`.

rifont:
With the current API validation and implementation, there are no cases where `default` preferences will be persisted as `null`. The frontend always calls the `PUT /v2/workflows/{id}` endpoint with the object retrieved from `GET /v2/workflows{id}`, so we can safely close this PR.

Any cases where the `default: null` issue arose are from a previous broken implementation during development of Dashboard V2, and should be considered "corrupt".

PR #7070, opened 2024-11-19
feat(dashboard): signup and login page design update

scopsy:
- Updated design for sign in and sign up pages
- Data Residency region picker
- Updated Auth Layout

### Screenshots

CleanShot 2024-11-20 at 11 47 40@2x
CleanShot 2024-11-20 at 11 47 54@2x
CleanShot 2024-11-20 at 11 48 02@2x

<details>

</details>


apps/dashboard/src/components/icons/circle-check.tsx
```diff
@@ -1,11 +1,11 @@
 import type { HTMLAttributes } from 'react';
 
-export const CircleCheck = (props: HTMLAttributes<HTMLOrSVGElement>) => {
+export const CircleCheck = (props: HTMLAttributes<HTMLOrSVGElement> & { color?: string }) => {
   return (
     <svg xmlns="http://www.w3.org/2000/svg" width="28" height="28" viewBox="0 0 17 16" fill="none" {...props}>
       <path
         d="M6.50004 7.99999L7.83337 9.33333L10.5 6.66666M15.1667 7.99999C15.1667 11.6819 12.1819 14.6667 8.50004 14.6667C4.81814 14.6667 1.83337 11.6819 1.83337 7.99999C1.83337 4.3181 4.81814 1.33333 8.50004 1.33333C12.1819 1.33333 15.1667 4.3181 15.1667 7.99999Z"
-        stroke="#1FC16B"
+        stroke={props.color ?? '#1FC16B'}
```
scopsy, line 8: Added the abilty to update the color from the default green

apps/dashboard/src/components/auth/auth-feature-row.tsx
```diff
@@ -0,0 +1,19 @@
...
+  description: string;
+}
+
+export function AuthFeatureRow({ icon, title, description }: AuthFeatureRowProps) {
+  return (
+    <div className="inline-flex items-start justify-start gap-3.5">
+      <div className="flex items-center justify-center p-1">{icon}</div>
+      <div className="inline-flex shrink grow basis-0 flex-col items-start justify-start gap-2">
+        <div className="text-sm font-medium text-[#0d111b]">{title}</div>
+        <div className="text-xs font-medium text-[#99a0ad]">{description}</div>
```
scopsy: @LetItRock what is the suggested way of working with brand colors and typography? do we have some class names for those?

apps/dashboard/src/components/auth/auth-side-banner.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Plug } from '../icons/plug';
+import { Sparkling } from '../icons/sparkling';
+import { AuthFeatureRow } from './auth-feature-row';
+import { TrustedCompanies } from './trusted-companies';
+
+export function AuthSideBanner() {
+  return (
+    <div className="inline-flex w-full max-w-[476px] flex-col items-center justify-center gap-[50px] p-5">
+      <div className="flex flex-col items-start justify-start gap-4">
+        <div className="inline-flex items-center justify-start gap-3">
```
scopsy, line 12: Is there a recommended way of working with brand spacing? @LetItRock

apps/dashboard/src/components/auth/auth-feature-row.tsx
rifont: Brand colors should be defined in HSL format in the [index.css](https://github.com/novuhq/novu/blob/next/apps/dashboard/src/index.css#L52) file, they should then be made available to Tailwind by adding an entry referencing the variable definition in the [tailwind.config.js](https://github.com/novuhq/novu/blob/next/apps/dashboard/tailwind.config.js#L65) file.

Then you can simply use `text-{var-name}` to use the color

apps/dashboard/src/components/auth/auth-side-banner.tsx
rifont: It's strongly preferred to use built-in Tailwind classes as you've already done, but it's also acceptable to use the `-[{whatever}]` escape hatch syntax in cases where it makes sense to provide a better UX + look/feel

apps/dashboard/src/components/auth/region-picker.tsx
```diff
@@ -0,0 +1,75 @@
...
+            </TooltipTrigger>
+            <TooltipContent>
+              Novu offers data residency in Europe (Germany) and United States. Account residency couldn't be modified
+              after sign-up.
+            </TooltipContent>
+          </Tooltip>
+        </TooltipProvider>
+      </div>
+      <Select value={selectedRegion} onValueChange={handleRegionChange}>
+        <SelectTrigger className="h-[22px] w-[64px] p-[4px] pl-[6px] text-[10px] leading-[14px]">
```
rifont: Tailwind operates with multiples of 4 for all spacing properties, so here is an example of what you would use here:
```suggestion
        <SelectTrigger className="h-[22px] w-16 p-1 pl-1.5 text-[10px] leading-[14px]">
```

Futhermore, you should always try the builtin default multiples first to see if they feel appropriate. This ensures that the base variable spacing values can be modified and the rest of the application updates dynamically with them.

apps/dashboard/src/components/auth/region-picker.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <div className="inline-flex w-full items-center justify-center gap-[6px]">
+      <div className="text-xs font-medium leading-none text-[#99a0ad]">
+        Data Residency
+        <TooltipProvider delayDuration={100}>
+          <Tooltip>
+            <TooltipTrigger className="ml-1">
+              <BsFillInfoCircleFill className="text-foreground-300 -mt-0.5 inline size-3" />
+            </TooltipTrigger>
+            <TooltipContent>
+              Novu offers data residency in Europe (Germany) and United States. Account residency couldn't be modified
```
rifont: Using correct grammar

```suggestion
              Novu offers data residency in Europe (Germany) and the United States. Data residency cannot be modified
```

apps/dashboard/src/components/auth/region-picker.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <SelectTrigger className="h-[22px] w-[64px] p-[4px] pl-[6px] text-[10px] leading-[14px]">
+          <SelectValue placeholder="Select a country" />
+        </SelectTrigger>
+        <SelectContent>
+          {Object.values(REGION_MAP).map((option) => (
+            <SelectItem key={option} value={option} className="w-[100px] w-full">
+              <div className="flex items-center gap-[6px]">
+                {' '}
+                {option === REGION_MAP.US ? (
+                  <USFlag className="h-[10px] w-[10px]" />
```
rifont: I would try either `h-2` or `h-3` first to see if if feels appropriately sized

apps/dashboard/src/components/icons/circle-check.tsx
rifont: Variant styles should be applied via Tailwind classes on the `className` prop only, not via new props. So you can simply pass something like: `<CircleCheck className="stroke-accent" />`, or replace `accent` with an override using `stroke-[red]` or whichever color is necessary.

apps/dashboard/src/utils/clerk-appearance.ts
```diff
@@ -0,0 +1,23 @@
+import { SignUpTheme, SignInTheme } from '@clerk/types';
+
+export const clerkAppearance: SignUpTheme | SignInTheme = {
```
rifont: I'd suggest a more accurate name in case further appearance props are added here

```suggestion
export const clerkSignupAppearance: SignUpTheme | SignInTheme = {
```

apps/dashboard/src/components/icons/flags/eu.tsx
rifont, line 1: chore: Please parse all SVGs through https://jakearchibald.github.io/svgomg/ to clean up redundant values and optimize size, most often we can reduce the total bytes by 50%

apps/dashboard/src/components/auth/trusted-companies.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface CompanyLogoProps {
+  name: string;
+}
+
+function CompanyLogo({ name }: CompanyLogoProps) {
+  return (
+    <div className="inline-flex flex-col items-start justify-start px-5">
+      <div className="flex flex-col items-start justify-start">
+        <div className="relative h-[20px]">
+          <img src={`/images/auth/${name}-customer.svg`} alt={name} />
```
rifont, line 39: nice use of`alt` - accessibility 😍

apps/dashboard/public/images/auth/background.svg
rifont, line 1: Beautiful!!

apps/dashboard/src/components/auth/region-picker.tsx
scopsy: 🩷

apps/dashboard/src/components/auth/region-picker.tsx
scopsy: Perfect, on it. Thank you

scopsy: In the example of h-[22px] is it ok to use h-5.5 similiar to pl-1.5?

apps/dashboard/src/components/auth/trusted-companies.tsx
djabarovgeorge: was it Dima or Cursor, we will never know ;)

apps/dashboard/src/components/icons/circle-check.tsx
scopsy: @rifont issue is that I need this to be applied to the specific path element which is a child of the SVG parent. Will check if I can somehow convert it to use nesting.

djabarovgeorge (approved):
🚀

PR #7080, opened 2024-11-20
feat(dashboard): org creation screen

scopsy:

- Updated design of the org create screen

Stacked PR with: https://github.com/novuhq/novu/pull/7070

### Screenshots
CleanShot 2024-11-20 at 13 53 05@2x

<details>

</details>


apps/dashboard/src/components/primitives/card.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+
+import { cn } from '@/utils/ui';
+
+const Card = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(({ className, ...props }, ref) => (
```
scopsy, line 5: Added a new Shadcn Card component

apps/dashboard/public/images/auth/ui-org.svg
djabarovgeorge, line 1: i wonder if it would make sense to reduce the opacity on the right side? To me, it feels like it draws too much attention at first glance.
i added a question in figma as well.

scopsy: Will check it out 💪

scopsy: Added opacity :)

apps/dashboard/src/components/auth/create-organization.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <img src="/images/auth/ui-org.svg" alt="create-org-illustration" className="opacity-70" />
+        </div>
+      </AuthCard>
+    </div>
+  );
+}
+
+function StepIndicator(): JSX.Element {
+  return (
+    <div className="inline-flex items-center gap-[2px] pl-[20px] text-[#525866]">
```
desiprisg: Doesn't this text color correspond to one from the `foreground` scale?

apps/dashboard/src/components/primitives/card.tsx
desiprisg, line 1: This won't work out of the box. You will need to replace the `card`, `card-foreground`, `muted`, `muted-foreground` colors with other ones, or add these variables to the css file. I like the former more, since we don't have any cards that reuse some kind of color.

scopsy: What will not work? It did work and rendered 😄 Can we do a quick 1-o-1 to talk this out? I also have some other questions about using colors, etc..

desiprisg: I'm saying that those css variables are not declared, and these colors are not even present in the tailwind config, so they do nothing.

desiprisg: Regarding colors in general, we are mostly using the `neutral` and `foreground` scale because there was not a lot of alignment between figma and the codebase at the beginning. As we bring them closer together, we can start adopting some new css variables, if they make sense. Generally, you can take a look at the `index.css` file for the variables declared there, as well as the tailwind.config.ts and primitives for implementation.

apps/dashboard/src/components/auth/create-organization.tsx
scopsy: Yep found it!

desiprisg (approved):

PR #7088, opened 2024-11-20
feat(worker): add defer duration validation

djabarovgeorge:

this PR will solve the tickets from https://github.com/novuhq/novu/pull/7035
https://linear.app/novu/issue/NV-4739/digest-add-api-restriction-on-the-tier-plan
https://linear.app/novu/issue/NV-4736/delay-add-api-restriction-on-the-tier-plan
This change was made because this PR have a refactoring of the logic that were made in 7035, this way we are having all of the logic in one place.

https://linear.app/novu/issue/NV-4739/digest-add-api-restriction-on-the-tier-plan
https://linear.app/novu/issue/NV-4736/delay-add-api-restriction-on-the-tier-plan

### Screenshots

<details>

</details>


apps/worker/src/app/workflow/usecases/add-job/add-job.usecase.ts
```diff
@@ -157,9 +161,41 @@ export class AddJob {
...
+        deferDuration: delay,
+        stepType: job.type,
+        organizationId: command.organizationId,
+      })
+    );
+
+    if (Object.keys(issues).length > 0) {
+      const tierIssues = (issues as Record<string, ContentIssue[]>).tier;
+      if (tierIssues) {
+        Logger.warn({ issues: tierIssues, jobId: job._id }, 'Defer duration limit exceeded for job', LOG_CONTEXT);
```
rifont: What does `Defer` mean in the context of a maximum duration limit? The naming is weird to me. I would just go with:
```suggestion
        Logger.warn({ issues: tierIssues, jobId: job._id }, 'Maximum duration limit exceeded for job', LOG_CONTEXT);
```

libs/application-generic/src/usecases/validate-control-by-tier/validate-control-by-tier.command.ts
```diff
@@ -0,0 +1,16 @@
...
+import { IsEnum, IsNumber, IsOptional } from 'class-validator';
+
+import { OrganizationLevelCommand } from '../../commands';
+
+export type Milliseconds = number;
+
+export class ValidateControlByTierCommand extends OrganizationLevelCommand {
+  @IsNumber()
+  @IsOptional()
+  deferDuration?: Milliseconds;
```
rifont: As above `defer` is a confusing new term. I would simplify to "maximum", also add `ms` to make the units clear
```suggestion
  maximumDurationMs?: Milliseconds;
```

libs/application-generic/src/usecases/validate-control-by-tier/validate-control-by-tier.command.ts
```diff
@@ -0,0 +1,16 @@
+import { StepTypeEnum } from '@novu/shared/';
+import { IsEnum, IsNumber, IsOptional } from 'class-validator';
+
+import { OrganizationLevelCommand } from '../../commands';
+
+export type Milliseconds = number;
```
rifont: Nothing is gained by creating a new type here aside from more code. If one hovers over the `deferDuration ` value, they will still see `number` in intellisense. The correct solution would be to use a Typescript brand strategy, but I think that needs wider discussion before going ahead with.

apps/worker/src/app/workflow/usecases/add-job/add-job.usecase.ts
djabarovgeorge: I think we should stick with this term since it's already used in the 'add job' use case. If needed, I can ensure it's applied consistently across the project, and i will raise it with the team.

We need a clear way to define two things in the project:  
1. The time we delay or notification, without explicitly using 'delay' or 'digest.'  
2. A term for actions that aren’t executed immediately (deferred actions).  

Having consistent terms is important because the current lack of clarity causes confusion in the code.  

I'm open to using a different term as long as we use it consistently and everyone understands it.

libs/application-generic/src/usecases/validate-control-by-tier/validate-control-by-tier.command.ts
djabarovgeorge: Let's continue the talk about the name on this [comment](https://github.com/novuhq/novu/pull/7088#discussion_r1850485669) 

I loved the Ms suffix, i will add it :)

libs/application-generic/src/usecases/validate-control-by-tier/validate-control-by-tier.command.ts
djabarovgeorge: I found that we have a lot of code that is not straightforward forward and this is why i added it, I think the Ms suffix could resolve it for now.

libs/application-generic/src/usecases/validate-control-by-tier/validate-control-by-tier.usecase.ts
djabarovgeorge, line 1: refactor the code to be more generic and not return control value issues

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
```diff
@@ -220,4 +243,70 @@ export class PrepareAndValidateContentUsecase {
...
+  private async computeTierIssues(
+    defaultControlValues: Record<string, unknown>,
+    user: UserSessionData,
+    stepType?: StepTypeEnum
+  ): Promise<Record<string, ContentIssue[]>> {
+    const deferDuration =
+      isValidDigestUnit(defaultControlValues.unit) && isNumber(defaultControlValues.amount)
+        ? calculateMilliseconds(defaultControlValues.amount, defaultControlValues.unit)
+        : 0;
+
```
tatarco, line 350: don't you want this calculation inside the usecase, i feel you will be doing it again in the worker

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
```diff
@@ -220,4 +243,70 @@ export class PrepareAndValidateContentUsecase {
...
+          message: restrictionsError.message,
+        },
+      ];
+      result.unit = [
+        {
+          issueType: StepContentIssueEnum.TIER_LIMIT_EXCEEDED,
+          message: restrictionsError.message,
+        },
+      ];
+    }
```
tatarco, line 375: is the text what we want inside the usecase? maybe keep the usecase result structured, like max_allowed amount , max_allowed_unit and a pass/ fail for it, I would also maybe encapsulate it so it support move validations of similar nature in the future, so like have a nestring with an enum called DEFFER_DURATION

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
```diff
@@ -220,4 +243,70 @@ export class PrepareAndValidateContentUsecase {
...
+      return 0;
+  }
+}
+
+function isValidDigestUnit(unit: unknown): unit is DigestUnitEnum {
+  return Object.values(DigestUnitEnum).includes(unit as DigestUnitEnum);
+}
+
+function isNumber(value: unknown): value is number {
+  return !Number.isNaN(Number(value));
```
tatarco, line 405: would expect it inside the use case

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
djabarovgeorge: worker is a more complex place, you are calculating it separately today for delay or digest, i wouldn't touch this part as part of this PR.

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -199,8 +203,31 @@ describe('Workflow Controller E2E API Testing', () => {
...
+          unit: [
+            {
+              issueType: 'TIER_LIMIT_EXCEEDED',
+              message:
+                'The maximum delay allowed is 90 days.Please contact our support team to discuss extending this limit for your use case.',
+            },
+          ],
+        });
+      });
     });
```
tatarco: I think the msgs are too specific, every basic change to the script will cause tests to fail, maybe just validate the amount.issue_type and unit.issue_type

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -1062,10 +1092,11 @@ function buildEmailStep(): StepCreateDto {
     type: StepTypeEnum.EMAIL,
   };
 }
-function buildDigestStep(): StepCreateDto {
+function buildDigestStep(overrides: Partial<StepCreateDto> = {}): StepCreateDto {
   return {
     name: 'Digest Test Step',
     type: StepTypeEnum.DIGEST,
+    ...overrides,
```
tatarco, line 1090: what would you add here?

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -1062,10 +1092,11 @@ function buildEmailStep(): StepCreateDto {
     type: StepTypeEnum.EMAIL,
   };
 }
-function buildDigestStep(): StepCreateDto {
+function buildDigestStep(overrides: Partial<StepCreateDto> = {}): StepCreateDto {
```
tatarco, line 1086: control values should not be updated from here anymore, we have a patch step, use it, I'm going to delete this functionality soon

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
djabarovgeorge: we can encapsulate and refactor it to support other cases when we will have them, for not it will make the code more complex with no reason.

apps/worker/src/app/workflow/usecases/add-job/add-job.usecase.ts
```diff
@@ -157,9 +161,38 @@ export class AddJob {
...
+          ...ExecutionLogRouteCommand.getDetailsFromJob(job),
+          detail: DetailEnum.DEFER_DURATION_LIMIT_EXCEEDED,
+          source: ExecutionDetailsSourceEnum.INTERNAL,
+          status: ExecutionDetailsStatusEnum.FAILED,
+          isTest: false,
+          isRetry: false,
+        })
+      );
+
+      throw new Error(DetailEnum.DEFER_DURATION_LIMIT_EXCEEDED);
```
tatarco, line 191: isn't this lacking some info, why, by how much and so on?

tatarco: maybe a specific error class for it

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
```diff
@@ -39,9 +39,10 @@ export class ComputeJobWaitDurationService {
       const delay = differenceInMilliseconds(new Date(delayDate), new Date());
 
       if (delay < 0) {
-        throw new ApiException(
-          `Delay date at path ${delayPath} must be a future date`,
-        );
+        throw new ApiException({
+          message: `Delay date at path must be a future date`,
```
tatarco, line 43: does it even go to API? isn't it a worker class?

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
djabarovgeorge: not at the current use case structure. at the moment the use case expects the computed amount, and then it applies the product restriction on top of the data.

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.command.ts
```diff
@@ -0,0 +1,14 @@
+import { IsEnum, IsNumber, IsOptional } from 'class-validator';
+
+import { StepTypeEnum } from '@novu/shared';
+import { OrganizationLevelCommand } from '../../commands';
+
+export class TierRestrictionsValidatorCommand extends OrganizationLevelCommand {
```
tatarco: shouldn't usecase describe an action and a verb

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
djabarovgeorge: sure, we can test issueType instead of all of the object.

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.consts.ts
```diff
@@ -0,0 +1,7 @@
+export const MILLISECONDS_IN_DAY = 24 * 60 * 60 * 1000;
+export const FREE_TIER_MAX_DELAY_DAYS = 30;
+export const BUSINESS_TIER_MAX_DELAY_DAYS = 90;
+export const MAX_DELAY_FREE_TIER =
+  FREE_TIER_MAX_DELAY_DAYS * MILLISECONDS_IN_DAY; // 30 days in milliseconds
+export const MAX_DELAY_BUSINESS_TIER =
+  BUSINESS_TIER_MAX_DELAY_DAYS * MILLISECONDS_IN_DAY; // 90 days in milliseconds
```
tatarco: is it being reused? I feel it should just be inside the usecase

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
djabarovgeorge: controlValue

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.response.ts
```diff
@@ -0,0 +1,10 @@
+export enum ErrorEnum {
+  TIER_LIMIT_EXCEEDED = 'TIER_LIMIT_EXCEEDED',
+}
+
+type TierValidationError = {
+  error: ErrorEnum;
+  message: string;
+};
+
+export type TierRestrictionsValidatorResponse = TierValidationError[] | null;
```
tatarco: maybe the response here should be a little more future looking, allowing multiple types of validation? of that's what the array is for, usually i find record better here to segment different errors

apps/worker/src/app/workflow/usecases/add-job/add-job.usecase.ts
djabarovgeorge: this is a "safe-guard" error throw as the executionLogRoute will throw the same error inside of it. meaning we will never get to this line.

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    if (tier === ApiServiceLevelEnum.FREE) {
+      if (deferDurationMs > MAX_DELAY_FREE_TIER) {
+        issues.push({
+          error: ErrorEnum.TIER_LIMIT_EXCEEDED,
+          message:
+            `The maximum delay allowed is ${FREE_TIER_MAX_DELAY_DAYS} days.` +
+            'Please contact our support team to discuss extending this limit for your use case.',
+        });
+      }
```
tatarco: I would make something more structured here, the issue passed: true/ false and some metadata to build the msg, I wouldn't contain the msgs here.

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.usecase.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const issues: TierRestrictionsValidatorResponse = [];
+    const organization =
+      await this.organizationRepository.findById(organizationId);
+    const tier = organization?.apiServiceLevel;
+
+    if (
+      tier === undefined ||
+      tier === ApiServiceLevelEnum.BUSINESS ||
+      tier === ApiServiceLevelEnum.ENTERPRISE
```
tatarco: maybe encapsulate it to make it easier to read

libs/application-generic/src/services/calculate-delay/compute-job-wait-duration.service.ts
djabarovgeorge: this is a valid point, it is a known issue that we still have not solved.

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.usecase.ts
tatarco: I feel that use the if statments to define the max values, than validate against the comuted value and return the pass / fail, max value , max unit would be a cleaner approach

packages/shared/src/dto/workflows/step-content-issue.enum.ts
```diff
@@ -2,7 +2,9 @@ export enum StepContentIssueEnum {
   MISSING_VARIABLE_IN_PAYLOAD = 'MISSING_VARIABLE_IN_PAYLOAD',
   ILLEGAL_VARIABLE_IN_CONTROL_VALUE = 'ILLEGAL_VARIABLE_IN_CONTROL_VALUE',
   MISSING_VALUE = 'MISSING_VALUE',
+  TIER_LIMIT_EXCEEDED = 'TIER_LIMIT_EXCEEDED',
```
tatarco, line 6: is this the correct word? will it be true to other things as well? or is it specific to the world of delay & digest

tatarco (approved):

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
tatarco: is it, wouldn't it be better moving the worker logic inside and both be using it?

tatarco: however you feel

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
tatarco: my issue is that the usecase is not being really shared, and is just thorwing an exception on the worker no? no point for all the text there

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.command.ts
djabarovgeorge: good point i will update it

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.consts.ts
djabarovgeorge: felt like it add to much to the use-case, it is used i the usecase.
it lives in the usecase namespace so it is ok that it is not reused outside of it.

apps/worker/src/app/workflow/usecases/add-job/add-job.usecase.ts
tatarco: I would make it more clear that if we got here we have a bug, so it will pop up in someones eyes if this is ever called, better that the current one

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.response.ts
djabarovgeorge: id refactor it once we will have more issues in this use case.

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
tatarco: should not be used, use patch instead

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
tatarco: can't we change it? there is a worker method computing this also, lets merge this logic

libs/application-generic/src/usecases/tier-restrictions-validator/tier-restrictions-validator.usecase.ts
djabarovgeorge: not sure i am following your thought here, can you elaborate?

packages/shared/src/dto/workflows/step-content-issue.enum.ts
djabarovgeorge: i belive we could reuse it other aspects as well, 
```
          body: [
            {
              issueType: 'TIER_LIMIT_EXCEEDED',
              message:
                'The maximum body length allowed is 280 characters. Please contact our support team to discuss extending this limit for your use case.',
            },
          ]
```

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
djabarovgeorge: worker throwing an exception if there is an error, not the new usecase.

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
djabarovgeorge: as i said worker logic is more complex and fragile i would not touch it as part of this work.

PR #7090, opened 2024-11-20
fix(dashboard): light theme for new domain

ChmaraX:
### Screenshots

<details>

</details>


apps/dashboard/src/context/opt-in-provider.tsx
```diff
@@ -10,6 +10,9 @@ export const OptInProvider = (props: PropsWithChildren) => {
     if (isLoaded && status !== NewDashboardOptInStatusEnum.OPTED_IN) {
       redirectToLegacyDashboard();
     }
+
+    // set light theme on the new domain for both legacy and new dashboard
+    localStorage.setItem('mantine-theme', 'light');
```
ChmaraX, line 15: dashboard-v2 and dashboard have a different localStorage, we need to set it in both places

or alternatively just here, and the legacy domain can keep the theme unchanged, but we are not yet sure if we keep 2 apps under the same domain or user will switch between 2 different ones, so I will keep it for now

SokratisVidros (approved):

PR #7101, opened 2024-11-21
revert(api): Return all preference channels for Subscriber Preference API

rifont:
* Return all preference channels for Subscriber Preference API
  * A recent change to preference retrieval across Subscriber, Inbox, and Widget API started returning only the active workflow channels for `POST`, `PUT`, and `GET` endpoints. This was a breaking change and resulted in some users experiencing broken integrations as the Preferences had been retrieved for external consumption. This change reverts the behaviour only on Subscriber Preference endpoints, Inbox & Widget continue to return only active channel preferences.
  * This change additionally adds a `includeInactiveChannels` query parameter for the Subscriber Preference endpoints to control whether to return only active channels. The flag is `true` by default, meaning both active + inactive preferences are returned, restoring functionality to the previous behaviour

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/subscribers/e2e/get-preferences.e2e.ts
+++ b/apps/api/src/app/subscribers/e2e/get-preferences.e2e.ts
@@ -2,9 +2,10 @@ import { UserSession } from '@novu/testing';
 import { expect } from 'chai';
 import { NotificationTemplateEntity } from '@novu/dal';
 
-import { getPreference } from './helpers';
+import { PreferenceLevelEnum } from '@novu/shared';
+import { getPreference, getPreferenceByLevel } from './helpers';
 
-describe('Get Subscribers preferences - /subscribers/preferences/:subscriberId (GET)', function () {
+describe('Get Subscribers workflow preferences - /subscribers/:subscriberId/preferences (GET)', function () {
   let session: UserSession;
   let template: NotificationTemplateEntity;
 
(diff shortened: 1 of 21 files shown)


apps/api/src/app/subscribers/e2e/get-preferences.e2e.ts
```diff
@@ -16,13 +17,44 @@ describe('Get Subscribers preferences - /subscribers/preferences/:subscriberId (
...
+    });
+  });
+
+  it('should get subscriber workflow preferences with inactive channels when includeInactiveChannels is true', async function () {
+    const response = await getPreference(session, session.subscriberId, {
+      includeInactiveChannels: true,
+    });
+    const data = response.data.data[0];
+
+    expect(data.preference.channels).to.deep.equal({
```
SokratisVidros, line 39: Shouldn't we see a difference in the returned channels between this test and the previous one since the `includeInactiveChannels` flag is included?

SokratisVidros (approved):

apps/api/src/app/subscribers/e2e/get-preferences.e2e.ts
rifont: Yes we do expect different behaviour, but for the following test instead. The default flag value is `true`, so these first two tests are expected to have the same behaviour.

See [this test below](https://github.com/novuhq/novu/pull/7101/files/c2001924b30d04facd0346e79108aad86be446c6#diff-c41eb3bd24095c41b35226dc58ce7c641d6f3c25d679c92dd151952c5ff03c30R48) for `includeInactiveChannels: false` which has some channels omitted.

PR #7102, opened 2024-11-21
fix(dashboard,web): opt-in page flickering; login/logout redirect

ChmaraX:

* fixes a temporary flick of legacy workflows page when redirecting to new dashboard
* redirect for opt-in users will happen on any domain where the legacy web is deployed

Fixes: NV-4838, NV-4825

### DIFF
--- a/apps/web/src/hooks/useOptInRedirect.ts
+++ b/apps/web/src/hooks/useOptInRedirect.ts
@@ -1,4 +1,4 @@
-import { useLayoutEffect } from 'react';
+import { useCallback, useLayoutEffect } from 'react';
 import { useLocation } from 'react-router-dom';
 import { FeatureFlagsKeysEnum, NewDashboardOptInStatusEnum } from '@novu/shared';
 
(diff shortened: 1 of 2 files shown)


apps/web/src/hooks/useOptInRedirect.ts
```diff
@@ -13,29 +13,26 @@ export const useOptInRedirect = () => {
   const { pathname } = useLocation();
   const isNewDashboardEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_NEW_DASHBOARD_ENABLED);
   const { status, isLoaded, redirectToNewDashboard } = useNewDashboardOptIn();
-  const isUserOptedIn = isLoaded && status && status === NewDashboardOptInStatusEnum.OPTED_IN;
-  const dashboardV2HostName = window.location.hostname.includes('dashboard-v2');
```
ChmaraX, line 17: this condition is not needed - we should redirect to new dashboard when legacy web is deployed on both `dashboard-v2` and `dashboard` domains

SokratisVidros (approved):

PR #7106, opened 2024-11-21
fix(api): Nv 4836 v2 dashboard workflows show error in old dashboard

tatarco:
make the v1 list not return v2 gui workflows. 

### Screenshots

<details>

</details>


libs/dal/src/repositories/notification-template/notification-template.repository.ts
```diff
@@ -187,6 +187,68 @@ export class NotificationTemplateRepository extends BaseRepository<
...
+   *
+   * @param {string} organizationId - The ID of the organization.
+   * @param {string} environmentId - The ID of the environment.
+   * @param {number} [skip=0] - The number of items to skip.
+   * @param {number} [limit=10] - The maximum number of items to return.
+   * @param {string} [query] - Optional query string to filter results.
+   * @returns {Promise<{ totalCount: number; data: NotificationTemplate[] }>} - A promise
+   * that resolves to an object containing the total count of items and the list of notification templates.
+   */
+  async getListExcludeNewDashboardWorkflows(
```
SokratisVidros: Can we reuse `getList` with one more option for this to avoid code duplication?

apps/api/src/app/workflows-v1/workflow-v1.controller.ts
```diff
@@ -227,7 +227,7 @@ export class WorkflowControllerV1 {
         data: body.data,
         __source: query?.__source,
         type: WorkflowTypeEnum.REGULAR,
-        origin: WorkflowOriginEnum.NOVU_CLOUD,
+        origin: WorkflowOriginEnum.NOVU_CLOUD_V1,
```
SokratisVidros, line 230: Do we need a data migration in Mongo for existing workflows?

SokratisVidros (changes requested):

libs/dal/src/repositories/notification-template/notification-template.repository.ts
tatarco: We can, I chose not to in order to easily get rid of this code after, it's deprecated, the ide shows it's deprecated and removing it soon will have zero impact on our existing production

apps/api/src/app/workflows-v1/workflow-v1.controller.ts
tatarco: We don't need but we can, there is still way to distinguish without migrating

libs/dal/src/repositories/notification-template/notification-template.repository.ts
SokratisVidros: The code will stay around for a while so let's reuse to avoid changing one method but not the other.

SokratisVidros (approved):

PR #7115, opened 2024-11-24
fix(node): add includeInactiveChannels option in sdk method

jainpawan21:
- Add `includeInactiveChannels` option in subscriber.getPreference method

Ref: [Discord Post](https://discord.com/channels/895029566685462578/1310170281700229180)

### Screenshots

<details>

</details>

### DIFF
--- a/packages/node/src/lib/subscribers/subscriber.interface.ts
+++ b/packages/node/src/lib/subscribers/subscriber.interface.ts
@@ -34,7 +34,10 @@ export interface ISubscribers {
    */
   unsetCredentials(subscriberId: string, providerId: string);
   updateOnlineStatus(subscriberId: string, online: boolean);
-  getPreference(subscriberId: string, includeInactiveChannels: boolean);
+  getPreference(
+    subscriberId: string,
+    { includeInactiveChannels }: { includeInactiveChannels: boolean },
+  );
   getGlobalPreference(subscriberId: string);
   getPreferenceByLevel(subscriberId: string, level: PreferenceLevelEnum);
   updatePreference(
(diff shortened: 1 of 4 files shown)


packages/node/src/lib/subscribers/subscriber.interface.ts
```diff
@@ -34,7 +34,7 @@ export interface ISubscribers {
    */
   unsetCredentials(subscriberId: string, providerId: string);
   updateOnlineStatus(subscriberId: string, online: boolean);
-  getPreference(subscriberId: string);
+  getPreference(subscriberId: string, includeInactiveChannels: boolean);
```
SokratisVidros:
```suggestion
  getPreference(subscriberId: string, { includeInactiveChannels: boolean });
```

packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -91,8 +91,13 @@ export class Subscribers extends WithHttp implements ISubscribers {
     return await this.http.delete(`/subscribers/${subscriberId}`);
   }
 
-  async getPreference(subscriberId: string) {
-    return await this.http.get(`/subscribers/${subscriberId}/preferences`);
+  async getPreference(
+    subscriberId: string,
+    includeInactiveChannels: boolean = true,
```
SokratisVidros:
```suggestion
    { includeInactiveChannels: boolean = true } 
```

packages/node/src/lib/subscribers/subscriber.interface.ts
SokratisVidros: Passing an object is more scalable and allows future addition of more parameters.

SokratisVidros (approved):

packages/node/src/lib/subscribers/subscribers.ts
```diff
@@ -93,7 +93,7 @@ export class Subscribers extends WithHttp implements ISubscribers {
 
   async getPreference(
     subscriberId: string,
-    includeInactiveChannels: boolean = true,
+    { includeInactiveChannels }: { includeInactiveChannels: boolean },
```
scopsy: We are lacking the defaults here:

```
    { includeInactiveChannels = false }: { includeInactiveChannels: boolean } = {},
```

SokratisVidros (approved):

PR #7116, opened 2024-11-24
feat(api): add framework workflow payload in preview

djabarovgeorge:

In App Step issue: Payload is not generated for the preview tab based on the payload schema for v2 workflows

1. Create a v2 code workflow with in app step
2. Specify the payload schema for the workflow
3. In app step in preview shows empty object

Expected behaviour:
* Payload schema is rendered as with UI based workflows

Also, use the preview response to see if errors still exist in combination with frontend validations

image
image

<details>

</details>


apps/api/src/app/bridge/bridge.module.ts
```diff
@@ -18,6 +18,38 @@ import { PreferencesRepository } from '@novu/dal';
...
+  BuildDefaultPayloadUsecase,
+  CollectPlaceholderWithDefaultsUsecase,
+  PrepareAndValidateContentUsecase,
+  ValidatePlaceholderUsecase,
+} from '../workflows-v2/usecases/validate-content';
+import { BuildAvailableVariableSchemaUsecase } from '../workflows-v2/usecases/build-variable-schema';
+import { ExtractDefaultValuesFromSchemaUsecase } from '../workflows-v2/usecases/extract-default-values-from-schema';
+import { HydrateEmailSchemaUseCase } from '../environments-v1/usecases/output-renderers/hydrate-email-schema.usecase';
+
+const SHARED_USECASES = [
```
djabarovgeorge: We should avoid using use cases from another module (as we're already doing) because it can create tight coupling, make maintenance more difficult, and potentially cause circular dependencies. Although this isn’t an issue right now since we don’t import the module directly, it could still lead to problems in the future. Instead, we should identify all shared use cases and move them to a dedicated shared module.

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
```diff
@@ -38,17 +38,22 @@ export class BuildAvailableVariableSchemaUsecase {
...
       },
       additionalProperties: false,
     } as const satisfies JSONSchemaDto;
   }
 }
-function safePayloadSchema(workflow: NotificationTemplateEntity): JSONSchemaDto | undefined {
+function safePayloadSchema(workflow: NotificationTemplateEntity): JSONSchemaDto {
   try {
-    return JSON.parse(workflow.payloadSchema);
+    // TODO: fix the type, we need to make sure we have the same data structure in the code
```
djabarovgeorge: we have inconsistency in the database where for framework workflows we have object and dashboard workflow, and we have it stringified to string

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -228,14 +229,20 @@ export class UpsertWorkflowUseCase {
   }
 
   private buildBaseStepEntity(
-    step: StepDto | StepUpdateDto,
+    step: StepCreateDto | StepUpdateDto,
     foundPersistedStep?: NotificationStepEntity
   ): NotificationStep {
+    const controlDto = step.controlSchema ? { schema: step.controlSchema } : null;
+    const outputDto = step.outputSchema ? { schema: step.outputSchema } : null;
+    const outputFallback = { schema: stepTypeToDefaultDashboardControlSchema[step.type].schema };
```
djabarovgeorge, line 238: To align framework and dashboard workflows, I added a fallback for the output schema from controls schema. 
In framework workflows, the output schema is always present, while in dashboard workflows, the controls currently match the outputs one-to-one. If this changes in the future, this section will need to be updated.

apps/api/src/app/workflows-v2/usecases/validate-content/collect-placeholders/collect-placeholder-with-defaults.usecase.ts
djabarovgeorge, line 1: I don't think this is the best solution. Currently, we couple the placeholders to the control values and extract all used variables from the control values and then process them further, including handling issues etc... 
However, for framework external workflows, we need to ignore variables for payload in the control values and use **only** those defined in the workflow definition. To address this, I created a filter that removes any `{{payload.VAR}}` variables found and then adds the variables from the payload schema.

A better solution would be to see if this logic can be refactored and extracted, so we don’t have to handle this filtering within nested code logic.

packages/shared/src/dto/workflows/update-workflow-dto.ts
```diff
@@ -1,11 +1,6 @@
 import { PreferencesRequestDto, StepCreateDto, StepUpdateDto, WorkflowCommonsFields } from './workflow-commons-fields';
 
 export type UpdateWorkflowDto = WorkflowCommonsFields & {
-  /**
-   * We allow to update workflow id only for code first workflows
-   */
-  workflowId?: string;
```
djabarovgeorge, line 7: was added to FrameworkCommonsFields

packages/shared/src/dto/workflows/create-workflow-dto.ts
```diff
@@ -2,8 +2,6 @@ import { PreferencesRequestDto, StepCreateDto, WorkflowCommonsFields } from './w
 import { WorkflowCreationSourceEnum } from '../../types';
 
 export type CreateWorkflowDto = WorkflowCommonsFields & {
-  workflowId: string;
```
djabarovgeorge, line 5: was added to FrameworkCommonsFields

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
djabarovgeorge: https://linear.app/novu/issue/NV-4874/align-the-workflowpayloadschema-type

apps/api/src/app/bridge/bridge.module.ts
```diff
@@ -13,11 +13,42 @@ import {
...
+  BuildDefaultPayloadUsecase,
+  CollectPlaceholderWithDefaultsUsecase,
+  PrepareAndValidateContentUsecase,
+  ValidatePlaceholderUsecase,
+} from '../workflows-v2/usecases/validate-content';
+import { BuildAvailableVariableSchemaUsecase } from '../workflows-v2/usecases/build-variable-schema';
+import { ExtractDefaultValuesFromSchemaUsecase } from '../workflows-v2/usecases/extract-default-values-from-schema';
+import { HydrateEmailSchemaUseCase } from '../environments-v1/usecases/output-renderers/hydrate-email-schema.usecase';
+
+const SHARED_USE_CASES = [
```
SokratisVidros: Since these aren't shared with anything else, we should move them to the PROVIDERS array.

apps/api/src/app/bridge/e2e/sync.e2e.ts
```diff
@@ -112,7 +113,6 @@ describe('Bridge Sync - /bridge/sync (POST)', async () => {
 
     expect(workflowData.steps.length).to.equal(1);
     expect(workflowData.steps[0].stepId).to.equal('send-email');
-    expect(workflowData.steps[0].uuid).to.equal('send-email');
```
SokratisVidros, line 115: @djabarovgeorge This tells us that the mock data generator for the e2e is probably not realistic and needs an update. Let's fix that in a follow up PR.

apps/api/src/app/bridge/e2e/sync.e2e.ts
```diff
@@ -561,93 +558,143 @@ describe('Bridge Sync - /bridge/sync (POST)', async () => {
...
+      workflowsRepository,
+      firstSyncResponse,
+      workflowId
+    );
 
-    const firstControlValueResponse = await controlValuesRepository.find({
-      _environmentId: session.environment._id,
-      _workflowId: firstWorkflowResponse._id,
-    });
+    await validateControlValues(controlValuesRepository, session, secondWorkflowResponse, workflowId);
```
SokratisVidros: Please inline the expectations to the tests, it makes reading the test much easier without having to jump across functions.

apps/api/src/app/bridge/e2e/sync.e2e.ts
```diff
@@ -561,93 +558,143 @@ describe('Bridge Sync - /bridge/sync (POST)', async () => {
...
-    const firstStepResponse = await session.testAgent.get(`/v1/bridge/controls/${workflowId}/send-email`);
-    expect(firstStepResponse.body.data.controls.subject).to.equal('Hello World again');
+  expect(secondControlValueResponse.length).to.equal(1);
+  expect(secondControlValueResponse[0].controls.subject).to.equal('Hello World again');
+
+  const secondStepResponse = await session.testAgent.get(`/v1/bridge/controls/${workflowId}/send-email`);
+  expect(secondStepResponse.body.data.controls.subject).to.equal('Hello World again');
+}
+
+async function updateWorkflow(
```
SokratisVidros: Ditto about inlining this function in the tests it belongs. This code contains test steps along with its expectations so the name is misleading.

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -142,149 +161,134 @@ export class Sync {
...
     return Promise.all(
       workflowsFromBridge.map(async (workflow) => {
-        const workflowExist = await this.notificationTemplateRepository.findByTriggerIdentifier(
+        const fetchedWorkflow = await this.notificationTemplateRepository.findByTriggerIdentifier(
           command.environmentId,
           workflow.workflowId
         );
 
-        let savedWorkflow: NotificationTemplateEntity | undefined;
+        const upsertWorkflowCommand = fetchedWorkflow
```
SokratisVidros, line 175: Commands should follow the use case pattern. That is, we should encapsulate the create or update logic of a command inside the upsert command or reuse the create or update upsert logic of the usecase.

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
```diff
@@ -0,0 +1,19 @@
...
+    try {
+      return JSON.parse(schema);
+    } catch (error) {
+      throw new Error(
+        'Invalid payload schema: Unable to parse provided JSON string. Please ensure the string contains valid JSON syntax.'
+      );
+    }
+  }
+
+  if (schema && typeof schema === 'object') {
```
SokratisVidros, line 14: Let's check this first and return the object. 
```suggestion
  if (typeof schema === 'object') {
```

Then we can try/catch with `JSON.parse` on anything else without extra ifs. Please also make sure to add a type assertion on JSON.parse.

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
```diff
@@ -38,17 +39,17 @@ export class BuildAvailableVariableSchemaUsecase {
...
         steps: buildPreviousStepsSchema(previousSteps, workflow.payloadSchema),
-        payload: safePayloadSchema(workflow) || { type: 'object', description: 'Payload for the current step' },
+        payload: safePayloadSchema(workflow),
       },
       additionalProperties: false,
     } as const satisfies JSONSchemaDto;
   }
 }
-function safePayloadSchema(workflow: NotificationTemplateEntity): JSONSchemaDto | undefined {
+function safePayloadSchema(workflow: NotificationTemplateEntity): JSONSchemaDto {
```
SokratisVidros, line 48: We already do this in `parsePayloadSchema`. If we want to always return an object, we can tweak the error handling there.

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
```diff
@@ -38,17 +39,17 @@ export class BuildAvailableVariableSchemaUsecase {
...
   try {
-    return JSON.parse(workflow.payloadSchema);
+    return parsePayloadSchema(workflow.payloadSchema);
   } catch (e) {
-    return undefined;
+    return { type: 'object', description: 'Payload for the current step' };
```
SokratisVidros, line 52: Do we need this hardcoded schema object or should we just return `{}` or `null` if parse fails? The latter feels safer and more future proof.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
SokratisVidros: I don't know if that's true. Controls don't much outputs 100% in the Dashboard.

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -46,15 +46,29 @@ export type StepResponseDto = StepDto & {
   issues?: StepIssuesDto;
 };
 
-export type StepUpdateDto = StepCreateDto & {
+export type StepUpdateDto = StepDto & {
```
SokratisVidros, line 49: Nice 😍

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -74,7 +88,17 @@ export type StepDto = {
   type: StepTypeEnum;
 };
 
-export type WorkflowCommonsFields = {
+/**
+ * Common fields used by framework workflows to store schemas and metadata
+ */
+export type FrameworkCommonsFields = {
```
SokratisVidros, line 94: This goes against our goal, which is to unify the framework and dashboard workflows. Which are the attributes that prevent us from defining a single unified workflow?

SokratisVidros (changes requested):

apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -142,149 +161,134 @@ export class Sync {
...
+        return StepTypeEnum.SMS;
+      case 'push':
+        return StepTypeEnum.PUSH;
+      case 'chat':
+        return StepTypeEnum.CHAT;
+      case 'digest':
+        return StepTypeEnum.DIGEST;
+      case 'delay':
+        return StepTypeEnum.DELAY;
+      case 'custom':
```
tatarco, line 287: All of this including all the changes here belongs to a diffrent PR, it has nothing to do with the "bug" we needed to fix. that bug assumes payload schema already persisted

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
tatarco: I think if it don't exist it will say that typeof is undefined so it's not needed

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
```diff
@@ -0,0 +1,19 @@
...
+    return schema as JSONSchemaDto;
+  }
+
+  throw new Error('Invalid payload schema: Payload schema must be either a valid JSON string or an object');
```
tatarco, line 18: don't you think it should be an internal server error of a specific kind

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
tatarco: I agree it should merge the logic, but the default is important, as it's the schema eq to {}, you are stating a general object with no additional information in the schema, having no payload schema doesn't mean you acannot use it as a variable

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -56,6 +57,7 @@ export class GeneratePreviewUsecase {
       controlDataSchema: stepData.controls.dataSchema,
       variableSchema: stepData.variables,
       previewPayloadFromDto: dto.previewPayload,
+      origin: stepData.origin || WorkflowOriginEnum.NOVU_CLOUD_V1,
```
tatarco, line 60: not relevant to the PR, why was this change made?

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
```diff
@@ -45,15 +45,16 @@ export class OverloadContentDataOnWorkflowUseCase {
...
         workflow,
         stepDatabaseId: step._templateId,
       });
       validatedStepContent[step._templateId] = await this.prepareAndValidateContentUsecase.execute({
         stepType: step?.template?.type,
         controlDataSchema: controls.schema,
         controlValues,
-        variableSchema: jsonSchemaDto,
+        variableSchema,
+        origin: workflow.origin || WorkflowOriginEnum.NOVU_CLOUD_V1,
```
tatarco, line 57: this feels weird, why do we need this for validation?

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -132,15 +131,16 @@ export class UpsertWorkflowUseCase {
...
-      payloadSchema: {},
+      rawData: workflowDto,
       active: isWorkflowActive,
       description: workflowDto.description || '',
       tags: workflowDto.tags || [],
       userPreferences: command.workflowDto.preferences?.user ?? null,
       defaultPreferences: command.workflowDto.preferences?.workflow ?? DEFAULT_WORKFLOW_PREFERENCES,
-      triggerIdentifier: slugify(workflowDto.name),
+      triggerIdentifier: workflowDto.workflowId || slugify(workflowDto.name),
+      payloadSchema: workflowDto.payloadSchema,
```
tatarco, line 143: is that meant to support code first? not really relevant for that PR

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -228,14 +230,20 @@ export class UpsertWorkflowUseCase {
...
+    const outputFallback = { schema: stepTypeToDefaultDashboardControlSchema[step.type].schema };
+
     return {
       template: {
         type: step.type,
         name: step.name,
-        controls: foundPersistedStep?.template?.controls || stepTypeToDefaultDashboardControlSchema[step.type],
+        controls:
+          controlDto || foundPersistedStep?.template?.controls || stepTypeToDefaultDashboardControlSchema[step.type],
+        output: outputDto || foundPersistedStep?.template?.output || outputFallback,
```
tatarco, line 246: this is not required, not even completely clear  where and how it's used, for sure not needed for dashboard first and for sure not a part of this PR

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -228,14 +230,20 @@ export class UpsertWorkflowUseCase {
...
+    const outputDto = step.outputSchema ? { schema: step.outputSchema } : null;
+    const outputFallback = { schema: stepTypeToDefaultDashboardControlSchema[step.type].schema };
+
     return {
       template: {
         type: step.type,
         name: step.name,
-        controls: foundPersistedStep?.template?.controls || stepTypeToDefaultDashboardControlSchema[step.type],
+        controls:
+          controlDto || foundPersistedStep?.template?.controls || stepTypeToDefaultDashboardControlSchema[step.type],
```
tatarco, line 245: this multiple || behavior worries me, it's not clear what we are trying to achieve

apps/api/src/app/workflows-v2/usecases/validate-content/collect-placeholders/collect-placeholder-with-defaults.command.ts
```diff
@@ -1,5 +1,8 @@
 import { BaseCommand } from '@novu/application-generic';
+import { JSONSchemaDto, WorkflowOriginEnum } from '@novu/shared';
 
 export class CollectPlaceholderWithDefaultsCommand extends BaseCommand {
   controlValues?: Record<string, unknown>;
+  origin?: WorkflowOriginEnum;
+  payloadSchema?: JSONSchemaDto;
```
tatarco, line 7: why do you do that, collect is simple and iterates over the values, this should not be a part of it

apps/api/src/app/workflows-v2/usecases/validate-content/collect-placeholders/collect-placeholder-with-defaults.command.ts
```diff
@@ -1,5 +1,8 @@
 import { BaseCommand } from '@novu/application-generic';
+import { JSONSchemaDto, WorkflowOriginEnum } from '@novu/shared';
 
 export class CollectPlaceholderWithDefaultsCommand extends BaseCommand {
   controlValues?: Record<string, unknown>;
+  origin?: WorkflowOriginEnum;
```
tatarco, line 6: also completely irrelevant for this usecase

apps/api/src/app/workflows-v2/usecases/validate-content/collect-placeholders/collect-placeholder-with-defaults.usecase.ts
```diff
@@ -11,16 +12,57 @@ export class CollectPlaceholderWithDefaultsUsecase {
...
+
+      placeholderAggregation.regularPlaceholdersToDefaultValue = Object.fromEntries(
+        Object.entries(regularPlaceholders).filter(([key]) => !key.includes('{{payload.'))
+      );
+      placeholderAggregation.nestedForPlaceholders = Object.fromEntries(
+        Object.entries(nestedPlaceholders).filter(([key]) => !key.includes('{{payload.'))
+      );
+    }
+
+    return resultPlaceholders;
```
tatarco, line 63: unclear to me why this usecase was changed, will appreciate clarification, seems at first glance unnecessary to me

apps/api/src/app/workflows-v2/usecases/validate-content/collect-placeholders/collect-placeholder-with-defaults.usecase.ts
```diff
@@ -11,16 +12,57 @@ export class CollectPlaceholderWithDefaultsUsecase {
...
+  }
+
```
tatarco, line 65: I would expect all the work done here to be done at an additional step in the validator looking at the payload schema exmple vs the preview payload defaults generated

apps/api/src/app/workflows-v2/usecases/validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase.ts
```diff
@@ -40,7 +41,12 @@ export class PrepareAndValidateContentUsecase {
   ) {}
 
   async execute(command: PrepareAndValidateContentCommand): Promise<ValidatedContentResponse> {
-    const controlValueToPlaceholders = this.collectPlaceholders(command.controlValues);
+    const controlValueToPlaceholders = this.collectPlaceholders(
+      command.controlValues,
+      command.variableSchema.properties?.payload as JSONSchemaDto,
+      command.origin
```
tatarco, line 47: please remove

packages/shared/src/dto/workflows/create-workflow-dto.ts
tatarco: why? it's not a part of the create no? it's generatred on the backend, what it has to do with this PR

packages/shared/src/dto/workflows/workflow-commons-fields.ts
```diff
@@ -46,15 +46,29 @@ export type StepResponseDto = StepDto & {
...
 export type StepCreateDto = StepDto & {
   /**
    * @deprecated This field is deprecated and will be removed in future versions, use the patch step data.
    */
   controlValues?: Record<string, unknown>;
+  /**
+   * for code-first workflows we allow to store the control schema on create/update step
+   */
+  controlSchema?: JSONSchemaDto;
+  outputSchema?: JSONSchemaDto;
```
tatarco, line 71: I really dislike the mix here, we should not be mixing our DTO's we have a DTO for the code first, we run upsert as it is and run the rest outside of it, we should not break the upsert or anything else and make the contract not as good in order to support code first which is already structured

packages/shared/src/dto/workflows/workflow-commons-fields.ts
tatarco: where was that goal defined? 
there is already a DTO for code first, we should not alter DTO's for our API's to support specifically code first, IMO  the sync should reuse as much as possible and add on top of it the unuiqe features of code first

tatarco (changes requested):
I'm sorry for blocking this PR, but there are some choices here that I would like us to explore changing them:

1. Altering upsert
2. Altering the validator with origin & payload schema.
3. Altering the working and proven DTO's

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: you are correct, we need to map the email one. i forgot about it, thanks!

self-note: check if there are any other misalignments besides email.

packages/shared/src/dto/workflows/create-workflow-dto.ts
djabarovgeorge: It generated on the back for the dashboard workflow, however, now we reuse upsert for framework workflows which define there one workflowId.

apps/api/src/app/bridge/bridge.module.ts
djabarovgeorge: I planned to refactor this part and move it into a separate NestJS module so that both the Sync Module and the WorkflowV2 Module can use it. However, I'll postpone this for now since including it in this PR would slow down its progress.

For now, i will do as you suggested ;)

apps/api/src/app/bridge/e2e/sync.e2e.ts
djabarovgeorge: It is not related to the mock, this param was generated in the API.
it is related to the fact that we do not create this duplicated uuid in the workflow entity anymore, so there is nothing to check.

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
```diff
@@ -94,6 +95,6 @@ export class OverloadContentDataOnWorkflowUseCase {
       finalPayload = _.merge(finalPayload, value.finalPayload.payload);
     }
     // eslint-disable-next-line no-param-reassign
-    workflow.payloadSchema = JSON.stringify(convertJsonToSchemaWithDefaults(finalPayload));
+    workflow.payloadSchema = convertJsonToSchemaWithDefaults(finalPayload);
```
djabarovgeorge, line 98: removing the stringify that resulted with two payloadSchema type schema | string

PR #7118, opened 2024-11-24
fix(api): Migrate subscriber global preferences before workflow preferences

rifont:
* Migrate subscriber global preferences before workflow preferences
  * This is necessary to preserve the workflow preference, as the subscriber upsert behaviour deletes workflow preferences for updated global channels
* Add queue mechanism to decouple fetching and processing
* Improve logging

### Screenshots

<details>

</details>


apps/api/package.json
```diff
@@ -117,6 +118,7 @@
     "@types/passport-jwt": "^3.0.3",
     "@types/sinon": "^9.0.0",
     "@types/supertest": "^2.0.8",
+    "async": "^3.2.0",
```
rifont, line 121: Async lib with queue API added as `devDependency`

SokratisVidros (approved):

SokratisVidros (approved):

PR #7119, opened 2024-11-24
perf(api): Lookup subscriber preferences with a single database query

rifont:
* Lookup each subscriber preference set with a single database query
* Extract preference merging logic to a separate use-case for shared use

### Screenshots

<details>

</details>


libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -49,30 +59,119 @@ export class GetSubscriberPreference {
...
+          _templateId: { $in: workflowList.map((template) => template._id) },
+          type: {
+            $in: [
+              PreferencesTypeEnum.WORKFLOW_RESOURCE,
+              PreferencesTypeEnum.USER_WORKFLOW,
+            ],
+          },
+        },
+      ],
+    });
```
rifont: Et voila, a single database query!

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -49,30 +60,138 @@ export class GetSubscriberPreference {
...
-            organizationId: command.organizationId,
-            subscriberId: command.subscriberId,
-            environmentId: command.environmentId,
-            template,
-            subscriber,
-            includeInactiveChannels: command.includeInactiveChannels,
-          }),
-        ),
-      ),
+    const allPreferences = await this.preferencesRepository.find({
```
SokratisVidros: ❓ Is it maybe faster to split this into these into parallel queries, one for each Preference type?

libs/application-generic/src/usecases/merge-preferences/merge-preferences.usecase.ts
```diff
@@ -0,0 +1,189 @@
...
+ * 3. Subscriber global preferences
+ * 4. Subscriber workflow preferences
+ *
+ * If a workflow has the readOnly flag set to true, the subscriber preferences are ignored.
+ *
+ * If the workflow does not have the readOnly flag set to true, the subscriber preferences are merged with the workflow preferences.
+ *
+ * If the subscriber has no preferences, the workflow preferences are returned.
+ */
+export class MergePreferences {
```
SokratisVidros, line 23: 😍 This is exactly what we needed in the code. A central place to do the resolution.

SokratisVidros (approved):

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -49,30 +60,138 @@ export class GetSubscriberPreference {
...
+          },
+        },
+        {
+          _subscriberId: subscriber._id,
+          type: PreferencesTypeEnum.SUBSCRIBER_GLOBAL,
+        },
+      ],
+    });
+
+    const subscriberGlobalPreferences = allPreferences.filter(
```
SokratisVidros: If we split it into separate Mongo queries, we can fetch this set of preferences from Mongo.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -49,30 +60,138 @@ export class GetSubscriberPreference {
...
+      (preference) =>
+        preference._subscriberId === subscriber._id &&
+        preference.type === PreferencesTypeEnum.SUBSCRIBER_GLOBAL,
+    );
+
+    const workflowPreferenceSets = allPreferences.reduce<
```
SokratisVidros: Ditto. If we split it into separate Mongo queries, we can fetch this set of preferences from Mongo.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -49,30 +60,138 @@ export class GetSubscriberPreference {
...
+          preferences: [...preferences, ...subscriberGlobalPreferences],
+        });
+        const merged = MergePreferences.merge(mergeCommand);
+
+        const includedChannels = this.getChannels(
+          workflow,
+          command.includeInactiveChannels,
+        );
+
+        const initialChannels = filteredPreference(
```
SokratisVidros, line 156: Should we move these two processing functions into the MergeCommand to have everything in one place?

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
```diff
@@ -49,30 +60,138 @@ export class GetSubscriberPreference {
...
-    const nonCriticalWorkflowPreferences = subscriberWorkflowPreferences.filter(
-      (preference) => preference.template.critical === false,
+    const nonCriticalWorkflowPreferences = mergedPreferences.filter(
+      (preference) => !preference.template.critical,
     );
 
     return nonCriticalWorkflowPreferences;
   }
+
+  private getChannels(
```
SokratisVidros, line 203:
```suggestion
  private getActiveChannels(
```

libs/application-generic/src/usecases/merge-preferences/merge-preferences.usecase.ts
```diff
@@ -0,0 +1,189 @@
...
+     * Order is (almost exactly) reversed of that above because 'readOnly' should be prioritized
+     * by the Dashboard (userPreferences) the most.
+     */
+    const orderedPreferencesForReadOnly = [
+      subscriberWorkflowPreferences,
+      subscriberGlobalPreferences,
+      workflowResourcePreferences,
+      workflowUserPreferences,
+    ]
+      .filter((preference) => preference !== undefined)
```
SokratisVidros:
```suggestion
      .reduce((preference) => ...)
```

libs/application-generic/src/usecases/merge-preferences/merge-preferences.usecase.ts
```diff
@@ -0,0 +1,189 @@
...
+ *
+ * If a workflow has the readOnly flag set to true, the subscriber preferences are ignored.
+ *
+ * If the workflow does not have the readOnly flag set to true, the subscriber preferences are merged with the workflow preferences.
+ *
+ * If the subscriber has no preferences, the workflow preferences are returned.
+ */
+export class MergePreferences {
+  public static merge(
+    command: MergePreferencesCommand,
```
SokratisVidros, line 25: I feel the code would be simplified if we passed each type of preference separately to this use case. Currently, we fetch from Mongo, we do some mini-processing before calling the Merge use case, and then we combine them and split them again inside the use case.

My suggestion is:

1. Fetch the four types of preferences separately (with a single or separate queries from Mongo).
2. Feed them to MergeUseCase separately via the command.
3. Do all the processing and merging in the MergePreferences use case.

libs/application-generic/src/usecases/merge-preferences/merge-preferences.usecase.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    // ensure we don't merge on an empty list
+    if (preferences.length === 0) {
+      return { preferences: undefined, type: undefined, source };
+    }
+
+    const readOnlyFlag = workflowPreferences?.all?.readOnly;
+
+    // Determine the most specific preference applied
+    let mostSpecificPreference: PreferencesTypeEnum | undefined;
```
SokratisVidros: We can get this automatically from `deepMerge` if we provide it during function invocation. The last type will win.

SokratisVidros (changes requested):

SokratisVidros (approved):
Beauty!

libs/application-generic/src/usecases/merge-preferences/merge-preferences.usecase.ts
rifont: Great suggestion, this helped to simplify the use-case a lot. Done!

libs/application-generic/src/usecases/merge-preferences/merge-preferences.usecase.ts
rifont: Yes! Done

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
rifont: I have left it as `getChannels` because it's possible for the method to return inactive channels when specifying the `includeInactiveChannels` flag, and the refined naming would be misleading.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
rifont: I think we probably could, but there will be some related things that are removed in the process of removing the V1 Preferences. I suggest we get back to this after their removal.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
rifont: I have split into a single query for each preference, the code reads nicer and the size of the collection means it will be more performant to query in parallel.

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
rifont: Done

libs/application-generic/src/usecases/get-subscriber-preference/get-subscriber-preference.usecase.ts
rifont: Yes! Done

PR #7123, opened 2024-11-25
fix(js): Truncate workflow name and center empty notifications text

desiprisg:
### Screenshots
Screenshot 2024-11-25 at 1 56 38 PM
Screenshot 2024-11-25 at 1 58 21 PM

<details>

</details>

### DIFF
--- a/packages/js/src/ui/components/elements/Preferences/Preferences.tsx
+++ b/packages/js/src/ui/components/elements/Preferences/Preferences.tsx
@@ -148,17 +148,14 @@ const PreferencesRow = (props: {
         <div
           class={style(
             'workflowLabelContainer',
-            'nt-flex nt-justify-between nt-flex-nowrap nt-self-stretch nt-cursor-pointer nt-items-center'
+            'nt-flex nt-justify-between nt-flex-nowrap nt-self-stretch nt-cursor-pointer nt-items-center nt-overflow-hidden'
           )}
           onClick={() => setIsOpen((prev) => !prev)}
           data-open={isOpen()}
         >
-          <div>
+          <div class={style('workflowLabelHeader', 'nt-overflow-hidden')}>
             <div
-              class={style(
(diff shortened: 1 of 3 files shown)


packages/js/src/ui/components/elements/Preferences/Preferences.tsx
```diff
@@ -148,17 +148,14 @@ const PreferencesRow = (props: {
...
           class={style(
             'workflowLabelContainer',
-            'nt-flex nt-justify-between nt-flex-nowrap nt-self-stretch nt-cursor-pointer nt-items-center'
+            'nt-flex nt-justify-between nt-flex-nowrap nt-self-stretch nt-cursor-pointer nt-items-center nt-overflow-hidden'
           )}
           onClick={() => setIsOpen((prev) => !prev)}
           data-open={isOpen()}
         >
-          <div>
+          <div style={{ overflow: 'hidden' }}>
```
rifont: If someone tried to override the default behavior via the appearance prop classnames, wouldn't this take precedence and prevent them from modifying behavior?

desiprisg: Correct. Don't see any case for this to happen now though. We have no descriptor here so you can't override anyway. Can you come up with a good name for it?

rifont: I don't think there's a particularly strong case either, but still good to keep it flexible. Here are some options:

1. workflowLabelSection
2. workflowLabelHeader
3. workflowLabelDisplay
4. workflowLabelPanel

Personally I would go with `workflowLabelHeader`.

desiprisg: Done! Nice suggestion.

SokratisVidros (approved):

PR #7124, opened 2024-11-25
chore(dashboard): user journey smoke tests

LetItRock:

Enterprise PR: https://github.com/novuhq/packages-enterprise/pull/273

Dashboard - the core user journeys smoke tests that cover:
- dashboard-created workflow 
- code-created workflow

### Screenshots

Screenshot 2024-11-25 at 13 59 28


apps/api/.gitignore
```diff
@@ -121,3 +121,4 @@ backups/
 
 # Nest.js auto-generated metadata (https://docs.nestjs.com/recipes/swc#monorepo-and-cli-plugins)
 src/metadata.ts
+src/.env.test
```
LetItRock: removed the `.env.test` file from git, because the local file that I used for the tests had sensitive clerk credentials

apps/dashboard/package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "start": "vite",
+    "start:test": "vite --mode test",
```
LetItRock, line 8: run dashboard and use `.env.test` file

apps/dashboard/package.json
```diff
@@ -83,9 +84,12 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@clerk/testing": "^1.3.27",
     "@clerk/types": "^4.30.0",
     "@eslint/js": "^9.9.0",
     "@hookform/devtools": "^4.3.0",
+    "@novu/dal": "workspace:*",
+    "@novu/testing": "workspace:*",
```
LetItRock, line 117: used to initialize the api session

apps/dashboard/playwright.config.ts
```diff
@@ -4,9 +4,16 @@ import { defineConfig, devices } from '@playwright/test';
...
-// import path from 'path';
-// dotenv.config({ path: path.resolve(__dirname, '.env') });
+import dotenv from 'dotenv';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { dirname } from 'path';
+
+const fileName = fileURLToPath(import.meta.url);
+const dirName = dirname(fileName);
+dotenv.config({ path: path.resolve(dirName, '.env.playwright') });
```
LetItRock, line 14: Because the Dashboard project is configured with `"type": "module"` this is how we can import the playwright env variables.

apps/dashboard/playwright.config.ts
```diff
@@ -38,9 +51,22 @@ export default defineConfig({
   },
   /* Configure projects for major browsers */
   projects: [
+    {
+      name: 'setup',
+      testMatch: /global\.setup\.ts/,
+    },
```
LetItRock: The global setup test file will initialize the Clerk session for all tests. Using this approach, we can say that for some tests, we don't want to have a session, and then we will be able to verify, for example, the sign-in/up flows.

apps/dashboard/src/components/primitives/tag-input.tsx
```diff
@@ -78,8 +78,10 @@ const TagInput = forwardRef<HTMLInputElement, TagInputProps>((props, ref) => {
           <div className="flex flex-wrap gap-2">
             {tags.map((tag, index) => (
               <Badge key={index} variant="outline" size="tag" className="gap-1">
-                <span style={{ wordBreak: 'break-all' }}>{tag}</span>
-                <button type="button" onClick={() => removeTag(tag)}>
+                <span style={{ wordBreak: 'break-all' }} data-testid="tags-badge-value">
```
LetItRock, line 101: For some components, I had to add `data-tested` as there was no accessible way to reference the elements to perform some actions.

apps/dashboard/src/utils/auth.ts
```diff
@@ -6,5 +8,9 @@ declare namespace Clerk {
 }
 
 export async function getToken(): Promise<string> {
+  if (MODE === 'test') {
+    return localStorage.getItem('nv_auth_token') ?? '';
+  }
```
LetItRock: For tests, we store the API session-generated JWT token in the local storage because we don't use the one that is provided by the Clerk as the token signature won't match on the API, and we will not be able to authorize the requests. 

In short the `@novu/testing` library needs some refactoring which is out of scope and might take a lot of time.

apps/dashboard/tests/code-created-flow.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+import { BridgeServer } from './utils/bridge-server';
+import { initializeSession } from './utils/session';
+
+const workflowId = 'code-created-workflow';
+const inAppStepId = 'send-in-app';
+const body = 'To join the Novu project, click the link below';
+
+let bridgeServer: BridgeServer;
+test.beforeEach(async ({ page }) => {
+  const session = await initializeSession({ page });
```
LetItRock: Initialize the API session. 
I tried to move this to the fixture, but we also have the global setup that is running before all tests, which will also run the fixture, resulting in 2x session initialization, like fixture run --> global setup, fixture run --> test.

apps/dashboard/tests/code-created-flow.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+            name: { type: 'string', default: 'John' },
+          },
+        } as const,
+      }
+    );
+  });
+  await bridgeServer.start({ workflows: [newWorkflow] });
+  await session.testAgent.post(`/v1/bridge/sync`).send({
+    bridgeUrl: bridgeServer.serverPath,
+  });
```
LetItRock: Run the bridge server and sync the workflow

apps/dashboard/tests/code-created-flow.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  await page.waitForTimeout(2000);
```
LetItRock: should be removed

apps/dashboard/tests/global.setup.ts
```diff
@@ -0,0 +1,33 @@
...
+
+setup('global setup', async ({ page }) => {
+  await clerkSetup();
+
+  /**
+   * TODO: Currently we are running tests with a single Clerk user and organization.
+   * This approach has a drawback that the tests are not independent from each other and write the data to the same organization.
+   * We should consider creating a new Clerk user and organization for each test using the @clerk/backend package.
+   * Then initialize the BE session with the new Clerk user and organization and update Clerk user metadata from the newly created session.
+   */
```
LetItRock: hopefully, it explains the problem well

apps/dashboard/tests/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "commonjs"
```
LetItRock, line 2: The `dal` and `testing` packages are CJS modules which can't be imported by the ESM modules. 
So this is the trick (together with tsconfig.json) to say "please allow me to import" in my test files :D

libs/testing/src/ee/ee.organization.service.ts
```diff
@@ -1,11 +1,16 @@
...
+import { getEEMemberRepository, getEEOrganizationRepository, getEERepository } from './ee.repository.factory';
 
 export class EEOrganizationService {
-  private organizationRepository = getEERepository<OrganizationRepository>('OrganizationRepository');
+  private organizationRepository: OrganizationRepository;
   private communityOrganizationRepository = new CommunityOrganizationRepository();
-  private memberRepository = getEERepository<MemberRepository>('MemberRepository');
+  private memberRepository: MemberRepository;
+
+  constructor({ mockClerkClient = true }: { mockClerkClient?: boolean } = {}) {
```
LetItRock: For the Dashboard E2E tests we don't want to mock the Clerk client, because we initialize the Clerk session with a real user that has the org and metadata that is used by the Dashboard code.

apps/dashboard/tests/dashboard-created-flow.spec.ts
```diff
@@ -0,0 +1,115 @@
+import { StepTypeEnum } from '@novu/shared';
```
SokratisVidros: How about naming it from the user journey perspective? For example: `manage-workflows.e2e.ts`.

I added the e2e extension as we have the convention with e2e testing file naming on the API.

apps/dashboard/tests/code-created-flow.spec.ts
```diff
@@ -0,0 +1,90 @@
+import { test } from '@playwright/test';
```
SokratisVidros: How about naming it from the user journey perspective? For example: `sync-workflow.e2e.ts.`

apps/dashboard/tests/code-created-flow.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+  await session.testAgent.post(`/v1/bridge/sync`).send({
+    bridgeUrl: bridgeServer.serverPath,
+  });
+});
+
+test.afterEach(async () => {
+  await bridgeServer.stop();
+});
+
+test('code defined workflow user journey', async ({ page }) => {
```
SokratisVidros:
```suggestion
test('sync workflow', async ({ page }) => {
```

apps/dashboard/tests/dashboard-created-flow.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+import { WorkflowEditorPage } from './page-models/workflow-editor-page';
+import { StepConfigSidebar } from './page-models/step-config-sidebar';
+import { TriggerWorkflowPage } from './page-models/trigger-workflow-page';
+import { initializeSession } from './utils/session';
+
+test.beforeEach(async ({ page }) => {
+  await initializeSession({ page });
+});
+
+test('dashboard defined workflow user journey', async ({ page }) => {
```
SokratisVidros:
```suggestion
test('manage workflows', async ({ page }) => {
```

apps/dashboard/tests/global.setup.ts
SokratisVidros: Correct! We should leverage [Clerk Backend API](https://clerk.com/docs/reference/backend-api), create a user before every run, sign in with that user, and then delete the user at the end.

Regarding the need for Clerk webhooks during user and organization creation, I suggest connecting to Mongo directly from the test code using the response of the Clerk API to create the user and organization record the application needs to function.

On user deletion, I don't think we need to do anything in Mongo as we run in a dockerized environment.

apps/dashboard/tests/page-models/create-workflow-sidebar.ts
```diff
@@ -0,0 +1,70 @@
+import { expect, type Page } from '@playwright/test';
```
SokratisVidros: Nit-picky: As per definition `page-models` -> `page-objects`

apps/dashboard/tests/page-models/create-workflow-sidebar.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export class CreateWorkflowSidebar {
+  constructor(private page: Page) {}
+
+  async createBtnClick({ awaitResponse = false } = {}): Promise<void> {
+    const createWorkflowBtn = await this.page.getByRole('button', { name: 'Create workflow' });
+    await createWorkflowBtn.click();
+
+    if (awaitResponse) {
```
SokratisVidros: Is this parameter needed? Shouldn't we always await for the response?

apps/dashboard/tests/page-models/create-workflow-sidebar.ts
```diff
@@ -0,0 +1,70 @@
...
+      await this.page.waitForResponse(
+        (resp) => resp.url().includes('/v2/workflows') && resp.request().method() === 'POST' && resp.status() === 201
+      );
+    }
+  }
+
+  async expectNameValidationError(): Promise<void> {
+    await expect(await this.page.getByText('Name is required')).toBeVisible();
```
SokratisVidros: Page object models are usually selectors and request helpers that keep the e2e code dry. To me, expectations belong to the test. Moreover, E2E tests have to be readable and self-explanatory. Dryness is not the most critical requirement.

So, I suggest adding all the expectations in the e2e file and using the page object models for selectors, event listeners and waiting logic. This ensures that page object models are highly reusable and can work with any expectation the e2e will require.

Most importantly, when the e2e is read by the future maintainer, they don't have to jump into every function to see what it does.

apps/dashboard/tests/page-models/create-workflow-sidebar.ts
```diff
@@ -0,0 +1,70 @@
...
+    workflowId: string;
+    workflowDescription: string;
+    tags: string[] | number;
+  }): Promise<void> {
+    // fill the workflow name
+    await this.page.locator('input[name="name"]').fill(workflowName);
+    const workflowIdInput = await this.page.locator('input[name="workflowId"]');
+
+    // check the workflow id
+    await expect(await workflowIdInput.inputValue()).toEqual(workflowId);
```
SokratisVidros: Following on my previous comment, this is a good example of overloading a page object function with expectation logic that belongs to the test.

apps/dashboard/tests/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
```
SokratisVidros, line 1: Is the extra tsconfig.json required?

apps/dashboard/tests/utils/session.ts
```diff
@@ -0,0 +1,54 @@
...
+   * Most Playwright tests, create a sessions using some utility functions. The session
+   * is injected in the test, but also needs to be injected in the browser storage, so that
+   * the app can work as expected. Currently, the apps shares token and environment information
+   * between React hooks and the api.client.ts via the localStorage. This needs to be revised
+   * in favor of an in-memory approach.
+   */
+  await page.addInitScript((currentSession) => {
+    window.addEventListener('DOMContentLoaded', () => {
+      localStorage.setItem('nv_auth_token', currentSession.token);
+      localStorage.setItem('nv_last_environment_id', currentSession.environment._id);
```
SokratisVidros: This is no longer required in the new dashboard. We can safely remove it.

libs/testing/src/ee/ee.repository.factory.ts
```diff
@@ -35,29 +35,43 @@ export function getEERepository<T>(className: 'OrganizationRepository' | 'Member
...
 
-function getEEMemberRepository() {
+export function getEEMemberRepository({ mockClerkClient = true }: { mockClerkClient?: boolean } = {}) {
   // nx-ignore-next-line
   const enterpriseModule = require('@novu/ee-auth');
   const EEMemberRepository = enterpriseModule?.EEMemberRepository;
-  const ClerkClientMock = enterpriseModule?.ClerkClientMock;
 
-  return new EEMemberRepository(new CommunityOrganizationRepository(), new ClerkClientMock());
+  if (mockClerkClient) {
```
SokratisVidros: Please refer to my previous comment about creating a Clerk user before each test run and doing an actual sign-in so that we can test the actual interaction between Clerk and the application, not a mocked version of it.

SokratisVidros (changes requested):

apps/dashboard/tests/page-models/create-workflow-sidebar.ts
LetItRock: it's used to await for the server response, and in case of validation errors it's not called

apps/dashboard/tests/page-models/create-workflow-sidebar.ts
LetItRock: I agree.

apps/dashboard/tests/tsconfig.json
LetItRock: yes it is for the tests to treat them as `commonjs` module, otherwise the `@novu/testing` package won't work

apps/dashboard/tests/utils/session.ts
LetItRock: partially, the `nv_auth_token` is used in tests setup

apps/dashboard/tests/global.setup.ts
LetItRock: I'll think about this, as it will require reimplementing all the current setup.

apps/dashboard/tests/package.json
SokratisVidros: 😢

apps/dashboard/tests/utils/session.ts
```diff
@@ -0,0 +1,53 @@
...
+   *
+   * Most Playwright tests, create a sessions using some utility functions. The session
+   * is injected in the test, but also needs to be injected in the browser storage, so that
+   * the app can work as expected. Currently, the apps shares token and environment information
+   * between React hooks and the api.client.ts via the localStorage. This needs to be revised
+   * in favor of an in-memory approach.
+   */
+  await page.addInitScript((currentSession) => {
+    window.addEventListener('DOMContentLoaded', () => {
+      localStorage.setItem('nv_auth_token', currentSession.token);
```
SokratisVidros: Do you know if this is necessary in Dashboard v2? If not mistaken this is V1 logic for the community auth.

.github/workflows/reusable-dashboard-e2e.yml
```diff
@@ -68,6 +68,30 @@ jobs:
...
 
+      - name: Create .env file for the Dashboard app
+        working-directory: apps/dashboard
+        run: |
+          touch .env
+          echo VITE_LAUNCH_DARKLY_CLIENT_SIDE_ID=${{ secrets.LAUNCH_DARKLY_CLIENT_SIDE_ID }} >> .env
+          echo VITE_API_HOSTNAME=http://127.0.0.1:1336 >> .env
+          echo VITE_WEBSOCKET_HOSTNAME=http://127.0.0.1:1340 >> .env
+          echo VITE_LEGACY_DASHBOARD_URL=http://127.0.0.1:4200 >> .env
+          echo VITE_CLERK_PUBLISHABLE_KEY=${{ secrets.CLERK_E2E_PUBLISHABLE_KEY }} >> .env
```
LetItRock, line 79: .env file for the Dashboard

.github/workflows/reusable-dashboard-e2e.yml
```diff
@@ -68,6 +68,30 @@ jobs:
...
+          touch .env.playwright
+          echo NOVU_ENTERPRISE=true >> .env.playwright
+          echo NEW_RELIC_ENABLED=false >> .env.playwright
+          echo NEW_RELIC_APP_NAME=Novu >> .env.playwright
+          echo MONGO_URL=mongodb://127.0.0.1:27017/novu-test >> .env.playwright
+          echo API_URL=http://127.0.0.1:1336 >> .env.playwright
+          echo CLERK_ENABLED=true >> .env.playwright
+          echo CLERK_PUBLISHABLE_KEY=${{ secrets.CLERK_E2E_PUBLISHABLE_KEY }} >> .env.playwright
+          echo CLERK_SECRET_KEY=${{ secrets.CLERK_E2E_SECRET_KEY }} >> .env.playwright
+          echo NODE_ENV=test >> .env.playwright
```
LetItRock, line 93: .env file for the playwright, used to initialize the session, create, and sync users/orgs

apps/dashboard/package.json
```diff
@@ -105,9 +106,15 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "@clerk/backend": "^1.6.2",
+    "@clerk/testing": "^1.3.27",
```
LetItRock, line 110: used to create Clerk user/orgs and to sign-in into the app

apps/dashboard/src/components/primitives/tag-input.tsx
```diff
@@ -82,8 +82,20 @@ const TagInput = forwardRef<HTMLInputElement, TagInputProps>((props, ref) => {
...
           <div className="flex flex-wrap gap-2">
             {tags.map((tag, index) => (
-              <Tag key={index} variant="stroke" onDismiss={() => removeTag(tag)}>
-                <span style={{ wordBreak: 'break-all' }}>{tag}</span>
+              <Tag
+                key={index}
+                variant="stroke"
+                onDismiss={(e) => {
+                  e.preventDefault();
+                  e.stopPropagation();
```
LetItRock, line 95: fix the bug - the click on remove button was causing the form submission

apps/dashboard/src/components/workflow-editor/base-node.tsx
```diff
@@ -80,7 +80,7 @@ export const NodeBody = ({
 
   return (
     <HoverCard openDelay={300}>
-      <HoverCardTrigger>
+      <HoverCardTrigger asChild>
```
LetItRock, line 83: react warning `a` nested in `a`

apps/dashboard/src/hooks/use-on-element-unmount.ts
```diff
@@ -25,10 +25,6 @@ export const useOnElementUnmount = (props: { callback: () => void; condition: bo
       });
 
       observer.observe(element.parentNode!, { childList: true });
-
-      return () => {
-        observer.disconnect();
-      };
```
LetItRock, line 31: react warning on that the callback is never called

apps/dashboard/tests/manage-workflows.e2e.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  // check the trigger workflow page
+  const triggerWorkflowPage = new TriggerWorkflowPage(page);
+  await expect(page).toHaveTitle(`Trigger ${workflowNameUpdated} | Novu Cloud Dashboard`);
+
+  // trigger the workflow
+  await triggerWorkflowPage.triggerWorkflowBtnClick();
+  const activityPanel = await triggerWorkflowPage.getActivityPanel();
+  await expect(activityPanel).toBeVisible();
+  await expect(activityPanel.locator('span').filter({ hasText: workflowNameUpdated })).toBeVisible();
```
SokratisVidros, line 142: I suggest we run more assertions on the trigger page to ensure that the status is updated while running.

apps/dashboard/tests/package.json
SokratisVidros: @LetItRock Since we wrote dedicated E2E helpers in the Dashboard, do we need @novu/testing afterall?

apps/dashboard/tests/manage-workflows.e2e.ts
```diff
@@ -0,0 +1,172 @@
...
+import { test } from './utils/fixtures';
+import { InAppStepEditor } from './page-object-models/in-app-step-editor';
+import { WorkflowsPage } from './page-object-models/workflows-page';
+import { CreateWorkflowSidebar } from './page-object-models/create-workflow-sidebar';
+import { WorkflowEditorPage } from './page-object-models/workflow-editor-page';
+import { StepConfigSidebar } from './page-object-models/step-config-sidebar';
+import { TriggerWorkflowPage } from './page-object-models/trigger-workflow-page';
+
+test('manage workflows', async ({ page }) => {
+  const workflowName = 'test-workflow';
```
SokratisVidros, line 13: How about starting the flow from sign-up so that we test the onboarding?

SokratisVidros (approved):

apps/dashboard/tests/utils/session.ts
LetItRock: it's the leftover from the old implementation, will remove it

apps/dashboard/tests/manage-workflows.e2e.ts
LetItRock: I'll do that in the PR that I've opened for the activity panel fix: https://github.com/novuhq/novu/pull/7768

apps/dashboard/tests/package.json
LetItRock: the `@novu/testing` is not needed, I made it in a way that the tests setup is independent of it

apps/dashboard/tests/manage-workflows.e2e.ts
LetItRock: we should do it in a separate test as the `test` fixture is doing all the user setup logic

PR #7127, opened 2024-11-25
feat(dashboard): Nv 4800 implement arrays and nested forms in custom step controls

BiswaViraj:
### Screenshots

https://github.com/user-attachments/assets/a6d444ec-294e-403f-8023-cc66c5bc9e2f

<details>

</details>


apps/dashboard/src/components/workflow-editor/steps/controls/array-field-item-template.tsx
BiswaViraj, line 1: for the Array went with the UI similar to the objects instead of using the tag-input as it will make it complex to handle different scenarios and also doesn't play well with the rjsf form

apps/dashboard/src/components/workflow-editor/steps/controls/array-field-template.tsx
```diff
@@ -0,0 +1,91 @@
...
+            <ArrayFieldTitleTemplate
+              idSchema={idSchema}
+              title={uiOptions.title || title}
+              schema={schema}
+              uiSchema={uiSchema}
+              required={required}
+              registry={registry}
+            />
+          </span>
+          <div className="bg-background text-foreground-600 -mt-px ml-auto mr-4 flex items-center gap-1">
```
rifont, line 66: Could we add a hover state to the collapsible button to be consistent with the add/minus button hover states?

It's generally a good UX practice for all buttons to have a hover state to provide feedback to the user that they are interacting with something (in addition to varying the pointer)

apps/dashboard/src/utils/schema.ts
```diff
@@ -167,3 +167,31 @@ export const buildDefaultValues = (uiSchema: UiSchema): object => {
 
   return keys;
 };
+
+export const buildDefaultValuesOfDataSchema = (dataSchema: JSONSchemaDto) => {
```
LetItRock: This function is a copy of the above one but the arg is just a different type. Can we adjust the above one?

apps/dashboard/src/components/workflow-editor/steps/other-steps-tabs.tsx
```diff
@@ -16,25 +18,29 @@ import { ToastIcon } from '@/components/primitives/sonner';
...
 const tabsContentClassName = 'h-full w-full px-3 py-3.5 overflow-y-auto';
 
 export const OtherStepTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; step: StepDataDto }) => {
   const { stepSlug = '' } = useParams<{ workflowSlug: string; stepSlug: string }>();
-  const { dataSchema, uiSchema } = step.controls;
+  const { dataSchema, uiSchema, values } = step.controls;
   const navigate = useNavigate();
   const schema = buildDynamicZodSchema(dataSchema ?? {});
+  const newFormValues = useMemo(
+    () => merge(buildDefaultValues(uiSchema ?? {}), buildDefaultValuesOfDataSchema(dataSchema ?? {}), values),
```
LetItRock: Same here, shouldn't we just use the `dataSchema` defaults in this case?

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -39,18 +39,22 @@ export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; s
 
   const { dataSchema, uiSchema, values } = step.controls;
   const schema = useMemo(() => buildDynamicZodSchema(dataSchema ?? {}), [dataSchema]);
-  const newFormValues = useMemo(() => merge(buildDefaultValues(uiSchema ?? {}), values), [uiSchema, values]);
+  const newFormValues = useMemo(
+    () => merge(buildDefaultValues(uiSchema ?? {}), buildDefaultValuesOfDataSchema(dataSchema ?? {}), values),
```
LetItRock: Do we need to merge the `uiSchema` defaults with `dataSchema` defaults? 
Here, we should only care about the `uiSchema` defaults.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -39,18 +39,22 @@ export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; s
...
+    [uiSchema, values, dataSchema]
+  );
 
   const form = useForm({
     mode: 'onSubmit',
     resolver: zodResolver(schema),
     values: newFormValues,
+    defaultValues: newFormValues,
```
LetItRock: Any reason why we need the default values?

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -171,7 +175,8 @@ export const InAppTabs = ({ workflow, step }: { workflow: WorkflowResponseDto; s
     [formValues]
   );
 
-  const blocker = useBlocker(formState.isDirty || isPending);
+  const isNotDirty = Object.keys(formState.dirtyFields).length === 0;
+  const blocker = useBlocker(isNotDirty || isPending);
```
LetItRock: This blocker stops me from closing the drawer, because the form is not dirty on the first load.

https://github.com/user-attachments/assets/0f7702f5-110e-4653-b758-54e57447e4b5

LetItRock (changes requested):
I found a few things:

1. Inconsistencies in default state:
Old Web app state:
Screenshot 2024-11-28 at 01 48 31

New Dashboard:
Screenshot 2024-11-28 at 01 49 52

2. Can't remove the number value, but it's allowed in old Web.

apps/dashboard/src/utils/schema.ts
BiswaViraj: Modified it to handle nested defs

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
BiswaViraj: This is leftover from the rebase, updated it now to reflect the changes on next branch

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
BiswaViraj: ok yes, should be `!isNotDirty`

apps/dashboard/src/components/icons/inbox-bell-filled.tsx
BiswaViraj, line 1: Unrelated but fixed this small error that was always in the browser console
image

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -35,16 +35,20 @@ export const ConfigureStepTemplateForm = (props: ConfigureStepTemplateFormProps)
...
-    []
-  );
+  const defaultValues = useMemo(() => {
+    // Use the UI Schema to build the default values if it exists else use the data schema (code-first approach) values
+    if (Object.keys(step.controls.uiSchema ?? {}).length !== 0) {
+      return merge(buildDefaultValues(step.controls.uiSchema ?? {}), step.controls.values);
+    }
+
+    return merge(buildDefaultValuesOfDataSchema(step.controls.dataSchema ?? {}), step.controls.values);
+  }, []);
```
BiswaViraj: This updates the form to have the default control values from the code-first step controls if available

SokratisVidros: There is a TODO for the API to generate the default values from the API for all workflows, not just the one defined in Dashboard, so we don't have to do these calculations in the FE.

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -55,7 +59,14 @@ export const ConfigureStepTemplateForm = (props: ConfigureStepTemplateFormProps)
   };
 
   useEffect(() => {
-    form.reset(merge(buildDefaultValues(step.controls.uiSchema ?? {}), step?.controls.values));
+    // Use the UI Schema to build the default values if it exists else use the data schema (code-first approach) values
+    const defaultValues =
+      Object.keys(step.controls.uiSchema ?? {}).length !== 0
```
SokratisVidros: @BiswaViraj Since we calculated the value from the useMemo, do we have to calculate it again?

apps/dashboard/src/components/workflow-editor/steps/controls/text-widget.tsx
```diff
@@ -1,34 +1,64 @@
...
-                fontFamily="inherit"
-                placeholder={capitalize(label)}
-                id={label}
-                extensions={[autocompletion({ override: [completions([])] })]}
-                value={field.value}
-                onChange={(val) => field.onChange(val)}
-                readOnly={readonly}
-              />
+            <InputField size="fit" className="w-full">
+              {isNumberType ? (
```
SokratisVidros, line 33: Do we support other field types such as email or tel?

apps/dashboard/src/components/workflow-editor/steps/controls/text-widget.tsx
```diff
@@ -1,34 +1,64 @@
...
+                <Input
```
SokratisVidros, line 34: This means that there is no autocomplete for variables if the field is a number. Should we maybe always use the editor for now?

SokratisVidros (changes requested):

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
desiprisg: step could have been updated since then. memo only calculates once for the defaults which are cached. We can simplify with a function though.

apps/dashboard/src/components/workflow-editor/steps/controls/text-widget.tsx
BiswaViraj: The problem here is, the whole validation is handled by the user via the framework control schema.
If they will do a `numberFIeld: z.number()` then allowing them to write `{{payload.value}}` on UI will always fail as the zodSchema will expect it to be a number on UI

desiprisg (approved):

apps/dashboard/src/components/workflow-editor/steps/controls/text-widget.tsx
BiswaViraj: Everything is controlled by user via the schema
image
but this also faces the same problem like the number field
user defines `emailField: z.email()` and then types the `payload` this will error out coz of zod schema validation

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
BiswaViraj: The useMemo one is for form initailisation on mount
this one whenever the form is changed

SokratisVidros: Thanks for the explanation. In this case, we need to wrap this logic into a function and reuse it.

SokratisVidros (approved):

PR #7131, opened 2024-11-26
feat(dashboard): add delay step

ChmaraX:

Current know issues:
- [visual] the error border is missing on the amount field (will be done in separate PR)

https://github.com/user-attachments/assets/1b324235-4303-4fac-aa9b-1a419675418f


apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,96 @@
...
+  inputName: string;
+  selectName: string;
+  options: string[];
+  defaultOption?: string;
+  className?: string;
+  placeholder?: string;
+  isReadOnly?: boolean;
+};
+
+export const NumberInputWithSelect = (props: InputWithSelectProps) => {
```
ChmaraX: I did this as a simple component for now - with some effort it could be probably more generic, but I'm not sure yet if it needs to be more generic - its difficult to do abstractions now since we're yet to see how or if will the digest reuse this.

apps/dashboard/src/components/workflow-editor/add-step-menu.tsx
```diff
@@ -121,7 +121,13 @@ export const AddStepMenu = ({
...
               <MenuItemsGroup>
                 <MenuItem stepType={StepTypeEnum.DIGEST}>Digest</MenuItem>
-                <MenuItem stepType={StepTypeEnum.DELAY}>Delay</MenuItem>
+                <MenuItem
+                  stepType={StepTypeEnum.DELAY}
+                  disabled={true}
+                  onClick={() => handleMenuItemClick(StepTypeEnum.DELAY)}
+                >
+                  Delay
+                </MenuItem>
```
ChmaraX: We don't have launch darkly setup on new dashboard yet - this is just visually disabled but the button can be clicked - will be done in separate ticket.

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
```diff
@@ -0,0 +1,152 @@
...
+  classNames: { toast: 'ml-10 mb-4' },
+};
+
+const delayControlsSchema = z
+  .object({
+    type: z.enum(['regular']).default('regular'),
+    amount: z.number(),
+    unit: z.nativeEnum(TimeUnitEnum),
+  })
+  .strict();
```
ChmaraX: Copied from API for now to have some typing, we need a way to have schemas shared

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
```diff
@@ -0,0 +1,152 @@
...
+            <FormLabel tooltip="Delays workflow for the set time, then proceeds to the next step.">
+              Delay execution by
+            </FormLabel>
+            <NumberInputWithSelect
+              form={form}
+              inputName="amount"
+              selectName="unit"
+              options={unitOptions}
+              isReadOnly={isReadOnly}
+            />
```
ChmaraX, line 140: Should this be made like components in InApp with the mapping and stuff? e.g.:

```jsx
  const {
    [avatarKey]: avatar,
    [subjectKey]: subject,
    [bodyKey]: body,
    [redirectKey]: redirect,
    [primaryActionKey]: primaryAction,
    [secondaryActionKey]: secondaryAction,
  } = uiSchema.properties ?? {};
  ```

Seems like an abstraction overkill for one simple input field, also how we would map these 3 ui schema components:

```jsx
    "uiSchema": {
                "group": "DELAY",
                "properties": {
                    "amount": {
                        "component": "DELAY_AMOUNT",
                        "placeholder": "30"
                    },
                    "unit": {
                        "component": "DELAY_UNIT",
                        "placeholder": "seconds"
                    },
                    "type": {
                        "component": "DELAY_TYPE",
                        "placeholder": null
                    }
                }
            },
  ```

File: apps/dashboard/src/components/workflow-editor/workflow-canvas.tsx [hunk 5]
```diff
@@ -59,7 +59,7 @@ const mapStepToNode = (
 ): Node<NodeData, keyof typeof nodeTypes> => {
   let content = '';
   if (step.type === StepTypeEnum.DELAY) {
-    content = `Wait to send ~ 30 minutes`;
+    content = `Delay sending for a specified duration`;
```
ChmaraX: Had to change this to something that doesnt specify the time.

If we want to make this react to the actual delay duration we would need `controlValues` in the workflow.steps response.

libs/application-generic/src/usecases/tier-restrictions-validate/tier-restrictions-validate.usecase.ts
```diff
@@ -69,7 +69,7 @@ export class TierRestrictionsValidateUsecase {
         issues.push({
           error: ErrorEnum.TIER_LIMIT_EXCEEDED,
           message:
-            `The maximum delay allowed is ${FREE_TIER_MAX_DELAY_DAYS} days.` +
+            `The maximum delay allowed is ${FREE_TIER_MAX_DELAY_DAYS} days. ` +
```
ChmaraX, line 74: Kept this as it is for now but I would suggest making this shorter and adding a link to billing page.

image

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
```diff
@@ -0,0 +1,137 @@
...
+        <Form {...form}>
+          <form
+            className="flex h-full flex-col gap-2"
+            onChange={(e) => {
+              e.preventDefault();
+              e.stopPropagation();
+              debouncedSave(form.getValues());
+            }}
+          >
+            <DelayAmount dataSchema={dataSchema ?? {}} isReadOnly={isReadOnly} />
```
ChmaraX: Should this be made like components in InApp with the mapping and stuff? e.g.:

```jsx
  const {
    [avatarKey]: avatar,
    [subjectKey]: subject,
    [bodyKey]: body,
    [redirectKey]: redirect,
    [primaryActionKey]: primaryAction,
    [secondaryActionKey]: secondaryAction,
  } = uiSchema.properties ?? {};
  ```

Seems like an abstraction overkill for one simple input field, also how we would map these 3 ui schema components to single component - the abstraction works for the editor fields in other steps but I'm not sure about it here:

```jsx
    "uiSchema": {
                "group": "DELAY",
                "properties": {
                    "amount": {
                        "component": "DELAY_AMOUNT",
                        "placeholder": "30"
                    },
                    "unit": {
                        "component": "DELAY_UNIT",
                        "placeholder": "seconds"
                    },
                    "type": {
                        "component": "DELAY_TYPE",
                        "placeholder": null
                    }
                }
            },
  ```

File: apps/dashboard/src/components/primitives/form/form.tsx [hunk 8]
```diff
@@ -120,8 +120,8 @@ const FormMessagePure = React.forwardRef<
       className={formMessageVariants({ variant: error ? 'error' : 'default', className })}
       {...props}
     >
-      {error ? <RiErrorWarningFill className="size-4" /> : <RiInformationFill className="size-4" />}
-      <span className="mt-[1px] text-xs leading-3">{body}</span>
+      <span>{error ? <RiErrorWarningFill className="size-4" /> : <RiInformationFill className="size-4" />}</span>
+      <span className="mt-[1px] text-xs leading-4">{body}</span>
```
ChmaraX, line 124: The icon is still a bit suspicious - it would be ok if the content would be shorter.

Before:

image

After:

image

libs/application-generic/src/usecases/tier-restrictions-validate/tier-restrictions-validate.usecase.ts
ChmaraX: Something like: 
_The maximum delay allowed is 90 days. [Please upgrade your plan.](url)_

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
```diff
@@ -0,0 +1,136 @@
...
+  .object({
+    type: z.enum(['regular']).default('regular'),
+    amount: z.number(),
+    unit: z.nativeEnum(TimeUnitEnum),
+  })
+  .strict();
+
+export const DelayConfigure = () => {
+  const { stepSlug = '' } = useParams<{ workflowSlug: string; stepSlug: string }>();
+  const { step, refetch } = useStepEditorContext();
```
ChmaraX: we have 2 types of step :

```jsx
1. const { step } = useStep()
2. const { step } = useStepEditorContext() 
```
The 1st one refetches when workflow is updated (desired I would say), the 2nd one doesn't until you call `refetch()`.  Could we have this behaviour here as well or its intended?

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <>
+      <InputFieldPure className="h-7 rounded-lg border pr-0">
+        <FormField
+          control={control}
+          name={inputName}
+          render={({ field }) => (
+            <FormItem className="w-full overflow-hidden">
+              <FormControl>
+                <Input
+                  type="number"
```
LetItRock: We should not allow values less than 1 and this should come from BE `dataSchema`, but we can't have the max value because it will be different for different time units and I'm not sure if it's possible to define that with JsonSchema.
Please also ping the BE guys about the validation on their side.

Screenshot 2024-11-26 at 17 21 00

LetItRock: Also, with type number, I can erase the last character, which results in this weird state:
Screenshot 2024-11-26 at 17 26 02
The form schema should actually validate the input, check if it's a number, and show the corresponding error.

LetItRock: Also, when the field is focused, we should have a black border around the input.

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const handleChange = (value: { input: number; select: string }) => {
+    // we want to always set both values and treat it as a single input
+    setValue(inputName, value.input, { shouldDirty: true });
+    setValue(selectName, value.select, { shouldDirty: true });
+  };
+
+  return (
+    <>
+      <InputFieldPure className="h-7 rounded-lg border pr-0">
```
LetItRock: there should be a red border around if we have the validation error

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,97 @@
...
+  selectName: string;
+  options: string[];
+  defaultOption?: string;
+  className?: string;
+  placeholder?: string;
+  isReadOnly?: boolean;
+};
+
+export const NumberInputWithSelect = (props: InputWithSelectProps) => {
+  const { className, inputName, selectName, options, defaultOption, placeholder, isReadOnly } = props;
```
LetItRock: Regarding `inputName, selectName` we came up with the `fields` prop for multiple form keys, please take a look into `URLInput`.

apps/dashboard/src/components/primitives/form/form.tsx
LetItRock: ~Please talk to the BE team and ask for a shorter message.~ ok saw another comment that you updated it

apps/dashboard/src/components/workflow-editor/steps/configure-step-content.tsx
```diff
@@ -21,23 +22,33 @@ export const ConfigureStepContent = () => {
     [step]
   );
 
+  if (!step?.type) {
+    return null;
+  }
```
LetItRock: why this is needed?

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
LetItRock: The schema is dynamic that's why we didn't have types for the in-app. I feel in this case we can also remove it.

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
LetItRock: The thing is that the first one doesn't have step controls, so let's use the second for now.

apps/dashboard/src/components/workflow-editor/steps/delay/delay-configure.tsx
LetItRock: Let's follow the same approach we had with `in-app` and display the delay fields based on the `uiSchema.group` property and assume for now that we will render the delay input when there are `amount, unit` properties.

LetItRock: Also added the ticket to render custom controls.

apps/dashboard/src/components/workflow-editor/workflow-canvas.tsx
LetItRock: please let Sok and Radek know

apps/api/src/app/workflows-v2/shared/schemas/delay-control.schema.ts
```diff
@@ -26,15 +26,15 @@ export const delayUiSchema: UiSchema = {
   properties: {
     amount: {
       component: UiComponentEnum.DELAY_AMOUNT,
-      placeholder: '30',
```
ChmaraX, line 29: We probably can't have it like this, see [this thread](https://novu.slack.com/archives/C07HVLG7G2U/p1732895403877039?thread_ts=1732887674.583079&cid=C07HVLG7G2U) for more info.

apps/dashboard/src/components/number-input-with-select.tsx
ChmaraX: Validation was added.

The border issue is a bit complicated due to combination of 2 elements - will try to solve it in separate PR since its a visual issue and I dont want to block this PR to avoid further conflicts etc

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
```diff
@@ -27,7 +29,25 @@ export const StepProvider = ({ children }: { children: ReactNode }) => {
...
+   */
+  const issues = useMemo(() => {
+    const newIssues = workflow?.steps.find(
+      (s) =>
+        getEncodedId({ slug: s.slug, divider: STEP_DIVIDER }) ===
+        getEncodedId({ slug: stepSlug, divider: STEP_DIVIDER })
+    )?.issues;
+
+    return { ...newIssues };
+  }, [workflow, stepSlug]);
```
ChmaraX: This is a hack that was present in code before - the root cause for this is that we update steps with PUT workflow endpoint using `WorkflowProvider` and we don't refetch steps (from GET step endpoint) afterwards.

The proper fix is described in the comment here. Currently, step and workflow are not entirely separate resources on our client which causes various issues.

libs/application-generic/src/usecases/tier-restrictions-validate/tier-restrictions-validate.usecase.ts
```diff
@@ -71,8 +71,8 @@ export class TierRestrictionsValidateUsecase {
         issues.push({
           error: ErrorEnum.TIER_LIMIT_EXCEEDED,
           message:
-            `The maximum delay allowed is ${FREE_TIER_MAX_DELAY_DAYS} days.` +
-            'Please contact our support team to discuss extending this limit for your use case.',
+            `The maximum delay allowed is ${FREE_TIER_MAX_DELAY_DAYS} days. ` +
+            'Please upgrade your plan.',
```
ChmaraX, line 75: I will open a PR to be able to add HTML here so frontend can parse it as an URL to billing page or something like that.

apps/dashboard/src/components/workflow-editor/steps/configure-inline/configure-step-inline-form.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+type ConfigureStepInlineFormProps = StepInlineFormProps & {
+  issues?: StepIssuesDto;
+  update: (data: UpdateWorkflowDto) => void;
+  updateStepCache: (step: Partial<StepDataDto>) => void;
+};
+
+export const ConfigureStepInlineForm = (props: ConfigureStepInlineFormProps) => {
+  const { workflow, step, issues, update, updateStepCache } = props;
```
ChmaraX: we are drilling the props quite deep, we should think about using the hooks here, but I left it like this for consistency.

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
```diff
@@ -2,33 +2,54 @@ import { createContext, useMemo, type ReactNode } from 'react';
...
```
ChmaraX, line 48: The fact that we update both step and workflow with one endpoint, but fetch the updated data from 2 endpoints causes a lot of issues regarding stale data, caching etc etc. 

We should fix this asap by fetching everything from workflows response as I suggested here in the comment or some other solution.

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,104 @@
...
+    selectKey: string;
+  };
+  options: string[];
+  defaultOption?: string;
+  className?: string;
+  placeholder?: string;
+  isReadOnly?: boolean;
+};
+
+export const NumberInputWithSelect = ({
```
SokratisVidros:
```suggestion
export const DurationInput = ({
```

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,104 @@
+import { cn } from '@/utils/ui';
+import { FormControl, FormField, FormItem, FormMessagePure } from '@/components/primitives/form/form';
+import { Input } from '@/components/primitives/input';
+import { InputFieldPure } from '@/components/primitives/input';
+import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/primitives/select';
+import { useFormContext } from 'react-hook-form';
+import { useMemo } from 'react';
+
+type InputWithSelectProps = {
+  fields: {
```
SokratisVidros: Nested in props is always 👿 . how about using two flat props?

1. `unit: 'seconds' | 'minutes' | 'hours' | 'months'` 
2. `amount: number`

This should simplify the logic of the handleChange a lot.

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,104 @@
...
+    inputKey: string;
+    selectKey: string;
+  };
+  options: string[];
+  defaultOption?: string;
```
SokratisVidros: Do we need the default option prop? The unit prop should be the default option otherwise its `defaultOption[0]`.

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
```diff
@@ -2,33 +2,54 @@ import { createContext, useMemo, type ReactNode } from 'react';
...
 import { createContextHook } from '@/utils/context';
 import { Step } from '@/utils/types';
+import { STEP_DIVIDER } from '@/utils/step';
+import { getEncodedId } from '@/utils/step';
+import { useWorkflow } from '@/components/workflow-editor/workflow-provider';
 
 export type StepEditorContextType = {
   isPending: boolean;
   step?: StepDataDto;
-  refetch: (options?: RefetchOptions) => Promise<QueryObserverResult<StepDataDto, Error>>;
```
ChmaraX, line 13: unused

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export const NumberInputWithSelect = ({
+  fields,
+  options,
+  defaultOption,
+  className,
+  placeholder,
+  isReadOnly,
+}: InputWithSelectProps) => {
+  const { getFieldState, setValue, getValues, control } = useFormContext();
```
SokratisVidros: I suggest making the DurationInput a primitive and then wiring it to the RHF either in workflow editor or with a wrapper component under /components/workflow-editor if necessary.

It will give us way more flexibility and make the component reusable.

apps/dashboard/src/components/number-input-with-select.tsx
```diff
@@ -0,0 +1,104 @@
...
+                      </SelectItem>
+                    ))}
+                  </SelectContent>
+                </Select>
+              </FormControl>
+            </FormItem>
+          )}
+        />
+      </InputFieldPure>
+      <FormMessagePure error={error ? String(error.message) : undefined} />
```
SokratisVidros:
```suggestion
      {error && <FormMessagePure error='error.message' />}
```

apps/dashboard/src/components/workflow-editor/add-step-menu.tsx
```diff
@@ -41,7 +43,7 @@ const MenuItem = ({
 
   return (
     <span
-      onClick={onClick}
+      onClick={!disabled ? onClick : undefined}
```
SokratisVidros: A more neat approch could be defining `const noop = () => {}` at the top of the file outside of the component. Then: 

```suggestion
      onClick={disabled ? noop : onClick }
```

apps/dashboard/src/components/number-input-with-select.tsx
SokratisVidros: The wrapper component already exists. See [here](https://github.com/novuhq/novu/pull/7131/files#diff-2232d48a43ebf90e611f34a4f12b354dc571e6640164e03a8d62f78ce4950b76R1).

apps/dashboard/src/components/workflow-editor/workflow-canvas.tsx
SokratisVidros:
```suggestion
    content = `Delay execution for a specified duration`;
```

apps/dashboard/src/components/workflow-editor/steps/configure-inline/configure-step-inline-form.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { StepTypeEnum } from '@novu/shared';
+
+import { flattenIssues, updateStepControlValuesInWorkflow } from '@/components/workflow-editor/step-utils';
+import { buildDefaultValues, buildDefaultValuesOfDataSchema, buildDynamicZodSchema } from '@/utils/schema';
+import { Form } from '@/components/primitives/form/form';
+import { useFormAutosave } from '@/hooks/use-form-autosave';
+import { SaveFormContext } from '@/components/workflow-editor/steps/save-form-context';
+import { DelayForm } from '@/components/workflow-editor/steps/delay/delay-form';
+
+const STEP_TYPE_TO_INLINE_FORM: Record<StepTypeEnum, (args: StepInlineFormProps) => React.JSX.Element | null> = {
```
SokratisVidros: This code along with `calculateDefaultValues` is duplicated between this new component and `configure-step-template-form.tsx` We need to make it dry.

SokratisVidros (changes requested):
I strongly suggest decoupling the form logic from the form wrapper. That is, the form of every step editor should work both in the first side bar or the second sidebar. 

Currently, I noticed a lot of duplication between the template form and the inline form. Introducing a StepForm component to host the form handling that is reused from the different sidebar containers should simplify the logic drastically.

apps/dashboard/src/components/number-input-with-select.tsx
LetItRock: @SokratisVidros this is what we agreed before with @desiprisg and we use this pattern in other places where the component is using two form fields

apps/dashboard/src/components/number-input-with-select.tsx
LetItRock: I'm not sure we can do something about it because this component uses the two form fields, check the `FormField` usage.

apps/dashboard/src/components/workflow-editor/add-step-menu.tsx
```diff
@@ -73,6 +75,7 @@ export const AddStepMenu = ({
   onMenuItemClick: (stepType: StepTypeEnum) => void;
 }) => {
   const [isPopoverOpen, setIsPopoverOpen] = useState(false);
+  const isDelayDigestEmailEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_ND_DELAY_DIGEST_EMAIL_ENABLED);
```
LetItRock: what `ND` means? new dashboard?

apps/dashboard/src/components/workflow-editor/steps/configure-inline/configure-step-inline-form.tsx
LetItRock: I agree.
Also, it will result in the form nested under another form, which we wanted to avoid.
The `configure-step-form` should just register all the fields under the `controls` in the form I guess, because this is the shape of the `step` and that will be updated later with using the `update workflow` endpoint.
```
{
   "name": "In-App Step",
   "stepId": "some-step-id",
   "controls": { ... } 
}

Comment (LetItRock) [reply apps/dashboard/src/components/workflow-editor/steps/configure-inline/configure-step-inline-form.tsx]: Later here we will also need to connect `custom step controls` from the code created workflows.

File: apps/dashboard/src/components/number-input-with-select.tsx (hunk 19 above)
Comment (ChmaraX) [reply apps/dashboard/src/components/number-input-with-select.tsx]: Its a generic component not necessarily a duration, it can be also `10 horses`.

Lets see how it evolves with the digest and we will incrementally adjust this component, by either making it primitive or whatever we need - I dont want to make it too generic too early since I dont know what will digest need.

File: apps/dashboard/src/components/workflow-editor/add-step-menu.tsx (hunk 27 above)
Comment (ChmaraX) [reply apps/dashboard/src/components/workflow-editor/add-step-menu.tsx]: Yes

File: apps/dashboard/src/components/number-input-with-select.tsx (hunk 23 above)
Comment (ChmaraX) [reply apps/dashboard/src/components/number-input-with-select.tsx]: Lets see how it evolves with the digest and we will incrementally adjust this component, by either making it primitive or whatever we need - I dont want to make it too generic too early since I dont know what will digest need.

File: apps/dashboard/src/components/number-input-with-select.tsx (hunk 21 above)
Comment (ChmaraX) [reply apps/dashboard/src/components/number-input-with-select.tsx]: We dictate the default selected value in ui schema coming from backend, so its needed.

File: apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx [hunk 28]
```diff
@@ -23,70 +24,137 @@ import { Input, InputField } from '@/components/primitives/input';
...
+  );
+
+  const defaultValues = useMemo(
+    () => ({
+      name: step.name,
+      stepId: step.stepId,
+      ...registerInlineControlValues(step),
+    }),
+    [step, registerInlineControlValues]
+  );
```
ChmaraX, line 133: We need to register control values and controls schema here conditionally. I thought maybe it could be done in a separate component like:

```
const form = useFormContext()

form.register(`controlValues`, registerInlineControlValues)
form.resolver = { ...stepSchema, ...controlsSchema} // not possible
....
```
But thats not feasible because:
1.  the resolver can't be changed/extended
2. the `controlValues` cant be registered conditionally because then the `useFormAutosave` -> `data` object must have some type, unless forced to `any`.
3. Bunch of other issues  

If I removed the `isInlineConfigurableStep` condition altogether, then if try to change just name on In-App step, it would trigger validation on `controlValues` as well.

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -23,70 +24,137 @@ import { Input, InputField } from '@/components/primitives/input';
...
-    isReadOnly: isCodeCreatedWorkflow,
+    isReadOnly,
     save: (data) => {
-      update(updateStepInWorkflow(workflow, data));
-      updateStepCache(data);
+      // tranform form fields to step update dto
+      const updateStepData: Partial<StepUpdateDto> = {
+        name: data.name,
+        ...(data.controlValues ? { controlValues: data.controlValues } : {}),
+      };
```
ChmaraX, line 150: We always needs to do these mappings because what we get from API and display in form has different keys than what we send in payload and vice-versa:

- we don't want to send controlValues if they're not changed, otherwise controls validation in InApp step would be triggered on name change
- we dont want to send `stepId` to update payload

libs/application-generic/src/usecases/tier-restrictions-validate/tier-restrictions-validate.usecase.ts
djabarovgeorge: Totally agree here, the copy is too long.

packages/shared/src/types/feature-flags.ts
```diff
@@ -43,4 +43,5 @@ export enum FeatureFlagsKeysEnum {
   IS_CONTROLS_AUTOCOMPLETE_ENABLED = 'IS_CONTROLS_AUTOCOMPLETE_ENABLED',
   IS_USAGE_ALERTS_ENABLED = 'IS_USAGE_ALERTS_ENABLED',
   IS_NEW_DASHBOARD_ENABLED = 'IS_NEW_DASHBOARD_ENABLED',
+  IS_ND_DELAY_DIGEST_EMAIL_ENABLED = 'IS_ND_DELAY_DIGEST_EMAIL_ENABLED',
```
djabarovgeorge: It's not critical, just a suggestion. 🙃
If the ND means a new dashboard, I would add the full name or remove it all together.

apps/dashboard/src/components/workflow-editor/workflow-canvas.tsx
djabarovgeorge: thought: 
i saw all the comments here with Sokratis suggestion, the copy is even longer, and it started to truncate/disappear, i wonder if we should make it shorter.
`Delay action for a set time`?

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
```diff
@@ -2,33 +2,54 @@ import { createContext, useMemo, type ReactNode } from 'react';
...
     workflowSlug,
     stepSlug,
   });
 
+  /**
+   * We need to get the issues from the workflow response
+   * because the step is not re-fetched when workflow is updated
+   *
+   * TODO:
+   * 1. add all step data to workflow response
```
djabarovgeorge, line 37: You mean from the API, right? if this is the case, please let me know what is missing. :)

apps/dashboard/src/components/number-input-with-select.tsx
SokratisVidros: Isn't a flat structure much easier to work with?

apps/dashboard/src/components/number-input-with-select.tsx
SokratisVidros: Folks, I strongly disagree with this approach. We start by pairing all our components with a context, making them stateful, whereas it is obvious they should be stateless, with presentational components wired to their state on each page.

I think of them as our internal UI component library, which is a composition of primitives and composite stateless components. This number input with Select is a composite.

apps/dashboard/src/components/number-input-with-select.tsx
SokratisVidros: The it's an `AmountInput`!

apps/dashboard/src/components/workflow-editor/steps/delay/delay-amount.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { step } = useStep();
+  const { dataSchema, uiSchema } = step?.controls ?? {};
+
+  const unitOptions = useMemo(
+    () => (dataSchema?.properties?.[unitKey] as any).enum ?? defaultUnitValues,
+    [dataSchema?.properties]
+  );
+
+  const defaultUnitOption = useMemo(
+    () => (uiSchema?.properties?.[unitKey] as any).placeholder ?? defaultUnitValues[0],
```
djabarovgeorge: My personal opinion, because defaultUnitValues[0] is less predictable, we could change the order, and we would get months here.

```suggestion
    () => (uiSchema?.properties?.[unitKey] as any).placeholder ?? defaultUnitValues.SECONDS,
```

packages/shared/src/types/feature-flags.ts
SokratisVidros: Correct. We have `IS_NEW_DASHBOARD_ENABLED` env variable so the prefix should be the same.

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
SokratisVidros: @ChmaraX If we need to change the API, let's figure out what we must do to make it easier on the frontend. What's the key difference in this case?

apps/dashboard/src/components/number-input-with-select.tsx
djabarovgeorge: Tend to agree with Adam here, we tend to overengineer a lot lately without the need, i would say whoever will implement Digest should be aware of Delay components and refactor accordingly

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
ChmaraX: @SokratisVidros This would be my rough plan:

1. Add `controls`, or the same values we get from step endpoint, to the workflow steps
2. Remove the current complex logic around step vs workflow on frontend - have only `workflow` as a single frontend entity/resource - we always want to operate on workflow resource
3. align types properly for get, update and workflow forms
4. align property names a bit, if GET returns `controls.values` then PUT should not be `controlsValues` but thats opinionated, I dont have strong opinion about this, it just removes the mapping

I can open these changes gradually and discuss them with the team so it doesn't disrupt existing code, like a wrangler pattern.

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
ChmaraX: Yes, this would be my steps to do that, we can gradually sync on that and make the changes:
https://github.com/novuhq/novu/pull/7131#discussion_r1870956412

packages/shared/src/types/feature-flags.ts
ChmaraX: I agree the naming could be better but I will not change it now - Its already in LD (can't be deleted just archived), its used elsewhere and its temporary + the description in LD states specifically what is it.

apps/dashboard/src/components/number-input-with-select.tsx
ChmaraX: I prefer `NumberInputWithSelect` because its immediately clear from the name what it is. For `AmountInput` I would need to open the component and check.

That said I changed it to `AmountInput`.

apps/dashboard/src/components/workflow-editor/schema.ts
```diff
@@ -23,10 +23,12 @@ export const workflowSchema = z.object({
   description: z.string().max(MAX_DESCRIPTION_LENGTH).optional(),
 });
 
-export const stepSchema = z.object({
-  name: z.string().min(1).max(MAX_NAME_LENGTH),
-  stepId: z.string(),
-});
+export const buildStepSchema = (controlsSchema?: z.ZodObject<any>) =>
```
SokratisVidros, line 26: Please add a TODO to return all the step schemas from the API. Since we are returning step controls, we should return everything.

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
SokratisVidros: We are aligned.

SokratisVidros (approved):

PR #7133, opened 2024-11-26
feat(dashboard): add feature flags provider and hook

ChmaraX:

Adds LD provider and hooks to be able to use feature flags.

Fixes: NV-4901


apps/dashboard/src/context/feature-flags-provider.tsx
```diff
@@ -0,0 +1,32 @@
...
+  };
+});
+
+export function FeatureFlagsProvider({ children }: { children: React.ReactNode }) {
+  return (
+    <Suspense fallback={<div>Loading...</div>}>
+      <AsyncFeatureFlagsProvider>{children}</AsyncFeatureFlagsProvider>
+    </Suspense>
+  );
+}
```
ChmaraX, line 38: I think this is a bit more elegant and logically separated compared to what [LD docs suggests](https://docs.launchdarkly.com/sdk/client-side/react/react-web#initialize-the-client):

```jsx
(async () => {
  const FeatureFlagsProvider = await asyncWithLDProvider({
    ....
  });

  createRoot(document.getElementById('root')!).render(
    <StrictMode>
      <FeatureFlagsProvider>
        <RouterProvider router={router} />
      </FeatureFlagsProvider>
    </StrictMode>
  );
})();
```

SokratisVidros: I like the logic with Suspense but please make sure that there is no initial flickering involved between a render where LD flags are not loaded yet and the time they arrive. See how we did it in [the web dashboard](https://github.com/novuhq/novu/blob/a305ec690cbdd7b89e636379cbfd36657d55a662/apps/web/src/index.tsx#L17) where lazy was harder to work.

apps/dashboard/src/hooks/use-feature-flag.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return !!LAUNCH_DARKLY_CLIENT_SIDE_ID;
+}
+
+export const useFeatureFlag = (key: FeatureFlagsKeysEnum, defaultValue = false): boolean => {
+  const flags = useFlags();
+
+  if (!isLaunchDarklyEnabled()) {
+    const envValue =
+      // Check window._env_ first
+      window?._env_?.[key] ??
```
SokratisVidros: I don't think we have the `_env_` fallback on the window object in the new Dashboard. This was in the old web app. Instead, we should rely on VITE_ prefixed env variables as a fallback.

SokratisVidros (changes requested):

apps/dashboard/src/context/feature-flags-provider.tsx
ChmaraX: I can't reproduce any flickering - it waits until the `AsyncFeatureFlagsProvider` resolves to either LD or to empty wrapper, only after that the app is rendered, should be exactly as in the legacy web.

```tsx
const AsyncFeatureFlagsProvider = lazy(async () => {
  if (!LAUNCH_DARKLY_CLIENT_SIDE_ID) {
    return {
      default: ({ children }: { children: React.ReactNode }) => <>{children}</>,
    };
  }

  const LaunchDarklyProvider = await asyncWithLDProvider(LD_CONFIG);
  return {
    default: ({ children }: { children: React.ReactNode }) => <LaunchDarklyProvider>{children}</LaunchDarklyProvider>,
  };
});

export function FeatureFlagsProvider({ children }: { children: React.ReactNode }) {
  return (
    <Suspense>
      <AsyncFeatureFlagsProvider>{children}</AsyncFeatureFlagsProvider>
    </Suspense>
  );
}
```

SokratisVidros: Nice. Do we need a loader or a splash screen? PTAL at [this touch](https://github.com/novuhq/novu/blob/4408daa2c02c938efa4ca71a6e5e9fa6ede69f03/apps/web/src/ApplicationReadyGuard.tsx#L20). It fades in the app when loading of everything is done. It's pretty cool.

SokratisVidros (approved):

apps/dashboard/src/context/feature-flags-provider.tsx
ChmaraX: The LD initiation is very fast - we didn't had splash screen in the legacy web before, so it should be probably ok.

PR #7137, opened 2024-11-26
feat(api): fix framework workflow payload preview

djabarovgeorge:

https://www.loom.com/share/a598b4a81a54447ea542c6104ab0f7fd

In App Step issue: Payload is not generated for the preview tab based on the payload schema for v2 workflows

1. Create a v2 code workflow with in app step
2. Specify the payload schema for the workflow
3. In app step in preview shows empty object

Expected behaviour:
* Payload schema is rendered as with UI based workflows

Also, use the preview response to see if errors still exist in combination with frontend validations

image
image

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
```diff
@@ -89,11 +88,15 @@ export class OverloadContentDataOnWorkflowUseCase {
...
+      return;
+    }
+
     let finalPayload = {};
     for (const value of Object.values(stepIdToControlValuesMap)) {
       finalPayload = _.merge(finalPayload, value.finalPayload.payload);
     }
     // eslint-disable-next-line no-param-reassign
-    workflow.payloadSchema = JSON.stringify(convertJsonToSchemaWithDefaults(finalPayload));
+    workflow.payloadSchema = convertJsonToSchemaWithDefaults(finalPayload);
```
djabarovgeorge, line 101: It simplified the payloadSchema type to avoid ambiguous complex type, changing it from `JSONSchemaDto | string` to a single, more consistent type, JSONSchemaDto as it was before.

https://linear.app/novu/issue/NV-4874/align-the-workflowpayloadschema-type

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
```diff
@@ -89,11 +88,15 @@ export class OverloadContentDataOnWorkflowUseCase {
     workflow: WorkflowInternalResponseDto,
     stepIdToControlValuesMap: { [p: string]: ValidatedContentResponse }
   ) {
+    if (workflow.origin === WorkflowOriginEnum.EXTERNAL) {
+      return;
+    }
```
djabarovgeorge, line 93: was needed because this flow was overriding the payload schema on the upsert request.

apps/api/src/app/workflows-v2/workflow.controller.ts
djabarovgeorge, line 1: renaming

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge, line 1: renaming

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
djabarovgeorge, line 1: this code is needed only because of this part
https://github.com/novuhq/novu/pull/7137/files#r1859185424
We need to avoid ambiguous types in the data layer, especially when the param is any in the Entity.

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
```diff
@@ -0,0 +1,15 @@
+export function parsePayloadSchema(schema: unknown): Record<string, unknown> {
+  if (schema && typeof schema === 'string') {
```
rifont: Since we stopped using `JSON.stringify` on the payload schema [here](https://github.com/novuhq/novu/pull/7137/files#diff-79f869ed3027bf91fde97124ec18dd06bec6e4a682f23b9916ece345c88b5ee7R100), is this change just for backwards compatibility of workflows with a persisted payloadSchema `string` type?

If so, I'd suggest we just treat this as a breaking change and avoid dealing with a polymorphic `object` | `string` type and accept that some workflows will be broken and will need their `payloadSchema` to be regenerated.

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
```diff
@@ -0,0 +1,15 @@
...
+    try {
+      return JSON.parse(schema);
+    } catch (error) {
+      throw new Error('Invalid JSON string provided for payload schema');
+    }
+  }
+
+  if (schema && typeof schema === 'object') {
```
rifont: `typeof null` === `'object'`, we need better type guard protection if we decide to go in this direction

```bash
> node
Welcome to Node.js v20.13.1.
Type ".help" for more information.
> typeof null
'object'
```

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
djabarovgeorge: This part is backward compatibility only, I think if we wait 1-2 weeks we can safely assume all workflows are valid and remove this part.

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
djabarovgeorge: this one is scary... :/

apps/api/src/app/workflows-v2/shared/parse-payload-schema.ts
rifont: Ok that sounds reasonable, let's create a ticket to address it after 2 weeks👍

djabarovgeorge: already created NV-4874 :)

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -45,24 +55,80 @@ export class GeneratePreviewUsecase {
...
+    stepData: StepDataDto,
+    user: UserSessionData,
+    payloadSchema: JSONSchemaDto
+  ) {
     if (!stepData.controls?.dataSchema) {
       throw new StepMissingControlsException(stepData.stepId, stepData);
     }
 
+    const payloadVariablesFromSchema = payloadSchema ? this.generatePayloadVariableExample(payloadSchema) : {};
+    const payloadVariablesExample = { ...payloadVariablesFromSchema, ...dto.previewPayload?.payload };
```
tatarco: i'm not sure this does deep merge, we need to be curfull here, we should have a deep merging here, also

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
```diff
@@ -44,9 +45,10 @@ export class BuildAvailableVariableSchemaUsecase {
     } as const satisfies JSONSchemaDto;
   }
 }
+
 function safePayloadSchema(workflow: NotificationTemplateEntity): JSONSchemaDto | undefined {
```
SokratisVidros: You can move this into the `parsePayloadSchema` helper and pass a second option, `{ strict: boolean }` or `{ safe: boolean }`, that will make the parse function undefined/null or throw an error.

For type safety, should we return undefined, null or `{}`? What's more convenient? I'd probably go with `null`.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -41,10 +53,55 @@ export class GeneratePreviewUsecase {
...
+      if (typeof definition === 'boolean') {
+        return acc;
+      }
+
+      const currentPath = `${path}.${key}`;
+
+      if (definition.default) {
+        acc[key] = definition.default;
+      } else if (definition.type === 'object' && definition.properties) {
+        acc[key] = this.generatePayloadVariableExample(definition, currentPath);
```
SokratisVidros: Please add a TODO here to add a maximum number of iterations to avoid an infinite loop. These schemas come from code, so we need to prevent very deep structures to avoid resource exhaustion attacks.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -41,10 +53,55 @@ export class GeneratePreviewUsecase {
...
+   * Generates a payload based solely on the schema.
+   * Supports nested schemas and applies defaults where defined.
+   * @param JSONSchemaDto - Defining the structure. example:
+   *  {
+   *    firstName: { type: 'string', default: 'John' },
+   *    lastName: { type: 'string' }
+   *  }
+   * @returns - Generated payload. example: { firstName: 'John', lastName: '{{payload.lastName}}' }
+   */
+  private generatePayloadVariableExample(schema: JSONSchemaDto, path = 'payload'): Record<string, unknown> {
```
SokratisVidros:
```suggestion
  private generateSamplePayload(schema: JSONSchemaDto, path = 'payload'): Record<string, unknown> {
```

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -60,9 +117,23 @@ export class GeneratePreviewUsecase {
...
 
+  private normalizeExampleVariables(
+    validatedContent: ValidatedContentResponse,
+    origin?: WorkflowOriginEnum,
+    payloadSchema?: JSONSchemaDto
+  ) {
+    if (origin === WorkflowOriginEnum.EXTERNAL && payloadSchema) {
+      const payloadVariableExample = { payload: this.generatePayloadVariableExample(payloadSchema) };
+
+      return mergeObjects(payloadVariableExample, validatedContent.finalPayload as Record<string, unknown>);
```
SokratisVidros: Why is this step necessary? What's the `validatedContent` for?

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
SokratisVidros: This feels like a code-smell. We shouldn't treat workflows differently when generating the payload sample. Please elaborate on the underlying issue and let's work it out differently.

SokratisVidros (changes requested):

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
djabarovgeorge: Agree.

This is necessary because currently, the dashboard workflow doesn't have a payload schema. As a workaround, we scan the control values, collect all instances of `{{payload.VAR}}`, and store them.  

For now, I'm skipping this step to avoid overriding the framework-defined payload schema.  

I also agree on separating concerns. I need the code to be as stable as possible, without frequent changes or refactoring, so we can focus on making meaningful improvements.

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
djabarovgeorge: `For type safety, should we return undefined, null or {}? What's more convenient? I'd probably go with null.`

I also prefer using `null`. Returning `{}` is unpredictable and reveals issues at runtime. We need to find a reliable way to avoid using empty objects.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: done added max depth of 10, i belive it will be enough for now.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: validatedContent is the response of PrepareAndValidateContentUsecase returning (ValidatedContentResponse), it had the data about the example payload(finalPayload) and issues.

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
SokratisVidros: If the workflow doesn't have a payload schema, the use case that generates the sample payload, should use the presence or absence of the workflow.payload schema in the database to handle the extra.

Again the argument is the same, use a single flow and use case for that

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -1,68 +1,156 @@
...
-    if (!stepData.controls?.dataSchema) {
-      throw new StepMissingControlsException(stepData.stepId, stepData);
+  private buildVariablesExample(
+    workflow: WorkflowInternalResponseDto,
+    stepData: StepDataDto,
+    dto: GeneratePreviewRequestDto
+  ) {
+    const variablesExample = this.generateVariablesExample(dto, stepData);
+
+    if (workflow.origin === WorkflowOriginEnum.EXTERNAL) {
```
SokratisVidros, line 72: Not all code-first workflows have payload ;)
```suggestion
    if (workflow.payloadSchema) {
```

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -1,68 +1,156 @@
...
+    const controlValues = Object.values(dto.controlValues || stepData.controls.values).join('');
+    const variablesExample = pathsToObject(extractTemplateVars(controlValues), {
+      valuePrefix: '{{',
+      valueSuffix: '}}',
     });
+
+    return variablesExample;
+  }
+
+  private buildPayloadExample(
```
SokratisVidros: Is this the duplication we talked about?

apps/api/src/app/workflows-v2/util/template-variables/extract-template-variables.ts
```diff
@@ -0,0 +1,39 @@
...
+export const extractTemplateVars = function (str: string): string[] {
+  if (str === null || str === undefined || typeof str !== 'string') {
+    return [];
+  }
+
+  const variables = new Set<string>();
+  const engine = new Liquid();
+  const parsed = engine.parse(str);
+
+  function traverseTokens(templates) {
```
SokratisVidros: This inner function declaration seems redundant.

SokratisVidros (changes requested):

apps/api/src/app/workflows-v2/util/template-variables/extract-template-variables.ts
djabarovgeorge: true this function was recursive and i scoped out the recursion because we don't need it at the moment (if/for support etc..)

djabarovgeorge: i will remove it 💪

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: not sure, which duplication you refer to but this code is old leftovers and not used anymore i removed it.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: this one is connected to 
https://github.com/novuhq/novu/pull/7137/files#r1859187874
and be resolved there.

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
djabarovgeorge: this comment resolves https://github.com/novuhq/novu/pull/7137/files#r1861939413 as well.

SokratisVidros: This was resolved offline.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -1,72 +1,152 @@
...
+   *    firstName: { type: 'string', default: 'John' },
+   *    lastName: { type: 'string' }
+   *  }
+   * @returns - Generated payload. example: { firstName: 'John', lastName: '{{payload.lastName}}' }
+   */
+  @Instrument()
+  private generateSamplePayload(schema: JSONSchemaDto, path = 'payload', depth = 0): Record<string, unknown> {
+    const MAX_DEPTH = 10;
+    if (depth >= MAX_DEPTH) {
+      return {};
```
SokratisVidros:
```suggestion
      throw new Error('Schema has surpassed the maximum allowed depth (10). Please specify a more shallow payload schema.');
```

We need a BadRequestException exception for that.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -1,72 +1,152 @@
...
+    }
+
+    if (Object.values(schema.properties || {}).length === 0) {
+      return {};
+    }
+
+    if (schema.type !== 'object' || !schema.properties) {
+      throw new Error('Schema must define an object with properties.');
```
SokratisVidros: Ditto about a BadRequestException exception.

SokratisVidros (approved):

PR #7139, opened 2024-11-26
fix(api): fix step id on sync to env

djabarovgeorge:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
@@ -40,32 +40,32 @@ export class SyncToEnvironmentUseCase {
       throw new BadRequestException('Cannot sync workflow to the same environment');
     }
 
-    const workflowToClone = await this.getWorkflowToClone(command);
-    const preferencesToClone = await this.getWorkflowPreferences(workflowToClone._id, command.user.environmentId);
-    const externalId = workflowToClone.workflowId;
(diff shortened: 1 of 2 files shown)


apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -176,11 +176,8 @@ export class SyncToEnvironmentUseCase {
     }
   }
 
-  private findDatabaseIdInProdByExternalId(
-    existingWorkflowSteps: StepResponseDto[] | undefined,
-    step: StepResponseDto
-  ) {
-    return existingWorkflowSteps?.find((existingStep) => existingStep.stepId === step.stepId)?._id ?? step._id;
```
djabarovgeorge, line 183: Removed the fallback fixed the sync issue. The problem was that we used the origin `_id` as a fallback, which wasn't available in the target environment during upsert. This prevented us from locating the control entry to update.

All other changes are just renames to make the code easier to follow.

scopsy (approved):

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
scopsy: Awesome! Do we maybe have a small e2e for syncing the changes between envs?

djabarovgeorge: we have some, let me check what is the test scope there.

PR #7140, opened 2024-11-26
fix(api): step naming

djabarovgeorge:
### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/build-step-data/build-step-data.command.ts
```diff
@@ -1,13 +1,13 @@
 import { EnvironmentWithUserObjectCommand } from '@novu/application-generic';
-import { IsNotEmpty, IsString } from 'class-validator';
+import { IsNotEmpty, IsOptional, IsString } from 'class-validator';
 import { IdentifierOrInternalId } from '@novu/shared';
 
 export class BuildStepDataCommand extends EnvironmentWithUserObjectCommand {
   @IsString()
   @IsNotEmpty()
-  identifierOrInternalId: IdentifierOrInternalId;
+  workflowIdentifierOrInternalId: IdentifierOrInternalId;
```
SokratisVidros:
```suggestion
  workflowIdOrInternalId: string;
```

SokratisVidros:
@djabarovgeorge Defining the IdentifierOrInternalId type as a string doesn't add any value, so let's just drop it. Moreover, this PR is about alignment. So the naming should be the same between workflows and steps. That is `workflowIdOrInternalId` and `stepIdOrInternalId`.

SokratisVidros (changes requested):

djabarovgeorge:
> @djabarovgeorge Defining the IdentifierOrInternalId type as a string doesn't add any value, so let's just drop it. Moreover, this PR is about alignment. So the naming should be the same between workflows and steps. That is `workflowIdOrInternalId` and `stepIdOrInternalId`.

Thanks for the feedback! @SokratisVidros 
I’ve updated all the endpoints to explicitly use workflowIdOrInternalId and stepIdOrInternalId. The same changes were applied to nested use cases as well. This should make it clearer in nested cases which ID we’re working with.

SokratisVidros (approved):

PR #7141, opened 2024-11-26
fix(dashboard): Fix copywriting

SokratisVidros:

Tweak the copywriting of the test trigger toast

Screenshot 2024-11-27 at 11 02 25

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-tabs.tsx
+++ b/apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-tabs.tsx
@@ -53,14 +53,14 @@ export const TestWorkflowTabs = ({ testData }: { testData: WorkflowTestDataRespo
           <>
             <ToastIcon variant="default" />
             <div className="flex flex-col gap-2">
-              <span className="font-medium">The test trigger was completed</span>
-              <span className="text-foreground-600">
-                Workflow ${workflow?.workflowId} was triggered successfully from the Dashboard.
+              <span className="font-medium">Test workflow succeeded</span>
+              <span className="text-foreground-600 inline">
+                Workflow <strong>{workflow?.name}</strong> was triggered successfully.
               </span>
               <Link
                 to={`${LEGACY_ROUTES.ACTIVITY_FEED}?transactionId=${transactionId}`}
                 reloadDocument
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-tabs.tsx
```diff
@@ -53,8 +53,10 @@ export const TestWorkflowTabs = ({ testData }: { testData: WorkflowTestDataRespo
           <>
             <ToastIcon variant="default" />
             <div className="flex flex-col gap-2">
-              <span className="font-medium">Test workflow triggered successfully</span>
-              <span className="text-foreground-600">{`Test workflow ${workflowSlug} was triggered successfully`}</span>
+              <span className="font-medium">The test trigger was completed</span>
+              <span className="text-foreground-600">
+                Workflow ${workflow?.workflowId} was triggered successfully from the Dashboard.
```
rifont: We are following the pattern of wrapping resource names with strong styling in other places, we could do the same here

```suggestion
                Workflow <strong>${workflow?.workflowId}</strong> was triggered successfully from the Dashboard.
```

rifont (approved):
💯

apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-tabs.tsx
rifont: I saw the link to change from using slug to workflowId.

Don't we think showing the workflow name would be more relevant to the Dashboard user?

rifont, line 58: I saw the link to change from using slug to workflowId.

Don't we think showing the workflow name would be more relevant to the Dashboard user?

rifont: I saw the link to change from using slug to workflowId.

Don't we think showing the workflow name would be more relevant to the Dashboard user?

rifont: If we go with the slug, adding the strong emphasis probably isn't needed

PR #7142, opened 2024-11-27
chore(root): Release version `2.1.0` for all applications

rifont:
* Release version `2.1.0` for all applications
* Add `type:app` tags to application package nx configuration to simplify release flow
  * make tag configuration consistent following NX recommended naming (apps: `tag: "type:app"`, packages: `tag: "package:public"` -> `tag: "type:package"` - all packages are public and non-public packages are libraries, so there's no need to be explicit about a package being public)
* Organize releasable artifacts into NX groups in nx.json

### Screenshots

<details>

</details>


scripts/release.mjs
```diff
@@ -29,13 +29,15 @@ import { execa } from 'execa';
   }
 
   const { workspaceVersion, projectsVersionData } = await releaseVersion({
+    projects: ['tag:type:package'],
```
rifont, line 41: @SokratisVidros is this script intended to release packages only? If not, we need to remove this filter

nx.json
```diff
@@ -21,24 +21,46 @@
...
+          "@novu/react-native",
+          "@novu/nextjs",
+          "@novu/node",
+          "@novu/providers",
+          "@novu/shared",
+          "@novu/stateless"
+        ],
+        "projectsRelationship": "independent",
+        "version": {
+          "generatorOptions": {
```
SokratisVidros, line 62: Let's add `"conventionalCommits": true,` here.

scripts/release.mjs
SokratisVidros: It releases only packages.

scripts/release.mjs
```diff
@@ -29,13 +29,15 @@ import { execa } from 'execa';
...
     specifier,
     dryRun,
     verbose,
     firstRelease: true,
```
SokratisVidros: This was committed accidentally. Please remove.

SokratisVidros (approved):

PR #7145, opened 2024-11-27
fix(api): make access internal to self-managed bridge

tatarco:
Issue with preview endpoint taking too long, make the access internal

### Screenshots

<details>

</details>

### DIFF
--- a/libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
+++ b/libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
@@ -1,10 +1,10 @@
 import {
+  BadRequestException,
+  HttpException,
+  HttpStatus,
   Injectable,
   Logger,
   NotFoundException,
-  BadRequestException,
-  HttpStatus,
-  HttpException,
 } from '@nestjs/common';
 import got, {
   CacheError,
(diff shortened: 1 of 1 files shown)


libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
```diff
@@ -270,13 +270,7 @@ export class ExecuteBridgeRequest {
...
 
   private getApiUrl(): string {
-    const apiUrl = process.env.API_ROOT_URL;
-
-    if (!apiUrl) {
-      throw new Error('API_ROOT_URL environment variable is not set');
-    }
-
-    return apiUrl;
+    return `http://localhost:${process.env.PORT}`;
```
djabarovgeorge:
```suggestion
    return `http://127.0.0.1:${process.env.PORT}`;
```

djabarovgeorge: having this code running in `Worker` ECS instance will the `API` ECS instance will be in the same network?

rifont (approved):
🦾

libs/application-generic/src/usecases/execute-bridge-request/execute-bridge-request.usecase.ts
rifont: @djabarovgeorge could you please elaborate on the reason for this suggestion?

rifont: We opted for an approach to only use the localhost loopback for the preview action, where it's guaranteed to be called from the API today.

djabarovgeorge: can you elaborate please? not sure i follow how it is not an issue for preview.

PR #7147, opened 2024-11-27
feat(api): wip fix framework workflow issues

djabarovgeorge:

now framework workflow has issues ;)
image

https://www.loom.com/share/ae23d0ef53ff455a8f5fe5b521674308

### Screenshots

<details>

</details>


apps/api/src/app/bridge/usecases/sync/sync.usecase.ts
```diff
@@ -142,46 +158,79 @@ export class Sync {
...
+        savedWorkflow = await this.updateWorkflowUsecase.execute(
+          UpdateWorkflowCommand.create({
+            ...validatedWorkflowWithIssues,
+            id: validatedWorkflowWithIssues._id,
+            type: WorkflowTypeEnum.BRIDGE,
+            environmentId: command.environmentId,
+            organizationId: command.organizationId,
+            userId: command.userId,
+          })
         );
```
djabarovgeorge, line 191: this part is responsible for updating the issues on the workflow


image

apps/dashboard/src/components/workflow-editor/in-app-preview.tsx
```diff
@@ -174,6 +175,10 @@ const Markdown = (props: MarkdownProps) => {
 
   const tokens = useMemo(() => parseMarkdownIntoTokens(children || ''), [children]);
 
+  if (children === PreviewIssueEnum.PREVIEW_ISSUE_REQUIRED_CONTROL_VALUE_IS_MISSING) {
+    return null;
+  }
```
djabarovgeorge: this part ensures that the issue returned by the API from the backend is not rendered.  

https://www.loom.com/share/ae23d0ef53ff455a8f5fe5b521674308


I have two questions:  
1. Is this the correct approach to render null it in the new dashboard?  
2. Why are we returning this value from the backend in the first place? Don’t we already use issues to inform the frontend when there’s a problem?

scopsy: My main problem with it, is that the preview will not indicate **any** error state on the tab. This at least shows you that something is wrong.

scopsy (approved):
Looks good, let's just restore the issue rendering behaviour as per the slack thread https://github.com/novuhq/novu/pull/7147/files#r1865851863

PR #7148, opened 2024-11-27
feat(dashboard): saving modal on navigating outside editor

ChmaraX:

⚠️ TODO: needs to be rebased on top of https://github.com/novuhq/novu/pull/7149

- saving modal is shown only when leaving workflow editor and the save request is pending
- toasts are cleaned when leaving workflow editor
- the user navigation continues when changes are saved
- user can close the modal in case as a fallback in case of timeout from API or whatever so he is not stuck on that state

No network throttling:

(hard to capture but before it was showing modal when focusing on canvas or clicking on other node etc)

https://github.com/user-attachments/assets/c1d36fe4-9d63-48c2-ace2-7ebd684d6714

Network throttling: 

https://github.com/user-attachments/assets/0a18b9b1-a9d9-47d5-b606-d84241cfe150

Network throttling - cancel redirect:

https://github.com/user-attachments/assets/62d4f6ea-6460-4eb8-b17d-f512cbacb3d0


apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -153,7 +162,27 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+   */
+  const handleBlockedNavigation = useCallback(async () => {
+    return new Promise<void>((resolve) => {
+      setTimeout(() => {
+        blocker.proceed?.();
+        toast.dismiss();
+        resolve();
+      }, 500);
+    });
+  }, [blocker]);
```
ChmaraX: The async and timeout here I think its needed because it:
1. ensures users perceive the success state (better UX)
2. allows success animations to complete their transitions
3. prevents jarring immediate redirects that might make users question if the save completed

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -153,7 +162,27 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+
+  const handleCancelNavigation = useCallback(() => {
+    if (blocker.state === 'blocked') {
+      blocker.reset();
+    }
+  }, [blocker]);
```
ChmaraX: fallback in case the user would be stuck on the "Saving changes" dialog (API timeout etc). Now he can press the "X" button otherwise he would need to refresh to get out of that state.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -228,30 +257,37 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+              </div>
+              <AlertDialogDescription className="transition-all duration-300">
+                {!isPending ? 'Workflow has been saved successfully' : 'Please wait while we save your changes'}
               </AlertDialogDescription>
             </div>
+            {isPending && (
+              <button onClick={handleCancelNavigation} className="text-gray-500">
+                <RiCloseFill className="size-4" />
+              </button>
+            )}
```
ChmaraX: This should be an `<Dialog>` not `<AlertDialog>`

- Dialog - if user action is not that important -> has cancel button and cancel on background click by default
- AlertDialog - when user action is mandatory/important

In this case its more like a Dialog usecase but we have different styling for the `<Dialog>` and I didn't wanted to spent time unifying that in this PR - will refactor this in separate PR.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -108,6 +100,17 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+    const workflowEditorBasePath = buildRoute(ROUTES.EDIT_WORKFLOW, {
+      workflowSlug,
+      environmentSlug: currentEnvironment?.slug ?? '',
+    });
+
+    const isLeavingEditor = !nextLocation.pathname.startsWith(workflowEditorBasePath);
+
+    return isLeavingEditor && isPending;
+  });
+
```
ChmaraX: I tried many alternatives to this, this seems to be the simplest and most predictable one.

At first I was thinking we could react to the unmount of this component and then block when component is unmounting, but that doesn't work due to `ReactStrict` components are mounting/unmounting twice at render etc etc

SokratisVidros: Relying on the routing mechanism seems to be a good solution.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -228,30 +257,37 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+                )}
+              </div>
             </div>
             <div className="space-y-1">
-              <AlertDialogTitle>You might lose your progress</AlertDialogTitle>
-              <AlertDialogDescription>
-                This workflow has some unsaved changes. Save progress before you leave.
+              <div>
+                <AlertDialogTitle className="transition-all duration-300">
+                  {!isPending ? 'Changes saved!' : 'Saving changes'}
```
SokratisVidros, line 278: Do you think the success dialog is necessary? The UX on pending changes is great, but we can remove the saved changes dialog. We already have the saved toast.

SokratisVidros (approved):

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
SokratisVidros: Save one character! 😆 

```suggestion
                  {isPending ? 'Saving changes' : 'Changes saved!'}
```

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -228,30 +257,37 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+                </AlertDialogTitle>
+              </div>
+              <AlertDialogDescription className="transition-all duration-300">
+                {!isPending ? 'Workflow has been saved successfully' : 'Please wait while we save your changes'}
```
SokratisVidros: Save another one!

SokratisVidros:
```suggestion
  {isPending ? 'Saving changes, please wait…' : 'Workflow was saved successfully' }
```

SokratisVidros (approved):

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
SokratisVidros: Perfect. In the next iteration, I'd like us to examine the possibility of using a Sonner toast for this as well, so that we have one way to demonstrate feedback in the editor. It doesn't have to be centralized, and the no-click cloak can appear separately of the sonner message.

apps/dashboard/src/components/workflow-editor/workflow-editor-provider.tsx
```diff
@@ -228,30 +257,37 @@ export const WorkflowEditorProvider = ({ children }: { children: ReactNode }) =>
...
+            >
+              <div className="transition-opacity duration-300">
+                {!isPending ? (
+                  <CheckCircleIcon className="text-success animate-in fade-in size-6" />
+                ) : (
+                  <RiAlertFill className="text-warning animate-in fade-in size-6" />
+                )}
+              </div>
             </div>
             <div className="space-y-1">
```
SokratisVidros: Please ensure that there is no layout shift during the state transition and the dialog shape and dimensions remain the same.

PR #7149, opened 2024-11-27
fix(dashboard): Sibling form structure and autosave

desiprisg:

Changes:
* Opted for a more safe approach when rendering forms
  * Form components are now injected with the objects they rely on.
  * Form components no longer depend on hooks like useParams or useWorkflowProviderContext to receive ids, slugs or data.
  * It is the form component's parent concern to provide those objects when the component is rendered
* Workflows & Steps
  * Workflow is fetched on the <WorkflowProvider /> and can be accessed via useWorkflow()
  * Step is fetched on the <StepProvider /> and can be accessed via useStep()
    * All step related endpoints use useStep() as source of truth.
* We no longer have nested forms, but rather sibling forms.
  * One form for the configure workflow page
  * One form for the configure step sidebar
  * One form for the configure step template sheet
  * No forms for canvas
* Autosave across the board
  * useFormAutosave is used for all 3 forms, and we're passing in the same debounced function for all, ensuring consistency. This function can be accessed via useWorkflow . The variants are:
    * update for instant PUT
      * This calls debouncedUpdate and then flushes.
    * debouncedUpdate for debounced PUT
    * patch for instant PATCH (only used for the active field)
### Screenshots

<details>

</details>


apps/dashboard/src/components/primitives/editor.tsx
```diff
@@ -141,27 +52,33 @@ export const Editor = React.forwardRef<{ focus: () => void; blur: () => void },
...
-          { tag: t.variableName, color: 'hsl(var(--foreground-950))' },
-        ],
-        settings: {
-          background: 'transparent',
-          lineHighlight: 'transparent',
-          fontFamily: fontFamily === 'inherit' ? 'inherit' : undefined,
-        },
-      });
-    }, [fontFamily]);
+    const debouncedOnChange = useMemo(
```
SokratisVidros, line 75:
```suggestion
    const debouncedOnChange = useCallback(
```

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  const onPauseWorkflow = () => {
+    statusForm.setValue('active', false, { shouldValidate: true, shouldDirty: true });
+  };
+
+  useFormAutosave(form, (data) => {
+    debouncedUpdate({ ...workflow, ...data });
+  });
+
+  useFormAutosave(statusForm, (data) => {
```
SokratisVidros, line 61: 🛑  This function is defined twice!

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -0,0 +1,229 @@
...
+    patch(data);
+  });
+
+  // const setIssuesFromWorkflow = (workflow: WorkflowResponseDto) => {
```
SokratisVidros: Please remove this if not needed.

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -0,0 +1,229 @@
...
+        transition={{ duration: 0.1 }}
+      >
+        <SidebarHeader className="items-center text-sm font-medium">
+          <div className="flex items-center gap-1">
+            <RouteFill />
+            <span>Configure workflow</span>
+          </div>
+        </SidebarHeader>
+        <Separator />
+        <form>
```
SokratisVidros: The `<form/>` should be rendered by the `<Form/>` primitive component. See how they use it [here](https://react-hook-form.com/docs/formprovider). That is wrap `form` with the Form Provider.

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -0,0 +1,229 @@
...
+              <FormField
+                control={form.control}
+                name="description"
+                render={({ field }) => (
+                  <FormItem>
+                    <FormLabel>Description</FormLabel>
+                    <FormControl>
+                      <Textarea
+                        className="min-h-36"
+                        placeholder="Description of what this workflow does"
```
SokratisVidros:
```suggestion
                        placeholder="Describe what this workflow does"
```

apps/dashboard/src/components/workflow-editor/configure-workflow.tsx
```diff
@@ -1,215 +1,14 @@
...
-        <Button variant="outline" className="flex w-full justify-start gap-1.5 text-xs font-medium" type="button">
-          <RiSettingsLine className="h-4 w-4 text-neutral-600" />
-          Configure channel preferences <RiArrowRightSLine className="ml-auto h-4 w-4 text-neutral-600" />
-        </Button>
-      </SidebarContent>
-      <Separator /> */}
-      </motion.div>
-    </>
-  );
+  return <ConfigureWorkflowForm workflow={workflow} debouncedUpdate={debouncedUpdate} patch={patch} />;
```
SokratisVidros, line 13: I don't think we need both the `debouncedUpdate` and the patch. I'd pass a `save` prop.

apps/dashboard/src/components/workflow-editor/schema.ts
```diff
@@ -20,7 +20,7 @@ export const MAX_TAG_LENGTH = 32;
 export const MAX_NAME_LENGTH = 64;
 export const MAX_DESCRIPTION_LENGTH = 256;
 
-export const workflowMinimalSchema = z.object({
+export const configureWorkflowSchema = z.object({
```
SokratisVidros:
```suggestion
export const workflowSchema = z.object({
```

apps/dashboard/src/components/workflow-editor/schema.ts
```diff
@@ -35,7 +35,12 @@ export const workflowMinimalSchema = z.object({
   description: z.string().max(MAX_DESCRIPTION_LENGTH).optional(),
 });
 
-export const workflowSchema = workflowMinimalSchema.extend({
+export const configureStepSchema = z.object({
```
SokratisVidros:
```suggestion
export const stepSchema = z.object({
```

apps/dashboard/src/components/workflow-editor/schema.ts
```diff
@@ -35,7 +35,12 @@ export const workflowMinimalSchema = z.object({
...
+  name: z.string().min(1).max(MAX_NAME_LENGTH),
+  stepId: z.string(),
+});
+
+export const workflowSchema = configureWorkflowSchema.extend({
```
SokratisVidros: I suggest we define a single workflowSchema with all the fields and then we [pick/omit per use case](https://github.com/colinhacks/zod?tab=readme-ov-file#pickomit). That way, we will only have `stepSchema` and `workflowSchema` in Zod.

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { getFirstBodyErrorMessage, getFirstControlsErrorMessage } from '@/components/workflow-editor/step-utils';
+import { ConfigureInAppStepTemplateCta } from '@/components/workflow-editor/steps/in-app/configure-in-app-step-template-cta';
+import { zodResolver } from '@hookform/resolvers/zod';
+import { configureStepSchema } from '@/components/workflow-editor/schema';
+
+type ConfigureStepFormProps = {
+  workflow: WorkflowResponseDto;
+  environment: IEnvironment;
+  step: StepDataDto;
+  debouncedUpdate: (data: UpdateWorkflowDto) => void;
```
SokratisVidros, line 41: I am not a big fan of this prop duplication. I'd pass a `save` prop that is debounced by default and accepts an option for immediate invocation. For example `save(newWorkflowData, { immediate: true })`. `

The key is to have one `save` across all form elements that are debounced by default, but you can also opt-out. The immediate should be part of the debounce mechanism so that subsequent immediate invocations still invoke the same function, and there is a single chain of immediate and debounced calls. 

See [here](https://lodash.com/docs/4.17.15#debounce).

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -0,0 +1,220 @@
...
+          >
+            <Button variant="link" size="icon" className="size-4" type="button">
+              <RiCloseFill />
+            </Button>
+          </Link>
+        </SidebarHeader>
+
+        <Separator />
+
+        <form>
```
SokratisVidros: Ditto about the form element

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <>
+            <SidebarContent>
+              <Link to={'./edit'} relative="path" state={{ stepType: step.type }}>
+                <Button
+                  variant="outline"
+                  className="flex w-full justify-start gap-1.5 text-xs font-medium"
+                  type="button"
+                >
+                  <RiPencilRuler2Fill className="h-4 w-4 text-neutral-600" />
+                  Configure {step.type} template <RiArrowRightSLine className="ml-auto h-4 w-4 text-neutral-600" />
```
SokratisVidros: Please titleize/humanize the step.type so that it becomes In App

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -0,0 +1,78 @@
...
+};
+export const ConfigureStepTemplateForm = (props: ConfigureStepTemplateFormProps) => {
+  const { workflow, step, debouncedUpdate } = props;
+
+  const schema = useMemo(() => buildDynamicZodSchema(step.controls.dataSchema ?? {}), [step.controls.dataSchema]);
+
+  const form = useForm({
+    mode: 'onChange',
+    resolver: zodResolver(schema),
+    defaultValues: merge(buildDefaultValues(step.controls.uiSchema ?? {}), step.controls.values),
```
SokratisVidros: The merging can be memoized.

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  const setIssuesFromStep = (step: StepDataDto) => {
+    const issues = flattenIssues(step.issues?.controls);
+    Object.entries(issues).forEach(([key, value]) => {
+      form.setError(key as string, { message: value }, { shouldFocus: true });
+    });
+  };
+
+  //set issues from step
```
SokratisVidros: Please remove

apps/dashboard/src/components/workflow-editor/steps/configure-step.tsx
```diff
@@ -1,123 +1,23 @@
...
-            className="ml-auto flex items-center"
-          >
-            <Button variant="link" size="icon" className="size-4" type="button">
-              <RiCloseFill />
-            </Button>
-          </Link>
-        </SidebarHeader>
-
-        <Separator />
+  if (!step || !workflow || !currentEnvironment) {
```
SokratisVidros:
```suggestion
  if (!currentEnvironment || !workflow || !step) {
```

apps/dashboard/src/components/workflow-editor/steps/edit-step-template.tsx
```diff
@@ -0,0 +1,10 @@
+import { ConfigureStepTemplate } from '@/components/workflow-editor/steps/configure-step-template';
```
SokratisVidros, line 1: Are these edit components necessary? I feel that we can move the wrapping into the  Configure* components and remove two extra imports.

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
```diff
@@ -0,0 +1,55 @@
...
+  in_app: 'In-App Step',
+  sms: 'SMS Step',
+  push: 'Push Step',
+  digest: 'Digest Step',
+  delay: 'Delay Step',
+  trigger: 'Trigger Step',
+  custom: 'Custom Step',
+};
+
+export const createStep = (type: StepTypeEnum): Step => ({
```
SokratisVidros, line 49: You can also add this function in the provider value and get it from via `const { buildStep } = useStep()` hook.

apps/dashboard/src/components/workflow-editor/workflow-provider.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { useEnvironment } from '@/context/environment/hooks';
+import { useFetchWorkflow, useUpdateWorkflow } from '@/hooks';
+import { useDebounce } from '@/hooks/use-debounce';
+import { usePatchWorkflow } from '@/hooks/use-patch-workflow';
+import { createContextHook } from '@/utils/context';
+import { buildRoute, ROUTES } from '@/utils/routes';
+
+export type WorkflowContextType = {
+  isPending: boolean;
+  workflow?: WorkflowResponseDto;
```
SokratisVidros, line 16: Consider making the `workflow` DTO or null (when not loaded) instead of optional.

apps/dashboard/src/components/workflow-editor/workflow-provider.tsx
```diff
@@ -0,0 +1,177 @@
...
+  const navigate = useNavigate();
+
+  const { workflow, isPending, error } = useFetchWorkflow({
+    workflowSlug,
+  });
+
+  const { patchWorkflow } = usePatchWorkflow({
+    onMutate: () => {
+      setToastId(
+        showToast({
```
SokratisVidros, line 38: Please add a TODO to tidy these toast helpers and make them dry.

apps/dashboard/src/components/workflow-editor/workflow-tabs.tsx
```diff
@@ -1,19 +1,18 @@
...
-  const form = useFormContext<z.infer<typeof workflowSchema>>();
-  const steps = useWatch({
-    control: form.control,
-    name: 'steps',
-  });
+export const WorkflowTabs = () => {
+  const { workflow } = useWorkflow();
+  const { currentEnvironment } = useEnvironment();
+
+  if (!workflow || !currentEnvironment) {
```
SokratisVidros:
```suggestion
  if (!currentEnvironment || !workflow) {
```

SokratisVidros (changes requested):

apps/dashboard/src/components/primitives/editor.tsx
desiprisg: `useCallback` gives an eslint error here.

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
desiprisg: Yeap, it's for two different "forms". One is for the patch endpoint (status form) and one for the update of the rest.

apps/dashboard/src/components/workflow-editor/configure-workflow.tsx
desiprisg: patch is used for active field only. update is instant PUT. debouncedUpdate is debounced PUT

apps/dashboard/src/components/workflow-editor/schema.ts
desiprisg: actually, i don't think we're going to need any more, since the forms are only the simple cases now.

apps/dashboard/src/components/workflow-editor/steps/edit-step-template.tsx
desiprisg: The component is necessary because it wraps with the provider. The file isn't that much but I liked it separated as it's the entry point in main.tsx

apps/dashboard/src/components/workflow-editor/steps/step-provider.tsx
desiprisg: It doesn't really depend on the context. At some point, I'd like for our canvas nodes to not use the context at all if possible so this would be imported.

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
desiprisg: I think this works as expected right now. `update` and `debouncedUpdate` both call the same debounced func. `patch` is used only for the active field and it makes sense to be a different func.

SokratisVidros (approved):

LetItRock (approved):

PR #7151, opened 2024-11-27
fix(api): Resolve circular import issue for workflow update validation

tatarco:
update call was not validating as expected

### Screenshots

<details>

</details>


libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.command.ts
```diff
@@ -24,14 +24,11 @@ import { Type } from 'class-transformer';
...
-  ContentIssue,
-  IStepControl,
   MAX_DESCRIPTION_LENGTH,
   MAX_NAME_LENGTH,
-  MAX_TAG_ELEMENTS,
   MAX_TAG_LENGTH,
-  NotificationStep,
-} from '../..';
+} from './upsert-validation-constants';
+import { ContentIssue, IStepControl, NotificationStep } from '../..';
```
rifont: These import statements from index files (`'../..'` ) are usually the problematic ones for creating circular import dependencies. I think changing the line above to include an absolute path reference will be all that's required here.

That said, having the validation constants in a separate file is much cleaner.

rifont (approved):
💯

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.command.ts
SokratisVidros: I pushed a fix for this so that we can merge it asap.

SokratisVidros (approved):

PR #7161, opened 2024-11-29
fix(api): remove duplicate New Relic import in bootstrap file

merrcury:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/bootstrap.ts
+++ b/apps/api/src/bootstrap.ts
@@ -1,5 +1,3 @@
-import 'newrelic';
-import './config/env.config';
 import './instrument';
 
 import helmet from 'helmet';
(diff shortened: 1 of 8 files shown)


SokratisVidros (approved):

apps/api/src/bootstrap.ts
```diff
@@ -1,6 +1,6 @@
+import 'newrelic';
```
SokratisVidros: You can also move the import to `./instrument` and import `./instrument` as the first thing.

SokratisVidros:
@merrcury Please apply the same change to all bootstrap files for all services.

apps/api/src/bootstrap.ts
merrcury: Wouldn't it create problem with import './config/env.config'? As instrument does make use of it?

merrcury: Shall I move that inside instrument too?

SokratisVidros: Yes

PR #7164, opened 2024-11-29
fix(dashboard): polish autosave experience

LetItRock:

Polish the autosave UX with:
- saving changes on blur
- having 10 debounce for change event

### Screenshots

https://github.com/user-attachments/assets/22d2bcd3-093b-4748-a9bf-68ee158b353f

https://github.com/user-attachments/assets/81cd59dd-e68e-46d6-a19f-195642745bb8

https://github.com/user-attachments/assets/81ba7171-baa2-4d7e-9f12-7dfc405329bc

https://github.com/user-attachments/assets/ebc165c8-164e-40a6-8ead-76e42e5bb814


apps/dashboard/src/components/icons/inbox-bell-filled.tsx
```diff
@@ -6,22 +6,27 @@ type InboxBellFilledProps = {
   codeClassName?: string;
 };
 
-export function InboxBellFilled(props: HTMLAttributes<HTMLOrSVGElement> & InboxBellFilledProps) {
+export function InboxBellFilled({
+  bellClassName,
+  codeClassName,
+  ringerClassName,
+  ...props
+}: HTMLAttributes<HTMLOrSVGElement> & InboxBellFilledProps) {
```
LetItRock: react warning about props being used on dom els

apps/dashboard/src/components/primitives/form/avatar-picker.tsx
```diff
@@ -35,79 +35,82 @@ type AvatarPickerProps = {
...
+    const { error } = useFormField();
 
-  const handlePredefinedAvatarClick = (url: string) => {
-    onChange?.(url);
-    setIsOpen(false);
-  };
+    const handlePredefinedAvatarClick = (url: string) => {
+      onPick?.(url);
+      setIsOpen(false);
+    };
```
LetItRock, line 51: I had to differentiate the onChange from avatar image pick, because on pick the API should be called

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -20,61 +20,56 @@ import { Switch } from '../primitives/switch';
...
 export const ConfigureWorkflowForm = (props: ConfigureWorkflowFormProps) => {
-  const { workflow, debouncedUpdate, patch, onDirtyChange } = props;
+  const { workflow, update, onDirtyChange } = props;
   const isReadOnly = workflow.origin === WorkflowOriginEnum.EXTERNAL;
   const [isPauseModalOpen, setIsPauseModalOpen] = useState(false);
   const tagsQuery = useTagsQuery();
 
   const form = useForm<z.infer<typeof workflowSchema>>({
     defaultValues: {
+      active: workflow.active,
```
LetItRock, line 38: no need for two separate forms, toggle is calling api right away

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -20,61 +20,56 @@ import { Switch } from '../primitives/switch';
...
 
-  const statusForm = useForm({
-    defaultValues: {
-      active: workflow.active,
-    },
+  const { onBlur, flushFormUpdates } = useFormAutosave({
+    previousData: workflow,
+    form,
+    isReadOnly,
+    update,
```
LetItRock: The modified `useFormAutosave` that provides the form `onBlur` listener and the utility function to flush all form changes and call api. More about it later.

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -61,18 +59,19 @@ export const ConfigureStepForm = (props: ConfigureStepFormProps) => {
...
-    });
+  const { onBlur } = useFormAutosave({
+    previousData: step,
+    form,
+    isReadOnly: isCodeCreatedWorkflow,
+    update: (data) => {
+      update({
+        ...workflow,
+        steps: workflow.steps.map((s) => (s._id === step._id ? { ...s, ...data } : s)),
+      });
```
LetItRock: here we only want to update the step changes

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,81 @@
...
+  update: (data: U) => void;
+}) {
+  const formRef = useDataRef(propsForm);
   const watchedData = useWatch<T>({
-    control: form.control,
+    control: propsForm.control,
   });
 
+  const onUpdate = async (data: T) => {
+    // use the form reference instead of descructuring the props to avoid stale closures
```
SokratisVidros:
```suggestion
    // use the form reference instead of destructuring the props to avoid stale closures
```

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,81 @@
...
+    }
+    const isValid = await form.trigger();
+    if (!isValid) {
+      return;
+    }
+
+    const values = { ...previousData, ...data };
+    // reset the dirty fields right away because on slow networks the patch request might take a while
+    // so other blur/change events might trigger in the meantime
+    form.reset(values);
```
desiprisg: Doesn't this reset the entire form?

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -27,55 +28,65 @@ export type StepEditorProps = {
...
     () => merge(buildDefaultValues(step.controls.uiSchema ?? {}), step.controls.values),
-    []
+    [step]
   );
 
   const form = useForm({
-    mode: 'onChange',
     resolver: zodResolver(schema),
     defaultValues: defaultValues,
-    shouldFocusError: true,
```
LetItRock, line 54: There is an issue with automatically focusing the field that we will have to solve. 
Basically when this option is set to `true` the field receives the focus. And if not fix the issue in the field it's not possible to blur into any another field. That's because `onBlur` implementation triggers the validation again.

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -27,55 +28,65 @@ export type StepEditorProps = {
...
   const Editor = STEP_TYPE_TO_EDITOR[step.type];
+
+  const value = useMemo(() => ({ flushFormUpdates }), [flushFormUpdates]);
+
   return (
     <Form {...form}>
-      <form className="flex h-full flex-col">
-        <Editor workflow={workflow} step={step} debouncedUpdate={debouncedUpdate} />
+      <form className="flex h-full flex-col" onBlur={onBlur}>
+        <FlashFormUpdatesContext.Provider value={value}>
```
LetItRock: I didn't know how to better propagate `flushFormUpdates` to deeply nested form components.

apps/dashboard/src/components/workflow-editor/steps/controls/select-widget.tsx
```diff
@@ -24,7 +26,15 @@ export function SelectWidget(props: WidgetProps) {
         <FormItem className="my-2 py-1">
           <FormLabel>{capitalize(label)}</FormLabel>
           <FormControl>
-            <Select value={field.value} onValueChange={field.onChange} disabled={readonly} required={required}>
+            <Select
+              value={field.value}
+              onValueChange={(value) => {
+                field.onChange(value);
+                flushFormUpdates();
```
LetItRock: All selects, dropdowns, toggles etc, basically not text fields should call `flushFormUpdates` after field changes.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-action.tsx
```diff
@@ -81,11 +84,19 @@ export const InAppAction = () => {
...
           </DropdownMenuTrigger>
         </div>
-        <DropdownMenuContent className="p-1" align="end">
+        <DropdownMenuContent
+          className="p-1"
+          align="end"
+          onBlur={(e) => {
+            e.preventDefault();
+            e.stopPropagation();
+          }}
```
LetItRock: weird behaviour but it was raising `onBlur` when hovering over the menu items

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,81 @@
...
+    // use the form reference instead of descructuring the props to avoid stale closures
+    const form = formRef.current;
+    const isDirty = form.formState.isDirty;
+    if (!isDirty || isReadOnly) {
+      return;
+    }
+    const isValid = await form.trigger();
+    if (!isValid) {
+      return;
+    }
```
LetItRock: If the form is read-only, not dirty, or not valid we should not call `update`.

LetItRock: Maybe better to call it `save` like it was before.

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,81 @@
...
-        const values = form.getValues();
-        save(values);
-      }
-    };
-
-    handleSave();
-  }, [watchedData]);
+    const values = formRef.current.getValues();
+    debouncedOnUpdate(values);
+  }, [watchedData, debouncedOnUpdate, formRef]);
```
LetItRock: This part handles the form `onChange` events and is debouncing it for a 10s

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,81 @@
...
+  const flushFormUpdates = (): Promise<void> => {
+    return new Promise((resolve) => {
+      // await for the state to be updated
+      setTimeout(async () => {
+        await onFlushInternal();
+
+        resolve();
+      }, 0);
+    });
+  };
```
LetItRock: Flushes all the form updates and calls the `update` function if form is dirty and valid.
Implemented with a promise and `setTimeout` because we need to give a change to form to update it's internal state before checking if it's dirty/valid.

apps/dashboard/src/hooks/use-invocation-queue.ts
```diff
@@ -0,0 +1,57 @@
+import { useCallback, useRef } from 'react';
+
+type CallbackFunction = () => Promise<unknown>;
+
+export function useInvocationQueue<T extends CallbackFunction = CallbackFunction>({
```
LetItRock, line 5: The hook that queues all the autosave api calls and runs them one by one in order, to guarantee that there wouldn't be any race conditions between users changes.

apps/dashboard/src/hooks/use-form-autosave.ts
LetItRock: yes, but we don't care actually because it's updated with the latest form values, and also because of the above comment

apps/dashboard/src/components/create-workflow-button.tsx
```diff
@@ -152,7 +152,11 @@ export const CreateWorkflowButton = (props: CreateWorkflowButtonProps) => {
                       <FormLabel hint={`(max. ${MAX_TAG_ELEMENTS})`}>Add tags</FormLabel>
                     </div>
                     <FormControl>
-                      <TagInput suggestions={tagsQuery.data?.data.map((tag) => tag.name) || []} {...field} />
```
SokratisVidros, line 155: ❓ How did this work before without the value?

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -96,11 +91,11 @@ export const ConfigureWorkflowForm = (props: ConfigureWorkflowFormProps) => {
           </div>
         </SidebarHeader>
         <Separator />
-        <Form {...statusForm}>
-          <form>
+        <Form {...form}>
+          <form className="h-full" onBlur={onBlur}>
```
SokratisVidros, line 89: Does this onBlur work? The blur [event doesn't bubble](https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event).

apps/dashboard/src/components/workflow-editor/steps/configure-step-template-form.tsx
```diff
@@ -27,55 +28,65 @@ export type StepEditorProps = {
...
+    shouldFocusError: false,
+  });
+
+  const { onBlur, flushFormUpdates } = useFormAutosave({
+    previousData: defaultValues,
+    form,
+    save: (data) => {
+      update({
+        ...workflow,
+        steps: workflow.steps.map((s) => {
```
SokratisVidros: We will have to tweak steps and other nested structures several times. I suggest creating helper functions to handle such data changes. For example `update(buildNewStep(...))`;

This is not a blocker for this PR.

rgozdzialski:
@LetItRock Looking at the screen recordings, it looks like the user needs to click away in order for the saving to be done. Is this the case for any change in the canvas? It would be helpful to have a list of logic: what happens on blur, what on click, what on add, what on submit, what on keystroke, etc. This will help us identify potential problems early.

Not release-critical, but a good thing to have long with this PR.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-action.tsx
SokratisVidros: Please add it as a comment.

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,81 @@
...
+    e.stopPropagation();
+    propsForm.handleSubmit(onSave)(e);
+  };
+
+  // flush the form updates right away
+  const flushFormUpdates = (): Promise<void> => {
+    return new Promise((resolve) => {
+      // await for the state to be updated
+      setTimeout(async () => {
+        await onFlushInternal();
```
SokratisVidros: Nit picky comment: I'd inline onFlushInterval here to improve readability.

SokratisVidros (changes requested):

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -197,6 +188,10 @@ export const ConfigureWorkflowForm = (props: ConfigureWorkflowFormProps) => {
                     <FormControl className="text-xs text-neutral-600">
                       <TagInput
                         {...field}
+                        onChange={(tags) => {
+                          form.setValue('tags', tags, { shouldValidate: true, shouldDirty: true });
+                          flushFormUpdates();
+                        }}
```
desiprisg, line 188: One of the things that I'd like to be DRYers is this. It will get repetitive to add this logic on every field that doesn't require a blur. Maybe it's better to do `useFormAutosave()` with a 0 sec delay that will only affect specified fields (which will be those who don't update on blur).

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-action.tsx
```diff
@@ -30,12 +30,15 @@ import { cn } from '@/utils/ui';
 import { urlTargetTypes } from '@/utils/url';
 import { autocompletion } from '@codemirror/autocomplete';
 import { useStep } from '@/components/workflow-editor/steps/step-provider';
+import { useFlushFormUpdates } from '../flush-form-updates-context';
```
desiprisg: Please use absolute imports :D

apps/dashboard/src/hooks/use-form-autosave.ts
desiprisg: The only issue I can think of with resetting the form with its current values is that in case of the response being different (a field changes upon submission by the api or an error is returned for it), we're left with wrong data.

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
LetItRock: It's React event bubbling not DOM
Screenshot 2024-11-29 at 11 51 54

apps/dashboard/src/components/create-workflow-button.tsx
LetItRock: not sure, but might not be needed, I'll recheck

apps/dashboard/src/hooks/use-form-autosave.ts
LetItRock: ok, will do

apps/dashboard/src/components/create-workflow-button.tsx
LetItRock: the same is applied in the configure workflow sidebar and was added by George... tags are optional in the schema so this is to fix the ts issue

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
LetItRock: yes, but this creates another issue of independent forms (which in the UI is one form), and if one has the validation error, the other one should not be allowed to submit

apps/dashboard/src/hooks/use-form-autosave.ts
LetItRock: talked about that in DM

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
SokratisVidros: Oh yes! you are right!

BiswaViraj (approved):

apps/dashboard/src/components/workflow-editor/steps/configure-step.tsx
```diff
@@ -5,7 +5,7 @@ import { useEnvironment } from '@/context/environment/hooks';
 
 export const ConfigureStep = () => {
   const { step } = useStep();
-  const { workflow, debouncedUpdate, update, onDirtyChange } = useWorkflow();
+  const { workflow, update, onDirtyChange } = useWorkflow();
```
ChmaraX: 💡  not related to this PR just a thought

I feel like we should merge `useStep` to have just one `useWorkflow`, we almost always use them together since step save operation is calling update from useWorklow etc or they're somehow tangled together.

The only difference is the fetch request where step has a standalone endpoint but that could be solved by some minor adjustment like:

```jsx
const { workflow, update, onDirtyChange, step } = useWorkflow(stepId);
```

I would imagine the ideal DX something like you want to use/modify workflow -- here is the single hook to do that which has all the tools.

ChmaraX (approved):

LetItRock:
PLEASE DO NOT MERGE YET!

apps/dashboard/src/components/workflow-editor/steps/configure-step.tsx
LetItRock: I think it's fine for now as we are aiming to use separate API calls for the steps soon.

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
```diff
@@ -20,61 +20,50 @@ import { Switch } from '../primitives/switch';
...
 
-  const statusForm = useForm({
-    defaultValues: {
-      active: workflow.active,
-    },
+  const { onBlur, flushFormUpdates } = useFormAutosave({
+    previousData: workflow,
+    form,
+    isReadOnly,
+    save: update,
```
SokratisVidros, line 52: Nit-picky one: I suggest we use the same name and safe a few bytes :p

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
SokratisVidros: See [here](https://github.com/novuhq/novu/pull/7164/files#r1863233547)

apps/dashboard/src/components/workflow-editor/toasts.tsx
```diff
@@ -0,0 +1,57 @@
+import { ToastIcon } from '@/components/primitives/sonner';
```
SokratisVidros, line 1: 😍 Yes!

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,91 @@
...
+
+  // flush the form updates right away
+  const flushFormUpdates = (): Promise<void> => {
+    return new Promise((resolve) => {
+      // await for the state to be updated
+      setTimeout(async () => {
+        // use the form reference instead of destructuring the props to avoid stale closures
+        const form = formRef.current;
+        const values = form.getValues();
+        await onSave({ ...values });
```
SokratisVidros:
```suggestion
        await onSave(values);
```

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -1,24 +1,91 @@
...
+
+      // cancel the pending debounces for example on change events
+      debouncedOnSave.cancel();
+      onSave(values);
+    },
+    [formRef, onSave, debouncedOnSave]
+  );
+
+  // flush the form updates right away
+  const flushFormUpdates = (): Promise<void> => {
```
SokratisVidros:
```suggestion
  const flushForm = (): Promise<void> => {
```

or `saveForm`

SokratisVidros (approved):
👏 👏

apps/dashboard/src/components/workflow-editor/configure-workflow-form.tsx
LetItRock: I would stay with `save` as you suggested to rename `flushFormUpdates` to `saveForm`

PR #7173, opened 2024-11-30
fix(api): centralize upsert validation  + improve nested error handling 

tatarco:
feat(api): add upsert validation

1. Established a single validation entry point for Upsert use case.
2. Fixed issue with nested validation based on @djabarovgeorge  solution in an old open PR.
3. Removed the Base Command duplication so all commands use the same validation mechanism.
4. Fix Wrong Validations around issues in internal commands ( update + create) 
5. Enhance the Filter to produce better and more detailed  400 errors around validation issues.
6. Make sure post process process update issues if no issues exist instead of empty object, the class validator does not handle empty object well for Record<string,?>
7. Added tests to validate correct behavior around tags and missing step name, make sure both fire good validation 400 erors

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
```diff
@@ -72,7 +72,7 @@ export function toWorkflowsMinifiedDtos(templates: NotificationTemplateEntity[])
 }
 
 function toStepResponseDto(persistedStep: NotificationStepEntity): StepResponseDto {
-  const stepName = persistedStep.name || 'Missing Name';
+  const stepName = persistedStep.name || 'Missing Name || UPDATE IMMEDIATELY';
```
tatarco: This needs to be here, either that or an exception in case of missing name, the entity allows missing names, but our API no more @SokratisVidros

SokratisVidros: The entity shouldn't allow for missing names, and the API should return a 422 or 400 error if the name is missing for a step.

apps/api/src/app/inbox/e2e/update-notification-action.e2e.ts
```diff
@@ -109,6 +110,15 @@ describe('Update Notification Action - /inbox/notifications/:id/{complete/revert
       _templateId: template._id,
     })) as MessageEntity;
   });
+  function assertValidationMessages(e: AxiosError<any, any>, field: string, msg1: string) {
+    if (!(e instanceof AxiosError)) {
+      throw new Error(e);
+    }
+    console.log(JSON.stringify(e.response?.data));
```
djabarovgeorge: do we need a console log here?

```suggestion
```

apps/api/src/app/inbox/e2e/update-notification-action.e2e.ts
```diff
@@ -109,6 +110,15 @@ describe('Update Notification Action - /inbox/notifications/:id/{complete/revert
       _templateId: template._id,
     })) as MessageEntity;
   });
+  function assertValidationMessages(e: AxiosError<any, any>, field: string, msg1: string) {
```
djabarovgeorge: is this function used?

apps/api/src/exception-filter.ts
```diff
@@ -96,6 +80,41 @@ export class AllExceptionsFilter implements ExceptionFilter {
       message: `Internal server error, contact support and provide them with the errorId`,
     };
   }
+  private handleCommandValidation(exception: CommandValidationException): ResponseMetadata {
+    const DO_NOT_TRACK_CLASSES = [ProcessBulkTriggerCommand.name];
```
djabarovgeorge: why are we choosing not to track this use case?

apps/api/src/exception-filter.ts
```diff
@@ -96,6 +80,41 @@ export class AllExceptionsFilter implements ExceptionFilter {
...
+
+    let uuid: string | undefined;
+    if (!DO_NOT_TRACK_CLASSES.includes(exception.className)) {
+      uuid = this.getUuid(exception);
+      this.logError(uuid, exception);
```
djabarovgeorge: we added last week log for all 4XX in buildResponseBody the same place that 5XX logs. so i think we can remove this one.

apps/api/src/exception-filter.ts
```diff
@@ -96,6 +80,41 @@ export class AllExceptionsFilter implements ExceptionFilter {
...
+    }
+
+    return {
+      message: {
+        message: exception.message,
+        cause: exception.constraintsViolated,
```
djabarovgeorge: can you provide an example of the error we had previously and what we are getting now? I don’t recall seeing cause in the past, but it’s possible I haven’t interacted with command errors recently.

libs/application-generic/src/commands/base.command.ts
djabarovgeorge, line 1: can you provide an example of the response we would get from this execution?

djabarovgeorge, line 1: we need to update base.command.spec.ts with new tests to cover the new functionality.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/preferences-request-upsert.data.ts
```diff
@@ -0,0 +1,6 @@
+import { WorkflowPreferences } from '@novu/shared';
+
+export class PreferencesRequestUpsertData {
+  user: WorkflowPreferences | null;
+  workflow?: WorkflowPreferences | null;
```
djabarovgeorge: we need to add decorators here with new commands for user workflow

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.dto.ts
```diff
@@ -0,0 +1,55 @@
...
+  Length,
+  ValidateNested,
+} from 'class-validator';
+import { Type } from 'class-transformer';
+import { MAX_NAME_LENGTH } from '@novu/application-generic';
+import { WorkflowCreationSourceEnum } from '@novu/shared';
+import { UpsertStepData } from './upsert-step.data';
+import { PreferencesRequestUpsertData } from './preferences-request-upsert.data';
+
+export class UpsertWorkflowDto {
```
djabarovgeorge: **General note:**
now that we have a separate class for the API we need to rethink how we maintain the DTO's in shared, because not we could have misalignments

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-step.data.ts
```diff
@@ -0,0 +1,21 @@
+import { IsEnum, IsNotEmpty, IsOptional, IsString, Length } from 'class-validator';
+import { MAX_NAME_LENGTH } from '@novu/application-generic';
+import { StepTypeEnum } from '@novu/shared';
+
+export class UpsertStepData {
```
djabarovgeorge:
```suggestion

export class UpsertStepDataCommand {
```

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.dto.ts
djabarovgeorge: The usecase should not care that this is a DTO, this should be a Command.

```suggestion
export class UpsertWorkflowCommand {
```

apps/api/src/app/workflows-v2/usecases/upsert-workflow/preferences-request-upsert.data.ts
```diff
@@ -0,0 +1,6 @@
+import { WorkflowPreferences } from '@novu/shared';
+
+export class PreferencesRequestUpsertData {
```
djabarovgeorge:
```suggestion
export class PreferencesRequestUpsertDataCommand {
```

djabarovgeorge (commented):
Looks amazing i love that we moved forward on adding nested validation on command.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.dto.ts
SokratisVidros: Correct, the DTO needs to be an interface in `@novu/shared`. There is another draft PR where I replace all classes in DTOs with interfaces to make strict TS happy.

apps/api/src/app/workflows-v2/usecases/post-process-workflow-update/post-process-workflow-update.usecase.ts
```diff
@@ -94,7 +92,10 @@ export class PostProcessWorkflowUpdate {
     return issue.body && Object.keys(issue.body).length > 0;
   }
 
-  private validateSteps(steps: NotificationStepEntity[], _workflowId: string): Record<string, StepIssuesDto> {
+  private validateSteps(
+    steps: NotificationStepEntity[],
+    _workflowId: string
+  ): Record<string, StepIssuesDto> | undefined {
```
SokratisVidros, line 98: Nit-picky one:

```suggestion
  ): Record<string, StepIssuesDto> | null {
```

In TS null feels more explicit than undefined as undefined.

apps/api/src/app/workflows-v2/usecases/post-process-workflow-update/post-process-workflow-update.usecase.ts
```diff
@@ -155,18 +156,22 @@ export class PostProcessWorkflowUpdate {
 
   private updateIssuesOnWorkflow(
     workflow: WorkflowInternalResponseDto,
-    workflowIssues: Record<keyof WorkflowResponseDto, RuntimeIssue[]>,
-    stepIssuesMap: Record<string, StepIssues>
+    stepIssuesMap?: Record<string, StepIssues>,
```
SokratisVidros, line 159: This is an attempt to align workflow and step issue type definition as part of this function

```suggestion
    stepIssues?: Record<keyof StepResponseDto, StepIssues>,
```

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.dto.ts
SokratisVidros: Don't we also need an optional `_id` field for the Mongo Object ID?

Gal: it comes as a separate parameter

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
tatarco: I'm resolving the comment, we are discussing it offfline

apps/api/src/app/inbox/e2e/update-notification-action.e2e.ts
tatarco: answer is always no, thanks

apps/api/src/app/inbox/e2e/update-notification-action.e2e.ts
tatarco: great catch

apps/api/src/exception-filter.ts
tatarco: because it's a public API, why do I want to track 4xx errors in it? there is nothing to do with this information besides maybe approach the customer, but if the error is clear enough I see no reason to deal with it, it's not giving us any specific input to a bug in our system, maybe the docs? I just wanted this functionality to exist, the ability to quite down usecases that we don't care about

apps/api/src/exception-filter.ts
tatarco: this is not a common practice, why do we log all 4XX errors?

tatarco: And did you notice it doesn't go to sentry? is it on purpose? I would like to talk about this and make sure we got all the usecases covered

apps/api/src/exception-filter.ts
tatarco: I've refactored the structure of the validation in a similar way to what you did, but before it will just shoot out a string list in the message now we have a more structured response, you can see the old structure in the tests changed 
the actual value in cause now is: 
    public constraintsViolated: Record<string, ConstraintValidation>,

class ConstraintValidation {
  messages: string[];
  value: any;
}

libs/application-generic/src/commands/base.command.ts
tatarco: {status: 400, 
message: "Failed Validation",
cause: {name: {value: "", messages:['Name cannot be empty']
}

tatarco: the functionality is being tested in other e2e tests and being asserted

tatarco: no need to cover multiple times the same functionality

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.dto.ts
tatarco: Hold your horses people, interfaces and classes are the same for that reason 
image
The solution here, and it will take probably a month more is to move all DTO's into the API and have our open api generator generate them as a part of the SDK, this is the only thing that will achieve the correct seperation here, that's next on the list for the SDK work, I would appreciate if we wait for that

apps/api/src/app/workflows-v2/usecases/post-process-workflow-update/post-process-workflow-update.usecase.ts
tatarco: this gets persisted, I prefer to use undefined here as it goes into mongo, since it's a nit-pick I choose to not take the proposal here

apps/api/src/app/workflows-v2/usecases/post-process-workflow-update/post-process-workflow-update.usecase.ts
tatarco: you missunderstood the parameter, the string is the stepId

apps/api/src/exception-filter.ts
djabarovgeorge: please check the [PR](https://github.com/novuhq/novu/pull/7225) i explained the reasoning there.

apps/api/src/exception-filter.ts
djabarovgeorge: due to a recent error [REF](https://novu-r9.sentry.io/issues/6111320858/?alert_rule_id=15115141&alert_type=issue&notification_uuid=37c1028d-5ad6-442c-885b-874dd9ffb1ac&project=6250907&referrer=slack), I believe one reason to log such errors is that they can help identify edge cases or bugs similar to what occurred last week.

apps/api/src/app/workflows-v2/usecases/post-process-workflow-update/post-process-workflow-update.usecase.ts
djabarovgeorge: This is one of the reasons it is hard for me to work with Record<string, Foo> there is no good way to understand the object by the type.

libs/application-generic/src/commands/base.command.ts
djabarovgeorge: one of the issues with our API is the existence of multiple error types. It would be a good practice to use a single interface for all errors, making it easier for API clients to parse and handle them. I would prefer not to introduce a new error type into the system


The errors we have today are the following:

**command response**
{
  "message": [
    "type should not be null or undefined",
    "type must be one of the following values: in_app, email, sms, chat, push, digest, trigger, delay",
    "content should not be null or undefined"
  ],
  "error": "Bad Request",
  "statusCode": 400
}

**exception response**
{
  "message": "Please provide a valid CTA action",
  "error": "Bad Request",
  "statusCode": 400
}

**current nested validation response**
{
  "message": [],
  "error": "Bad Request",
  "statusCode": 400
}

djabarovgeorge: there is no need to test `Command` functionality in multiple places. We should trust it works as expected when we use it and it is by those unit tests we have. 

in other cases, we can test the `Usecase` functionality,  where we can test the `Usecase` interface that acts as expected it is not a `Command` functionality.

apps/api/src/exception-filter.ts
djabarovgeorge: Could you elaborate on why you chose this more structured object other than `"message": []`. is there a common practice to do it in this way to the API client can easier to consume it?

In addition, i am not sure i fully understand what the structure and what `string` stands for in:
`Record<string, ConstraintValidation>`
because if it is the first layer variables, i do not understand how this object is supporting nested validation. in another comment, you provided the following example:

> {
>     status: 400,
>     message: "Failed Validation",
>     cause: {name: {value: "", messages:['Name cannot be empty']
> }

But how will it represent an error if we have an issue with the following structure as an example `workflow.steps[3].template.name`

apps/api/src/app/workflows-v2/usecases/upsert-workflow/preferences-request-upsert-data.command.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  @IsObject()
+  @ValidateNested({ each: true })
+  channels: Record<ChannelTypeEnum, ChannelPreferenceData>;
+}
+
+export class PreferencesRequestUpsertDataCommand {
+  @IsOptional()
+  @ValidateNested()
+  @IsInstance(WorkflowPreferencesUpsertData)
```
djabarovgeorge: I am not familiar with @IsInstance as far as I know, and I tested the nested validation works well with @Type() of the class transformer. do we have a test that covers this functionality?

tatarco: it was a chatgpt glich, thank you for pointing out, changed it to type

apps/api/src/exception-filter.ts
tatarco: i'm removing it, let eveything be logged

apps/api/src/exception-filter.ts
tatarco: all resolved offline

apps/api/src/exception-filter.ts
tatarco: it will show 
{
status: 400 
message: "Failed Validation" 
cause :{
"workflow.steps[3].template.name": "name cannot be empty"
}
}
API client shows a body of the error, there is no standard practice here, it needs to be a clear error, also we don't automatically consume it, it just needs to be clear and readable, just dumping a json in the message doesn't make it more readable

libs/application-generic/src/commands/base.command.ts
tatarco: correct, and once an instance is tested it's enough to prove the functionality works.

tatarco: we have currently exactly one interface returning as an error
I added CTX to include additional context info
image
There are no surprises, specific errors can have additional context, the message is controlled by the message thrown from the exception

SokratisVidros (approved):

PR #7180, opened 2024-12-02
feat(dashboard): update delay control schema

ChmaraX:
### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/shared/schemas/delay-control.schema.ts
```diff
@@ -26,7 +26,7 @@ export const delayUiSchema: UiSchema = {
   properties: {
     amount: {
       component: UiComponentEnum.DELAY_AMOUNT,
-      placeholder: '30',
```
ChmaraX, line 29: we can't use a specific value here otherwise user will see both value and error - we don't want to support automatic default values on backend when step is inserted.

image

ChmaraX:
Closed - will be open as part of main PR.

PR #7187, opened 2024-12-02
fix(dashboard): don't use defaultValues for async content

ChmaraX:

Details: https://linear.app/novu/issue/NV-4941/step-form-values-are-not-being-properly-updated-after-revisiting

https://github.com/user-attachments/assets/8dd1ec79-7021-4090-84cb-ec4f72a3f2c9

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
+++ b/apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
@@ -57,10 +57,16 @@ export const ConfigureStepForm = (props: ConfigureStepFormProps) => {
     navigate(buildRoute(ROUTES.EDIT_WORKFLOW, { environmentSlug: environment.slug!, workflowSlug: workflow.slug }));
   };
 
+  const defaultValues = {
+    name: step.name,
+    stepId: step.stepId,
+  };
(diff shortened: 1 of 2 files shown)


apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -57,10 +57,16 @@ export const ConfigureStepForm = (props: ConfigureStepFormProps) => {
...
+  };
+
   const form = useForm<z.infer<typeof stepSchema>>({
-    defaultValues: {
-      name: step.name,
-      stepId: step.stepId,
+    defaultValues,
+    values: {
+      ...defaultValues,
+      ...step.controls.values,
```
ChmaraX, line 69: According to some posts `values` should be used for values that depend on async return and `defaultValues` on static fields, but the react-hook-form docs states various other things, quite confusing.

SokratisVidros: TBH I am not sure I understand the fix. If it works it works but I suggest adding a comment in the code.

SokratisVidros (approved):

PR #7186, opened 2024-12-02
fix(dashboard): in-app editor preview only call when tab is opened

LetItRock:
### What changed? Why was the change needed?

In-App Editor Preview - the api request for the preview should only be called when the tab is opened or when the `apply` button is pressed. 
In the attached video, there are 2x calls to the API because the React `effect` is called twice on the mount.

### Screenshots

https://github.com/user-attachments/assets/f0eb9102-10df-476d-bf5a-4b532a0e6ac0

https://github.com/user-attachments/assets/aa75a74f-721d-4d32-8624-b177250e7cd7

https://github.com/user-attachments/assets/4a74ed39-05bc-451e-ad6f-28e537731fab

https://github.com/user-attachments/assets/96e89175-af81-41c6-880f-46940ac56821

PR #7190, opened 2024-12-02
feat(api): gracefully preview

djabarovgeorge:
### Screenshots

<details>

</details>


SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -35,37 +37,53 @@ export class GeneratePreviewUsecase {
...
-    };
+      return {
+        result: {
+          preview: executeOutput.outputs as any,
+          type: stepData.type as unknown as ChannelTypeEnum,
+        },
+        previewPayloadExample: variablesExample,
+      };
+    } catch (error) {
+      Logger.error(
```
tatarco: That won't go to sentry, and it's the wrong logger.
Need a better solution here

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -35,37 +37,53 @@ export class GeneratePreviewUsecase {
...
+        `Unexpected error while generating preview`,
+        LOG_CONTEXT
+      );
+
+      return {
+        result: {
+          preview: {},
+          type: undefined,
+        },
+        previewPayloadExample: {},
```
tatarco, line 95: That is not better UX the failure

tatarco (changes requested):

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: We’re working on handling edge cases to ensure the client doesn’t fail.  
One example is dealing with a stale control schema stored in the database.  
How would you approach this with a better DX?

tatarco: I wrote UX, is an empty preview better than a toast error? I don't think so

Regarding the stale controls, whatever which OS under our control should not have backwards compatibility issues, the render can branch to support both v1 and v2 of a step, the errors should be rare yet we cannot stop them all, now the question is what UX is best for such issue while making sure we get alerted to fix the issue, this logger doesn't really log, it doesn't go to sentry and the response will still feel broken

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
SokratisVidros: Good catch. @djabarovgeorge please also add an `import { captureException } from '@sentry/node';`

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
SokratisVidros: Preview will only be invoked when the preview tab is selected or the mini preview is rendered - [See PR](https://github.com/novuhq/novu/pull/7186). 

The debounced preview mechanism after the introduction of autosave is redundant. Moreover, errors from the preview are not actionable by the user, as all issues will surface from the autosave. Preview, being a best-effort endpoint, will try to render as much as possible or return nothing (via a successful response).

Lastly, I agree that preview errors must be eliminated; that's why we report them to Sentry so that we can debug them, but this is an API concern. As far as the FE is concerned, the preview endpoint should always return.

tatarco: as long as they are in sentry i'm ok, I'm just wandering maybe a success with some error preview syntax will be better than nothing.
I think a preview with this: 
 "error has accrued, we are working on it , here is the error id to follow up with support.
 is better than just empty.  @SokratisVidros

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: missed this one, good catch indeed!

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -35,37 +38,57 @@ export class GeneratePreviewUsecase {
...
-    };
+      return {
+        result: {
+          preview: executeOutput.outputs as any,
+          type: stepData.type as unknown as ChannelTypeEnum,
+        },
+        previewPayloadExample: variablesExample,
+      };
+    } catch (error) {
+      Logger.error(
```
tatarco: This is the wrong logger, you should use pino instead

tatarco (approved):
please just update the correct logger , make sure the log has the sentry id as well

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
SokratisVidros: We will get that from Sentry and act on it. We also added a "No preview available" in the frontend that should address the concern.

PR #7192, opened 2024-12-02
refactor(api): generate payload in runtime

djabarovgeorge:
https://linear.app/novu/issue/NV-4938/calculate-payload-schema-for-the-dashboard-preview-and-trigger-pages

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/build-test-data/build-workflow-test-data.usecase.ts
```diff
@@ -10,29 +15,102 @@ import {
...
+  }
+
+  private extractTemplateVariables(controlValues: Array<{ controls: unknown | null }>): string[] {
+    const flattenedControls = controlValues
+      .map((item) => item.controls)
+      .filter((control): control is NonNullable<typeof control> => control != null);
+
+    const concatenatedControlValues = flattenedControls.map(flattenObjectValues).flat().join(' ');
+
+    return extractTemplateVars(concatenatedControlValues);
```
djabarovgeorge: after this week learning i know that this will throw an error if the control value has an invalid state because we use liquid js there.

i need to double-check this implementation.

djabarovgeorge: solved, we will extract valid and invalid separately.

apps/api/src/app/workflows-v2/util/template-variables/extract-template-variables.ts
```diff
@@ -1,37 +1,79 @@
...
+
+  let variables = new Set<string>();
+  const invalidVariables: { context: string; message: string; variable: string }[] = [];
+
+  for (const liquidExpression of liquidExpressions) {
+    try {
+      variables = new Set([...variables, ...parseByLiquid(liquidExpression)]);
+    } catch (error) {
+      invalidVariables.push(
+        ...error.errors.map((e) => ({ context: e.context, message: e.message, variable: liquidExpression }))
```
djabarovgeorge: This object is not 100% correct because here i am passing liquid js Output and not a variable.
Id suggest keeping it like this because there is no valid(transitivity) way to know what the variable is in an invalid Output.
wanted to hear your thoughts on this.

```suggestion
        ...error.errors.map((e) => ({ context: e.context, message: e.message, output: liquidExpression }))
```

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -120,18 +118,6 @@ export class GeneratePreviewUsecase {
     );
   }
 
-  @Instrument()
```
djabarovgeorge, line 123: not used

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const templateVars = this.extractTemplateVariables(controlValues);
+    if (templateVars.length === 0) {
+      return {};
+    }
+
+    const variablesExample = pathsToObject(templateVars, {
+      valuePrefix: '{{',
+      valueSuffix: '}}',
+    }).payload as Record<string, unknown>;
```
SokratisVidros: Is this type of assertion required? If not mistaken `pathsToObjects` return `Record<string, unknown>`

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    return convertJsonToSchemaWithDefaults(variablesExample);
+  }
+
+  private async buildControlValues(command: BuildPayloadSchemaCommand) {
```
SokratisVidros, line 36: @djabarovgeorge If I understand correctly, we use the provided control values and fetch them from the database if no control values are provided. Is there a scenario that we need to merge the provided values with the ones fetched from the database?

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -0,0 +1,61 @@
...
+    let aggregateControlValues = command.controlValues ? [command.controlValues] : [];
```
SokratisVidros: Nit picky comment: The aggregated prefix made me believe that we are merging control values. 

```suggestion
    let controlValus = command.controlValues ? [command.controlValues] : [];
```

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!aggregateControlValues.length) {
+      aggregateControlValues = (
+        await this.controlValuesRepository.find({
+          _environmentId: command.environmentId,
+          _organizationId: command.organizationId,
+          _workflowId: command.workflowId,
+          level: ControlValuesLevelEnum.STEP_CONTROLS,
+        })
+      )
+        .map((item) => item.controls)
```
SokratisVidros: Can we combine the mapping and the filtering into the mongoose query so that we fetch less data between Mongo and the application server? 

For example: 

```
{ controls: { $ne: null } }, // Filter to exclude null 'controls'
{ controls: 1, _id: 0 }     // Projection: Include 'controls', exclude '_id'
```

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -0,0 +1,61 @@
...
+        .filter((control): control is NonNullable<typeof control> => control != null);
+    }
+
+    return aggregateControlValues;
+  }
+
+  @Instrument()
+  private extractTemplateVariables(aggregateControlValues: Record<string, unknown>[]): string[] {
+    const concatenatedControlValues = aggregateControlValues.map(flattenObjectValues).flat().join(' ');
```
SokratisVidros:
```suggestion
    const controlValuesString = aggregateControlValues.map(flattenObjectValues).flat().join(' ');
```

apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
```diff
@@ -39,14 +44,35 @@ export class BuildAvailableVariableSchemaUsecase {
...
+
+  @Instrument()
+  private async resolvePayloadSchema(
+    workflow: NotificationTemplateEntity,
+    command: BuildAvailableVariableSchemaCommand
+  ): Promise<JSONSchemaDto> {
+    if (workflow.payloadSchema) {
+      return (
+        parsePayloadSchema(workflow.payloadSchema, { safe: true }) || {
+          type: 'object',
```
SokratisVidros: This parsePayloadSchema invocation is different from the one triggered in apps/api/src/app/workflows-v2/usecases/build-test-data/build-workflow-test-data.usecase.ts. If not mistaken, they should be aligned.

apps/api/src/app/workflows-v2/usecases/overload-content-data/overload-content-data-on-workflow.usecase.ts
```diff
@@ -44,9 +41,12 @@ export class OverloadContentDataOnWorkflowUseCase {
...
       const controlValues = stepIdToControlValuesMap[step._templateId];
 
-      const jsonSchemaDto = this.buildAvailableVariableSchemaUsecase.execute({
+      const jsonSchemaDto = await this.buildAvailableVariableSchemaUsecase.execute({
+        environmentId: user.environmentId,
+        organizationId: user.organizationId,
+        userId: user._id,
         workflow,
-        stepDatabaseId: step._templateId,
+        stepInternalId: step._templateId,
```
SokratisVidros, line 49: 😍

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.spec.ts
```diff
@@ -0,0 +1,97 @@
+import { expect } from 'chai';
+import { extractLiquidTemplateVariables } from './liquid-parser';
+
+describe('parseLiquidVariables', () => {
```
SokratisVidros, line 4: Lovely unit testing

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -0,0 +1,126 @@
+import { Template, Liquid, RenderError, LiquidError } from 'liquidjs';
```
SokratisVidros, line 1: Do you think this should live in application generic package?

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -0,0 +1,126 @@
...
+ * }
+ *
+ * @example
+ * // Mixed valid and invalid syntax
+ * parseLiquidVariables('{{user.name}} {{invalid..syntax}}')
+ * // Returns:
+ * {
+ *   validVariables: ['user.name'],
+ *   invalidVariables: [{
+ *     context: '>> 1| {{invalid..syntax}}\n                ^',
```
SokratisVidros, line 55: Do we need the `context` key in invalidVariables?

apps/api/src/app/workflows-v2/util/template-parser/parser-types.ts
```diff
@@ -0,0 +1,10 @@
+export type InvalidVariable = {
```
SokratisVidros: Consider moving these types into the liquide-parse.ts file to save us from yet another file.

apps/api/src/app/workflows-v2/util/template-parser/parser-types.ts
```diff
@@ -0,0 +1,10 @@
...
+  context: string;
+  message: string;
+  variable: string;
+};
+
+export type TemplateParseResult = {
+  validVariables: string[];
```
SokratisVidros: It's better DX for this parse function to return the same type of variables for valid and invalid. That is consider refactoring the types as follows:

```
export type Variable = { name: string, message? string, context?: string };
export type TemplateParseResult = { validVariables: Variable[], invalidVariables: Variable [] }
```

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * @param template - The Liquid template string to parse
+ * @returns Object containing arrays of valid and invalid variables found in the template
+ */
+export function extractLiquidTemplateVariables(template: string): TemplateParseResult {
+  if (!isValidTemplate(template)) {
+    return { validVariables: [], invalidVariables: [] };
+  }
+
+  const liquidRawOutput = extractLiquidExpressions(template);
```
SokratisVidros, line 69: @djabarovgeorge, Can you please explain why we use extra expressions using regex before passing them to Liquid? I'd expect Liquid to get the raw string and do the calculations.

SokratisVidros (changes requested):
Nicely done @djabarovgeorge. I added some nitty gritty comments that should be easily addressed but overall the flow is crystal clear.

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      const parsedVars = parseByLiquid(rawOutput);
+      parsedVars.forEach((variable) => variables.add(variable));
+    } catch (error: unknown) {
+      if (isLiquidErrors(error)) {
+        invalidVariables.push(
+          ...error.errors.map((e: RenderError) => ({
+            context: e.context,
+            message: e.message,
+            variable: rawOutput,
```
djabarovgeorge: raising again this matter, I wonder if we should change, or rename the name.

```suggestion
            rawOutput: rawOutput,
```

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
djabarovgeorge: Awesome question, this is the way we can provide `valid` and `invalid`.
because once we parse we get an exception.
we made a workaround it, so now we will parse every liquid js Output separately. 
That will allow as to return both `valid` and `invalid` variables.
this one is a bit complex, we can sync about it if needed.

apps/api/src/app/workflows-v2/util/template-parser/parser-types.ts
djabarovgeorge: No problem, i will merge it i wanted to split the logic as much as possible.

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
djabarovgeorge: In my opinion, we could benefit from it. this is how the CLI presents the issues.
Let me provide an example or given example:
context: '>> 1| {{invalid..syntax}}\n                ^',

the client will have the exact place where the issue is if it is render/split by a new line
`>> 1| {{invalid..syntax}}`
`                ^`

this is an indication that Liquid JS sees an error

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
djabarovgeorge: IMO once we reuse it in another place, we should move it to app-gen.

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
djabarovgeorge: Good question, not that i am aware of.
The provided control values are `only` for the preview purposes (i added comment in the command about it as well to make it more clear), and if it is provided it is the un-saved source of truth for control value otherwise, the stored control values are the source of truth.

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
djabarovgeorge: I refactored the code a lot, and added more types along the way. so maybe it is redundant right now.
thanks for the comment i will double-check it.

apps/api/src/app/workflows-v2/util/template-parser/parser-types.ts
djabarovgeorge: Love it, my main problem is this part https://github.com/novuhq/novu/pull/7192/files#r1871222390
if you want we can sync about it

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
SokratisVidros: Thanks for the explanation.

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
SokratisVidros: What about `name` as I suggested in the other comment?

djabarovgeorge: we decided to rename to `name`

SokratisVidros (approved):

PR #7197, opened 2024-12-03
chore(dashboard): beta label

LetItRock:

Add the Beta label to the new Dashboard.

### Screenshots
Screenshot 2024-12-03 at 11 20 25

Screenshot 2024-12-03 at 11 18 18
Screenshot 2024-12-03 at 11 18 09

### DIFF
--- a/apps/dashboard/src/components/primitives/badge.tsx
+++ b/apps/dashboard/src/components/primitives/badge.tsx
@@ -7,7 +7,7 @@ const badgeVariants = cva(
   {
     variants: {
       variant: {
-        neutral: 'border-neutral-500 bg-neutral-500',
+        neutral: 'border-neutral-100 bg-neutral-100 text-neutral-500',
         destructive: 'border-transparent bg-destructive/10 text-destructive',
         success: 'border-transparent bg-success/10 text-success',
         warning: 'border-transparent bg-warning/10 text-warning',
--- a/apps/dashboard/src/components/primitives/badge.tsx
+++ b/apps/dashboard/src/components/primitives/badge.tsx
@@ -3,7 +3,7 @@ import { cva, type VariantProps } from 'class-variance-authority';
 import * as React from 'react';
 
 const badgeVariants = cva(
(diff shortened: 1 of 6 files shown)


apps/dashboard/src/components/primitives/badge.tsx
```diff
@@ -14,24 +14,29 @@ const badgeVariants = cva(
         soft: 'bg-neutral-alpha-200 text-foreground-500 border-transparent',
         outline: 'border-neutral-alpha-200 bg-transparent font-normal text-foreground-600 shadow-sm',
       },
-      size: {
+      kind: {
```
LetItRock, line 17: renamed

apps/dashboard/src/components/primitives/badge.tsx
```diff
@@ -14,24 +14,29 @@ const badgeVariants = cva(
...
         default: 'rounded-md px-2 py-1',
         pill: 'rounded-full px-2',
         'pill-stroke': 'rounded-full px-2',
         tag: 'rounded-md py-0.5 px-2',
       },
+      size: {
```
LetItRock, line 23: added the new size prop

apps/dashboard/src/components/primitives/badge.tsx
scopsy: What's the default between variants and kinds?

scopsy (approved):

apps/dashboard/src/components/primitives/badge.tsx
SokratisVidros: ☝️ Same question by me. The naming feels a bit overlapping.

SokratisVidros (approved):

apps/dashboard/src/components/primitives/badge.tsx
desiprisg: size is better here as it affects paddings. Seems like a design flaw to have this many combinations. Maybe we should add more sizes instead of a new variant.

LetItRock: but it's not about the "size" only, but the "shape" and paddings

LetItRock: maybe we should move the paddings to the size

PR #7201, opened 2024-12-03
fix(dashboard): step form values were not updated when switching between steps

LetItRock:

When the step form values were updated, and the user selected the other step and then the first one again, the UI did not reflect the latest step changes for the first step. Not anymore.

### Screenshots

https://github.com/user-attachments/assets/ad5270bf-bed8-4b27-801f-e01bd4f042f2

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/nodes.tsx
+++ b/apps/dashboard/src/components/workflow-editor/nodes.tsx
@@ -12,6 +12,7 @@ import { STEP_TYPE_TO_COLOR } from '@/utils/color';
 import { useWorkflow } from '@/components/workflow-editor/workflow-provider';
 import { WorkflowOriginEnum } from '@novu/shared';
 import { createStep } from '@/components/workflow-editor/steps/step-provider';
+import { getEncodedId, STEP_DIVIDER } from '@/utils/step';
 
 export type NodeData = {
   name?: string;
@@ -52,7 +53,11 @@ const StepNode = (props: StepNodeProps) => {
     stepSlug: string;
   }>();
 
-  return <Node aria-selected={stepSlug === data.stepSlug} className={cn('group', className)} {...rest} />;
+  const isSelected =
+    getEncodedId({ slug: stepSlug ?? '', divider: STEP_DIVIDER }) ===
+    getEncodedId({ slug: data.stepSlug ?? '', divider: STEP_DIVIDER });
+
+  return <Node aria-selected={isSelected} className={cn('group', className)} {...rest} />;
 };
 
 export const EmailNode = ({ data }: NodeProps<NodeType>) => {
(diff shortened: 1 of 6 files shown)


apps/dashboard/src/components/workflow-editor/nodes.tsx
```diff
@@ -52,7 +53,11 @@ const StepNode = (props: StepNodeProps) => {
     stepSlug: string;
   }>();
 
-  return <Node aria-selected={stepSlug === data.stepSlug} className={cn('group', className)} {...rest} />;
+  const isSelected =
+    getEncodedId({ slug: stepSlug ?? '', divider: STEP_DIVIDER }) ===
+    getEncodedId({ slug: data.stepSlug ?? '', divider: STEP_DIVIDER });
+
+  return <Node aria-selected={isSelected} className={cn('group', className)} {...rest} />;
```
LetItRock, line 60: fixes the issue with disappearing node selection when step name changes

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -64,10 +65,6 @@ export const ConfigureStepForm = (props: ConfigureStepFormProps) => {
 
   const form = useForm<z.infer<typeof stepSchema>>({
     defaultValues,
-    values: {
-      ...defaultValues,
-      ...step.controls.values,
-    },
```
LetItRock, line 70: no need to use `values` here, the form should always be set with default values that should have the latest step changes

apps/dashboard/src/components/workflow-editor/steps/configure-step-form.tsx
```diff
@@ -78,6 +75,7 @@ export const ConfigureStepForm = (props: ConfigureStepFormProps) => {
     isReadOnly: isCodeCreatedWorkflow,
     save: (data) => {
       update(updateStepInWorkflow(workflow, data));
+      updateStepCache(data);
```
LetItRock, line 78: because we use two separate endpoints: to update workflow and to fetch the step data, we need to update the step that is in the query cache on the `useFetchStep` hook

apps/dashboard/src/hooks/use-fetch-step.tsx
```diff
@@ -1,29 +1,43 @@
...
-
-  const { data, isPending, isRefetching, error, refetch } = useQuery<StepDataDto>({
-    queryKey: [
+  const queryKey = useMemo(
+    () => [
       QueryKeys.fetchWorkflow,
       currentEnvironment?._id,
-      workflowSlug,
-      getEncodedId({ slug: stepSlug!, divider: STEP_DIVIDER }),
+      getEncodedId({ slug: workflowSlug, divider: WORKFLOW_DIVIDER }),
```
LetItRock, line 17: we should not use `workflowSlug` or `stepSlug` as the query keys but their ids

ChmaraX (approved):

PR #7203, opened 2024-12-03
feat(dashboard): Billing settings page in dashboard v2

scopsy:

CleanShot 2024-12-03 at 16 22 50@2x

<details>

</details>


apps/dashboard/src/components/billing/hooks/use-subscription.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const daysLeft = useMemo(() => {
+    const today = new Date();
+    if (!data?.data.trial.end) return 0;
+
+    return isSameDay(new Date(data.data.trial.end), today) ? 0 : differenceInDays(new Date(data.data.trial.end), today);
+  }, [data?.data.trial.end]);
+
+  if (isLoading || !data) {
+    return {
```
SokratisVidros: It's a good opportunity to refactor this hook a little bit and make it work with a subscription object from the database or null. Null in data means there is no subscription yet, and the page needs to load. 

This default state creates a duplication of code between the frontend and the API that we can remove.

apps/dashboard/src/components/billing/plan-action-button.tsx
```diff
@@ -0,0 +1,67 @@
...
+  selectedBillingInterval,
+  variant = 'default',
+  showIcon = false,
+  className,
+  size = 'default',
+}: PlanActionButtonProps) {
+  const { data: subscription } = useSubscription();
+  const { trial, apiServiceLevel } = subscription || {};
+  const isPaidSubscriptionActive =
+    subscription?.isActive && !trial?.isActive && apiServiceLevel !== ApiServiceLevelEnum.FREE;
```
SokratisVidros: If the subscription is null until it loads, we can easily return null at the top of the component (after the hooks) and avoid all the `?.` train wrecks.

apps/dashboard/src/components/billing/subscription-provider.tsx
```diff
@@ -0,0 +1,35 @@
+import { createContext, useContext } from 'react';
+import { ApiServiceLevelEnum } from '@novu/shared';
+import { useSubscription, type UseSubscriptionType } from './hooks/use-subscription';
+
+const SubscriptionContext = createContext<UseSubscriptionType>({
+  isLoading: false,
```
SokratisVidros: Ditto about the default state. Let's go with `null`

apps/dashboard/src/components/billing/utils/hubspot.constants.ts
```diff
@@ -0,0 +1,3 @@
+export const HUBSPOT_FORM_IDS = {
+  UPGRADE_CONTACT_SALES: '297d0b87-4c09-4ccf-a74c-28fd5e3b37f6',
```
SokratisVidros, line 2: We need a VITE environment variable on this to prevent Hubspot from working locally or on staging.

SokratisVidros (approved):

apps/dashboard/src/components/billing/hooks/use-subscription.ts
scopsy: I agree, first wanted to make it work 😄 Going to refactor a few of those older pieces 💪

apps/dashboard/src/components/billing/plan-action-button.tsx
scopsy: I will keep this, but will refactor slightly. I want to use the Button loading indication. I think it's looking better.

apps/dashboard/src/components/billing/utils/hubspot.constants.ts
scopsy: This is actually not the hubspot id itself, just an id of the form we use to submit. I would keep it as is, as we have a vite env for hubspot already.

apps/api/src/app/testing/billing/get-portal-link.e2e-ee.ts
```diff
@@ -35,7 +35,7 @@ describe('Get portal link', async () => {
     });
 
     expect(stub.lastCall.args.at(0)).to.deep.equal({
-      return_url: `${process.env.FRONT_BASE_URL}/settings/billing`,
+      return_url: `${process.env.FRONT_BASE_URL}/manage-account/billing`,
```
scopsy, line 38: The path was previously wrong

apps/dashboard/src/pages/settings.tsx
```diff
@@ -52,6 +56,10 @@ const clerkComponentAppearance: Appearance = {
 export function SettingsPage() {
   const navigate = useNavigate();
   const location = useLocation();
+  const isV2BillingEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V2_DASHBOARD_BILLING_ENABLED);
+
+  const TAB_TRIGGER_CLASSNAME =
```
scopsy, line 60: Extracted to be reused across tabs

apps/dashboard/src/components/side-navigation/free-trial-card.tsx
```diff
@@ -3,70 +3,95 @@ import { RiArrowRightDoubleLine, RiInformationFill } from 'react-icons/ri';
...
-      </div>
-      <div
-        className={`-mt-2 max-h-0 overflow-hidden opacity-0 ${transition} group-hover:max-h-8 group-hover:opacity-100`}
-      >
-        <Button className={`w-full translate-y-full ${transition} group-hover:translate-y-0`} variant="light" size="sm">
-          Upgrade now
-        </Button>
-      </div>
+    <a href={LEGACY_ROUTES.BILLING} className={cardClassName}>
+      <CardContent pluralizedDays={pluralizedDays} daysTotal={daysTotal} daysLeft={daysLeft} />
```
scopsy, line 94: Functionality left-as is, just wrapped it to provide backwards compatibility

apps/dashboard/src/components/billing/hooks/use-subscription.ts
```diff
@@ -0,0 +1,62 @@
...
+    daysLeft: number;
+  };
+}
+
+export interface UseSubscriptionType {
+  isLoading: boolean;
+  data: ISubscriptionData | null;
+}
+
+export function useSubscription(): UseSubscriptionType {
```
SokratisVidros: Let's reuse the `@hooks/use-subscription.ts`.

apps/dashboard/src/components/billing/hooks/use-subscription.ts
```diff
@@ -0,0 +1,62 @@
+import { useQuery } from '@tanstack/react-query';
+import { ApiServiceLevelEnum } from '@novu/shared';
+import { get } from '../../../api/api.client';
+import { differenceInDays, isSameDay } from 'date-fns';
+import { useMemo } from 'react';
+
+export interface ISubscriptionData {
```
SokratisVidros: Please move this to `@novu/shared`. There should be the same interface somewhere in the API.

apps/dashboard/src/components/billing/plan-action-button.tsx
```diff
@@ -0,0 +1,94 @@
...
+  size = 'default',
+}: PlanActionButtonProps) {
+  const { data, isLoading } = useSubscription();
+  const track = useTelemetry();
+
+  const isPaidSubscriptionActive = () => {
+    return data?.isActive && !data?.trial?.isActive && data?.apiServiceLevel !== ApiServiceLevelEnum.FREE;
+  };
+
+  const { mutateAsync: checkout, isLoading: isCheckingOut } = useMutation(
```
SokratisVidros: Nit-picky one, consider introducing a use-checkout hook so that this is reusuable.

apps/dashboard/src/components/billing/plan-action-button.tsx
```diff
@@ -0,0 +1,94 @@
...
+        track(TelemetryEvent.BILLING_UPGRADE_ERROR, {
+          error: error.message,
+          billingInterval: selectedBillingInterval,
+        });
+        toast.error(error.message || 'Unexpected error');
+      },
+    }
+  );
+
+  const { mutateAsync: goToPortal, isLoading: isGoingToPortal } = useMutation(() => get('/v1/billing/portal'), {
```
SokratisVidros: Ditto about hooks

apps/dashboard/src/components/side-navigation/free-trial-card.tsx
```diff
@@ -3,70 +3,95 @@ import { RiArrowRightDoubleLine, RiInformationFill } from 'react-icons/ri';
...
-        <Progress value={daysTotal - daysLeft} max={daysTotal} />
-      </div>
-      <div
-        className={`-mt-2 max-h-0 overflow-hidden opacity-0 ${transition} group-hover:max-h-8 group-hover:opacity-100`}
-      >
-        <Button className={`w-full translate-y-full ${transition} group-hover:translate-y-0`} variant="light" size="sm">
-          Upgrade now
-        </Button>
-      </div>
+    <a href={LEGACY_ROUTES.BILLING} className={cardClassName}>
```
SokratisVidros, line 93: Shouldn't we go to the new billing if the feature flag is on?

apps/dashboard/src/components/billing/active-plan-banner.tsx
```diff
@@ -0,0 +1,130 @@
...
+                      </span>{' '}
+                      <span className="text-muted-foreground">
+                        of {subscription.events.included.toLocaleString()} events
+                      </span>
+                    </div>
+                    <span className="text-muted-foreground text-xs">Updates hourly</span>
+                  </>
+                )}
+              </div>
+              {!subscription ? (
```
SokratisVidros: Save a character! 😆 

```suggestion
              {subscription ? ... : ...
```

SokratisVidros (changes requested):

apps/dashboard/src/components/side-navigation/free-trial-card.tsx
scopsy: We do, (that's the current logic). Or maybe I'm missing something? 🤔

apps/dashboard/src/components/billing/hooks/use-subscription.ts
scopsy: Good catch! Didn't noticed we have one

apps/dashboard/src/components/side-navigation/free-trial-card.tsx
SokratisVidros: I am confused by the href to LEGACY_ROUTES.BILLING in the free-trial-card of the new-dashboard.

SokratisVidros (approved):
💵

apps/dashboard/src/components/side-navigation/free-trial-card.tsx
scopsy: If the feature flag is OFF, we want to navigate to the legacy billing page, (I want to properly tests this in prod and staging before enabling this page)

PR #7208, opened 2024-12-04
feat(api): Nv 4939 e2e testing happy path events

tatarco:

1. testing trigger  / broadcast with SDK
2. Changes required to make it work in the actual repo

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/events/dtos/trigger-event-response.dto.ts
+++ b/apps/api/src/app/events/dtos/trigger-event-response.dto.ts
@@ -1,30 +1,38 @@
-import { IsBoolean, IsDefined, IsString } from 'class-validator';
+import { IsBoolean, IsDefined, IsEnum, IsOptional, IsString } from 'class-validator';
 import { TriggerEventStatusEnum } from '@novu/shared';
 import { ApiProperty } from '@nestjs/swagger';
 
 export class TriggerEventResponseDto {
   @ApiProperty({
-    description: 'If trigger was acknowledged or not',
+    description: 'Indicates whether the trigger was acknowledged or not',
+    type: Boolean,
   })
   @IsBoolean()
   @IsDefined()
   acknowledged: boolean;
 
   @ApiProperty({
-    description: 'Status for trigger',
+    description: 'Status of the trigger',
     enum: TriggerEventStatusEnum,
   })
   @IsDefined()
+  @IsEnum(TriggerEventStatusEnum)
   status: TriggerEventStatusEnum;
 
   @ApiProperty({
-    description: 'In case of an error, this field will contain the error message',
+    description: 'In case of an error, this field will contain the error message(s)',
+    type: [String], // Specify that this is an array of strings
+    required: false, // Not required since it's optional
   })
+  @IsOptional()
   error?: string[];
 
   @ApiProperty({
-    description: 'Transaction id for trigger',
+    description: 'Transaction ID for the trigger',
+    type: String, // Specify that this is a string
+    required: false, // Not required since it's optional
   })
+  @IsOptional()
   @IsString()
   transactionId?: string;
 }
--- a/apps/api/src/app/events/dtos/trigger-event-response.dto.ts
+++ b/apps/api/src/app/events/dtos/trigger-event-response.dto.ts
@@ -28,7 +28,7 @@ export class TriggerEventResponseDto {
   error?: string[];
 
   @ApiProperty({
-    description: 'Transaction ID for the trigger',
+    description: 'The returned transaction ID of the trigger',
     type: String, // Specify that this is a string
     required: false, // Not required since it's optional
   })
--- a/apps/api/src/app/events/dtos/trigger-event-to-all-request.dto.ts
+++ b/apps/api/src/app/events/dtos/trigger-event-to-all-request.dto.ts
@@ -22,9 +22,9 @@ export class TriggerEventToAllRequestDto {
       },
     },
     type: 'object',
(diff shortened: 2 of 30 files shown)


apps/api/src/app/events/dtos/trigger-event-response.dto.ts
```diff
@@ -1,30 +1,38 @@
...
+    description: 'In case of an error, this field will contain the error message(s)',
+    type: [String], // Specify that this is an array of strings
+    required: false, // Not required since it's optional
   })
+  @IsOptional()
   error?: string[];
 
   @ApiProperty({
-    description: 'Transaction id for trigger',
+    description: 'Transaction ID for the trigger',
```
SokratisVidros:
```suggestion
    description: 'The returned transaction ID of the trigger',
```

apps/api/src/app/events/dtos/trigger-event-to-all-request.dto.ts
```diff
@@ -15,14 +15,18 @@ export class TriggerEventToAllRequestDto {
...
-    description: `The payload object is used to pass additional custom information that could be used to render the template, or perform routing rules based on it. 
-      This data will also be available when fetching the notifications feed from the API to display certain parts of the UI.`,
     example: {
       comment_id: 'string',
       post: {
         text: 'string',
       },
     },
+    type: 'object',
+    description: `The payload object is used to pass additional custom information that
```
SokratisVidros:
```suggestion
    description: `The payload object is used to pass additional information that 
```

apps/api/src/app/events/dtos/trigger-event-to-all-request.dto.ts
```diff
@@ -15,14 +15,18 @@ export class TriggerEventToAllRequestDto {
...
     example: {
       comment_id: 'string',
       post: {
         text: 'string',
       },
     },
+    type: 'object',
+    description: `The payload object is used to pass additional custom information that 
+    could be used to render the template, or perform routing rules based on it. 
+      This data will also be available when fetching the notifications feed from the API to display certain parts of the UI.`,
```
SokratisVidros:
```suggestion
For In-App channel, payload data are also available in `<Inbox />`.
```

apps/api/src/app/events/e2e/cancel-event.e2e.ts
```diff
@@ -18,25 +20,24 @@ describe('Cancel event - /v1/events/trigger/:transactionId (DELETE)', function (
   let subscriber: SubscriberEntity;
   let subscriberService: SubscribersService;
   const jobRepository = new JobRepository();
-
+  let novuClient: Novu;
+
+  async function cancelEvent(transactionId: string | undefined) {
+    if (!transactionId) {
+      throw new Error('id missing to cancel');
```
SokratisVidros:
```suggestion
      throw new Error('Missing transactionId');
```

apps/api/src/app/events/events.controller.ts
```diff
@@ -132,6 +137,10 @@ export class EventsController {
     description: `Trigger a broadcast event to all existing subscribers, could be used to send announcements, etc.
       In the future could be used to trigger events to a subset of subscribers based on defined filters.`,
   })
+  @ApiCreatedResponse({
+    description: 'The trigger event has been successfully created.',
```
SokratisVidros: Please rephrase this description to something more user-centric.

packages/shared/src/dto/stateless-control-values/stateless-workflow-to-step-control-values.ts
```diff
@@ -0,0 +1,10 @@
+export class StatelessWorkflowToStepControlValues {
```
SokratisVidros: How about making it shorter?
```suggestion
export class StatelessControls {
```

We have to stick to a single name for control, control values, and control schemas in the codebase. My suggestion is that these are controls, and they have values and a schema. So naming should reflect that.

libs/dal/src/repositories/notification-template/notification-template.entity.ts
```diff
@@ -142,11 +141,11 @@ export class StepVariantEntity implements IStepVariant {
...
   bridgeUrl?: string;
-  /**
-   * @deprecated This property is deprecated and will be removed in future versions.
-   * Use `fullName` instead.
+  /*
+   * controlVariables exists
+   * only on none production environment in order to provide stateless control variables on fly
    */
-  controlVariables?: ControlsDto;
+  controlVariables?: Record<string, unknown>;
```
SokratisVidros, line 148: Is this property deprecated today or not yet?

SokratisVidros (changes requested):

packages/shared/src/dto/stateless-control-values/stateless-workflow-to-step-control-values.ts
tatarco: but those are not controls that include both, it's a weird hybrid made for stateless triggering, it's a higher level map transmitted in a trigger including a stepid --> controlId --> value nested map, not used anywhere else, I will rename, but the rest of the assumption is not correct, just a weird naming confusing all over

libs/dal/src/repositories/notification-template/notification-template.entity.ts
tatarco: it was a misuse that worked just because there is no real types in javascript, this was not supposed to be the type here. verified in details pairing with @djabarovgeorge

SokratisVidros (approved):

PR #7210, opened 2024-12-04
fix(dashboard): minor onboarding updates

scopsy:

- Precache usecase images
- Users who create account from v2, should be marked as opt in
- Fix an issue with workflow race condition on useEffect

### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/auth/inbox-playground.tsx
+++ b/apps/dashboard/src/components/auth/inbox-playground.tsx
@@ -102,21 +102,21 @@ export function InboxPlayground() {
   }, [telemetry]);
 
   useEffect(() => {
-    if (!data || isLoading) return;
+    if (!data) return;
 
     /**
      * We only want to create the demo workflow if it doesn't exist yet.
      * This workflow will be used by the inbox preview examples
      */
     const initializeDemoWorkflow = async () => {
-      const workflow = data?.workflows.find((workflow) => workflow.workflowId === ONBOARDING_DEMO_WORKFLOW_ID);
+      const workflow = data?.workflows.find((workflow) => workflow.workflowId?.includes(ONBOARDING_DEMO_WORKFLOW_ID));
       if (!workflow) {
         await createDemoWorkflow();
       }
     };
 
     initializeDemoWorkflow();
-  }, [data, isLoading]);
+  }, [data]);
 
   const handleSendNotification = async () => {
     try {
--- a/apps/dashboard/src/components/auth/inbox-playground.tsx
+++ b/apps/dashboard/src/components/auth/inbox-playground.tsx
@@ -91,7 +91,7 @@ export function InboxPlayground() {
   });
 
   const { triggerWorkflow, isPending } = useTriggerWorkflow();
-  const { data } = useWorkflows({ query: ONBOARDING_DEMO_WORKFLOW_ID });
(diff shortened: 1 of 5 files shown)


apps/dashboard/src/components/auth/inbox-playground.tsx
```diff
@@ -102,7 +102,7 @@ export function InboxPlayground() {
   }, [telemetry]);
 
   useEffect(() => {
-    if (!data) return;
+    if (!data || isLoading) return;
```
SokratisVidros: Isn't the lack of data the same `isLoading` in the react-query kingdom?

apps/dashboard/src/pages/usecase-select-page.tsx
```diff
@@ -41,6 +41,14 @@ export function UsecaseSelectPage() {
   const [selectedUseCases, setSelectedUseCases] = useState<ChannelTypeEnum[]>([]);
   const [hoveredUseCase, setHoveredUseCase] = useState<ChannelTypeEnum | null>(null);
 
+  useEffect(() => {
```
SokratisVidros: We can leverage browser [resource hints prefetching](https://web.dev/learn/performance/resource-hints) and let then decide when is the best time to prefetch them:

```import React from "react";
import { Helmet } from "react-helmet-async";

const ImagePrefetcher = () => {
  return (
    <Helmet>
      <link rel="prefetch" href="/path/to/image1.jpg" as="image" />
      <link rel="prefetch" href="/path/to/image2.jpg" as="image" />
    </Helmet>
  );
};
```

SokratisVidros (changes requested):

apps/dashboard/src/components/auth/inbox-playground.tsx
scopsy: Yep you are correct, my issue was actually from something else. Investigating

apps/dashboard/src/components/auth/inbox-playground.tsx
```diff
@@ -91,7 +91,7 @@ export function InboxPlayground() {
   });
 
   const { triggerWorkflow, isPending } = useTriggerWorkflow();
-  const { data } = useWorkflows({ query: ONBOARDING_DEMO_WORKFLOW_ID });
+  const { data, isLoading } = useWorkflows({ query: ONBOARDING_DEMO_WORKFLOW_ID });
```
LetItRock: `isLoading` not used?

apps/dashboard/src/pages/usecase-select-page.tsx
```diff
@@ -101,6 +101,11 @@ export function UsecaseSelectPage() {
 
   return (
     <>
+      <Helmet>
+        {channelOptions.map((option) => (
+          <link key={option.id} rel="prefetch" href={`/images/auth/${option.image}`} as="image" />
```
LetItRock, line 106: is this going to help? because it will run when the JS is executed by the browser, so like just a ms before the dom is rendered

SokratisVidros: It should, as the page will render, the user still needs to interact with it. I will check it in the preview URL.

SokratisVidros (approved):

PR #7211, opened 2024-12-04
feat(novu): Add `--studio-host` option on dev server

4rthem:

Added an option to start the studio server on a different host (than `localhost`).
I'm running the studio server in a docker compose project and need to bind it to the docker network with `--studio-host 0.0.0.0`

### DIFF
--- a/packages/novu/README.MD
+++ b/packages/novu/README.MD
@@ -34,15 +34,16 @@ npx novu@latest dev
 
 ## 🔥 Flags
 
-| flag | long form usage example | description                   | default value               |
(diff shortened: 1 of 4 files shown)


scopsy (approved):
Thank you for the contribution! Curious to learn more about your usecase in running the studio inside of Docker?

4rthem:
I don't have NodeJS on my machine, I only use Docker for every programming languages. Also my dev stacks are under docker compose with a Traefik reverse proxy using DNS instead of many many open ports :)

PR #7213, opened 2024-12-04
fix(dashboard): editor fields lagging

LetItRock:

Fix for the Editor (CodeMirror) fields lagging when typing fast.

### Screenshots

**Before:**

https://github.com/user-attachments/assets/54b54e65-d023-4303-883c-c3577534f951

**After:**

https://github.com/user-attachments/assets/df764701-bdc1-4eae-80ea-e2667013a396


apps/dashboard/src/components/primitives/editor.tsx
```diff
@@ -159,27 +169,27 @@ export const Editor = React.forwardRef<{ focus: () => void; blur: () => void },
...
+    const onChangeCallback = useCallback(
+      (value: string) => {
+        // when typing fast the onChange event is called multiple times during one render phase
+        // by default react batches state updates and only triggers one render phase
+        // which results in value not being updated and "jumping" effect in the editor
+        // to prevent this we need to flush the state updates synchronously
+        flushSync(() => {
+          onChange?.(value);
+        });
+      },
```
LetItRock, line 182: The "lagging" in the editor was caused by two problems:
- RHF `useWatch` causing re-render for many form components
- React batching state updates

This code solves the second.

apps/dashboard/src/components/primitives/editor.tsx
```diff
@@ -159,27 +169,27 @@ export const Editor = React.forwardRef<{ focus: () => void; blur: () => void },
       [fontFamily]
     );
 
-    const debouncedOnChange = useMemo(
-      () =>
-        debounce((value: string) => {
-          if (onChange) onChange(value);
-        }, 50),
```
LetItRock, line 167: No need to debounce because it affects the input (editor) responsiveness and if typing quickly the changes are dropped and resulting in text "jumps". 
React is quick in doing updates, so we had to solve the root cause of the issue and not "hack" it.

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -18,9 +18,6 @@ export function useFormAutosave<U extends Record<string, unknown>, T extends Fie
   save: (data: U) => void;
 }) {
   const formRef = useDataRef(propsForm);
-  const watchedData = useWatch<T>({
-    control: propsForm.control,
-  });
```
LetItRock, line 23: the first mentioned reason for "lagging" in the editor, `useWatch` is causing re-renders when the form state is updated

apps/dashboard/src/hooks/use-form-autosave.ts
```diff
@@ -79,11 +76,16 @@ export function useFormAutosave<U extends Record<string, unknown>, T extends Fie
...
-    debouncedOnSave(values);
-  }, [watchedData, debouncedOnSave, formRef]);
+    const form = formRef.current;
+
+    const { unsubscribe } = form.watch((partial) => {
+      const values = form.getValues();
+      debouncedOnSave({ ...values, ...partial });
+    });
+
+    return () => unsubscribe();
```
LetItRock, line 87: this is how to watch instead

desiprisg (approved):

PR #7223, opened 2024-12-05
fix(dashboard): Always trust the URL for the environment selection

SokratisVidros:
    
- [x] Pick an environment based on the URL and avoid race conditions using the local storage as we did in the old web app
- [x] Allow multiple environments to open in separate browser tabs
- [x] Unify conventions around API-related hooks. That is:

1. Name all of them similarly using the use-verb-resource pattern.
2. Unify their return types and add stricter types to hook arguments

Screenshot 2024-12-05 at 23 50 31

### What this PR doesn't contain?
-[] Unify the return types to always follow the { resourceName: data, ...restOfReactQuery } return type. This will be done separately in the future.


apps/dashboard/src/api/api.client.ts
```diff
@@ -17,28 +17,28 @@ type HttpMethod = 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH';
 const request = async <T>(
   endpoint: string,
   options?: {
-    data?: unknown;
+    environment?: IEnvironment;
```
SokratisVidros, line 20: Inject the environment to API calls

apps/dashboard/src/api/billing.ts
```diff
@@ -1,8 +1,7 @@
-import type { GetSubscriptionDto } from '@novu/shared';
+import type { IEnvironment, GetSubscriptionDto } from '@novu/shared';
 import { get } from './api.client';
 
-export async function getBillingSubscription() {
-  const { data } = await get<{ data: GetSubscriptionDto }>('/billing/subscription');
-
+export async function getSubscription({ environment }: { environment: IEnvironment }) {
```
SokratisVidros, line 4: Simplify naming

apps/dashboard/src/components/auth/inbox-playground.tsx
```diff
@@ -1,25 +1,26 @@
-import { useEffect, useState } from 'react';
```
SokratisVidros, line 1: Sort alphabetically to have it nice and tidy

apps/dashboard/src/components/auth/questionnaire-form.tsx
```diff
@@ -9,7 +9,7 @@ import { useForm, Controller } from 'react-hook-form';
 import { updateClerkOrgMetadata } from '../../api/organization';
 import { hubspotCookie } from '../../utils/cookies';
 import { identifyUser } from '../../api/telemetry';
-import { useTelemetry } from '../../hooks';
+import { useTelemetry } from '../../hooks/use-telemetry';
```
SokratisVidros, line 12: We should avoid the index.ts barrel file. It destroys tree shaking, chunking and introduces an extra import and file resolution.

apps/dashboard/src/hooks/index.ts
```diff
@@ -1,9 +0,0 @@
-export * from './use-billing-subscription';
```
SokratisVidros, line 1: https://tkdodo.eu/blog/please-stop-using-barrel-files

scopsy (approved):

apps/dashboard/src/components/create-workflow-button.tsx
```diff
@@ -36,8 +36,9 @@ export const CreateWorkflowButton = (props: CreateWorkflowButtonProps) => {
   const navigate = useNavigate();
   const { currentEnvironment } = useEnvironment();
   const [isOpen, setIsOpen] = useState(false);
+  // TODO: Move to a use-create-workflow.ts hook
   const { mutateAsync, isPending } = useMutation({
-    mutationFn: async (data: CreateWorkflowDto) => createWorkflow(data),
+    mutationFn: async (workflow: CreateWorkflowDto) => createWorkflow({ environment: currentEnvironment!, workflow }),
```
SokratisVidros, line 41: In a follow-up PR, `currentEnvironment` will never be undefined or null to avoid the `!` or `?._id` discomfort.

apps/dashboard/src/hooks/use-fetch-workflows.ts
```diff
@@ -10,18 +9,15 @@ interface UseWorkflowsParams {
...
-    queryFn: async () => {
-      const { data } = await getV2<{ data: ListWorkflowResponse }>(
-        `/workflows?limit=${limit}&offset=${offset}&query=${query}`
-      );
-      return data;
-    },
+    queryFn: () => getWorkflows({ environment: currentEnvironment!, limit, offset, query }),
     placeholderData: keepPreviousData,
+    enabled: !!currentEnvironment?._id,
+    refetchOnWindowFocus: true,
```
SokratisVidros, line 20: This caters to the use case where customers will have their dev and prod environments side by side, and they will sync workflows between environments. They should be able to see the update without a full page reload.

LetItRock (approved):
lgtm 🙌

PR #7230, opened 2024-12-06
feat(dashboard): delay amount input field polish

ChmaraX:

Before:

https://github.com/user-attachments/assets/c2125c5c-1c25-4d9d-bcfe-4570805600cc

After:

https://github.com/user-attachments/assets/daa35ba0-2554-43ba-a587-596e0c78eb3f

### DIFF
--- a/apps/dashboard/src/components/amount-input.tsx
+++ b/apps/dashboard/src/components/amount-input.tsx
@@ -1,0 +1,120 @@
+import { cn } from '@/utils/ui';
+import { FormControl, FormField, FormItem, FormMessagePure } from '@/components/primitives/form/form';
+import { Input } from '@/components/primitives/input';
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/components/amount-input.tsx
desiprisg, line 1: can we name this something more intuitive, like `duration-input` ?

ChmaraX: @desiprisg Its not necessarily a duration, its number + unit, e.g. `10 horses`, `15 minutes`, `20 meters`

 the most intuitive was `number-input-with-select` :D but we changed that to `amount-input`

SokratisVidros (approved):
@ChmaraX Nice touch. Can you please also align the error icon to the top of the message, not the center? I feel it will look much better especially for the long message of the 90 days.

PR #7235, opened 2024-12-07
feat(api): add full step data to workflow dto; refactor

ChmaraX:
### TL;DR:
- The external workflow DTO now includes full step data (with control values etc). This supports multiple upcoming features and simplifies both the frontend and backend **significantly**.
- A unified external step DTO replaces previously inconsistent step DTOs returned by `GET step` and `GET workflow` endpoints

💡  These changes do not affect FE-BE compatibility in any way - only new fields were added to `workflow.steps`, no external DTO property was removed or renamed.

### Key DTO changes:

**File renames and moves:**

- `step-data.dto` ➔ `step.dto.ts` (better reflects contents and avoids circular imports).
- All step-related DTOs consolidated in `step.dto.ts`.
- `workflow.dto.ts` ➔ `workflow.deprecated.dto` (contains only deprecated DTOs).
- `workflow-response.dto` ➔ `workflow.dto.ts`.
- All workflow-related DTOs consolidated in `workflow.dto.ts`.

**Simplified Step DTO:**

- Replaced `StepResponseDto` with `StepDataDto` (already used by GET step).
- This unification reduces redundancy and simplifies the codebase.

 
### Refactoring:
- `sync-to-environment.usecase` - improved naming, less code, clearer intent
- Removed redundant workflow e2e code (full step data is now available without extra fetches).

### Next steps:
1. Update the frontend to use step data from the `GET workflow` endpoint.
  1.1 this will enable us to get rid of react-query cache invalidation code, step hooks etc
2. Remove unused API `GET step` endpoint, (I think we should also remove `PATCH step` + usecases) since we are currently not using them - if we decide to use them again in future we can find them in git history
3. Simplify API e2e tests


apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
```diff
@@ -15,7 +15,10 @@ import { NotificationStepEntity, NotificationTemplateEntity } from '@novu/dal';
 import { WorkflowInternalResponseDto } from '@novu/application-generic';
 import { buildSlug } from '../../shared/helpers/build-slug';
 
-export function toResponseWorkflowDto(workflow: WorkflowInternalResponseDto): WorkflowResponseDto {
+export function toResponseWorkflowDto(
+  workflow: WorkflowInternalResponseDto,
+  fullSteps: StepDataDto[]
```
djabarovgeorge: personal opinion we can keep `steps` name, and the type should explain what we get here.

```suggestion
  steps: StepDataDto[]
```

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
```diff
@@ -30,7 +33,7 @@ export function toResponseWorkflowDto(workflow: WorkflowInternalResponseDto): Wo
     tags: workflow.tags,
     active: workflow.active,
     preferences: preferencesDto,
-    steps: getSteps(workflow),
+    steps: fullSteps,
```
djabarovgeorge: I believe we should keep the mapping here for two key reasons:
1. We’re adding extra parameters, such as `slug`, which are only exposed to the API client.  
2. We’re mapping the types—for instance, if the entity allows `undefined`, we convert those to stricter types in the DTO response.

djabarovgeorge: the first point is not relevant, i see that you added the slug to the BuildStepDataUsecase.

apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
```diff
@@ -21,6 +32,29 @@ export class GetWorkflowUseCase {
...
+  ): Promise<StepDataDto[]> {
+    const stepPromises = workflow.steps.map((step: NotificationStepEntity & { _id: string }) =>
+      this.buildStepDataUsecase.execute(
+        BuildStepDataCommand.create({
+          workflowIdOrInternalId: workflow._id,
+          stepIdOrInternalId: step._id,
+          user,
+        })
+      )
+    );
```
djabarovgeorge: maybe we could throw it here? this way we will have the step id that failed as well.

```suggestion
try{
      this.buildStepDataUsecase.execute(
        BuildStepDataCommand.create({
          workflowIdOrInternalId: workflow._id,
          stepIdOrInternalId: step._id,
          user,
        })
      )
    );
    } catch (error) {
      throw new InternalServerErrorException({ message: 'Failed to build full workflow steps:', error: error.message, workflowId: workflow._id, stepId: step._id );
    }
```

packages/shared/src/dto/workflows/workflow-deprecated.dto.ts
djabarovgeorge, line 1: please add js doc tag `@deprecated`.
/**
 * `@deprecated` ...
 */
 
today those types are used for /v1/workflows?

djabarovgeorge (commented):
looks good! :) 
left a couple of comments for your review.

packages/shared/src/dto/workflows/workflow-deprecated.dto.ts
ChmaraX: Yes, they seem to be used only in legacy create and update workflow usecases which are marked as deprecated.

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
```diff
@@ -15,7 +15,10 @@ import { NotificationStepEntity, NotificationTemplateEntity } from '@novu/dal';
 import { WorkflowInternalResponseDto } from '@novu/application-generic';
 import { buildSlug } from '../../shared/helpers/build-slug';
 
-export function toResponseWorkflowDto(workflow: WorkflowInternalResponseDto): WorkflowResponseDto {
+export function toResponseWorkflowDto(
```
tatarco, line 18: the point was letting this map to dto, if you send the DTO pre cooked I think it's taking the string out of the funciton no? I would expect to get the entity list and maps of the additinoal data you want to bake in

apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
```diff
@@ -21,6 +32,29 @@ export class GetWorkflowUseCase {
...
 
-    return toResponseWorkflowDto(workflowEntity);
+    const fullSteps = await this.getFullWorkflowSteps(workflowEntity, command.user);
+
+    return toResponseWorkflowDto(workflowEntity, fullSteps);
+  }
+
+  private async getFullWorkflowSteps(
+    workflow: WorkflowInternalResponseDto,
+    user: UserSessionData
```
tatarco, line 42: Regarding my first comment, maybe it's time to kill this mapper, and have it map inside the get use case and just return it as the response of upsert

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -61,14 +58,13 @@ export class SyncToEnvironmentUseCase {
   private async buildRequestDto(
     originWorkflow: WorkflowResponseDto,
     preferencesToClone: PreferencesEntity[],
-    command: SyncToEnvironmentCommand,
     targetWorkflow?: WorkflowResponseDto
   ) {
     if (targetWorkflow) {
-      return await this.mapWorkflowToUpdateWorkflowDto(originWorkflow, targetWorkflow, preferencesToClone, command);
+      return this.mapWorkflowToUpdateWorkflowDto(originWorkflow, targetWorkflow, preferencesToClone);
```
tatarco: I really like having await waiting for the result of an async function, I saw you removed them in bulk, is it a technical reason or preference? I like it for readability, it's clear to me those are both promises but since you didn't define the retuen type it's a little obscure

packages/shared/src/dto/workflows/workflow-deprecated.dto.ts
```diff
@@ -0,0 +1,29 @@
+import { IWorkflowStepMetadata } from '../../entities/step';
+import { BuilderFieldType, BuilderGroupValues, FilterParts } from '../../types';
+import { MessageTemplateDto } from '../message-template';
+
+export class StepVariantDto {
+  id?: string;
+  _id?: string;
```
tatarco, line 10: I was hoping to not see this variant ever again, why are we exposing it in all it's glory? is it really nessesery, what we expose is harder to remove in the future, I would like to not have this exposed, it's repetatice and super confusing

tatarco (approved):
It's looking OK, I'm missing some tests and added some comments, lets please resolve / reply to them before merging

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
ChmaraX: I see your point but the stricter types in this case are already coming from the `BuildStepDataUsecase`.

The return type from the `BuildStepDataUsecase` doesn't allow undefined and does explicit checks for missing values, therefore having additional response type for step and mapping, e.g. like this:

```ts
name: step.name || 'Missing step name' 
```

would be redundant . I think we should keep it simple and when the usecase emerges then we can have some mapping.

packages/shared/src/dto/workflows/workflow-deprecated.dto.ts
ChmaraX: It was already exposed, I didn't added this I only renamed the file.

Its used in legacy (v1) workflow endpoints and usecases.

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
ChmaraX: I would say its a preference, having `await` when returning async function inside async function is redundant - you can see that the function is a promise by hovering over that function.

image

Although the are some advantages to use redundant await when debugging. We can enforce and ESLint rule to enforce either way.

tatarco: If it's ok with you I would prefer to leave it there for now, lets add this stuff and have a standards call and decide on those stuff and enforce them by eslint

apps/api/src/app/workflows-v2/usecases/get-workflow/get-workflow.usecase.ts
ChmaraX: Yes could be an option, but I also think its fine for now if internally we use the `WorkflowInternalResponseDto` which is returned from the upsert etc, and use the full step data just for external DTO here. 

We can re-evaluate later, but for now I think this is the simplest and safest option.

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
ChmaraX: It still maps the workflow to DTO as name of the method suggests, only the steps are coming in already in "DTO form".

We simply don't need to do any mapping for the steps for now, its already in viable form, which is good. Each layer of mapping just adds an overhead

packages/shared/src/dto/workflows/workflow-deprecated.dto.ts
tatarco: resolving, it wasn't clear from git

apps/api/src/app/workflows-v2/mappers/notification-template-mapper.ts
```diff
@@ -71,19 +64,6 @@ export function toWorkflowsMinifiedDtos(templates: NotificationTemplateEntity[])
...
-
-  return {
-    _id: persistedStep._templateId,
-    slug: buildSlug(stepName, ShortIsPrefixEnum.STEP, persistedStep._templateId),
-    name: stepName,
-    stepId: persistedStep.stepId || 'Missing Step Id',
-    type: persistedStep.template?.type || StepTypeEnum.EMAIL,
-    issues: persistedStep.issues,
-  } satisfies StepResponseDto;
-}
```
ChmaraX, line 85: I moved this directly to the `build-step-data` use-case, because its **always** used together with the use-case.

If the usecase for unsanitized step data emerges, then we can split it again or create `GetStepDataInternal` usecase. 

Now I don't think its even possible to create step without name or stepId or without type, it seems to be here only because of tests. These things should be validated upon creation and not rely on fallbacks anyway.

PR #7240, opened 2024-12-08
fix(dashboard): nested payload gen

djabarovgeorge:
### Before
image

### After
image

### Before
image

### After
image

## Empty state case
image

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -134,17 +134,19 @@ export class GeneratePreviewUsecase {
...
     if (!examplePayloadSchema || Object.keys(examplePayloadSchema).length === 0) {
       return finalPayload;
     }
 
-    return _.merge(
-      finalPayload as Record<string, unknown>,
-      { payload: examplePayloadSchema },
-      commandVariablesExample || {}
-    );
+    return _.merge(finalPayload as Record<string, unknown>, examplePayloadSchema, commandVariablesExample || {});
```
djabarovgeorge: the change was needed because now createMockObjectFromSchema is a more generic function and not `payload` object based.

apps/api/src/app/workflows-v2/util/utils.ts
djabarovgeorge, line 1: was moved to shared and refactored to be a more generic function and not payload object based.

apps/dashboard/src/components/workflow-editor/schema.ts
djabarovgeorge, line 1: was deleted, less is more?

_now we reuse the createMockObjectFromSchema from shared the same one that BE used._

packages/shared/src/utils/schema/create-mock-object-from-schema.ts
djabarovgeorge, line 1: util from the API.

SokratisVidros (approved):

PR #7243, opened 2024-12-09
fix(dashboard): run queue invocations safely

ChmaraX:

API errors would result in client freezing, due to unhandled error state.

Simulating 0.33 error rate from API:

https://github.com/user-attachments/assets/d7dd5d24-fdfc-416d-9dcd-e5b7d55374f2

### DIFF
--- a/apps/dashboard/src/hooks/use-invocation-queue.ts
+++ b/apps/dashboard/src/hooks/use-invocation-queue.ts
@@ -1,4 +1,5 @@
 import { useCallback, useRef, useState } from 'react';
+import * as Sentry from '@sentry/react';
 
 type CallbackFunction = () => Promise<unknown>;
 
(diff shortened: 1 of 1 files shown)


apps/dashboard/src/hooks/use-invocation-queue.ts
```diff
@@ -58,6 +56,17 @@ export function useInvocationQueue<T extends CallbackFunction = CallbackFunction
...
   );
 
+  const safelyRunInvocation = useCallback(async (invocation: T | undefined) => {
+    if (!invocation) return;
+
+    try {
+      await invocation();
+    } catch (error) {
+      // If the invocation fails, we want to log the error and continue with the next invocation
+      console.error('Error processing queue item:', error);
```
LetItRock: please send the error to sentry

apps/dashboard/src/hooks/use-invocation-queue.ts
```diff
@@ -58,6 +57,18 @@ export function useInvocationQueue<T extends CallbackFunction = CallbackFunction
...
```
LetItRock: I'm not sure if the `console.error` brings any value here.

SokratisVidros: I agree, we can remove it.

SokratisVidros (approved):

LetItRock (approved):

PR #7244, opened 2024-12-09
feat(api): 'Missing' issue missing when the control value has empty string

tatarco:

@desiprisg found that the FE is sending default values  '' causing the issue mechanism to ignore the missing control value, found the problem lies in the order as we do sanitize for it, added a test as well to validate the behavior

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
+++ b/apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
@@ -223,6 +223,20 @@ describe('Workflow Controller E2E API Testing', () => {
           }
         }
       });
+      it('should show control value required when empty', async () => {
+        const { issues, status } = await createWorkflowAndReturnStepIssues(
+          { steps: [{ ...buildInAppStep(), controlValues: { body: '' } }] },
+          0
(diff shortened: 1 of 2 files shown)


apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
```diff
@@ -223,6 +223,20 @@ describe('Workflow Controller E2E API Testing', () => {
           }
         }
       });
+      it('should show control value required when empty', async () => {
```
ChmaraX, line 226: 😄
```suggestion

      it('should show control value required when empty', async () => {
```

tatarco: whats the change? It looks the same @ChmaraX

ChmaraX (approved):

apps/api/src/app/workflows-v2/workflow.controller.e2e.ts
ChmaraX: newline between functions

PR #7245, opened 2024-12-09
fix(dashboard): Fix stale data on test workflow page

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/hooks/use-fetch-workflow-test-data.ts
+++ b/apps/dashboard/src/hooks/use-fetch-workflow-test-data.ts
@@ -3,14 +3,22 @@ import type { WorkflowTestDataResponseDto } from '@novu/shared';
 import { QueryKeys } from '@/utils/query-keys';
 import { getWorkflowTestData } from '@/api/workflows';
 import { useEnvironment } from '@/context/environment/hooks';
+import { getWorkflowIdFromSlug, WORKFLOW_DIVIDER } from '@/utils/step';
 
-export const useFetchWorkflowTestData = ({ workflowSlug }: { workflowSlug?: string }) => {
+export const useFetchWorkflowTestData = ({ workflowSlug }: { workflowSlug: string }) => {
   const { currentEnvironment } = useEnvironment();
   const { data, isPending, error } = useQuery<WorkflowTestDataResponseDto>({
-    queryKey: [QueryKeys.fetchWorkflowTestData, currentEnvironment?._id, workflowSlug],
+    queryKey: [
+      QueryKeys.fetchWorkflowTestData,
+      currentEnvironment?._id,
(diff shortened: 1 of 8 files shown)


apps/dashboard/src/hooks/use-fetch-workflow-test-data.ts
```diff
@@ -3,11 +3,16 @@ import type { WorkflowTestDataResponseDto } from '@novu/shared';
...
 
-export const useFetchWorkflowTestData = ({ workflowSlug }: { workflowSlug?: string }) => {
+export const useFetchWorkflowTestData = ({ workflowSlug }: { workflowSlug: string }) => {
   const { currentEnvironment } = useEnvironment();
   const { data, isPending, error } = useQuery<WorkflowTestDataResponseDto>({
-    queryKey: [QueryKeys.fetchWorkflowTestData, currentEnvironment?._id, workflowSlug],
+    queryKey: [
+      QueryKeys.fetchWorkflowTestData,
+      currentEnvironment?._id,
+      getEncodedId({ slug: workflowSlug, divider: WORKFLOW_DIVIDER }),
```
SokratisVidros: Nit-picky one: 

```suggestion
      getWorkflowInternalId(workflowSlug),
```

apps/dashboard/src/hooks/use-fetch-workflow-test-data.ts
```diff
@@ -3,11 +3,16 @@ import type { WorkflowTestDataResponseDto } from '@novu/shared';
...
 import { getWorkflowTestData } from '@/api/workflows';
 import { useEnvironment } from '@/context/environment/hooks';
+import { getEncodedId, WORKFLOW_DIVIDER } from '@/utils/step';
 
-export const useFetchWorkflowTestData = ({ workflowSlug }: { workflowSlug?: string }) => {
+export const useFetchWorkflowTestData = ({ workflowSlug }: { workflowSlug: string }) => {
   const { currentEnvironment } = useEnvironment();
   const { data, isPending, error } = useQuery<WorkflowTestDataResponseDto>({
-    queryKey: [QueryKeys.fetchWorkflowTestData, currentEnvironment?._id, workflowSlug],
+    queryKey: [
```
SokratisVidros, line 11: @desiprisg I think the test data page is better if we don't cache it at all. It's a separate tab, so I don't see the value of caching. We'd rather fetch data every time we visit the tab instead of caching and invalidating.

SokratisVidros (approved):
I approved it but please see my comment about caching.

apps/dashboard/src/hooks/use-fetch-workflow-test-data.ts
desiprisg: Sure we can set `gcTime` to 0 to achieve this. `queryKey` remains required though, in order to trigger a refetch if an invalidation were to happen while the query is still used.

PR #7249, opened 2024-12-09
feat(dashboard): Activity Feed Page - Stacked PR

scopsy:
This PR utilized the existing API without any extra changes there.

- New activity feed page

Not in scope: Filters, Workflow Editor connectivity

### Screenshots

CleanShot 2024-12-09 at 14 31 05@2x

<details>

</details>


apps/dashboard/package.json
```diff
@@ -37,6 +37,7 @@
     "@radix-ui/react-collapsible": "^1.1.1",
     "@radix-ui/react-dialog": "^1.1.2",
     "@radix-ui/react-dropdown-menu": "^2.1.1",
+    "@radix-ui/react-hover-card": "^1.1.2",
```
scopsy, line 40: Added new shadcn hover component for time view component

apps/dashboard/src/api/activity.ts
```diff
@@ -0,0 +1,51 @@
+import { IActivity, IEnvironment } from '@novu/shared';
+import { get } from './api.client';
+
+export interface IActivityFilters {
+  channels?: string[];
+  templates?: string[];
+  email?: string;
+  subscriberId?: string;
+  transactionId?: string;
+}
```
scopsy: Filters functonality will be added in a later PR

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  const handleRowMouseEnter = (activity: IActivity) => {
+    hoverTimerRef.current = setTimeout(() => {
+      queryClient.prefetchQuery({
+        queryKey: [QueryKeys.fetchActivity, currentEnvironment?._id, activity._id],
+        queryFn: () => getNotification(activity._id, currentEnvironment!),
+      });
+    }, 1000);
+  };
```
scopsy: I'm not sure if it's needed, but the goal is to preload the hovered pages to eliminate loading times

apps/dashboard/src/components/activity/components/pagination.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function Pagination({ offset, limit, hasMore, onOffsetChange }: PaginationProps) {
+  return (
+    <div className="bottom-0 mt-auto border-t border-t-neutral-200 bg-white py-3">
+      <div className="flex items-center justify-end px-6">
+        <div className="border-input inline-flex items-center rounded-lg border bg-transparent">
+          <Button
+            variant="ghost"
+            size="icon"
```
scopsy: Since we don't have page numbers, we only use carets for pagination

apps/dashboard/src/components/primitives/table.tsx
```diff
@@ -3,23 +3,46 @@ import * as React from 'react';
...
+
+const LoadingRow = () => (
+  <TableRow>
+    <TableCell className="animate-pulse" colSpan={100}>
+      <div className="h-8 w-full rounded-md bg-neutral-100" />
+    </TableCell>
+  </TableRow>
+);
+
+const Table = React.forwardRef<HTMLTableElement, TableProps>(
```
scopsy, line 21: Added a default loading state to the table component

apps/dashboard/src/api/activity.ts
```diff
@@ -0,0 +1,51 @@
...
+  page = 0,
+  filters?: IActivityFilters
+): Promise<ActivityResponse> {
+  const searchParams = new URLSearchParams();
+  searchParams.append('page', page.toString());
+
+  if (filters?.channels?.length) {
+    searchParams.append('channels', filters.channels.join(','));
+  }
+  if (filters?.templates?.length) {
```
SokratisVidros: Since the template concept doesn't exist in V2 and it's just an internal modelling issue, let's rename this to `workflows` for now as it's customer-facing.

SokratisVidros:
```suggestion
  if (filters?.workflows?.length) {
```

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+export interface ActivityTableProps {
+  selectedActivityId: string | null;
+  onActivitySelect: (activity: IActivity) => void;
+}
+
+export function ActivityTable({ selectedActivityId, onActivitySelect }: ActivityTableProps) {
+  const queryClient = useQueryClient();
+  const { currentEnvironment } = useEnvironment();
+  const hoverTimerRef = useRef<NodeJS.Timeout | null>(null);
```
SokratisVidros: #til

apps/dashboard/src/api/activity.ts
```diff
@@ -0,0 +1,51 @@
...
+  pageSize: number;
+}
+
+export function getActivityList(
+  environment: IEnvironment,
+  page = 0,
+  filters?: IActivityFilters
+): Promise<ActivityResponse> {
+  const searchParams = new URLSearchParams();
+  searchParams.append('page', page.toString());
```
SokratisVidros, line 26: Does the API work with page or limit offset? I see both in different parts of the code.

apps/dashboard/src/components/activity/activity-table.tsx
SokratisVidros: I like the idea, but I am not sure if it's practical, as the activity page is usually visited with a specific transaction ID.

If we want to give it a shot, l'd also try prefetching all entries from the first list using resource hints and Hemnet (same strategy we did in a previous PR for images) so that the browser does the orchestration automatically, it uses its cache, etc...

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,171 @@
...
+        </div>
+      </TableCell>
+      <TableCell className="px-3">
+        <div className="h-4 w-36 rounded bg-neutral-100 font-mono" />
+      </TableCell>
+    </TableRow>
+  );
+}
+
+function getSubscriberDisplay(subscriber?: Pick<ISubscriber, '_id' | 'subscriberId' | 'firstName' | 'lastName'>) {
```
SokratisVidros, line 180:
```suggestion
function getSubscriberDisplayName(subscriber?: Pick<ISubscriber, '_id' | 'subscriberId' | 'firstName' | 'lastName'>) {
```

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,171 @@
...
+  if (!subscriber) return '';
+
+  if (subscriber.firstName || subscriber.lastName) {
+    return `• ${subscriber.firstName || ''} ${subscriber.lastName || ''}`;
```
SokratisVidros, line 184:
```suggestion
    return `• ${subscriber.firstName || ''} ${subscriber.lastName || ''}`.trim();
```

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  return '';
```
SokratisVidros, line 187: Shall we fallback to email and phone or ID? It would look better than an empty column.

apps/dashboard/src/components/activity/components/pagination.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Button } from '@/components/primitives/button';
+
+interface PaginationProps {
+  offset: number;
+  limit: number;
+  hasMore: boolean;
+  onOffsetChange: (newOffset: number) => void;
+}
+
+export function Pagination({ offset, limit, hasMore, onOffsetChange }: PaginationProps) {
```
SokratisVidros: I strongly suggest reusing the pagination component from the workflow table, even if the Activity API doesn't allow it yet. We will isolate the API's compatibility layer on the hook level and then refactor it later.

apps/dashboard/src/components/activity/constants.ts
```diff
@@ -0,0 +1,92 @@
+import { JobStatusEnum, StepTypeEnum } from '@novu/shared';
+import { RiForbidFill } from 'react-icons/ri';
+import { RiErrorWarningLine } from 'react-icons/ri';
+import { BadgeVariant } from '../primitives/badge';
+import { RiCheckboxCircleLine, RiLoader3Line, RiLoader4Fill } from 'react-icons/ri';
+import { IconType } from 'react-icons/lib';
+
+export const STEP_TYPE_LABELS: Record<StepTypeEnum, string> = {
+  [StepTypeEnum.EMAIL]: 'Email',
```
SokratisVidros: I am sure this constant exists elsewhere! I've seen it 10 times :p

apps/dashboard/src/components/time-display-hover-card.tsx
```diff
@@ -0,0 +1,40 @@
+import { format, formatDistanceToNow } from 'date-fns';
+import { HoverCard, HoverCardContent, HoverCardTrigger } from './primitives/hover-card';
+
+interface TimeDisplayHoverCardProps {
+  date: Date;
+  children?: React.ReactNode;
+}
+
+export function TimeDisplayHoverCard({ date, children }: TimeDisplayHoverCardProps) {
```
SokratisVidros: Lovely. We need this component in every timestamp we display in the Dashboard. This is already in the dashboard table for the last updated at column so let's reuse it.

apps/dashboard/src/hooks/use-activities.ts
```diff
@@ -0,0 +1,38 @@
...
+  const { currentEnvironment } = useEnvironment();
+  const [searchParams] = useSearchParams();
+
+  const offset = parseInt(searchParams.get('offset') || '0');
+  const limit = parseInt(searchParams.get('limit') || '10');
+
+  const { data, isLoading, isFetching } = useQuery<ActivityResponse>({
+    queryKey: ['activitiesList', currentEnvironment?._id, offset, limit, filters],
+    queryFn: () => getActivityList(currentEnvironment!, Math.floor(offset / limit), filters),
+    staleTime: 0,
```
SokratisVidros: Please make it a dependent query on the environment.
```suggestion
    enabled: currentEnvironment._id
```

apps/dashboard/src/hooks/use-activities.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export function useActivities({ filters }: UseActivitiesOptions = {}) {
+  const { currentEnvironment } = useEnvironment();
+  const [searchParams] = useSearchParams();
+
+  const offset = parseInt(searchParams.get('offset') || '0');
+  const limit = parseInt(searchParams.get('limit') || '10');
+
+  const { data, isLoading, isFetching } = useQuery<ActivityResponse>({
```
SokratisVidros:
```suggestion
  const { data, ...rest } = useQuery<ActivityResponse>({
```

apps/dashboard/src/hooks/use-activities.ts
```diff
@@ -0,0 +1,38 @@
...
+    queryKey: ['activitiesList', currentEnvironment?._id, offset, limit, filters],
+    queryFn: () => getActivityList(currentEnvironment!, Math.floor(offset / limit), filters),
+    staleTime: 0,
+  });
+
+  return {
+    activities: data?.data || [],
+    hasMore: data?.hasMore || false,
+    pageSize: limit,
+    isLoading,
```
SokratisVidros:
```suggestion
    ...rest
```

packages/shared/src/entities/activity-feed/activity.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+  _id: string;
+  _templateId: string;
+  _environmentId: string;
+  _organizationId: string;
+  _subscriberId: string;
+  _digestedNotificationId?: string;
+  transactionId: string;
+  channels: ChannelTypeEnum[];
+  to: {
+    subscriberId: string;
```
SokratisVidros, line 21: What if this is a topic? Is the nesting necessary at this point or `to: string` is enough?

SokratisVidros (changes requested):

apps/dashboard/src/hooks/use-activities.ts
```diff
@@ -0,0 +1,38 @@
+import { useQuery } from '@tanstack/react-query';
```
SokratisVidros: Please rename to `use-fetch-activities.ts` to align with the hook naming conventions.

packages/shared/src/entities/activity-feed/activity.interface.ts
scopsy: Even for a topic, this data will always be the result of a fanout for a given subscriber. This aligns with the API result schema, so I left it as-is.

apps/dashboard/src/api/activity.ts
scopsy: It uses a page, I will adjust it in the other prs 💪

apps/dashboard/src/components/activity/activity-table.tsx
scopsy: I think this can be mostly useful during the initial building and onboarding phases, where a user might be triggering something and then going to check it. I will leave it as I think it makes the page feel more snappy. Regarding using browser hints, I'm not sure if it will work as we don't use cookies and etc.. for auth.

apps/dashboard/src/components/activity/activity-table.tsx
scopsy: I was considering this, let's add it. Why not :)

apps/dashboard/src/components/activity/constants.ts
scopsy: I couldn't find anything related :) We have one for icons, maybe that's the one you've seen

apps/dashboard/src/components/activity/components/pagination.tsx
scopsy: I initially tried, but the difference is too big and makes a lot of weird props and if statements. I would prefer to keep this one for the arrow navigations. We might need this also on the subscribers page.

scopsy: Renamed it to ArrowNavigation

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,172 @@
...
+}
+
+export function ActivityTable({ selectedActivityId, onActivitySelect, filters }: ActivityTableProps) {
+  const queryClient = useQueryClient();
+  const { currentEnvironment } = useEnvironment();
+  const hoverTimerRef = useRef<NodeJS.Timeout | null>(null);
+  const [searchParams] = useSearchParams();
+  const location = useLocation();
+  const navigate = useNavigate();
+  const { activities, isLoading, hasMore } = useFetchActivities({ filters });
```
LetItRock: We need to handle the cases when there will be no activity feed or when the request fails. For example, I can now type `page=asfa` in the URL address bar, and it will fail.

apps/dashboard/src/components/activity/activity-table.tsx
LetItRock: What's the point of having this short timeout? to debounce a bit requests when moving the mouse around?

The problem is that we will fire many requests just for nothing in the case when the user is just trying to find the activity item. Another problem is that the requests are not cached, meaning if I have on the same item twice, it will fire requests twice.

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,172 @@
...
+  return format(new Date(date), 'MMM d yyyy, HH:mm:ss');
+}
+
+function SkeletonRow() {
+  return (
+    <TableRow className="animate-pulse">
+      <TableCell className="px-3">
+        <div className="flex flex-col gap-1">
+          <div className="h-5 w-36 rounded bg-neutral-200" />
+          <div className="h-2.5 w-20 rounded bg-neutral-100" />
```
LetItRock: Please use the `skeleton` primitive.

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,172 @@
...
+              onMouseEnter={() => handleRowMouseEnter(activity)}
+              onMouseLeave={handleRowMouseLeave}
+            >
+              <TableCell className="px-3">
+                <div className="flex flex-col">
+                  <span className="text-foreground-950 font-medium">
+                    {activity.template?.name || 'Deleted workflow'}
+                  </span>
+                  <span className="text-foreground-400 text-[10px] leading-[14px]">
+                    {activity.transactionId} {getSubscriberDisplay(activity.subscriber)}
```
LetItRock: We should allow to copy this value with having a copy button.

apps/dashboard/src/components/activity/components/status-badge.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  const status = getActivityStatus(jobs);
+
+  const { variant, icon: Icon, label } = JOB_STATUS_CONFIG[status] || JOB_STATUS_CONFIG[JobStatusEnum.PENDING];
+  const displayLabel =
+    status === JobStatusEnum.FAILED ? `${errorCount} ${errorCount === 1 ? 'ERROR' : 'ERRORS'}` : label;
+
+  return (
+    <Popover open={isOpen}>
```
LetItRock, line 50: When I hover over the popover content, it blinks.
https://github.com/user-attachments/assets/3c4f57be-4cbf-4922-91d6-d7b273afd457

apps/dashboard/src/components/activity/components/status-badge.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { JOB_STATUS_CONFIG } from '../constants';
+
+export interface StatusBadgeProps {
+  jobs: IActivityJob[];
+}
+
+export function StatusBadge({ jobs }: StatusBadgeProps) {
+  const [isOpen, setIsOpen] = useState(false);
+  const errorCount = jobs.filter((job) => job.status === JobStatusEnum.FAILED).length;
+  let hoverTimeout: NodeJS.Timeout;
```
LetItRock:
```suggestion
  let hoverTimeout: ReturnType<typeof setTimeout> ;
```

apps/dashboard/src/components/activity/components/status-badge.tsx
LetItRock: It might be related to the mouse enter/leave events. There is no option to delay the initial mouse enter, but maybe we can use the `Tooltip` here instead which has that option. But I'm not sure.

apps/dashboard/src/components/activity/components/status-badge.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const { variant, icon: Icon, label } = JOB_STATUS_CONFIG[status] || JOB_STATUS_CONFIG[JobStatusEnum.PENDING];
+  const displayLabel =
+    status === JobStatusEnum.FAILED ? `${errorCount} ${errorCount === 1 ? 'ERROR' : 'ERRORS'}` : label;
+
+  return (
+    <Popover open={isOpen}>
+      <PopoverTrigger onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+        <Badge variant={variant as BadgeVariant} className="cursor-pointer gap-1.5">
+          <Icon className="h-3.5 w-3.5" />
```
LetItRock, line 53: Why are we not using the same icons as what we have in the workflows list page for the success and error? The UI components should be everywhere consistent.

apps/dashboard/src/components/activity/constants.ts
LetItRock: Please then move to the new reusable constant.

apps/dashboard/src/components/activity/constants.ts
```diff
@@ -0,0 +1,92 @@
...
+  [StepTypeEnum.DELAY]: 'Delay',
+  [StepTypeEnum.TRIGGER]: 'Trigger',
+  [StepTypeEnum.CUSTOM]: 'Custom',
+};
+
+export const STATUS_STYLES = {
+  completed: 'border-[1px] border-[#b4e6c5] bg-[#e8f9ef] text-[#b4e6c5]',
+  failed: 'border-[1px] border-[#fca5a5] bg-[#fde8e8] text-[#fca5a5]',
+  delayed: 'border-[1px] border-[#f7d794] bg-[#fef9e7] text-[#f7d794]',
+  default: 'border-[1px] border-[#e0e0e0] bg-[#f0f0f0] text-[#e0e0e0]',
```
LetItRock: can we please use the color names and not hardcoded hashes?

apps/dashboard/src/components/auth/inbox-playground.tsx
```diff
@@ -118,6 +118,7 @@ export function InboxPlayground() {
     };
 
     initializeDemoWorkflow();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
LetItRock, line 121: why not fix it?

apps/dashboard/src/components/primitives/table.tsx
LetItRock: Why it's needed if every table implementation is different?

apps/dashboard/src/components/time-display-hover-card.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const timeAgo = formatDistanceToNow(date, { addSuffix: true });
+
+  return (
+    <HoverCard openDelay={100} closeDelay={100}>
+      <HoverCardTrigger asChild className="hover:cursor-default">
+        <span>{children}</span>
+      </HoverCardTrigger>
+      <HoverCardContent className="w-fit" align="end" sideOffset={4}>
+        <div className="flex flex-col gap-2">
+          <div className="text-muted-foreground text-[10px] font-medium uppercase">Time Details</div>
```
LetItRock: we have `2xs` font size

apps/dashboard/src/hooks/use-fetch-activities.ts
```diff
@@ -0,0 +1,38 @@
...
+  hasMore: boolean;
+  pageSize: number;
+}
+
+export function useFetchActivities({ filters }: UseActivitiesOptions = {}) {
+  const { currentEnvironment } = useEnvironment();
+  const [searchParams] = useSearchParams();
+
+  const page = parseInt(searchParams.get('page') || '0');
+  const limit = parseInt(searchParams.get('limit') || '10');
```
LetItRock: we should parse these safely, as a user can type anything

packages/shared/src/entities/activity-feed/activity.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+import { IExecutionDetail } from '../execution-details';
+import { IJob as IJobBase } from '../job';
+import { INotificationTemplate } from '../notification-template';
+import { ISubscriber } from '../subscriber';
+
+export interface IActivityJob extends IJobBase {
+  executionDetails: IExecutionDetail[];
+}
+
+export interface IActivity {
```
LetItRock, line 11: We don't use `I` to mark the interface.

apps/dashboard/src/components/activity/activity-table.tsx
```diff
@@ -0,0 +1,172 @@
...
+              className={cn(
+                'relative cursor-pointer hover:bg-neutral-50',
+                selectedActivityId === activity._id &&
+                  'bg-neutral-50 after:absolute after:right-0 after:top-0 after:h-[calc(100%-1px)] after:w-[5px] after:bg-neutral-200'
+              )}
+              onClick={() => onActivitySelect(activity)}
+              onMouseEnter={() => handleRowMouseEnter(activity)}
+              onMouseLeave={handleRowMouseLeave}
+            >
+              <TableCell className="px-3">
```
LetItRock: Missing the workflow icon for the code created workflows vs dashboard defined.
Btw please check the designs as we have a purple icon for the dashboard created ones.

apps/dashboard/src/components/activity/activity-table.tsx
scopsy: Well, if you decided to change the page to asfa you deserve it! Hehe, just kidding, going to add an error state handling good catch!

apps/dashboard/src/components/activity/activity-table.tsx
scopsy: I also wanted, the problem is that we don't have spacing for properly placing it on hover because of the subscriber details on the right side, and adding padding on the left made it look weird. @twentyone24 do you have any suggestions? we can make it copiable without the button just by hovering with a tooltip Click to Copy?

apps/dashboard/src/components/activity/components/status-badge.tsx
scopsy: Unfortuently tooltip is very different style-wise, i've implemented a simple timeout delay to handle the mouse transitioning and the flicker is now preveneted

apps/dashboard/src/components/activity/constants.ts
scopsy: Yes! Just learned it can be achieved with /40 thank you! :)

packages/shared/src/entities/activity-feed/activity.interface.ts
scopsy: This is located in the @novu/shared package, all of the interface here are with I prefix unless I'm missing something. For consistency across the package I will leave it as-is. We can do a complete refactor but it has to be across the board.

apps/dashboard/src/components/activity/activity-table.tsx
scopsy: This was removed by Naveen, he forgot to update the designs

apps/dashboard/src/components/primitives/table.tsx
scopsy: My thinking was to provide an easy way of doing loading states, unless more complex table structure exists

apps/dashboard/src/components/activity/activity-table.tsx
LetItRock: I didn't get the problem, here is how it's in the workflows list:
Screenshot 2024-12-12 at 09 52 57

We can do the same in the activity table:
Screenshot 2024-12-12 at 09 53 24

packages/shared/src/entities/activity-feed/activity.interface.ts
LetItRock: For the new interfaces, we decided not to use the `I`, all the rest of the old interfaces will change with time, in small steps.

apps/dashboard/src/components/activity/activity-table.tsx
LetItRock: But don't we need to have a way to differentiate?

LetItRock:
when the first row is hovered, the table header border disappears
https://github.com/user-attachments/assets/4ad96cb6-ec7b-48fa-8242-3ab862b1d0eb

PR #7264, opened 2024-12-10
refactor(dashboard): use workflow as a single resource

ChmaraX:

Full step data are now part of `GET workflow` response. We treat workflow as a single CRUD resource.

**Following issues are solved:**
- no need to explicitly refetch step data when updating step
- no need to do clunky react-query cache invalidation
- no need to get step issues from workflow dto, since step data were stale ([ref](https://github.com/novuhq/novu/compare/nv-4976-use-only-getput-workflow-endpoints-on-frontend?expand=1#diff-357e03675541ce1e41c6ab7ed5593f59ff568fc4fbb2d84891b1b20f0f76d748L32-L48))
- no need to use nest multiple workflow and step providers, no need to wrap route paths that need step data in provider 
- we are able to use step data when rendering workflow on canvas - we can display control values directly on node 
- straightforward implementation: you update workflow/step -> you get updated data back -> everything invalidates and updates automatically as it should 
- all data needed to build a workflow are complete and returned from single endpoint - could be useful for export/import etc in future when we open API

**Potential downside:**
The size of transferred data in responses will be larger, but considering the average amount of steps in workflow it should not be an issue and if it will be, like in case of high volume of traffic where every bit of size matters, then thats a good problem to have. 

**Next steps:**
- Update e2e API tests to reflect the actual user flow - we are using patch/get step endpoints in API e2e tests but we never use nor we used before any of these endpoints on the web


desiprisg (approved):
💯

apps/dashboard/src/components/workflow-editor/step-utils.ts
```diff
@@ -48,3 +57,11 @@ export const updateStepInWorkflow = (
     }),
   };
 };
+
+export const createStep = (type: StepTypeEnum): Step => ({
+  name: STEP_NAME_BY_TYPE[type],
+  stepId: '',
+  slug: '_st_',
```
SokratisVidros, line 64: Does the slug start with `_`? Shouldn't this be? 
```suggestion
  slug: 'st_',
```

SokratisVidros (approved):

apps/dashboard/src/components/workflow-editor/step-utils.ts
ChmaraX: Should be correct, its `delay-step_st_<id>`

i just copied it to different file so I guess it worked before

PR #7265, opened 2024-12-10
feat(dashboard): identify provider centralization

scopsy:
- Added Segment identify
- Added LaunchDarkly Identify
- Moved the useMonitoring for sentry to the same provider

### Screenshots

<details>

</details>


apps/dashboard/src/context/identity-provider.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useSegment } from './segment/hooks';
+import { setUser as sentrySetUser } from '@sentry/react';
+
+export function IdentityProvider({ children }: { children: React.ReactNode }) {
+  const ldClient = useLDClient();
+  const segment = useSegment();
+  const { currentUser, currentOrganization } = useAuth();
+  const hasIdentified = useRef(false);
+
+  const isNovuUser = currentUser && currentUser._id && !currentUser._id.startsWith('user_');
```
SokratisVidros: The `user_` check lead me [to this](https://github.com/novuhq/novu/blob/f2bfd0a6045a6bb6e8c26bd8effa1a0683db003a/apps/dashboard/src/context/auth/mappers.ts#L19) that needs to be revised as its an unexpected hack.

Since the hack is related to useMonitoring, I suggest, returning the `_id` to the current_user only if the `clerkUser.externalId` is set and then this condition becomes:

```suggestion
  const isNovuUser = currentUser && currentUser._id;
```

SokratisVidros (changes requested):

apps/dashboard/src/context/identity-provider.tsx
scopsy: Wouldn't it lead to unexpected race conditions during signup flow/onboarding? We use the useAuth there aswell, I'm just not fully aware of potentially effects of this change.

scopsy: (Wondering at what scenario the externalId will be empty)

scopsy: Did a small test with the change, and it seemed to work. Just hope  it won't bite us back on some strange race condition 😅

SokratisVidros (approved):

apps/dashboard/src/context/auth/mappers.ts
```diff
@@ -42,11 +42,11 @@ export const toOrganizationEntity = (clerkOrganization: OrganizationResource): I
    * entity identifier) that gets used further in the app. There are a few consumers that
```
SokratisVidros: Let's rephrase this comment with the latest status.

PR #7267, opened 2024-12-10
feat(api): add query parser

djabarovgeorge:
### Screenshots

<details>

</details>


packages/shared/src/dto/workflows/step.dto.ts
```diff
@@ -84,6 +85,7 @@ export enum UiComponentEnum {
   CHAT_BODY = 'CHAT_BODY',
   PUSH_BODY = 'PUSH_BODY',
   PUSH_SUBJECT = 'PUSH_SUBJECT',
+  SKIP_RULES = 'SKIP_RULES',
```
djabarovgeorge: was not sure how to call. should we switch to SKIP?

```suggestion
  SKIP = 'SKIP',
```

apps/api/src/app/shared/services/query-parser/query-parser.service.ts
```diff
@@ -0,0 +1,47 @@
+import jsonLogic, { AdditionalOperation, RulesLogic } from 'json-logic-js';
+
+const initializeCustomOperators = (): void => {
```
djabarovgeorge, line 55: self note: 
validate that we cover all the `react-querybuilder` operators.

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -199,7 +209,25 @@ export class ConstructFrameworkWorkflow {
 
     return foundWorkflow;
   }
+
+  private processSkipOption(controlValues: { [x: string]: unknown }, variables: FullPayloadForRender) {
```
djabarovgeorge, line 214: we can convert this to `renderer`.

as a side note, at the moment i am not sure we need to have Injectable use-cases for the `renderer` feels like overkill.

apps/api/src/app/shared/services/query-parser/query-parser.service.ts
djabarovgeorge: done

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -170,22 +178,24 @@ export class ConstructFrameworkWorkflow {
...
-      ...this.constructCommonStepOptions(staticStep),
+      ...this.constructCommonStepOptions(staticStep, fullPayloadForRender),
     };
   }
 
   @Instrument()
-  private constructCommonStepOptions(staticStep: NotificationStepEntity): Required<StepOptions> {
+  private constructCommonStepOptions(
+    staticStep: NotificationStepEntity,
+    fullPayloadForRender: FullPayloadForRender
```
djabarovgeorge, line 194: we need FullPayloadForRender variables(===frameweork.payload) data in the skip method to execute the skip function.

apps/api/src/app/events/e2e/trigger-event.e2e.ts
```diff
@@ -3223,10 +3226,169 @@ describe(`Trigger event - /v1/events/trigger (POST)`, function () {
       tenant,
       actor,
     };
-    console.log('request111', JSON.stringify(request, null, 2));
 
     return (await novuClient.trigger(request)).result;
   }
+
+  describe('Trigger Event New Dashboard - /v1/events/trigger (POST)', function () {
```
djabarovgeorge: should we rename it to v2 workflow?

```suggestion
  describe('Trigger Event v2 workflow - /v1/events/trigger (POST)', function () {
```

apps/api/src/app/shared/services/query-parser/query-parser.service.spec.ts
djabarovgeorge, line 1: the main reason for this spec is that we are adding new operators to the json logic js, which are not tests in json logic js. Therefore, we need to make sure we are testing it here.

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -103,65 +108,68 @@ export class ConstructFrameworkWorkflow {
...
+          this.constructActionStepOptions(staticStep, fullPayloadForRender)
         );
       case StepTypeEnum.DELAY:
         return step.delay(
           stepId,
           async (controlValues) => {
             return this.delayOutputRendererUseCase.execute({ controlValues, fullPayloadForRender });
           },
-          this.constructActionStepOptions(staticStep)
+          this.constructActionStepOptions(staticStep, fullPayloadForRender)
```
SokratisVidros, line 160: Nit-picky: This function is the same across all steps, we can remove it out side the switch.

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -170,22 +178,24 @@ export class ConstructFrameworkWorkflow {
...
+  ): Required<StepOptions> {
     return {
       // TODO: fix the `JSONSchemaDto` type to enforce a non-primitive schema type.
       controlSchema: staticStep.template!.controls!.schema as JsonSchema,
-      /*
-       * TODO: add conditions
-       * Used to construct conditions defined with https://react-querybuilder.js.org/ or similar
-       */
-      skip: (controlValues) => false,
+      skip: (controlValues: { [x: string]: unknown }) => this.processSkipOption(controlValues, fullPayloadForRender),
```
SokratisVidros:
```suggestion
      skip: (controlValues: Record<string, unknown>) => this.processSkipOption(controlValues, fullPayloadForRender),
```

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -199,7 +209,25 @@ export class ConstructFrameworkWorkflow {
 
     return foundWorkflow;
   }
+
+  private processSkipOption(controlValues: { [x: string]: unknown }, variables: FullPayloadForRender) {
+    const skipRules = controlValues.skip as RulesLogic<AdditionalOperation>;
+
+    // TODO: check if we need to add more validation here (is empty object, etc.)
+    if (!skipRules) {
```
SokratisVidros: We can use lodash/isEmpty and cover all of the empty cases of an object ;)

```suggestion
    if (isEmpty(skipRules)) {
```

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
```diff
@@ -199,7 +209,25 @@ export class ConstructFrameworkWorkflow {
...
+      return false;
+    }
+
+    const { result, error } = evaluateRules(skipRules, variables);
+
+    if (error) {
+      this.logger.error({ err: error }, 'Failed to evaluate skip rule', LOG_CONTEXT);
```
SokratisVidros, line 224: Shall we pause workflow execution when there is a skip condition error? I think we should and we need to show it in activity feed.

apps/api/src/app/events/e2e/trigger-event.e2e.ts
SokratisVidros: yes

apps/api/src/app/shared/services/query-parser/query-parser.service.ts
SokratisVidros: Nice! @djabarovgeorge Isn't there an npm package that does this already so that it saves us from adding custom operators? 

Do they maybe have it for another JSON representation of conditionals that is more complete out of the box? 

What about [this](https://github.com/react-querybuilder/json-logic-js)?

If not, it's a great opportunity to make an OS contribution to react-querybuilder.

packages/shared/src/dto/workflows/step.dto.ts
SokratisVidros:
```suggestion
  SKIP_RULES = 'QUERY_EDITOR',
```

SokratisVidros (approved):
👏 👏 👏

apps/api/src/app/shared/services/query-parser/query-parser.service.ts
djabarovgeorge: Good idea, although i was not able to find any npm package that has it.

from the code looks like [@react-querybuilder/json-logic-js](https://github.com/react-querybuilder/json-logic-js/blob/main/src/logic.ts) is not aligning the param the react-querybuilder default operates support.

In this PR, I aligned [react query builder default operators](https://react-querybuilder.js.org/api/react-querybuilder/type-aliases/DefaultOperatorName) with [json logic js default operators](https://jsonlogic.com/operations.html).

From the conversation I saw on GitHub, it seems the community has strong opinions about the current operator support. If someone needs a new operator, they are directed to the "Add Operator" documentation.

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
djabarovgeorge: It is not the same one, we have one for Action step and one for Channel step, let's keep it like this for one.

apps/api/src/app/environments-v1/usecases/construct-framework-workflow/construct-framework-workflow.usecase.ts
djabarovgeorge: Good point. It’s better to avoid sending a notification altogether than to send an invalid or unintended one.

PR #7279, opened 2024-12-11
fix(novu): Respect .env values for API URL and SECRET_KEY

SokratisVidros:
This is just a DX improvement for `npx novu sync` command. If these env vars are set, novu sync command should use them instead of requiring the corresponding CLI Argumentthe


scopsy (approved):
❤️

PR #7292, opened 2024-12-12
fix(api): Crate of fixes part 2

SokratisVidros:

This PR applies some product changes:

1. Make the email body and subject optional
2. Do not return preview errors in the UI. If the step can't be previewed, we will tackle it separately with new product specs informing users about the missing variables. For now, return an empty preview.
3. Fix the naming of the email body step control.
4. Clean up some unused shared typed


desiprisg (approved):

apps/api/src/app/environments-v1/usecases/output-renderers/render-email-output.usecase.ts
```diff
@@ -14,8 +15,13 @@ export class RenderEmailOutputUsecase {
 
   @InstrumentUsecase()
   async execute(renderCommand: RenderEmailOutputCommand): Promise<EmailRenderOutput> {
-    const { emailEditor, subject } = EmailStepControlZodSchema.parse(renderCommand.controlValues);
-    const expandedSchema = this.transformForAndShowLogic(emailEditor, renderCommand.fullPayloadForRender);
+    const { body, subject } = EmailStepControlZodSchema.parse(renderCommand.controlValues);
```
djabarovgeorge, line 18: we will have regression here because people will already have stored in the DB emailEditor in the control value. or did we fix it somewhere else?

SokratisVidros: We haven't released the email in production. So I think there is no danger for prod.

PR #7296, opened 2024-12-13
fix(dal): adds & removes indicies across repository to make it consistent across environments

merrcury:

This fixes INF-457

### Screenshots

<details>

</details>

### DIFF
--- a/libs/dal/src/repositories/topic/topic-subscribers.schema.ts
+++ b/libs/dal/src/repositories/topic/topic-subscribers.schema.ts
@@ -40,6 +40,26 @@ const topicSubscribersSchema = new Schema<TopicSubscribersDBModel>(
   schemaOptions
 );
 
+topicSubscribersSchema.index({
+  _environmentId: 1,
+});
+
+topicSubscribersSchema.index({
+  _organizationId: 1,
+});
+
+topicSubscribersSchema.index({
+  _subscriberId: 1,
+});
+
+topicSubscribersSchema.index({
+  _topicId: 1,
+});
+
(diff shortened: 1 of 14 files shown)


libs/dal/src/repositories/topic/topic-subscribers.schema.ts
```diff
@@ -40,6 +40,26 @@ const topicSubscribersSchema = new Schema<TopicSubscribersDBModel>(
   schemaOptions
 );
 
+topicSubscribersSchema.index({
```
SokratisVidros, line 43: Are all these indices used in production? I'd expect a compound index on env and org.

libs/dal/src/repositories/topic/topic.schema.ts
```diff
@@ -29,5 +29,13 @@ const topicSchema = new Schema<TopicDBModel>(
   schemaOptions
 );
 
+topicSchema.index({
```
SokratisVidros, line 32: Same here, I think we can replace all these indices with a single one combining env and org. This can be done in a follow up PR. I will add a task to Linear.

SokratisVidros (approved):

libs/dal/src/repositories/topic/topic-subscribers.schema.ts
merrcury: Yes

libs/dal/src/repositories/topic/topic.schema.ts
merrcury: Cool

PR #7299, opened 2024-12-13
fix(dashboard): wait for save when switching preview tab

ChmaraX:

The preview components already had the logic to wait for preview data - the issue was that it was incorrectly mounted in `in-app` due to different tabs implementation compared to `email`.

- extracted the common tabs logic to wrapper, since it was the reason for the bug in the first place - we had slightly different implementation in each of them
- unified email and in-app editors

https://github.com/user-attachments/assets/25d47911-bf39-4c1a-8569-afe8798b53d1

slow network:

https://github.com/user-attachments/assets/455c5e1e-9054-4fb2-8327-950d69404071


apps/dashboard/src/components/workflow-editor/steps/email/email-tabs.tsx
```diff
@@ -1,76 +1,41 @@
...
+  const editorContent = (
+    <>
+      {isNovuCloud && <EmailEditor uiSchema={uiSchema} />}
+      {isExternal && (
+        <EmailTabsSection>
+          <CustomStepControls dataSchema={dataSchema} origin={workflow.origin} />
+        </EmailTabsSection>
+      )}
+    </>
+  );
```
ChmaraX, line 29: Added more explicit conditions so its unified across steps

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
```diff
@@ -12,8 +12,8 @@ const redirectKey = 'redirect';
 const primaryActionKey = 'primaryAction';
 const secondaryActionKey = 'secondaryAction';
 
-export const InAppEditor = ({ uiSchema }: { uiSchema?: UiSchema }) => {
-  if (!uiSchema || uiSchema?.group !== UiSchemaGroupEnum.IN_APP) {
```
ChmaraX, line 16: Not sure if its better to check for `uiSchema` in child or parent - moved to parent for consistency with email-editor

apps/dashboard/src/components/workflow-editor/steps/email/email-tabs.tsx
```diff
@@ -1,76 +1,41 @@
...
-          <TabsTrigger value="editor" className="gap-1.5">
-            <RiPencilRuler2Line className="size-5 p-0.5" />
-            <span>Editor</span>
-          </TabsTrigger>
-          <TabsTrigger value="preview" className="gap-1.5">
-            <Notification5Fill className="size-5 p-0.5" />
-            <span>Preview</span>
-          </TabsTrigger>
-        </TabsList>
+  const isNovuCloud = workflow.origin === WorkflowOriginEnum.NOVU_CLOUD && uiSchema;
```
SokratisVidros, line 17: Our goal should be to align Dashboard and Code First workflow management. That is, uiSchema should be available in both. Let's remove it from this check.

apps/dashboard/src/components/workflow-editor/steps/email/email-editor.tsx
```diff
@@ -2,15 +2,17 @@ import { Separator } from '@/components/primitives/separator';
...
 
 const subjectKey = 'subject';
 const emailEditorKey = 'emailEditor';
 
-type EmailEditorProps = { uiSchema: UiSchema };
-export const EmailEditor = (props: EmailEditorProps) => {
-  const { uiSchema } = props;
-  const { [emailEditorKey]: emailEditor, [subjectKey]: subject } = uiSchema?.properties ?? {};
+export const EmailEditor = ({ uiSchema }: { uiSchema: UiSchema }) => {
+  if (uiSchema.group !== UiSchemaGroupEnum.EMAIL) {
```
SokratisVidros, line 11: I suggest moving this logic outside the editor to the code that decides which group corresponds to each component. It's a bit of a code smell that the email editor needs to check for the enum.

It might not be necessary due to the existence of `getComponentByType`.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-editor.tsx
```diff
@@ -12,8 +12,8 @@ const redirectKey = 'redirect';
...
+export const InAppEditor = ({ uiSchema }: { uiSchema: UiSchema }) => {
+  if (uiSchema.group !== UiSchemaGroupEnum.IN_APP) {
```
SokratisVidros, line 16: Ditto

SokratisVidros (changes requested):

apps/dashboard/src/components/workflow-editor/steps/email/email-tabs.tsx
ChmaraX: Currently its not available in both and the uiSchema type is optional on the response, there needs to be a check somewhere and this is the only place where the check makes sense.

If we want to achieve what you are describing there needs to be additional changes on API which I don't want to include as part of this UI bugfix.

apps/dashboard/src/components/workflow-editor/steps/email/email-editor.tsx
ChmaraX: Email editor is the component that decides which group corresponds to each component. The only other place where this could live is in `<EmailTabs/>` which is arguably worse place.

SokratisVidros (approved):

PR #7300, opened 2024-12-13
feat(dashboard): design system tokens

scopsy:
- Added and aligned on all design tokens from Align UI
- Added missing tailwind classes and typography

### Screenshots

<details>

</details>


apps/dashboard/tailwind.config.js
```diff
@@ -1,20 +1,367 @@
...
+    },
+  ],
+  'paragraph-xs': [
+    '.75rem',
+    {
+      lineHeight: '1rem',
+      letterSpacing: '0em',
+      fontWeight: '400',
+    },
+  ],
```
twentyone24, line 131:
```suggestion
  ],
  'paragraph-2xs': [
  '0.625rem', 
  {
    lineHeight: '0.875rem', 
    letterSpacing: '0em',
    fontWeight: '400',
  },
],

```

twentyone24 (approved):
😍

apps/dashboard/src/index.css
```diff
@@ -49,33 +90,253 @@
...
+    --purple-900: 258 64% 32%;
+    --purple-800: 258 64% 40%;
+    --purple-700: 258 64% 48%;
+    --purple-600: 256 72% 56%;
+    --purple-500: 256 88% 64%;
+    --purple-400: 256 88% 64% / 80%;
+    --purple-300: 256 88% 64% / 60%;
+    --purple-200: 256 88% 64% / 40%;
+    --purple-100: 256 88% 64% / 20%;
+    --purple-50: 256 88% 64% / 10%;
```
desiprisg, line 178: Isn't it a bit weird that some colors have opacity and some not ?

scopsy: I tend to agree, @twentyone24 do you think we can make all of the opacity based? Or is it too much effort?

desiprisg:
I'm beginning to think that we should not add all of these right now. We started with the mentality that we will be adding on demand. All of these colors are unused. Maybe let's just do the semantic colors?

scopsy:
@desiprisg what do you mean by semantic colors? you mean removing the sky, and pink, teal and the ones that are unused? @twentyone24 what do you think? IMO, if we have a color on the design system on figma we should also have it here. Otherwise let's remove it from Figma.

apps/dashboard/src/index.css
```diff
@@ -18,7 +35,27 @@
     --foreground-900: 226 21% 12%;
     --foreground-950: 222 32% 8%;
 
-    --primary: 346 73% 50%;
+    --novu-800: 346 72% 42%;
```
SokratisVidros, line 38: Is Novu color the primary color?

apps/dashboard/src/index.css
```diff
@@ -49,33 +90,253 @@
...
+    --information-light: var(--blue-200);
+    --information-lighter: var(--blue-50);
+
+    --warning: var(--orange-500);
+    --warning-dark: var(--orange-950);
+    --warning-base: var(--orange-500);
+    --warning-light: var(--orange-200);
+    --warning-lighter: var(--orange-50);
+
+    --error-dark: var(--red-950);
```
SokratisVidros, line 286: Does Align UI suggest this practice? I am not well-versed in how modern design systems prefer to do this, but I'd avoid overloading CSS with too many semantics. As a pattern, it makes sense when the color needs to be reused in two semantic tokens, for example, error and destructive in our case, but I'd resolve this in the component layer.

That is, on the CSS level, we can define a red var, and then we will define a message that is an error and uses red or a destructive button that also uses red.

Let me know if this sounds reasonable.

apps/dashboard/tailwind.config.js
```diff
@@ -62,43 +420,120 @@ export default {
...
+        lighter: 'hsl(var(--warning-lighter))',
+      },
+      away: {
+        dark: 'hsl(var(--away-dark))',
+        base: 'hsl(var(--away-base))',
+        light: 'hsl(var(--away-light))',
+        lighter: 'hsl(var(--away-lighter))',
+      },
+      error: {
+        DEFAULT: 'hsl(var(--error))',
```
SokratisVidros, line 490: Please refer to my previous comment about adding too many tokens and variables on the CSS level.

apps/dashboard/src/index.css
scopsy: Align doesa separation between the color pallete, and their semantic usage. Novu is the color palette, while primary is the semantic usage of the novu color.

apps/dashboard/src/index.css
scopsy: Yes it is how align works, and destructive actually doesn't exists in align. They just call it error which uses the red color palette. I kept here the current destructive naming during the migration period. It will be removed in favor of the align tokens for errors

desiprisg:
> @desiprisg what do you mean by semantic colors? you mean removing the sky, and pink, teal and the ones that are unused? @twentyone24 what do you think? IMO, if we have a color on the design system on figma we should also have it here. Otherwise let's remove it from Figma.

Yes, I mean add what makes sense for us. Align ui provides 100x what we need, to  I don't understand why we need to use the colors in the boilerplate here. It was supposed to be a starting point and we would create our own design system. This never happened.

twentyone24:
> @desiprisg what do you mean by semantic colors? you mean removing the sky, and pink, teal and the ones that are unused? @twentyone24 what do you think? IMO, if we have a color on the design system on figma we should also have it here. Otherwise let's remove it from Figma.

Makes sense, I agree. 
but, we use most of the colors (eg: x-light of all colors) for steps. 
we might not use all three variations of a color, but I feel it is OK to have the group if atleast one is used.

apps/dashboard/src/index.css
LetItRock: If we can make it, then we don't need to define all the tokens in this CSS file. Instead, we will just have "base" colors and generate all the "semantic" ones in the tailwind config. Otherwise still we can simplify the opacity variables by defining them in the tw.

apps/dashboard/tailwind.config.js
scopsy: I would highly suggest keeping the align UI based approach, as it will ensure on modification of a color token all relevant components will be updated, in-case we would want to change things like color-feature for example. it is mapped perfectly to what naveen is having on Figma and it will simplify our maintenance for new components assell. As copy-paste from align UI snippets will work out of the box.

scopsy:
Team, I'm merging this PR as is. To maintain consistency and speed up front-end implementation, let's avoid creating custom patterns when using the Align UI design system. Following their definitions will help us achieve a more uniform UI with less work. I'm currently updating some of our components to align their prop terminology with the design system, ensuring that our Figma screens and code components use similar language. Once I have a proof of concept ready, I'll schedule an engineering session to review it and establish guidelines for modifying design system components and going forward.

PR #7302, opened 2024-12-13
feat(api): Nv 5045 update the api to have same behavior as preference

BiswaViraj:
- created delete control values usecase
- allow deleting the control values when null value is supplied
- updated the upsetworkflow and patchstep usecases to have same functionality of deletion of control values like preferences

### Screenshots

https://github.com/user-attachments/assets/a76b6c01-fd4c-43df-98ed-c52c1a1a5897

<details>

</details>


apps/dashboard/src/components/workflow-editor/steps/email/email-tabs.tsx
```diff
@@ -58,7 +58,7 @@ export const EmailTabs = (props: StepEditorProps) => {
       <Separator />
       <TabsContent value="editor" forceMount className={tabsContentClassName}>
         {workflow.origin === WorkflowOriginEnum.NOVU_CLOUD && uiSchema && <EmailEditor uiSchema={uiSchema} />}
-        {workflow.origin === WorkflowOriginEnum.EXTERNAL && (
+        {workflow.origin === WorkflowOriginEnum.EXTERNAL && !uiSchema && (
```
SokratisVidros: @BiswaViraj A short-tem goal is to unify uiSchema across all control types. That is, use it for custom step controls as well. That is, I suggest removing this check from the UI.

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-tabs.tsx
```diff
@@ -54,8 +54,11 @@ export const InAppTabs = (props: StepEditorProps) => {
       </header>
       <Separator />
       <TabsContent value="editor" className={tabsContentClassName}>
-        <InAppEditor uiSchema={uiSchema} />
-        <CustomStepControls dataSchema={dataSchema} origin={workflow.origin} />
+        {uiSchema ? (
```
SokratisVidros: Ditto

libs/application-generic/src/usecases/delete-control-values/delete-control-values.usecase.ts
```diff
@@ -0,0 +1,41 @@
...
+import { DeleteControlValuesCommand } from './delete-control-values.command';
+import { Instrument, InstrumentUsecase } from '../../instrumentation';
+
+@Injectable()
+export class DeleteControlValuesUseCase {
+  constructor(private controlValuesRepository: ControlValuesRepository) {}
+
+  @InstrumentUsecase()
+  public async execute(command: DeleteControlValuesCommand): Promise<void> {
+    const existingControlValues = await this.controlValuesRepository.findOne({
```
SokratisVidros: You can delete the values with a single query, instead of fetching and deleting. If the control is not found the delete query is just a Noop.

SokratisVidros (changes requested):

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -299,7 +299,7 @@ export class UpsertWorkflowUseCase {
   ): Promise<WorkflowInternalResponseDto> {
     for (const step of workflow.steps) {
       const controlValues = this.findControlValueInRequest(step, command.workflowDto.steps);
-      if (!controlValues) {
+      if (controlValues === undefined) {
```
SokratisVidros, line 302:
```suggestion
      if (!controlValues) {
```

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
BiswaViraj: It needs to be undefined, coz there is a null value too and we need to process the null value

PR #7303, opened 2024-12-13
refactor(api): refactor none-email control value preview usecase

djabarovgeorge:

refactoring all the none-email(tiptap) control value previews.
the first step for solving addressing ticket
https://linear.app/novu/issue/NV-5001/unify-liquidjs-processing-across-all-steps

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -241,3 +334,171 @@ class FrameworkError {
...
+    const validate = ajv.compile(controlSchema);
+    const isValid = validate(schemaCompliantControls);
+    const errors = validate.errors as null | ErrorObject[];
+
+    if (!isValid && errors && errors?.length !== 0 && schemaCompliantControls) {
+      fixedValues = replaceInvalidControlValues(schemaCompliantControls, errors);
+
+      return { issues, schemaCompliantControls: fixedValues };
+    }
+  }
```
djabarovgeorge: this part, with the help of strict JSON schema/Zod, will work well for generating `issues` on context of type validation (`"<param> required"`, `<param> must be URL`, etc...). However, in this context, I believe we can focus on improving `normalizeControlValues` to handle the heavy lifting. At this stage, the control values would then be clean and valid (aligned with the control schema and framework output schema), ensuring the preview runs smoothly.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -111,10 +111,72 @@ export class GeneratePreviewUsecase {
...
+
+    const { validVariables: validSchemaVariables, invalidVariables: invalidSchemaVariables } =
+      validateValidVariablesAgainstSchema(variableSchema, validVariables);
+
+    const resultVariables = {
+      valid: validSchemaVariables,
+      invalid: [...invalidVariables, ...invalidSchemaVariables],
+    };
+
+    return resultVariables;
```
djabarovgeorge: `resultVariables.invalid` will work well for generating issues based on the liquid errors (for example `{{subscriber..firstName}} is invalid`, etc...).

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -111,10 +111,72 @@ export class GeneratePreviewUsecase {
...
+  private buildVariableExample(
+    controlValueForPreview: Record<string, unknown>,
+    workflow: WorkflowInternalResponseDto,
+    commandVariablesExample: PreviewPayload | undefined
+  ) {
+    const templateVars = this.extractTemplateVariables([controlValueForPreview]);
+
+    if (templateVars.length === 0) {
+      // eslint-disable-next-line no-console
+      console.log('No template variables found');
```
djabarovgeorge: todo remove.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -202,6 +264,37 @@ export class GeneratePreviewUsecase {
...
+      const EMPTY_STRING = '';
+      const invalidVariableExpression = invalidVariable.template
+        .replace('{{', EMPTY_STRING)
+        .replace('}}', EMPTY_STRING);
+      const liquidJsParsableString = PREVIEW_ERROR_MESSAGE_PLACEHOLDER.replace(
+        INVALID_VARIABLE_PLACEHOLDER,
+        invalidVariableExpression
+      );
+
+      controlValuesString = replaceAll(controlValuesString, invalidVariable.template, liquidJsParsableString);
```
djabarovgeorge: This part will filter out all non-parsable content (which would otherwise throw errors in Liquid JS) and replace it with the placeholder error message PREVIEW_ERROR_MESSAGE_PLACEHOLDER. Invalid output without {{ }} will also be handled  (example `[[Invalid Variable: subscriber.. email | upcase]]`), ensuring the preview (in frameworks that use Liquid for parsing) doesn’t fail at runtime.

At this point, we have a single location where invalid "text" is identified. Now, we need to decide what to replace it with to provide the best possible UX. we could use an empty string here as well.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: please disregard the logic in the `if (controlSchema)`  statement (with the zod validation and replace by issues), as I will remove it in the future commit it in favor of utilizing `normalizeControlValues` more effectively, as explained earlier.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -241,3 +426,154 @@ class FrameworkError {
   message: string;
   name: string;
 }
+
+/**
+ * Validates liquid template variables against a schema, the result is an object with valid and invalid variables
+ * @example
+ * const variables = [
+ *   { name: 'subscriber.firstName' },
+ *   { name: 'subscriber.orderId' }
```
scopsy, line 435: How will we handle the data object on the subscriber?

scopsy (approved):

scopsy (approved):

apps/api/src/app/workflows-v2/shared/schemas/email-control.schema.ts
```diff
@@ -28,3 +31,8 @@ export const emailStepUiSchema: UiSchema = {
     skip: skipControl.uiSchema.properties.skip,
   },
 };
+
+export const emailStepControl = {
```
SokratisVidros, line 35: Nit-pick comment: It's a bit of an overstretch to re-export as an object. It also adds an extra level of nesting when we import. Why was it necessary?

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -17,74 +18,100 @@ import {
...
+  Variable,
+} from '../../util/template-parser/liquid-parser';
+import { pathsToObject } from '../../util/path-to-object';
+import { HydrateEmailSchemaUseCase } from '../../../environments-v1/usecases/output-renderers';
+import { PrepareAndValidateContentUsecase } from '../validate-content/prepare-and-validate-content/prepare-and-validate-content.usecase';
 
 const LOG_CONTEXT = 'GeneratePreviewUsecase';
 
+type DestructuredControlValues = {
+  tiptapControlValues: { emailEditor?: string | null; body?: string | null } | null;
```
SokratisVidros, line 42: @djabarovgeorge Didn't we replace emailEditor with body? I am surprised we still use it here. Is this maybe for backward compatibility? In production we don't have any email step controls yet so it shouldn't be a concern.

libs/application-generic/src/utils/index.ts
```diff
@@ -12,3 +12,4 @@ export * from './subscriber';
 export * from './variants';
 export * from './deepmerge';
 export * from './generate-id';
+export * from './sanitize-preview-control-values';
```
SokratisVidros, line 15: I suggest renaming this file to `sanitize-control-values.ts`. If not mistaken, it should be reused for preview and actual execution. A more generic name will work better.

libs/application-generic/src/utils/sanitize-preview-control-values.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  return normalized;
+}
+
+function sanitizeEmail(controlValues: Record<string, unknown>) {
+  if (!controlValues) return controlValues;
+
+  const emailControls: Record<string, unknown> = {};
+
+  /*
```
SokratisVidros, line 96: Please remove the commented code if not necessary.

libs/application-generic/src/utils/sanitize-preview-control-values.ts
```diff
@@ -0,0 +1,222 @@
...
+   * if (controlValues.body != null) {
+   *   emailControls.body = controlValues.body || '';
+   * }
+   */
+  emailControls.subject = controlValues.subject || '';
+  emailControls.body = controlValues.body || EMPTY_TIP_TAP_OBJECT;
+  emailControls.data = controlValues.data || null;
```
SokratisVidros, line 103: What is `emailControls.data`? We have a `data` field on `inApp` notifications but not on email.

libs/application-generic/src/utils/sanitize-preview-control-values.ts
```diff
@@ -0,0 +1,222 @@
...
+  };
+}
+
+function sanitizePush(controlValues: Record<string, unknown>) {
+  if (!controlValues) return controlValues;
+
+  const mappedValues = {
+    subject: controlValues.subject || '',
+    body: controlValues.body || '',
+    data: controlValues.data || null,
```
SokratisVidros, line 123: Ditto about the data field on push.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -111,39 +137,149 @@ export class GeneratePreviewUsecase {
...
+    } catch (e) {
+      return null;
+    }
   }
 
-  @Instrument()
-  private buildVariablesExample(
+  private async handleTipTapControl(
+    tiptapControlValue: {
+      emailEditor?: string | null;
```
SokratisVidros, line 168: Ditto about `emailEditor` vs `body`

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -111,39 +137,149 @@ export class GeneratePreviewUsecase {
...
-  ) {
-    if (workflow.origin !== WorkflowOriginEnum.EXTERNAL) {
-      return finalPayload;
+    commandVariablesExample: PreviewPayload | undefined
+  ): ProcessedControlResult {
+    const variables = this.processControlValueVariables(controlValues, variableSchema);
+    const processedControlValues = this.fixControlValueInvalidVariables(controlValues, variables.invalid);
+    const extractedTemplateVariables = variables.valid.map((variable) => variable.name);
+    const payloadVariableExample =
+      workflow.origin === WorkflowOriginEnum.EXTERNAL
```
SokratisVidros, line 206: Ideally, we need to do payloadVariable extraction only if the workflow has a `payloadSchema` without checking if it's external. 

Currently, only code-first workflows have `payloadSchema` but soon all workflows will have. Meanwhile, we still need to clear the payload schema from all the dashboard workflows in the Mongo database; we will do that next week. Regardless, the code should be ready for that. It's another step towards unifying the pipeline across all types of workflows.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -17,74 +18,100 @@ import {
...
+      const sanitizedValidatedControls = sanitizePreviewControlValues(initialControlValues, stepData.type);
+
+      if (!sanitizedValidatedControls) {
+        throw new Error(
+          // eslint-disable-next-line max-len
+          'Control values normalization failed: The normalizeControlValues function requires maintenance to sanitize the provided type or data structure correctly'
+        );
+      }
+
+      const destructuredControlValues = this.destructureControlValues(sanitizedValidatedControls);
```
SokratisVidros, line 82: @djabarovgeorge I think we can simplify this part even more. There is no need to use intermediate structures such as destructuredControlValues that group variables into two categories (tiptap and simple). Alternatively, we can use a single function that parses the controls one after the other and checks whether the control is tiptap or simple.

That function will return the final previewData object at once without doing a final merge. 

Effectively, it's a `reduce` function on the step controls array that returns the preview Data at once.

Let me know if this makes sense.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -111,39 +137,149 @@ export class GeneratePreviewUsecase {
...
 
-  /**
-   * Merges the payload schema into the variables schema to enable proper validation
-   * and sanitization of control values in the prepareAndValidateContentUsecase.
-   */
-  @Instrument()
-  private buildVariablesSchema(variables: Record<string, unknown>, payloadSchema: JSONSchemaDto) {
-    if (Object.keys(payloadSchema).length === 0) {
-      return variables;
+  private async safeAttemptToParseEmailSchema(
```
SokratisVidros, line 140: Consider renaming to `safeParseTiptapState`

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -111,39 +137,149 @@ export class GeneratePreviewUsecase {
...
+  private async initializePreviewContext(command: GeneratePreviewCommand) {
+    const stepData = await this.getStepData(command);
+    const controlValues = command.generatePreviewRequestDto.controlValues || stepData.controls.values || {};
+    const workflow = await this.findWorkflow(command);
+    const variableSchema = await this.buildVariablesSchema(stepData.variables, command, controlValues);
+
+    return { stepData, controlValues, variableSchema, workflow };
+  }
+
+  private processControlValueVariables(
```
SokratisVidros, line 242: Nice! 👏

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: this one was before we introduced email.body, we already removed the old emailEditor

PR #7310, opened 2024-12-16
feat(dashboard): new integrations page view

scopsy:

- Introduces a new integrations page list view
- SegmentedControl component as a standalone control
- TabsList component refactored to use the SegmentedControl
- Improved some alignments of other pages like workflows and settings page to be according to the design

* This PR contains a lot of new logo svg files moved from web to show integration icons

### Screenshots
CleanShot 2024-12-16 at 15 51 13@2x

<details>

</details>


apps/api/src/app/integrations/usecases/check-integration/check-integration.usecase.ts
```diff
@@ -21,7 +21,7 @@ export class CheckIntegration {
         );
       }
 
-      throw e;
+      throw new BadRequestException(e.message);
```
scopsy, line 24: Throw a proper validation error, previously it was filtered by the new exception pipe middleware

apps/dashboard/src/api/api.client.ts
```diff
@@ -48,7 +48,7 @@ const request = async <T>(
 
     if (!response.ok) {
       const errorData = await response.json();
-      throw new NovuApiError(`Novu API error`, errorData, response.status);
+      throw new NovuApiError(parseErrorMessage(errorData), response.status, errorData);
```
scopsy, line 51: Provided an easier DX for consuming error messages, using error.message. Also, the server responds to the error as a JSON string and parses it as JSON for easier consumption on the front end. @galezra BTW, is there a specific reason for us stringifying it?

apps/dashboard/src/components/primitives/segmented-control.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+>(({ children, className, floatingBgClassName, ...rest }, forwardedRef) => {
+  const [lineStyle, setLineStyle] = React.useState({ width: 0, left: 0 });
+
+  const { mounted, listRef } = useTabObserver({
+    onActiveTabChange: (_, activeTab) => {
+      const { offsetWidth: width, offsetLeft: left } = activeTab;
+      setLineStyle({ width, left });
+    },
+  });
```
scopsy, line 25: This new component is later needed by the create and update sidebars, I have extracted to a reusable SegmentedControl component, which is also used by the tabs component. For instances where the segmented control is used as a form element and not tabbed functionality

https://github.com/user-attachments/assets/533ae16c-0a0a-441b-9cf5-197e38dbc502

apps/dashboard/src/components/primitives/tabs.tsx
```diff
@@ -22,24 +25,64 @@ const tabsListVariants = cva('inline-flex', {
...
+            hidden: !mounted,
+          })}
+          style={{
+            transform: `translate3d(${lineStyle.left}px, 0, 0)`,
+            width: `${lineStyle.width}px`,
+            transitionTimingFunction: 'cubic-bezier(0.65, 0, 0.35, 1)',
+          }}
+          aria-hidden="true"
+        />
+      </TabsPrimitive.List>
```
scopsy, line 72: I also updated the regular tabs to have a floating island. 

https://github.com/user-attachments/assets/76c19e08-794f-4bda-9045-45388e9281ed

apps/dashboard/src/pages/test-workflow.tsx
```diff
@@ -18,7 +18,7 @@ export const TestWorkflowPage = () => {
     <>
       <PageMeta title={`Trigger ${workflow?.name}`} />
       <EditWorkflowLayout headerStartItems={<EditorBreadcrumbs />}>
-        {testData && <TestWorkflowTabs testData={testData} />}
+        <TestWorkflowTabs testData={testData} />
```
scopsy, line 21: This was causing a flicker when switching tabs which didn't looked good, now the transition is smoother

apps/dashboard/src/pages/workflows.tsx
```diff
@@ -33,14 +31,13 @@ export const WorkflowsPage = () => {
...
-            <Input placeholder="Search workflows" />
-          </div>
+        <div className="flex justify-between px-2.5 py-2.5">
+          <div className="invisible flex w-[20ch] items-center gap-2 rounded-lg bg-neutral-50 p-2"></div>
 
           <CreateWorkflowButton asChild>
-            <Button variant="primary">Create workflow</Button>
+            <Button variant="primary" size="sm">
+              Create workflow
+            </Button>
```
scopsy, line 40: The page header was not aligned as per the designs. This was quite noticeable between other page transitions from integration store

apps/dashboard/src/pages/settings.tsx
```diff
@@ -57,9 +57,6 @@ export function SettingsPage() {
   const location = useLocation();
   const isV2BillingEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_V2_DASHBOARD_BILLING_ENABLED);
 
-  const TAB_TRIGGER_CLASSNAME =
```
scopsy, line 60: Refactored to use the existing designs of the tab list

apps/dashboard/src/hooks/use-tab-observer.ts
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react';
+
+interface TabObserverOptions {
+  onActiveTabChange?: (index: number, element: HTMLElement) => void;
+}
+
+export function useTabObserver({ onActiveTabChange }: TabObserverOptions = {}) {
```
scopsy, line 7: Required for floating island

apps/dashboard/src/components/workflow-editor/workflow-tabs.tsx
```diff
@@ -10,19 +10,15 @@ export const WorkflowTabs = () => {
   const { workflow } = useWorkflow();
   const { currentEnvironment } = useEnvironment();
 
-  if (!currentEnvironment || !workflow) {
-    return null;
-  }
```
scopsy, line 15: This prevents a full page flicker when switching tabs and reloading the page

apps/dashboard/src/api/api.client.ts
galezra: From my perspective, if all service use cases return the same error interface, there is no need to stringify the error result, especially if the API handles this for each request. To create an ideal interface, I suggest that each use case inherit from a base class. The base class should implement an execute method that calls the concrete execute method of the specific use case, following the method pattern

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
+import { Badge } from '@/components/primitives/badge';
```
LetItRock, line 1: Please move all the component files from the `src/pages/integrations/components` directory to the `src/components/integrations`. If the utils are reusable between the modules please move them to the `utils` folder, otherwise, keep them under the `src/components/integrations`.

apps/dashboard/src/pages/integrations/integrations-list-page.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <TabsList variant="regular" className="border-b-0 border-t-2 border-transparent p-0 !px-2">
+            <TabsTrigger value="providers" variant="regular">
+              Providers
+            </TabsTrigger>
+            <TabsTrigger value="data-warehouse" variant="regular" disabled>
+              Data{' '}
+              <Badge kind="pill" size="2xs">
+                SOON
+              </Badge>
+            </TabsTrigger>
```
LetItRock: Maybe we should have a tooltip explaining what this will do?

apps/dashboard/src/pages/integrations/components/integrations-list.tsx
```diff
@@ -0,0 +1,91 @@
...
+    </div>
+  );
+}
+
+export function IntegrationsList({ onRowClickCallback }: IntegrationsListProps) {
+  const { currentEnvironment, environments } = useEnvironment();
+  const { integrations } = useFetchIntegrations();
+  const availableIntegrations = novuProviders;
+
+  if (!integrations || !availableIntegrations || !currentEnvironment) {
```
LetItRock: - we don't need to check for the `availableIntegrations` because it's from the `@novu/shared` package.
- using `integrations` as the `loading` indicator is a bad idea because in case of an API error the screen will still show the loading state instead of an empty list component, which is btw missing

apps/dashboard/src/pages/integrations/components/integrations-list.tsx
```diff
@@ -0,0 +1,91 @@
...
+    (acc, integration) => {
+      const channel = integration.channel;
+      if (!acc[channel]) {
+        acc[channel] = [];
+      }
+      acc[channel].push(integration);
+      return acc;
+    },
+    {} as Record<ChannelTypeEnum, typeof integrations>
+  );
```
LetItRock: Please wrap with `useMemo` otherwise, it slows down a bit the rendering phase.

apps/dashboard/src/pages/integrations/components/integrations-list.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <IntegrationCardSkeleton />
+        <IntegrationCardSkeleton />
+      </div>
+    </div>
+  );
+}
+
+export function IntegrationsList({ onRowClickCallback }: IntegrationsListProps) {
+  const { currentEnvironment, environments } = useEnvironment();
+  const { integrations } = useFetchIntegrations();
```
LetItRock: Question: shouldn't we split integrations now per env? Because the URL justifies that.

apps/dashboard/src/pages/integrations/types.ts
```diff
@@ -0,0 +1,14 @@
+import { ChannelTypeEnum } from '@novu/shared';
+
+export interface ITableIntegration {
```
LetItRock: please let's not use `I` and prefer the `type` over `interface` this is our convention

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const tableIntegration: ITableIntegration = {
+    integrationId: integration._id ?? '',
+    name: integration.name,
+    identifier: integration.identifier,
+    provider: provider.displayName,
+    channel: integration.channel,
+    environment: environment.name,
+    active: integration.active,
+  };
```
LetItRock: please define this object under the `handleConfigureClick`

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleConfigureClick = (e: React.MouseEvent) => {
+    if (integration.channel === 'in_app' && !integration.connected) {
```
LetItRock:
```suggestion
    if (integration.channel === ChannelTypeEnum.IN_APP && !integration.connected) {
```

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+  integration: IIntegration;
+  provider: IProviderConfig;
+  environment: IEnvironment;
+  onRowClickCallback: (item: { original: ITableIntegration }) => void;
+}
+
+export function IntegrationCard({ integration, provider, environment, onRowClickCallback }: IntegrationCardProps) {
+  const navigate = useNavigate();
+
+  const tableIntegration: ITableIntegration = {
```
LetItRock: we don't have a "table" anywhere

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+    provider: provider.displayName,
+    channel: integration.channel,
+    environment: environment.name,
+    active: integration.active,
+  };
+
+  const handleConfigureClick = (e: React.MouseEvent) => {
+    if (integration.channel === 'in_app' && !integration.connected) {
+      e.stopPropagation();
+      navigate(ROUTES.INBOX_EMBED);
```
LetItRock: When it redirects me to the onboarding flow, it shows that I successfully configured the inbox, which is not true because I didn't for the `production` env.
Screenshot 2024-12-17 at 16 54 26

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+                {provider.channel === 'email' ? 'emails' : 'sms'} per month. Not suitable for production use.
+              </p>
+            </TooltipContent>
+          </Tooltip>
+        )}
+      </div>
+
+      <div className="mt-auto flex items-center gap-2">
+        {integration.channel === 'in_app' && !integration.connected ? (
+          <Button size="xs" className="h-[26px]" variant="outline" onClick={handleConfigureClick}>
```
LetItRock, line 100: if that's the size we would like to use, let's introduce it to the button component

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const isDemo = isDemoIntegration(provider.id);
+
+  return (
+    <div
+      className={cn(
+        'bg-card shadow-xs group relative flex min-h-[125px] cursor-pointer flex-col gap-2 overflow-hidden rounded-xl border border-neutral-100 p-3 transition-all hover:shadow-lg',
+        !integration.active && 'opacity-75 grayscale'
+      )}
+      onClick={handleConfigureClick}
+      data-test-id={`integration-${integration._id}-row`}
```
LetItRock: Novu integrations should have a gradient background:
Screenshot 2024-12-17 at 17 01 09

apps/dashboard/src/pages/integrations/components/integration-card.tsx
```diff
@@ -0,0 +1,117 @@
...
+          {environment.name}
+        </Badge>
+      </div>
+    </div>
+  );
+}
+
+export function isDemoIntegration(providerId: string) {
+  return providerId === 'novu-email' || providerId === 'novu-sms';
+}
```
LetItRock: no need for the export, and please define it with using enums and move it to the utils

apps/dashboard/src/pages/integrations/components/integration-card.tsx
LetItRock: Please also update the border: https://www.figma.com/design/iOS9MnuJjLR7avdSIHemFy?node-id=9685-198980&m=dev#1067690831

apps/dashboard/src/pages/integrations/components/integration-card.tsx
scopsy: Oh my 🤦‍♂️ was prototyping initially and forgot to move them. I blame keyboard shortcuts for this hehe

apps/dashboard/src/pages/integrations/components/integrations-list.tsx
scopsy: Good catch, I think they still be global. But good catch, let me try to see what happens when it's not under env id

apps/dashboard/src/pages/integrations/types.ts
scopsy: My bad, missed to refactor this one.

apps/dashboard/src/pages/integrations/components/integration-card.tsx
scopsy: It's not a size that will be used outside of this component. @twentyone24 I've added this Configure button on the In App provider in case the user hasn't connected yet the integration. Any suggestions on other UX pattern for it? I had to adjust the button size to be in the same size of the Env pill

apps/dashboard/src/pages/integrations/components/integration-card.tsx
scopsy: This design was updated, just have some old things there still cc. @twentyone24 for posterity ler's remove anything that's not relevant from figma https://www.figma.com/design/iOS9MnuJjLR7avdSIHemFy/novu-v3.0-Design-System-%E2%9A%A1%EF%B8%8F?node-id=9685-199066&m=dev

apps/dashboard/src/pages/integrations/components/integration-card.tsx
scopsy: Going to address this in the second PR to avoid merge conflict from hell

apps/dashboard/src/pages/integrations/components/integration-card.tsx
twentyone24: my two cents:
- It doesn't look like a clickable button at glance, alongside the env pill. 
- since the whole card is clickable and does the same action to what clicking on configure does, I'd suggest to have a warning Icon to the the top-right instead. so, this adds an emphasis of action required / missing, and urge to configure. 

WDYT?

scopsy: @twentyone24 I'm not sure if it should be a warning. It's more of a Get Started state.

LetItRock (approved):
awesome! 🙌

PR #7318, opened 2024-12-17
feat(dashboard): Nv 4884 push mini preview

BiswaViraj:
### Screenshots

https://github.com/user-attachments/assets/bb0a321b-0d5d-4863-8d1c-fae5b1efc4ce

<details>

</details>


apps/dashboard/src/components/workflow-editor/steps/push/push-preview.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <PushContentContainerPreview className="relative z-10">
+            <PushSubjectPreview isPending />
+            <PushBodyPreview isPending />
+          </PushContentContainerPreview>
+        </PushNotificationContainer>
+      </PushBackgroundWithPhone>
+    );
+  }
+
+  if (previewData?.result.type !== ChannelTypeEnum.PUSH) {
```
SokratisVidros, line 27: Don't we already have guarantees in parent components that this push-preview will only be rendered when the channel type is a PUSH?

If yes, we can remove this conditional.

SokratisVidros (approved):

apps/dashboard/src/components/workflow-editor/steps/push/push-preview.tsx
BiswaViraj: Yes, but this also helps TS narrowing the types

ChmaraX (approved):
I've put the phone used in SMS to `public/phones/iphone-sms.svg` 

maybe we could follow same pattern e.g. `iphone-notification.svg`

apps/dashboard/src/components/workflow-editor/steps/push/push-preview.tsx
SokratisVidros: hahaha Ok, not a good enough reason but not a blocker as well. If we change the preview DTO to a much flatter alternative it would be much easier for us to fix this issue.

PR #7322, opened 2024-12-18
feat(dashboard): digest aggregated by key field editor with autosuggestions

LetItRock:

Digest - aggregated by key field support the editor with autosuggestions for the variables.

### Screenshots

https://github.com/user-attachments/assets/73411320-eb6f-4f42-a725-0d6f40c7d498


ChmaraX (approved):

BiswaViraj (approved):
Nice 👌🏼 

Not a blocker: We may need to add a [custom scrollbar for the editor](https://codemirror.net/5/demo/simplescrollbars.html). the current one looks a bit clunky 😅

scopsy:
@LetItRock do we need to have the {{curly}} braces for it in the design? I was expecting not to have it and select the key itself. I'm concerned it's confusing and people won't understand the {{ are needed. cc. @twentyone24

SokratisVidros (approved):

apps/dashboard/src/components/workflow-editor/steps/digest/digest-key.tsx
```diff
@@ -22,18 +30,21 @@ export const DigestKey = () => {
...
                 <Code2 className="-ml-1.5 size-5" />
-                <span className="text-foreground-600 mt-[1px] text-xs font-normal">subscriberId</span>
+                <span className="text-foreground-600 text-xs font-normal">subscriberId</span>
               </FormLabel>
               <FormControl>
-                <Input
-                  className={cn(
-                    'min-w-[20ch] [appearance:textfield] [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-outer-spin-button]:appearance-none'
-                  )}
+                <Editor
```
SokratisVidros, line 39: Note that as part of future work, we need to introduce a single line editor component that leverages the `<Editor/>` but behaves like an input. That is, on Enter press no new limes should be added, and on tab press we should move to the next input in the form instead of applying indentation to the current line of text.

LetItRock:
> @LetItRock do we need to have the {{curly}} braces for it in the design? I was expecting not to have it and select the key itself. I'm concerned it's confusing and people won't understand the {{ are needed. cc. @twentyone24

@scopsy This is our approach in "almost" all editor fields; the field is free text plus variables. I know this one is a bit different, but I don't remember if we allow having multiple keys separated by a comma (like we had in v1)

PR #7329, opened 2024-12-18
feat(dashboard,api): New maily variables function usage & update maily

desiprisg:
### Screenshots

<details>

</details>


apps/dashboard/src/utils/parseStepVariablesToLiquidVariables.ts
```diff
@@ -5,38 +5,62 @@ export interface LiquidVariable {
...
 
   function extractProperties(obj: JSONSchemaDefinition, path = ''): void {
-    if (typeof obj === 'boolean') return; // Handle boolean schema
+    if (typeof obj === 'boolean') return;
+
+    if (obj.type === 'object') {
+      // Handle object with additionalProperties
+      if (obj.additionalProperties === true) {
+        result.namespaces.push({
+          type: 'variable',
```
SokratisVidros, line 33: Since we split variables by type into three arrays, is the type: 'variable' necessary? If the arrays contain just a string of labels, then it will be much simpler to work with them in the components and we won't need the useMemo.

apps/dashboard/src/components/workflow-editor/steps/email/maily.tsx
```diff
@@ -74,11 +113,37 @@ export const Maily = (props: MailyProps) => {
...
     />
   );
 };
+
+const dedupAndSortVariables = (
+  variables: { name: string; required: boolean }[],
+  query: string
+): { name: string; required: boolean }[] => {
+  // Filter variables that match the query
+  const filteredVariables = variables.filter((variable) => variable.name.toLowerCase().includes(query.toLowerCase()));
```
SokratisVidros, line 130: To deduplicate: 

```suggestion
  const filteredVariables = new Set(variables.filter((variable) => variable.name.toLowerCase().includes(query.toLowerCase())));
```

apps/dashboard/src/components/workflow-editor/steps/email/maily.tsx
```diff
@@ -74,11 +113,37 @@ export const Maily = (props: MailyProps) => {
...
+
+  // Deduplicate based on name property
+  const uniqueVariables = Array.from(new Map(filteredVariables.map((item) => [item.name, item])).values());
+
+  // Sort variables: exact matches first, then starts with query, then alphabetically
+  return uniqueVariables.sort((a, b) => {
```
SokratisVidros, line 136: Is this algorithm the same one that Maily used internally? I am curious if we need a more sophisticated string comparison/fuzzy search library to match query with results like https://www.npmjs.com/package/bloodhound-js.

SokratisVidros (approved):

apps/dashboard/src/utils/parseStepVariablesToLiquidVariables.ts
desiprisg: Besides maily, we need to pass them like this in codemirror.

apps/dashboard/src/components/workflow-editor/steps/email/maily.tsx
desiprisg: looks good to me as is, but we can improve. I didn't look at maily internals.

PR #7333, opened 2024-12-19
fix(api): add additional properties true in case schema is computed

tatarco:
### Screenshots

<details>

</details>


scopsy (approved):

SokratisVidros:
We tested this again. For new workflows, it is working as expected. The issue with missing `additionalProperties` appears only in workflows with invalid persisted payloadSchema in Mongo.

PR #7334, opened 2024-12-19
fix(api-service): digest schema - remove the schema defaults as it doesn't work with the framework ajv validation

LetItRock:

Removed the defaults from the digest control values schema as it doesn't work with the framework AJV schema validation.

### Screenshots

https://github.com/user-attachments/assets/7926942f-a58c-40a0-a77a-b576c7eaf753


apps/api/src/app/environments-v1/usecases/output-renderers/digest-output-renderer.usecase.ts
```diff
@@ -14,22 +14,15 @@ export class DigestOutputRendererUsecase {
   @InstrumentUsecase()
   execute(renderCommand: RenderCommand): DigestRenderOutput {
     const parse: DigestControlSchemaType = DigestControlZodSchema.parse(renderCommand.controlValues);
-    if (
-      isDigestRegularControl(parse) &&
-      parse.amount &&
-      parse.unit &&
-      parse.lookBackWindow &&
-      parse.lookBackWindow.amount &&
-      parse.lookBackWindow.unit
```
LetItRock, line 23: the `lookBackWindow` is an optional field in the framework

apps/api/src/app/workflows-v2/shared/schemas/digest-control.schema.ts
```diff
@@ -14,12 +14,12 @@ const DigestRegularControlZodSchema = z
   .object({
     amount: z.union([z.number().min(1), z.string().min(1)]),
     skip: skipControl.schema,
-    unit: z.nativeEnum(TimeUnitEnum).default(TimeUnitEnum.SECONDS),
```
LetItRock, line 17: Removed the `default` here as AJV strict validation is breaking with it. We will need to rethink that when will be working on default values refactor.

SokratisVidros:
@LetItRock What doesn't work with these default values? Is it their format, an AJV error or something else? I watched the screencast but it's not very clear.

packages/shared/src/dto/workflows/preview-step-response.dto.ts
```diff
@@ -43,7 +43,7 @@ class DigestRegularOutput {
   amount: number;
   unit: TimeUnitEnum;
   digestKey?: string;
-  lookBackWindow: {
+  lookBackWindow?: {
```
djabarovgeorge, line 46: 👏

apps/api/src/app/workflows-v2/shared/schemas/digest-control.schema.ts
djabarovgeorge: I'm not sure that i understand the error and its fix, can you share the error you got here please?

LetItRock:
> @LetItRock What doesn't work with these default values? Is it their format, an AJV error or something else? I watched the screencast but it's not very clear.

@SokratisVidros there should be an error example in the ticket. Somehow the AJV in strict mode doesnt like the default values that were on JsonSchema.

SokratisVidros:
> > @LetItRock What doesn't work with these default values? Is it their format, an AJV error or something else? I watched the screencast but it's not very clear.
> 
> @SokratisVidros there should be an error example in the ticket. Somehow the AJV in strict mode doesnt like the default values that were on JsonSchema.

Isn't that unusual? Is there maybe another underlying root cause and we need to investigate deeper?

LetItRock:
> > > @LetItRock What doesn't work with these default values? Is it their format, an AJV error or something else? I watched the screencast but it's not very clear.
> > 
> > 
> > @SokratisVidros there should be an error example in the ticket. Somehow the AJV in strict mode doesnt like the default values that were on JsonSchema.
> 
> Isn't that unusual? Is there maybe another underlying root cause and we need to investigate deeper?

@SokratisVidros yes, it is, and I thought that we could investigate when we will be working on the default values for all steps of the schema

SokratisVidros (approved):

PR #7335, opened 2024-12-19
fix(dashboard): fixed view execution logs button triggering the workflow

LetItRock:

Workflow Trigger - Activity Log:
Fixed the issue with the `View Execution` button that was triggering the workflow. 

### Screenshots

https://github.com/user-attachments/assets/d569ed74-ea5e-4990-a8eb-80114e7d1070


apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-logs-sidebar.tsx
```diff
@@ -59,7 +51,7 @@ export const TestWorkflowLogsSidebar = ({ transactionId }: TestWorkflowLogsSideb
       ) : activityId ? (
         <ActivityPanel
           activityId={activityId}
-          onActivitySelect={() => {}}
```
LetItRock, line 68: This was not implemented at all and the click on View Execution was triggering the workflow.

apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-logs-sidebar.tsx
```diff
@@ -1,52 +1,44 @@
...
-    // Only stop refetching if we have an activity and it's not pending
-    setShouldRefetch(isPending);
-
-    queryClient.invalidateQueries({
-      queryKey: [QueryKeys.fetchActivity, currentEnvironment?._id, activity._id],
-    });
-  }, [activities]);
+    if (activityId) {
+      setShouldRefetch(false);
+    }
```
LetItRock, line 29: Refetch the activities (notifications) until we have some by the transactionId. We need to refetch because the worker doesn't create them right away, but processes after some time depending on where the jobs are in the queue.

apps/dashboard/src/components/activity/activity-panel.tsx
```diff
@@ -21,7 +26,28 @@ export function ActivityPanel({
...
+
+    const isPending = activity.jobs?.some((job) => job.status === JobStatusEnum.PENDING);
+
+    // Only stop refetching if we have an activity and it's not pending
+    setShouldRefetch(isPending);
+
+    queryClient.invalidateQueries({
+      queryKey: [QueryKeys.fetchActivity, currentEnvironment?._id, activityId],
+    });
+  }, [activity, queryClient, currentEnvironment, activityId]);
```
LetItRock: refetch the `activity` details until we have any pending jobs

apps/dashboard/src/components/activity/activity-panel.tsx
```diff
@@ -77,7 +103,10 @@ export function ActivityPanel({
               ctaClassName="text-foreground-950"
               variant={'tip'}
               ctaLabel="View Execution"
-              onCtaClick={() => {
+              onCtaClick={(e) => {
+                e.stopPropagation();
+                e.preventDefault();
```
LetItRock, line 114: don't propagate click event

apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-logs-sidebar.tsx
scopsy: Good catch, missed this during the smaller PR's split

apps/dashboard/src/components/activity/activity-panel.tsx
scopsy: @LetItRock I've faced this issue already on other multiple product areas. Wdyt if we change the primitive Button component to have type="button", so it won't be treated as type submit when wrapped in a form?

apps/dashboard/src/components/workflow-editor/test-workflow/test-workflow-logs-sidebar.tsx
scopsy: @LetItRock what if we just set the activityId when onActivity select happens? Because we want to override it here? And then not fetching logic needs to happen here on the test-workflow-logs-sidebar as you moved all of the logic to the activity panel itself.

scopsy (approved):
Looking good. Did you had a chance to also test the activity feed page for regressions?

LetItRock:
> Looking good. Did you had a chance to also test the activity feed page for regressions?

sorry for the late response, but yes it's working well :)

Issue #6346, opened 2024-08-17
🐛 Bug Report: @novu/framework cannot be used in Cloudflare Pages


fell-lucas:
### 📜 Description

We get build errors when deploying to CF pages because the package imports unsupported node dependencies.

Even though we are not using the JSON Schema features of @novu/framework, we still get these import errors. Is it possible to ask for conditional importing of the problematic dependencies?

### 👟 Reproduction steps

1. Install @novu/framework on a SvelteKit project using @sveltejs/cloudflare-adapter.
2. Create a workflow using the `workflow` import from @novu/framework.
3. Create `api/novu/+server.ts` and import the `serve` function from @novu/framework/sveltekit
4. Run build script / build project

### 👍 Expected behavior

Not throw build errors because we're not relying on JSON Schema to create our workflows.

### 👎 Actual Behavior with Screenshots

<img width="1120" alt="image" src="https://github.com/user-attachments/assets/529bc3ca-38b7-4ecb-8db4-852c3ca2526d">

### Novu version

Novu SaaS

### npm version

NA

### node version

NA

### 📃 Provide any additional context for the Bug.

@novu/framework version: 2.0.0-canary.5

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find a similar issue

### 🏢 Have you read the Contributing Guidelines?

- [X] I have read the [Contributing Guidelines](https://github.com/novuhq/novu/blob/main/CONTRIBUTING.md)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

rifont:
Thanks for reporting this. The `require`d modules shown in your screenshot are not critical runtime dependencies of Framework and we should be able to make those dependencies conditional.

We plan to investigate conditional importing for these dependencies within the next 2 weeks.

jkhaui:
> Thanks for reporting this. The `require`d modules shown in your screenshot are not critical runtime dependencies of Framework and we should be able to make those dependencies conditional.
> 
> We plan to investigate conditional importing for these dependencies within the next 2 weeks.

Hey, just wondering if there's any update?

CF Pages is becoming an increasingly popular deployment platform :)

SokratisVidros:
We are actively working on a fix for this, aiming to ship it by next week.

SokratisVidros:
[...]

PR #7337, opened 2024-12-19
fix(framework): Remove @novu/shared dependency temporarily

SokratisVidros:

This PR copied all the `@novu/shared` code directly into the @novu/framework package to eliminate a dependency that is causing ESM/CJS issues on `@novu/framework`.

This is a temporary patch until we refactor @novu/shared to use strict TS and support ESM/CJS.

It should address:

-[] https://github.com/novuhq/novu/issues/6346


packages/framework/src/shared.ts
```diff
@@ -0,0 +1,252 @@
+/**
+ * ==========
+ * | NOTICE |
+ * ==========
+ *
+ * This file contains copied code from @novu/shared in order to temporarily eliminate the dependency of
+ * framework on the shared package.
+ *
+ * The shared package, doesn't support ESM/CJS with strict TS yet.
+ * So we sacrified a bit code duplication in order to address ESM/CJS issues reported on the @novu/framework
```
scopsy:
```suggestion
 * So we sacrificed a bit code duplication in order to address ESM/CJS issues reported on the @novu/framework
```

scopsy (approved):

PR #7343, opened 2024-12-20
chore(dashboard): scheduled digest descope the yearly option

LetItRock:

Scheduled Digest - descope the yearly option according to [this thread](https://novu.slack.com/archives/C07HVLG7G2U/p1734700090045209?thread_ts=1734699397.524019&cid=C07HVLG7G2U)

### Screenshots
Screenshot 2024-12-20 at 17 51 33

Screenshot 2024-12-20 at 17 51 38


apps/dashboard/src/components/workflow-editor/steps/digest/period.tsx
```diff
@@ -8,7 +8,8 @@ const PERIOD_OPTIONS = [
   { value: PeriodValues.DAY, label: 'day' },
   { value: PeriodValues.WEEK, label: 'week' },
   { value: PeriodValues.MONTH, label: 'month' },
-  { value: PeriodValues.YEAR, label: 'year' },
+  // disable the year option for now as it might affect Redis memory and performance
+  // { value: PeriodValues.YEAR, label: 'year' },
```
djabarovgeorge: I think we can remove it entirely. There's no need to keep those lines for now.
we can revisit and add support when needed.

djabarovgeorge (approved):
🚀

apps/dashboard/src/components/workflow-editor/steps/digest/period.tsx
LetItRock: if we want to remove it, there is more code... but it might be hard to bring it back after some time, so maybe let's leave it for now

SokratisVidros: Let's remove it completely as an option for now from the UI and the API. Detailed specs about the max delay and digest duration will arrive soon.

PR #7344, opened 2024-12-21
refactor(api-service): issues management

djabarovgeorge:

Issue Management:
- Responsibility for issues has been shifted to schemas and LiquidJS instead of custom code , that make us aligned with the preview flow.

https://www.loom.com/share/710a5a211e4a4cbaa236fbbcf882008a?provider=slack-openid-connect

### Screenshots

<details>

</details>


apps/api/src/app/workflows-v2/usecases/patch-workflow/patch-workflow.usecase.ts
```diff
@@ -11,18 +10,12 @@ export class PatchWorkflowUsecase {
...
 
   async execute(command: PatchWorkflowCommand): Promise<WorkflowResponseDto> {
     const persistedWorkflow = await this.fetchWorkflow(command);
-    let transientWorkflow = this.patchWorkflowFields(persistedWorkflow, command);
-
-    transientWorkflow = await this.postProcessWorkflowUpdate.execute({
-      workflow: transientWorkflow,
-      user: command.user,
-    });
+    const transientWorkflow = this.patchWorkflowFields(persistedWorkflow, command);
```
djabarovgeorge, line 18: what is the status of using the patch? do we want to keep it??
this change means that we won't update issues on patch.
if we need it i can recreate the issue generation on patch as well.

apps/api/src/app/environments-v1/usecases/output-renderers/hydrate-email-schema.usecase.ts
```diff
@@ -234,3 +233,5 @@ export const TipTapSchema = z.object({
   text: z.string().optional(),
   attrs: z.record(z.unknown()).optional(),
 });
+
+const buildLiquidJSDefault = (mailyFallback: string) => (mailyFallback ? ` | default: '${mailyFallback}'` : '');
```
SokratisVidros: Consider combining this with `const finalValue = `{{ ${variableName} ${buildLiquidJSDefault(fallback)} }}`;` under a `buildLiquidJSVariable`.

apps/api/src/app/environments-v1/usecases/output-renderers/render-email-output.usecase.ts
```diff
@@ -50,7 +56,19 @@ export class RenderEmailOutputUsecase {
...
 }
+
+export const stringifyDataStructureWithSingleQuotes = (value: unknown, spaces: number = 0): string => {
+  if (Array.isArray(value) || (typeof value === 'object' && value !== null)) {
+    const valueStringified = JSON.stringify(value, null, spaces);
+    const valueSingleQuotes = valueStringified.replace(/"/g, "'");
+    const valueEscapedNewLines = valueSingleQuotes.replace(/\n/g, '\\n');
+
+    return valueEscapedNewLines;
+  } else {
```
SokratisVidros, line 71: `else` is not required.

```suggestion
  }
```

apps/api/src/app/environments-v1/usecases/output-renderers/render-email-output.usecase.ts
```diff
@@ -50,7 +56,19 @@ export class RenderEmailOutputUsecase {
   }
 
   @Instrument()
-  private transformForAndShowLogic(body: string, fullPayloadForRender: FullPayloadForRender) {
+  private transformMailyDynamicBlocks(body: string, fullPayloadForRender: FullPayloadForRender) {
     return this.expandEmailEditorSchemaUseCase.execute({ emailEditorJson: body, fullPayloadForRender });
   }
 }
+
+export const stringifyDataStructureWithSingleQuotes = (value: unknown, spaces: number = 0): string => {
```
SokratisVidros, line 64: @djabarovgeorge can you please elaborate with an example? Why do we need this extra handling on single quotes?

apps/api/src/app/workflows-v2/shared/step-type-to-control.mapper.ts
```diff
@@ -1,11 +1,12 @@
 import { ActionStepEnum, ChannelStepEnum } from '@novu/framework/internal';
```
SokratisVidros, line 1: Nit-picky: This file could be replaced by a mapper returned from the `./schemas/index.ts` to avoid introducing a new filename, mapper.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -77,58 +61,46 @@ export class GeneratePreviewUsecase {
...
+        previewTemplateData.controlValues
       );
 
       return {
         result: {
           preview: executeOutput.outputs as any,
           type: stepData.type as unknown as ChannelTypeEnum,
         },
-        previewPayloadExample: finalVariablesExample,
+        previewPayloadExample: mergedVariablesExample,
```
SokratisVidros, line 103: As per lexicon, these are the [preview control data](https://www.notion.so/novuhq/Product-Terminology-Lexicon-d6ef923953744bc1b967ba8f297dee7f?pvs=4#15ad03ad838f80cbb958e9906eddee46). Let's add a TODO for the renaming in a follow up PR.

I also have a task in my list, to remove the nesting added by the `result` attribute from this endpoint.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -293,12 +245,12 @@ export class GeneratePreviewUsecase {
...
 
       for (const invalidVariable of invalidVariables) {
-        if (!controlValuesString.includes(invalidVariable.template)) {
+        if (!controlValuesString.includes(invalidVariable.output)) {
           continue;
         }
 
         const EMPTY_STRING = '';
-        controlValuesString = replaceAll(controlValuesString, invalidVariable.template, EMPTY_STRING);
+        controlValuesString = replaceAll(controlValuesString, invalidVariable.output, EMPTY_STRING);
```
SokratisVidros, line 253: Nice!

apps/api/src/app/workflows-v2/usecases/patch-workflow/patch-workflow.usecase.ts
SokratisVidros: For now, patch should only toggle the active status of a workflow. So you can safely remove everything else and simplify the PATCH usecase. See [here](https://www.notion.so/novuhq/Workflow-V2-API-159d03ad838f8065a49feaa44a7b3efe?pvs=4#159d03ad838f802d885df64ebd5b5869).

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -148,45 +144,68 @@ export class UpsertWorkflowUseCase {
...
       userPreferences: command.workflowDto.preferences?.user ?? null,
       defaultPreferences: command.workflowDto.preferences?.workflow ?? DEFAULT_WORKFLOW_PREFERENCES,
       triggerIdentifier: slugify(workflowDto.name),
+      status: this.computeStatus(command.workflowDto, steps),
     };
   }
 
-  private convertCreateToUpdateCommand(
+  private computeStatus(workflow: UpsertWorkflowDataCommand, steps: NotificationStep[]) {
+    const hasIssues = steps.filter((step) => this.hasControlIssues(step.issues)).length > 0;
```
SokratisVidros: Nit picky one, please compute `hasIssues`, after checking on the active flag to avoid unnecessary computations for inactive workflows.

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/patch-workflow/patch-workflow.usecase.ts
djabarovgeorge: amazing, thanks for the ref as well!

apps/api/src/app/environments-v1/usecases/output-renderers/render-email-output.usecase.ts
djabarovgeorge: A few reasons:  
1. Alignment with the Liquid.js client used in the framework.  
2. I noticed that when a variable is missing, the framework currently parses it as `unknown` (we have it today in the framework). This option causes this behavior. I believe it's an issue, but I couldn't fix it quickly, so I suggest we address it separately.  
For now, we've ensured consistent behavior across channel steps.

apps/api/src/app/workflows-v2/shared/step-type-to-control.mapper.ts
djabarovgeorge: I like the idea and have used this pattern myself, but I noticed that not everyone on the team is aligned with it. I suggest introducing it to the team first.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: agree :)

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: done! :)

PR #7345, opened 2024-12-22
feat(dashboard): changelog news widget

scopsy:
- Added a new stacked changelog feed
- Cached for 4 hours in memory
- Conditionally shows the trial card UI 
- Only show the card if free trial is not visible

https://github.com/user-attachments/assets/8ae4120b-5346-423c-929b-77663edb31a7

<details>

</details>


apps/dashboard/src/components/side-navigation/changelog-cards.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const track = useTelemetry();
+
+  const fetchChangelogs = async (): Promise<Changelog[]> => {
+    const cachedData = storage.getCachedData();
+    if (cachedData) {
+      return filterChangelogs(cachedData, storage.getDismissedChangelogs());
+    }
+
+    const response = await fetch(CONSTANTS.CHANGELOG_API_URL);
+    const rawData = await response.json();
```
djabarovgeorge: do we have any interface of this data?

apps/dashboard/src/components/side-navigation/changelog-cards.tsx
```diff
@@ -0,0 +1,207 @@
...
+          />
+        ))}
+      </div>
+    </div>
+  );
+}
+
+const storage = {
+  getDismissedChangelogs: (): string[] => {
+    const dismissed = localStorage.getItem(CONSTANTS.DISMISSED_STORAGE_KEY);
```
djabarovgeorge: i think we started to use Clerk context for such cases maybe we can utilize it in here as well

apps/dashboard/src/components/side-navigation/changelog-cards.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const fetchChangelogs = async (): Promise<Changelog[]> => {
+    const cachedData = storage.getCachedData();
+    if (cachedData) {
+      return filterChangelogs(cachedData, storage.getDismissedChangelogs());
+    }
+
+    const response = await fetch(CONSTANTS.CHANGELOG_API_URL);
+    const rawData = await response.json();
+
+    storage.cacheData(rawData);
```
djabarovgeorge: why do we need to cache data if we use react query here?

djabarovgeorge (approved):
looks amazing :)

apps/dashboard/src/components/side-navigation/changelog-cards.tsx
scopsy: I want to cache it between reloads, so we don't always make api calls to product lane on each page load as this info rarely changes. But, good point

PR #7347, opened 2024-12-22
feat(api): fail inactive notification trigger

djabarovgeorge:

We added issue management under the workflow entity, but we never enforced its usage. This PR ensures that notifications won't be triggered for statuses that are not active.

At the same time, to improve the user experience, we're allowing users to trigger the "Trigger" action directly from the Dashboard. This helps them quickly identify issues while debugging the workflow.

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
+++ b/apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
@@ -28,7 +28,9 @@ import {
   ReservedVariablesMap,
   TriggerContextTypeEnum,
   TriggerEventStatusEnum,
+  TriggerSourceEnum,
   WorkflowOriginEnum,
+  WorkflowStatusEnum,
 } from '@novu/shared';
 import {
   EnvironmentEntity,
@@ -139,6 +141,20 @@ export class ParseEventRequest {
       const message = workflowOverride ? 'Workflow is not active by workflow override' : 'Workflow is not active';
       Logger.log(message, LOG_CONTEXT);
 
+      return {
+        acknowledged: true,
+        status: TriggerEventStatusEnum.NOT_ACTIVE,
+      };
+    }
+
+    if (template.status === WorkflowStatusEnum.INACTIVE && !this.isGracefulTriggerAllowed(command.payload.__source)) {
+      return {
+        acknowledged: true,
+        status: TriggerEventStatusEnum.NOT_ACTIVE,
+      };
+    }
+
+    if (template.status === WorkflowStatusEnum.ERROR && !this.isGracefulTriggerAllowed(command.payload.__source)) {
       return {
         acknowledged: true,
         status: TriggerEventStatusEnum.NOT_ACTIVE,
@@ -404,6 +420,10 @@ export class ParseEventRequest {
 
   private isBlockedEmail(email: string): boolean {
     return BLOCKED_DOMAINS.some((domain) => email.includes(domain));
+  }
+
+  private isGracefulTriggerAllowed(source?: string): boolean {
+    return source === TriggerSourceEnum.DASHBOARD;
   }
 }
 
--- a/apps/api/src/app/events/e2e/trigger-event.e2e.ts
+++ b/apps/api/src/app/events/e2e/trigger-event.e2e.ts
@@ -37,9 +37,17 @@ import {
   WorkflowCreationSourceEnum,
   WorkflowResponseDto,
   ExecutionDetailsStatusEnum,
+  WorkflowStatusEnum,
+  TriggerSourceEnum,
 } from '@novu/shared';
 import { EmailEventStatusEnum } from '@novu/stateless';
-import { DetailEnum } from '@novu/application-generic';
+import {
+  buildNotificationTemplateIdentifierKey,
+  CacheInMemoryProviderService,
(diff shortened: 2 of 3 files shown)


apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -405,6 +421,10 @@ export class ParseEventRequest {
   private isBlockedEmail(email: string): boolean {
     return BLOCKED_DOMAINS.some((domain) => email.includes(domain));
   }
+
+  private isGracefulTriggerAllowed(source?: string): boolean {
+    return source === TriggerSourceEnum.DASHBOARD;
```
SokratisVidros, line 426: @djabarovgeorge, why do we need to check the source for this? I suggest just using the status for now.

SokratisVidros (approved):

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
djabarovgeorge: please see the description of the PR 

> At the same time, to improve the user experience, we're allowing users to trigger the "Trigger" action directly from the Dashboard. This helps them quickly identify issues while debugging the workflow.

djabarovgeorge: Let me know what you think, i am fine removing it because now the trigger is less predicted, but in the same time it will cost workflow debugging flow.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
```diff
@@ -145,6 +147,20 @@ export class ParseEventRequest {
...
+      return {
+        acknowledged: true,
+        status: TriggerEventStatusEnum.NOT_ACTIVE,
+      };
+    }
+
+    if (template.status === WorkflowStatusEnum.ERROR && !this.isGracefulTriggerAllowed(command.payload.__source)) {
+      return {
+        acknowledged: true,
+        status: TriggerEventStatusEnum.NOT_ACTIVE,
```
scopsy, line 160: Can we report a better status here? And potentially also pass the actual errors in the response? I would also actually throw a 4XX error here instead of returning 200

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
scopsy: I tend to agree with sok, I think trigger should be predictable and throw an error when in a bad step, not the type of things you want to learn about in prod

SokratisVidros: Following on this, @djabarovgeorge is right. Without a more sophisticated way to test an inactive workflow from the Dashboard we will need this hack in the logic to allow Novu users to test workflows from the Dashboard.

But this means we must ensure the API request comes from the Dashboard. That is, we should verify the source and the fact that a JWT (not an API key) was used for authentication.

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
SokratisVidros: The status code depends on how we handled the trigger endpoint. Returning 4XX can be a breaking change for some customers who don't handle errors when doing `novu.trigger`. So, for now, we should return whatever the API was returning.

scopsy: Didn't thought of the active one, but for the Error checks, I would still return an error

apps/api/src/app/events/usecases/parse-event-request/parse-event-request.usecase.ts
djabarovgeorge: How do we feel about adding additional indicators for this case?  
1. If the status is `ERROR`, we could either:  
   - Change the button color, or  
   - Add a warning icon to highlight the workflow status.  
2. If the status is `ERROR`, the payload could include an extra parameter to show the trigger override, for example:  
```
payload:
{
  "__debugModeBypassStatus": true,
  ...otherParams
}
```  
the API will check the __source and this new param __debugModeBypassStatus

LetItRock: I think that the API should behave consistently, doesn't matter if it's a direct HTTP, sdk, or dashboard call. If we want to simplify the life of Dashboard users, I think it will be enough if we show some kind of indication that the workflow has errors or is inactive in the trigger tab, but the API behavior should be consistent. Like I don't want to see different responses when triggered with curl as we suggest that option.

SokratisVidros: Let's continue our work at this [DX guide](https://www.notion.so/novuhq/1d4fb9bf6d2346b08b113b86ca8059ef?v=1b43f42e1e6045e98cb24428e9d23247&p=169d03ad838f8020b0bfd029607d1176&pm=s)

djabarovgeorge:
closing this one for now, need to rethink all the trigger flow.

PR #7351, opened 2024-12-23
fix(dashboard): Tweak arbitrary variable handling

SokratisVidros:

- A minor cleanup on empty JSON Schema
- Treat inline variables from payload or subscriber.data as not required in Maily

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
@@ -5,7 +5,7 @@ import { Instrument, InstrumentUsecase } from '@novu/application-generic';
 import { flattenObjectValues } from '../../util/utils';
 import { pathsToObject } from '../../util/path-to-object';
 import { extractLiquidTemplateVariables } from '../../util/template-parser/liquid-parser';
-import { convertJsonToSchemaWithDefaults } from '../../util/jsonToSchema';
+import { convertJsonToSchemaWithDefaults, emptyJsonSchema } from '../../util/jsonToSchema';
 import { BuildPayloadSchemaCommand } from './build-payload-schema.command';
 import { transformMailyContentToLiquid } from '../generate-preview/transform-maily-content-to-liquid';
 import { isStringTipTapNode } from '../../util/tip-tap.util';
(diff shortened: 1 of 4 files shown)


apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -19,20 +19,12 @@ export class BuildPayloadSchema {
     const controlValues = await this.buildControlValues(command);
 
     if (!controlValues.length) {
-      return {
-        type: 'object',
-        properties: {},
-        additionalProperties: true,
-      };
+      return emptyJsonSchema();
```
scopsy, line 22: 👏

apps/dashboard/src/components/workflow-editor/steps/email/maily.tsx
```diff
@@ -98,9 +98,15 @@ export const Maily = (props: MailyProps) => {
                       editor?.commands.deleteRange({ from, to });
                       editor?.commands.insertContent({
                         type: 'variable',
-                        attrs: { id: queryWithoutSuffix, label: null, fallback: null, showIfKey: null },
+                        attrs: {
+                          id: queryWithoutSuffix,
+                          label: null,
+                          fallback: null,
+                          showIfKey: null,
+                          required: false,
```
scopsy, line 106: question: What the required prop controls here?

scopsy (approved):

apps/dashboard/src/components/workflow-editor/steps/email/maily.tsx
SokratisVidros: The ⚠️ icon in Maily jargon. For now, all of our variables should not be required.

scopsy: But does is solve the actual problem then? Because the issue was happening when maily autocomplete returned "empty results" instead of allowing to add a new variable from the dropdown

SokratisVidros: A database migration solves the issue. I was torn between teaching the code to handle invalid payload vs just correcting the DB value and I did the latter as for code-first any user-defined payload schema should be accepted.

Issue #42, opened 2021-10-12
Add text field for IEmailOptions


ComBarnea:
- **I'm submitting a ...**
  [ ] bug report
  [X] feature request
  [ ] question about the decisions made in the repository
  [ ] question about how to use this project

- **Summary**
For fallback, it is required for some providers to supply simple text (not HTML), this field is just for that.

- **Other information** (e.g. detailed explanation, stack traces, related issues, suggestions how to fix, links for us to have context, eg. StackOverflow, personal fork, etc.) 
New field should be added to the interface called text

```ts
export interface IEmailOptions {
  to: string | string[];
  subject: string;
  html: string;
  from?: string;
  text?: string; // <- this is new
}
```

PR #7359, opened 2024-12-23
feat(api-service): refactor issue error messages

djabarovgeorge:

https://www.loom.com/share/91f0b186dea3470f987ff902ff9b8933

### Screenshots

<details>

</details>


apps/api/src/app/environments-v1/usecases/output-renderers/chat-output-renderer.usecase.ts
```diff
@@ -8,8 +7,8 @@ import { RenderCommand } from './render-command';
 export class ChatOutputRendererUsecase {
   @InstrumentUsecase()
   execute(renderCommand: RenderCommand): ChatRenderOutput {
-    const body = renderCommand.controlValues.body as string;
+    const { skip, ...outputControls } = renderCommand.controlValues ?? {};
 
-    return { body };
+    return outputControls as any;
```
djabarovgeorge, line 12: todo: we need to refactor all the renders to have output controls typed.

[issue just in case ](https://linear.app/novu/issue/NV-5118/api-make-the-outputs-typed)

libs/application-generic/src/schemas/control/in-app-control.schema.ts
```diff
@@ -0,0 +1,110 @@
...
+  url: z.string().regex(templateUrlPattern),
+  target: z
+    .enum(['_self', '_blank', '_parent', '_top', '_unfencedTop'])
+    .default('_blank'),
+});
+
+const actionZodSchema = z
+  .object({
+    label: z.string(),
+    redirect: redirectZodSchema,
```
djabarovgeorge: self note make redirect optional

scopsy (approved):

apps/api/src/app/environments-v1/usecases/output-renderers/digest-output-renderer.usecase.ts
```diff
@@ -2,38 +2,20 @@ import { Injectable, InternalServerErrorException } from '@nestjs/common';
...
-      };
-    }
-    if (isDigestTimedControl(parse) && parse.cron) {
-      return {
-        cron: parse.cron,
-        digestKey: parse.digestKey,
-      };
+    const { skip, ...outputControls } = renderCommand.controlValues ?? {};
+
+    if (outputControls.length === 0) {
```
SokratisVidros, line 12: Why is this check necessary for digest output but not for the other steps?

apps/api/src/app/workflows-v2/workflow.controller.ts
```diff
@@ -112,7 +113,7 @@ export class WorkflowController {
   ): Promise<WorkflowResponseDto> {
     return await this.upsertWorkflowUseCase.execute(
       UpsertWorkflowCommand.create({
-        workflowDto: updateWorkflowDto,
+        workflowDto: { ...updateWorkflowDto },
```
SokratisVidros, line 116: This object copy seems unnecessary.

apps/api/src/app/environments-v1/usecases/output-renderers/render-email-output.usecase.ts
```diff
@@ -19,17 +15,30 @@ export class RenderEmailOutputUsecase {
 
   @InstrumentUsecase()
   async execute(renderCommand: RenderEmailOutputCommand): Promise<EmailRenderOutput> {
-    const { body, subject } = emailControlZodSchema.parse(renderCommand.controlValues);
+    const { body, subject } = renderCommand.controlValues;
 
-    if (isEmpty(body)) {
-      return { subject, body: '' };
+    if (!body || typeof body !== 'string') {
```
SokratisVidros, line 20: Nit picky comment: Since tip tap processing takes place if the body exists, it is more readable to change the logic to 

let renderedBody = '';

``` if (body) {
  const expandedMailyContent = this.transformMailyDynamicBlocks(body, renderCommand.fullPayloadForRender);
  const parsedTipTap = await this.parseTipTapNodeByLiquid(expandedMailyContent, renderCommand);
  renderedBody = await this.renderEmail(parsedTipTap);
}

return { subject, renderedBody }
```

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -309,3 +321,106 @@ class FrameworkError {
   message: string;
   name: string;
 }
+
+function sanitizeControlValuesByOutputSchema(
+  controlValues: Record<string, unknown>,
+  type: StepTypeEnum
+): Record<string, unknown> {
+  const outputSchema = channelStepSchemas[type].output || actionStepSchemas[type].output;
```
SokratisVidros, line 329: 💯 Nice, reusing schema definitions from @novu/framework!

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -309,3 +321,106 @@ class FrameworkError {
...
+
+  if (!outputSchema || !controlValues) {
```
SokratisVidros, line 331: We should throw an error at this point. The inability to identify an `outputSchema` from the framework indicates an invalid workflow definition.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -309,3 +321,106 @@ class FrameworkError {
...
+    return controlValues;
+  }
+
+  const ajv = new Ajv({ allErrors: true });
+  addFormats(ajv);
+  const validate = ajv.compile(outputSchema);
+  const isValid = validate(controlValues);
+  const errors = validate.errors as null | ErrorObject[];
```
SokratisVidros, line 339: I tested it locally and TS doesn't complain if we do `const { errors } = validate;`

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -309,3 +321,106 @@ class FrameworkError {
...
+
+  if (isValid || !errors || errors?.length === 0) {
+    return controlValues;
+  }
+
+  return replaceInvalidControlValues(controlValues, errors);
```
SokratisVidros, line 345: @djabarovgeorge, did we try using the `useDefault` option from AJV so that AJV can make the replacement?

> With [option useDefaults](https://ajv.js.org/guide/options.html#options) Ajv will assign values from default keyword in the schemas of properties and items (when it is the array of schemas) to the missing properties and items.

The flow might be a bit different as AJV will first fill in the gaps with defaults and then do the validation, but I think it's more correct to do it that way than validating and assigning defaults later. The latter has the risk of assigning invalid defaults post-validation.

Note also this `in schemas inside anyOf, oneOf and not (see [#42 (opens new window)](https://github.com/ajv-validator/ajv/issues/42))`

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
```diff
@@ -309,3 +321,106 @@ class FrameworkError {
...
+    },
+  ],
+};
+
+/**
+ * Default control values used specifically for preview purposes.
+ * These values are designed to be parsable by Liquid.js and provide
+ * safe fallback values when generating preview.
+ */
+export const previewControlValueDefault = {
```
SokratisVidros, line 412: Following on the previous comment, these defaults should probably belong to the framework so that preview is more graceful in code-first workflows.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -111,28 +112,26 @@ export class UpsertWorkflowUseCase {
     if (existingWorkflow && isWorkflowUpdateDto(command.workflowDto, command.workflowIdOrInternalId)) {
       return await this.updateWorkflowGenericUsecase.execute(
         UpdateWorkflowCommand.create(
-          await this.convertCreateToUpdateCommand(command.workflowDto, command.user, existingWorkflow)
+          await this.buildUpdateWorkflowCommand(command.workflowDto, command.user, existingWorkflow)
```
SokratisVidros, line 115: 💅

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -491,29 +516,26 @@ function processControlValuesBySchema(
...
-  if (!controlSchema || !cleanedControlValues) {
+  if (!controlSchema || !controlValues) {
     return issues;
   }
 
   const ajv = new Ajv({ allErrors: true });
   addFormats(ajv);
   const validate = ajv.compile(controlSchema);
-  const isValid = validate(cleanedControlValues);
+  const isValid = validate(controlValues);
```
SokratisVidros, line 526: This validation logic is duplicated between upsert and preview. Should we consider consolidating it into a validate workflow use case?

In Upsert, we want this logic to generate 422 errors or issues. In Preview, we need this logic to fill in the invalid controls with defaults. 

It feels like both functionalities can be handled by a reusable use case that only handles the AJV part and returns errors that are handled accordingly per use case.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -553,6 +584,23 @@ function cleanObject(
   );
 }
 
+function convertEmptyStringsToNull(obj: Record<string, unknown> | undefined): Record<string, unknown> | undefined {
```
SokratisVidros, line 587: @djabarovgeorge can you please elaborate why is this necessary?

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -563,3 +611,19 @@ function mapAjvErrorToIssueType(error: ErrorObject): StepContentIssueEnum {
...
+  if (error.keyword === 'required') {
+    return `${_.capitalize(error.params.missingProperty)} is required`;
+  }
+  if (
+    error.keyword === 'pattern' &&
+    error.message?.includes('must match pattern') &&
+    error.message?.includes('mailto') &&
+    error.message?.includes('https')
+  ) {
+    return 'Invalid URL format. Must be a valid absolute URL, path, or valid variable';
```
SokratisVidros, line 625: 'Invalid URL format. Must be a valid absolute URL, path starting with /, or Novu {{variable}}';

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -162,7 +162,7 @@ function extractProps(template: any): { valid: boolean; props: string[]; error?:
    * Invalid: {{user.first name}} - postfix length would be 2 due to space
    */
   if (initial.postfix.length > 1) {
-    return { valid: false, props: [], error: 'Novu does not support variables with spaces' };
+    return { valid: false, props: [], error: 'Variables with spaces are not supported' };
```
SokratisVidros, line 165: Invalid variable name containing whitespaces. Variables must follow the [dot notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors#dot_notation)

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -181,7 +181,11 @@ function extractProps(template: any): { valid: boolean; props: string[]; error?:
    * Invalid: {{firstName}} - No namespace
    */
   if (validProps.length === 1) {
-    return { valid: false, props: [], error: 'Novu variables must include a namespace (e.g. user.firstName)' };
+    return {
+      valid: false,
+      props: [],
+      error: `Variables must include a namespace (e.g. payload.${validProps[0]})`,
```
SokratisVidros, line 187: Invalid variable name missing namespace. Variables must follow the [dot notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors#dot_notation) (e.g. payload.${validProps[0]}).

libs/application-generic/src/schemas/control/in-app-control.schema.ts
```diff
@@ -0,0 +1,109 @@
...
+ * 2. Full URLs that may contain template variables
+ *    - Excludes mailto: links
+ *    - Example: https://example.com, https://example.com/{{variable}}, https://{{variable}}.com
+ *
+ * 3. Paths starting with / that may contain template variables
+ *    - Example: /path/to/page, /path/{{variable}}/page
+ *
+ * Pattern is optimized to prevent exponential backtracking while maintaining all functionality
+ */
+const templateUrlPattern =
```
SokratisVidros, line 26: Consider renaming to `redirectUrlRegex`

libs/application-generic/src/schemas/control/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './in-app-control.schema';
```
SokratisVidros, line 1: These schemas are strictly connected to the built-in Dashboard UI powered by the new Novu workflow V2 API. Do they belong in the application generic? I don't feel strongly about it, but I'd consider it a code smell if these schemas were reused in other services.

apps/api/src/app/environments-v1/usecases/output-renderers/digest-output-renderer.usecase.ts
djabarovgeorge: not that i am aware of probably legacy code. i will remove it

apps/api/src/app/workflows-v2/workflow.controller.ts
djabarovgeorge: agree could be debugging leftovers.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: haha thanks, i thought it was an obvious move here ✨

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: agreed, but I tried to handle it as gracefully as possible. maybe we could log and return the controlValues instead? are you sure we need to throw an error here?

SokratisVidros: Yes. Without output schema, there is no step. So we need to fail fast.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: looks good, will do!

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: Currently, our codebase with zod-to-json-schema and ajv is very complex and unpredictable. For now, I’ve removed all defaults from the control schemas. I suggest scheduling a meeting to discuss and decide on a solution that meets all our requirements.

apps/api/src/app/workflows-v2/usecases/generate-preview/generate-preview.usecase.ts
djabarovgeorge: I'd argue that having defaults on the schema makes the flow more unpredictable. let's discuss about the best solution.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: agree :)
i will create [task](https://linear.app/novu/issue/NV-5129/api-refactor-ro-reusable-ajv-validator) for it.

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
djabarovgeorge: definitely! this is necessary because it allows us to alert the user when they've entered a `nullish` value in the Dashboard. Sending and storing an empty string counts as a valid required string, even though the user has "removed" the text.  
https://www.loom.com/share/6c100325b9a140f6b97fccb0b8fc7991

Another solution could be adding a UI helper component to indicate that the user deleted the value. This way, the Dashboard would send a `nullish` value instead of an empty string, preserving validation.

libs/application-generic/src/schemas/control/index.ts
djabarovgeorge: Interesting and valid point, i can share why they are in app-gen. 
The are in app-gen because i reused the types from the schema in the dashboardSanitizeControlValues function so i needed them close to it, and dashboardSanitizeControlValues is used in Worker.
Let me know what you think about such an implementation.

PR #7360, opened 2024-12-23
fix(api-service,dashboard): Crate of fixes for variable suggestions

SokratisVidros:

https://linear.app/novu/issue/NV-5104/adding-variable-on-the-fly-in-a-button-is-not-working
https://linear.app/novu/issue/NV-5105/when-trying-to-add-variables-on-fly-from-button-or-image-block-link

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
@@ -5,7 +5,6 @@ import { Instrument, InstrumentUsecase } from '@novu/application-generic';
 import { flattenObjectValues } from '../../util/utils';
 import { pathsToObject } from '../../util/path-to-object';
 import { extractLiquidTemplateVariables } from '../../util/template-parser/liquid-parser';
-import { convertJsonToSchemaWithDefaults, emptyJsonSchema } from '../../util/jsonToSchema';
 import { BuildPayloadSchemaCommand } from './build-payload-schema.command';
 import { transformMailyContentToLiquid } from '../generate-preview/transform-maily-content-to-liquid';
 import { isStringTipTapNode } from '../../util/tip-tap.util';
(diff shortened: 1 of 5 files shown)


scopsy (approved):

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -16,26 +15,13 @@ export class BuildPayloadSchema {
 
   @InstrumentUsecase()
   async execute(command: BuildPayloadSchemaCommand): Promise<JSONSchemaDto> {
-    const controlValues = await this.buildControlValues(command);
+    const controlValues = await this.getControlValues(command);
+    const sanitizedControlValues = await this.sanitizeControlValues(controlValues);
```
djabarovgeorge: i am not sure if the naming is correct here we do not sanitizeControlValues we create templateVariableNames and the function is returning not control value but string[] of variables

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -87,4 +73,30 @@ export class BuildPayloadSchema {
 
     return processedValue;
   }
+
+  private async buildVariablesSchema(variables: string[]) {
```
djabarovgeorge, line 77: i think we lost the case of

 if (!controlValues.length) {
    return this.buildVariablesSchema(sanitizedControlValues);
      return emptyJsonSchema();	
    }
    
    
   if i remember correctly this case was meant to cover the mini preview on step create, before we create the initial control values

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
```diff
@@ -87,4 +73,30 @@ export class BuildPayloadSchema {
...
+
+    const schema: JSONSchemaDto = {
+      type: 'object',
+      properties: {},
+      required: [],
+      additionalProperties: true,
+    };
+
+    if (variablesObject) {
+      for (const [key, value] of Object.entries(variablesObject)) {
```
djabarovgeorge, line 92: will this for loop cover nested payload as well, for example, `payload.bar.data.foo...` etc..?

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
SokratisVidros: If I am not mistaken, this should be covered as it as the functions can all handle empty variables.

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
SokratisVidros: Note, this is the reason e2e is failing, and I am pushing a fix.

apps/api/src/app/workflows-v2/usecases/build-payload-schema/build-payload-schema.usecase.ts
SokratisVidros: Fixed

PR #7365, opened 2024-12-24
chore(dashboard,web): Clean up feature flags

SokratisVidros:

Cleaning up feature flags


apps/web/src/ee/clerk/components/UserProfileButton.tsx
```diff
@@ -1,29 +1,24 @@
...
 export function UserProfileButton() {
   const { optIn } = useNewDashboardOptIn();
-  const isNewDashboardEnabled = useFeatureFlag(FeatureFlagsKeysEnum.IS_NEW_DASHBOARD_ENABLED);
 
   return (
     <UserButton
       afterSignOutUrl={`${WEB_APP_URL}${ROUTES.AUTH_LOGIN}`}
       userProfileUrl={ROUTES.MANAGE_ACCOUNT_USER_PROFILE}
     >
-      {isNewDashboardEnabled && (
```
scopsy, line 18: This will also show this to v1 users right? I'm not sure how their experience will be 🤔

scopsy (approved):

apps/web/src/ee/clerk/components/UserProfileButton.tsx
SokratisVidros: They should see the button, go to the new Dashboard, see all their workflows but for V1 workflows go back to the old web app on workflow selection.

I will verify it on staging.

PR #7366, opened 2024-12-24
fix(api-service): workflow creation api flow didn't generate payload schema

scopsy:
- Fixes the issue where API-generated workflows are lacking the payload schema, and it only works on upsert
- 

### Screenshots

<details>

</details>

### DIFF
--- a/apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
+++ b/apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
@@ -62,7 +62,7 @@ export class BuildAvailableVariableSchemaUsecase {
     workflow: NotificationTemplateEntity | undefined,
     command: BuildAvailableVariableSchemaCommand
   ): Promise<JSONSchemaDto> {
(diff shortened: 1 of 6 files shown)


apps/api/src/app/workflows-v2/usecases/build-variable-schema/build-available-variable-schema.usecase.ts
```diff
@@ -62,15 +62,15 @@ export class BuildAvailableVariableSchemaUsecase {
     workflow: NotificationTemplateEntity | undefined,
     command: BuildAvailableVariableSchemaCommand
   ): Promise<JSONSchemaDto> {
-    if (!workflow) {
+    if (workflow && !workflow?.steps.length) {
```
scopsy: This is what caused the issue, as the workflow didn't existed here yet.

SokratisVidros:
```suggestion
    if (workflow && workflow.steps && workflow.steps.length === 0) {
```

SokratisVidros (approved):

PR #7379, opened 2024-12-25
feat(dashboard): variable management for all editor fields

scopsy:
- Add support for variables pill in the FieldEditor component
- Replace all existing body and subject to use the field editor
- Introduce a variable popover
- Allow managing liquid syntax in a visual way
- Reordering liquid filters and managing filter params

### Screenshots
https://www.loom.com/share/cd1ec264b44249f88770d1ffe0d37727

CleanShot 2024-12-30 at 20 36 29@2x

<details>

</details>


apps/dashboard/public/images/code.svg
scopsy, line 1: Had to add it since I need to reference it from ::after in css

apps/dashboard/src/utils/liquid-autocomplete.ts
```diff
@@ -1,159 +1,114 @@
...
+        type: 'variable',
+      }));
+
+    if (payloadPath && !existingVariables.some((v) => v.label === `payload.${payloadPath}`)) {
+      suggestions.unshift({
+        label: `payload.${payloadPath}`,
+        type: 'variable',
+        boost: 99,
+      });
+    }
```
scopsy: Add support for adding new payload variables

apps/dashboard/src/components/workflow-editor/url-input.tsx
```diff
@@ -48,13 +44,12 @@ export const URLInput = ({
                 <FormItem className="min-w-px max-w-full basis-full">
                   <FormControl>
                     {asEditor ? (
-                      <Editor
-                        asInput
+                      <FieldEditor
                         fontFamily="inherit"
                         placeholder={placeholder}
-                        extensions={extensions}
```
scopsy, line 59: Extensions now live inside of the editor and not outside

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,171 @@
...
+  onChange: (value: string) => void;
+  variables: LiquidVariable[];
+  placeholder?: string;
+  autoFocus?: boolean;
+  size?: 'default' | 'lg';
+  fontFamily?: 'inherit';
+  id?: string;
+};
+
+export const FieldEditor = ({
```
scopsy: A new component that wraps the <Editor component with the needed behaviour for content management

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+  onChange: (value: string) => void;
+  variables: LiquidVariable[];
+  placeholder?: string;
+  autoFocus?: boolean;
+  size?: 'default' | 'lg';
+  fontFamily?: 'inherit';
+  id?: string;
+};
+
+export function FieldEditor({
```
scopsy: A new Field Editor component that uses the underlying Editor component and extends it for usage inside of content inputs

scopsy:
cc. @twentyone24

apps/dashboard/src/components/workflow-editor/steps/base/base-body.tsx
```diff
@@ -33,18 +22,18 @@ export const BaseBody = () => {
...
-                id={field.name}
-                extensions={extensions}
-                basicSetup={basicSetup}
-                ref={field.ref}
-                value={field.value}
-                onChange={field.onChange}
-                height="100%"
-              />
+            <InputField className="flex h-36 w-full items-start px-1">
+              <div className="w-full">
```
SokratisVidros: Is this div necessary since the InputField already contains a `w-full`?

apps/dashboard/src/components/workflow-editor/steps/base/base-body.tsx
```diff
@@ -33,18 +22,18 @@ export const BaseBody = () => {
...
+                <FieldEditor
+                  fontFamily="inherit"
```
SokratisVidros: This should be redundant.

apps/dashboard/src/components/workflow-editor/steps/email/email-subject.tsx
```diff
@@ -22,23 +18,20 @@ export const EmailSubject = () => {
...
-                onChange={(val) => field.onChange(val)}
-              />
-            </FormControl>
-            <FormMessage />
-          </FormItem>
-        </>
+        <FormItem className="w-full">
+          <FormControl>
+            <FieldEditor
+              size="lg"
```
SokratisVidros: Why do we use the `default` size in all other control fields except the email?

apps/dashboard/src/utils/liquid-autocomplete.ts
```diff
@@ -1,159 +1,114 @@
...
+}
+
+function getVariableCompletions(
+  word: { text: string; from: number; to?: number } | null,
+  pos: number,
+  variables: LiquidVariable[]
+): CompletionResult | null {
+  if (!word && !variables.length) return null;
+
+  if (word?.text.startsWith('payload.')) {
```
SokratisVidros: We need to add the same logic for `subscriber.data` and `digest.events[].payload`. They both allow arbitrary variables.

Subscriber.data should be supported by the API today. For digest, I need to test.

apps/dashboard/src/utils/liquid-autocomplete.ts
```diff
@@ -1,159 +1,114 @@
...
+    const payloadPath = word.text.slice(8);
+    const existingVariables = variables.filter((v) => v.label.startsWith('payload.'));
```
SokratisVidros: Is this logic here to support autocompleting arbitrary variables that are not synced with the API yet? So far, the logic is that existing variables are sent from the server inside each step.

So, we need to use the ones from the server and combine them with the "draft" variables on the client side.

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
+import { EditorView } from '@uiw/react-codemirror';
```
SokratisVidros: How about calling this `ControlField` or `ControlInput`? I prefer the latter as it is an input for Novu Controls. It maps better with the domain.

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+  size = 'default',
+  fontFamily = 'inherit',
+  id,
+}: FieldEditorProps) {
+  const viewRef = useRef<EditorView | null>(null);
+  const lastCompletionRef = useRef<CompletionRange>(null);
+
+  const { selectedVariable, setSelectedVariable, handleVariableSelect, isUpdatingRef } = useVariableSelection();
+
+  const handleVariableUpdate = useVariableUpdate(
```
SokratisVidros: I strongly suggest merging `useVariableSelection` and `useVariableSelection` into a `useVariables` hook. The hooks are strongly connected, as the output of the first one is the input of the second. Moreover, they share the same state using common ref objects.

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+    selectedVariable,
+    viewRef,
+    isUpdatingRef,
+    onChange,
+    setSelectedVariable
+  );
+
+  const completionSource = useCompletionSource(variables, lastCompletionRef);
```
SokratisVidros:
```suggestion
  const completionSource = useAutocomplete(variables, lastCompletionRef);
```

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+      <Popover
+        open={!!selectedVariable}
+        onOpenChange={(open) => {
+          if (!open) {
+            setTimeout(() => setSelectedVariable(null), 0);
+          }
+        }}
+      >
+        <PopoverTrigger asChild>
+          <div />
```
SokratisVidros: Is this empty trigger necessary?

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+ * 2. Prevents recursive updates when variables are being modified
+ */
+function useVariableSelection() {
+  const [selectedVariable, setSelectedVariable] = useState<SelectedVariable>(null);
+  const isUpdatingRef = useRef(false);
+
+  const handleVariableSelect = useCallback((value: string, from: number, to: number) => {
+    if (isUpdatingRef.current) return;
+
+    requestAnimationFrame(() => {
```
SokratisVidros: `requestAnimationFrame` needs a clean-up function on unmount, similar to setTimeout. See [here](https://css-tricks.com/using-requestanimationframe-with-react-hooks/)

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+    },
+    [selectedVariable, onChange, viewRef, isUpdatingRef, setSelectedVariable]
+  );
+}
+
+function useCompletionSource(variables: LiquidVariable[], lastCompletionRef: React.MutableRefObject<CompletionRange>) {
+  return useCallback(
+    (context: CompletionContext) => {
+      // Match text that starts with {{ and capture everything after it until the cursor position
+      const word = context.matchBefore(/\{\{([^}]*)/);
```
SokratisVidros: We should move all the parsing logic into the completions utility and simplify the hook.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const VARIABLE_REGEX = /{{([^{}]+)}}/g;
```
SokratisVidros: Nit-picky one: Conside moving this contents to the main file. Three constants don't justify another file that adds an overhead to the live reload and building process.

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+export function VariablePopover({ variable, onClose, onUpdate }: VariablePopoverProps) {
+  const searchInputRef = useRef<HTMLInputElement>(null);
+  const { parsedName, parsedDefaultValue, parsedTransformers } = useVariableParser(variable);
+  const [name, setName] = useState(parsedName);
+  const [defaultVal, setDefaultVal] = useState(parsedDefaultValue);
+  const [showRawLiquid, setShowRawLiquid] = useState(false);
+  const [searchQuery, setSearchQuery] = useState('');
+
+  useEffect(() => {
+    const handleEscape = (e: KeyboardEvent) => {
```
SokratisVidros: I am surprised we have to handle escape manually. Doesn't the default ShadCN popover close on ESC?

apps/dashboard/src/components/primitives/field-editor/variable-popover/components/draggable-transformer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  draggingItem: number | null;
+  dragOverIndex: number | null;
+  onDragStart: (index: number) => void;
+  onDragEnd: () => void;
+  onDrag: (e: any, info: any) => void;
+  onRemove: (value: string) => void;
+  onParamChange: (index: number, params: string[]) => void;
+};
+
+export function DraggableTransformer({
```
SokratisVidros: The draggable list should be a shared primitive as we will need to elsewhere.

SokratisVidros (changes requested):

SokratisVidros:
As I was testing this I noticed the following issues:

1. Multiple open variable autosuggestion at the same time Screenshot 2024-12-31 at 11 00 58
2. There are too many re-renderings on Codemirror, as we are again facing the lagging input we had in the past.
3. The handling of the Enter key is problematic. Hitting enter in single-line inputs such as in-app subjects is problematic and can also split variable pills in two.
4. Faced the following error while typing Screenshot 2024-12-31 at 11 09 47

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
LetItRock: I'm not sure that we need to request an animation frame. Do we?

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  const { selectedVariable, setSelectedVariable, handleVariableSelect, isUpdatingRef } = useVariableSelection();
+
+  const handleVariableUpdate = useVariableUpdate(
+    selectedVariable,
+    viewRef,
+    isUpdatingRef,
+    onChange,
+    setSelectedVariable
+  );
```
LetItRock: These two seem to be connected and can be one hook. We pass three args from `useVariableSelection` to `useVariableUpdate`.

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+          // If we're not already inside liquid brackets, wrap the completion with {{}}
+          // If we're inside liquid brackets (detected by checking previous chars), just insert the variable name
+          const content = view.state.doc.toString();
+          const before = content.slice(Math.max(0, from - 2), from);
+          const insert = before !== '{{' ? `{{${text}}} ` : `${text}}} `;
+
+          view.dispatch({
+            changes: { from, to, insert },
+          });
+        },
```
LetItRock: this function is never called because there is no apply function on the `CompletionResult`

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+ * 3. Maintains cursor position after updates
+ * 4. Prevents recursive updates during the edit process
+ *
+ * The update process:
+ * 1. Checks if the new value already has Liquid syntax
+ * 2. Detects if there are existing closing brackets after the cursor
+ * 3. Calculates the correct range to replace
+ * 4. Updates the editor state while maintaining proper cursor position
+ */
+function useVariableUpdate(
```
LetItRock: please move to a separate file (imo easier to read side by side)

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
```diff
@@ -0,0 +1,66 @@
+import { EditorView } from '@uiw/react-codemirror';
+
+export const variablePillTheme = EditorView.baseTheme({
+  '.cm-variable-pill': {
+    backgroundColor: 'hsl(var(--bg-weak))',
+    color: 'hsl(var(--text-sub))',
+    border: '1px solid hsl(var(--stroke-soft))',
+    borderRadius: '9999px',
+    padding: '2px 6px 2px 18px',
+    margin: '0 4px',
```
LetItRock: because of the margin:
- it's not aligned well when it is the first thing in the field, like there is a big gap on the left side
- weird cursor behaviour

here are examples:
Screenshot 2024-12-31 at 10 25 04

https://github.com/user-attachments/assets/cce7e015-ce3b-473b-a853-e02e6e17dc1b

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
```diff
@@ -0,0 +1,66 @@
+import { EditorView } from '@uiw/react-codemirror';
+
+export const variablePillTheme = EditorView.baseTheme({
+  '.cm-variable-pill': {
+    backgroundColor: 'hsl(var(--bg-weak))',
+    color: 'hsl(var(--text-sub))',
+    border: '1px solid hsl(var(--stroke-soft))',
+    borderRadius: '9999px',
```
LetItRock: 10px looks enough

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
```diff
@@ -0,0 +1,66 @@
+import { EditorView } from '@uiw/react-codemirror';
+
+export const variablePillTheme = EditorView.baseTheme({
+  '.cm-variable-pill': {
```
LetItRock: there should be equal gap between `before` and `after` (image and dot), which is missing rn, and the items seems to close

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
```diff
@@ -0,0 +1,66 @@
...
+  },
+  '.cm-variable-pill.has-modifiers': {
+    paddingRight: '12px',
+  },
+  '.cm-variable-pill::before': {
+    content: '""',
+    position: 'absolute',
+    left: '4px',
+    top: '50%',
+    transform: 'translateY(-50%)',
```
LetItRock: there is no need to render positioned absolutely, the parent is inline flex which positions `before - content - after` in a row

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
LetItRock: please use class constant `VARIABLE_PILL_CLASS`

apps/dashboard/src/components/primitives/field-editor/variable-plugin/plugin-view.ts
```diff
@@ -0,0 +1,79 @@
...
+    view: EditorView,
+    private viewRef: MutableRefObject<EditorView | null>,
+    private lastCompletionRef: MutableRefObject<{ from: number; to: number } | null>,
+    private onSelect?: (value: string, from: number, to: number) => void
+  ) {
+    this.decorations = this.createDecorations(view);
+    viewRef.current = view;
+  }
+
+  update(update: any) {
```
LetItRock: when the variable is picked from the autosuggestions dropdown the value does not change to pill

https://github.com/user-attachments/assets/47257c72-8145-435b-a806-5ab3fb30f26f

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-pill-widget.ts
```diff
@@ -0,0 +1,82 @@
...
+
+      // setTimeout is used to defer the selection until after CodeMirror's own click handling
+      // This prevents race conditions where our selection might be immediately cleared by the editor
+      setTimeout(() => {
+        this.onSelect?.(this.fullVariableName, this.start, this.end);
+      }, 0);
+    };
+
+    span.addEventListener('mousedown', handleClick);
+    (span as any)._variableClickHandler = handleClick;
```
LetItRock: can't we set it as a private field instead?

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+          </div>
+        </div>
+        <div className="grid gap-1.5">
+          <div className="grid gap-2 p-2">
+            <FormItem>
+              <FormControl>
+                <div className="grid gap-1">
+                  <label className="text-text-sub text-label-xs">Variable name</label>
+                  <InputField size="fit" className="min-h-0">
+                    <Code2 className="text-text-sub h-4 w-4" />
```
LetItRock: the icon is shrunk: 

Screenshot 2024-12-31 at 10 26 01

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+                      if (value) {
+                        handleTransformerToggle(value);
+                        setSearchQuery('');
+                      }
+                    }}
+                  >
+                    <SelectTrigger className="!text-paragraph-xs text-text-soft !h-7 min-h-0 !p-1.5">
+                      <SelectValue placeholder="Add a modifier..." />
+                    </SelectTrigger>
+                    <SelectContent
```
LetItRock: the content should be the side of the select input and look like a dropdown and not another popover

Screenshot 2024-12-31 at 13 19 53

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+                  <label className="text-text-sub text-label-xs">Modifiers</label>
+                  <Select
+                    value=""
+                    onValueChange={(value) => {
+                      if (value) {
+                        handleTransformerToggle(value);
+                        setSearchQuery('');
+                      }
+                    }}
+                  >
```
LetItRock: the height should be the same of the other inputs

https://github.com/user-attachments/assets/48eb8260-06b1-4dca-9743-9454b424e1f4

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+          if (!open) {
+            setTimeout(() => setSelectedVariable(null), 0);
+          }
+        }}
+      >
+        <PopoverTrigger asChild>
+          <div />
+        </PopoverTrigger>
+        {selectedVariable && (
+          <VariablePopover
```
LetItRock: the popover jumps, I feel it should be aligned with the start of the variable, but not sure if that is possible

https://github.com/user-attachments/assets/07d3cc9f-83ca-41a8-a332-3a2ddf44ff9e

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+import { Separator } from '../../separator';
+import { Select, SelectContent, SelectItem, SelectTrigger, SelectValue } from '@/components/primitives/select';
+import { TransformerItem } from './components/transformer-item';
+import { TransformerList } from './components/transformer-list';
+import { useVariableParser } from './hooks/use-variable-parser';
+import { useTransformerManager } from './hooks/use-transformer-manager';
+import { formatLiquidVariable } from './utils';
+import type { VariablePopoverProps } from './types';
+
+export function VariablePopover({ variable, onClose, onUpdate }: VariablePopoverProps) {
```
LetItRock: Screenshot 2024-12-31 at 10 03 20

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-subject.tsx
```diff
@@ -30,13 +23,13 @@ export const InAppSubject = () => {
         <InputField size="fit">
           <FormItem className="w-full">
             <FormControl>
-              <Editor
+              <FieldEditor
```
LetItRock: the font size is too big

Screenshot 2024-12-31 at 10 05 02

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
LetItRock: when the body has validation errors they are shown in the popover fields

https://github.com/user-attachments/assets/e657edc3-3132-44c7-a65c-53e997515848

LetItRock: the borders and headers of two seems to be a bit different

Screenshot 2024-12-31 at 10 05 40

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+                          </div>
+                        ) : (
+                          getFilteredTransformers(searchQuery).map((transformer) => (
+                            <SelectItem
+                              key={transformer.value}
+                              value={transformer.value}
+                              className="relative [&>*:first-child]:p-0"
+                            >
+                              <TransformerItem transformer={transformer} />
+                            </SelectItem>
```
LetItRock: the scroll does not work

https://github.com/user-attachments/assets/ae49681f-b87e-4090-b713-c205286851e8

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
LetItRock: on search it changes it's size

https://github.com/user-attachments/assets/c19534ce-5b20-4ae1-b811-92805b63ee02

LetItRock: when clicking outside of the modifiers it closes the variable popover 

https://github.com/user-attachments/assets/0b212216-8dd7-4ead-a8a0-9e9e47102a5b

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,257 @@
...
+        </div>
+        <div className="grid gap-1.5">
+          <div className="grid gap-2 p-2">
+            <FormItem>
+              <FormControl>
+                <div className="grid gap-1">
+                  <label className="text-text-sub text-label-xs">Variable name</label>
+                  <InputField size="fit" className="min-h-0">
+                    <Code2 className="text-text-sub h-4 w-4" />
+                    <Input value={name} onChange={(e) => handleNameChange(e.target.value)} className="h-7 text-sm" />
```
LetItRock: When there are modifiers and if the variable name gets removed all the modifiers appear in the editor.
I feel like we should only apply the changes when the popover is closed.

https://github.com/user-attachments/assets/33b2c496-ac7b-4c34-94a2-dfd25028513a

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-body.tsx
```diff
@@ -33,17 +22,16 @@ export const InAppBody = () => {
       render={({ field }) => (
         <FormItem className="w-full">
           <FormControl>
-            <InputField className="h-36 px-1">
-              <Editor
+            <InputField className="flex h-36 items-start">
+              <FieldEditor
```
LetItRock: weird focus behaviour when clicking inside/outside of the field

https://github.com/user-attachments/assets/e7c722a8-8cb5-4d72-ad64-7c2cc8e8508b

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,232 @@
...
+    isUpdatingRef,
+    onChange,
+    setSelectedVariable
+  );
+
+  const completionSource = useCompletionSource(variables, lastCompletionRef);
+
+  const extensions = useMemo(
+    () => [
+      autocompletion({
```
LetItRock: somehow now it's not closing the autosuggestions when clicking outside


https://github.com/user-attachments/assets/c22317cb-8b20-4e3e-9740-c02195cfebb7

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
LetItRock: react errors in the console when the popover is opened

Screenshot 2024-12-31 at 12 21 27

apps/dashboard/src/components/primitives/field-editor/variable-popover/hooks/use-variable-parser.ts
```diff
@@ -0,0 +1,72 @@
+import { useMemo } from 'react';
+import { TransformerWithParam } from '../types';
+import { TRANSFORMERS } from '../constants';
+
+export function useVariableParser(variable: string) {
```
LetItRock: Can't we use liquid library for all this parsing? 
I'm not sure if that's the right thing but seems like that https://liquidjs.com/api/classes/Tokenizer.html#constructor

LetItRock (changes requested):
learned a ton! 🤝 
there are a few suggestions for the improvement

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-pill-widget.ts
djabarovgeorge, line 1: should we educate the user that the order is important? 

image

apps/dashboard/src/components/workflow-editor/steps/email/email-subject.tsx
scopsy: The subject in email is in bigger font size compared to the rest of the form fields we have

apps/dashboard/src/utils/liquid-autocomplete.ts
scopsy: subscriber data is not supporting by API, but I will take a look at this

apps/dashboard/src/utils/liquid-autocomplete.ts
scopsy: Yep it works 💪

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
scopsy: Yep, it helps with the location of the popover relative to the editor. I want to try to make it more precise by attaching to the variable not, will see if it's feasible in a reasonable amount of time

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
scopsy: It's a tricky one, i gave up on this. Will make another try with a clean head :)

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-pill-widget.ts
scopsy: We do provide the drag and drop to reorder,  I will consider adding a tooltip there 💪

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
scopsy: Regarding the error state visible on all field it happens because of the InputField, I will refactor this to the new stateless inputs in the Design System PR

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
scopsy: I think I will leave it for now as it's seem to be not a critical scenario,

apps/dashboard/src/components/workflow-editor/steps/in-app/in-app-body.tsx
scopsy: Will be handleed by the aligned inputs PR

apps/dashboard/src/components/primitives/field-editor/variable-plugin/constants.ts
scopsy: Agree, But they are used across different files, and I didn't want to end up in a weird cyclical dependencies scenario.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/plugin-view.ts
scopsy: Unfortuently couldn't find on how we can achieve this, spent quite a lot of time on this. Will leave it as-is for now

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
scopsy: There is a small delay when the parent sheet is closing, which annoyed me, but it's not significant tho. Will remove it.

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
scopsy: Refactored this to use a command component

scopsy:
@SokratisVidros I've improved some useMemo usecases, but other than that I couldn't notice any significant degradation compared to staging or prod. Let me know if you have some specific scenarios we might want to optimize.

apps/api/src/app/workflows-v2/shared/map-step-type-to-result.mapper.ts
```diff
@@ -31,9 +31,13 @@ function buildDigestResult(payloadSchema?: JSONSchema) {
             time: {
               type: 'string',
             },
-            payload: payloadSchema || {
-              type: 'object',
-            },
+            payload:
```
SokratisVidros, line 34: Was this a bug on digest? Why additional properties are required?

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { Popover, PopoverTrigger } from '@/components/primitives/popover';
+import { createAutocompleteSource } from '@/utils/liquid-autocomplete';
+import { LiquidVariable } from '@/utils/parseStepVariablesToLiquidVariables';
+import { useCallback, useMemo, useRef } from 'react';
+import { useVariables } from './hooks/use-variables';
+import { createVariablePlugin } from './variable-plugin';
+import { variablePillTheme } from './variable-plugin/variable-theme';
+import { VariablePopover } from './variable-popover';
+
+type CompletionRange = {
```
SokratisVidros: Nitpicky one; it isn't very intuitive to define a type as null. The CompletionRange type variable, can be of that type or null.

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+type FieldEditorProps = {
+  value: string;
+  onChange: (value: string) => void;
+  variables: LiquidVariable[];
+  placeholder?: string;
+  autoFocus?: boolean;
+  size?: 'default' | 'lg';
+  id?: string;
+  singleLine?: boolean;
```
SokratisVidros: Nit picky DX, most of the time the FieldEditor is single line, so multiline is the exception. That is, we should revert the prop to `multiline` that defaults to false.

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,115 @@
+import { autocompletion } from '@codemirror/autocomplete';
```
SokratisVidros: If I am not mistaken, I wrote this comment again, but the field terminology is foreign to the Novu domain. This is about step controls, so the naming of such a pivotal component should reflect the domain. That is, we can call it ControlEditor or ControlInput

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
```diff
@@ -0,0 +1,115 @@
...
+      autocompletion({
+        override: [completionSource],
+        closeOnBlur: true,
+        defaultKeymap: true,
+        activateOnTyping: true,
+      }),
+    [completionSource]
+  );
+
+  const variablePlugin = useMemo(
```
SokratisVidros: Since everything else is an extension...

```suggestion
  const variableExtension = useMemo(
```

apps/dashboard/src/components/primitives/field-editor/hooks/use-variables.ts
```diff
@@ -0,0 +1,78 @@
+import { EditorView } from '@uiw/react-codemirror';
+import { useCallback, useRef, useState } from 'react';
+
+type SelectedVariable = {
```
SokratisVidros: Ditto about null.

apps/dashboard/src/components/primitives/field-editor/hooks/use-variables.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Manages variable selection and updates in the editor.
+ *
+ * This hook combines variable selection and update logic:
+ * 1. Tracks which variable is currently selected
+ * 2. Prevents recursive updates when variables are being modified
+ * 3. Handles proper Liquid syntax maintenance
+ * 4. Manages cursor position and editor state updates
+ */
+export function useVariables(viewRef: React.RefObject<EditorView>, onChange: (value: string) => void) {
+  const [selectedVariable, setSelectedVariable] = useState<SelectedVariable>(null);
```
SokratisVidros:
```suggestion
  const [selectedVariable, setSelectedVariable] = useState<SelectedVariable | null>(null);
```

apps/dashboard/src/components/primitives/field-editor/hooks/use-variables.ts
```diff
@@ -0,0 +1,78 @@
...
+    (newValue: string) => {
+      if (!selectedVariable || !viewRef.current || isUpdatingRef.current) return;
+
+      try {
+        isUpdatingRef.current = true;
+        const { from, to } = selectedVariable;
+        const view = viewRef.current;
+
+        // Ensure the new value has proper liquid syntax
+        const hasLiquidSyntax = newValue.match(/^\{\{.*\}\}$/);
```
SokratisVidros: The regex also matches variables with multiple brackets such as `{{{{foo}}}}` or `{{{{{{{foo}}}`. We should either consider it invalid or squash multiple {{{{ and }}} to two on each side.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/index.ts
```diff
@@ -0,0 +1,33 @@
+import { EditorView, ViewPlugin, Decoration } from '@uiw/react-codemirror';
+import type { PluginState } from './types';
+import { VariablePluginView } from './plugin-view';
+
+export function createVariablePlugin({ viewRef, lastCompletionRef, onSelect }: PluginState) {
```
SokratisVidros:
```suggestion
export function createVariableExtension({ viewRef, lastCompletionRef, onSelect }: PluginState) {
```

apps/dashboard/src/components/primitives/field-editor/variable-plugin/index.ts
```diff
@@ -0,0 +1,33 @@
...
+      provide: (plugin) =>
+        EditorView.atomicRanges.of((view) => {
+          return view.plugin(plugin)?.decorations || Decoration.none;
+        }),
+    }
+  );
+}
+
+export * from './types';
+export * from './constants';
```
SokratisVidros: Nit-picky one, I'd just move constants here to remove an extra file.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+export function parseVariable(match: RegExpExecArray): VariableMatch {
+  const start = match.index;
+  const end = start + match[0].length;
+  const fullVariableName = match[1].trim();
+  const parts = fullVariableName.split('|').map((part) => part.trim());
+  const variableName = parts[0];
+  const hasModifiers = parts.length > 1;
+
+  return {
+    fullVariableName,
```
SokratisVidros:
```suggestion
    expression,
```

apps/dashboard/src/components/primitives/field-editor/variable-plugin/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+    variableName,
```
SokratisVidros:
```suggestion
    name,
```

apps/dashboard/src/components/primitives/field-editor/variable-plugin/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+    start,
+    end,
+    hasModifiers,
```
SokratisVidros: Let's return all the liquid filters in an array instead of a boolean. I am sure we will use them later. Also since this is about Liquid Filters, we should rename the variable.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+  };
+}
+
+/**
+ * Handles backspace behavior for variables in the editor.
+ * When backspace is pressed near or within a variable, it removes the entire variable including its brackets.
+ */
+export function handleVariableBackspace(view: EditorView, pos: number, content: string): boolean {
```
SokratisVidros: :bulp If I get it right, this ensures the pill is treated as a single character during keyboard interactions. Code mirror uses content-editable divs. So, to create pills that behave as a single character inside a content-editable div, there is a technique to nest non-content-editable spans.

See [here](https://jsbin.com/mohuyokena/edit?html,output). Since we fully control the rendering process with the view plugin, maybe it provides a non-JS solution to the problem.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-pill-widget.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Determines if two VariablePillWidget instances are equal by comparing all their properties.
+   * Used by CodeMirror to optimize re-rendering.
+   */
+  eq(other: VariablePillWidget) {
+    return (
+      other.variableName === this.variableName &&
+      other.fullVariableName === this.fullVariableName &&
```
SokratisVidros: The full expression of the variable along with the start and end position should be enough for the comparison.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
```diff
@@ -0,0 +1,61 @@
...
+    color: 'hsl(var(--text-sub))',
+    border: '1px solid hsl(var(--stroke-soft))',
+    borderRadius: '10px',
+    padding: '2px 6px 2px 6px',
+    fontFamily: 'inherit',
+    display: 'inline-flex',
+    alignItems: 'center',
+    height: '16px',
+    lineHeight: '16px',
+    fontSize: '12px',
```
SokratisVidros: Do we have font and typography-related CSS root variables to use as we did with colors?

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
```diff
@@ -0,0 +1,61 @@
...
+  },
+  '.cm-variable-pill::before': {
+    content: '""',
+    left: '4px',
+    width: '12px',
+    minWidth: '12px',
+    height: '12px',
+    marginRight: '3px',
+    backgroundColor: 'hsl(var(--feature-base))',
+    maskImage: `url("/images/code.svg")`,
```
SokratisVidros: If not mistaken we can do `content: url("/images/code.svg");`. Is the maskImage required for anything else? If not, let's simplify it.

apps/dashboard/src/components/primitives/field-editor/variable-popover/components/draggable-transformer.tsx
```diff
@@ -0,0 +1,131 @@
...
+  draggingItem: number | null;
+  dragOverIndex: number | null;
+  onDragStart: (index: number) => void;
+  onDragEnd: () => void;
+  onDrag: (e: any, info: any) => void;
+  onRemove: (value: string) => void;
+  onParamChange: (index: number, params: string[]) => void;
+};
+
+export const DraggableTransformer = forwardRef<HTMLDivElement, DraggableTransformerProps>(
```
SokratisVidros: This component deserves being a top level primitive called Sortable, used from the variable popover, to drag and sort Liquid filters.

apps/dashboard/src/components/primitives/field-editor/variable-popover/components/transformer-item.tsx
```diff
@@ -0,0 +1,27 @@
+import TruncatedText from '@/components/truncated-text';
+import { Transformer } from '../types';
+
+type TransformerItemProps = {
+  transformer: Transformer;
+};
+
+export function TransformerItem({ transformer }: TransformerItemProps) {
```
SokratisVidros: As per Liquid.js jargon and [Novu Lexicon](https://www.notion.so/novuhq/Product-Terminology-Lexicon-d6ef923953744bc1b967ba8f297dee7f?pvs=4#15bd03ad838f8070a697c74fa7f8efa7) these are Liquid Filters, not transformers.

I strongly advice aligning the terminology across the board to easy the future reader.

apps/dashboard/src/components/primitives/field-editor/variable-popover/components/transformer-list.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onDragEnd,
+  onDrag,
+  onRemove,
+  onParamChange,
+}: TransformerListProps) {
+  if (transformers.length === 0) return null;
+
+  return (
+    <div className="rounded-8 border-stroke-soft flex flex-col gap-0.5 border px-1 py-1.5">
+      <AnimatePresence mode="popLayout">
```
SokratisVidros: Ditto about the sortable primitive.

apps/dashboard/src/components/primitives/field-editor/variable-popover/hooks/use-transformer-manager.ts
```diff
@@ -0,0 +1,113 @@
+import { useCallback, useState } from 'react';
+import { TRANSFORMERS } from '../constants';
+import type { TransformerWithParam } from '../types';
+
+type UseTransformerManagerProps = {
+  initialTransformers: TransformerWithParam[];
+  onUpdate: (transformers: TransformerWithParam[]) => void;
+};
+
+export function useTransformerManager({ initialTransformers, onUpdate }: UseTransformerManagerProps) {
```
SokratisVidros: It feels that these hooks belong to the sortable component and they need to be generalized to control the sortable state.

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  const handleDefaultValueChange = useCallback(
+    (newDefaultVal: string) => {
+      setDefaultVal(newDefaultVal);
+      debouncedUpdate(name, newDefaultVal, transformers);
+    },
+    [name, transformers, debouncedUpdate]
+  );
+
+  const handleRawLiquidChange = useCallback((value: string) => {
```
SokratisVidros: Shouldn't this code be part of the useVariableParser logic? Ideally, we should centralize all the logic about variable parsing and formatting to a single file, so that if we need to modify or expand the syntax, we do it in one place. Currently the logic is scattered across files.

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+          <div className="flex flex-col gap-1 p-2">
+            <FormItem>
+              <FormControl>
+                <div className="grid gap-1">
+                  <label className="text-text-sub text-label-xs">Modifiers</label>
+                  <Popover open={isCommandOpen} onOpenChange={setIsCommandOpen}>
+                    <PopoverTrigger asChild>
+                      <button className="text-text-soft bg-background flex h-[30px] w-full items-center justify-between rounded-md border px-2 text-sm">
+                        <span>Add a modifier...</span>
```
SokratisVidros: I suggest we call them filters in the UI as well. Since we will talk about Novu using Liquid.js we need to stick to the terminology of the template engine.

apps/dashboard/src/components/primitives/field-editor/variable-popover/variable-popover.tsx
```diff
@@ -0,0 +1,263 @@
...
+                          )}
+                        </CommandList>
+                      </Command>
+                    </PopoverContent>
+                  </Popover>
+                </div>
+              </FormControl>
+            </FormItem>
+
+            <TransformerList
```
SokratisVidros:
```suggestion
            <FilterSelector
```

SokratisVidros (changes requested):

apps/api/src/app/workflows-v2/shared/map-step-type-to-result.mapper.ts
scopsy: Yes, it was not allowed to add new variables when used with the auto complete flow. cc @djabarovgeorge

apps/dashboard/src/components/primitives/field-editor/field-editor.tsx
scopsy: I aggree! Done.

apps/dashboard/src/components/primitives/field-editor/variable-plugin/variable-theme.ts
scopsy: nope, currently all of the definitions are managed inside of the tailwind config file only. I will take a look at this at later stage as part of the new tailwind version upgrade, where they are moving away from the tailwind config file into a CSS variable native approach.

apps/dashboard/src/components/primitives/field-editor/variable-popover/components/draggable-transformer.tsx
scopsy: I have left it out intentionally for the time being, I'm not sure entirely where and how this is going to be used in other places, and would prefer to extract it when needed as it's not a primitive on our design system yet.

SokratisVidros (approved):

PR #7389, opened 2024-12-26
feat(api-service): polish action steps

djabarovgeorge:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/step-utils.ts
+++ b/apps/dashboard/src/components/workflow-editor/step-utils.ts
@@ -1,14 +1,23 @@
 import { flatten } from 'flat';
 import type {
   ContentIssue,
+  StepCreateDto,
   StepIssuesDto,
-  StepTypeEnum,
   StepUpdateDto,
   UpdateWorkflowDto,
   WorkflowResponseDto,
 } from '@novu/shared';
-import { Step } from '@/utils/types';
-import { STEP_TYPE_LABELS } from '@/utils/constants';
+import { StepTypeEnum } from '@novu/shared';
+import {
+  DEFAULT_CONTROL_DELAY_AMOUNT,
+  DEFAULT_CONTROL_DELAY_TYPE,
+  DEFAULT_CONTROL_DELAY_UNIT,
+  DEFAULT_CONTROL_DIGEST_AMOUNT,
+  DEFAULT_CONTROL_DIGEST_CRON,
+  DEFAULT_CONTROL_DIGEST_DIGEST_KEY,
+  DEFAULT_CONTROL_DIGEST_UNIT,
+  STEP_TYPE_LABELS,
+} from '@/utils/constants';
 
 export const getFirstBodyErrorMessage = (issues?: StepIssuesDto) => {
   const stepIssuesArray = Object.entries({ ...issues?.body });
@@ -58,10 +67,25 @@ export const updateStepInWorkflow = (
   };
 };
 
-export const createStep = (type: StepTypeEnum): Step => ({
-  name: STEP_TYPE_LABELS[type] + ' Step',
-  stepId: '',
-  slug: '_st_',
-  type,
-  _id: crypto.randomUUID(),
-});
+export const createStep = (type: StepTypeEnum): StepCreateDto => {
+  const controlValue: Record<string, unknown> = {};
+
+  if (type === StepTypeEnum.DIGEST) {
+    controlValue.amount = DEFAULT_CONTROL_DIGEST_AMOUNT;
+    controlValue.unit = DEFAULT_CONTROL_DIGEST_UNIT;
+    controlValue.digestKey = DEFAULT_CONTROL_DIGEST_DIGEST_KEY;
+    controlValue.cron = DEFAULT_CONTROL_DIGEST_CRON;
+  }
+
+  if (type === StepTypeEnum.DELAY) {
+    controlValue.amount = DEFAULT_CONTROL_DELAY_AMOUNT;
+    controlValue.unit = DEFAULT_CONTROL_DELAY_UNIT;
+    controlValue.type = DEFAULT_CONTROL_DELAY_TYPE;
+  }
+
+  return {
+    name: STEP_TYPE_LABELS[type] + ' Step',
+    type,
+    controlValues: controlValue,
+  };
+};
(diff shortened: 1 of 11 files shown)


apps/dashboard/src/components/workflow-editor/step-utils.ts
```diff
@@ -58,10 +67,25 @@ export const updateStepInWorkflow = (
...
+    controlValue.unit = DEFAULT_CONTROL_DELAY_UNIT;
+    controlValue.type = DEFAULT_CONTROL_DELAY_TYPE;
+  }
+
+  return {
+    name: STEP_TYPE_LABELS[type] + ' Step',
+    type,
+    controlValues: controlValue,
+  };
+};
```
djabarovgeorge, line 91: The summary of the findings of the best approach, why the change is necessary, and what is happening in the current implementation.

The main reason for this change is that we cannot create a step with control values during step creation. This limitation causes unexpected behavior, like the [bug](https://linear.app/novu/issue/NV-5124/dashboard-save-control-value-on-add-step) we are experiencing.  

Currently, we cannot move schemas with default values to the `Shared` package and reuse them in the `Dashboard` because these schemas depend on `Zod` and `Zod-to-JSON-Schema`.  

The solution we originally wanted—to move default values from the `UiSchema` to the `dataSchema`—is not supported natively. The solution should be `Zod` schema with default values (used by the `Dashboard`) and make certain attributes, like `body`, required for generating `issues`. However, this approach fails because `zod-to-json-schema` removes the required attributes if there is a default value meaning we will need to manipulate the schema manually.  

This solution is reducing the amount of code. In addition once we store the default values during step creation,  we wouldn't need to handle parsing them (placeholders/defaults) in the `Dashboard` or maintain "placeholders" or default values in the API.

**Next Steps**  
- Remove placeholders from the API.  
- Eliminate placeholder parsing in the `Dashboard`.

libs/application-generic/src/utils/sanitize-control-values.ts
```diff
@@ -222,37 +237,63 @@ export function dashboardSanitizeControlValues(
...
+        break;
+      default:
+        normalizedValues = filterNullishValues(controlValues);
+    }
+
+    console.log('normalizedValues 333222 ', normalizedValues);
+
+    return normalizedValues;
+  } catch (error) {
+    console.error('Error sanitizing control values', error);
```
djabarovgeorge: i made this sanitization more safe, i think it will be a good idea to pass here a logger and print Errors as well.

libs/application-generic/src/utils/sanitize-control-values.ts
```diff
@@ -177,15 +176,31 @@ function sanitizeDigest(controlValues: DigestControlSchemaType) {
...
+    ...(parseAmount(controlValues) as { amount?: number }), // TODO FIX THIS - should be controlValues.amount
+    type: controlValues.type,
+    unit: controlValues.unit,
     skip: controlValues.skip || undefined,
   };
 
   return filterNullishValues(mappedValues);
 }
 
+function parseAmount(amount?: unknown) {
```
djabarovgeorge, line 196: safe parse just in case

libs/application-generic/src/utils/sanitize-control-values.ts
```diff
@@ -177,15 +176,31 @@ function sanitizeDigest(controlValues: DigestControlSchemaType) {
 
 function sanitizeDelay(controlValues: DelayControlType) {
   const mappedValues: DelayControlType = {
-    type: controlValues.type || 'regular',
-    amount: controlValues.amount || 0,
-    unit: controlValues.unit || TimeUnitEnum.SECONDS,
+    // Cast to trigger Ajv validation errors - possible undefined
+    ...(parseAmount(controlValues) as { amount?: number }), // TODO FIX THIS - should be controlValues.amount
```
djabarovgeorge: self note: execute the todo

libs/application-generic/src/usecases/tier-restrictions-validate/tier-restrictions-validate.command.ts
```diff
@@ -5,9 +5,9 @@ import { OrganizationLevelCommand } from '../../commands';
 
 export class TierRestrictionsValidateCommand extends OrganizationLevelCommand {
   @IsOptional()
-  @Transform(({ value }) => (value ? String(value) : value))
-  @IsString()
-  amount?: string;
+  @Transform(({ value }) => (value ? Number(value) : value))
+  @IsNumber()
+  amount?: number;
```
djabarovgeorge, line 10: this should be a number as we use it as number in the usecase

apps/dashboard/src/components/workflow-editor/step-utils.ts
```diff
@@ -58,10 +67,25 @@ export const updateStepInWorkflow = (
...
 };
 
-export const createStep = (type: StepTypeEnum): Step => ({
-  name: STEP_TYPE_LABELS[type] + ' Step',
-  stepId: '',
-  slug: '_st_',
-  type,
-  _id: crypto.randomUUID(),
-});
+export const createStep = (type: StepTypeEnum): StepCreateDto => {
```
djabarovgeorge, line 70: we used the wrong type here, this is why we passed redundant data to the API such as slug prefix

libs/application-generic/src/utils/sanitize-control-values.ts
```diff
@@ -222,37 +237,63 @@ export function dashboardSanitizeControlValues(
...
-      normalizedValues = sanitizeDelay(controlValues as DelayControlType);
-      break;
-    default:
-      normalizedValues = filterNullishValues(controlValues);
+  try {
+    if (!controlValues) {
+      return null;
+    }
+
+    console.log('controlValues 333222 ', controlValues);
```
scopsy:
```suggestion
```

libs/application-generic/src/utils/sanitize-control-values.ts
```diff
@@ -222,37 +237,63 @@ export function dashboardSanitizeControlValues(
...
+        );
+        break;
+      case StepTypeEnum.DELAY:
+        normalizedValues = sanitizeDelay(controlValues as DelayControlType);
+        break;
+      default:
+        normalizedValues = filterNullishValues(controlValues);
+    }
+
+    console.log('normalizedValues 333222 ', normalizedValues);
```
scopsy:
```suggestion
```

libs/application-generic/src/utils/sanitize-control-values.ts
scopsy: Why not logger now?

scopsy (approved):

libs/application-generic/src/utils/sanitize-control-values.ts
djabarovgeorge: 🤦‍♂️ thanks!

libs/application-generic/src/utils/sanitize-control-values.ts
djabarovgeorge: done

PR #7396, opened 2024-12-27
fix(js): Inbox DX fixes

SokratisVidros:

Update naming for improved DX. See commits for more details.

### DIFF
--- a/packages/js/src/preferences/preference.ts
+++ b/packages/js/src/preferences/preference.ts
@@ -1,9 +1,10 @@
 import { InboxService } from '../api';
 
 import { NovuEventEmitter } from '../event-emitter';
-import { ChannelPreference, PreferenceLevel, Result, Workflow } from '../types';
+import { ChannelPreference, PreferenceLevel, Result, Workflow, Prettify } from '../types';
 import { updatePreference } from './helpers';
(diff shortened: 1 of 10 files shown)


packages/js/src/preferences/preference.ts
```diff
@@ -42,15 +43,19 @@ export class Preference {
     this.workflow = preference.workflow;
   }
 
-  update({ channelPreferences }: { channelPreferences: ChannelPreference }): Result<Preference> {
+  update({
+    channels,
+    // @deprecated use channels instead
+    channelPreferences,
```
scopsy, line 49: We are already doing a breaking change for the apiUrl, so I think we can just continue 😅 Just need to make sure to tell about this on release notes and changelog

scopsy: Oh nvm, just noticed it was on the httpclient only

scopsy (approved):

packages/js/src/preferences/preference.ts
LetItRock: We are not doing a breaking change as we have deprecated the old `backendUrl` and introduced the `apiUrl` on the `NovuOptions`. Similarly, on the update preferences args.

PR #7399, opened 2024-12-27
fix(application-generic): Bump got

SokratisVidros:

Update Got to its latest version to get all fixes and use its timing reporting so as to investigate timeouts with Bridge endpoints


SokratisVidros:
Got will be bumped as part of the #7424

PR #7400, opened 2024-12-27
fix(api-service): change webhook URL property to optional 

m1heng:

make webhook in UpdateSubscriberChannelRequestDto optional so that align to the definition in UpdateSubscriberChannelCommand `libs/application-generic/src/usecases/subscribers/update-subscriber-channel/update-subscriber-channel.command.ts`


scopsy:
Was already fixed on next 🙏 thank you!

PR #7401, opened 2024-12-27
chore(root): polish after review

djabarovgeorge:
### Screenshots

<details>

</details>


packages/shared/src/dto/workflows/step.dto.ts
```diff
@@ -2,7 +2,7 @@ import type { JSONSchemaDto } from './json-schema-dto';
 import { Slug, StepTypeEnum, WorkflowOriginEnum } from '../../types';
 import { StepContentIssueEnum, StepIssueEnum } from './step-content-issue.enum';
 
-export type StepDataDto = {
+export type StepResponseDto = {
```
djabarovgeorge, line 5: if we used this param as request DTO, my guess is that it is not self-explanatory enough

package.json
```diff
@@ -72,7 +72,8 @@
     "start": "npm run jarvis",
     "symlink:submodules": "pnpm --filter \"@novu/ee-*\" exec node \"$(pwd)/scripts/symlink-ee.mjs\"",
     "test:providers": "cross-env pnpm --filter './packages/providers/**' test",
-    "test": "cross-env CI=true lerna run test:watch --parallel"
+    "test": "cross-env CI=true lerna run test:watch --parallel",
+    "start:dev:services": "concurrently \"pnpm start:api:dev | cat\" \"pnpm start:worker | cat\" \"pnpm start:dashboard | cat\" --names \"API,WORKER,DASHBOARD\" --prefix-colors \"blue,green,yellow\" --kill-others --restart-after-error"
```
djabarovgeorge: for local easy services spin

apps/dashboard/src/components/primitives/form/form.tsx
djabarovgeorge, line 1: i guess now we support link the the issue error message 🙃

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -184,7 +190,8 @@ function extractProps(template: any): { valid: boolean; props: string[]; error?:
     return {
       valid: false,
       props: [],
-      error: `Variables must include a namespace (e.g. payload.${validProps[0]})`,
+      // eslint-disable-next-line max-len
+      error: `Invalid variable name missing namespace. Variables must follow the ${DOT_ANNOTATION_MARKDOWN_LINK} (e.g. payload.${validProps[0]})`,
```
scopsy: Is this long text renders nicely in the ui?

apps/dashboard/src/components/primitives/form/form.tsx
```diff
@@ -120,6 +121,27 @@ const FormMessagePure = React.forwardRef<
...
+    const markdownLinkRegex = /\[([^\]]+)\]\(([^)"']+)(?:\s+"[^"]*")?\)/g;
+
+    const htmlContent = text.replace(markdownLinkRegex, (_, text, url) => {
+      const isValidUrl = /^(https?:\/\/|\/)[^\s<>{}\\^~[\]`]+$/i.test(url);
+      if (!isValidUrl) return text;
+
+      return `<a href="${url}" class="underline text-primary hover:text-primary/80" target="_blank" rel="noopener noreferrer nofollow">${text}</a>`;
+    });
+
+    return DOMPurify.sanitize(htmlContent, {
```
scopsy: I wonder what if we just use react markdown here? Might be a little overkill tho 🤔

scopsy (approved):

apps/dashboard/src/components/primitives/form/form.tsx
SokratisVidros: What is the use case for this? Is it rendering links in validation error messages? If that's the case, we don't need Markdown at all. 

```import React from 'react';

const ErrorMessage = ({ errorMessage }) => {
  // Regular expression to extract the URL
  const urlRegex = /(https?:\/\/[^\s]+)/g;

  // Split the message into parts: text before, the URL, and text after
  const parts = errorMessage.split(urlRegex);

  return (
    <div className="error-message">
      {parts.map((part, index) => {
        if (urlRegex.test(part)) {
          // Render the link
          return (
            <a key={index} href={part} target="_blank" rel="noopener noreferrer">
              {part}
            </a>
          );
        }
        // Render regular text
        return <span key={index}>{part}</span>;
      })}
    </div>
  );
};

export default ErrorMessage;```

File: package.json [hunk 5]
```diff
@@ -186,7 +187,8 @@
     "typescript": "5.6.2",
     "typescript-eslint": "^8.13.0",
     "wait-port": "^0.3.0",
-    "yargs": "^17.7.2"
+    "yargs": "^17.7.2",
+    "concurrently": "^8.2.2"
```
SokratisVidros: @djabarovgeorge Can we achieve the same parallel execution using NX, avoiding another dependency?

SokratisVidros (approved):

package.json
djabarovgeorge: I was not able to make it work with NX.

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
djabarovgeorge: image

package.json
SokratisVidros: It should be `pnpm nx run-many --targets=start --projects=@novu/api-service,@novu/worker,@novu/dashboard --parallel`.

See https://nx.dev/recipes/running-tasks/run-tasks-in-parallel. I don't think NX has the ability to set color per task yet thought but that's not too critical.

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
```diff
@@ -144,7 +144,10 @@ export class SyncToEnvironmentUseCase {
     });
   }
 
-  private buildStepCreateOrUpdateDto(step: StepDataDto, existingInternalId?: string): StepUpdateDto | StepCreateDto {
+  private buildStepCreateOrUpdateDto(
+    step: StepResponseDto,
+    existingInternalId?: string
+  ): StepUpdateDto | StepCreateDto {
```
LetItRock, line 150: it's weird now to see that some functions require `StepResponseDto`

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
```diff
@@ -162,7 +165,11 @@ function extractProps(template: any): { valid: boolean; props: string[]; error?:
    * Invalid: {{user.first name}} - postfix length would be 2 due to space
    */
   if (initial.postfix.length > 1) {
-    return { valid: false, props: [], error: 'Variables with spaces are not supported' };
+    return {
+      valid: false,
+      props: [],
+      error: `<span>Invalid variable name containing whitespaces. Variables must follow the ${DOT_ANNOTATION_MARKDOWN_HREF}</span>`,
```
LetItRock: I'm unsure about this approach to building HTML code on the API side. If this API is used by RN or any other platform different from the web, the error message would be useless. 
Instead, we should return the pure message to the Dashboard and have a hint or a link in the UI pointing to our documentation teaching how to use the editors.
cc @SokratisVidros

apps/dashboard/src/components/primitives/form/form.tsx
djabarovgeorge: this approach make the code a bit more complicated because you need to calculate that text is the `part` in `<a>part</a>`, therefore i suggested another approach in order to simplify it

package.json
djabarovgeorge: Yeah, but the color makes it easier to differentiate them, which helps with running all the services in a single terminal. I'll remove it for now since it's not a priority at the moment.

apps/api/src/app/workflows-v2/usecases/sync-to-environment/sync-to-environment.usecase.ts
djabarovgeorge: exactly why I renamed it.

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
djabarovgeorge: The assumption right now is that the only consumer of the new API is the Dashboard, this is why this approach felt appropriate to me.

package.json
djabarovgeorge: was removed

apps/api/src/app/workflows-v2/util/template-parser/liquid-parser.ts
SokratisVidros: I agree with Pawel. We also use react, so returning HTML from a JSON API will eventually bite us. 

See also this that is strictly relevant. The complexity for error rendering should be in the Dashboard UI. The message text is a the responsibility of the API.

djabarovgeorge: Ok so it is decided I will remove the link, but please take into consideration that providing links will become much more complex in the client.

PR #7406, opened 2024-12-30
feat(dashboard): Open step template drawer on Add Step

desiprisg:
To make sure that the user can focus on modifying the content immediately after adding a step and doesn't get distracted by errors for missing content, we are opening the configure drawer right after the step is added.

### Screenshots

https://github.com/user-attachments/assets/ccdc2a53-73ac-4800-9c2e-710ef04f6e21

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/workflow-editor/edges.tsx
+++ b/apps/dashboard/src/components/workflow-editor/edges.tsx
@@ -1,8 +1,11 @@
-import { BaseEdge, Edge, EdgeLabelRenderer, EdgeProps, getBezierPath } from '@xyflow/react';
-import { AddStepMenu } from './add-step-menu';
-import { useWorkflow } from '@/components/workflow-editor/workflow-provider';
-import { WorkflowOriginEnum } from '@novu/shared';
 import { createStep } from '@/components/workflow-editor/step-utils';
+import { useWorkflow } from '@/components/workflow-editor/workflow-provider';
+import { TEMPLATE_CONFIGURABLE_STEP_TYPES } from '@/utils/constants';
+import { buildRoute, ROUTES } from '@/utils/routes';
+import { WorkflowOriginEnum } from '@novu/shared';
+import { BaseEdge, Edge, EdgeLabelRenderer, EdgeProps, getBezierPath } from '@xyflow/react';
+import { useNavigate } from 'react-router-dom';
+import { AddStepMenu } from './add-step-menu';
 
 export type AddNodeEdgeType = Edge<{ isLast: boolean; addStepIndex: number }>;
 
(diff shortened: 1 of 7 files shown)


apps/dashboard/src/components/workflow-editor/edges.tsx
```diff
@@ -60,8 +63,18 @@ export function AddNodeEdge({
...
+                      data: {
+                        ...workflow,
+                        steps: updatedSteps,
+                      },
+                      onSuccess: (data) => {
+                        navigate(
+                          buildRoute(ROUTES.EDIT_STEP_TEMPLATE, {
+                            workflowSlug: workflow.slug,
+                            stepSlug: data.steps[indexToAdd].slug,
+                          })
```
scopsy: There is an issue here, when we add a non sidebar item like delay, it opens a random step email editor or custom controls

desiprisg: Hmm ur right. I'll filter those out.

apps/dashboard/src/utils/constants.ts
```diff
@@ -28,6 +28,14 @@ export const STEP_TYPE_LABELS: Record<StepTypeEnum, string> = {
   [StepTypeEnum.CUSTOM]: 'Custom',
 };
 
+export const AUTO_OPEN_DRAWER_AFTER_CREATION_STEP_TYPES: StepTypeEnum[] = [
```
SokratisVidros: Every TEMPLATE_CONFIGURABLE_STEP_TYPES should allow for the same behavior. So we don't need a separate enum for this.

Moreover, these enums have a bit of a code smell as the UI around drawers should work independently of the step types. We will get back to that soon; meanwhile, let's not add another constant array of array steps.

apps/dashboard/src/components/workflow-editor/workflow-provider.tsx
```diff
@@ -105,9 +111,13 @@ export const WorkflowProvider = ({ children }: { children: ReactNode }) => {
   const isUpdatePatchPending = isPatchPending || isUpdatePending || hasPendingItems;
 
   const update = useCallback(
-    (data: UpdateWorkflowDto) => {
+    ({ data, onSuccess }: { data: UpdateWorkflowDto; onSuccess?: (workflow: WorkflowResponseDto) => void }) => {
```
SokratisVidros: I think it would be better DX to do 
``` const update = useCallback(
    (data: UpdateWorkflowDto, { onSuccess }: { onSuccess?: (workflow: WorkflowResponseDto) => void }) => {```

It's also a safer change and doesn't introduce the super generic data key across the board.

SokratisVidros (changes requested):

SokratisVidros (approved):

PR #7407, opened 2024-12-30
fix(api-service): when step is deleted remove all associated variables

LetItRock:

When the step is removed delete the associated control values and as a result step variables.

### Screenshots

https://github.com/user-attachments/assets/7376a7d7-8722-4473-b09d-5c0d1bf7b985


apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
```diff
@@ -77,6 +77,7 @@ export class UpsertWorkflowUseCase {
   async execute(command: UpsertWorkflowCommand): Promise<WorkflowResponseDto> {
     const workflowForUpdate = await this.queryWorkflow(command);
     const persistedWorkflow = await this.createOrUpdateWorkflow(workflowForUpdate, command);
+    // TODO: this upsertControlValues logic should be moved to the create/update workflow usecase
```
LetItRock, line 77: IMO the logic to create/update the control values should be moved to the create/update workflow usecase, now it's disconnected from the place where we manage steps.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
```diff
@@ -753,6 +757,16 @@ export class UpdateWorkflow {
           workflowType: command.type,
         }),
       );
+
+      await this.deleteControlValuesUseCase.execute(
```
LetItRock: when the step is removed --> remove also the control values

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
SokratisVidros: In that case, won't we duplicate much of the control value logic? I guess we can get away with a shared use case about control values.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
SokratisVidros: Nit-picky one: The deleteControlValues use case is just a single delete call to a repository. The existence of a use case should be justified by the use case doing at least two things. So, we could replace it with a repository call and remove two extra files (use case & command)

SokratisVidros (approved):

apps/api/src/app/workflows-v2/usecases/upsert-workflow/upsert-workflow.usecase.ts
LetItRock: I'm not sure why you are saying that the logic will be duplicated. This is the only place where we create/update the control values, all I'm saying is to move this logic to the corresponding workflow usecase.

libs/application-generic/src/usecases/workflow/update-workflow/update-workflow.usecase.ts
LetItRock: I'm ok with that.

PR #7408, opened 2024-12-30
feat(dashboard): Make the whole workflow row clickable

desiprisg:
### Screenshots

<details>

</details>

### DIFF
--- a/apps/dashboard/src/components/workflow-row.tsx
+++ b/apps/dashboard/src/components/workflow-row.tsx
@@ -1,18 +1,4 @@
-import { useState } from 'react';
-import { FaCode } from 'react-icons/fa6';
-import {
-  RiDeleteBin2Line,
-  RiFileCopyLine,
-  RiFlashlightLine,
-  RiGitPullRequestFill,
-  RiMore2Fill,
-  RiPauseCircleLine,
-  RiPlayCircleLine,
-  RiPulseFill,
-} from 'react-icons/ri';
-import { Link } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


apps/dashboard/src/components/workflow-row.tsx
```diff
@@ -160,47 +180,43 @@ export const WorkflowRow = ({ workflow }: WorkflowRowProps) => {
   };
 
   return (
-    <TableRow key={workflow._id} className="relative">
+    <TableRow key={workflow._id} className="group relative isolate">
```
scopsy, line 183: What if we just do:

```
    <TableRow key={workflow._id} className="hover:bg-accent/50 relative hover:cursor-pointer" onClick={handleRowClick}>
```

On the main row, and add to the onHoverCopy and actions cell:
```
          onClick={(e) => e.stopPropagation()}
```

Might simplify this a bit, wdyt?

SokratisVidros: The implementation feels like an overkill, but if I am not mistaken, it's the only way to maintain valid Table/tr/td HTML while treating each row as a link, making right-clicking Open in a new Tab work.

@desiprisg Please confirm.

SokratisVidros (approved):

apps/dashboard/src/components/workflow-row.tsx
desiprisg: Yes that's the thinking behind this. Also checked with some other implementations like stripe's and they seem to follow the same approach.

